PR #1355, opened 2022-03-03
Updated h2h WS, Added support for ftp Multi Hit, Reworked cartio rang…

Kayanesa:
Reworked cratio cap for ws range. Added support for ftp on all hit of multi hit ws. Updated hand to hand weaponskill in accordance with new value.
Also change the Critical_Hit_Rate to follow curve instead of guesstimate

**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] that I agree to LandSandBoat's [Limited Contributor License Agreement](https://github.com/LandSandBoat/server/blob/base/.github/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [x] that I have **read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)**
- [x] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

PR #1380, opened 2022-03-07
[Dagger] Add fTP replication WS and Refresh

Kayanesa:
Adding fTP relpication to Dancing Edge, Evisceration, Exenterator and Pyrrhic Kleos. General refresh of the ws parameter where needed. Added mp drain and scale base on tp to Energy Steal and Energy Drain.

**_I affirm:_**

### DIFF
--- a/scripts/globals/weaponskills/cyclone.lua
+++ b/scripts/globals/weaponskills/cyclone.lua
@@ -28,8 +28,9 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     params.skill = xi.skill.DAGGER
     params.includemab = true
 
-    if (xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES == true) then
+    if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
         params.dex_wsc = 0.4 params.int_wsc = 0.4
+        params.ftp300 = 3.75 -- http://wiki.ffo.jp/html/685.html
     end
 
     local damage, criticalHit, tpHits, extraHits = doMagicWeaponskill(player, target, wsID, params, tp, action, primary)
--- a/scripts/globals/weaponskills/energy_drain.lua
+++ b/scripts/globals/weaponskills/energy_drain.lua
@@ -23,9 +23,11 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     end
 
     damagemod = damagemod * xi.settings.WEAPON_SKILL_POWER
-    local mpReturn = target:addMP(-damagemod)
-    player:addMP(mpReturn) -- this isnt the right way but the only way I could figure out
-    -- to do make undead undrainable
+
+    if not target:isUndead() then
+        local mpReturn = target:addMP(-damagemod)
+        player:addMP(mpReturn)
+    end
 
     return 1, 0, false, 0
 end
--- a/scripts/globals/weaponskills/energy_drain.lua
+++ b/scripts/globals/weaponskills/energy_drain.lua
@@ -23,11 +23,11 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     end
 
     damagemod = damagemod * xi.settings.WEAPON_SKILL_POWER
-    mpReturn = target:addMP(-damagemod)
+    local mpReturn = target:addMP(-damagemod)
     player:addMP(mpReturn) -- this isnt the right way but the only way I could figure out
     -- to do make undead undrainable
 
-    return 1, 0, false, damagemod
+    return 1, 0, false, 0
 end
 
 return weaponskill_object
--- a/scripts/globals/weaponskills/energy_drain.lua
+++ b/scripts/globals/weaponskills/energy_drain.lua
@@ -8,9 +8,9 @@ require("scripts/globals/weaponskills")
 local weaponskill_object = {}
 
 weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary, action, taChar)
-   -- TODO: Should not wake the mob involved. Not tested on sleeping mob
-
-    local damage = 60 -- Base mp absorbtion from http://wiki.ffo.jp/html/686.html saying it was between 60 and 120 and always higher than Energy Steal
+    -- TODO: Should not wake the mob involved. Not tested on sleeping mob. Need to add Darkness element and resistance. More testing needed to understand retail behavior. 
+    --[[
+    local damage = 60 -- Base mp absorbtion from http://wiki.ffo.jp/html/687.html saying it was between 60 and 120 and always higher than Energy Steal this is before adoulin era
 
     -- params.ftp100 = 1 params.ftp200 = 1.5 params.ftp300 = 2 -- to follow earlier statement of been always higher than Energy Steal
     -- params.mnd_wsc = 1
@@ -26,7 +26,7 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     if not target:isUndead() then
         player:addMP(target:addMP(-damagemod))
     end
-
+    --]]
     return 1, 0, false, 0
 end
 
--- a/scripts/globals/weaponskills/energy_drain.lua
+++ b/scripts/globals/weaponskills/energy_drain.lua
@@ -18,8 +18,7 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     local damagemod = damage * ((tp - 1000) / 2000 + 1) -- Simplified formula for getting difference in tp return from 1000 tp to 3000 tp
 
     if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
-        local attMind = player:getStat(xi.mod.MND)
-        damagemod = damagemod + attMind * 1
+        damagemod = damagemod + player:getStat(xi.mod.MND)
     end
 
     damagemod = damagemod * xi.settings.WEAPON_SKILL_POWER
--- a/scripts/globals/weaponskills/energy_drain.lua
+++ b/scripts/globals/weaponskills/energy_drain.lua
@@ -8,25 +8,30 @@ require("scripts/globals/weaponskills")
 local weaponskill_object = {}
 
 weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary, action, taChar)
-    -- TODO: Should not wake the mob involved. Not tested on sleeping mob. Need to add Darkness element and resistance. More testing needed to understand retail behavior. 
+    -- TODO: Should not wake the mob involved. Not tested on sleeping mob. Need to add Darkness element and resistance. More testing needed to understand retail behavior.
 
-    -- local damage = 60 -- Base mp absorbtion from http://wiki.ffo.jp/html/687.html saying it was between 60 and 120 and always higher than Energy Steal this is before adoulin era
+    --[[
+    local damage = 60 -- Base mp absorbtion from http://wiki.ffo.jp/html/687.html saying it was between 60 and 120 and always higher than Energy Steal this is before adoulin era
 
     -- params.ftp100 = 1 params.ftp200 = 1.5 params.ftp300 = 2 -- to follow earlier statement of been always higher than Energy Steal
     -- params.mnd_wsc = 1
 
-    -- local damagemod = damage * ((tp - 1000) / 2000 + 1) -- Simplified formula for getting difference in tp return from 1000 tp to 3000 tp
+    local damagemod = damage * ((tp - 1000) / 2000 + 1) -- Simplified formula for getting difference in tp return from 1000 tp to 3000 tp
 
-    -- if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
-    --     damagemod = damagemod + player:getStat(xi.mod.MND)
-    -- end
+    if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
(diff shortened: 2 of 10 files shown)


Kayanesa:
So I couldn't find a way to make Energy drain and Energy steal to drain mp from the weaponskills.lua mabe I missed it but I don't think the code to make it possible exist. So I did all the calculation inside the ws themself but I realize this is not standard. If this acceptable than great. But I'm open to any help/suggestion as to how to implement it correctly. Not sure where in the weaponskills.lua it should be taken into account.

WinterSolstice8:
Another thing that needs to be verified, which I can check for you later tonight for you is whether or not Energy Drain/Energy Steal work on mobs without MP due to neither their job/subjob not having MP. It's worth noting that if it's like Sanguine Blade, it may not be an actual "drain". Sanguine Blade can kill a mob with 10 hp left for 1000 damage and you still get 1000 HP. 

Probably, if the mob has >0 MP left it will work. That part is a bit more lengthy to verify... thankfully Aspir II/III is a thing now :)

scripts/globals/weaponskills/exenterator.lua
```diff
@@ -30,8 +30,15 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     params.acc100 = 0.0 params.acc200= 0.0 params.acc300= 0.0
     params.atk100 = 1.0; params.atk200 = 1.0; params.atk300 = 1.0
 
-    if (xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES == true) then
+    if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
         params.agi_wsc = 0.7 + (player:getMerit(xi.merit.EXENTERATOR) * 0.03)
+        params.ftp100 = 2.0 params.ftp200 = 2.0 params.ftp300 = 2.0 -- http://wiki.ffo.jp/html/25595.html
+    end
+
+    if xi.settings.USE_MULTI_HIT_FTP_WEAPON_SKILL_CHANGES then
```
claywar: This setting is undefined, and also appears to only be duplicating the `USE_ADOULIN_WEAPON_SKILL_CHANGES` values for everything in this PR.

Kayanesa: it's a setting that will be added if my other ws pr goes trough.

Kayanesa: should I duplicate the setting here it's just one line of code in the main.lua?

Kayanesa:
Actually I have tested the mp ability and id doesn't return mp if the mob as no mp to start with because of that line
local mpReturn = target:addMP(-damagemod)
with this mpReturn is only positive if the mob as mp to loose.

WinterSolstice8:
Did you check on an aspired mob or just a WAR/WAR etc?

Kayanesa:
If memory serve it was a cactuar and and antican but not sur what job they are

Kayanesa:
More testing definitely welcome

WinterSolstice8:
probably WARs and MNKs. I will check an aspired elemental for you in a few hours.

scripts/globals/weaponskills/exenterator.lua
Kayanesa: with the fTP relpication many ws ftp value became weaker like dragon kick 
from
2.00 | 3.875 | 7.00 | ×2 | STR50% VIT50%
to this
1.70 | 3.00 | 5.00 | ×2 | STR50% VIT50%
or 
exentrerator from
2.0
to 
1.1875

WinterSolstice8:
Thunder V to waste MP, Aspir 3 to show 0 MP, energy drain for...
image
Yup. Just needs MP > 0 to get any MP back, just like Aspir. Other than undead, of course.

scripts/globals/weaponskills/energy_drain.lua
```diff
@@ -1,11 +1,33 @@
...
+
+    if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
+        local attMind = player:getStat(xi.mod.MND)
+        damagemod = damagemod + attMind * 1
+    end
+
+    damagemod = damagemod * xi.settings.WEAPON_SKILL_POWER
+    local mpReturn = target:addMP(-damagemod)
+    player:addMP(mpReturn) -- this isnt the right way but the only way I could figure out
+    -- to do make undead undrainable
```
WinterSolstice8: Looks like the way to check if something is undead is with `target:isUndead()` , cannibal blade uses it for the same purpose but with drain

Kayanesa: I will add and test it a bit later thanks

Kayanesa: Added the check to undead and tested on bogy and that was just what I needed.

Kayanesa:
Unless claywar want me to do away with making the ftp replication optional and just add it to post adoulin then this should be ready but I will wait for issue #1355 to be done with before undrafting since this code is partially dependent on it.

scripts/globals/weaponskills/cyclone.lua
```diff
@@ -28,8 +28,9 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     params.skill = xi.skill.DAGGER
     params.includemab = true
 
-    if (xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES == true) then
+    if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
         params.dex_wsc = 0.4 params.int_wsc = 0.4
+        params.ftp300 = 3.75 -- http://wiki.ffo.jp/html/685.html
```
Xaver-DaRed, line 33: As per JP wiki, this is missing params.ftp200 = 2375

scripts/globals/weaponskills/exenterator.lua
Xaver-DaRed: I'm not too sure about this new setting, if I'm being honest.

Firstly we would have to determine what the setting is for. The adoulin setting being ON is meant for modern retail WS parameters. It being off is meant for older eras.

So if anything, I think the new setting should be removed, becouse either this parameters are the new ones, or they arent.

If they are the new ones, then we should be using the setting we already have.
If they are the old ones, then the old values not inside the adoulin setting should be the ones being edited

Kayanesa: Removing the setting is fine with me, I just saw the adoulin setting existed because of the big change to ws and ftp replication is another big change so I made it a choice. I just need to know what is best for this project so I can change it accordingly.

TeoTwawki: adoulin one is basically mis-named. it should say "modern" instead of "adoulin" and the aim is for an easy toggle for the old school crowd who wants 75 cap gameplay and the rest of us, who are matching modern retail (by default).

Kayanesa: New setting was removed.

scripts/globals/weaponskills/energy_drain.lua
```diff
@@ -1,11 +1,35 @@
...
+    local damage = 60 -- Base mp absorbtion from http://wiki.ffo.jp/html/686.html saying it was between 60 and 120 and always higher than Energy Steal
+
+    -- params.ftp100 = 1 params.ftp200 = 1.5 params.ftp300 = 2 -- to follow earlier statement of been always higher than Energy Steal
+    -- params.mnd_wsc = 1
+
+    local damagemod = damage * ((tp - 1000) / 2000 + 1) -- Simplified formula for getting difference in tp return from 1000 tp to 3000 tp
+
+    if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
+        local attMind = player:getStat(xi.mod.MND)
+        damagemod = damagemod + attMind * 1
```
Xaver-DaRed: Why are we multiplying by 1 here?

This 2 lines could be simplified to:
`damagemod = damagemod + player:getStat(xi.mod.MND)`

Also, why isnt this a params.mnd_wsc = 1 or something similar?

I ask becouse a rdm lvl 75 WILL have 80+ MND, witch will make it so he steals quite an amount of MP
Granted, idk how the param is actually handled in weaponskills.lua

scripts/globals/weaponskills/energy_drain.lua
```diff
@@ -1,11 +1,35 @@
...
+    end
+
+    damagemod = damagemod * xi.settings.WEAPON_SKILL_POWER
+
+    if not target:isUndead() then
+        local mpReturn = target:addMP(-damagemod)
+        player:addMP(mpReturn)
+    end
```
Xaver-DaRed, line 29: Couldnt this be simplified to:
```
    if not target:isUndead() then
        target:addMP(-damagemod)
        player:addMP(damagemod)
    end
```

Xaver-DaRed (changes requested):
Same comments found in energy_drain.lua can apply to energy_steal.lua

Also, I'd be interested in knowing why this isnt using
`doPhysicalWeaponskill`
or
`doMagicalWeaponskill`
Or some other function. (I dont know much about WS, hence my question)
I ask becouse this doesnt seem to take into account things like darkness resistances or SDT

scripts/globals/weaponskills/energy_drain.lua
WinterSolstice8:
```
int32 CBattleEntity::addMP(int32 mp)
{
    int32 cap = std::clamp(health.mp + mp, 0, GetMaxMP());
    mp        = health.mp - cap;
    health.mp = cap;
    if (mp != 0)
    {
        updatemask |= UPDATE_HP;
    }
    return abs(mp);
}
```

It's done like that to prevent giving more MP than the mob has left

TeoTwawki: now here I would have just done a check for mp with the undead check, or [clamped](https://github.com/LandSandBoat/server/blob/base/scripts/globals/utils.lua#L70) it depending if we needed mobs mp to just be nonzero or we needed the aspired amount to not be bigger than the remaining amount (its been years since I've had to care how energy drain worked lol) and aspir the spell is set up this way https://github.com/LandSandBoat/server/blob/base/scripts/globals/spells/black/aspir.lua#L45

¯\\\_(ツ)_/¯

scripts/globals/weaponskills/energy_drain.lua
Kayanesa: I might have removed some of my explaination while cleaning up.
I couldn't find the code that would let me make this mp drain ws work from the weaponskills.lua and I don't know where in the weaponskills.lua it should be added so I did all the relevent calculation from the ws file. This ws does no damage and only drain mp from mob with mp that aren't undead. The time 1 is really not useful it's a place older in the formula to implement the ws with the commented out parameter params.mnd_wsc = 1.

scripts/globals/weaponskills/energy_drain.lua
Kayanesa: In this case the mp shouldn't be more than what the target as left.

scripts/globals/weaponskills/energy_drain.lua
Kayanesa: there is no need to pass it to a variable at that point so I can remove it.

Kayanesa: Your point about resistance is a good question. While I'm not 100% sure I believe it is a straight mp steal that can not be resisted. The description on https://ffxiclopedia.fandom.com/wiki/Energy_Drain does specify that it's not align with elemental gorgets or elemental belts and does not have skill chain property. But it does have a darkness element.

scripts/globals/weaponskills/energy_drain.lua
Kayanesa: I could change it to this 
player:addMP(target:addMP(-damagemod))
but I would need to test it since I don't know if this is actually proper coding and possible to do. I can't test right away but will later.

Kayanesa: After testing player:addMP(target:addMP(-damagemod)) still work as intended and lua as no problem with the syntax.
This is as simplified as I can make it.

WinterSolstice8:
I think these changes for WS make sense. Though because it seems (collectively) we know nothing about Energy Drain/Steal, maybe those should be commented out for a later time when we understand it more.  It seems it does drain a static amount of MP with the same TP/gear/target, though. 

image (I had to use a TON of MP+ gear to record this result)

If you want to put in some time to figure out the modifiers it should be pretty simple to test. SAM main job lets you get 1000 TP exactly with meditate, so you should be able to use meditate once to figure it out. Most magic WS deal static damage against the same target at the same level (use battlemod to see a target's level with /check for non NMs)

Kayanesa:
That is a lot of mp return is this from the server or from retail?

WinterSolstice8:
Retail, 3k TP, BLM main job with ilvl 119 gear so a ton of MND

Kayanesa:
Could you tell me the mind number that would help me know if the 2 time + mind modifier looks good or not. ^^

Kayanesa:
Right now you would need 496 mind to make this happen

WinterSolstice8:
I'll get that later today.

WinterSolstice8:
206 MND, it looks like. Don't forget there's probably a dMND/dINT bonus

Kayanesa:
Thanks for the help. Since I do not know how to add elemental resistance and the numbers need more testing as they don't match up I disabled the ws for now so the push for the other ws tpreplication can happen. I will open an issue on those two ws so they have a chance of getting fixed.

WinterSolstice8 (approved):
The changes check out to me. Cyclone is the only one that wasn't researched by NA in this PR.

Xaver-DaRed (approved):

PR #1381, opened 2022-03-07
Update 'UNKNOWN' RELEASE_TYPE name to 'PLAYERINPUT'.

atom0s:
**_I affirm:_**

`RELEASE_TYPE::UNKNOWN` is used during events where the user is prompted to input a password. (ie. Castle Oztroja furnace, Strange Apparatus', etc.) It is also used when prompting the player for numerical input, such as returning seals from Shami in Port Jeuno and similar.

When the client inputs a password, it sends a `0x60` packet and sets an internal flag `RecPassWdWindowFlag` waiting for the server to respond. When the server responds, the flag is cleared and the prompt is closed.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -3658,7 +3658,7 @@ void SmallPacket0x060(map_session_data_t* const PSession, CCharEntity* const PCh
     luautils::OnEventUpdate(PChar, updateString);
 
     PChar->pushPacket(new CReleasePacket(PChar, RELEASE_TYPE::EVENT));
-    PChar->pushPacket(new CReleasePacket(PChar, RELEASE_TYPE::UNKNOWN));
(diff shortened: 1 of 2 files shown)


claywar (approved):

atom0s:
Looking more into this tonight, it looks like this release mode is used for more than just password input. It's for any input expected from the player.

String input is expected for passwords from events like the Castle Ozstroja Brass Furnace, however, there is also numerical inputs that use this as well such as the Magian Moogle's storage with the crate to pull items back out, as well as asking for quantities from the player with storage things like Shami's seal storage in Port Jeuno.

Might be better to name this `RELEASE_TYPE::PLAYERINPUT` instead.

claywar:
I can get onboard with `PLAYERINPUT`; we're at a naming convention at this point.  Player needs to type something that's not a predetermined response.

zach2good (approved):

PR #1388, opened 2022-03-08
[RUN] Implement Vallation, Valiance, Battuta, Inspiration

WinterSolstice8:
Fairly meaty again, started to go too far so cut new features at this line. The result of this PR will help shape the future for the complication that will be Effusion, for sure.

**_I affirm:_**
- [X] that I agree to LandSandBoat's [Limited Contributor License Agreement](https://github.com/LandSandBoat/server/blob/base/.github/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [X] that I have **read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)**
- [X] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

### DIFF
--- a/scripts/globals/abilities/battuta.lua
+++ b/scripts/globals/abilities/battuta.lua
@@ -1,0 +1,21 @@
+-----------------------------------
+-- Ability: Battuta
+-- Increases the likelihood of parrying and deals counter damage after parrying dependent upon harbored runes.
+-- Obtained: Rune Fencer Level 75 (merit)
+-- Recast Time: 5:00
+-- Duration: 1:30
+-----------------------------------
+require("scripts/globals/status")
+require("scripts/globals/job_utils/rune_fencer")
+-----------------------------------
+local ability_object = {}
+
+ability_object.onAbilityCheck = function(player, target, ability)
+    return xi.job_utils.rune_fencer.checkHaveRunes(player)
+end
+
+ability_object.onUseAbility = function(player, target, ability, action)
+    return xi.job_utils.rune_fencer.useBattuta(player, target, ability, action)
+end
+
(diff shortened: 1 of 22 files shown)


src/map/modifier.h
```diff
@@ -876,8 +879,8 @@ enum class Mod
     // 261 to 287
     // 888
     // 936
-    // 
-    // SPARE = 1021, and onward
+    //
+    // SPARE = 1023, and onward
```
Xaver-DaRed: 1024

WinterSolstice8: counting is hard :)

zach2good:
Seems fine to me, but since we're targeting a feature branch at the moment it might be a safer idea to PR those additional modifiers to `base`, to save on any possible heartbreak later

WinterSolstice8:
> Seems fine to me, but since we're targeting a feature branch at the moment it might be a safer idea to PR those additional modifiers to `base`, to save on any possible heartbreak later

created #1394 to address this

WinterSolstice8:
Here we go again with rebasing doing things I wasn't expecting by adding so many commits. I need to understand why this keeps happening -- at least I know how to fix this particular issue.

WinterSolstice8:
Looks like some commits went missing from feature/runefencer? As a bonus I figured out how to rebase properly

zach2good:
Yeah, I think I screwed up when I was rebasing :( I can try and fix it after work, or you can grab the missing commits as part of this branch and we'll fix things that way (those commits will still exist/be visible through the previous PRs you've done)

zach2good (approved):

PR #1392, opened 2022-03-09
[Katana] ftp replication and refresh

Kayanesa:
Added ftp replication to katana ws Blade: Jin, Blade: Ku, Blade: Shun.
Refresh of ws parameter.

**_I affirm:_**

### DIFF
--- a/scripts/globals/weaponskills/blade_kamu.lua
+++ b/scripts/globals/weaponskills/blade_kamu.lua
@@ -32,7 +32,11 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
 
     if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
         params.str_wsc = 0.6 params.int_wsc = 0.6
-        params.atk100 = 3; params.atk200 = 3; params.atk300 = 3 -- http://wiki.ffo.jp/html/15893.html
+        params.ignoresDef = true
+        params.ignored100 = 0.25
+        params.ignored200 = 0.25
+        params.ignored300 = 0.25
+        params.atk100 = 2.25; params.atk200 = 2.25; params.atk300 = 2.25 -- http://wiki.ffo.jp/html/15893.html
     end
 
     -- Apply Aftermath
--- a/scripts/globals/weaponskills/blade_kamu.lua
+++ b/scripts/globals/weaponskills/blade_kamu.lua
@@ -28,11 +28,11 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     params.crit100 = 0.0 params.crit200 = 0.0 params.crit300 = 0.0
     params.canCrit = false
     params.acc100 = 0.0 params.acc200 = 0.0 params.acc300 = 0.0
-    params.atk100 = 1; params.atk200 = 1; params.atk300 = 1
+    params.atk100 = 1.3; params.atk200 = 1.3; params.atk300 = 1.3
 
     if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
         params.str_wsc = 0.6 params.int_wsc = 0.6
-        params.atk100 = 1.3125; params.atk200 = 1.3125; params.atk300 = 1.3125
+        params.atk100 = 3; params.atk200 = 3; params.atk300 = 3 -- http://wiki.ffo.jp/html/15893.html
     end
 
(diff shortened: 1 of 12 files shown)


scripts/globals/weaponskills/blade_kamu.lua
```diff
@@ -28,11 +28,11 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     params.crit100 = 0.0 params.crit200 = 0.0 params.crit300 = 0.0
     params.canCrit = false
     params.acc100 = 0.0 params.acc200 = 0.0 params.acc300 = 0.0
-    params.atk100 = 1; params.atk200 = 1; params.atk300 = 1
+    params.atk100 = 1.3; params.atk200 = 1.3; params.atk300 = 1.3
 
     if xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES then
         params.str_wsc = 0.6 params.int_wsc = 0.6
-        params.atk100 = 1.3125; params.atk200 = 1.3125; params.atk300 = 1.3125
+        params.atk100 = 3; params.atk200 = 3; params.atk300 = 3 -- http://wiki.ffo.jp/html/15893.html
```
WinterSolstice8: The JP wiki claims it's def ignore +25% attack bonus +125%.  Can you adjust the attack bonuses to match, and add the following?

``` 
params.ignoresDef = true
params.ignored100 = 0.25
params.ignored200 = 0.25
params.ignored300 = 0.25
```

scripts/globals/weaponskills/blade_ku.lua
```diff
@@ -30,11 +30,13 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
...
     params.crit100 = 0.0 params.crit200 = 0.0 params.crit300 = 0.0
     params.canCrit = false
-    params.acc100 = 0.8 params.acc200= 0.9 params.acc300= 1.0
+    params.acc100 = 0.9 params.acc200= 1 params.acc300= 1.1
+    --  the TP bonus for TP 1,000 and above has been raised, which has substantially raised the hit rate http://wiki.ffo.jp/html/732.html
+    -- data on ws accuracy are difficult to come by, would need to decide on sane value for the accuracy boost stated in JPWiki for now I'm only adding 0.1 per tier
     params.atk100 = 1; params.atk200 = 1; params.atk300 = 1
 
-    if (xi.settings.USE_ADOULIN_WEAPON_SKILL_CHANGES == true) then
-        params.ftp100 = 1.25 params.ftp200 = 1.25 params.ftp300 = 1.25
```
WinterSolstice8, line 37: No need to remove fTP bonus. It's legitimate according to BG research. I can't really speak to the accuracy bonus of course. 
```params.ftp100 = 1.25 params.ftp200 = 1.25 params.ftp300 = 1.25```

scripts/globals/weaponskills/blade_kamu.lua
Kayanesa: I had to triple check before I finally saw the note at the bottom good catch

Kayanesa:
Strange question but the build should not fail as far as I know I didn't change anything that would have an impact. How to I rerun the check without  changing the file?

WinterSolstice8:
No worries, that error had nothing to do with you.

WinterSolstice8:
Everything but Blade: Metsu checks out, there's claims of it being 3 different fTP values. Byrth took the data in a bg thread and put 5.0 in bg wiki, when I do the math it's close to 5.0 but a little higher. I don't claim to be an expert on the calculations but it seems like it's probably 5.0...

I can run my own test, as I have a Kikoku, will get a few samples later.

Kayanesa:
Sounds good, I can't really do test myself as I dont have a leveled character anymore.

WinterSolstice8:
For Kikoku 119 III
image
qRatio caps at 4.25 as far as we know.

image
Sure looks like 5.0 fTP to me, it's the closest fit, though i don't think i'm _actually_ capped on fSTR. I had very lowball numbers based on my stats before, off by about 800. I pumped it even more and i'm still off but by a lot less. I'm at 102+173 STR, can't really get more without better gear that I don't have for NIN.

Kayanesa:
Looks close enough to me

Xaver-DaRed (approved):

TeoTwawki:
I sure do like seeing all those url references in comments so we know where data came from :+1:

Issue #885, opened 2021-10-27
[synth recipes] invalid data from automatic scrape, disabled recipes from lack of data


TeoTwawki:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: All

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
So back in topaz, an automated scrape+ compare of known available data sources was performed and synth_recipes.sql was remade from scratch (with some column order changes too). I believe this attempt at automation came about due to multiple community members back channel complaints and non-reports of the projects synth system being "all kinds of wrong" without any contribution towards correcting it in the way of data, pr, or even issue reports and a hope that by not having any "wrong" recipes it would encourage people to add the missing ones and we could verify the addition was correct and not guessed at.

In my opinion that hasn't worked out. Server owners that actually updated just re-enabled/filled in with their own guess work where they needed to, and little to no retail verification of anything has been performed and pull requested back.

Since then, many previously craftable items cannot be synthed. In many cases we just need the required skill leveled figured out and filled in. But in some others the recipe itself is obviously wrong due to being outright impossible and the column count of the row is wrong. Take the Nagan, which has a duplicate recipe containing a diff skill level and a gold ingot. This recipe uses 9 materials. There are only 8 material slots in game, no recipe EVER uses more than 8 slots and accordingly we only have 8 material input columns in our database structure. This is just one example. not all of the invalid rows are duplicated. The scrape made a lot of mistakes.

Additionally, we did not (and cannot) auto scrape the JP wiki, which can fill in many of the unknowns if someone can read it (I've tried with translation but while I can read a given page well enough, I have a hard time FINDING the page I need in the first place).

We need more hands on this. I am asking **YOU** dear reader, to consider pitching in if you are able, by checking in game on retail or by comparing our recipes table to the JP wiki and sharing corrections. If you aren't comfortable making a pull request you cans hare right here in this issue thread.

Skerxan:
[...]

PR #1395, opened 2022-03-10
update more darksteel/adaman items, start ilvl reviews

RAIST5150:
**_I affirm:_**

Some more from the list in #885, moving into reviewing lower level ilevel pieces that start using SoA area items.

Found duplicate recipes for Nagan (14566 and 14578).  The first had an extra item listed, which was corrected, but leaving both commented out for now.


Xaver-DaRed (approved):
Looks reasonable

PR #1422, opened 2022-03-17
[Hold] Change spell damage resist check from 3 checks to 1 check

Enedin9000:
**_I affirm:_**

BLMs got a lot more resists recently, and this is because Magic Hit Rate is now checked 3 times instead of just once. This means that even with 95% MHR, you only get a 85.7% unresist rate. This drops down dramatically: at 75% MHR, you get a 42.2% unresist rate:
image

This change is even more pronounced when we average damage over X nukes:
_EDIT: dumbass me forgot to triple the numbers for 1/2 and 1/4 resists (there are 3 ways to get a 1/2 resist on 3 rolls, not one) so the change isn't as dramatic but the point still stands._
image

The [BG wiki Resist page](https://www.bg-wiki.com/ffxi/Resist) shows how it used to be done, and that's what I changed it to now: we pre-calculate the tiers based on the MHR and random once. This brings it in line with idea of "rolling again if you resisted" as described on the BG page.

### DIFF
--- a/scripts/globals/magic_utils/spell_damage.lua
+++ b/scripts/globals/magic_utils/spell_damage.lua
@@ -414,14 +414,22 @@ xi.magic_utils.spell_damage.calculateResist = function(caster, target, spell, sk
     -- STEP 4: Get Resist Tier
     -----------------------------------
     local resistTier = 0
-    local randomVar  = math.random(1, 100)
+    local randomVar  = math.random()
 
-    -- 3 random rolls.
-    for i = 3, 1, -1 do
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
I'm gonna put this on hold for now, until I get to properly determine how mob elemental resistances should be handled

Xaver-DaRed:
I applied this changes in this PR #1438 with proper formating while I was messing with magic. I made sure to co-author you.
Those charts were invaluable. TYVM.
Closing this.

PR #1423, opened 2022-03-17
[LandKingSystem] Unhide ???

Kayanesa:
**_I affirm:_**

When using the LandKingSystem with a value of 2(both time spawn and force spawn possible) you get in a situation where
the NM spawn and make a call to hide the QM (from mob lua). But unlike trading where the npcUtil.popFromQM take care to unhide the QM after mob despawn (from npcs.qm lua) the QM get hidden but never come back(until next server reset).


TeoTwawki:
I mentioned this situation on the last pr that removed what you are adding. I really wish ppl would learn how it was intended to work before deleting code, these poor scripts have been redone more times than Halver.

TeoTwawki (approved):

TeoTwawki:
ty for putting the `???` back

zach2good (approved):

claywar (approved):

PR #1452, opened 2022-03-23
[ROV] mission 2-6

Kayanesa:
**_I affirm:_**
I'm gonna need help on this one I have tried to make the cs work in every way I could think of but so far the best I got is this.
I get a black screen when loading the cs and the mission does get flagged completed but the cs never load and you get black screen till you zone.
I know that the cs work fine with the command !cs 164 0 0 0 0 0 0 0 0 0
and doesn't work without that 9th parameter.

### DIFF
--- a/scripts/missions/rov/2_06_Desert_Winds.lua
+++ b/scripts/missions/rov/2_06_Desert_Winds.lua
@@ -1,5 +1,5 @@
 -----------------------------------
--- Numbering Days
+-- Desert Winds
 -- Rhapsodies of Vana'diel Mission 2-6
 -----------------------------------
 -- !addmission 13 54
@@ -7,7 +7,6 @@
 require('scripts/globals/missions')
 require('scripts/globals/interaction/mission')
 require('scripts/globals/zone')
-require("scripts/globals/rhapsodies")
 -----------------------------------
 
 local mission = Mission:new(xi.mission.log_id.ROV, xi.mission.id.rov.DESERT_WINDS)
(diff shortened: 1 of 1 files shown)


claywar:
I'm going to level with you.  If you're writing this based on guesswork alone, it will not be accepted.  Even in the past few missions that I rewrote, I know there's some items missing, but we don't have caps to verify them.  We must be focused on real data from retail to recreate these missions, and fill in the gaps as they become clear.

Kayanesa:
This is really not ready for release it need to update differently depending on where we are on aht mission. Base on that and the difference in conversation on the cs the first optional argument is not much of leap. But I definitely won't be able to test all that in release. If that's what is needed I will refrain from doing work on the missions.
edit: sorry I read too many mission lately this one doesn't need to update differently it's the next one.

Kayanesa:
Even if it's not gonna get accepted here, I would love to know how to make that cs work. I will still implement it on my server.

Kayanesa:
Good new is I kept at it and found a way to make it work. For reason I don't understand I can't make the cs work with the onZoneIn fonction but it's working well with afterZoneIn.

I will now explain how the code is working I only did the first optional variable.

This cs is linking you toward the next cs at the whitegate palace door that cs only get available when you have flag the TOAU mission 12: Royal Pupeteer.
https://ffxiclopedia.fandom.com/wiki/Ever_Forward

with option 0 they are sending you to Salaheem's Sentinels saying something happened to gessho with a crazy mithra (so you don't know about Salaheem's sentinels yet)
with option 1 they are sending you to the palace but says you will be unwelcome and should work on your reputation (you know about Salaheem's sentinels but you are not ready for next cs so not at Royal Pupeteer)
with option 2 they are just sending you to the palace no warning (so you are atleast at mission Royal Pupeteer)

It is based on the next mission requirement and I can't deny it been guess work. But now you have all the information to make a decision if this pr is good or not.

If it's considered to be too much guess work I can either leave all option to 0 or if there are no answer in the next two week or so I will just delete this PR.

Xaver-DaRed:
First off, the reason you dont get it to work with onZoneIn is becouse that function doesn't admit additional parameters, just the eventID.
Normally, there is an onEventUpdate right after that feeds the CS with them.

And that it's the reason we try as much as possible not to do guesswork. It's both an insane amount of effort that could be saved by watching a capture. And we also have no way of knowing if it's actually the correct way of coding the mission/quest.

One of the reasons Missions were in such a bad state before the conversion was becouse ppl just read the progress from wikis. And they were not correct. So the resulting implementation wasn't correct either.

Kayanesa:
Where can I watch a capture?
I did try with the onEventUpdate had no luck.

Xaver-DaRed:
[This wiki page](https://github.com/LandSandBoat/server/wiki/Resources) Capture server discord link

Kayanesa:
I will go back to draft and see if I can find the capture

Kayanesa:
Sadly there is no capture for my mission since they had the boarding permit and the mission never got flag, no too sure what I can do. But it wont be this weekend

PR #1454, opened 2022-03-24
[Geo] Basic luopan positioning and corrected Luopan size

Omnione:
Not much here yet, doing very basic changes to get the essentials in place in small stages,
Luopan now gets placed on the target not just the caster as before,
Full Circle now works, as before it did nothing and couldn't despawn a luopan once placed.

**_I affirm:_**


scripts/globals/job_utils/geomancer.lua
```diff
@@ -40,9 +40,22 @@ end
 -- TODO: After elements are aligned in the codebase, this should become:
 -- xi.job_utils.geomancer.spawnLuopan = function(player, target, spell, tickEffect, tickPower, targetType)
 xi.job_utils.geomancer.spawnLuopan = function(player, target, modelID, tickEffect, tickPower, targetType, spell)
+    if target == nil then
```
TeoTwawki:
```lua
 if target then
    xi.pet.spawnPet(player, xi.pet.id.LUOPAN)
else
    return
end
```

Omnione: oof

scripts/globals/job_utils/geomancer.lua
```diff
@@ -52,9 +65,11 @@ xi.job_utils.geomancer.spawnLuopan = function(player, target, modelID, tickEffec
 
     -- Change the luopans appearance to match the effect
     -- TODO: This is should be the element of the spell being cast added as an offset
-    --       on top of a base model ID in core.
+    -- on top of a base model ID in core.
     luopan:setModelId(modelID)
 
+    -- Size is effected by EntityFlags
+    player:setPetFlags(luopanSize)
```
TeoTwawki: technically, we should be setting just be using bitewise ops on its existing value just as a pattern for the folks who might copy paste from here for something else. (2nd and 3rd bits in the mask) rather then replacing the entire mask with a decimal value. But practically, it does not matter for this entity as it has not other flags, so I'm not going to insist, just noting it here in case it needs revisiting later or if someone disagrees with my dismissing that minor issue.

Its just a single value input implies (incorrectly) that the number = how big it is, rather than the number just so happening to be bits in a mask. they might try 3 or 5 or 50 not knowing that isn't how this works.

TeoTwawki (dismissed):

scripts/globals/job_utils/geomancer.lua
Omnione: changed now, much better order of opertation

scripts/globals/abilities/full_circle.lua
```diff
@@ -0,0 +1,39 @@
...
+    end
+    return 0,0
+end
+
+ability_object.onUseAbility = function(player, target, ability)
+    local luopan        = player:getPet()
+    local hpp_remaining = luopan:getHPP()
+    local mp_cost       = luopan:getLocalVar("MP_COST")
+    local mp_returned   = (0.5 * mp_cost) * hpp_remaining / 100
+    local hp_returned   = 0
```
claywar, line 32: This declaration doesn't seem necessary, since the conditional both sets and restores a value

scripts/globals/abilities/full_circle.lua
```diff
@@ -0,0 +1,39 @@
...
+
+    if crMerit > 0 then
+       hp_returned = mp_returned + ((mp_returned *0.5) *crMerit)
+       player:restoreHP(hp_returned)
+    end
+    player:restoreMP(mp_returned)
+    target:despawnPet()
+end
+
+return ability_object
```
claywar: Newline please :)

scripts/globals/abilities/full_circle.lua
TeoTwawki: and this is why morning teo should make afternoon teo double check before approving xD I 100% missed this

Omnione: This should be good now, also used the oppertunity to correct full circle formula for mp and hp returned

Omnione:
corrected merit values for full circle effect and curative recantation also on that last commit

scripts/globals/abilities/full_circle.lua
```diff
@@ -0,0 +1,38 @@
...
+require("scripts/settings/main")
+require("scripts/globals/msg")
+-----------------------------------
+local ability_object = {}
+
+ability_object.onAbilityCheck = function(player, target, ability)
+    if not player:hasPet() or not player:getPetID() == 75 then
+        return xi.msg.basic.REQUIRE_LUOPAN, 0
+    end
+    return 0,0
```
zach2good: Can we extract this block into the job utils for geo?

```lua
ability_object.onAbilityCheck = function(player, target, ability)
    return xi.job_utils.geomancer.checkLuopan(player, target, ability)
end

...

xi.job_utils.geomancer.checkLuopan = function(player, target, ability)
    if not player:hasPet() or not player:getPetID() == 75 then
        return xi.msg.basic.REQUIRE_LUOPAN, 0
    end

    return 0, 0
end
```

scripts/globals/job_utils/geomancer.lua
zach2good: I'd say just add a comment saying "NOTE: This is overwriting the whole value, rather than just setting individual flags"

Then copy/pasters have been warned

src/map/lua/lua_baseentity.cpp
```diff
@@ -12193,6 +12193,48 @@ uint32 CLuaBaseEntity::getMobFlags()
...
 }
 
+/************************************************************************
+ *  Function: setPetFlags()
+ *  Purpose : Manually set Pet flags
+ *  Example : player:setPetFlags(flags)
+ *  Notes   :
+ ************************************************************************/
+
+void CLuaBaseEntity::setPetFlags(uint32 flags)
```
zach2good: Do we need these new bindings? When we could just have:
```lua
local pet = master:getPet()
if pet then
    pet:setFlags(...)
end
```

scripts/globals/job_utils/geomancer.lua
Omnione: I'm going to leave entity flags alone for now until a better solution / refactor has been put in place, as this i feel needs to be addressed before going further.

scripts/globals/abilities/full_circle.lua
Omnione: this has been done now, makes more sense

src/map/lua/lua_baseentity.cpp
Omnione: This does not work for pets as is, would have been easier but not a viable solution in its current form,
as montioned in a prevous comment, hopefully a refactor of the entity flags will provide a solution.

scripts/globals/job_utils/geomancer.lua
Omnione: No longer a problem, removed all the silly flag setting,
The flags dont seem to be setting correctly from the database so now they are set before spawn,
Props to Teo for helping with this.

scripts/globals/abilities/full_circle.lua
```diff
@@ -0,0 +1,36 @@
...
+ability_object.onAbilityCheck = function(player, target, ability)
+    return xi.job_utils.geomancer.geoOnAbilityCheck(player, target, ability)
+end
+
+ability_object.onUseAbility = function(player, target, ability)
+    local luopan        = player:getPet()
+    local hpp_remaining = luopan:getHPP()
+    local mp_cost       = luopan:getLocalVar("MP_COST")
+    local fcMerit       = player:getMerit(xi.merit.FULL_CIRCLE_EFFECT)
+    local mp_returned   = ((0.5 + (fcMerit /100)) *mp_cost) *hpp_remaining / 100
```
zach2good: Order of operations is unclear:
` ((0.5 + (fcMerit / 100)) * mp_cost) * (hpp_remaining / 100)`

scripts/globals/abilities/full_circle.lua
```diff
@@ -0,0 +1,36 @@
...
+    local crMerit       = player:getMerit(xi.merit.CURATIVE_RECANTATION)
+
+    if crMerit > 0 then
+       player:restoreHP((1.2 *(mp_cost *crMerit) *(hpp_remaining /100)))
```
zach2good: Spacing
`player:restoreHP((1.2 * (mp_cost * crMerit) * (hpp_remaining / 100)))`

zach2good (approved):
Looks good, just some small styling points

src/map/utils/petutils.cpp
```diff
@@ -1596,7 +1596,8 @@ namespace petutils
             PPet->SetMLevel(PMaster->GetMLevel());
             PPet->health.maxhp = (uint32)floor((250 * PPet->GetMLevel()) / 15);
             PPet->health.hp    = PPet->health.maxhp;
-
+            // This sets the correct visual size for the luopan
+            PPet->m_flags      = 0x0000008B;
```
TeoTwawki, line 1602: so the reason this wasn't loaded from sql is it was never added to the query for building the entity:
https://github.com/LandSandBoat/server/blob/base/src/map/utils/petutils.cpp#L140

TeoTwawki: it would take it from mob pools normally by being added to that and using another handle entry below the linked section to populare PPet->m_flags

TeoTwawki: adjust the comment here to mention it

Omnione: Made a note of the entity flags not being used with a TODO

scripts/globals/abilities/full_circle.lua
Omnione: Should be a bit simpler to follow now

PR #1460, opened 2022-03-26
pr: changed addItem to accept a table of augments properly

m241dan:
**_I affirm:_**

I know this doesn't really affect anyone but I had used it for some custom code that I'm reimplementing and it's a real bummer that it doesn't work like it used to. More details in the commit.

Also, const'd it because const is good.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -3233,11 +3233,10 @@ bool CLuaBaseEntity::addItem(sol::variadic_args va)
                     if (augmentsObj.is<sol::table>())
                     {
                         auto augmentsTable = augmentsObj.as<sol::table>();
-                        for (auto& entryPair : augmentsTable)
+                        for (const auto& entryPair : augmentsTable)
                         {
-                            auto   pair   = entryPair.second.as<sol::table>();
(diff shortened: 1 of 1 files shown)


TeoTwawki:
```
This matches the "old way" of passing a table. Where the table
could look like
{
   [100] = 1,
   [50] = 2,
}

where the index is the augid and the value is the augval

The code (as it was converted) was looking at something like
{
   [1] = {100, 1}
}
which is not how this "pass augments to addItem via table" used to work.
```

Unfortunately this was an intentional change and we don't want the old format anywhere in project scripts. And since we'd reject that format in pr's, we do not want to enable it core side again and then have to answer why its possible.

We also already use the new formatting in several place in project, with a lot more not yet merged that uses it in the pipes. Multiple other servers have used the newer form in their custom scripts as well. It would be better to either make a custom alternate function then find+replace the function name in your custom scripts, or to write a lua function to intercept and transform your old tables on the fly. Or just do some regex to reformat yours in script. 

In any case, I have to say no to this pr, as we need our newer format to do what we want properly.

m241dan:
Ah, did not know there was work out on branches. Thought it was just an
accident. Cool, I'll fix my stuff.

PR #1461, opened 2022-03-26
Add support for calculated Traverser Stones and add migration script

claywar:
**_I affirm:_**

Going to slowly start pulling apart the old Abyssea branch into sane chunks.  This does not implement function yet, but adds core support for calculated vs claimed traverser stones based on an epoch that is set upon starting abyssea questlines.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -13213,6 +13213,103 @@ void CLuaBaseEntity::addDropListModification(uint16 id, uint16 newRate, sol::var
 }
 
 /************************************************************************
+ *  Function: getAvailableTraverserStones()
+ *  Purpose : Returns the number of Traverser Stones available for claim
+ *  Note: Does not yet account for KI reduction
+ ************************************************************************/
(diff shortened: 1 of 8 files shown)


Xaver-DaRed (approved):

zach2good (approved):

claywar:
Oh, doh, forgot about the Sql changes

src/map/lua/lua_baseentity.cpp
```diff
@@ -13212,6 +13212,103 @@ void CLuaBaseEntity::addDropListModification(uint16 id, uint16 newRate, sol::var
...
+        return 0;
+    }
+
+    auto* PChar = static_cast<CCharEntity*>(m_PBaseEntity);
+    return charutils::getTraverserEpoch(PChar);
+}
+
+/************************************************************************
+ *  Function: setTraverserEpoch()
+ *  Purpose : Returns the number of Traverser Stones claimed by the player
```
kaincenteno, line 13250: Function comment purpose should have the "return" part removed

PR #1468, opened 2022-03-26
[CI] Add python checks to CI

zach2good:
**_I affirm:_**

Includes a small pre-change to make sure dbtool is hit by CI on this run

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,8 +12,8 @@ on:
       - everything
 
 jobs:
-  Linux_2004_Clang11_64bit:
-    runs-on: ubuntu-20.04
+  Linux_Clang11_64bit:
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -132,7 +132,8 @@ jobs:
     - name: Install Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y software-properties-common clang-format cppcheck luajit-5.1-dev luarocks mariadb-server mariadb-client libmariadb-dev-compat
+        sudo apt-get install -y software-properties-common clang-format cppcheck luajit-5.1-dev luarocks mariadb-server mariadb-client libmariadb-dev-compat python3.7
```
claywar: Is 3.7 the intended version? I thought we required greater than that now?

zach2good: It's copy/paste from the other CI entries, I don't think we have a hard reliance on any Python 3.x version. Happy to have CI use latest

claywar (approved):

Issue #1470, opened 2022-03-27
Navmesh - Boat Ride : Agro into CargoHold / Mobs path to Pilot's raised area


Kalifriki:
![Bug_Boat_Ver](https://user-images.githubusercontent.com/46100825/160272106-6d505b38-61ed-4841-a795-79e1b616fbef.png)

- [X] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [X] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [X] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [X] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [X] I have specified what branch this happens on branch: EDITME

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Navmesh for boat rides (all: Selbina to Mhuara; Mhuara to Whitegate, Whitegate to Nashamu): mobs seem to ignore traditional pathing and are capable of agro into the cargo hold as well as moving into the pilot's perch area.

Behavior has been observed on Dawnbreak & Rebirth.
![Bug_BoatAgro](https://user-images.githubusercontent.com/46100825/160271570-9dd108b0-48dd-4ed0-8eea-be6d1df4b02d.png)

claywar:
It might be a good idea to raise this issue on the navmesh repo here https://github.com/LandSandBoat/xiNavmeshes as well

zach2good:
I've done a little research:

To determine aggro and line of sight stuff, we use `raycasts`, this is a regular raycast:
<img width="1269" alt="image" src="https://user-images.githubusercontent.com/1389729/160285180-45b996fc-0873-4b1f-b132-30bcfe085be4.png">

Bad raycast:
<img width="1269" alt="image" src="https://user-images.githubusercontent.com/1389729/160285274-c44c7348-cc96-488b-819c-f13acb37777c.png">

(Note, that raycasts on our navmeshes aren't in 3D space, they're projected along the flat plane of the navmesh)

The navmesh is split up between multiple `regions`. Presumably, a raycast is meant to continue to work between regions (which is fine).

Apart from in this case, where there are regions that shouldn't be connected, being connected:
<img width="1269" alt="image" src="https://user-images.githubusercontent.com/1389729/160285396-53784f1c-e634-4f09-a7ac-cd6a6f65eb53.png">

[...]

PR #294, opened 2021-06-08
[navmesh] Reduce values for findNearestPoly search area

zach2good:
I did a live stream, I profiled some things. All the navmesh calls have gone from 200us to 50us (or less). 
Probably warrants some testing, but looks like a pretty easy win.

The extants passed to findNearestPoly appear to just be defining a box in which to search. We generate the vast majority of points that we're searching around, and later validating, by lerping between points given to us from detour's findPath, so they're very likely to be valid, or extremely close to a valid point 

So it seems to be safe to reduce this search area. 

There is a tiny chance this might affect wyrm's draw in. If they can search a massive area, they won't be able to validate a broken-add point to warp you to, so they'll work better. This is just untested conjecture though, could be worth a look. 

**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] that I agree to LandSandBoat's [Limited Contributor License Agreement](https://github.com/LandSandBoat/server/blob/base/.github/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [x] that I have **read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)**
- [x] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

PR #1474, opened 2022-03-27
[WIP] SightMesh impl

zach2good:
**_I affirm:_**
- [nope] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

I did an analysis in this ticket:
https://github.com/LandSandBoat/server/issues/1470

Without using _filthy_ hacks like reducing the vertical detection range for mobs etc., recast+detour isn't able to handle navigation/raycasting between multiple floors, and it never will be.

So I did a little googling, and found a lib that can take an OBJ file, and then allows you to raycast between points of it arbitrarily. It also _seems_ fast, but I've only done very very basic measurements so far.

```
zachtoogood@FVFFK42EQ05P build % ./raycastmeshtest ../models/Ship_bound_for_Mhaura.obj
Loading wavefront obj file '../models/Ship_bound_for_Mhaura.obj'
Creating Raycast Mesh. Building AABB
Performing 65536 raycasts.
Finished raycasting; converting image to grayscale.
Saving image file RaycastMesh.png
Average Time: 22 ns
Min Time: 0 ns
Max Time: 92500 ns
```

Based on my optimisations in this PR: https://github.com/LandSandBoat/server/pull/294

Our current raycast is is in the region of `18us`, this implementation _seems_ to be `22ns` -> `1000x faster`

In the end it might not even work properly. But it might!

### DIFF
--- a/ext/CMakeLists.txt
+++ b/ext/CMakeLists.txt
@@ -2,6 +2,7 @@ add_subdirectory(detour)
 # add_subdirectory(lua) # Handled globally 
 # add_subdirectory(mysql) # Handled globally
 add_subdirectory(recast)
+add_subdirectory(sightmesh)
 add_subdirectory(sol)
 add_subdirectory(spdlog)
 # add_subdirectory(zmq) # Handled globally
--- a/ext/sightmesh/CMakeLists.txt
+++ b/ext/sightmesh/CMakeLists.txt
@@ -1,0 +1,11 @@
+set(SOURCES
+    RaycastMesh.cpp
+    RaycastMesh.h
+    wavefront.cpp
+    wavefront.h)
+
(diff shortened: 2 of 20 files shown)


Kayanesa:
If this can solve the problem like the big link from multy floor in dyna Jeuno I will cross my finguers and hope for your success.

zach2good:
While I was super hyped about this solution, it had a number of issues:
- The sightmesh files were generated from the obj files, and were approximately the same size, or even larger than the obj files once I serialized their data out (11gb total)
- They added _considerably_ to startup time and RAM usage
- They were excellent and crazy accurate in some zones, but totally funky in other zones.

So, I'm closing this in favour of my other PR: https://github.com/LandSandBoat/server/pull/1483

PR #1476, opened 2022-03-29
updated Catastrophe drain to 30-70%

GravesXI:
added random drain 30-70% of damage return got info from BG
https://www.bg-wiki.com/ffxi/Catastrophe
credit to Xaver for helping with it as well

**_I affirm:_**

### DIFF
--- a/scripts/globals/weaponskills/catastrophe.lua
+++ b/scripts/globals/weaponskills/catastrophe.lua
@@ -39,8 +39,7 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, wsID, params, tp, action, primary, taChar)
 
     if not target:isUndead() then
-        local drain = math.floor(damage * 0.4)
-        player:addHP(drain)
+    player:addHP(math.floor(damage * math.random(3, 7) / 10))
     end
 
     return tpHits, extraHits, criticalHit, damage


scripts/globals/weaponskills/catastrophe.lua
```diff
@@ -39,8 +39,7 @@ weaponskill_object.onUseWeaponSkill = function(player, target, wsID, tp, primary
     local damage, criticalHit, tpHits, extraHits = doPhysicalWeaponskill(player, target, wsID, params, tp, action, primary, taChar)
 
     if not target:isUndead() then
-        local drain = math.floor(damage * 0.4)
-        player:addHP(drain)
+    player:addHP(math.floor(damage * math.random(3, 7) / 10))
```
claywar, line 42: Should we be increasing this by 10% at a time?  This might be something that's handled with a bit more randomness, but would need retail confirmation.

WinterSolstice8:
The source on the 30-70% is from [Taint](https://www.bluegartr.com/threads/114547-Random-Question-Thread?p=5580845&viewfull=1#post5580845), who is a well known DRK main, but an offhand comment by him isn't exactly... data. Even he says "assumed" in said comment. 

Additionally, drains only work with the remaining HP of the mob, you can't drain more HP than they have remaining, the current implementation will allow that, though that's just leftover from the original implementation. 

[JP wiki](https://wiki-ffo-jp.translate.goog/html/3088.html?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp&_x_tr_sch=http) <s>has no mention of the HP absorb percentages</s> but it does rearrfirm the utsusemi bypass.

WinterSolstice8:
Whoops, looked back at the page just to make sure I wasn't going crazy because I thought JP wiki would absolutely document that

`追加効果のHP吸収は、最大でWSダメージの半分を自身のHPに吸収する。`
through deepL comes out as 
`The additional effect of HP absorption absorbs up to half of the WS damage into its own HP.`
so JP wiki thinks it's "up to 50%" which sounds more reasonable for the power of a relic WS at 75.

I think the actual values need testing here, as well as other stuff like does obi/dark MAB+ effect the drain? though we could probably punt that part for now. It looks like the original 40% drain in LSB was too low at the very least.

PR #1482, opened 2022-04-01
[core] Add bindings for custom menus

zach2good:
**_I affirm:_**

### DIFF
--- a/scripts/commands/menu.lua
+++ b/scripts/commands/menu.lua
@@ -10,18 +10,30 @@ cmdprops =
 }
 
 function onTrigger(player)
-    player:customMenu({ "Test Menu (Play Effect)",
-        { "Option 1: Hearts", function(playerArg)
-            playerArg:PrintToPlayer("Option 1 Selected", xi.msg.channel.NS_SAY)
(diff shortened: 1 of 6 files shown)


zach2good:
image

scripts/commands/menu.lua
```diff
@@ -0,0 +1,27 @@
...
+-----------------------------------
+
+cmdprops =
+{
+    permission = 1,
+    parameters = ""
+}
+
+function onTrigger(player)
+    player:customMenu({ "Test Menu (Play Effect)",
```
claywar: This styling kills me, and I'd hate to set the example for further users.  Perhaps define the table outside of the function call with our style guides in mind?

zach2good: Happy to accept other stylings, all Allman?

TeoTwawki: I think we sort of cover this case:

https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md#inline-tables

zach2good: I think I'd be OK adjusting this rule to be "define your table beforehand and pass it in", I never really liked this edge case

claywar (approved):

PR #1497, opened 2022-04-04
Fix Server Message Packet

zach2good:
**_I affirm:_**

Thanks Ino and Eden!

### DIFF
--- a/src/map/packets/server_message.cpp
+++ b/src/map/packets/server_message.cpp
@@ -34,24 +34,23 @@ CServerMessagePacket::CServerMessagePacket(const string_t& message, int8 languag
     ref<uint8>(0x06)  = 1;
     ref<uint8>(0x07)  = language;
     ref<uint32>(0x08) = (uint32)(timestamp == 0 ? time(nullptr) : timestamp);
-    ref<uint32>(0x0C) = 0; // Message Length.. (Total)
-    ref<uint32>(0x10) = 0; // Message Offset..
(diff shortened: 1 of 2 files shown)


src/map/packets/server_message.cpp
```diff
@@ -34,24 +34,23 @@ CServerMessagePacket::CServerMessagePacket(const string_t& message, int8 languag
...
+        ref<uint32>(0x10) = message_offset;    // Message Offset
+        ref<uint32>(0x14) = (uint32)sndLength; // Message Length
 
-        memcpy((data + (0x18)), message.c_str() + message_offset, sndLength);
+        memcpy((data + (0x18)) , message.c_str() + message_offset, sndLength);
 
         auto textSize = (uint8)(sndLength + sndLength % 2);
-        // TODO: Verify the below math
-        this->setSize(((((0x14 + textSize) + 4) >> 1) & 0xFE) * 2);
+        this->size = ((((0x14 + textSize) + 4) >> 1) & 0xFE);
```
claywar: Remember that this->size (previous implementation) was 16bit values, not by byte.  This line should most likely not change.

zach2good: Quite right

zach2good:
Tested now, the annoying 'g' at the end of the server message is now gone 🙌

claywar (approved):

zach2good:
Tried a bunch of RUN abilities, still works

Issue #1108, opened 2021-12-14
Duplicate NM Spawns in Dynamis Jeuno


CatsEyeXI:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: everything

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

It appears to me that NM's have duplicate entries for Dynamis Jeuno. This appears to be causing them to spawn in multiple locations simultaneously. I think I've isolated this to multiple entries in mob_spawn_points.sql, but after speaking with @Xaver-DaRed I'm not sure if zone init is simply supposed to "pick one" and not "all of them". If it's a matter of dupes in the SQL file I'd be happy to try a PR to fix it, but I have a feeling its more complicated.

CatsEyeXI:
For example:

```
INSERT INTO `mob_spawn_points` VALUES (17547378,'Humnox_Drumbelly','Humnox Drumbelly',34,26.200,-1.069,-65.960,1);
INSERT INTO `mob_spawn_points` VALUES (17547382,'Humnox_Drumbelly','Humnox Drumbelly',34,29.090,-0.498,-46.690,44);
INSERT INTO `mob_spawn_points` VALUES (17547388,'Humnox_Drumbelly','Humnox Drumbelly',34,29.478,-0.495,-31.992,200);
```

TeoTwawki:
I can confirm: its normal for the dats themselves (and thus our moblist) to have multiple copies of some NMs, but the code to spawn them should limit to one copy up at a time. If its not somebody made an oopsie.

PR #1498, opened 2022-04-04
[Dynamis Jeuno] mob group spawn

Kayanesa:
Added a REFILL_MOB_GROUP to spawn monster in different grouping instead of all individual spawn. Created the table of possible spawn.
All following documentation
https://www.dropbox.com/s/p3zpuq93bs2vw90/Dynamis%20Jeuno%20IDs%20Groups%20Mechanics.zip?dl=0
It also line up with retail video of the zone.
Also added NM spawn condition for both group NM and Place holder NM type.
Will solve #1108 

**_I affirm:_**

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -774,39 +774,40 @@ INSERT INTO `mob_droplist` VALUES (141,0,0,1000,1118,220);
 INSERT INTO `mob_droplist` VALUES (141,0,0,1000,1540,30);
 INSERT INTO `mob_droplist` VALUES (141,0,0,1000,16995,450);
 INSERT INTO `mob_droplist` VALUES (142,0,0,1000,846,60);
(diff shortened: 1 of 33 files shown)


sql/mob_droplist.sql
```diff
@@ -17137,6 +17202,19 @@ INSERT INTO `mob_droplist` VALUES (3209,0,0,1000,14168,1000); -- (Sabotender Bai
 INSERT INTO `mob_droplist` VALUES (3209,0,0,1000,916,140); -- (Sabotender Bailarin) Cactaur Needle
```
claywar, line 17202: Looks like this file isn't based on LSB.  You've got additional entries here for Bailarin, and also missing a couple hundred entries near the end of the file.

Kayanesa: hum strange I didnt select that file for this pr but I tryied to base it on my other pr I must have mess something up.

PR #1499, opened 2022-04-04
[Dynamis Jeuno] DropList 5NM Update

Kayanesa:
Rework of the drop list and adding missing item like bijou and forgotten. Changed Vanguard group for the odious item drop.
Regrouped all NM that could be grouped. Remove timer on Vanguard to make them pop in group instead of having all of them on the map at the same time. (there were too many mob on the map compare to retail) 
Following the pop mechanic shown here https://www.dropbox.com/s/p3zpuq93bs2vw90/Dynamis%20Jeuno%20IDs%20Groups%20Mechanics.zip?dl=0
also goes hand in hand with my other PR #1502

Item drop done with http://www.ffxidb.com/zones/188/

**_I affirm:_**


Kayanesa:
I can't find the conflict strangely so just point me in the right direction and I will fix it.

PR #1510, opened 2022-04-07
[Magic] Convert Enhancing spells

Xaver-DaRed:
**_I affirm:_**

Convert White, Black and Ninjutsu spells to use a global file.
Removes global files used exclusively for Bar-spells

- Spells not existing yet:
Animus- (2), Adoquium, - storm II (8), Temper III

- New spell scripts/parameters implemented:
Refresh III, Regen V

- Other notes:
Base implementation of Embolden Effect. Doesnt account for merits/JP whatever.
Auspice is treated as an en-spell. That should be checked.
To review, focus on spell_enhancing.lua, spell_enhancing_ninjutsu.lua and spell_enhancing_teleport.lua
Other file changes were mostly cleanup.
Sanity fail due to magic.lua


scripts/globals/spells/black/retrace.lua
```diff
@@ -2,28 +2,16 @@
 -- Spell: Retrace
 -- Transports player to their Allied Nation. Can cast on allies.
 -----------------------------------
-require("scripts/globals/teleports")
-require("scripts/globals/status")
-require("scripts/globals/msg")
+require("scripts/globals/spells/spell_enhancing_teleport")
```
claywar, line 5: If there's multiple enhancing scripts in the global, perhaps we should move this into a sub-directory?  Just thinking about really long names.

scripts/globals/spells/songs/fowl_aubade.lua
```diff
@@ -11,7 +11,7 @@ spell_object.onMagicCastingCheck = function(caster, target, spell)
 end
 
 spell_object.onSpellCast = function(caster, target, spell)
-    return xi.spells.spell_song_enhancing.useEnhancingSong(caster, target, spell)
+    return xi.spells.spell_enhancing_song.useEnhancingSong(caster, target, spell)
```
claywar, line 14: Similar to the naming convention idea for sub-directories.  The global table and function name seem to be a bit redundant.  Not sure if we need to be as verbose (for example, the table could be xi.spells.enhancing, and the function wouldn't really need to have enhancing in its name as well.

scripts/globals/spells/spell_enhancing.lua
```diff
@@ -0,0 +1,344 @@
...
+xi.spells = xi.spells or {}
+xi.spells.spell_enhancing = xi.spells.spell_enhancing or {}
+-----------------------------------
+-- File structure:
+-- 2 Basic Functions to calculate final potency. Called by the main function.
+-- 1 Basic Function to calculate final duration. Called by the main function.
+-- 1 Main function, called by spell scripts.
+
+-- Table variables.
+local enhancingTable = xi.spells.parameters.enhancingSpell
```
claywar, line 21: I would almost prefer having parameters used in this script here for easier reference instead of one massive file containing params for all different spell types.

scripts/globals/spells/spell_enhancing.lua
```diff
@@ -0,0 +1,344 @@
...
+    end
+
+    ------------------------------
+    -- Level penalty to duration.
+    ------------------------------
+    if targetLevel < spellLevel then
+        duration = duration * targetLevel / spellLevel
+    end
+
+    ----------
```
claywar: Bit of an unnecessary comment at the end of this function

TeoTwawki (commented):
I think we could reduce the if/else-yness of the global code with some tabled function calling per discord discussion - worth a looking into at least

WinterSolstice8:
> I think we could reduce the if/else-yness of the global code with some tabled function calling per discord discussion - worth a looking into at least

per discussion we could mark it TODO in the code to get this merged

scripts/globals/spells/spell_enhancing.lua
Xaver-DaRed: Removed

Xaver-DaRed:
I spoke with Claywar and his comments will be adressed in a different PR. One that doesn't involve any new coding, just simplifying function names and getting rid of the global parameters.lua file

Xaver-DaRed:
Ready.
CI fail is because magic.lua file (cyclomatic complexity).
So: not my fault^^

claywar (approved):

PR #1515, opened 2022-04-08
Additional functionality for Allied Notes NPCs

Era-Lusiphur:
Adds:

- Discount for players of the same allegiance as the nation the NPC represents on most items, as on retail.
- Mapping for ciphers / information on how to enable. These should ultimately depend on an active campaign unless a server wants them both active all of the time.

Fixes incorrect values for Cobra Unit Cloche/Cap.

These NPCs could use a refactor once they're more fully implemented. They still need some aspects sorted, like the Can Equip check in the event update.  Did a little bit of cleanup work on styling for now and left them mostly as-is per side discussion with Teo.

**_I affirm:_**


claywar (approved):
I'll most likely end up refactoring this a bit, but the data looks sane

PR #1522, opened 2022-04-09
Interaction - Missions: Chains of Promathia Chapter 6

claywar:
**_I affirm:_**


claywar:
Completed testing.  Added two additional issues regarding PM6-4 BCNM; however, all missions for Chapter 6 are now converted and able to be completed.

Note: PM6-4 BCNM had the ability to skip the 32001 CS; this was removed, since the event is only a fade-out, and handled by other cutscenes.  If the player has completed this mission, those additional cutscenes will not fire.

Xaver-DaRed (approved):

PR #1524, opened 2022-04-09
[core] Load cmbSkill for Trusts

### LINKED ISSUE #1521: Trusts that would typically Dual Wield, or have two H2H attacks, only have one atttack at standard delay.
- [x] I have specified what branch this happens on branch: base

Any Trust that dual wields or has hand to hand attacks only performs a single attack per round and does not perform the left hand attack as normal. Changing the cmbSkill to match that of a dual wielding mob (Orcish Grappler) in the mobpool did not fix.

zach2good:
**_I affirm:_**

Fixes https://github.com/LandSandBoat/server/issues/1521


src/map/utils/trustutils.cpp
```diff
@@ -344,18 +349,26 @@ namespace trustutils
         auto adjustedDamage   = baseDamage * damageMultiplier;
         auto finalDamage      = std::max(adjustedDamage, 1.0f);
 
-        (dynamic_cast<CItemWeapon*>(PTrust->m_Weapons[SLOT_MAIN]))->setDamage(static_cast<uint16>(finalDamage));
-        (dynamic_cast<CItemWeapon*>(PTrust->m_Weapons[SLOT_RANGED]))->setDamage(static_cast<uint16>(finalDamage));
-        (dynamic_cast<CItemWeapon*>(PTrust->m_Weapons[SLOT_AMMO]))->setDamage(static_cast<uint16>(finalDamage));
+        // Trust do not really have weapons, but they are modelled internally as
+        // if they do.
+        auto* mainWeapon   = (CItemWeapon*)PTrust->m_Weapons[SLOT_MAIN];
```
claywar: Is there a reason for the C-style casts here?

claywar: Also, is there a chance that any of these weapon pointers could be null?

zach2good: Just while I copy/paste to get it working. They've never failed _yet_ with brutal casting

src/map/utils/trustutils.cpp
```diff
@@ -209,56 +211,59 @@ namespace trustutils
...
+                trust->m_MobSkillList = (uint16)sql->GetUIntData(14);
+
+                std::ignore = (uint16)sql->GetUIntData(15); // SpellID
+
+                trust->radius    = sql->GetUIntData(16);
+                trust->EcoSystem = (ECOSYSTEM)sql->GetIntData(17);
+                trust->HPscale   = sql->GetFloatData(18);
+                trust->MPscale   = sql->GetFloatData(19);
+
+                trust->speed     = (uint8)sql->GetIntData(20);
```
TeoTwawki: ya know we can possibly trim this right out - retail seems to have a static \*155\* for _all_ of them in client memory, not sure what the packets claim but its unusual to be that value.

(similarly speedSub is always 50)

src/map/utils/trustutils.cpp
```diff
@@ -344,18 +349,45 @@ namespace trustutils
...
+        if (auto* ammoWeapon = dynamic_cast<CItemWeapon*>(PTrust->m_Weapons[SLOT_AMMO]))
+        {
+            ammoWeapon->setDamage(static_cast<uint16>(finalDamage));
+            ammoWeapon->setDelay((trustData->cmbDelay * 1000) / 60);
+            ammoWeapon->setBaseDelay((trustData->cmbDelay * 1000) / 60);
+        }
 
-        (dynamic_cast<CItemWeapon*>(PTrust->m_Weapons[SLOT_AMMO]))->setDelay((trustData->cmbDelay * 1000) / 60);
-        (dynamic_cast<CItemWeapon*>(PTrust->m_Weapons[SLOT_AMMO]))->setBaseDelay((trustData->cmbDelay * 1000) / 60);
+        if (trustData->mJob == JOB_NIN ||
```
TeoTwawki:
```cpp
if (trust->cmbSkill == 1)
{
    PTrust->m_dualWield = true;
}
```
~~I think thats the value we use on mobs pools anyway..~~

TeoTwawki:
as I was asked to note: AAHM does 2 hits in my battle log

so did matsui-p when he was around

unsure about aldo will edit when I can check

Edit: aldo also hits twice

src/map/utils/trustutils.cpp
zach2good:
```
    SKILL_NONE         = 0,
    SKILL_HAND_TO_HAND = 1,
    SKILL_DAGGER       = 2,
    SKILL_SWORD        = 3,
    SKILL_GREAT_SWORD  = 4,
    SKILL_AXE          = 5,
    SKILL_GREAT_AXE    = 6,
    SKILL_SCYTHE       = 7,
    SKILL_POLEARM      = 8,
    SKILL_KATANA       = 9,
    SKILL_GREAT_KATANA = 10,
    SKILL_CLUB         = 11,
    SKILL_STAFF        = 12,
    ```

We set AAHM to 3 for  Swords, I wonder if we need some other way of handling this, since `MOBMOD_DUAL_WIELD` is marked as "please deprecate me"

Comment (TeoTwawki) [reply src/map/utils/trustutils.cpp]: yeah this is why calling it combat skill and pretending these object HAVE an actual weapon is bad.  \*

What we need in the future are to rip that column out and replace it with just 2 values: what damage type it is, and is it double hit (for dw and h2h).

There are lots of mobs using jobs that would normally dw/double hit but do not so we can't use job checks on this either.

Mimas in upper delkfut does not have a great axe, his model looks like it does. AAHM does not have swords, his model looks like it does. All mobs basically are h2h behind the scenes even if they single hit, and some have their attacks count as various dmg types (can use bst pets to confirm their damage type) and the same goes for trusts. Shantotto II's melee counts as magic damage (she can damage through pld's invincible) so the damage type and "skill" are not even connected for them like players. AAHM would be doublehit with slashing damage on retail, not "sword". For our purposes, we set it to 1 so he will double hit and disregard my previous mistake of "2". All values here other than 1 essentially do the same thing for us.

Check all our existing ninja mobs that dw. you will see all their pools say 1.


\* _This is why SE has to say things like such and such trusts "will count **as if** they wield a great sword for the purposes of" a job trait - because they do not genuinely have a great sword._

Comment (TeoTwawki) [reply src/map/utils/trustutils.cpp:392]: edited for readability/clearity.

**Review by TeoTwawki — APPROVED** [review]
~~lgtm~~

File: src/map/utils/trustutils.cpp [hunk 4]
```diff
@@ -273,56 +277,63 @@ namespace trustutils
...
-                trust->thunder_meva = (int16)sql->GetIntData(47);
-                trust->water_meva   = (int16)sql->GetIntData(48);
-                trust->light_meva   = (int16)sql->GetIntData(49);
-                trust->dark_meva    = (int16)sql->GetIntData(50);
+                trust->slash_sdt  = (uint16)(sql->GetFloatData(32) * 1000);
+                trust->pierce_sdt = (uint16)(sql->GetFloatData(33) * 1000);
+                trust->hth_sdt    = (uint16)(sql->GetFloatData(34) * 1000);
+                trust->impact_sdt = (uint16)(sql->GetFloatData(35) * 1000);
+
+                trust->fire_sdt    = (int16)sql->GetFloatData(36); // Modifier 54, base 10000 stored as signed integer. Positives signify less damage.
```
TeoTwawki, line 321: actually I missed this. elemental SDT isn't float data

TeoTwawki (dismissed):

zach2good:
They can 2-hit now:
image

TeoTwawki (approved):

PR #1535, opened 2022-04-10
[core] changed base mount speed to be 80

m241dan:
**_I affirm:_**

So, turns out, chocobos have been slower than walking. map.conf says this should be 80 (which is double oldschool movement speed).

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -222,7 +222,7 @@ int32 CBattleEntity::GetMaxMP() const
 
 uint8 CBattleEntity::GetSpeed()
 {
-    int16 startingSpeed = isMounted() ? 40 + map_config.mount_speed_mod : speed;
+    int16 startingSpeed = isMounted() ? 80 + map_config.mount_speed_mod : speed;
     // Mod::MOVE (169)
     // Mod::MOUNT_MOVE (972)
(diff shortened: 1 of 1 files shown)


TeoTwawki (changes requested):
Wrong. Wrong wrong wrong didn't read the commenting I left for you all over creation except somehow the one spot I should have:

I can PROVE that the move speed is actually 40 in client, but that you will "stride" twice as fast so 40 is not really 40 when mounted. The client itself handles this doubling effect. If the chocobo feels slow its because you are used to higher than retail walking speed on foot, and a separate setting exists to buff the bird.

image

Printed out retail speeds via addon.

m241dan:
Well, Mr. wrong wrong wrong, perhaps you should go read the comment in the conf file that says it should be 80 default and change that comment?

Jesus man.

TeoTwawki:
> Well, Mr. wrong wrong wrong, perhaps you should go read the comment in the conf file that says it should be 80 default and change that comment?
> 
> Jesus man.

Sorry. I've been over the speed shit to many times. I apologize apologize apologize apologize. But it really is 40 and moves like 80. so at 80 you'd be moving at 160. The conf may say 80 but its expecting everyone to think "twice as fast as walking" and if we tell them the base is 40 they will just think thats wrong anyway, so I've not cared what it technically said there for a long while.

PR #1537, opened 2022-04-10
Add textTable support to onZoneIn packet

claywar:
**_I affirm:_**

* Adjusts byte 40 in 0x0A (zone_in) packet to allow for text table swapping
* Converts luautils::OnZoneIn to void function, and sets event data directly
* Allows for table input to OnZoneIn to allow for a second parameter to change text table.  This was the simpler solution, since multiple returns from interaction functions would require significant refactors.

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -1653,7 +1653,7 @@ namespace luautils
      *                                                                       *
      ************************************************************************/
 
-    int32 OnZoneIn(CCharEntity* PChar)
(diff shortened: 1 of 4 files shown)


TeoTwawki (approved):
so long as existing scripts work as is (looks like they should), this is fine

claywar:
> so long as existing scripts work as is (looks like they should), this is fine

Tested with garbage data as well.  Working as expected for default values on table, and will always default to -1 if neither of the two conditions are valid.

PR #1545, opened 2022-04-12
[RUN] Implement Enspell damage of Rune Enhancement

WinterSolstice8:
After many, many hours of poking around doing research, I think it's done.
**_I affirm:_**
- [X] that I agree to LandSandBoat's [Limited Contributor License Agreement](https://github.com/LandSandBoat/server/blob/base/.github/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [X] that I have **read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)**
- [X] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

### DIFF
--- a/documentation/message.log
+++ b/documentation/message.log
@@ -226,7 +226,7 @@
 226 - The <player> uses .. .. TP drained from <target>.
 227 - The <player> casts (NULL). .. HP drained from <target>.
 228 - The <player> casts (NULL). .. MP drained from <target>.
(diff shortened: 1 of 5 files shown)


zach2good (approved):
Oh cool, you found a way around that thing I mentioned to you, nice 👍

PR #1564, opened 2022-04-14
[core] Replace setName/PacketName with renameEntity

zach2good:
**_I affirm:_**

Finally, we don't have to use the renamer addon anymore!

image
image

### DIFF
--- a/src/map/entities/baseentity.cpp
+++ b/src/map/entities/baseentity.cpp
@@ -29,6 +29,7 @@
 
 CBaseEntity::CBaseEntity()
 : status(STATUS_TYPE::DISAPPEAR)
+, isRenamed(false)
 {
     id       = 0;
     targid   = 0;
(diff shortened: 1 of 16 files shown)


src/map/entities/baseentity.cpp
```diff
@@ -83,6 +84,11 @@ const int8* CBaseEntity::GetName()
     return (const int8*)name.c_str();
 }
 
+const int8* CBaseEntity::GetPacketName()
```
claywar, line 87: One day, one day this will not be int8*

src/map/packets/entity_update.cpp
```diff
@@ -209,11 +209,11 @@ CEntityUpdatePacket::CEntityUpdatePacket(CBaseEntity* PEntity, ENTITYUPDATE type
     {
         this->setSize(0x48);
 
-        bool isNPC      = PEntity->objtype == TYPE_NPC;
         auto name       = PEntity->name;
         auto nameOffset = (PEntity->look.size == MODEL_EQUIPPED) ? 0x44 : 0x34;
 
-        if ((!isNPC && !PEntity->packetName.empty()) ||
```
claywar, line 216: Is the empty case still necessary?

claywar (approved):

zach2good:
Tested and happy with it (checked mobs, trusts, npcs etc.)

PR #1573, opened 2022-04-15
[geo] Core work for GEO basic functionality

Omnione:
A bit big this one, take ya time going through it, still wip with debug stuff in it,
mainly to get feedback on systems.

no more spells yet, just indi-regen, geo-regen, indi-posion and geo-poison scripted,

done alot of backend systems to make things work correctly,
can now place a luopan using the control button to set a new location as an offset from the target,
luopan positions now set in core,
potency and geocolre effects are all worked out from the spell id not in spell scripts as with model of the loupan.
entrust targeting flags now in place ready for the ability to be setup corectly,
skillups now work with handbell if a handbell is equipped when casting a geo/indi spell.
Many more changes.

Please note, as stated this is a work in pregress and still being worked on and is not in a state to be used yet,
If you do, please do not make any issues related to geo as it will be ignored.

Any questions welcome.

**_I affirm:_**


src/map/packet_system.cpp
```diff
@@ -735,6 +735,12 @@ void SmallPacket0x01A(map_session_data_t* const PSession, CCharEntity* const PCh
     // uint32 ID = data.ref<uint32>(0x04);
     uint16 TargID = data.ref<uint16>(0x08);
     uint8  action = data.ref<uint8>(0x0A);
+    position_t actionOffset =
+    {
+        data.ref<float>(0x10),
```
claywar, line 740: General question, are we guaranteed that float on all architectures will always be 4 bytes?  If not, we might want to pull this in as something else and cast.

zach2good: Yeah, that's a pretty safe assumption 👍

claywar (approved):

src/map/packet_system.cpp
zach2good: I think we can also rely on the endianness being correct, as we pull floats out of packets in other places

zach2good (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -10383,6 +10383,35 @@ bool CLuaBaseEntity::delLatent(uint16 condID, uint16 conditionValue, uint16 mID,
...
 }
 
+/************************************************************************
+ *  Function: getEquippedItemsMaxModValue()
+ *  Purpose : Returns the highest integer value of a specified Mod on all equiped items
+ *  Example : local maxValue = player:getEquippedItemsMaxModValue(xi.mod.GEOMANCY_BONUS)
+ *  Notes   :
+ ************************************************************************/
+
+int16 CLuaBaseEntity::getEquippedItemsMaxModValue(Mod modId)
```
TeoTwawki: kinda long name. `getMaxGearMod()` ?

TeoTwawki: related: do we mistakenly stack movement speed from gear? we must if we didn't already have a function to do this.

Omnione: Yep, gets clamped is all

Omnione: That’s a better name, I like getMaxGearMod, I’ll get that changed if others agree

zach2good:
Yup

src/map/lua/lua_baseentity.cpp
zach2good: Sounds good to me

Omnione: ok changed to getMaxGearMod()

TeoTwawki (approved):

PR #1580, opened 2022-04-15
[dbtool] Automatically pick up migrations

zach2good:
**_I affirm:_**

### DIFF
--- a/tools/dbtool.py
+++ b/tools/dbtool.py
@@ -6,6 +6,7 @@ import re
 import time
 import fileinput
 import shutil
+import importlib
 
 # Pre-flight sanity checks
 def preflight_exit():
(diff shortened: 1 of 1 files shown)


claywar (approved):

cocosolos:
At this point it probably doesn't matter much, but  `extend_mission_log` needs to be run before `mission_blob_extra` and `cop_mission_ids` or you could end up with some malformed mission logs. `os.scandir` reads in arbitrary order so you could rename `extend_mission_log.py` to something like `0extend_mission_log.py` then sort before processing with `sorted(os.scandir("migrations"), key=lambda e: e.name)`.

zach2good:
I did have some concerns about this, and considered prepending new
migrations with 1_, 2_ etc.

Ill probably end up doing it, just to be on the safe side

cocosolos:
AFAIK it's only that one migration that needs to run first since the other 2 rely on the size of the mission blob (this is admittedly not very safe or future proof, my bad), but I haven't really looked into the newer ones.

cocosolos:
Actually after looking at it again `extend_mission_log` needs to run first, then `mission_blob_extra`, then order doesn't really matter after those 2.

PR #1621, opened 2022-04-23
[core] changed how loading group items works

m241dan:
**_I affirm:_**

This could screw up people who have dropType=1 and dropId=0 type stuff, but LSB as a project does not have any and seems to want dropId's for groups to start at 1 (a sane number for an Id).

First, this fixes it so that unless you start with groupId 0,
which we don't do, you don't get an extra group (empty)

Second, this fixes it so that if SQL loads a groupId 2 first,
groupId 1 doesn't get groupId 2's droprate.

### DIFF
--- a/src/map/utils/itemutils.cpp
+++ b/src/map/utils/itemutils.cpp
@@ -564,11 +564,12 @@ namespace itemutils
                 {
                     uint8  GroupId   = (uint8)sql->GetIntData(4);
                     uint16 GroupRate = (uint16)sql->GetIntData(5);
-                    while (GroupId >= dropList->Groups.size())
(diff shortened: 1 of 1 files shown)


TeoTwawki (approved):
lgtm, hadn't thought of this the other day when looking at drop rate problems on groups. makes sense.

Future note: may do away with the dropType for groups and just check for non zero group ID, since we already start at 1 for all grouped things. Its a case of multiple cooks here and forgetting wtf we were doing. I added the drop type and made it a combine-able mask value way back in dsp, then another guy did the work to actually use it based on my notes. I plan to revisit this area of the codebase soon to add some features we don't yet have, will probably start a discussion thread when I am ready to start.

PR #1623, opened 2022-04-23
[core] added code to prevent modules for re-caching

m241dan:
**_I affirm:_**

This is seen when loading mobs. For example, loadMOBList() loops through all its mobs, each time it luautils::OnEntityLoad(PMob) once the Mobs stats are set. This in turn called CacheLuaObjectFromFile()
on the mob even if it has already loaded it.
    
Ultimately, this creates a bug for people who load modules something say the "Peryton" in Valley Of
Sorrows. When it loads the Peryton for the first time, it applies the override, and then marks that as
applied. However, once the second Peryton is loaded, it pulls directly from the file and overwrites its
cache value (which had been previously overwritten by the module, essentially undoing that). However,
when the "TryApplyLuaModules" goes to run that override again, the applied flag is already set to true
so it does not try to load it again. Bother...

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -674,7 +674,15 @@ namespace luautils
         {
             if (part == parts.back())
             {
-                table[sol::override_value][part] = file_result;
+                // if this isn't checked, you will re-load the same thing many times
+                // this generally isn't a problem unless you have applied on override
+                // from a module already, in this situation, the override gets overwritten.
+                bool not_cached_already = !table[part].valid();
+
+                if (not_cached_already)
+                {
+                    table[sol::override_value][part] = file_result;
+                }
             }
             else
             {


zach2good:
I've not got time to look at this properly, but my intention (in time) is to allow hot-reloading of cached functions, and module overrides in any order, at any time. Which would play out by re-caching whatever function, then re-applying the override on top afterwards.

m241dan:
> I've not got time to look at this properly, but my intention (in time) is to allow hot-reloading of cached functions, and module overrides in any order, at any time. Which would play out by re-caching whatever function, then re-applying the override on top afterwards.

Ah, that's a fair point. This will likely prevent hot reloading which is useful on the dev side of things. But not something I'd want in production.

Let me... poke at this some more. But, know that it is a problem for anything that is not singular in existence (ie, an NPC there is only one, a dynamis mob? Not so much...).

PR #1626, opened 2022-04-24
[geo] Added bolster and blaze of glory

Omnione:
Added the SP ability Bolster and the ability Blaze Of Glory,
Aslo some cleanup

https://www.bg-wiki.com/ffxi/Bolster
https://www.bg-wiki.com/ffxi/Blaze_of_Glory

**_I affirm:_**

### DIFF
--- a/scripts/globals/job_utils/geomancer.lua
+++ b/scripts/globals/job_utils/geomancer.lua
@@ -141,6 +141,9 @@ local potencyData =
     [xi.effect.GEO_HASTE]               = {divisor =  30.1,  minPotency = 2.4, maxPotency = 29.9, geoModMultiplier = 1.1},
 }
 
+-----------------------------------
+-- Ability Check Functions
+-----------------------------------
 xi.job_utils.geomancer.geoOnAbilityCheck = function(player, target, ability)
     if player:getPetID() == xi.pet.id.LUOPAN then
         return 0,0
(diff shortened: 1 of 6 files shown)


claywar (approved):

scripts/globals/job_utils/geomancer.lua
```diff
@@ -245,9 +285,57 @@ xi.job_utils.geomancer.spawnLuopan = function(player, target, spell)
...
+    local bolsterJP = target:getJobPointLevel(xi.jp.BOLSTER_EFFECT)
+
+    -- Luopan Geo effect
+    if target:getPet() and target:getPet():getPetID() == xi.pet.id.LUOPAN then
+        local pet            = target:getPet()
+        local geoPotency     = pet:getLocalVar("GEO_POTENCY")
+        local currentPotency = pet:getStatusEffect(xi.effect.COLURE_ACTIVE):getSubPower()
+        if currentPotency == geoPotency * 2 then -- will always be this value with Bolster or Blaze of Glory active
+            pet:getStatusEffect(xi.effect.COLURE_ACTIVE):setSubPower(geoPotency)
+            pet:setMod(xi.mod.REGEN_DOWN, pet:getMod(xi.mod.REGEN_DOWN) +bolsterJP)
```
Omnione: ah, need to space out the + and bolsterJP here

Omnione: ok done

src/map/utils/petutils.cpp
```diff
@@ -1602,11 +1602,19 @@ namespace petutils
         {
             PPet->SetMLevel(PMaster->GetMLevel());
             PPet->health.maxhp = (uint32)floor((250 * PPet->GetMLevel()) / 15);
-            PPet->health.hp    = PPet->health.maxhp;
+
+            if (PMaster->StatusEffectContainer->HasStatusEffect(EFFECT_BOLSTER))
+            {
+                uint8 bolsterJPVal = dynamic_cast<CCharEntity*>(PMaster)->PJobPoints->GetJobPointValue(JP_BOLSTER_EFFECT);
```
zach2good, line 1608: No harm here, and no need to change it, but:

`dynamic_cast` is only required over `static_cast` if you're going to check the result afterwards. Otherwise there's no additional benefit.
```cpp
// If thing wasn't valid for casting, PChar will be nullptr
auto* PChar = dynamic_cast<CCharEntity*>(thing);
if (PChar)
{
    // Only possible to come in here if the dynamic_cast was successful
}

...

// static_cast is pretty much stuffing bits into a CCharEntity ptr, so it might "succeed" even if thing wasn't valid
auto* PChar = static_cast<CCharEntity*>(thing);
if (PChar)
{
    // We can sometimes make it in here, even if we aren't supposed to!
}
```

Note that in the static_cast example, we do this when dealing with Allies as Pets, and we really shouldn't be. But the code relies on that so 🤷

zach2good (approved):

src/map/utils/petutils.cpp
Omnione: ah ok ty for the examples

PR #1628, opened 2022-04-25
Add Auto Regen at level 25 to Carbuncle pet

RunAwayDSP:
https://ffxiclopedia.fandom.com/wiki/Carbuncle
"Gains Auto Regen at level 25."

**_I affirm:_**

### DIFF
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -1462,6 +1462,15 @@ namespace petutils
             {
                 ((CItemWeapon*)PPet->m_Weapons[SLOT_MAIN])->setDelay((uint16)(floor(1000.0 * (280.0f / 60.0f))));
             }
+  
+            if (PetID == PETID_CARBUNCLE)
+            {
+                if (PPet->GetMLevel() >= 25)
+                {
+                    PPet->setModifier(Mod::REGEN, 1);
+                }
+            }
+
 
             // In a 2014 update SE updated Avatar base damage
             // Based on testing this value appears to be Level now instead of Level * 0.74f


src/map/utils/petutils.cpp
```diff
@@ -1462,6 +1462,15 @@ namespace petutils
...
                 ((CItemWeapon*)PPet->m_Weapons[SLOT_MAIN])->setDelay((uint16)(floor(1000.0 * (280.0f / 60.0f))));
             }
+  
+            if (PetID == PETID_CARBUNCLE)
+            {
+                if (PPet->GetMLevel() >= 25)
+                {
+                    PPet->setModifier(Mod::REGEN, 1);
+                }
+            }
```
zach2good, line 1472: This can be simplified to:
```cpp
if (PetID == PETID_CARBUNCLE && PPet->GetMLevel() >= 25)
{
    PPet->setModifier(Mod::REGEN, 1);
}
```

zach2good (commented):
Does the amount of regen here increase by 75 or 99?

src/map/utils/petutils.cpp
RunAwayDSP: There is no mention of the regen getting stronger at higher levels.
I left it expanded in case there is any other carbuncle mods that also need to be added.

Xaver-DaRed:
Carbuncle is believed to be a WHM, witch gains Auto-Regen I at 25 and Auto-Regen II at 76.
I think job traits should be handling this by setting it's job to WHM, instead of hard-coding it.

RunAwayDSP:
> Carbuncle is believed to be a WHM, witch gains Auto-Regen I at 25 and Auto-Regen II at 76. I think job traits should be handling this by setting it's job to WHM, instead of hard-coding it.

There is no confirmation of this, only auto-regen 1 at 25. In addition there are no other whm traits that have been mentioned for carbuncle. 

But if we can get some kind of confirmation that they get other job traits, or carby getting higher regen at higher levels the job trait coding would be better

Omnione:
done some testing on retail and this is what i found,
as we know,
Auto Regen I is 1 HP/tick
Auto Regen II is 2 HP/tick

from testing i got this:

I and a few others tested carby's HP% over time and coalted the data
tested from 90% onwards at level 99 with no pet regen gear
All data was from packets
this is assuming carby has what we belive to be 1600 Max HP 

(100% = 1600) 80 ticks
(99%  = 1584) 72 ticks
(98%  = 1568) 64 ticks
(97%  = 1552) 56 ticks
(96%  = 1536) 48 ticks
(95%  = 1520) 40 ticks
(94%  = 1504) 32 ticks
(93%  = 1488) 24 ticks
(92%  = 1472) 16 ticks
(91%  = 1456) 8  ticks
(90%  = 1440)
diff for each % is 16 hp
16 HP recovered from 8 ticks over 24 seconds
16 / 8 = 2
HP needed to refil from 90% -> 100% is 160
160 HP/ 80 ticks = 2
so carby should have 2 hp a tick regen
So, Auto Regen II 76 onwards would be a good guess like a WHM
hope this helps

src/map/utils/petutils.cpp
TeoTwawki: even simpler is to see if he has MDB and if he does, set mainjob to whm because 2 traits being whm traits is evidence enough for me.

TeoTwawki: > There is no mention of the regen getting stronger at higher levels. I left it expanded in case there is any other carbuncle mods that also need to be added.

the code snippet isn't scaling anything like that, its just combining your checks

RunAwayDSP:
> done some testing on retail and this is what i found, as we know, Auto Regen I is 1 HP/tick Auto Regen II is 2 HP/tick
> 
> from testing i got this:
> 
> I and a few others tested carby's HP% over time and coalted the data tested from 90% onwards at level 99 with no pet regen gear All data was from packets this is assuming carby has what we belive to be 1600 Max HP
> 
> (100% = 1600) 80 ticks (99% = 1584) 72 ticks (98% = 1568) 64 ticks (97% = 1552) 56 ticks (96% = 1536) 48 ticks (95% = 1520) 40 ticks (94% = 1504) 32 ticks (93% = 1488) 24 ticks (92% = 1472) 16 ticks (91% = 1456) 8 ticks (90% = 1440) diff for each % is 16 hp 16 HP recovered from 8 ticks over 24 seconds 16 / 8 = 2 HP needed to refil from 90% -> 100% is 160 160 HP/ 80 ticks = 2 so carby should have 2 hp a tick regen So, Auto Regen II 76 onwards would be a good guess like a WHM hope this helps

Thank you! Yes it appears avatard get there job traits so this code would be incorrect 

Instead need to apply job traits to each avatar instead

PR #1632, opened 2022-04-26
[module] Add claim_shield module

zach2good:
**_I affirm:_**

We'll need `m_IsClaimable` later down the line for things like DI etc., this just adds it and uses it for a lua module claimshield. Seems to work nicely. Avoids any core changes

### DIFF
--- a/modules/custom/claim_shield.lua
+++ b/modules/custom/claim_shield.lua
@@ -1,0 +1,93 @@
+-----------------------------------
+-- Claim Shield
+-----------------------------------
+require("modules/module_utils")
+require("scripts/globals/msg")
+require("scripts/globals/utils")
+-----------------------------------
+local m = Module:new("claim_shield")
+m:setEnabled(false)
+
+local claimshieldTime = 7500
+
+-- NOTE: These names are as they are as filenames.
+-- Example: Behemoth's Dominion => Behemoths_Dominion
+-- Example: King Behemoth       => King_Behemoth
+-- { zone name, mob name }
+-- Format:
+local nmsToShield =
+{
+    { "Attohwa_Chasm", "Citipati" },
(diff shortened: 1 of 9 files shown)


modules/custom/claim_shield.lua
```diff
@@ -0,0 +1,93 @@
+-----------------------------------
+-- Claim Shield
+-----------------------------------
+require("modules/module_utils")
+require("scripts/globals/msg")
+require("scripts/globals/utils")
+-----------------------------------
+local m = Module:new("claim_shield")
+m:setEnabled(true)
```
claywar: Should we default to enabled=false for custom modules?

src/map/lua/lua_baseentity.cpp
```diff
@@ -9819,6 +9822,39 @@ sol::table CLuaBaseEntity::getNotorietyList()
...
+/************************************************************************
+ *  Function: setClaimable(...)
+ *  Purpose : sets m_IsClaimable for a mob
+ *  Example : mob:setClaimable(false)
+ *  Notes   :
+ ************************************************************************/
+
+void CLuaBaseEntity::setClaimable(bool claimable)
+{
+    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
```
claywar: Do we want to continue to use breaks in these bindings, or return gracefully and log?

claywar (approved):

modules/custom/claim_shield.lua
zach2good: Whoops, yes

src/map/lua/lua_baseentity.cpp
zach2good: Yeah, agreed

PR #1633, opened 2022-04-26
separate mobs stats out of CalculateStats into CalculateMobStats, add argument to optionally prevent refilling mobs hp/mp.…

TeoTwawki:
…  By default all behavior stays as is.

**_I affirm:_**
- [:crossed_fingers: ] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits
Didn't thoroughly test yet. assistance with that appreciated, as my time is very limited.

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -551,7 +551,7 @@ void CMobEntity::Spawn()
     SetMLevel(level);
     SetSLevel(level); // calculated in function
 
-    mobutils::CalculateStats(this);
+    mobutils::CalculateMobStats(this);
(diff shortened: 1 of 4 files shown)


zach2good (approved):
This is to enable mob levelling up/down in combat yeah?

TeoTwawki:
> This is to enable mob leveling up/down in combat yeah?

function can be used at pretty much any time. some custom content on my own sues uses it to scale the mob to whoever spawned it/claimed it. that part already existed, but would always full heal the mob. The extra parameter makes the heal part optional, but still happens by default.

I just wrapped the hp/mp parts in a check for the param.

PR #1635, opened 2022-04-26
[modules] Add renamer module

zach2good:
**_I affirm:_**

Yet another thing to try and replace the renamer addon. Or at least to remove the need to distribute your own renamer lists to your users. This uses custom packets and handlers defined in a cpp module to handle requests from the client, which is also using a custom addon to ask for and receive information.


zach2good:
Moving these changes to https://github.com/LandSandBoat/server/pull/1647

PR #1638, opened 2022-04-27
[debug] Re-add LuaJIT special exception handling

zach2good:
**_I affirm:_**

I've seen repeated `0xE24C4A02` exception crashes downstream. I _thought_ that Wheaty handled all of this, but it seems as though it doesn't. This re-adds the special block we had to just before Wheaty starts handling things.


zach2good:
For expediency, we used to run this, and Eden is currently running this, so I think it's a pretty safe snippet.

claywar (approved):

PR #1641, opened 2022-04-28
Set BB with unkillable flag

soapytoast:
Set BB with unkillable flag
Add DB entry for buccaneer knife adding additional effect and set to 100% chance
Add additional effect to remove unkillable flag if BB is hit with buccaneer knife.

**_I affirm:_**


Xaver-DaRed:
Please read the style guide and format the code properly.
Also, remove the additional, unrelated script. !fafnir could be renamed, but we dont need an aditional copy of an already existing script

scripts/globals/additional_effects.lua
```diff
@@ -120,7 +120,8 @@ xi.additionalEffect.attack = function(attacker, defender, baseAttackDamage, item
         ABSORB_STATUS = 10,
         SELF_BUFF = 11,
         DEATH = 12,
-    }
+		REMOVE_UNKILLABLE = 13, --Solely for Buccaneer's Knife with Brigandish Blade
```
claywar, line 123: Should we really be adding an AE for a single mob?  This seems like something that could more elegantly be handled elsewhere.

soapytoast:
sorry issues with my branching and committing. will remake the pull req.

PR #1643, opened 2022-04-28
Brigandish Blade unkillable effect

soapytoast:
Set BB with unkillable flag
Add DB entry for buccaneer knife adding additional effect and set to 100% chance
Add additional effect to remove unkillable flag if BB is hit with buccaneer knife.

**_I affirm:_**


soapytoast:
If there is a better way to handle this please let me know, but as far as we can tell it can't be handled on the mob directly so I am not sure where else we would handle this.

TeoTwawki:
> If there is a better way to handle this please let me know, but as far as we can tell it can't be handled on the mob directly so I am not sure where else we would handle this.

I'd go with a listener to trigger instead. could possibly make a mixin to have recyclable code for mobs in case someday there is another mob that can only be killed by another different item. We used to handle sky gods and winds stripping their enspell damage the way you are trying to do brigandish blade here, before the code got global'd. it got removed and hasn't been redone yet.

p.s. next time your pr gets all fubar chances are we can help you fix it without having to close it and open another. just let us know you've hit a dead end. the worst that happens is you wind up doing it again anyway, but we might be able to rescue it.

p.p.s. this IS how you'd add more additional effects though :D except for that extra end

eraffxi-personal:
> > If there is a better way to handle this please let me know, but as far as we can tell it can't be handled on the mob directly so I am not sure where else we would handle this.
> 
> I'd go with a listener to trigger instead. could possibly make a mixin to have recyclable code for mobs in case someday there is another mob that can only be killed by another different item. We used to handle sky gods and winds stripping their enspell damage the way you are trying to do brigandish blade here, before the code got global'd. it got removed and hasn't been redone yet.
> 
> p.s. next time your pr gets all fubar chances are we can help you fix it without having to close it and open another. just let us know you've hit a dead end. the worst that happens is you wind up doing it again anyway, but we might be able to rescue it.

I fucked up my whole repo and just remade it, still learning the right way to handle the PR's. 
We will look into mixin's still learning all the changes to LSB here.

TeoTwawki:
> I fucked up my whole repo and just remade it, still learning the right way to handle the PR's.

Anytime you need help unfucking things, let me know if I am free I may be able to help :+1: I think I have seen almost every way imaginable to break a repo by now. Sometimes just starting over is faster, but I usually can roll back anything that has happened.

eraffxi-personal:
In regards to the additional effect aspect, part of why we went with this method is because the weapon does actually have an additional effect of water dmg as well, we just didn't have solid numbers on that currently and had not added those yet.

TeoTwawki:
> In regards to the additional effect aspect, part of why we went with this method is because the weapon does actually have an additional effect of water dmg as well, we just didn't have solid numbers on that currently and had not added those yet.

I know Bartholomew's has additional effect water (pretty dang strong one too), but to my knowledge buccaneer's doesn't.

eraffxi-personal:
https://ffxiclopedia.fandom.com/wiki/Buccaneer%27s_Knife per here it does specifically when fighting BB

TeoTwawki:
> https://ffxiclopedia.fandom.com/wiki/Buccaneer%27s_Knife per here it does specifically when fighting BB

huh. other wiki didn't even mention, and I'd never noticed. still, we can handle that as a latent if need be. I'm curious if it procs on any other mobs in game (weapons perhaps). I think I'll go get one and find out. To limit it to BB, we can make sure he's his own family and that can be the trigger.

zach2good:
Ignoring the additional effect, here is how I'd handle the unkillable thing:
```lua
    -- TODO: This should be damage immunity, rather than being unkillable.
    --     : Replace this with that when it is available.
    mob:setUnkillable(true)
    mob:addListener("TAKE_DAMAGE", "BRIGANDISH_BLADE_TAKE_DAMAGE", function(mob, amount, attacker, attackType, damageType)
        if
            attackType == xi.attackType.PHYSICAL and
            attacker:getObjType() == xi.objType.PC and
            attacker:getEquipID(xi.slot.MAIN) == xi.items.BUCCANEERS_KNIFE
        then
            mob:setUnkillable(false)
        end
    end)
end
```

PR #1647, opened 2022-04-29
Reorganize Modules and Module loading, more example modules, some fixes

zach2good:
**_I affirm:_**


zach2good:
Ready to go

modules/example/lua/example_lua_module.lua
```diff
@@ -0,0 +1,14 @@
...
+-- Old setting from settings.lua: HOMEPOINT_HEAL
+-- Set if you want Home Points to heal you like in single-player Final Fantasy games.
+-----------------------------------
+require("modules/module_utils")
+require("scripts/globals/homepoint")
+-----------------------------------
+local m = Module:new("homepoint_heal")
+
+m:addOverride("xi.homepoint.onTrigger", function(player, csid, index)
+    super(player, csid, index)
```
claywar: Just noticed this, but looks like this module doesn't apply any additional function (not sure if we want to have homepoint heal actually implemented)

modules/init.txt
```diff
@@ -0,0 +1,40 @@
...
+# | custom/lua/claim_shield.lua
+# |
+# ---------------------
+#
+# Will load only custom/cpp/custom_module.cpp and custom/lua/claim_shield.lua.
+#
+# Live example:
+
+custom/lua/test_npcs_in_gm_home.lua
+example/cpp/example_cpp_module.cpp
```
claywar: Should this be enabled by default?

modules/renamer/lua/list.lua
```diff
@@ -0,0 +1,814 @@
...
+        { 17216287, "799" }, -- Slime Tarou
+        { 17216288, "800" }, -- blank
+        { 17216289, "801" }, -- blank
+        { 17216290, "802" }, -- Orgis
+        { 17216291, "803" }, -- blank
+        { 17216292, "804" }, -- ???
+        { 17216293, "805" }, -- blank
+        { 17216511, "1023" },
+    },
+}
```
claywar, line 813: Quite the list for this.  Why the numeric renames?

scripts/globals/player.lua
```diff
@@ -41,10 +41,13 @@ local startingJobGear =
...
 -----------------------------------
--- local functions
+-- public functions
 -----------------------------------
 
-local function CharCreate(player)
+xi = xi or {}
+xi.player = {}
+
+xi.player.charCreate = function(player)
```
claywar, line 50: 👍

src/map/lua/luautils.cpp
```diff
@@ -1685,10 +1685,11 @@ namespace luautils
     {
         TracyZoneScoped;
 
-        auto name = PChar->m_moghouseID ? "Residential_Area" : (const char*)zoneutils::GetZone(PChar->loc.destination)->GetName();
+        auto name     = PChar->m_moghouseID ? "Residential_Area" : (const char*)zoneutils::GetZone(PChar->loc.destination)->GetName();
+        auto filename = fmt::format("./scripts/zones/{}/Zone.lua", name);
 
         auto onZoneInFramework = lua["xi"]["globals"]["interaction"]["interaction_global"]["onZoneIn"];
-        auto onZoneIn = lua["xi"]["zones"][name]["Zone"]["onZoneIn"];
+        auto onZoneIn = GetCacheEntryFromFilename(filename)["onZoneIn"];
```
claywar, line 1692: Assuming this is to support modules loading beforehand?  Just wondering, since we seem to load these pretty early

src/map/packet_system.cpp
```diff
@@ -315,12 +315,15 @@ void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PCh
...
     {
-        if (PChar->loc.zone != nullptr)
-        {
-            ShowWarning("Client cannot receive packet or key is invalid: %s", PChar->GetName());
-        }
+        ShowWarning("Client cannot receive packet or key is invalid: %s, Zone: %s (%i)",
+            PChar->GetName(), PChar->loc.zone->GetName(), PChar->loc.zone->GetID());
+
+        // Write a sane location for them
+        auto newZone = PChar->loc.prevzone ? PChar->loc.prevzone : ZONE_VALKURM_DUNES;
```
claywar, line 324: Interesting choice of default

claywar (approved):

modules/example/lua/example_lua_module.lua
zach2good: Good catch, I'm not sure what to leave in the Lua example, if anything

modules/init.txt
zach2good: I suppose not, can remove

modules/renamer/lua/list.lua
zach2good: This was the big test list Teo had in his renamed repo, makes for good stress testing

src/map/lua/luautils.cpp
zach2good: This is mainly to stop the lua panic+crash that happens if you try to zone into "unknown", then this chain of accesses will explode: `auto onZoneIn = lua["xi"]["zones"][name]["Zone"]["onZoneIn"];`

src/map/packet_system.cpp
zach2good: So, this doesn't nicely forward someone out of death into a real zone, but when they eventually get bored of their black loading screen and restart the game, they'll appear in a sane zone

TeoTwawki: We normally used 214 for that, which is basically the jeuno mog house. we have some magic in core attached to it that will dump you to a valid prevZone on exit because we direct you landed there "abnormally".

its kinda obscure you'd prolly never notice it without me telling the story. :)

modules/renamer/lua/list.lua
TeoTwawki: Actually what it was originally made for too - at the time I wasn't sure if ashita's onRender() was fast enough for all 0-1023 dat index slots in a single rendering frame (it was, and then some)

PR #1648, opened 2022-04-29
[sql] ModID 370 (Regen) with a negative power value changed to ModID 404 (Regen Down) with positive power value

EpicTaru:
This adjustment is only in item_latents.sql.  
-This is so that items that have a latent effect of draining HP wake the player up if the player has sleep status effect.

Also checked and verified no above uses in item_mods.sql/item_mods_pet.sql.

**_I affirm:_**


Omnione (approved):
Looks good
Frenzy Sallet changed to use mod regen down rather than - regen
Berserker's Torque using the same

zach2good (approved):

PR #1659, opened 2022-05-02
Various improvements after stress testing

zach2good:
**_I affirm:_**

Thanks Wings 3.0 for the stress  testing, we learned a bunch of things.

This is a mixed bag of things, split into sane commits.

- More things for tracy
- Comment out the campaign map packet logic - it isn't safe across processes!
- Add sql->Async. I thought this would make the command handler faster, but it seems like it's slow because of `stringstream` and friends :(


The-Aerec:
Can you ping me when this is ready? And we will get the box updated for second round of testing this weekend. Also should we come up with a 'Set of tests' that we should do as to stress in particular ways?

claywar (approved):

PR #1660, opened 2022-05-02
[core] Add utils to translate auto-translate bytes

zach2good:
**_I affirm:_**

Once this is done, we can log chat and gm audits in plantext instead of blobs!


zach2good:
Ready to go, data entry can happen some other time. This fixes gm_audit.

claywar (approved):

PR #1662, opened 2022-05-03
[Fix] Banishing Gates interactions

Xaver-DaRed:
**_I affirm:_**

Fixes to gates behaviour and interaction. Left a TODO regarding the lever behaviour.
Fixed possible message spam when opening a gate with lstopnes KI and repeatedly steping on a lever.
Please check Banishing Gate 3, since I left a very good, but not retail accurate case of "Better than retail".


claywar (approved):

scripts/zones/Garlaige_Citadel/npcs/_5ki.lua
```diff
@@ -11,18 +12,25 @@ entity.onTrade = function(player, npc, trade)
...
+    else
+        -- South side EXPECTED regular interaction.
+        if player:getZPos() < 359 then
+            player:messageSpecial(ID.text.YOU_COULD_OPEN_THE_GATE, xi.ki.POUCH_OF_WEIGHTED_STONES)
         end
-        player:messageSpecial(ID.text.BANISHING_GATES + 2) -- Third Banishing gate opening
-        return 1
     end
+
+    return 1
```
claywar: This can be removed (returning 1 has no impact in onTrigger)

PR #1668, opened 2022-05-05
[Quest] Limit Break 1 and 2 KI cleanup and other fixes

Xaver-DaRed:
**_I affirm:_**

Changes:
- Add missing message after LB1 quest completion (But before next quest is available)
- Add missing message after LB2 quest completion (But before next quest is available)
- Add cleanup method on quest complete

### DIFF
--- a/scripts/quests/jeuno/LB01_In_Defiant_Challenge.lua
+++ b/scripts/quests/jeuno/LB01_In_Defiant_Challenge.lua
@@ -31,6 +31,25 @@ local garlaigeID = require('scripts/zones/Garlaige_Citadel/IDs')
 
 local quest = Quest:new(xi.quest.log_id.JEUNO, xi.quest.id.jeuno.IN_DEFIANT_CHALLENGE)
 
+-- Key Item removals
+local function cleanKIMold(player)
+    player:delKeyItem(xi.ki.EXORAY_MOLD_CRUMB1)
+    player:delKeyItem(xi.ki.EXORAY_MOLD_CRUMB2)
+    player:delKeyItem(xi.ki.EXORAY_MOLD_CRUMB3)
+end
+
+local function cleanKICoal(player)
+    player:delKeyItem(xi.ki.BOMB_COAL_FRAGMENT1)
+    player:delKeyItem(xi.ki.BOMB_COAL_FRAGMENT2)
+    player:delKeyItem(xi.ki.BOMB_COAL_FRAGMENT3)
+end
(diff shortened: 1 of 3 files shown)


claywar (approved):

scripts/quests/jeuno/LB01_In_Defiant_Challenge.lua
```diff
@@ -282,7 +298,7 @@ quest.sections =
             {
                 onTrigger = function(player, npc)
                     if
-                        not player:hasItem(xi.items.PIECE_OF_ANCIENT_PAPYRUS) and
+                        not player:findItem(xi.items.PIECE_OF_ANCIENT_PAPYRUS) and
```
TeoTwawki: I would just like to note this is effectively the same thing here.

has item just returns true or false, finditem returns the actual item object that we aren't using here. so either way this check will be true if the item is found and false if not. I've noted a lot of the scripts were similarly switched over to finditem, but then the item object isn't used. doesn't hurt anything so far as I know. and at this point the only reason to need one or over the other is the 2nd param of has item, which enables specifying a single container to check instead of all player inventory containers  (which almost never is done, very few quests care).

I'm wondering if we shouldn't just combine these back to one function with both capabilities.

TeoTwawki (approved):

scripts/quests/jeuno/LB01_In_Defiant_Challenge.lua
claywar: Odd, I thought hasItem only searched one inventory

TeoTwawki: > Odd, I thought hasItem only searched one inventory

only if you feed it the 2nd arg for what container to search.

incidentally If we do combine them later, I prefer has over find naming convention wise. 

the binding calls the charutils function which has a loop:
```cpp
    bool HasItem(CCharEntity* PChar, uint16 ItemID)
    {
        for (uint8 LocID = 0; LocID < CONTAINER_ID::MAX_CONTAINER_ID; ++LocID)
        {
            if (PChar->getStorage(LocID)->SearchItem(ItemID) != ERROR_SLOTID)
            {
                return true;
            }
        }
        return false;
    }
```

Xaver-DaRed: Reverted those changes. I like "hasItem" more than "findItem" too.

PR #1672, opened 2022-05-05
[Battlefields] Fix exiting when a wipe happens and implement related messages

Xaver-DaRed:
**_I affirm:_**

Changes:
- Fixes battlefields not waiting 3 full minutes after last party member dies before kicking you out.
- Implement wipe and 3-minute-reminder messages.


zach2good:
Are there multiple kickout reminders? Because there are multiple kickout messages (that include seconds):
Boneyard Gullly from Dec 2021:
```
7408 : Everybody's fallen in room E, and they can't get up! Leaving room E!
7409 : If all party members' HP are still zero after # minute[/s] # second[/s], the party will be removed from the battlefield.
7410 : If all party members' HP are still zero after # minute[/s], the party will be removed from the battlefield.
```

EDIT: Oh, my bad, that's minutes and minutes+seconds

EpicTaru:
> [Battlefields] Fix kick-out

And here I was thinking for some reason that something was wrong with the TP move **Kick Out** for the behemoths in Limbus, stupid brains....

claywar (approved):

PR #1673, opened 2022-05-06
[build] Update a bunch of "useless casts", make compiler happy

WinterSolstice8:
* Note, this commit doesn't fix the build entirely for GCC
dependencies still error. Even when build is successful, topaz_game crashes on login.
Please scrutinize this commit to the furthest extent possible.

**_I affirm:_**
- [☠️] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -192,7 +192,7 @@ void PrintPacket(CBasicPacket data)
         // TODO: -Wno-format-overflow - writing between 4 and 53 bytes into destination of 50
         // TODO: FIXME
         // cppcheck-suppress sprintfOverlappingData
-        snprintf(message, sizeof(message), "%s %02hhx", message, *((uint8*)data[(const int)y]));
+        snprintf(message, sizeof(message), "%s %02hhx", message, *((uint8*)data[(int)y]));
         if (((y + 1) % 16) == 0)
         {
             message[48] = '\n';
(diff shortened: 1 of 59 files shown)


src/map/packet_system.cpp
```diff
@@ -3825,7 +3827,12 @@ void SmallPacket0x064(map_session_data_t* const PSession, CCharEntity* const PCh
         return;
     }
 
-    memcpy(&PChar->keys.tables[KeyTable].seenList, data[0x08], 0x40);
+    // memcpy(&PChar->keys.tables[KeyTable].seenList, data[0x08], 0x40);
+    // TODO: verify this works?
+    for (int i = 0; i < 0x40; i++)
+    {
+        PChar->keys.tables[KeyTable].seenList[i] = data[0x08 + i];
```
WinterSolstice8, line 3834: Please double check this. No idea if this is the correct syntax, memset specifically was called out as bad to use against std::bitset by gcc.

zach2good: https://stackoverflow.com/questions/18513259/how-to-initialize-a-bitset-type-all-bits-to-1

Looks like `set` is a good option here for `bitset`

WinterSolstice8: Looking at it more, memset here was copying 64 bytes and this for loop is only setting 64 bits. So this implementation is definitely wrong

zach2good (approved):
This is good work, I'm going to merge this in (since it's going into my other working branch, and it'll need a lot more work anyway)!

PR #1689, opened 2022-05-09
[Interaction] Bastok 3-2 fixes

Xaver-DaRed:
**_I affirm:_**

Changes:
- Removed distance check on trade.
- Changed gateguards message to event

### DIFF
--- a/scripts/missions/bastok/3_2_To_the_Forsaken_Mines.lua
+++ b/scripts/missions/bastok/3_2_To_the_Forsaken_Mines.lua
@@ -17,11 +17,7 @@ require('scripts/settings/main')
 require('scripts/globals/interaction/mission')
 require('scripts/globals/zone')
 -----------------------------------
-local bastokMarketsID = require('scripts/zones/Bastok_Markets/IDs')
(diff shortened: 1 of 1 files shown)


scripts/missions/bastok/3_2_To_the_Forsaken_Mines.lua
```diff
@@ -137,7 +133,6 @@ mission.sections =
             {
                 onTrade = function(player, npc, trade)
                     if
-                        player:checkDistance(npc) <= 1.6 and
```
claywar, line 140: This should not be deleted, and was tested.

Xaver-DaRed: I performed the trade succesfuly at arround 5 yalms of distance
Since its a repeatable mission, i can do a video as proof though, if needed.

claywar: Possible it changed, but this was the case in the past.

claywar (approved):

PR #1690, opened 2022-05-09
Implement Beadeaux Afflictor messages

Xaver-DaRed:
Co-authored-by: KnowOne134 <35616771+KnowOne134@users.noreply.github.com>

**_I affirm:_**

As title suggest.
Animation when getting afflicted by curse not impleneted

Many thanks to KnowOne and Eden peeps^^


scripts/zones/Beadeaux/Zone.lua
```diff
@@ -16,13 +16,20 @@ require("scripts/globals/zone")
...
 zone_object.onInitialize = function(zone)
-    -- Regions 1-6 are for the Afflictor System (RegionID, X, Radius, Z)
-    zone:registerRegion(1, -163, 10, -137, 0, 0, 0)
-    zone:registerRegion(2, -209, 10, -131, 0, 0, 0)
-    zone:registerRegion(3, -140, 10,   20, 0, 0, 0)
-    zone:registerRegion(4,  261, 10,  140, 0, 0, 0)
-    zone:registerRegion(5,  340, 10,  100, 0, 0, 0)
-    zone:registerRegion(6,  380, 10,   60, 0, 0, 0)
+    -- The Afflictor System (RegionID, X, Radius, Z) for curse
+    zone:registerRegion( 1, -163, 15, -137, 0, 0, 0) -- 17379801 The Afflictor
```
claywar: Would suggest not using NPC ID, but some other method for locating it, since these are subject to change

scripts/zones/Beadeaux/Zone.lua
```diff
@@ -50,17 +57,66 @@ zone_object.onConquestUpdate = function(zone, updatetype)
 end
 
 zone_object.onRegionEnter = function(player, region)
-    if region:GetRegionID() <= 6 then
-        if not player:hasStatusEffect(xi.effect.CURSE_I) and not player:hasStatusEffect(xi.effect.SILENCE) then
-            player:addStatusEffect(xi.effect.CURSE_I, 50, 0, 300)
-            if player:getQuestStatus(xi.quest.log_id.BASTOK, xi.quest.id.bastok.THE_CURSE_COLLECTOR) == QUEST_ACCEPTED and player:getCharVar("cCollectCurse") == 0 then
-                player:setCharVar("cCollectCurse", 1)
+    local regionID = region:GetRegionID()
+    local pos      = player:getYPos()
```
claywar: Can we change this var name to yPos to be more specific?  This happens below as well.

scripts/zones/Beadeaux/Zone.lua
```diff
@@ -50,17 +57,66 @@ zone_object.onConquestUpdate = function(zone, updatetype)
 end
 
 zone_object.onRegionEnter = function(player, region)
-    if region:GetRegionID() <= 6 then
```
claywar, line 53: Why are these quest options being removed or commented out?  It appears that some sort of implementation exists.

Xaver-DaRed: I was capturing the quest and planned on converting it. I can leave it there for the time being though

scripts/zones/Beadeaux/Zone.lua
TeoTwawki: I recommend identifying in the comment by map sector like `3rd Afflictor (J-7)` or w/e its placement is.

claywar (approved):

PR #1692, opened 2022-05-09
[Interaction] Bastok 3-3 fixes

Xaver-DaRed:
**_I affirm:_**

Changes:
- Change gate guard messages to events.
- Made last event with Goggehn non-mandatory, becouse it isn't. After rescuing Senator, you can interact with door directly.

### DIFF
--- a/scripts/missions/bastok/3_3_Jeuno.lua
+++ b/scripts/missions/bastok/3_3_Jeuno.lua
@@ -18,18 +18,14 @@ require('scripts/settings/main')
 require('scripts/globals/interaction/mission')
 require('scripts/globals/zone')
 -----------------------------------
-local bastokMarketsID = require('scripts/zones/Bastok_Markets/IDs')
-local bastokMinesID   = require('scripts/zones/Bastok_Mines/IDs')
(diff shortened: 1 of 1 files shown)


scripts/missions/bastok/3_3_Jeuno.lua
```diff
@@ -171,7 +163,7 @@ mission.sections =
             ['_542'] =
             {
                 onTrade = function(player, npc, trade)
-                    -- TODO: As of the April 2009 update, trading the key to any Cermet Door or the elevator will turn
```
claywar, line 174: This still should be marked as a TODO, as it is not implemented at this time (or at least fully implemented)

Xaver-DaRed: While true, thats not part of the actual mission logic, as it can be done at any time.
Also, I'm pretty sure we already handle it

claywar: After additional searching, yep, there's some refactoring to do, but its in place.

claywar (approved):

PR #1697, opened 2022-05-10
Improve Char and Entity update packet flow, async-ify a number of db calls, more info for Tracy

zach2good:
**_I affirm:_**

It would be worth going through each commit, as they're somewhat sane chunks of work. 

Huge thanks to InoUno and Eden for their packet work, it's truly a lightspeed step forward!

### DIFF
--- a/scripts/globals/trust.lua
+++ b/scripts/globals/trust.lua
@@ -232,6 +232,11 @@ xi.trust.onTradeCipher = function(player, trade, csid, rovCs, arkAngelCs)
 end
 
 xi.trust.canCast = function(caster, spell, not_allowed_trust_ids)
+    -- GMs can do what they want
+    if caster:getGMLevel() > 0 and caster:checkNameFlags(0x04000000) then
+        return 0
+    end
+
     -- Trusts must be enabled in settings
     if xi.settings.ENABLE_TRUST_CASTING == 0 then
         return xi.msg.basic.TRUST_NO_CAST_TRUST
--- a/scripts/globals/trust.lua
+++ b/scripts/globals/trust.lua
@@ -232,14 +232,14 @@ xi.trust.onTradeCipher = function(player, trade, csid, rovCs, arkAngelCs)
 end
 
 xi.trust.canCast = function(caster, spell, not_allowed_trust_ids)
-    -- GMs can do what they want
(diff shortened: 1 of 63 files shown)


scripts/globals/trust.lua
```diff
@@ -232,6 +232,11 @@ xi.trust.onTradeCipher = function(player, trade, csid, rovCs, arkAngelCs)
 end
 
 xi.trust.canCast = function(caster, spell, not_allowed_trust_ids)
+    -- GMs can do what they want
+    if caster:getGMLevel() > 0 and caster:checkNameFlags(0x04000000) then
```
TeoTwawki: hey, some of us have minimalist GM's we need to have a flag on but don't give access to a lot of powerful commands on. GM commands give access to trusts if they are enabled. if they are disabled NOBODY server wide should be casting. this should be below line 241-243 for this reason.

src/map/entities/baseentity.h
```diff
@@ -158,18 +158,29 @@ enum UPDATETYPE
...
     UPDATE_ALL_MOB  = 0x0F,
     UPDATE_ALL_CHAR = 0x1F,
+    UPDATE_DESPAWN  = 0x20,
 };
 
 enum ENTITYFLAGS
 {
     FLAG_NONE          = 0x000,
     FLAG_INFO_ICON     = 0x001, // (I) Icon next to name
+    FLAG_LARGE_MODEL   = 0x002,
```
TeoTwawki: no, wrong.

these flags values are
1. not consistent across entity types (that bit isn't size for players), and
2. sizing flags do not have 100% consistent  behaviors across mob models (in some cases, this exact value of 2 will make a model _SMALLER_ so this is a bad way to name the flag)

scripts/globals/trust.lua
zach2good: Agreed 👍

src/map/entities/baseentity.h
zach2good: How about `FLAG_ALTERNATE_APPEARANCE` ?

zach2good: We talked about this on discord, going to leave this out for now

claywar (approved):

PR #1705, opened 2022-05-11
Fixes SQL Async Causes High CPU Load

neuromancerxi:
Noticed very high CPU Load on an idle server instance. With the guidance
of zach2good, added a sleep condition to prevent the process from
running wild and eating CPU for breakfast, lunch, and dinner.

**_I affirm:_**

### DIFF
--- a/src/common/sql.cpp
+++ b/src/common/sql.cpp
@@ -46,6 +46,7 @@ void AsyncThreadBody(const char* user, const char* passwd, const char* host, uin
     while (asyncRunning)
     {
         con.HandleAsync();
+        std::this_thread::sleep_for(200ms); // TODO: This is bad and ugly. Replace with something better.
     }
 }
 


src/common/sql.cpp
```diff
@@ -46,6 +46,7 @@ void AsyncThreadBody(const char* user, const char* passwd, const char* host, uin
     while (asyncRunning)
     {
         con.HandleAsync();
+        std::this_thread::sleep_for(200ms); // TODO: This is bad and ugly. Replace with something better.
```
WinterSolstice8, line 49: @zach2good how about [std::this_thread::yield()](https://en.cppreference.com/w/cpp/thread/yield)? as I understand it, that essentially tells the scheduler that we're done and other stuff can use the CPU so it's not busy looping.

WinterSolstice8: Hmm. Reading more, yield() still causes a high CPU load to be displayed even if other process and threads can still use the CPU time that would normally be used by a busy loop without yield().

zach2good: Yeah, I'm not familiar with yield, but just looks like an os-bound tiny sleep that we can't control the length of. I'd say the answer here is to eventually use a condition variable to signal when there is some more work to do, but a sleep is fine for now

zach2good (approved):

PR #1707, opened 2022-05-12
[Mission] Stop Magicite mission from appearing in gate guard list

Xaver-DaRed:
**_I affirm:_**

Magicite mission (Any-Nation 4-1) is always given by the consulate of the proper nation and never from a nation gate guard.


claywar:
This shouldn't be happening, since 2 is to *not* add to the mask.
`0 = Not Repeatable, 1 = Repeatable, 2 = Do Not Add`

If there's something broken in that logic, then that is what should be fixed. 0 _is_ added to the list.

claywar (approved):

PR #1727, opened 2022-05-15
[Packets] Fix utsusemi handling in action packet

WinterSolstice8:
Utsusemi message fixes, WIP because need to check counter packet in retail.

**_I affirm:_**


WinterSolstice8:
Verified counter packet with retail.

Odd detail, it seems that some mobs can't counter or at least produce no visible message on the client. Lizards with counter at least on LSB don't produce a log message but Mandragoras do.

zach2good (approved):

PR #1737, opened 2022-05-15
[Interaction] Convert RDM AF1 and AF2 Quests

Xaver-DaRed:
**_I affirm:_**
- [?] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits


N3ckB3ard (approved):

scripts/quests/sandoria/RDM_AF1_The_Crimson_Trial.lua
```diff
@@ -0,0 +1,167 @@
...
+                onTrigger = function(player, npc)
+                    return quest:messageSpecial(davoiID.text.AN_ORCISH_STORAGE_HOLE)
+                end,
+
+                onTrade = function(player, npc, trade)
+                    if
+                        npcUtil.tradeHasExactly(trade, xi.items.DAVOI_STORAGE_KEY) and
+                        not player:hasKeyItem(xi.ki.ORCISH_DRIED_FOOD)
+                    then
+                        player:tradeComplete()
```
claywar, line 107: Prefer confirmTrade here, also, not sure if trades can fall through to other functions, so might want to return quest:keyItem for the below item.

claywar (approved):
Seems reasonable to me.  Personal preference is to just use the quest name instead of prefixing them for the filenames, but we use that in other places as well.

Issue #1694, opened 2022-05-10
[MAP] Crashing due to trusts or zoning? [Commit b51586a]


kaincenteno:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: base

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

logs:

```
/09/22 22:13:05:741][map][info][lua] fTP error: TP value is not between 1000-3000!  (print:437)
[05/09/22 22:13:10:194][map][info][info] parse: 01A | 31E2 308A 0E from user: Raist (parse:660)
[05/09/22 22:13:10:194][map][info][action] CLIENT Raist PERFORMING ACTION Spellcast (0x03) (SmallPacket0x01A:1158)
[05/09/22 22:13:10:540][map][error][navmesh] CPathFind::FindPath Entity (Eschan_Ilaerns_Euvhi - 17961058) could not find path (FindPath:346)
[05/09/22 22:13:10:944][map][error][navmesh] CPathFind::FindPath Entity (Eschan_Ilaerns_Euvhi - 17961058) could not find path (FindPath:346)
[05/09/22 22:13:11:340][map][error][navmesh] CPathFind::FindPath Entity (Eschan_Ilaerns_Euvhi - 17961058) could not find path (FindPath:346)
[05/09/22 22:13:11:740][map][error][navmesh] CPathFind::FindPath Entity (Eschan_Ilaerns_Euvhi - 17961058) could not find path (FindPath:346)
[05/09/22 22:13:12:140][map][error][navmesh] CPathFind::FindPath Entity (Eschan_Ilaerns_Euvhi - 17961058) could not find path (FindPath:346)
/bin/sh: 1: Syntax error: Bad fd number
[05/09/22 22:13:12:188][map][critical][fatalerror] --- gdb backtrace --- (dump_backtrace:141)
[05/09/22 22:13:12:188][map][critical][fatalerror]  (dump_backtrace:142)
terminate called without an active exception
```

backtrace:

```
oot@4fc18c239935:/server# gdb topaz_game core.38 
[...]

Issue #1724, opened 2022-05-14
[CRASH] on Everything branch commit 19168d (latest release)


AuroraFlare:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: EVERYTHING

There is no error message just after updating to the latest everything the server crashes on death and random things. Checked error logs and everything just know it was reported on death and random things even being AFK.

TeoTwawki:
just updated the branch, see if it still happens

RAIST5150:
We are a bit behind on xiweb... think we are right before #1715, or thereabouts.

My SMN was afk with a pet out and 4 trusts (5th tends to crash the client atm)  The pet and all four trusts were taken out before my face was eaten.

Death warped and both client and server are still up and running.

AuroraFlare:
Weird it worked okay for you? i updated, I played everything seemed great, died and then the server crashed no error message. I did reboot the server, etc. the whole IT stuff you would do. No issues.

```
[05/14/22 16:50:19:958][map][info][action] CLIENT Digitalgeek PERFORMING ACTION Spellcast (0x03) (SmallPacket0x01A:1158)
[05/14/22 16:50:19:958][map][info][info] parse: 050 | 0B1C 0B1A 04 from user: Digitalgeek (parse:681)
[05/14/22 16:50:19:971][map][info][info] parse: 050 | 0B1C 0B1A 04 from user: Digitalgeek (parse:681)
[05/14/22 16:50:19:982][map][info][info] parse: 050 | 0B1C 0B1A 04 from user: Digitalgeek (parse:681)
[05/14/22 16:50:19:992][map][info][info] parse: 050 | 0B1C 0B1A 04 from user: Digitalgeek (parse:681)
[05/14/22 16:50:20:769][map][info][info] parse: 050 | 0B1E 0B1C 04 from user: Digitalgeek (parse:681)
[05/14/22 16:50:20:781][map][info][info] parse: 050 | 0B1E 0B1C 04 from user: Digitalgeek (parse:681)
[...]

Issue #1666, opened 2022-05-04
Crash on call for help mobs when trusts in different parties are killed


CatsEyeXI:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: base

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

1. Tested in Dynamis - Jeuno on final boss.
2. 4 separate players with 3-5 trusts each
3. Boss was CFH
4. All participants engaged boss and trusts were defeated by Seismostomp (or other TP moves)
5. Server eventually crashed.

log attached
[topaz_game.exe_3-5_23-12-35.log](https://github.com/LandSandBoat/server/files/8617700/topaz_game.exe_3-5_23-12-35.log)

UPDATE: 
After further testing, it appears this also consistently happens with a single player with trusts, fighting claimed mob normally.
We can reproduce this consistently by popping Goblin Golem in Dyna Jeuno, and feeding him 3000 TP until he seismostomps all the trusts to death.

CatsEyeXI:
I've updated the description of the issue. Seems that this is a broader issue, causing a crash against even not CFH mobs, with a single player (plus trusts)... should easy to reproduce. Latest build.

jeffnavy14:
I too have found that in normal solo party (trusts) in the open world once all trusts die it instantly crashes the server, I have had to disable trusts on my server (Omicron) due to players being close do dieing from AoE's but trusts fall to the floor it Instant crashes. I keep getting Enmity Container fail's. I recently updated to the latest pull and still fails/crashes on trust death.

zach2good:
Stack traces?

CatsEyeXI:
> Stack traces?

There's one attached to the issue, would you like a new one?

zach2good:
I've already patched the area from the original crash:
https://github.com/LandSandBoat/server/commit/4243e4234daf31f39de2f3eb87e385823c0bdb52

So the crash has moved somewhere else, so a new trace is needed

CatsEyeXI:
[...]

Issue #1654, opened 2022-05-01
[CRASH]  Map server [Commit b500127]


kaincenteno:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: base

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

```
#0  __pthread_clockjoin_ex (threadid=140450195306240, thread_return=0x0, clockid=<optimized out>, abstime=<optimized out>, block=<optimized out>) at pthread_join_common.c:145
        __oldtype = 0
        __err = <optimized out>
        tid = 4763
        _buffer = {__routine = 0x7fbd23e9fa60 <cleanup>, __arg = 0x7fbd1bfffd28, __canceltype = 11, __prev = 0x0}
        pd = 0x7fbd1bfff700
        self = <optimized out>
        result = 0
        pd_result = <optimized out>
#1  0x00007fbd240fc047 in std::thread::join() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#2  0x0000560a22069420 in do_final (code=1) at /home/kain/server/src/map/map.cpp:314
No locals.
#3  0x0000560a21ff9a07 in do_abort () at /home/kain/server/src/map/map.cpp:336
No locals.
#4  sig_proc (sn=<optimized out>) at /home/kain/server/src/common/kernel.cpp:171
        is_called = 0
        __FUNCTION__ = <optimized out>
#5  <signal handler called>
No locals.
#6  std::_Rb_tree<CBattleEntity*, CBattleEntity*, std::_Identity<CBattleEntity*>, std::less<CBattleEntity*>, std::allocator<CBattleEntity*> >::find (__k=<optimized out>,
    this=<optimized out>) at /usr/include/c++/9/bits/stl_tree.h:2559
        __j = <optimized out>
        __j = <optimized out>
#7  std::set<CBattleEntity*, std::less<CBattleEntity*>, std::allocator<CBattleEntity*> >::find (__x=<optimized out>, this=<optimized out>) at /usr/include/c++/9/bits/stl_set.h:795
[...]

PR #1740, opened 2022-05-16
[core] Be more careful with Trust destruction

zach2good:
**_I affirm:_**

Tested before and after:
- `!godmode`, `!immortal`, `!togglegm`, `!changejob WHM 30`
- Summon 5x melee trusts
- `!fafnir`, and give him `!tp 3000`
- Engage and hope he does some nasty AOE immediately and wiped out your trusts.

Managed to get a repro a couple of times, and then none after.

Should fix:
https://github.com/LandSandBoat/server/issues/1694
https://github.com/LandSandBoat/server/issues/1724
https://github.com/LandSandBoat/server/issues/1666
https://github.com/LandSandBoat/server/issues/1654

### DIFF
--- a/src/common/WheatyExceptionReport.cpp
+++ b/src/common/WheatyExceptionReport.cpp
@@ -251,7 +251,6 @@ LONG WINAPI WheatyExceptionReport::WheatyUnhandledExceptionFilter(
             m_hDumpFile, MiniDumpWithIndirectlyReferencedMemory, &info, &additionalStreamInfo, nullptr);
 
         CloseHandle(m_hDumpFile);
-
     }
 
     if (m_hReportFile)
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):

claywar (approved):

zach2good:
I can't remember what project it was associated with, but we had a similar crash and fix with pets being destructed in instances, which is why I've copied the same pattern onto trusts. This will eventually all be moot once we get rid of the manual management of pointers to entities in favour of just using shared/unique/weak ptrs.

Xaver-DaRed:
IIrc it was a matter of entities (pets, pups, etc) being out of range or something similar

PR #1741, opened 2022-05-16
Simplify Shami onTrade seal event options

claywar:
**_I affirm:_**
- [🤞] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

### DIFF
--- a/scripts/zones/Port_Jeuno/npcs/Shami.lua
+++ b/scripts/zones/Port_Jeuno/npcs/Shami.lua
@@ -91,22 +91,15 @@ entity.onTrade = function(player, npc, trade)
     -- Trading Seals/Crests
     local sealOption = getSealTradeOption(trade)
     if sealOption ~= nil then
+        local eventParams = { 321, 0, 0, 0, 0, 0 }
         local storedSeals = player:getSeals(sealOption)
         local itemCount = trade:getItemCount()
 
-        if sealOption == 0 then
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

scripts/zones/Port_Jeuno/npcs/Shami.lua
```diff
@@ -91,22 +91,15 @@ entity.onTrade = function(player, npc, trade)
...
-        elseif sealOption == 3 then
-            player:startEvent(321, 0, 0, 0, (storedSeals + itemCount) * 65536)
-        else
-            player:startEvent(321, 0, 0, 0, 0, (storedSeals + itemCount) * 65536)
-        end
+        eventParams[sealOption + 2] = bit.lshift(storedSeals + itemCount, 16)
+        player:startEvent(unpack(eventParams))
         player:addSeals(itemCount, sealOption)
         player:confirmTrade()
+        return
```
TeoTwawki, line 102: maybe just else to the other condition instead of returning

its just not something we normally do and a fear people copying without understanding in the far future

claywar: That was my original intent, but I was trying to avoid calling getOrbEvent twice, since it loops over the shamiOrbItems table

PR #1746, opened 2022-05-17
[core] Add ConsoleService

zach2good:
**_I affirm:_**

Slowly chipping away at porting my changes from 2022 refresh

### DIFF
--- a/src/common/socket.cpp
+++ b/src/common/socket.cpp
@@ -168,7 +168,7 @@ int32 makeConnection(uint32 ip, uint16 port, int32 type)
 
     if (fd == -1)
     {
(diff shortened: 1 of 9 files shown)


src/common/socket.cpp
```diff
@@ -55,6 +55,8 @@ typedef int socklen_t;
 SOCKET sock_arr[FD_SETSIZE];
 int    sock_arr_len = 0;
 
+std::array<std::unique_ptr<socket_data>, FD_SETSIZE> session;
```
zach2good, line 58: I'm not sure how we've ever been compiling, with this missing from a compilation unit 👀

Xaver-DaRed (approved):

PR #1747, opened 2022-05-17
[connect server] Delete session in call to delete_session

zach2good:
**_I affirm:_**

With the use of extra debug logging, we discovered on stream that we're filling the `session` array but never clearing entries out of it. In this context, sessions are the active reservation and use of fd's with sockets. So, once we run out (1000+ on windows, now moved up to 4k), no more will be able to be allocated. 

With this change, the number shrinks back down as sockets are released (ish).

Won't know if this fixes the login server until it's out in the wild 🤷


scripts/commands/ver.lua
```diff
@@ -1,19 +0,0 @@
...
------------------------------------
-
-cmdprops =
-{
-    permission = 0,
-    parameters = ""
-}
-
-function onTrigger(player, target)
-    local version = GetServerVersion()
```
TeoTwawki, line 13: I'll take "things we forgot existed" for 200, Alex

Xaver-DaRed (approved):

PR #1751, opened 2022-05-19
[NPC] condor eye added missing dialog

kaincenteno:
condor eye doesnt turn around to talk to you

**_I affirm:_**

adds dialog for condor eye when not inside the port. Currently he is not acting as in retail and rotating 180 to talk to you and gives you only 1 dialog, he should have 1 dialog for inside the port and 1 outside.


Xaver-DaRed (approved):

kaincenteno:
Tested, working as expected

PR #1763, opened 2022-05-20
Dbtool.py Protect List Changes

The-Aerec:
**_I affirm:_**

Add zone_settings.sql to default protected list

Use Dbtool.py


The-Aerec:
@cocosolos - Separated

The-Aerec:
After discussion with @zach2good:

We came to a conclusion that the best course of action for now is to add zone_settings.sql to the protected list in dbtool.py

This adds the desired functionality and changes nothing for:

Local developers

And removes the extra step of protecting zone_settings.sql for server maintainters.

cocosolos:
I thought to go-to solution for this kind of stuff was making a custom.sql file to import after an update? My only complaint about this is losing the automatic update function for the occasional changes to zone_settings, which is rare enough to miss unless you're paying attention.

TeoTwawki:
Its a bad idea to ever modify the stock sql files. its far safer to make a new file as coco says and run it after the stock files are done.

An example, from my own server :
```sql
-- Update zone ports: move all BCNM type and instance type battlefields to Battlefield server
UPDATE zone_settings SET zoneport = 54237 WHERE zonetype = 4 or zonetype = 6;

-- Update zone ports: move all Dynamis and Limbus zones to Dynamis server
UPDATE zone_settings SET zoneport = 54238 WHERE zonetype = 5;
```

My IP set also lives in that file so i don't start my map server at 127.0.0.1 by mistake.

In a sub-folder named custom as its own file. Everything in the custom folder gets run after any changed from the stock files. As no char data exists in zone_settings it makes more sense to me to do this than to protect it.

I've just used a small `.cmd` (same thing as `.bat`) file to run the sql's from command line for the last 6 yrs

p.s. didn't we recently get module support for sql also?

TeoTwawki:
p.s. didn't we recently get module support for sql also?

PR #1793, opened 2022-05-24
Zulkheim Region NM audit

CriticalXI:
**_I affirm:_**

Bringing in my work from the Wings repo to correct issues with NMs in the Zulkheim region. 
https://gitlab.com/ffxiwings/wings/-/merge_requests/1198

Test any lua adjustments to mob behavior listed in the changed files.


sql/mob_droplist.sql
```diff
@@ -17428,6 +17428,7 @@ INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15123,10);
 INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15140,10);
 INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15142,10);
 INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,16360,10);
+INSERT INTO `mob_droplist` VALUES (3223,0,0,1000,16017,212);
```
Xaver-DaRed, line 17431: From this page data: http://ffxidb.com/zones/193/bombast
It would seem that this drop would fall either into very common or common per the chart I linked you.

The data-sample we have is not that great, unfortunately, so I would say change it to common (150)

scripts/zones/Konschtat_Highlands/mobs/Stray_Mary.lua
```diff
@@ -6,8 +6,14 @@ require("scripts/globals/hunts")
 require("scripts/globals/titles")
 require("scripts/quests/tutorial")
 -----------------------------------
+local ID = require("scripts/zones/Konschtat_Highlands/IDs")
```
Xaver-DaRed: I dont think this require is needed

Xaver-DaRed (changes requested):
Other than those 2 things, I think this is good to go

scripts/zones/Gusgen_Mines/Zone.lua
```diff
@@ -36,4 +36,41 @@ end
...
+
+    if totd == xi.time.NEW_DAY or totd == xi.time.MIDNIGHT then
+        local asphyxiated = GetMobByID(ID.mob.ASPHYXIATED_AMSEL)
+        local burned = GetMobByID(ID.mob.BURNED_BERGMANN)
+        local crushed = GetMobByID(ID.mob.CRUSHED_KRAUSE)
+        local pulverized = GetMobByID(ID.mob.PULVERIZED_PFEFFER)
+        local smothered = GetMobByID(ID.mob.SMOTHERED_SCHMIDT)
+        local wounded = GetMobByID(ID.mob.WOUNDED_WURFEL)
+
+        if not asphyxiated:isSpawned() and os.time() > asphyxiated:getLocalVar("cooldown") then
```
TeoTwawki: I have morning brain and didn't test my syntax here but something ***like*** this could save you some repeating those conditionals and compact this down to a loop:
```lua
        local ghost
        local ghostTable =
        {
            [1] = {id = ID.mob.ASPHYXIATED_AMSEL, nm = GetMobByID(ID.mob.ASPHYXIATED_AMSEL)},
            [2] = {id = ID.mob.BURNED_BERGMANN, nm = GetMobByID(ID.mob.BURNED_BERGMANN)},
            [3] = {id = ID.mob.CRUSHED_KRAUSE, nm = GetMobByID(ID.mob.CRUSHED_KRAUSE)},
            [4] = {id = ID.mob.PULVERIZED_PFEFFER, nm = GetMobByID(ID.mob.PULVERIZED_PFEFFER)},
            [5] = {id = D.mob.SMOTHERED_SCHMIDT, nm = GetMobByID(ID.mob.SMOTHERED_SCHMIDT)},
            [6] = {id = ID.mob.WOUNDED_WURFEL, nm = GetMobByID(ID.mob.WOUNDED_WURFEL)},
        }

        for i = 1, 6 do
            ghost = ghostTable[i].nm
            if not ghost:isSpawned() and os.time() > ghost:getLocalVar("cooldown") then
                SpawnMob(ghost.id)
            end
        end
```

sql/mob_droplist.sql
The-Aerec: Are we converting all droprates to the Eden / Limitbreak implementation of rarity bases? (I'm guessing yes since that is the retail accurate way)

TeoTwawki: > Are we converting all droprates to the Eden / Limitbreak implementation of rarity bases? (I'm guessing yes since that is the retail accurate way)

not exactly. We're making our rates fit into what seems to be the most logical rate "bucket" for a base with no TH, and at some later future date we'll work out wtf exactly we want to do about the old incorrect dsp TH implementation and what kind of table format we want to hold it all in. This has been the plan since before any servers made their own implementations, and I dunno what exactly they do since I haven't looked at their code at all.

The-Aerec: iirc Eden uses a non numeric entry in their drop tables: ie: Common, VC, Rare etc

And then based off those groupings affects the base droprates with the TH affects table that was shown in another MR recents.

Limitbreak changed all the droprates in a similar manner except instead of a string they used the: Common = 240, Rare = 100 etc method where everything that was determined to fit in those pools are assigned that pools respective numerical base droprate.

I think we talked about each of these in the ASB discord.

TeoTwawki: eventually I think I want presets instead of raw numbers, but with an option for a "custom" alternate drop rates that don't fit the retail values.

I believe this same system SE uses for drop rates matches what they do for other systems like harvesting. We already used those terms back in the day on old wiki without knowing the numbers behind them - I bet the numbers fit and SE reused the same table for multiple things.

The-Aerec: Yea I believe either the Eden/Lb way is fine.

I think Edens is much easier. Just reference the rarity level and then you can adjust what those rarity levels mean elsewhere. Making it easier to modulize and customize if people want to.

Xaver-DaRed: I stand in the oposite side. i would rather have all values be numeric. No need to just know the tier rates is one. the other for consistency, since the "paired" drops, like okote or dring dont have a % rate that fit a tier

As for the reason asking to change this to fit those tiers, its because the task of converting all drops to a fit proper tier is abysmal.
However, if adding or editing just a few, the effort to check them an do a good judjement on what tier they should have is much lower

The-Aerec: Can probably write a script to fit everything in the sql to their closest tier for now? And then update as we go for any outliers? or once we have more confirmations. (If we are converting the system that is)

TeoTwawki: > I stand in the opposite side. i would rather have all values be numeric.

@xaver I was thinking maybe sql variables to sort of macro it. so the numbers don't go away, just a little abstracted. Like when you cast something to a lua variable and then reuse that variable in a dozen places. SQL can actually do that. I haven't mocked up an example yet been workin irl to much to _think_ about things.

The-Aerec: I think this is pretty much Edens way of doing it. But once again we can easily writeup a script to go throught he sql and fit the current data to a 'Rarity' as a first pass. Then as users PR things in that have drops, just request for them to verify actual drop rates. (Although Limitbreak did a vast amount of the work for us there up to ToAU) So we could start with their #s and then fit the rest.

TeoTwawki: @The-Aerec I think we have a long way to go on manually correcting rates yet tho, even if a script replaced existing values to nearest official rate many would still be wrong. I'm sure we'll have many future PRs either way. We're kind of Cluttering crits PR now. If we need to we can make a discussion topic for further TH things if they aren't for the purpose of aiding his PR.

CriticalXI: Drop rate has been adjusted to the closest tier.

scripts/zones/Ordelles_Caves/mobs/Bombast.lua
```diff
@@ -8,33 +8,10 @@ local entity = {}
 
 entity.onMobInitialize = function(mob)
     mob:setMobMod(xi.mobMod.AUTO_SPIKES, 1)
-    mob:addStatusEffect(xi.effect.BLAZE_SPIKES, 45, 0, 0)
+    mob:addStatusEffect(xi.effect.BLAZE_SPIKES, 15, 0, 0)
```
TeoTwawki, line 12: Please leave a TODO note here for us to revisit this after scripted spikes are repaired, as this will not give retail accurate spikes behavior as is.

TeoTwawki (dismissed):

scripts/zones/Ordelles_Caves/mobs/Bombast.lua
CriticalXI: TODO comment added

Xaver-DaRed (approved):

scripts/zones/Gusgen_Mines/Zone.lua
```diff
@@ -40,35 +40,22 @@ zone_object.onGameHour = function(zone)
...
+            [1] = {id = ID.mob.ASPHYXIATED_AMSEL, nm = GetMobByID(ID.mob.ASPHYXIATED_AMSEL)},
+            [2] = {id = ID.mob.BURNED_BERGMANN, nm = GetMobByID(ID.mob.BURNED_BERGMANN)},
+            [3] = {id = ID.mob.CRUSHED_KRAUSE, nm = GetMobByID(ID.mob.CRUSHED_KRAUSE)},
+            [4] = {id = ID.mob.PULVERIZED_PFEFFER, nm = GetMobByID(ID.mob.PULVERIZED_PFEFFER)},
+            [5] = {id = ID.mob.SMOTHERED_SCHMIDT, nm = GetMobByID(ID.mob.SMOTHERED_SCHMIDT)},
+            [6] = {id = ID.mob.WOUNDED_WURFEL, nm = GetMobByID(ID.mob.WOUNDED_WURFEL)},
+        }
+
+        for i = 1, 6 do
+            ghost = ghostTable[i].nm
```
TeoTwawki, line 55: you mean my snippet worked as is 1st try? go me I guess!

CriticalXI: Yes, thank you! Only thing that needed to be adjusted is the info passed into the SpawnMob line. I'm learning a lot about utilizing tables this past week.

TeoTwawki: I was sure I wrote the for...do part wrong lol

TeoTwawki (approved):

PR #1795, opened 2022-05-24
[Mission] CoP7-5 Warriors Path Mechanics

The-Aerec:
**_I affirm:_**

Implements Scripted Tenzen Mechanics
Adds the Tarutaru Triplets and their mechanics

Basics from BG-WIki:

image

https://youtu.be/DuYTkqhT3RY?t=3487

!togglegm
!zone Sealions Den
Enter the mission BCNM "The_Warriors_Path"


The-Aerec:
Have some styling to do, but everything 'works'. Will open for review when its all cleaned. Passes local CI checks.

scripts/globals/mobskills/amatsu_hanaikusa.lua
```diff
@@ -9,7 +9,13 @@ require("scripts/globals/mobskills")
 local mobskill_object = {}
 
 mobskill_object.onMobSkillCheck = function(target, mob, skill)
-    return 0
+    if mob:getObjType() == xi.objType.TRUST then
```
claywar: This and the condition below can be simplified (if trust or animation == 0)

scripts/globals/mobskills/amatsu_kazakiri.lua
```diff
@@ -9,7 +9,13 @@ require("scripts/globals/mobskills")
 local mobskill_object = {}
 
 mobskill_object.onMobSkillCheck = function(target, mob, skill)
-    return 0
+    if mob:getObjType() == xi.objType.TRUST then
```
claywar: Same as above

scripts/globals/mobskills/amatsu_torimai.lua
```diff
@@ -9,7 +9,13 @@ require("scripts/globals/mobskills")
 local mobskill_object = {}
 
 mobskill_object.onMobSkillCheck = function(target, mob, skill)
-    return 0
+    if mob:getObjType() == xi.objType.TRUST then
```
claywar: Same as above

scripts/globals/mobskills/amatsu_tsukioboro.lua
```diff
@@ -9,7 +9,13 @@ require("scripts/globals/mobskills")
 local mobskill_object = {}
 
 mobskill_object.onMobSkillCheck = function(target, mob, skill)
-    return 0
+    if mob:getObjType() == xi.objType.TRUST then
```
claywar: Same as above

scripts/globals/mobskills/amatsu_yukiarashi.lua
```diff
@@ -9,7 +9,13 @@ require("scripts/globals/mobskills")
 local mobskill_object = {}
 
 mobskill_object.onMobSkillCheck = function(target, mob, skill)
-    return 0
+    if mob:getObjType() == xi.objType.TRUST then
```
claywar: Same as above

scripts/mixins/warriors_path_taru.lua
```diff
@@ -0,0 +1,68 @@
...
+    [2] = { 'guts', 'kuya', 'nagu', 'yoro', 'ffr1', 'awat' },
+    [3] = { 'awat', 'guts', 'kuya', 'nagu', 'yoro', 'ffr1' },
+}
+
+local randChance = { 0.166, 0.333, 0.500, 0.666, 0.833, 1}
+
+g_mixins.warriors_path_taru = function(mob)
+
+    mob:addListener("COMBAT_TICK", "TARU_CTICK", function(mobArg)
+
```
claywar, line 21: Newline can be removed

scripts/zones/Sealions_Den/mobs/Tenzen.lua
```diff
@@ -3,20 +3,102 @@
 --  Mob: Tenzen
 -----------------------------------
 local ID = require("scripts/zones/Sealions_Den/IDs")
+local TenzenFunctions = require("scripts/zones/Sealions_Den/helpers/TenzenFunctions")
```
claywar: Lower camel please for locals

scripts/mixins/warriors_path_taru.lua
```diff
@@ -0,0 +1,67 @@
...
+        local battletime = mob:getBattleTime()
+        local ID = zones[mob:getZoneID()]
+        local randOffset = math.random(0, 2) -- each taru most commonly uses text offset 0-2+ for during fight
+        local animationchance = math.random()
+        local battlefield = mob:getBattlefield()
+        if -- every 5 seconds the taru's will say something and use a random animation
+            battletime - changetime >= 5 and
+            battlefield:getLocalVar("fireworks") == 0
+        then
+            if cheru == 1 then
```
claywar, line 35: Is it intentional that only one taru can change per tick?

scripts/zones/Sealions_Den/helpers/TenzenFunctions.lua
```diff
@@ -0,0 +1,135 @@
+local ID = require("scripts/zones/Sealions_Den/IDs")
+
+local TenzenFunctions = {}
```
claywar: uppercase localvar needs renamed

scripts/mixins/warriors_path_taru.lua
The-Aerec: Nice catch, fixed will test in a bit

scripts/zones/Sealions_Den/IDs.lua
```diff
@@ -31,6 +35,26 @@ zones[xi.zone.SEALIONS_DEN] =
     {
         AIRSHIP_DOOR_OFFSET = 16908420,
     },
+    aWarriorsPath = {
```
claywar: Not a stopper, but I bet we could consolidate this down to a single offset in the future

The-Aerec: I did it this way because I referred to these mobs using their appropriate BFID (which as you have noted probably could have been an offset value instead) I can make this adjustment if you want.

claywar: Since it seems to be linear (and a multiplier of 3) it may be easier to maintain.  Again, not a show-stopper, but just trying to simplify things

TeoTwawki: all 12 are sequential so for example the others in the same battle could be tenzen+1, 2, 3.  and diff between 2 tenzen's is +4 per bf id.

The-Aerec: this has been fixed!

claywar (approved):

scripts/globals/mobskills/cosmic_elucidation.lua
```diff
@@ -0,0 +1,35 @@
...
+mobskill_object.onMobWeaponSkill = function(target, mob, skill)
+    local dmgmod = 2
+local info = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg()*21, xi.magic.ele.LIGHT, dmgmod, 3)
+    local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.LIGHT, 0)
+    target:takeDamage(dmg, mob, xi.attackType.MAGICAL, xi.damageType.LIGHT)
+
+    -- TODO: Cosmic Elucidation has a special message shown below
+    -- However it always displays 0 DMG even though its sopposed to show the amount of dmg a player takes
+    -- Need to figure out how to pass DMG through to this specific message.
+    --skill:setMsg(302) -- cosmic elucidation messageid
```
WinterSolstice8: do you have a cap for this message? I can look into it

The-Aerec: I don't unfortunately it was based on a few old videos of people losing to this. And the original author and I found it by cycling through messages. I couldn't find any 'capture' captures of people losing to this.

The-Aerec:
Conflicts fixed. Ready to go again

scripts/globals/mobskills/cosmic_elucidation.lua
```diff
@@ -0,0 +1,35 @@
...
+---------------------------------------------------
+local mobskill_object = {}
+
+mobskill_object.onMobSkillCheck = function(target, mob, skill)
+    return 1 -- only scripted use
+end
+
+mobskill_object.onMobWeaponSkill = function(target, mob, skill)
+    local dmgmod = 2
+local info = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg()*21, xi.magic.ele.LIGHT, dmgmod, 3)
```
WinterSolstice8: a few things:

1) indentation needs fixed
2) you are missing the final param of mobMagicalMove "tpvalue", which is nil here. You have set the tpeffect to "3", see function header: `xi.mobskills.mobMagicalMove = function(mob, target, skill, damage, element, dmgmod, tpeffect, tpvalue)
`
3) please update the modifier to the enum value and not just a number (see below)
4) assign the tpvalue final parameter to something sane (from the mobskill.lua):

```
-- MAGICAL MOVE
-- Call this on every magical move!
-- mob/target/skill should be passed from onMobWeaponSkill.
-- dmg is the base damage (V value), accmod is a multiplier for accuracy (1 default, more than 1 = higher macc for mob),
-- ditto for dmg mod but more damage >1 (equivalent of M value)
-- tpeffect is an enum from one of:
-- 0 = xi.mobskills.magicalTpBonus.NO_EFFECT
-- 1 = xi.mobskills.magicalTpBonus.MACC_BONUS
-- 2 = xi.mobskills.magicalTpBonus.MAB_BONUS
-- 3 = xi.mobskills.magicalTpBonus.DMG_BONUS
-- tpvalue affects the strength of having more TP along the following lines:
-- xi.mobskills.magicalTpBonus.NO_EFFECT -> tpvalue has no xi.effect.
-- xi.mobskills.magicalTpBonus.MACC_BONUS -> direct multiplier to macc (1 for default)
-- xi.mobskills.magicalTpBonus.MAB_BONUS -> direct multiplier to mab (1 for default)
-- xi.mobskills.magicalTpBonus.DMG_BONUS -> direct multiplier to damage (V+dINT) (1 for default)
--Examples:
-- xi.mobskills.magicalTpBonus.DMG_BONUS and TP=100, tpvalue = 1, assume V=150  --> damage is now 150*(TP*1)/100 = 150
-- xi.mobskills.magicalTpBonus.DMG_BONUS and TP=200, tpvalue = 1, assume V=150  --> damage is now 150*(TP*1)/100 = 300
-- xi.mobskills.magicalTpBonus.DMG_BONUS and TP=100, tpvalue = 2, assume V=150  --> damage is now 150*(TP*2)/100 = 300
-- xi.mobskills.magicalTpBonus.DMG_BONUS and TP=200, tpvalue = 2, assume V=150  --> damage is now 150*(TP*2)/100 = 600
```
4) you can now uncomment skill:setMsg(302) because the script was returning early and causing the return value to be nil, see pic related (I hardcoded it to 666)
image

5) I am sure nobody is sure if this is "light" elemental or not. I would probably assign it to be both `xi.attackType.SPECIAL` and `xi.damageType.ELEMENTAL`.  I don't think it matters because all this move is does is be pretty and the real damage is mostly irrelevant. Special and elemental should bypass most absorb checks while still using stoneskin/phalanx as intended.
6) There is no healed message for Cosmic Elucidation, so set `dmg = math.min(0, dmg)` after the `xi.mobskills.mobFinalAdjustments` call so the message doesn't return some huge bugged number.

WinterSolstice8 (changes requested):

scripts/globals/mobskills/cosmic_elucidation.lua
WinterSolstice8: also, make sure to test the whole fight again and check for lua errors in xi_map's output

I only found out about the missing parameter because I wanted to see why the message didn't work

WinterSolstice8 (approved):

PR #1811, opened 2022-05-28
[lua] untargetable -> setUntargetable

zach2good:
**_I affirm:_**

Renames methods and bindings from untargetable -> setUntargetable

CTRL+SHIFT+F to find the old calls and bindings, make sure there aren't any warnings on startup


src/map/lua/lua_baseentity.cpp
```diff
@@ -14505,7 +14505,7 @@ void CLuaBaseEntity::Register()
     SOL_REGISTER("setAggressive", CLuaBaseEntity::setAggressive);
     SOL_REGISTER("setTrueDetection", CLuaBaseEntity::setTrueDetection);
     SOL_REGISTER("setUnkillable", CLuaBaseEntity::setUnkillable);
-    SOL_REGISTER("untargetable", CLuaBaseEntity::untargetable);
+    SOL_REGISTER("setUntargetable", CLuaBaseEntity::setUntargetable);
```
claywar, line 14531: Though unused, should we go ahead and create a getUntargetable lua binding to match?

claywar (approved):

src/map/lua/lua_baseentity.cpp
zach2good: I did think about that, will do

Gradius2: Yes, I'm fully updated and of course, compiled thereafter.

zach2good:
https://github.com/LandSandBoat/server/runs/6717730179?check_suite_focus=true

None of those errors in our comprehensive Windows and Linux startup checks.

Gradius2:
I updated and compiled multiple times over the last month.

This is the first time I saw those errors, it has nothing to do with the link you provided.

I compiled just now, same errors.  I wonder what could be have gone wrong.

Gradius2:
I fully updated the /scripts folder by hand.  Same issue.  Could be due VS2022 ?  I'm running the last updated VS2022.

However is weird, never got such error before.

TeoTwawki:
"by hand"? proper updating is just using git pull

And what the link has to do with it is showing you a normal run doesn't have those errors. The 1st error message is telling you that function doesn't exist. you've done something wrong and it doesn't exist in your source but you have scripts calling it.

Xaver-DaRed:
This isnt a place to have a conversation
And this only goes to prove that, up to now, you were not building properly. I'm getting the same kernel error, witch means there is no way you were building up to this point

PR #1821, opened 2022-05-28
New commands for easier in game testing

CriticalXI:
**_I affirm:_**

Adding several commands to LSB to help with easier in game testing of mobs

- getenmity: Command to spit out current CE/VE values on targeted mob
- getmobmod/setmobmod: Commands to interact with mobmods in game
- reset: Now also wipes any debilitating effects from the player
- mobstun: Stuns the targeted mob for an hour for an in game testing dummy
- godmode: Adds toned down version of godmode to help with testing incoming/outgoing damage

Load this branch and attempt the above listed commands.


scripts/commands/softgodmode.lua
```diff
@@ -0,0 +1,43 @@
+---------------------------------------------------------------------------------------------------
+-- func: SoftGodMode
+-- desc: Toggles a toned-down version of godmode that allows the GM to test incoming/outgoing damage
```
TeoTwawki: We've got `!immortal` for this

scripts/commands/setmobmod.lua
```diff
@@ -0,0 +1,46 @@
+---------------------------------------------------------------------------------------------------
+-- func: semobtmod
```
TeoTwawki: typo: mob**t**mod

scripts/commands/reloadmixin.lua
```diff
@@ -0,0 +1,21 @@
...
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 4,
+    parameters = "ss"
+}
+
+function onTrigger(player, mixinLua, other)
+    if (mixinLua ~= nil and other == nil) then
```
TeoTwawki: outer parenthesis not needed

TeoTwawki: you can actually do `if thing then` and it will pass if `thing` exists.

scripts/commands/reloadzone.lua
```diff
@@ -0,0 +1,21 @@
...
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 4,
+    parameters = "ss"
+}
+
+function onTrigger(player, zoneLua, other)
+    if (zoneLua ~= nil and other == nil) then
```
TeoTwawki: same here

scripts/commands/getmobmod.lua
```diff
@@ -0,0 +1,51 @@
...
+    for k, v in pairs(xi.mobMod) do
+        modNameByNum[v]=k
+    end
+
+    -- validate modID
+    id = string.upper(id)
+    local modId = tonumber(id)
+    local modName = nil
+
+    if modId ~= nil then
```
TeoTwawki: `if modId then` will ensure it isn't nil

TeoTwawki: same for the checks after here.

scripts/commands/getenmity.lua
```diff
@@ -0,0 +1,30 @@
...
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!getenmity")
+end
+
+function onTrigger(player)
+
+    local targ = player:getCursorTarget()
+
+    if targ == nil or targ:isMob() == false then
```
TeoTwawki: ` if not targ or not targ:isMob() then`

scripts/commands/entityanimation.lua
```diff
@@ -0,0 +1,40 @@
...
+        targ = player:getCursorTarget()
+        animationId = arg1
+    else
+        -- player provided npcId and animationId.
+        targ = GetNPCByID(tonumber(arg1))
+        animationId = arg2
+    end
+
+    -- validate target
+    if (targ == nil) then
```
TeoTwawki: `if not targ then`

scripts/commands/entityanimation.lua
```diff
@@ -0,0 +1,40 @@
...
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!entityanimation {npcID} <animationID>")
+end
+
+function onTrigger(player, arg1, arg2)
+    local targ
+    local animationId
+
+    if (arg2 == nil) then
```
TeoTwawki: `if not arg2 then`

scripts/commands/entityanimation.lua
```diff
@@ -0,0 +1,40 @@
+---------------------------------------------------------------------------------------------------
+-- func: entityanimation
```
TeoTwawki: so turns out we have commands for this already, but they are type specific - one for players one for non players.

https://github.com/LandSandBoat/server/blob/base/scripts/commands/animatenpc.lua#L53

scripts/commands/letmego.lua
```diff
@@ -0,0 +1,43 @@
...
+    parameters = ""
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!letmego")
+end
+
+function onTrigger(player)
+    player:delStatusEffect(xi.effect.TERROR)
```
TeoTwawki: While this appears to be useful for some things, we don't believe this is necessary to test content. we're already able to del effects on demand and with immortal toggle we can survive anything thrown at us. We can even temporarily become immune to these effects if they are too annoying.

scripts/commands/mobstun.lua
```diff
@@ -0,0 +1,28 @@
...
+
+cmdprops =
+{
+    permission = 3,
+    parameters = ""
+}
+
+function onTrigger(player)
+    local targ = player:getCursorTarget()
+    if targ ~= nil and targ:isMob() then
```
TeoTwawki: ` if targ and targ:isMob() then`

scripts/commands/softgodmode.lua
CriticalXI: Immortal just makes you impossible to die. I found this command to be extremely useful for my testing on Wings if I wanted to test outgoing damage without all the huge damage boosts that godmode gives you.

scripts/commands/letmego.lua
CriticalXI: I assume you would need to know all the effect IDs to remove these on demand as you're mentioning? And become immune? I found this command to be a very useful all in one if I was testing a mob that had a lot of debuffs (the final fight in A Shantotto Ascension comes to mind, their TP moves apply a lot of debuffs).

scripts/commands/softgodmode.lua
TeoTwawki: What is it you need from this exactly? Immortal leaves your damage normal, and you can set mp and tp as needed if you need to test ws or spells. Or just `!addeffect regain 1000 0 0`  instead of setting tp. Just trying to understand the reason.

scripts/commands/letmego.lua
TeoTwawki: Nope, don't need to know the IDs since we map a human readable name to them all. just gotta spell them right and not type like ol Teo  ._. ;; There are some blocking effects like poison screen too.

scripts/commands/softgodmode.lua
CriticalXI: I guess I just found it to make testing easier without having to spam commands so much or add effects one by one. It was just a one command that would give me everything I needed to tune and balance fights/NMs.

TeoTwawki: Additional thought: we could add an optional parameter to the existing godmode to set "tiers" of it. It looks like this version you have is lost on zone intentionally so it wouldn't clutter up the player.lua so I don't think anyone would oppose that modification

scripts/commands/entityanimation.lua
CriticalXI: Removed command entirely

scripts/commands/letmego.lua
CriticalXI: I can remove this command from the PR if this isn't wanted in LSB. I just found it to be the same reason for the softgodmode command, it made testing easier. I used this command a lot.

TeoTwawki: discussing with team about getting some easy toggle for debuf immunity in for you instead of mass removal.

TeoTwawki: I think we're going to take this in for now and possibly change the method later, Zach is going to have a look at the  PR before anything is final.

scripts/commands/reloadzone.lua
```diff
@@ -0,0 +1,21 @@
...
+function onTrigger(player, zoneLua, other)
+    if zoneLua and other then
+        local String = table.concat({"scripts/zones/", zoneLua})
+        package.loaded[String] = nil
+        require(String)
+        player:PrintToPlayer(string.format("Lua file '%s' has been reloaded.", String))
+    else
+        player:PrintToPlayer("Must Specify a [zone]/file lua file.")
+    end
+end
```
zach2good: For this, another special handler could be added to `luautils::CacheLuaObjectFromFile` to make it hot-reload, instead of having to do it by hand. We already do this for the base global files:
```cpp

        // Handle Globals then return
        // Globals need to be nil'd before they're reloaded
        if (parts.size() == 2 && parts[0] == "globals")
        {
            std::string requireName = fmt::format("scripts/globals/{}", parts[1]);

            auto result = lua.safe_script(fmt::format(R"(package.loaded["{}"] = nil; require("{}");)", requireName, requireName));
            if (!result.valid())
            {
                sol::error err = result;
                ShowError("luautils::CacheLuaObjectFromFile: Load global error: %s: %s", filename, err.what());
                return;
            }

            ShowInfo("[FileWatcher] GLOBAL %s -> \"%s\"", filename, requireName);
            return;
        }
```

The logic would be something like: 
```cpp
if (parts.size() == 3 && parts[0] == "zones" && (parts[2] == "IDs" || parts[2] == "Zone"))
{
    // do the nil-ing and reload etc.
}
```

scripts/commands/reloadmixin.lua
```diff
@@ -0,0 +1,21 @@
+---------------------------------------------------------------------------------------------------
```
zach2good: mixins could be handled the same as I mentioned for zone globals

scripts/commands/mobstun.lua
```diff
@@ -0,0 +1,28 @@
+---------------------------------------------------------------------------------------------------
+-- func: mobstun
```
zach2good: Any reason this can't just be `!stun`? and then just check against "anything that isn't an NPC"

scripts/commands/letmego.lua
zach2good: I think what we came up with yesterday, was this would make more sense if it went into `!reset`. The big list of effects could also be a table, or, I think we have a binding that does erase or something? Worth looking into.

scripts/commands/getenmity.lua
```diff
@@ -0,0 +1,30 @@
...
+function onTrigger(player)
+
+    local targ = player:getCursorTarget()
+
+    if not targ or not targ:isMob() then
+        error(player, "you must select a target monster with the cursor first")
+        return
+    end
+
+    player:PrintToPlayer(string.format("your enmity against %s is ... CE = %u ... VE = %u", targ:getName(), targ:getCE(player), targ:getVE(player)))
```
zach2good: These prints could be capitalised

scripts/commands/reloadzone.lua
zach2good: need newline at end of file (we added a CI check for this last night, so you wouldn't have caught that yet)

scripts/commands/softgodmode.lua
```diff
@@ -0,0 +1,43 @@
...
+---------------------------------------------------------------------------------------------------
+
+cmdprops =
+{
+    permission = 2,
+    parameters = ""
+}
+
+function onTrigger(player)
+    if (player:getCharVar("SoftGodMode") == 0) then
```
zach2good: No need for parens here

scripts/commands/softgodmode.lua
zach2good: +1 for the tiers. I think it's best to remove this from this PR and we'll talk about reworking `godmode` to be more flexible later on

scripts/commands/getenmity.lua
```diff
@@ -0,0 +1,30 @@
...
+    parameters = ""
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!getenmity")
+end
+
+function onTrigger(player)
+
```
zach2good: Extra whitespace lurking around in this file, please remove

zach2good (changes requested):

scripts/commands/letmego.lua
TeoTwawki: I'm not sure but I think that erasable only covers part of these.  Regardless of what happens here I'd like to go ahead with eventually adding the option to be immune to these during testing. Can't be annoyed by them if they never land amiright?

scripts/commands/softgodmode.lua
CriticalXI: Please take a look, I've gone ahead and added some adjustments to godmode to take the argument of `1` to enable this version of soft god mode.

scripts/commands/letmego.lua
CriticalXI: I've removed the !letmego file and added this into !reset with the eraseAllStatusEffect() function. There is still a table to remove effects that aren't able to be erased.

scripts/commands/reloadzone.lua
CriticalXI: I'll be honest, I am extremely new to C++ so I would approach doing something like this in the future. Otherwise I'll leave in these reload commands unless you don't want them added to LSB.

scripts/commands/reloadmixin.lua
CriticalXI: As before, I'd tackle doing something like this later down the line. I'll remove these commands if you don't want them in LSB.

scripts/commands/mobstun.lua
CriticalXI: Renamed this command to `!stun` and adjusted the check to check if not NPC.

scripts/commands/getenmity.lua
CriticalXI: I think I've cleared up the whitespace you're referring to.

CriticalXI:
I've gone ahead and removed the reloadmixin and reloadzone from the MR. Please let me know if anything else is needed.

scripts/commands/reset.lua
```diff
@@ -30,7 +30,34 @@ function onTrigger(player, target)
...
+        xi.effect.CHARM_I,
+        xi.effect.CHARM_II,
+        xi.effect.POISON,
+    }
+
+    for _, v in pairs(effects) do
+        player:delStatusEffect(v)
+    end
+
+
```
zach2good: No need for this extra whitespace

scripts/commands/godmode.lua
```diff
@@ -1,16 +1,17 @@
...
 cmdprops =
 {
     permission = 1,
-    parameters = ""
+    parameters = "i"
 }
 
-function onTrigger(player)
-    if (player:getCharVar("GodMode") == 0) then
+function onTrigger(player, arg1)
```
zach2good: 'arg1' doesn't really tell us what this is doing. Try something like 'mode'

scripts/commands/godmode.lua
```diff
@@ -1,16 +1,17 @@
...
+    if (not arg1 or arg1 == 0) and player:getCharVar("GodMode") == 0 then
```
zach2good: You only need to handle arg1 once at the top of the function, rather than in both blocks:

```lua
local mode = utils.clamp(arg1 or 0, 0, 2)
local state = player:getCharVar("GodMode")

...
```

scripts/commands/godmode.lua
```diff
@@ -70,5 +72,40 @@ function onTrigger(player)
...
+        player:addHP( 50000 )
+        player:setMP( 50000 )
+
+        player:PrintToPlayer("Soft God Mode enabled.")
+    elseif (not arg1 or arg1 == 1) and player:getCharVar("GodMode") == 2 then
+        -- Toggle Soft GodMode off..
+        player:setCharVar("GodMode", 0)
+
+        -- Remove bonus effects..
+        player:delStatusEffect(xi.effect.MAX_HP_BOOST)
```
zach2good: There's a lot going on in this function now, break these out into their own functions:
```lua
local god_mode_on = function(player)
    ...
end

local god_mode_off = function(player)
    ...
end

local soft_god_mode_on = function(player)
    ...
end

local soft_god_mode_off = function(player)
    ...
end

Then the main function will be a lot clearer
```

scripts/commands/reset.lua
```diff
@@ -30,7 +30,34 @@ function onTrigger(player, target)
...
+        xi.effect.BLINDNESS,
+        xi.effect.AMNESIA,
+        xi.effect.CHARM_I,
+        xi.effect.CHARM_II,
+        xi.effect.POISON,
+    }
+
+    for _, v in pairs(effects) do
+        player:delStatusEffect(v)
+    end
```
zach2good, line 60: 👍

scripts/commands/setmobmod.lua
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    if target:isMob() == false then
+        error(player, "No valid target found. Place cursor on a mob. ")
+        return
+    end
+
+    local oldmod = target:getMobMod(modID)
+    target:setMobMod(modID, amount)
+    player:PrintToPlayer(string.format("Target name: %s (Target ID: %i) | Old %s modifier value: %i | New %s modifier value: %i", target:getName(), target:getID(), modifier, oldmod, modifier, amount))
```
zach2good, line 45: 👍

zach2good (changes requested):
Getting there 👍

zach2good:
Aside from these commands, whatever wacky commands you want to have for you own testing can now exist as modules:
https://github.com/LandSandBoat/server/tree/base/modules/custom/commands

It detects the presence of `cmdprops` and `onTrigger`; which normal commands have
https://github.com/LandSandBoat/server/blob/base/src/map/utils/moduleutils.cpp#L179

scripts/commands/godmode.lua
TeoTwawki: "tier" works

scripts/commands/godmode.lua
TeoTwawki: I dislike the terminology "soft god mode" for the lesser tier don't think helper functions should be named by tier anyway. I propose a section for shared bits both tiers do, then calling helpers named by what they do rather than which part they are for (effects/mods/immortality) as needed

opinions?

CriticalXI: I've adjusted it according to what Zach posted above. Also adjusted it so that you can toggle between the two modes without having to turn it off.

scripts/commands/godmode.lua
CriticalXI: Adjusted to "tier"

scripts/commands/reset.lua
CriticalXI: Removed

TeoTwawki (approved):

zach2good (approved):

PR #1831, opened 2022-05-29
use SQL vars in the droplist

TeoTwawki:
**_I affirm:_**

Uses SQL variables in our droplist table to make the tiers human readable.
```sql
-- Variables
SET @VCOMMON = 240;  -- Very common, 24%
SET @COMMON = 150;   -- Common, 15%
SET @UNCOMMON = 100; -- Uncommon, 10%
SET @RARE = 50;      -- Rare, 5%
SET @VRARE = 10;     -- Very rare, 1%
SET @SRARE = 5;      -- Super Rare, 0.5%
SET @URARE = 1;      -- Ultra rare, 0.1%
```
I wanted to avoid wrapping the variable names in back-tics in every table row, so I had to avoid the underscores I would normally use. This means no snake case. I could also do all one word all caps, or I could do camel case. If either of those are preferred or this shortening of the 1st word let me know, easy to mass replace to change it. Right now we don't really have a set standard on SQL variables in our style guide. I think we should add one.

Import the SQL file, then check your table data. You should see the numbers in place of the variable in your live table.


claywar (approved):

TeoTwawki:
the force push was to amend a comment line. :+1:

zach2good (approved):

Xaver-DaRed (approved):

PR #1839, opened 2022-05-30
[bcnm] added phaseChange to Apollyon SW

m241dan:
**_I affirm:_**

Apollyon - SW is set to "WON" shortly after the first player enters. This creates a weird situation where whoever follows gets a weird "you cannot enter, you need to wait" message. However, it still lets them enter. And then, when the original person casts on one of the people who entered after them, it will crash that client. When the caster fully DC / exits, the BCNM will clean up and everything will despawn, but the other party members will have the battlefield effect flag stuck on them (needs to be manually removed).

This PR adds a "phaseChange" to the Apollyon - SW BCNM since it has no "kill this specific mob" to win mechanic (ie, the "kill to win" changes based on day you enter). This leads to the malarkey described above. 

1. Bring two characters, (one rdm, one anything).
2. Enter on RDM.
3. Enter on other.
4. Notice weird message on "other".
5. RDM cast haste on "other", client will crash.
6. Profit from insanity.

### DIFF
--- a/scripts/zones/Apollyon/bcnms/sw_apollyon.lua
+++ b/scripts/zones/Apollyon/bcnms/sw_apollyon.lua
@@ -20,6 +20,11 @@ battlefield_object.onBattlefieldInitialise = function(battlefield)
     battlefield:setLocalVar("restorePH", ID.npc.APOLLYON_SW_CRATE[3] + random + 1)
     battlefield:setLocalVar("itemPH", ID.npc.APOLLYON_SW_CRATE[3] + random + 2)
 
+    -- Since this BC does not have a default "kill to win" mob set in the database like the other Apollyons
+    -- We have to manage it with this phaseChange variable. This allows the "kill to win" mobs to be the
+    -- elementals that correspond with the day you enter
+    battlefield:setLocalVar("phaseChange", 1)
+
     xi.limbus.setupArmouryCrates(battlefield:getID())
 end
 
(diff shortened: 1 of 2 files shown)


Xaver-DaRed:
Line 14 in scripts/zones/Apollyon/bcnms/sw_apollyon.lua
`battlefield:setLocalVar("lootSpawned", 1)`

I wonder if commenting out/deleting this line would fix this issue aswell. It would be a cleaner way if it does.

This variable isn't used in the other 3 apollyons, witch don't have this issue. It also isnt used in any of the 4 helpers.

m241dan:
> Line 14 in scripts/zones/Apollyon/bcnms/sw_apollyon.lua `battlefield:setLocalVar("lootSpawned", 1)`
> 
> I wonder if commenting out/deleting this line would fix this issue aswell. It would be a cleaner way if it does.
> 
> This variable isn't used in the other 3 apollyons, witch don't have this issue. It also isnt used in any of the 4 helpers.

No, not at all because this is the line that was borking things up: https://github.com/LandSandBoat/server/blob/e3e7d186965b55bee4d42c4a7f32c2c73368aafe/scripts/globals/battlefield.lua#L118

Unless somehow phaseChange is related to that?

Xaver-DaRed:
My point is more that the other 3 do not use phases and yet, still work properly.
The win condition for all 4 should be the same: "open the last chest in the last floor".
The winning chest in all 4 is spawned on battlefield creation, they are just set to be untargetable and invisible.
The mob that makes the chest visible is variable in several apollyons, not only in this one

Bottom line: This apollyon, somewhere, for some reason, is doing something different, and we should locate what it is.

m241dan:
I already did.

They all have mobs that have the "kill to win" condition. This one has none by default because they change. Because of that, it goes to load "what mobs do I need to kill to set to win?" And it gets back none, so it thinks they are all dead and sets the win.

Xaver-DaRed:
Aha, and that is set in the database.
It's a matter of ALL mobs in this apollyon being variable.

If thats the case, please add a comment where both are set, stating why, so we get it documented/we know why we are doing this.

m241dan:
> Aha, and that is set in the database. It's a matter of ALL mobs in this apollyon being variable.
> 
> If thats the case, please add a comment where both are set, stating why, so we get it documented/we know why we are doing this.

Will do.

m241dan:
@Xaver-DaRed comments added

Xaver-DaRed (approved):
We should not be assuming the victory condition is to kill anything. So, until we rework how we handle BCNMs victory conditions, this is the best option.

PR #1843, opened 2022-05-31
Lycopodium Mixin; Granite Door requires Uggalepih Key

CriticalXI:
**_I affirm:_**

- Granite Door at E-8 in Temple of Uggalepih now requires an Uggalepih Key
- Lycopodiums in the past will now follow players for a short time similarly to Hpemdes in Al'teaiu

- Door: Attempt to open the door from both sides, ensure it opens from insde the room and not from outside.
- Lycopodiums: With GM flag turned off, move around Lycopodiums and see that they follow you around for a short time.


kaincenteno:
thats awesome! I love lycopodiums moving in the past, this makes past zones more lively. Great work :D 

I have a quick question, do lycopodiums keep following all over the map or do they stop following you a couple meters away from where they spawn?

TeoTwawki:
> thats awesome! I love lycopodiums moving in the past, this makes past zones more lively. Great work :D
> 
> I have a quick question, do lycopodiums keep following all over the map or do they stop following you a couple meters away from where they spawn?

on retail at least, you could get them to follow long distances with some difficulty, if you didn't move to fast for them. But they'd also randomly forget me sometimes after a while even if I was not moving. ADHD mandies :)

scripts/zones/Temple_of_Uggalepih/npcs/_mf6.lua
```diff
@@ -0,0 +1,36 @@
...
+end
+
+entity.onTrigger = function(player, npc)
+    if player:getXPos() >= -208 then
+        player:messageSpecial(ID.text.THE_DOOR_IS_LOCKED, xi.items.UGGALEPIH_KEY)
+    else
+        npc:openDoor(11)
+    end
+
+    return 1
```
claywar: Is this a carryover from old code?  From what I can see, the only retval from onTrigger that causes action in core is -1

CriticalXI: Removed that, definitely a copy paste on my end from Wings that I should have removed.

scripts/zones/Temple_of_Uggalepih/npcs/_mf6.lua
```diff
@@ -0,0 +1,34 @@
...
+local ID = require("scripts/zones/Temple_of_Uggalepih/IDs")
+require("scripts/globals/npc_util")
+-----------------------------------
+local entity = {}
+
+entity.onTrade = function(player, npc, trade)
+    if npcUtil.tradeHasExactly(trade, xi.items.UGGALEPIH_KEY) then
+        player:confirmTrade()
+        player:messageSpecial(ID.text.YOUR_KEY_BREAKS, 0, xi.items.UGGALEPIH_KEY)
+        npc:openDoor(6.5)
```
TeoTwawki: I am reasonably sure this does not actually give you the .5 and that either 6 or 7 is close enough.

(7 happens to be the default because a majority of retail doors do that length - at least as far as we can tell with japan latency)

claywar: To echo that, 7 is also the default if no parameter is provided

scripts/zones/Batallia_Downs_[S]/mobs/Lycopodium.lua
```diff
@@ -0,0 +1,17 @@
...
+-- Area: Batallia Downs [S]
+--  Mob: Lycopodium
+-----------------------------------
+mixins = {require("scripts/mixins/families/lycopodium")}
+-----------------------------------
+
+local entity = {}
+
+entity.onMobInitialize = function(mob)
+    mob:setMobMod(xi.mobMod.ALWAYS_AGGRO, 1)
```
Comment (TeoTwawki) [inline scripts/zones/Batallia_Downs_[S]/mobs/Lycopodium.lua:11]: Why not either set them agro via their pool or bake this into their mixin script?

The extant example of this in SEA (Hpemde) use the pool value, if I recall correctly but it could easily be an onSpawn in the mixin as well.

Comment (CriticalXI) [reply scripts/zones/Batallia_Downs_[S]/mobs/Lycopodium.lua:11]: I am not familiar with something you can set in their pool to make them always aggro, please let me know if there is. They should always follow players regardless of the player's level so without this, at a certain point you won't be followed by the mandies if you out level their aggro range. Hpemdes don't have this issue because they're all lvl 70+.

The reason I placed this in the onMobInitiialize is that the ALWAYS_AGGRO mod was not properly being set on spawn through the mixin. I assume the mixin doesn't get applied to the mandies until after they have spawned during server start up so the intial round of lycopodiums spawned into the world would not properly have the mod set.

Comment (TeoTwawki) [reply scripts/zones/Batallia_Downs_[S]/mobs/Lycopodium.lua:11]: I'll look into why it's not working at the mixin level, it should. I can always move this there later, so marking this convo resolved and letting it through per mob. A comment about it in file would be good though :+1: 

Was mainly looking for why because I figured there had to be a problem for you to do it this way.

Things not applying on the initial spawn is a problem in itself - that 1st round of spawns should be firing off all the "on spawn" things after their onInit is done.

Xaver-DaRed (changes requested):
Lets keep the door duration at 7, unless it actually lasts less than 7. in that case, lets go with 6, or a round number in general.

Everything else looks good^^

CriticalXI:
Adjusted the door to be open for the default 7 seconds.

claywar (approved):

Xaver-DaRed (approved):

PR #1844, opened 2022-05-31
[lua] Make "Full Speed Ahead!" resistant to ID shifts

zach2good:
**_I affirm:_**

Removes the hard-coded IDs needed to complete FSA and replaces them with a by-name lookup of the main NPC, and then uses offsets from there to work out everything else.

Play, _and complete_ FSA, either naturally, or through `!minigame` menu

### DIFF
--- a/scripts/quests/full_speed_ahead.lua
+++ b/scripts/quests/full_speed_ahead.lua
@@ -136,3 +136,5 @@ xi.full_speed_ahead.completeGame = function(player)
 end
 
 xi.fsa = xi.full_speed_ahead
+
+return xi.full_speed_ahead -- NOTE: This return does nothing apart from silence the hot-reloader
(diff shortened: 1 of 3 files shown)


TeoTwawki:
nice :+1:

TeoTwawki (approved):
I know the CI isn't done yet, but I don't see anything that could fail it here.

zach2good:
I like this pattern of "look up the ID by name on zone init, or when you need it", but I don't like the binding. We should all talk about how we'd like it to look in the future (we're only using it in a couple of places so far, so we can be flexible)

TeoTwawki:
> I like this pattern of "look up the ID by name on zone init, or when you need it", but I don't like the binding. We should all talk about how we'd like it to look in the future (we're only using it in a couple of places so far, so we can be flexible)

I like it for things where we keep it brief don't have to deal with any side effects of it like here. I had some concerns that per my discussion with Claywar don't apply to our repo but did in old dsp regarding long iterations to find a name.

PR #1847, opened 2022-05-31
Drop list: Variablize 100% drop items as well

TeoTwawki:
**_I affirm:_**
- [ ] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

adds one more sql variable in mob_droplist.sql to cover dropslots that are 100%

import sql, observe rows that have 1000 imported successfully


claywar (approved):

zach2good (approved):

zach2good:
Do we do the thing where higher than this amount acts like dropping multiple?

Xaver-DaRed (approved):

Xaver-DaRed:
> Do we do the thing where higher than this amount acts like dropping multiple?

They should be different drops. as in same item listed multiple times

TeoTwawki:
> Do we do the thing where higher than this amount acts like dropping multiple?

nope, anything above 1000 just acts like 1000 coz it beats the random roll. multiple drops should use multiple db rolls instead.

PR #1854, opened 2022-06-01
Disable Quick Edit Mode by default on Windows Consoles, Add runFlag checking to ZMQ and Console threads

claywar:
**_I affirm:_**

* Disables Quick Edit Mode (Mark) in Windows Consoles to prevent accidental clicks in the window from freezing the server

* Uses runFlag to signal exit to ZMQ and Console threads to properly exit without SIGABRT
Windows consoles (cmd, VS debug window) should not register mouse highlight or clicks unless the user right-clicks to "Mark"


zach2good:
Could `int runFlag` be changed to `std::atomic<bool> runFlag` or something, not necessarily for safety, but to signal what it is and how it's used

claywar:
Done, should be good to go

src/common/console_service.h
```diff
@@ -19,6 +19,9 @@
 ===========================================================================
 */
 
+#ifndef _CONSOLE_SERVICE_H_
+#define _CONSOLE_SERVICE_H_
```
zach2good, line 23: No need for header guards like this, #pragma once does the same thing, it's just not part of the standard (or something)

src/common/kernel.cpp
```diff
@@ -46,7 +46,8 @@
 #endif
 #endif
 
-int    runflag = 1;
+std::atomic<bool> runFlag = true;
```
zach2good: Maybe rename this gRunFlag? Since it's mega-global

zach2good (approved):

src/common/console_service.h
claywar: Was failing on GCC10 without these

src/common/kernel.cpp
claywar: Will do 👍

PR #1858, opened 2022-06-02
Correct Geomancy and Handbell skill bonus mod

### LINKED ISSUE #1855: Temple Cyclas gives Geomancy and Handbell skill.

## Branch affected by issue 

`base`

## Expected behavior
The Temple Cyclas is the MNK arifact body piece. It currently gives Geomancy and Handbell skill.

ari-osm:
**_I affirm:_**

Moves the mod skill bonus check for Geomancy and Handbell skill down to the correct area and skips the normal check used for every other skill.  This makes it so "Enhances Chakra" gear doesn't erroneously grant Geomancy and Handbell skill.

Closes #1855 

Equip Temple Cyclas before and after patch.  Skill is now gone.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -2847,14 +2847,6 @@ namespace charutils
                     skillBonus += PChar->getMod(Mod::DARK_ARTS_SKILL);
                 }
             }
-            else if (i == 44)
-            {
-                skillBonus += PChar->getMod(Mod::GEOMANCY);
-            }
-            else if (i == 45)
-            {
-                skillBonus += PChar->getMod(Mod::HANDBELL);
-            }
             else if (i >= 22 && i <= 24)
             {
                 if (PChar->PAutomaton)
@@ -2866,7 +2858,18 @@ namespace charutils
             skillBonus += PChar->PMeritPoints->GetMeritValue(skillMerit[meritIndex], PChar);
             meritIndex++;
 
-            skillBonus += PChar->getMod(static_cast<Mod>(i + 79));
+            if (i == 44)
+            {
+                skillBonus += PChar->getMod(Mod::GEOMANCY); //1026
+            }
+            else if (i == 45)
+            {
+                skillBonus += PChar->getMod(Mod::HANDBELL); //1027
+            }
+            else
+            {
+                skillBonus += PChar->getMod(static_cast<Mod>(i + 79));
+            }
 
             PChar->WorkingSkills.rank[i] = battleutils::GetSkillRank((SKILLTYPE)i, PChar->GetMJob());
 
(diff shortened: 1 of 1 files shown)


src/map/utils/charutils.cpp
```diff
@@ -2866,7 +2858,18 @@ namespace charutils
             skillBonus += PChar->PMeritPoints->GetMeritValue(skillMerit[meritIndex], PChar);
             meritIndex++;
 
-            skillBonus += PChar->getMod(static_cast<Mod>(i + 79));
+            if (i == 44)
```
claywar: Call me crazy, but I believe the proper fix for Temple Cyclas would be to just edit the item mod entries in the db for that item?

TeoTwawki: the item mod entries are actually correct tho

claywar: Has someone confirmed this issue with retail client, or is this something happening with an outdated one?

TeoTwawki: The issue appears to be that it was assumed all the skills would be in the same order, but those chakra mods showed up in between and the geo ones were given an ID far away

And this is why magic numbers are bad kids.

edit: no, this STILL doesn't make sense.

ari-osm: yeah that's correct.  It's converting the xi.skill > xi.mod value by adding 79 which works for all except for geomancy and handbell.  44+79 = 123 which is enhance chakra effect.

TeoTwawki: I kind of want to swap the mod IDs just to put the skills in order, but that's my OCD talking

TeoTwawki: And I seriously dislike that nothing was making it obvious wtf `i` actually was and had to be inferred

TeoTwawki: lucy our predecessors have some splaining' to do!

TeoTwawki: opinion: 

1. for now, I think we should swap the IDs back that my ocd isn't far off from sane for a change
2. raw magic numbers should get replaced by their skill IDs name from the enumeration
3. a comment above `for (int32 i = 1; i < 48; ++i)` should be added to inform editors we're iterating skill IDs and offsetting by 79 to get the modifier ID
4. in the future (and for now a TODO comment) this whole thing should eventually get a refactored to be less dependent on arbitrary ordering of modifier IDs and conditionals on skill ranges.

Xaver-DaRed: I'll take care of swapping the modifiers in all necesary places.

If when I do that ari-osm hasnt gotten to it already, i'll refactor this... mess

ari-osm: 2, 3, and 4 are done.  I removed the geomancy and handbell logic so that when Xaver swaps the mod ids, it will all work.

Xaver-DaRed (approved):

TeoTwawki (approved):

PR #726, opened 2021-09-16
Refactor and expand gear_sets

neuromancerxi:
This update brings gear_sets up to date with all gear sets currently in the game as of September 2021, including a number of lower tier sets that were missing as well as all AF3 90/109/119 sets.
Arbitrarily reorders/renumbers the set ids in Gearsets starting from lowest level armor set and moving upwards.
This serves little purpose other than to help me identify which sets were missing as I added them in.
Updated and lined up comments where possible.
Moved Jailer weapons to the end of the stack.
Updated Gearset logic to apply buffs to pets (Breeders set and SMN AF1 119 +2/+3 currently).
addGearSetMod() now expects a 0 for standard sets and a 1 for petmod sets (see examples in `doSpecialHandling` and `addGearSetMod`)
Updated Gearset logic to rename `cappedTierSets` to the more aptly named `specialHandlingSets`.
Updated Gearset logic to standardize AF3 set bonuses where possible, and manually defined the outliers.
Added in missing `specialHandlingSets` and ensured that migrated sets matched their new id after refactor.
Wired up `AUGMENT_BLU_MAGIC` in bluemagic.lua.
Created new MODs for missing Empyrean Gear Sets:

Implemented with logic:
AUGMENT_DAMAGE_HP
AUGMENT_DAMAGE_PET_HP
AUGMENT_SONG_STAT*
AUGMENT_BLU_MAGIC

Place Holders: --TODO
AUGMENT_CONSERVE_MP
AUGMENT_BLOOD_BOON
AUGMENT_COMPOSURE
GEOMANCY_MP_NO_DEPLETE

*Regarding AUGMENT_SONG_STAT, due to the way that songs are applied, this is implemented in three different ways (depending on the existing song script):

1. The Lions share have the AUGMENT_SONG_STAT value passed as subPower in the song, and applied in the effect.
2. For etude, subpower is in use. The value is passed instead via power (elements of songs match the stat buff) and the etude effect, remaining unchanged, applies the bonus within the power.
3. For carol, subpower is also in use. The `AUGMENT_SONG_STAT` value is multiplied by 100, added to the element value, then deconstructed in the effect. This ensures that the buff and the base value can be cleanly separated and applied. Details and logic are commented in the effect.

**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
[...]

PR #1866, opened 2022-06-03
Rework Gearsets - Implement BLU AF3 Sets and Calcs

neuromancerxi:
Continued rework of #726 into smaller chunks for easier review/ingestion.

**_I affirm:_**

Updates the `bluemagic` script to apply bonus WSC for chain affinity and the AF3 set (when it 
procs) 
Adds the associated BLU AF3 sets to `gear_sets`.

References: 
<https://www.bg-wiki.com/ffxi/Hashishin_Attire_Set>
<https://www.bg-wiki.com/ffxi/Chain_Affinity>

**Testing Gear Set MOD Values**

Login a test character
`!changejob BLU 99`
Remove all currently equipped gear
`!getmod AUGMENT_BLU_MAGIC` should return 0

Add the following test items: `!additem <itemid>`
```
11079 \
11099 /  Set 218 samples
27082 \
27268  > Set 219 samples
27442 /
```
Mix and Match equipped items, check results with `!getmod AUGMENT_BLU_MAGIC`.
Sets 218 and 219 should activate independently of one another, but behave with the same logic, two items from the set minimum, with a bonus of 1 for each additional piece (2, 3, 4, 5).

Mix and match gear, and test for expected resulting value of `!getmod AUGMENT_BLU_MAGIC`. Combing at least 2 items of either set will activate the bonus:

1. AF3 85 Head + AF3 109 Hands will not activate bonus 
2. AF3 85 Head + AF3 89 Body will activate bonus (2) (Set 218 is active)
3. AF3 109 hands + AF3 119 Legs will activate bonus (2) (Set 219 is active)
4. AF3 85 Head + AF3 89 Body + AF3 109 hands + AF3 119 Legs will activate bonus (4) (Set 218 AND Set 219 are active)
5. Equipping all 5 example test items will replicate having set 218 active (2) and set 219 active (3) for a total of 5.
6. Bonus testing: Add all 15 item IDs referenced in the PR and mix and match to confirm expected values.

**Testing WSC Calculations**

Login a test character
`!changejob BLU 99`
`!addallspells`
`!capallskills`
Set `Foot Kick` as an available blu spell
Head to any starter zone (e.g. West Ron.)
Remove all currently equipped gear
Select a specific Test Mob for consistency (e.g. Tunnel Worm or Wild Rabbit)
Use `Foot Kick` against the mob and note the damage. (1x WSC)
Use `Chain Affinity` then use `Foot Kick` again, noting the increased damage output. (2x WSC)
Target the character and `!setmod AUGMENT_BLU_MAGIC 100`
Use `Foot Kick` against the mob and note the increased damage output. (3x WSC)
Use `Chain Affinity` then use `Foot Kick` again, noting the increased damage output. (4x WSC)

~~Note: Testers may un-comment lines 272, 273 and 275 in the updated `bluemagic.lua` to see pre- and post- values in the map server logs to help evaluate the changes.~~ Removed commented debug prints.

### DIFF
--- a/scripts/globals/gear_sets.lua
+++ b/scripts/globals/gear_sets.lua
@@ -33,6 +33,9 @@ local absorbDamageBaseRate = 2
 local absorbDamagePieceBonus = 1
 local instantCastBaseRate = 2
 local instantCastPieceBonus = 1
+local baseRate = 2
+local pieceBonus = 1
+--TODO: Migrate remaining AF3 variables to baseRate/pieceBonus and update comment.
 
 --              {id, {item, ids, in, no, particular, order}, minimum matches required, match type, mods{id, value, modvalue for each additional match, additional whole set bonus}
 local gearSets =
@@ -177,7 +180,9 @@ local gearSets =
              {id = 214, items = {26673, 26849, 27025, 27201, 27377}, matches = 2, matchType = matchtype.any, mods = {{xi.mod.DOUBLE_ATTACK, 4, 2, 0}} }, -- Argosy +1
              {id = 215, items = {25616, 25689, 27120, 27305, 27476}, matches = 2, matchType = matchtype.any, mods = {{xi.mod.MATT, 20, 10, 0}} }, -- Amalric +1
              {id = 216, items = {18947, 15818}, matches = 2, matchType = matchtype.any, mods = {{xi.mod.ACC, 5, 0, 0}, {xi.mod.SOULEATER_EFFECT, 2, 0, 0}} }, -- Moliones's Sickle/Ring
-             -- next id = 218
+             {id = 218, items = {11079, 11099, 11119, 11139, 11159}, matches = 2, matchType = matchtype.any, mods = {{xi.mod.AUGMENT_BLU_MAGIC, baseRate, pieceBonus, 0}} }, -- Mavi +2 Set: Occasionally triples the WSC of Blue Magic Spells. Will stack with Chain Affinity.
+             {id = 219, items = {26770, 26771, 26928, 26929, 27082, 27083, 27267, 27268, 27441, 27442}, matches = 2, matchType = matchtype.any, mods = {{xi.mod.AUGMENT_BLU_MAGIC, baseRate, pieceBonus, 0}} },  -- AF3 BLU 109/119 Set: Occasionally triples the WSC of Blue Magic Spells. Will stack with Chain Affinity.
+             -- next id = 220
         }
 
              -- increment id by the number of mods in previous gearset (e.g. id 199 has 3 mods, 199 + 3 = 202)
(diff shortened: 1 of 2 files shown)


scripts/globals/gear_sets.lua
```diff
@@ -178,7 +181,9 @@ local gearSets =
              {id = 214, items = {26673, 26849, 27025, 27201, 27377}, matches = 2, matchType = matchtype.any, mods = {{xi.mod.DOUBLE_ATTACK, 4, 2, 0}} }, -- Argosy +1
```
claywar, line 180: Can we find a way to incorporate the items.lua enum into this table?  It might mean expanding out the single-line format, but I think it would improve readability by a huge amount.

neuromancerxi: If it becomes a requirement, I'll consider it as an option as I continue iterating through the original PR to get to a finished state in future PRs. Though, tbh, I'm not sure it's a good fit here. Opinion follows:

I would point out that moving to ENUMs instead of Item IDs would likely add an additional layer of complexity to the process since we'd have to not only enumerate all the item IDs, but then migrate them into `items.lua` and additionally spot check for typos and item name inconsistencies across both files (e.g. would `27442` be `HASHISHIN_BASMAK_+1` or `HASHI._BASMAK_+1`?).

In a finished state, this might make this a bit more human readable, but any gains seem like they'd be outweighed by the time spent doing the conversion and the added complexity. My goal is to have the comments on each set line give enough additional contextual information to be used in conjunction with a cursory SQL look up to confirm the items:

```sql
SELECT * FROM xidb.item_basic
WHERE itemid in (26770, 26771, 26928, 26929, 27082, 27083, 27267, 27268, 27441, 27442);
```
`-- AF3 BLU 109/119 Set: Occasionally triples the WSC of Blue Magic Spells. Will stack with Chain Affinity.`

In a "finished" state, we'd see each line commented, with an easily selected array of item ids that can be plugged into SQL if there was any need to get more info.

Thoughts?

TeoTwawki: my 2 cents: I think while dealing with item names won't be that different from dealing with item ids (they get typo'd too) I think there's already enough to chew on in the incoming work here, and putting it to names can be done as its own thing at a later date if it really needs doing (I may even volunteer myself for it, I am good with that sort of drudgery when my clients at work take 20 min to figure out how to unmute their mic).

TeoTwawki: p.s. we've been using the convention `_P1` and `_p2` to mean `+1` and `+2` and stripping our symbols like periods, in the existing enum file, just so you know about that.

neuromancerxi: Any additional feedback on the PR? Does the dialog regarding itemid vs utilizing items.lua prevent this from moving forward in it's current state?

TeoTwawki: @claywar I've got no further input for the man, You?

neuromancerxi:
@claywar Following up on this PR. Are there remaining issues that should be addressed?

LSB Reviewers: It has been suggested that we break out PRs into smaller bites to make them easier to review. I am all for that, time is a scarce commodity and I want to be mindful of that. This request attempts to break up #726 into smaller, easier to review chunks.

As to the proposed changes, I'd like to get an understanding if this can move forward or if it's even desirable. If there's no interest in the PR, please let me know in comments, and I'll close it out instead of leaving it open for review.

Thanks in advance!

scripts/globals/gear_sets.lua
claywar: Only other thing I would suggest is removing the debug prints if they're no longer needed

Xaver-DaRed (approved):

claywar (approved):

PR #1875, opened 2022-06-04
Puk mob adjustments

CriticalXI:
**_I affirm:_**

- Adds correct Puk behavior where they react with a TP move if you hurt them with an element equal to the current day. In addition, all Puks will absorb Wind magic.
- Adjusted Puk TP list so that only Puks in Mamook can use Ill Wind.

Fight a Puk and see that they react with a TP move depending on the element used to damage them.


scripts/mixins/families/puk.lua
```diff
@@ -0,0 +1,45 @@
...
+        -- If the element corresponding to the elemental day of the in-game Vana'diel week is used on a Puk, it will get 100% TP instantly.
+        for k, v in pairs(elements) do
+            if damageType == v[1] and VanadielDayOfTheWeek() == v[2] then
+                puk:addTP(3000)
+            end
+        end
+    end)
+
+    mob:addListener("ROAM_TICK", "PUK_ROAM_TICK", function(puk)
+        if VanadielDayOfTheWeek() == xi.day.WINDSDAY and puk:getMod(xi.mod.REGAIN) == 0 then
```
TeoTwawki: They do this in wind weather. Anecdotal but feels like it doubles up in double weather.

Can't say they don't also get it from day, but I've not noticed it and wiki's don't mention it. They do mention wind weather though.

TeoTwawki: Found another paragraph of info farther down old wiki's family page:
> Auto-Regain: On Lightsday and Windsday, and during Light or Wind Weather, Puks have the ability to regenerate TP, much like the Eruca of Mount Zhayolm.

so both day and weather, and allegedly light as well as wind but that has a verification tag so we can omit that pending an in game test on retail.

Xaver-DaRed: On lightsday, Puks do not get Regain

Xaver-DaRed (dismissed):

scripts/mixins/families/puk.lua
TeoTwawki: looks like we got day of week wind twice, one just needs to be weather and the pr is good

TeoTwawki: oh I missed that one was for removing

Need an `or` in them then

CriticalXI: I adjusted the mixin to give regain on Wind day or weather. I also added another listener to do this if they're in combat as well.

TeoTwawki (approved):

PR #1881, opened 2022-06-05
Retail Accuracy for Brigandish Blade

m241dan:
**_I affirm:_**

Many have tried, many have failed but... this should make Brigandish Blade retail accurate.

Behaviour matches: https://youtu.be/LaXD0oh307A?t=371

1. Load it up
2. Fight Brigandish
3. Use Buccaneer's Knife (notice the add effect works)
4. Swap to some other weapon
5. Notice how he takes zero damage from everything once he's at 1 HP
6. Swap to Buccaneer's Knife, hit does "0" damage still, but the water add effect kills him.

### DIFF
--- a/scripts/globals/additional_effects.lua
+++ b/scripts/globals/additional_effects.lua
@@ -120,10 +120,11 @@ xi.additionalEffect.attack = function(attacker, defender, baseAttackDamage, item
         ABSORB_STATUS = 10,
         SELF_BUFF = 11,
         DEATH = 12,
+        DAMAGE_SPECIFIC_MOB = 13,
     }
 
     -- If we're not going to proc, lets not execute all those checks!
-    if chance < math.random(100) then
+    if chance < math.random(100) or (addType == procType.DAMAGE_SPECIFIC_MOB and defender:getTargID() ~= power) then -- power is the targID for type DAMAGE_SPECIFIC_MOB
         return 0,0,0
     end
 
@@ -137,7 +138,7 @@ xi.additionalEffect.attack = function(attacker, defender, baseAttackDamage, item
     end
     --------------------------------------
 
-    if addType == procType.DAMAGE then
+    if addType == procType.DAMAGE or addType == procType.DAMAGE_SPECIFIC_MOB then
         damage = xi.additionalEffect.calcDamage(attacker, element, defender, damage)
         msgID = xi.msg.basic.ADD_EFFECT_DMG
         if damage < 0 then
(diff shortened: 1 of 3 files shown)


scripts/globals/additional_effects.lua
```diff
@@ -120,10 +120,11 @@ xi.additionalEffect.attack = function(attacker, defender, baseAttackDamage, item
         ABSORB_STATUS = 10,
         SELF_BUFF = 11,
         DEATH = 12,
+        DAMAGE_SPECIFIC_MOB = 13,
```
TeoTwawki, line 123: Sorry but we already said we didn't want to do this exact thing on a previous pr, tho at the time our understanding of how the knife worked was off. I recently did a kill of brig. blade to be sure I understood, and some additional tests on the knife itself.

These are my notes from that: 

> tested hitting it with the dagger then unequipping. the dagger does not "remove" its immunity. It becomes the only way to damage it once it gets to 1% hp Melee hit did zero, add effect killed it. 
>
> add effect damage ~~does proc on other evil weapon type mobs NEARBY but not elsehwere in other zones, and for far less dmg. 43 vs a mystic weapon and it~~ did not proc vs a goobue gardener at all nor an evil weapon in rollanberry _I had a trust out who sneaky enspelled, and [log lag happened](https://user-images.githubusercontent.com/6871475/172252469-530c6224-39eb-4b21-b935-75ba67abf55f.png)_
>
> it does work from the offhand - all my hits were offhand hits, mainhand was a Shijo.
> so we need latent effect for that proc. vs sub-family only. weapons in ve lugunon gonna need diff subfamily ID than the ones in the rest of the game.
> 
> we can on-hit listener, and after dealing 0 dmg from the melee hit remove immortal status and remove his damage taken reduction so the en-damage kills him.

I don't think I finished hooking up that latent. Either main family or subfamily wasn't hooked up yet. its an easy thing to do I just forgot about it is all.


edit: immortal status = the same setUnkillable you used. we just don't want a special add effect type just for him is all.

TeoTwawki: pardon me, misspoke subfamily isn't a thing, its family and super family:

```sql
  `familyID` smallint(4) unsigned NOT NULL,
  `family` tinytext,
  `superFamilyID` smallint(4) unsigned NOT NULL DEFAULT 0,
  `superFamily` tinytext,
  `ecosystemID` tinyint(2) unsigned NOT NULL DEFAULT 0,
  `ecosystem` tinytext,
```
```
"family, superFamily, ecosystem" relationship: Korrigan, Mandragora, Plantoid
```

TeoTwawki: the family latent this would need is hooked up, it was superfamily I never finished, which I'll get in soon (workin on it right now) so I can put _bano del sol_ to rest

m241dan:
Well hurry up then! :D

Another one bites the dust.

scripts/globals/additional_effects.lua
LegenDAirie: @TeoTwawki Did you ever finish the "superfamily"?
I'm looking to fix the Bano Del Sol proc rate bug.

PR #1883, opened 2022-06-05
Steam Cleaner spawn mechanics retail accuracy

m241dan:
**_I affirm:_**

Made it so Steam Cleaner spawns from Detectors on the lower levels.

1. Pull detect from the lower level
2. Kill Caretakers until Steam Cleaner spawns (set spawn chance to 100 to speed up the process)


scripts/zones/VeLugannon_Palace/mobs/Detector.lua
```diff
@@ -10,25 +10,51 @@ entity.onMobSpawn = function(mob)
     mob:setLocalVar("petCount", 1)
 end
 
+local getMobToSpawn = function(detector)
+    local ID = require("scripts/zones/VeLugannon_Palace/IDs")
```
Xaver-DaRed: `local ID = require("scripts/zones/VeLugannon_Palace/IDs")`

This line should go above current line 5 `require("scripts/globals/regimes")`

scripts/zones/VeLugannon_Palace/mobs/Detector.lua
```diff
@@ -10,25 +10,51 @@ entity.onMobSpawn = function(mob)
...
+        -- this check is needed to prevent the detector from spawning BOTH steam cleaner and a caretaker
+        -- by essentially "re-returning" steam cleaner, the subsequent logic will know that he's already spawned as well
+        -- and won't try to respawn him
+        -- This silly check is basically needed because we fall back on a "my id + 1" convention instead of actually
+        -- doing some sort of "this detector owned/spawned this caretaker and while he is alive, I shouldn't summon
+        -- something else".
+        if detector:getLocalVar("SpawnedSC") == 1 then
+            return steamCleaner
+        end
+    elseif canSpawnSteamCleaner and math.random(100) < steamCleanerSpawnChance then
```
Xaver-DaRed: I don't know if I'm following this correctly but...
Wouldn't this prevent ALL detectors from spawning adds the moment Steam Cleaner is spawned?

m241dan: No, because if it isn't the detector that spawned it, it will fall through to the bottom return. That's whole purpose huge but obviously insufficient comment for the if statement on line 27.

scripts/zones/VeLugannon_Palace/mobs/Detector.lua
m241dan: Will do.

scripts/zones/VeLugannon_Palace/mobs/Detector.lua
WinterSolstice8: It's not obvious at a glance, but looking carefully:

The set local var of "SpawnedSC" is assigned to the detector that spawned it, and only that one specific detector will `return SteamCleaner` on subsequent checks. The other detectors will fall through to the bottom return. 

I think adding a comment on the get/set local vars about how they are used on _only_ the detector that spawns it should clear that up for future code spelunkers

scripts/zones/VeLugannon_Palace/mobs/Detector.lua
```diff
@@ -10,25 +10,51 @@ entity.onMobSpawn = function(mob)
     mob:setLocalVar("petCount", 1)
 end
 
+local getMobToSpawn = function(detector)
+    local ID = require("scripts/zones/VeLugannon_Palace/IDs")
+    local detectorID = detector:getID()
+    local canSpawnSteamCleaner = GetServerVariable("[POP]SteamCleaner") < os.time() and utils.contains(detectorID, ID.mob.STEAM_CLEANER_DETECTORS)
```
WinterSolstice8, line 16: what happens if someone holds steam cleaner for 4 hours? Will steam cleaner be attempted to spawn again by another detector? I suppose an `and not steamCleaner:isSpawned()` after reorganizing it after `local steamCleaner = GetMobByID(ID.mob.STEAM_CLEANER)` would fix that.

WinterSolstice8: I just realized TOD would not be set until it's killed so this is probably a non issue

scripts/zones/VeLugannon_Palace/mobs/Detector.lua
m241dan: Moved it above line 5 as instructed.

scripts/zones/VeLugannon_Palace/mobs/Detector.lua
m241dan: > It's not obvious at a glance, but looking carefully:
> 
> The set local var of "SpawnedSC" is assigned to the detector that spawned it, and only that one specific detector will `return SteamCleaner` on subsequent checks. The other detectors will fall through to the bottom return.
> 
> I think adding a comment on the get/set local vars about how they are used on _only_ the detector that spawns it should clear that up for future code spelunkers

Let me know if you think these comments are better.

Xaver-DaRed (approved):

PR #1894, opened 2022-06-06
 hook up super family to actually use it and add latent effect trigger for it

TeoTwawki:
**_I affirm:_**

Awhile back i added super families to the mob sql table in prep for some things. This PR continues with it and makes use of super families in the mob_family_system sql table so that we can't check against those for example with the latent this PR also adds.

while targeting a mob:
`!exec player:PrintToPlayer(tostring(target:getSuperfamily()))`

A follow up PR will be using the latent this PR adds. Latent was only as far as ensuring some prints happened and nothing else got broken.


WinterSolstice8 (commented):
Looks good, though it seems CI is complaining about tabs vs spaces.

TeoTwawki:
> Looks good, though it seems CI is complaining about tabs vs spaces.

yep they predate me but ima tack on nuking them anyways!

TeoTwawki:
We'll learn them tabs! And those Cyrillic comments too!

WinterSolstice8 (approved):

Issue #943, opened 2021-11-10
Rolanberry Turkey (food) not usable, missing multiple elements to make usable


RAIST5150:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: Everything

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Got some of these during last login campaign, but cannot use them.

https://ffxiclopedia.fandom.com/wiki/Rolanberry_Turkey
Food Effects: (4 hours, All Races)

    STR +10
    Attack +20% (Max. 120 @ 600 Base Attack)
    Ranged Attack +20% (Max. 120 @ 600 Base Ranged Attack)
    Counter +10
    Resist Amnesia +10

It is present in item_basic.sql:
```lua
INSERT INTO `item_basic` VALUES (6576,0,'turkey_with_rolanberry_sauce','rol._turkey',1,26184,0,1,0);
```
It shows proper data on it when you view it in inventory and such, just does not appear to have the relevant LUA and SQL entries to be able to use them. Initially, thought it might be a group food, but the wikis still list it as player only.  So it may be something that could be loosely modeled after existing 4hr foods (roast turkey, etc.)

I expect we have some means of handling the counter effect/mod, but there was a question in discord as to whether we have the resist amnesia effect/mod coded yet to be able to mark it up or not.  There does not appear to be an addMod scripted for the baramnesia spell effect, so this may be true?

I did spot entries for resisting amnesia in item_mods.sql though.  Not sure if any of that data could be useful in marking up this food item and possibly the baramnesia spell effect or not though:

```lua
INSERT INTO `item_mods` VALUES (10546,1,27); -- URJA GLOVES, DEF:27 AGI+6 Attack+5 Enhances "Resist Amnesia" effect
INSERT INTO `item_mods` VALUES (10546,253,1); -- RESAMNESIA
INSERT INTO `item_mods` VALUES (10547,1,28); -- STHIRA GLOVES, DEF:28 AGI+7 Attack+6 Enhances "Resist Amnesia" effect
[...]

PR #1898, opened 2022-06-07
[SQL] [Lua] Add "Turkey with Rolanberry Sauce" food issue #943

CatsEyeXI:
**_I affirm:_**

Closes: #943
^ Making Rolanberry Turkey usable.

1. Check attributes
2. !additem 6576
3. Use Item
4. Check attributes

I am not aware of how to check my current "Counter" and "Amnesia Resist" but assuming these mods work there should be no problem here. 
STR, Attack and Ranged Attack confirmed.

### DIFF
--- a/scripts/globals/items/turkey_with_rolanberry_sauce.lua
+++ b/scripts/globals/items/turkey_with_rolanberry_sauce.lua
@@ -1,0 +1,50 @@
+-----------------------------------
+-- ID: 6576
+-- Item: rolanberry_turkey
+-- Food Effect: 4 Hours, All Races
+-----------------------------------
+-- STR +10
+-- Attack +20% (Max. 120)
+-- Ranged Attack +20% (Max. 120)
+-- "Counter" +10
+-- "Resist Amnesia" +10
+-- https://www.bg-wiki.com/ffxi/Rol._Turkey
+-----------------------------------
+require("scripts/globals/status")
+require("scripts/globals/msg")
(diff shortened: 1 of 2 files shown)


scripts/globals/items/turkey_with_rolanberry_sauce.lua
```diff
@@ -0,0 +1,50 @@
...
+-- https://www.bg-wiki.com/ffxi/Rol._Turkey
+-----------------------------------
+require("scripts/globals/status")
+require("scripts/globals/msg")
+-----------------------------------
+local item_object = {}
+
+item_object.onItemCheck = function(target)
+    local result = 0
+    if (target:hasStatusEffect(xi.effect.FOOD) or target:hasStatusEffect(xi.effect.FIELD_SUPPORT_FOOD)) then
```
Xaver-DaRed: Remove the outer parenthesis

CatsEyeXI:
looks like startup check failed because port was already in use

Xaver-DaRed (approved):
Item cannot be easily captured, so I'm OK with the TODO

zach2good:
Teo also agrees in the other issue:
> Do foods have diff use animations? I'd never noticed/paid attention. Even so, you can get away with copying the value from a similar food and putting a todo comment if the animation is wrong. I have a couple of these and when I finally use them, I can fix it if nobody else has.

Issue #835, opened 2021-10-21
Interrupting an item use with movement doesn't display a message


ShiyoKozuki:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: EDITME

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
"Uses Echo Drops, but it fails to activate" - Retail.
MSG ID 62.
https://github.com/Windower/Lua/wiki/Message-IDs

TeoTwawki:
:star:

CatsEyeXI:
We display the wrong msg (445)
![image](https://user-images.githubusercontent.com/65316697/172285155-e58f934e-9d9a-44c5-8edf-cbd08b72d824.png)

Verified in retail, should be 62, going to try to fix this

CatsEyeXI:
I think the fix is pretty simple, will submit a PR shortly.
![image](https://user-images.githubusercontent.com/65316697/172286352-c579a19f-fb34-43d8-b722-442072bbfd7f.png)

PR #1901, opened 2022-06-07
[Fix] Update message when interrupting item usage

CatsEyeXI:
**_I affirm:_**

Addresses issue #835 where incorrect message is displayed when interrupting an item use.

1. Attempt to use an item while moving

### DIFF
--- a/src/map/ai/states/item_state.cpp
+++ b/src/map/ai/states/item_state.cpp
@@ -240,10 +240,11 @@ void CItemState::TryInterrupt(CBattleEntity* PTarget)
         UpdateTarget(m_PEntity->IsValidTarget(m_targid, m_PItem->getValidTarget(), m_errorMsg));
     }
 
-    uint16 msg = 62; // the item fails to activate
+    uint16 msg = 445; // you cannot use items at this time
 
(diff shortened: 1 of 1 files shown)


src/map/ai/states/item_state.cpp
```diff
@@ -240,7 +240,7 @@ void CItemState::TryInterrupt(CBattleEntity* PTarget)
         UpdateTarget(m_PEntity->IsValidTarget(m_targid, m_PItem->getValidTarget(), m_errorMsg));
     }
 
-    uint16 msg = 445; // you cannot use items at this time
+    uint16 msg = 62; // the item fails to activate
 
     if (HasMoved() || m_PEntity->StatusEffectContainer->HasPreventActionEffect())
     {
```
Xaver-DaRed, line 246: instead of changing the default message, change the message used inside this block. Like so:
```c++
if (HasMoved() || m_PEntity->StatusEffectContainer->HasPreventActionEffect())
{
    msg           = MSGBASIC_ITEM_FAILS_TO_ACTIVATE;
    m_interrupted = true;
}
```

TeoTwawki (approved):
Bonus points if you add ID 445 to the ENUM and use it on line 243.

The points are like those in "who's line is it anyway?" though.

Approved either way.

PR #1903, opened 2022-06-07
[core] Add some basic warnings for SQL query latency after startup

zach2good:
**_I affirm:_**

Adds some optional additional logging to track when SQL queries run long. Will be helpful in identifying when db's need to be cleared out. 

Run an artificially long query, or set `m_LatencyWarning` to true during startup, to see those long queries generate warnings.


TeoTwawki (approved):
looks reasonable

PR #1796, opened 2022-05-24
Begin implementing Garrison (in feature branch)

RunAwayDSP:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

Co-Authored-by: Zach2good
Additional Coding support from entire ASB dev teams
Additional Capture data from Wiggo
Other Credits: Big thanks to Ophelia, TeoTwawki, and Aerec for help with various aspects along the way
 
## What does this pull request do?
All Zones have a working Garrison

## Included Settings
changing these allows for easier testing as well
    GARRISON_LOCKOUT             = 1800;    -- Time in seconds before a new garrison can be started (default: 1800)
    GARRISON_TIME_LIMIT          = 1800; -- Time in seconds before lose ongoing garrison (default: 1800)
    GARRISON_ONCE_PER_WEEK       = 0;    -- Set to 1 to bypass the limit of one garrison per Conquest Tally Week.
    GARRISON_PARTY_LIMIT         = 18;   -- Set to max party members you want to do garrison (default: 18).
    GARRISON_NATION_BYPASS       = 0;    -- Set to 1 to bypass the nation requirement.
    GARRISON_RANK                = 2;    -- Set to minumum Nation Rank to start Garrison (default: 2).

## What is not included / will not be completed in this PR
Wave Data Captures to get correct wave order
Capture look/name of all NPCs
Update npcs table on Conquest Update with new look and name

## Unable to complete
Check to see if there is Ballista in the Zone
Battlefield Effect (need another way to prevent outside help)
Get names to actually display

## How to play test party size
	--gets party size for spawning each NPC
    party = #party
Change the #party to up to 18

PR #429, opened 2021-07-05
Mannequin Packets, Logic + Unlock Quest

zach2good:
image

For the fourth or fifth time in recent memory:
```
That looks easy, shouldn't take me more than a few hours...
                       -zach2good
```

This was a neverending nightmare.

**Review Guide**
- SmallPacket0x03B: Mannequin Equip
- CInventoryCountTo80Packet: Regular and Mannequin update
- Quest: Its_Raining_Mannequins.lua (Not tested yet)
- Fyi_Chalmwoh.lua logic, ie. the mannequin shop/trader
- New bindings to modify exdata of items given to the player: https://github.com/LandSandBoat/server/pull/429/commits/1346b65fbc0d099cc4043297944782f6ce67cfbf

**After noticing that the mannequin exdata is mapped out by Windower after I did Soultrappers that way, I thought this should be quick and easy to figure out**
https://github.com/Windower/Lua/blob/dev/addons/libs/extdata.lua#L1891-L1916

**Inspecting exdata**
image

**Getting UI to respond**

image

**Pestering Windower Folks for clues**

image

**Writing addons to debug**

image

**Packet deep dive**
image

**@InoUno extending PVLV to help me out**
image

**Not pictured: Me running Garrison with Lusi to get my own mannequin so I could test properly!**

**Success!**
image

**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] that I agree to LandSandBoat's [Limited Contributor License Agreement](https://github.com/LandSandBoat/server/blob/base/.github/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [x] that I have **read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)**
- [x] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

PR #424, opened 2021-07-04
Garrison Plumbing

zach2good:
My confrontation helper from the AMK Cardian fight established some very useful patterns. Garrison is pretty easy 👀 

**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] that I agree to LandSandBoat's [Limited Contributor License Agreement](https://github.com/LandSandBoat/server/blob/base/.github/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [x] that I have **read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)**
- [x] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

PR #1905, opened 2022-06-07
[WIP] Merge feature/garrison into base

zach2good:
**_I affirm:_**

This is a long-term tracking PR to watch `feature/garrison` and its journey from feature branch + `everything` into `base`.

Previous PRs:
https://github.com/LandSandBoat/server/pull/1796
https://github.com/LandSandBoat/server/pull/429
https://github.com/LandSandBoat/server/pull/424

Play some Garrison, build a mannequin. etc.


scripts/settings/default/main.lua
```diff
@@ -199,4 +199,12 @@ xi.settings =
     ENM_COOLDOWN                 = 120,  -- Number of hours before a player can obtain same KI for ENMs (default: 5 days)
     FORCE_SPAWN_QM_RESET_TIME    = 300,  -- Number of seconds the ??? remains hidden for after the despawning of the mob it force spawns.
     GOBBIE_BOX_MIN_AGE           = 45,   -- Minimum character age in days before a character can sign up for Gobbie Mystery Box
+
+    -- GARRISON
+    GARRISON_LOCKOUT             = 1800; -- Time in seconds before a new garrison can be started (default: 1800)
```
zach2good: These lines should end with commas, not `;`'s

scripts/zones/Beaucedine_Glacier/IDs.lua
```diff
@@ -29,6 +29,7 @@ zones[xi.zone.BEAUCEDINE_GLACIER] =
         YOU_CANNOT_ENTER_DYNAMIS       = 7882,  -- You cannot enter Dynamis - [Dummy/San d'Oria/Bastok/Windurst/Jeuno/Beaucedine/Xarcabard/Valkurm/Buburimu/Qufim/Tavnazia] for <number> [day/days] (Vana'diel time).
         PLAYERS_HAVE_NOT_REACHED_LEVEL = 7884,  -- Players who have not reached level <number> are prohibited from entering Dynamis.
         DYNA_NPC_DEFAULT_MESSAGE       = 7894,  -- There is an unusual arrangement of branches here.
+        GARRISON                       = 8244,  -- Hm? What is this? How do I know this is not some trick?.
```
zach2good: This could do with a better name, also, `?.` looks kind of fishy, my extracts list it as:
```
Hm? What is this? %? How do I know this is not some [San d'Orian/Bastokan/Windurstian] trick?
```

(we need this to line up with the extract, for our update tooling to continue to work)

https://raw.githubusercontent.com/zach2good/UpdateExtractor/versions/2021-december/out/scripts/zones/Beaucedine_Glacier/Text.lua.raw

scripts/zones/Beaucedine_Glacier/npcs/Parledaire_RK.lua
```diff
@@ -14,11 +16,48 @@ local guardRegion = xi.region.FAUREGANDI
...
+        player:messageSpecial(text.GARRISON, trade:getItem():getID(), player:getNation())
+    end
 end
 
 entity.onTrigger = function(player, npc)
-    xi.conq.overseerOnTrigger(player, npc, guardNation, guardType, guardEvent, guardRegion)
+    local zoneId = npc:getZoneID()
+    local text = zones[zoneId].text
+    local win = player:getZone():getLocalVar(string.format("[GARRISON]Treasure_%s", zoneId))
+    local won = player:getCharVar("Garrison_Won")
```
zach2good: Could these be condensed down into a single `GarrisonStatus` var?

scripts/zones/Beaucedine_Glacier/IDs.lua
zach2good: Also `GARRISON_BASE` would be an OK name

scripts/zones/Buburimu_Peninsula/npcs/Bonbavour_RK.lua
```diff
@@ -14,11 +16,48 @@ local guardRegion = xi.region.KOLSHUSHU
...
+    elseif
+        win < os.time() and
+        win > 0
+    then
+        -- Trader took too long to claim prize lose
+        player:messageSpecial(text.GARRISON + 37)
+        xi.garrison.onLose(player, npc)
+    else
+        xi.conq.overseerOnTrigger(player, npc, guardNation, guardType, guardEvent, guardRegion)
+    end
```
zach2good: This logic looks quite similar between guards, could probably be extracted into a shared function

sql/mob_droplist.sql
```diff
@@ -17416,29 +17416,31 @@ INSERT INTO `mob_droplist` VALUES (3222,2,0,1000,1455,0); -- one_byne_bill
...
+INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15101,10);
+INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15117,10);
+INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15122,10);
+INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15123,10);
+INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15140,10);
+INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,15142,10);
+INSERT INTO `mob_droplist` VALUES (3222,0,0,1000,16360,10);
+INSERT INTO `mob_droplist` VALUES (3223,0,0,1000,16017,150);
+INSERT INTO `mob_droplist` VALUES (3224,0,0,1000,1604,100); -- Mannequin Legs
+INSERT INTO `mob_droplist` VALUES (3224,0,0,1000,1605,100); -- Mannequin Feet
```
zach2good, line 17443: I chewwed up these db changes during my rebasing. They'll need to adopt the new `@COMMON` syntax. Also, these could be PR'd to `base` ahead of time so that they don't conflict with any other db changes.

scripts/zones/Beaucedine_Glacier/npcs/Parledaire_RK.lua
RunAwayDSP: needs better naming:
win is when the trader has won the garrison but has yet to trigger the win with the guard.
won is after the trader has triggered the win other players can now talk to npc to remove effects and get the won text.

needs 2 different vars though for the 2 different triggers

RunAwayDSP: fixed in #1919

scripts/zones/Buburimu_Peninsula/npcs/Bonbavour_RK.lua
RunAwayDSP: fixed in #1919

scripts/zones/Beaucedine_Glacier/npcs/Akane_IM.lua
```diff
@@ -15,10 +17,12 @@ local guardEvent  = 32761
 
 entity.onTrade = function(player, npc, trade)
     xi.conq.overseerOnTrade(player, npc, trade, guardNation, guardType)
+    xi.garrison.onTrade(player, npc, trade, guardNation)
 end
 
 entity.onTrigger = function(player, npc)
     xi.conq.overseerOnTrigger(player, npc, guardNation, guardType, guardEvent, guardRegion)
+    xi.garrison.onTrigger(player, npc)
 end
```
zach2good, line 26: These sort of things should probably not run immediately after each other, even if the implementations inside these functions are safe to do so.

```lua
if xi.conq.overseerOnTrade(player, npc, trade, guardNation, guardType) then
    return
else if xi.garrison.onTrade(player, npc, trade, guardNation) then
    return
end
```

This guarantees that only one of these runs, never both. Looks like all the guards need this

scripts/zones/Buburimu_Peninsula/Zone.lua
```diff
@@ -27,6 +28,10 @@ zone_object.onInitialize = function(zone)
     xi.helm.initZone(zone, xi.helm.type.LOGGING)
 end
 
+zone_object.onZoneTick = function(zone)
+    xi.garrison.buildNpcTable(zone)
+end
+
```
zach2good: This looks very suspicious.

Should you be building the table on every tick?

If it's a thing that needs to be built once, it can go in zone init, or whatever.

If it's something that _should_ happen every tick, this should probably indicate that it's only for when garrison is active

```
if xi.garrison.isGarrisonActive(zone) then
    xi.garrison.tickGarrison(zone)
end
```

etc.

zach2good (commented):
Looks like it's getting closer 👍

scripts/zones/Buburimu_Peninsula/Zone.lua
RunAwayDSP: not every tick its there for testing for now.. needs to be moved to something like on conquest update and need to run a check before start to possibly build it as well

scripts/zones/Beaucedine_Glacier/npcs/Akane_IM.lua
RunAwayDSP: so the trade items dont overlap I found no issues getting both to work but I will double check this

zach2good: It doesn't matter if they logically won't overlap, at a glance it's impossible to tell if you're a passer-by. These should be split in a way that makes it obvious that only one can be firing at a time (if/else blocks)

zach2good:
With this: https://github.com/LandSandBoat/server/pull/2106
You should be able to get the NPCs working correctly

zach2good:
I don't know if the special spawning animation is used in Garrison, but its available: https://github.com/LandSandBoat/server/pull/2144

RunAwayDSP:
> I don't know if the special spawning animation is used in Garrison, but its available: #2144

There is its on my todo list just been busy

zach2good:
The majority of the useful things from this have been ported into base with:

https://github.com/LandSandBoat/server/pull/2673
https://github.com/LandSandBoat/server/pull/2675
https://github.com/LandSandBoat/server/pull/2681
https://github.com/LandSandBoat/server/pull/2853
https://github.com/LandSandBoat/server/pull/2684
https://github.com/LandSandBoat/server/pull/2203
https://github.com/LandSandBoat/server/pull/2144
https://github.com/LandSandBoat/server/pull/2953

PR #1909, opened 2022-06-07
[SOA] 3-6-3 Stonewalled fight

zach2good:
**_I affirm:_**

Adds the SOA 3-6-3 fight

!instance 27100


scripts/zones/Cirdas_Caverns_U/IDs.lua
```diff
@@ -14,6 +14,7 @@ zones[xi.zone.CIRDAS_CAVERNS_U] =
         ITEM_OBTAINED                 = 6390, -- Obtained: <item>.
         GIL_OBTAINED                  = 6391, -- Obtained <number> gil.
         KEYITEM_OBTAINED              = 6393, -- Obtained key item: <keyitem>.
+        KEYITEM_LOST                  = 6394, -- Lost key item: <keyitem>.
```
claywar, line 14: This has been brought up to me before, but I'm curious now for general thoughts: Since the "Lost KI" message always seems to be obtained+1, should we migrate that message and delete to an npcUtil function?

TeoTwawki: I think its clearer to actually see it separately from the obtained one

PR #1940, opened 2022-06-10
Remove pet godmode rings

bope12:
**_I affirm:_**
- [ ] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

Removes mods from Evoker's Ring and Pacifist Ring that gives pets godmode like stats

equip the two rings watch your pets be normal


Xaver-DaRed (approved):
How has this gone unnoticed until now?

claywar (approved):

TeoTwawki:
> How has this gone unnoticed until now?

Was a recent PR that they slipped in with. No doubt somebodies custom they thought was removed before making the PR.

Went unnoticed because we weren't able to line by line every single mod in a sea of numbers.

PR #1945, opened 2022-06-10
Settings (and Lua) refactor

zach2good:
**_I affirm:_**

### For Users

- ALL settings now live in the `settings/` folder, which has a `default/` subfolder. The files in the defaults subfolder are read to populate the defaults. If you use all defaults (like for local dev), you don't need to bother copying any files anywhere.
- If you copy defaults into the main settings folder, the defaults will get read one-by-one, and then your main settings will be read to overwrite the defaults.
- Network settings have been moved into `settings/network.lua`, and are shared by all exes
- Each exe gets a file for its settings, plus a "main" settings which is for the game in general

### For Devs

- Split out basic Lua init into common, and give it to all exes
- Lua-based settings loading
- Easy, type-safe-ish access to settings from C++
- Consolidate all settings into a single place

## Review Guide

This is a big boy, so here is how to review it:

#### Lua:

- At the top of a lot of scripts, `require("scripts/settings/")` has been replaced with `require("scripts/globals/settings")`. Go read `scripts/globals/settings` for an explanation of why.
-  All instances of `xi.settings` have been swapped out for `xi.settings.main` in Lua land

#### CPP:

- Introduced `common/lua.h/cpp` and `common/settings.h/cpp`. You can inspect these if you want, but it's mostly voodoo.
- All `config` structs, and loading have been ripped out, in favour of loading from the Lua settings files at startup.
- All accesses to config structs are replaced with the equivalent calls to `settings:get<type>("name")`
- Some additions to string utils

#### Settings:

- `conf` files are gone, RIP
- The settings files in `scripts/settings` are now in `settings`

#### Logging:

- Overhaul of the logging categories, replace old calls with consolidated new ones

### Other

- dbool has been updated to handle the new layout
- Loading settings from environment variables

## Before merging

- Go through each setting in `main` and make sure they're properly accounted for, and defaulted, in the new system.
- Play the game a bunch. Combat, missions, quests, crafting, airships, etc.
- Updating various readmes
- Updating the wiki
- Updating all the comments in the settings


zach2good:
C++-side testing:
image

cocosolos:
I'm sure you're already aware of this but dbtool uses a few of the conf files for its functions. There's obviously the db connect info, but also the default CLIENT_VER is compared to the user copy to see if that should be updated. There's also a DB_VER that's inserted as a commented line into version.conf that is compared to the current hash for express update purposes. This one never got additional use in core so could probably be moved into dbtool's config.yaml file.

cocosolos:
This is great btw. I love having all the settings in one place and reading the defaults instead of requiring more user input is icing on the cake.

zach2good:
It had crossed my mind, but I forgot soon after, I'll make sure to add it to my list 👍

zach2good:
Added a review guide if anyone wants to take a look before this is ready-ready

tools/dbtool.py
```diff
@@ -139,65 +140,61 @@ def fetch_credentials():
...
+    current_client = release_version = release_client = None
+
     try:
         release_version = repo.git.rev_parse(repo.head.object.hexsha, short=4)
     except: # lgtm [py/catch-base-exception]
         print(colorama.Fore.RED + 'Unable to read current version hash.')
+
+    filename = '../settings/default/login.lua'
+    if os.path.exists('../settings/login.lua'):
+        filename = '../settings/login.lua'
```
cocosolos: Since this is checking for the "release" client (upstream client version) I believe we should only be checking the default settings here, followed by a separate check of the user settings for the "current" client. This way if the client version has been edited it will detect the diff.

tools/dbtool.py
```diff
@@ -230,26 +229,43 @@ def fetch_module_files():
...
                     import_files.append(diff.a_path[4:])
                 express_enabled = True
             else:
                 express_enabled = False
-                if len(repo.commit(current_version).diff(release_version,paths='tools/migrations')) > 0:
+                if len(repo.commit(db_ver).diff(release_version,paths='tools/migrations')) > 0:
                     express_enabled = True
         except: # lgtm [py/catch-base-exception]
-            print(colorama.Fore.RED + 'Error checking diffs.\nCheck that hash is valid in ../conf/version.conf.')
+            print(colorama.Fore.RED + 'Error checking diffs.\nCheck that hash is valid in ../settings/login.lua.')
```
cocosolos: This should point to `config.yaml` now instead of `../settings/login.lua`.

tools/dbtool.py
```diff
@@ -265,28 +281,28 @@ def fetch_files(express=False):
...
+    global db_ver
     success = False
     update_client = auto_update_client
     if not silent and current_client != release_client:
         update_client = input('Update client version? [y/N] ').lower() == 'y'
     update_client = update_client and release_client
+    db_ver = release_version
+    write_configs()
+
+    filename = '../settings/default/login.lua'
```
cocosolos: This function shouldn't check default files. If the client version hasn't been changed and is just following upstream then the rest of this function can be skipped.

tools/dbtool.py
```diff
@@ -505,7 +517,7 @@ def restore_backup():
                 backup_file = backups[choice - 1]
                 print(colorama.ansi.clear_screen())
                 print(colorama.Fore.RED + 'If this is a full backup created by this tool, it is recommended to manually change \n'
-                    'the DB_VER in ../conf/version.conf to the hash sequence in the filename, after \n'
+                    'the DB_VER in ../settings/login.lua to the hash sequence in the filename, after \n'
```
cocosolos: This should point to `config.yaml` now instead of `../settings/login.lua`.

tools/dbtool.py
```diff
@@ -545,8 +557,8 @@ def menu():
     print(colorama.Fore.GREEN + 'o' + colorama.Fore.RED + '---------------------------------------' + colorama.Fore.GREEN + 'o\n' + colorama.Fore.RED +
           '| ' + colorama.Style.RESET_ALL + 'LandSandBoat Database Management Tool ' + colorama.Fore.RED + '|\n'
           '| ' + colorama.Style.RESET_ALL + str('Connected to ' + database).center(30) + colorama.Fore.RED + '        |')
-    if current_version:
-        print(colorama.Fore.RED + '| ' + colorama.Style.RESET_ALL + str('#' + current_version).center(30) + colorama.Fore.RED + '        |')
+    if db_ver:
+        print(colorama.Fore.RED + '| ' + colorama.Style.RESET_ALL + str('#' + db_ver).center(30) + colorama.Fore.RED + '        |')
     print(colorama.Fore.GREEN + 'o' + colorama.Fore.RED + '---------------------------------------' + colorama.Fore.GREEN + 'o')
     if express_enabled:
         print(colorama.Fore.RED + '|' + colorama.Fore.GREEN + 'e' + colorama.Style.RESET_ALL + '. Express Update ' + str('(#' + release_version + ')').ljust(14) + colorama.Fore.RED + '       |')
```
cocosolos: Minor formatting issue bugging me here with the centering.
```py
          '| ' + colorama.Style.RESET_ALL + 'LandSandBoat Database Management Tool' + colorama.Fore.RED + ' |\n'
          '| ' + colorama.Style.RESET_ALL + str('Connected to ' + database).center(37) + colorama.Fore.RED + ' |')
    if db_ver:
        print(colorama.Fore.RED + '| ' + colorama.Style.RESET_ALL + str('#' + db_ver).center(37) + colorama.Fore.RED + ' |')
    print(colorama.Fore.GREEN + 'o' + colorama.Fore.RED + '---------------------------------------' + colorama.Fore.GREEN + 'o')
    if express_enabled:
        print(colorama.Fore.RED + '|' + colorama.Fore.GREEN + 'e' + colorama.Style.RESET_ALL + '. Express Update ' + str('(#' + release_version + ')').ljust(21) + colorama.Fore.RED + ' |')
```

tools/dbtool.py
zach2good: I don't think I understand what you mean, can you PR what you think it should be?

tools/dbtool.py
zach2good: Same here

tools/dbtool.py
cocosolos: #2097
Basically it should try to read from both default/login.lua _and_ settings/login.lua, but only write to settings/login.lua.

zach2good:
OK, I think this is now safe and ready to go, once this gets some green ticks, I'll adjust all of our docs to reflect the changes and then go ahead with the merge

claywar (approved):

PR #1959, opened 2022-06-12
[mobskills] add 1.0 to breath attacks combinedDamageTaken

m241dan:
**_I affirm:_**

This fixes Breath attacks doing 0 damage most of the time. IE, the
damage line that uses combinedDamageTaken pretty much just mutliplies
by zero all the time. In fact, if you have breath damage taken, you
get healed! Fun, but not accurate.

1. Force Greater Manticore to only use Heat Breath
2. Go to Cape Terrigan and fight, watch 0 dmg Heat Breaths


Xaver-DaRed (approved):

claywar (approved):
Seems reasonable to me.  Should we be capping the upper bound though? (Note, this existed prior, so not PR-specific)

m241dan:
> Seems reasonable to me. Should we be capping the upper bound though? (Note, this existed prior, so not PR-specific)

I had that same thought, but I'm not sure the harm at the moment. 

I'd be more concerned about the mob MAB / MDB caps.

PR #1961, opened 2022-06-12
[trust] Addition of trust Sylvie UC

Omnione:
**_I affirm:_**

Adds the trust Sylvie UC to a somewhat decent state

The Indi spell selection gambit for macc is not perfect but its enough to get let her decide if the player needs more macc or not.
All the main stuff is there, No Nott ws yet same as Apururu, so thats a TODO

Tested on multiple levels and jobs

### DIFF
--- a/scripts/globals/gambits.lua
+++ b/scripts/globals/gambits.lua
@@ -75,6 +75,8 @@ ai.select =
     BEST_AGAINST_TARGET = 6,
     BEST_SAMBA          = 7,
     HIGHEST_WALTZ       = 8,
+    ENTRUSTED           = 9,
+    BEST_INDI           = 10,
 }
 ai.s = ai.select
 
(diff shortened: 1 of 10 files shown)


zach2good:
Coooool!

scripts/globals/gambits.lua
```diff
@@ -75,6 +75,8 @@ ai.select =
     BEST_AGAINST_TARGET = 6,
     BEST_SAMBA          = 7,
     HIGHEST_WALTZ       = 8,
+    ENTRUSTED           = 9,
+    BESTINDI            = 10,
```
zach2good: If this is two words, can it be `BEST_INDI`?

src/map/ai/helpers/gambits_container.cpp
```diff
@@ -341,6 +341,25 @@ namespace gambits
                         //    controller->Cast(target->targid, static_cast<SpellID>(spell_id.value()));
                         //}
                     }
+                    else if (action.select == G_SELECT::BESTINDI)
```
zach2good: Same here: `BEST_INDI`

zach2good (approved):
Looks good 👍 

We don't have unity rankings and stuff do we? I haven't been paying attention. How do you use her?

scripts/globals/gambits.lua
Omnione: yes you'r right, it should be

PR #1969, opened 2022-06-12
[Crash + Bugfix] Aern Crash Fixes + NO_DROP fix

### LINKED ISSUE #1955: 🐛 Aern's Reraise causes Server Crash If Pet is Killer and Pet is Despawned During Reraise

## Branch affected by issue 

`base`

## Steps to reproduce

Fight any Aern with the Aern family mixin. 

Attack with Carbuncle
!hp 0 (while ensuring aern is fighting carby)
release carby when aern dies

server crash

## Expected behavior

Aern aggros master since carby is no longer alive. 

No Crash

### Crash DMP

[xi_map.exe_11-6_22-14-49.log](https://github.com/LandSandBoat/server/files/8885094/xi_map.exe_11-6_22-14-49.log)

### LINKED ISSUE #1964: 🐛 Geodes/Avatarites drop when Mobmod NO_DROPs is set

## Branch affected by issue 

`base`

## Steps to reproduce

Give a mob the mobmod: NO_DROPS = 1
Watch mobs still drop avatarites and geodes.

image
image

## Expected behavior

No drops whatsoever

The-Aerec:
**_I affirm:_**

Fixes the following bugs:
The server will no longer crash when pets are despawned during Aern reraise mechanics. Fixes https://github.com/LandSandBoat/server/issues/1955
The mobmod: NO_DROP - Now also stops Seals + Avatarites + Geodes from dropping as well. Fixes https://github.com/LandSandBoat/server/issues/1964

Kill a mob with mobmob no_drop = 1 and observe no drops of any kind
Kill an Aern and do a load of things during its reraise mechanic - Observe no crash


src/map/entities/mobentity.cpp
```diff
@@ -986,15 +986,15 @@ void CMobEntity::DropItems(CCharEntity* PChar)
 
     bool validZone = ((Pzone > 0 && Pzone < 39) || (Pzone > 42 && Pzone < 134) || (Pzone > 135 && Pzone < 185) || (Pzone > 188 && Pzone < 255));
 
-    if (validZone && charutils::CheckMob(m_HiPCLvl, GetMLevel()) > EMobDifficulty::TooWeak)
+    if (!getMobMod(MOBMOD_NO_DROPS) && validZone && charutils::CheckMob(m_HiPCLvl, GetMLevel()) > EMobDifficulty::TooWeak)
```
zach2good, line 989: 👍

zach2good (approved):

scripts/mixins/families/aern.lua
```diff
@@ -20,21 +20,51 @@ g_mixins.families.aern = function(aernMob)
                     reraises = 1
                 end
             end
+            mob:setMobMod(xi.mobMod.NO_DROPS, 0) -- Drops on if Not reraising
             if curr_reraise < reraises then
+                mob:setMobMod(xi.mobMod.NO_DROPS, 1)  -- Drops off if reraising
                 local dropid = mob:getDropID()
-                mob:setDropID(0)
+                local owner = nil
```
TeoTwawki: `local thing` is valid syntax for defining a local var as nil, but we don't even need it here as `target` is already a local var on the next line and not a function arg like killer was we can just redefine it as needed and use it instead of owner. `target = target:getMaster()` legit works and then the rest of the code can safely assume its not a pet and not nil.

scripts/mixins/families/aern.lua
```diff
@@ -20,21 +20,51 @@ g_mixins.families.aern = function(aernMob)
...
                 local target = mob:getTarget()
-                if target then killer = target end
```
TeoTwawki, line 27: I believe the likely intend was to use target if it existed and fallback to killer if it didn't, but the originator of this line overwrite killer with target rather than overwriting target with killer (backwards assignment). I am not sure what retail actually does if the target is died just before the aern performs its self-raise. If it does NOT go for another target, such a fallback isn't needed.

TeoTwawki: looking in the history before that mistake came in, it didn't fall back, just gave up if no target was found. there is a loop further down looking for additional targets.

from discord:
- the comment was more to highlight no we don't ever want to reassign killer
- we MIGHT want to reassign target

scripts/mixins/families/aern.lua
```diff
@@ -20,21 +20,51 @@ g_mixins.families.aern = function(aernMob)
...
-                    if mobArg:checkDistance(killer) < 40 then
-                        mobArg:updateClaim(killer)
-                        mobArg:updateEnmity(killer)
+                    if
+                        mobArg:checkDistance(target) < 40 and
+                        target:isAlive()
+                    then
+                        mobArg:updateClaim(target)
+                        mobArg:updateEnmity(target)
+                    elseif -- If pet has despawned then aggro owner
```
TeoTwawki: if we just overwrite target you won't need the elseif

scripts/mixins/families/aern.lua
The-Aerec: So I did this because it catches a specific case:

We don't automatically want to set the target as the owner of the pet in the case that the pet is still alive and well, Aerns should if possibly reaggro whoever the target was. 

I set up the owner nil scenario as a placeholder for the owner in case the target was dead and was a pet, and then and only then should we aggro the owner.

Maybe i'm mis-reading what you are saying.  

Are you saying on the owner ~= nil line we can:

target:getMaster() ~= nil 

instead? (I would get that if the target isn't a pet getMaster() returns nil?)

scripts/mixins/families/aern.lua
The-Aerec: It aggros other party members. IE the reason for my getAlliance() section now

scripts/mixins/families/aern.lua
TeoTwawki: > We don't automatically want to set the target as the owner of the pet in the case that the pet is still alive and well, Aerns should if possibly reaggro whoever the target was.

that's not what we are doing though. we don't need to know the owner beyond ensuring we switch target TO the owner if target was a pet. this happens straight away so that all code after is dealing with a player only and never a pet.
```lua
local target = mob:getTarget()
if not target then
     -- hold up target doesn't exist
end
if target:isPet() then
    -- Whoa targets a pet! we don't wnat that!
    target = target:getMaster()
end
if target:isDead() then
    -- Whoa targets dead! we need a new one
end
```
etc

src/map/entities/mobentity.cpp
TeoTwawki: just want to point out that mob drops and aern crashes normally wouldn't go in the same pr, and I'd have rubber stamped this part in and merged already if it were standalone - its now held up by reviewing aern fixes

scripts/mixins/families/aern.lua
The-Aerec: the whole: 
```lua
if target:isPet() then
    -- Whoa targets a pet! we don't wnat that!
    target = target:getMaster()
end
```

We do want a pet though, if the pet is alive and well, the aern SHOULD aggro the pet and not the master.

The-Aerec: Which is why I left the original stuff, the current script just cleans it up a bit.

If i'm a pet and not alive, then target should be my master.

If my master isn't alive then ill skip the first if and goto the 'find a valid target in my alliance' loop.

If all else fails, deaggro.

TeoTwawki: I was sure it went for the master on retail. that changes things and is a pain because they share a claim and functions that don't work on pets etc.

will come back to this in a bit have to afk

TeoTwawki: confirmed in game by WinterSolstice8 that Aerec is correct - its just difficult for the pet to have more hate than the master, but it goes for highest hate.

scripts/zones/The_Garden_of_RuHmet/mobs/Ixaern_DRK.lua
```diff
@@ -15,25 +15,46 @@ require("scripts/globals/status")
...
+                    target:isAlive()
+                then
+                    mobArg:updateClaim(target)
+                    mobArg:updateEnmity(target)
+                else
+                    local partySize = killer:getPartySize() -- Check for other available valid aggro targets
+                    local i = 1
+                    if killer ~= nil then
+                        for _, partyMember in pairs(killer:getAlliance()) do
+                            if partyMember:isAlive() and mobArg:checkDistance(partyMember) < 40 then
```
WinterSolstice8: Went and tested this on retail, a party member that never had enmity on the aern within 10~ units didn't get targeted and it didn't go after my SMN that killed it with a garuda melee attack. The SMN was over 25 units away.

Additionally, in a separate test I made sure my SMN had hate by flashing/meleeing down to low hp and letting Garuda kill the Aern with a melee attack. See log
[aern RR testing.txt](https://github.com/LandSandBoat/server/files/8886762/aern.RR.testing.txt)


tl;dr: change distance check to <25, and add a `-- TODO, change to check highest in enmity list`. There's no binding for the mob/player highest on the list _yet_

WinterSolstice8 (changes requested):

WinterSolstice8 (approved):

TeoTwawki (approved):

zach2good:
Squash on merge please

PR #1970, opened 2022-06-12
Pr/fix mother globe and slave globe behaviour mechanics to be more retail accurate

m241dan:
**_I affirm:_**

Makes the Slave Glove spawning mechanics and behaviour a little more retail. I will admit that this is not perfect because of the crappiness of "onMobRoam" timing  / mechanics. The slave globes tend to surge instead of follow closely, but it's mostly pretty good.

This is probably "best" until there is a "multiple pet" mechanic.

1. Start Server
2. Visit MG
3. Watch, observe, chuckle, maybe frown a little at the surging, but be mostly happy with this.


TeoTwawki:
> crappiness of "onMobRoam" timing / mechanics.

yeah somewhere in the future we need some new functionality. Some Goblins/Moblins in oldton have this behavior of follow-the-leader as well

scripts/zones/The_Shrine_of_RuAvitau/mobs/Mother_Globe.lua
```diff
@@ -4,40 +4,109 @@
...
+end
+
+local handleSlaveGlobesRoam = function(mg)
+    local mgPos = mg:getPos()
+
+    for _, slaveGlobeID in ipairs(ID.mob.SLAVE_GLOBES) do
+        local slaveGlobe = GetMobByID(slaveGlobeID)
+
+        if slaveGlobe:isSpawned() then
+            local slavePositionSlot = slaveGlobe:getLocalVar("slavePositionSlot")
```
m241dan: missed this in my refactor, should be using the calculateSlaveGlobePosition() method

scripts/zones/The_Shrine_of_RuAvitau/mobs/Mother_Globe.lua
```diff
@@ -4,40 +4,109 @@
...
 -----------------------------------
 local entity = {}
 
+local findSlaveGlobeToSpawn = function()
+    local spawnCount = 0
+
+    for _, slaveGlobeID in ipairs(ID.mob.SLAVE_GLOBES) do
+        local slaveGlobe = GetMobByID(slaveGlobeID)
+
+        spawnCount = spawnCount + 1
```
m241dan: This should just be the index since it's not --really-- doing what I wanted...

m241dan (commented):
I'll clean this up at some point.

scripts/globals/utils.lua
```diff
@@ -551,3 +551,23 @@ function utils.mobTeleport(mob, hideDuration, pos, disAnim, reapAnim)
...
 end
+
+local ffxiRotConversionFactor = 360.0 / 255.0
+
+function utils.ffxiRotToDegrees(ffxiRot)
+    return ffxiRotConversionFactor * ffxiRot
+end
+
+
+function utils.lateralTranslateWithOriginRotation(origin, translation)
```
zach2good, line 561: The dangerous thing about this is that it has no awareness of the navmesh, and if the new end-location is valid. We should probably add a binding to lua_zone to ask the navmesh "is the new x, y, z coord valid" for this sort of thing. (Or better yet, move this all down into pathfind/navmesh and make a binding to access it from Lua)

m241dan: Added a raycast check.

I don't think this is strictly what you wanted. But, I think it works. In terms of pulling this stuff fully in to the core, I'll leave that to someone who is more familiar with our overall navmesh / mob pathing situation.

zach2good: See my writeup here for why raycast doesn't solve this issue (and why its generally pretty awful):
https://github.com/LandSandBoat/server/issues/1470

You're better off making a binding for `zone:isValidPosition(x, y, z)` that hooks up to `navmesh::validPosition`

m241dan: Ah, hmmm okay. It didn't seem like that would actually give me what I wanted. It just seemed like it was valid within the zone, but didn't necessarily preclude something on the other side of a wall or "inside" a column?

m241dan: Okay, bound and switched to validPosition. Though, on the Lua side I named it to "isNavigablePoint" because I feel like it's more expressive (see commit msg).

Left the raycast stuff in there but will drop the commit if you want but I figure "more" bindings can't hurt us.

zach2good: I think I'd rather you drop the raycast binding, since it's a fairly well-known term that doesn't operate in the well-known way.

m241dan:
Done.

scripts/globals/utils.lua
m241dan: removed.

src/map/lua/lua_zone.cpp
```diff
@@ -187,6 +187,19 @@ void CLuaZone::reloadNavmesh()
...
+        point["x"].get_or<float>(0),
+        point["y"].get_or<float>(0),
+        point["z"].get_or<float>(0),
+        point["moving"].get_or<uint16>(0),
+        point["rot"].get_or<uint8>(0)
+    };
+
+    return m_pLuaZone->m_navMesh->validPosition(position);
+}
+
```
zach2good, line 202: 👍

scripts/zones/The_Shrine_of_RuAvitau/IDs.lua
```diff
@@ -35,6 +35,11 @@ zones[xi.zone.THE_SHRINE_OF_RUAVITAU] =
         ULLIKUMMI        = 17506418,
         OLLAS_OFFSET     = 17506667,
         KIRIN            = 17506670,
+        MOTHER_GLOBE     = 17506396,
+        SLAVE_GLOBES     =
+        {
+            17506397, 17506398, 17506399, 17506400, 17506401, 17506402
```
zach2good: You don't need to define more IDs here, when you can know for sure that:
```lua
local SLAVE_GLOBE_1 = ID.mobs.MOTHER_GLOBE + 1
local SLAVE_GLOBE_2 = ID.mobs.MOTHER_GLOBE + 2
local SLAVE_GLOBE_3 = ID.mobs.MOTHER_GLOBE + 3
local SLAVE_GLOBE_4 = ID.mobs.MOTHER_GLOBE + 4
local SLAVE_GLOBE_5 = ID.mobs.MOTHER_GLOBE + 5
local SLAVE_GLOBE_6 = ID.mobs.MOTHER_GLOBE + 6
```

So:
```lua
    for i = 1, 6 do
        local slaveGlobe = GetMobByID(ID.mob.MOTHER_GLOBE + i)
        -- ...
    end
```

zach2good (changes requested):

scripts/zones/The_Shrine_of_RuAvitau/IDs.lua
m241dan: I'd rather not rely on convention where possible, but I'll make some change.

m241dan: Resolved?

zach2good (approved):

Issue #1986, opened 2022-06-14
🐛 Add Spell Can crash the server


The-Aerec:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Branch affected by issue 

`base`

## Steps to reproduce

!addspell certain spells

Observer server crash

## Expected behavior

server shouldn't crash.

dmp attached. 

[xi_map.exe_13-6_18-2-38.log](https://github.com/LandSandBoat/server/files/8895747/xi_map.exe_13-6_18-2-38.log)

TeoTwawki:
got an example spell? I've seen this happen when using addspell on trust spells instead of addtrust way back in dsp, and I think I'd seen it when db info was missing but its been multiple years now.

zach2good:
From deep in those logs (I dug these out for you dude, you could have pasted them)

it's from:!addspell 14331

TeoTwawki:
Max id in the clients dat is faaaar below that value. This was super out of bounds (stops at 1023).

zach2good:
Needs an OOB check in charutills::addSpell

TeoTwawki:
> Needs an OOB check in charutills::addSpell

I _almost_ opted to do the check in the lua side instead before deciding to prevent people from making mistakes outside of the GM command too.

PR #1988, opened 2022-06-14
Limit ID range that can be passed into `addSpell()` to be 1-1023

TeoTwawki:
**_I affirm:_**
- [:crossed_fingers:] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

Limits spell IDs the core will try to set on a player to the legit range that exists in the client dat file.

Try and `!addspell` spell ID of 0 or 1024 or higher.

If anyone has a quick suggestion on how to check if the spell ID being passed also exists in our database, I'd like to add that as well - not all 1023 slots are used at present. I tried it and it didn't seem to crash me at least, but it did bug out my clients list display till I zoned. Wasn't super thorough with this.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -3201,7 +3201,8 @@ namespace charutils
 
     int32 addSpell(CCharEntity* PChar, uint16 SpellID)
     {
-        if (!hasSpell(PChar, SpellID))
+        // Todo: come up with a good way to validate that the SpellID exists in the database also.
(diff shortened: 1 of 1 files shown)


zach2good (approved):
SpellID is a uint, so will always be larger than 0

zach2good:
Wait, it's 5 am, I'm stupid, carry on

zach2good:
Should close https://github.com/LandSandBoat/server/issues/1986 (or be the first step to fixing it)

PR #2002, opened 2022-06-15
[Login] Add IP Limit Exceptions

The-Aerec:
Co-Authored-By: Damarus <leblanc.jlj@gmail.com>

**_I affirm:_**

Adds Exceptions to the IP Login Limit via dates: (Allows for Temporary time gated vs permanent)

Login on multi accounts.

### DIFF
--- a/sql/ip_exceptions.sql
+++ b/sql/ip_exceptions.sql
@@ -1,0 +1,8 @@
+DROP TABLE IF EXISTS `ip_exceptions`;
+CREATE TABLE IF NOT EXISTS `ip_exceptions` (
+    `accid` int(10) unsigned NOT NULL DEFAULT '0',
+    `exception` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
+    `comment` varchar(512) DEFAULT NULL,
+    PRIMARY KEY (`accid`)
+)
+ENGINE=InnoDB DEFAULT CHARSET=utf8;
--- a/src/login/lobby.cpp
+++ b/src/login/lobby.cpp
@@ -356,8 +356,22 @@ int32 lobbydata_parse(int32 fd)
                         sessionCount = (uint16)sql->GetIntData(0);
                     }
 
+                    fmtQuery = "SELECT UNIX_TIMESTAMP(exception) \
+                                FROM ip_exceptions \
+                                WHERE accid = %u;";
+
+                    int64 exceptionTime = 0;
+
+                    if (sql->Query(fmtQuery, sd->accid) != SQL_ERROR && sql->NumRows() != 0)
+                    {
+                        sql->NextRow();
+                        exceptionTime = sql->GetIntData(0);
+                    }
+
+                    int64_t timeStamp = std::chrono::duration_cast<std::chrono::seconds>(server_clock::now().time_since_epoch()).count();
                     bool isNotMaint   = maint_config.maint_mode == 0;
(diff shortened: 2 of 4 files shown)


The-Aerec:
This is a draft PR for collaboration, this was my idea on how to proceed and my 'best attempt' (I am not a cpp guru obviously, and i'm very much still learning.)

Server owner sets a date in exceptions: yyyymmddhhmmss format. (We could even change the sql data type to a char or varchar type and make the date format more readable and just substr the '-' and ':' out of: yyyy-mm-dd-hh:mm:ss)

We convert to epoch and check differences between epoch and exception date and if the epoch is after the exception date then well the user can't login if there is a session already on their IP.

Owner can set date years in the future for a more permanent exception, or an actual date for a temporary one.

Please rip this apart... I suck at cpp.

CatsEyeXI:
just stopping by to say i fully support this <3

zach2good:
> Server owner sets a date in exceptions: yyyymmddhhmmss format. (We could even change the sql data type to a char or varchar type and make the date format more readable and just substr the '-' and ':' out of: yyyy-mm-dd-hh:mm:ss)

Why not jus use a datetime or timestamp type in SQL?
```sql
  `exception_time` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
```

image

So this is an infinite exception, rather than "you can have this custom limit"?

If you're going to modify an existing table, which will have existing user data in it, you'll need to provide a migration (look in tools). It would be easier for you, and anyone else if you just create a whole new table, then it'll just appear as part of a regular update without needing a migration (I think?):

 ```
account_ip_exceptions.sql

id |  login_limit | exception_expires | notes
```

Then you can just query using the account number, or add it to nearby queries.

You can get away with not using whole time structures, and just use unix epoch:
https://database.guide/unix_timestamp-examples-mysql/#:~:text=In%20MySQL%2C%20you%20can%20use,or%20another%20specified%20date%2Ftime.

We use that here:
```cpp
int32 CCharEntity::GetTimeCreated()
{
    TracyZoneScoped;
    const char* fmtQuery = "SELECT UNIX_TIMESTAMP(timecreated) FROM chars WHERE charid = %u;";

    int32 ret = sql->Query(fmtQuery, id);

    if (ret != SQL_ERROR && sql->NumRows() != 0 && sql->NextRow() == SQL_SUCCESS)
    {
        return sql->GetIntData(0);
    }

    return 0;
}
```

Here is the modern cpp way of getting that same epoch to compare with:
```cpp
int64_t timestamp = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
```

https://godbolt.org/z/5PTjx5Ejc

zach2good:
Adding on to this, you WILL need a migration, but it'll be very simple.

Since you're probably going to add the exception list to the protected tables list.

if exceptions table doesn't exist, create it.

claywar:
> If you're going to modify an existing table, which will have existing user data in it, you'll need to provide a migration (look in tools). It would be easier for you, and anyone else if you just create a whole new table, then it'll just appear as part of a regular update without needing a migration (I think?):

Since it will need to be set up as a protected table, will still need a migration, since those aren't touched by the tool except for initial db setup

zach2good (approved):

zach2good:
Don't merge yet, mroe changes incoming

The-Aerec:
> Don't merge yet, mroe changes incoming

Ready to go: Changed id -> accid to be in line with other sql files when it comes to referring to an Account ID

zach2good (approved):

PR #2021, opened 2022-06-16
Update menu_merit, break out monipulator packets

claywar:
**_I affirm:_**

* Refactors menu_merit packet, and adds support for Blue Mage job points
* Breaks out monipulator packets 1 and 2 into two separate packet types

* Add Blue Mage job points for points, and observe increase
* All other function should remain

NOTE: Need to verify that we're sending the proper packets for various requests, as we may be over-sending

### DIFF
--- a/src/map/packets/monipulator2.cpp
+++ b/src/map/packets/monipulator2.cpp
@@ -1,0 +1,36 @@
+﻿/*
+===========================================================================
+
+  Copyright (c) 2022 LandSandBoat Dev Teams
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


src/map/packets/monipulator2.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+CMonipulatorPacket2::CMonipulatorPacket2(CCharEntity* PChar)
+{
+    this->setType(0x63);
+    this->setSize(0xB4);
+
+    memset(data + 4, 0, PACKET_SIZE - 4);
+
+    uint8 packet2[] = { 0x04, 0x00, 0xB0 };
+    memcpy(data + (0x04), &packet2, sizeof(packet2));
```
zach2good: This could be a C++ <array>:
```cpp
#include <array>
#include <iostream>

using uint8  = std::uint8_t;

int main()
{
    uint8 packet2_old[] = { 0x04, 0x00, 0xB0 };
    std::cout << packet2_old << " " << sizeof(packet2_old) << "\n"; // 3

    std::array<uint8, 3> packet2 = { 0x04, 0x00, 0xB0 };
    std::cout << packet2.data() << " " << packet2.size() << "\n";

    return 0;
}
```

zach2good (approved):

PR #2024, opened 2022-06-16
Dbtool Updates

cocosolos:
**_I affirm:_**

A few changes to dbtool:
- Removes the ability to edit protected tables in config since this config setting no longer exists.
- No longer reads DB_VER from version.conf and checks the dbtool config instead. (in anticipation of #1945)
- Checks for missing protected tables and imports them if they're missing.

On first use, express update will not be available until db_ver is moved from /conf/version.conf into /tools/config.yaml. Dbtool will do this for you the first time you perform a regular update, or if you manually copy the hash from version.conf into config.yaml.
It will look like this in version.conf: `#DB_VER: abc123`
And should look like this in config.yaml: `- db_ver: abc123`

**This should work as expected but I'm marking it as a WIP for now until I can test the changes a little more thoroughly.**

- Use dbtool to update. No protected tables should be selected for import if none are missing. Drop one or more protected tables and attempt another update. Dbtool should detect this and offer to import missing tables even if they're marked as protected player data. Test the same things using the CLI options. At no point should dbtool ever attempt to import an existing protected table.
- Confirm db_ver is correctly written, read, and used for express updates.

### DIFF
--- a/tools/dbtool.py
+++ b/tools/dbtool.py
@@ -91,6 +91,7 @@ player_data = [
 ]
 
 import_files = []
+import_protected = []
 backups = []
 database = None
 host = None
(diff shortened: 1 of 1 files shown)


zach2good:
I've pointed this PR at my Settings/Lua refactor PR, so I can deploy the two of these together, whatever is left to do I'll handle it 👍

zach2good (approved):

PR #2025, opened 2022-06-16
[Mission] CoP3-3 The Road Forks - Balance Fixes

Frankie-hz:
**_I affirm:_**

- Fixed an issue where anitlions that used ambush would be hidden, untargetable and stuck underground.
- Fixed an issue where when disengaged did not return to its original spawn point
- Lioumere and Overgrown Ivy have been scripted towards retail accuracy.

**Lioumere  References**
- https://www.youtube.com/watch?v=T-hXDvNXZ-U&ab_channel=ScryerFinalFantasyXI
- https://www.youtube.com/watch?v=OHbPqIzO6yE&t=2168s&ab_channel=heorotlinea
- https://ffxiclopedia.fandom.com/wiki/Lioumere

**Overgrown Ivy References**
- https://youtu.be/7Jrnhdg7PFs?t=836
- https://ffxiclopedia.fandom.com/wiki/Overgrown_Ivy

!addmission 6 325

For Overgrown Ivy:
!spawnmob 16785709
!gotoid 16785709
Fight it.

For Lioumere: Follow the windy mission in this order.
Ohbiru-Dohbiru: !pos 23 -5 -193 238
Yoran-Oran: !pos -109.987 -14 203.338 239
Kyume-Romeh: !pos -58 -4 23 238
Honoi-Gomoi: !pos -195 -11 -120 238
Loose Sand: !pos 478.8 20 41.7 7

Once at the Loose Sand click on it and Lioumere will perform the ambush mixin and be visible.
Fight it


scripts/zones/Attohwa_Chasm/mobs/Lioumere.lua
```diff
@@ -8,6 +8,30 @@ require("scripts/globals/missions")
 -----------------------------------
 local entity = {}
 
+entity.onMobWeaponSkill = function(player, mob, skill)
+    --Sets entire party enmity CE/VE to 0.
+    local alliance = player:getAlliance() -- Gets member pointers for alliance
+    for _, member in pairs(alliance) do
+        mob:resetEnmity(member) -- Resets player enmity on the mob
+    end
```
Xaver-DaRed, line 20: Shouldn't this check for player pets aswell?

sql/mob_pools.sql
```diff
@@ -3124,7 +3124,7 @@ INSERT INTO `mob_pools` VALUES (3068,'Osschaarts_Wyvern','Osschaarts_Wyvern',193
 INSERT INTO `mob_pools` VALUES (3069,'Oupire','Oupire',284,0x0000350700000000000000000000000000000000,4,5,7,220,100,0,1,1,0,2,7,0,0,1157,8,0,150,0,0,284,284);
 INSERT INTO `mob_pools` VALUES (3070,'Ouryu','Ouryu',259,0x0000620200000000000000000000000000000000,7,4,7,240,125,1040,1,1,0,2,0,32,297,1179,0,0,109,0,0,259,259);
 INSERT INTO `mob_pools` VALUES (3071,'Ouzelum','Ouzelum',125,0x0000530100000000000000000000000000000000,4,5,12,240,100,0,0,0,0,2,0,0,1,1179,0,0,266,0,0,125,125);
-INSERT INTO `mob_pools` VALUES (3072,'Overgrown_Ivy','Overgrown_Ivy',186,0x00007C0100000000000000000000000000000000,2,2,7,360,100,0,1,0,0,0,0,0,1261,1155,8,0,0,0,0,186,186);
+INSERT INTO `mob_pools` VALUES (3072,'Overgrown_Ivy','Overgrown_Ivy',186,0x00007C0100000000000000000000000000000000,2,2,7,100,100,0,1,0,0,0,0,0,1261,1155,8,0,0,0,0,186,186);
```
TeoTwawki, line 3127: Do we have a source for that 100 delay\*? Wiki old wiki says fast and compares to hundred fists, and that's mostly likely bull. They've been wrong just about everytime they say that (usually some kind of multihit involved). Bg doesn't mention it at all. JP Wiki also has nothing. I see it is listed as mnk mob on old wiki and a warrior on bg tho. 

I have a hunch it just has a high rate of double attacks as war/war or is war/mnk or mnk/war.

_*as in for the number, I know you cited sources, but nothing says the speed. Felt I should clarify_

TeoTwawki: Err on side of caution of you want to get the pr in and do further adjustments later.

TeoTwawki: ps. I **did** watch the video. it didn't look like the swings were constant to me. I can't measure by eyeballing it.

Frankie-hz: I honestly don't have a straight answer for you other than watching like 4-5 videos and trying to get it to  match the best I can. 100 Delay seems to fit with the constant auto attacks with a 10% double attack proc. Any higher percentage, even with a slower delay just didn't feel right. I also found all the same thing you did so I can't say for certain this is 100% accurate but it's the closest I can get it. If someone else to play around with it I will gladly change it.

The inconsistent swings must be a double attack percentage bc in nowhere of the few videos ive watched did I ever see a triple attack. In the cases it may have looked like a triple attack I had a few other ppl also look at it and they agreed it seemed to be the animation that happens when it doubles+autos right after the 2nd hit, which also led be to believe it must be a constant low delay.

Frankie-hz: And it does not look like a multi-hit either otherwise this thing would be going completely crazy. It also doesnst match the attacks in any video I found

scripts/zones/Attohwa_Chasm/mobs/Lioumere.lua
Frankie-hz: Yeah you are right. I'll fool around with it tomorrow

sql/mob_pools.sql
TeoTwawki: There were a few pauses when I watched that made me think it might be just a slow conistant2-hit occasionally being 3, or it might have just been a bit of frame stutter with non stop single swings. Def not a stream of high speed triple attack like I've seen with most mob wiki says "fast like hundred fists".

Didn't really look fast enough to be 100 to my own eyes, maybe 150. But Wouldn't be able to say without some kind of reliable measurement. I think we want to err on the side of making the mob harder than should be rather than being to easy so its probably going to merge as is unless I get some better data but wanted to check in case you did something to figure out it was 100 or had some reliable source for the number. 

Winter was playing around some math to see if there was a way we could figure it out better but we couldn't get anything concrete without popping it ourselves yet.

TeoTwawki: Can you tag a todo comment here for me? I'll help me revisit in the future
`-- Todo: verify melee delay`

Frankie-hz: > Winter was playing around some math to see if there was a way we could figure it out better but we couldn't get anything concrete without popping it ourselves yet.

Yeah I would do the capture myself but I did CoP on my retail char a long time ago. Sorry :( had to rely on what information I had.

TeoTwawki: Its ok man, I just always gotta check where the numbers are obtained so I know, ya know? :)

Frankie-hz: Of course :D

scripts/mixins/families/antlion_ambush.lua
```diff
@@ -15,12 +15,21 @@ g_mixins.families.antlion_ambush = function(antlionAmbushMob)
...
+        end)
     end)
     antlionAmbushMob:addListener("DISENGAGE", "ANTLION_AMBUSH_DISENGAGE", function(mob)
         mob:hideName(true)
         mob:setUntargetable(true)
         mob:setAnimationSub(0)
         mob:wait(2000)
+        mob:timer(2000, function(ant)
+            local spawn = mob:getSpawnPos()
+            mob:setPos(spawn.x, spawn.y, spawn.z, spawn.rot)
```
TeoTwawki: This doesn't seem to happen on retail.

It appears that when they go back underground, they stay right there for awhile. if they were to far from "home" (pop area) and are non NM they can despawn instead (sometimes before they go underground). 

NMs (or at least the one I checked) appear to "walk home" _while underground_ o.O which seams like such a pain in the butt.

scripts/mixins/families/antlion_ambush.lua
```diff
@@ -15,12 +15,21 @@ g_mixins.families.antlion_ambush = function(antlionAmbushMob)
     antlionAmbushMob:addListener("ENGAGE", "ANTLION_AMBUSH_ENGAGE", function(mob, target)
         mob:useMobAbility(278) -- Pit Ambush
         mob:setMobMod(xi.mobMod.NO_MOVE, 0)
+        mob:timer(2000, function(ant)
+            ant:hideName(false)
```
TeoTwawki: this stuff already happens in the mobskill itself, and takes care of the timing aspect so the timer call isn't needed - you get problems when you launch these things before the skill animation finishes

https://github.com/LandSandBoat/server/blob/d9d3a7b6494d17f2178b140c9b0f6d9e983d7459/scripts/globals/mobskills/pit_ambush.lua#L32

scripts/zones/Carpenters_Landing/mobs/Overgrown_Ivy.lua
```diff
@@ -4,6 +4,23 @@
...
+
+entity.onMobFight = function(mob, target)
+    if mob:getHPP() <= 15 then
+        mob:setMod(xi.mod.REGAIN, 200)
+    end
+end
+
+entity.onMobWeaponSkillPrepare = function(mob, target)
+    if mob:getHPP() <= 15 then
+        return 319
```
TeoTwawki: comment what we're returning, nobody will know what 319 is at a glance

sql/mob_pools.sql
TeoTwawki: by todo comment, I meant here in file as `-- Todo: verify melee delay` at the end of the line.

PR #2026, opened 2022-06-16
Module to Remove Guild Kits

RunAwayDSP:
**_I affirm:_**

Could do other things with this module as well but this is basic just remove the crafting kits from the guild stock.

Load module and see that there are no more kits on the guild shops 🥳 
Before
img_20220615_232300
After
img_20220615_232306


modules/custom/lua/guild_no_kits.lua
```diff
@@ -0,0 +1,205 @@
+-----------------------------------
+-- Guilds: Remove Kits from Guild Shops
+-----------------------------------
+require("modules/module_utils")
+require("scripts/globals/shop")
+-----------------------------------
+local m = Module:new("guild_no_kits")
+
+m:addOverride("xi.shop.generalGuild", function(player, stock, guildSkillId)
```
zach2good, line 9: Teo mentioned something very smart to me:

You can just modify `stock` here:

```lua
local itemsToRemove =
{
 ...
}

local function modify(stock)
    -- Find and remove entries from stock, looked up from itemsToRemove
    -- (and also remove the next two entries, since it's a linear table
end

m:addOverride("xi.shop.generalGuild", function(player, stock, guildSkillId)
    -- do your modification
    super(player, modify(stock), guildSkillId)
end
```

That way the tables will survive changes upstream, and you don't have to re-implement everything

zach2good (changes requested):

modules/custom/lua/guild_no_kits.lua
RunAwayDSP: Wtf I searched for something like that too well that makes things way easier!

PR #2036, opened 2022-06-17
[Lua] inherit sneak/invis multiplier from global settings for tonko & monomi

CatsEyeXI:
**_I affirm:_**

Factor in global multiplier for ninjutsu sneak/invis

Cast Monomi/Tonko before and after change.


TeoTwawki:
iirc, didn't somebody remove it from the white magic spell? should prolly add it back if we're doing it for ninjutsu.

CatsEyeXI:
> iirc, didn't somebody remove it from the white magic spell? should prolly add it back if we're doing it for ninjutsu.

I'll check these as well as spectral jig then

Xaver-DaRed:
Honestly, I think those multipliers should be removed.
You can acomplish the same at this point by editing the parameter table. It gives as much or even more control over what spells do

TeoTwawki:
> Honestly, I think those multipliers should be removed. You can acomplish the same at this point by editing the parameter table. It gives as much or even more control over what spells do

fair point about the params. also the durations were much shorter back when the setting was made

PR #1878, opened 2022-06-05
[core] [lua] [RUN] Implement Liement, add MDT/DT/Phalanx/Stoneskin checks where appropriate, general Absorb mechanic improvements

WinterSolstice8:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [X] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [X] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

Implement liement, 

* add MDT/DT/Phalanx/Stoneskin checks where appropriate
* Adjust spikes damage to allow for absorbing
* Adjust spikes calls to use some actual message IDs instead of hardcodes
* Add core handler to check for Liement absorb
* Change some RUN effect removal to be silent
* Add range to Liement, which is not normally used unless you have Epeolatry's mod by design
* Update SteamJacket handler to pass in the element instead of hardcoded "ELEMENTAL" damage
* Adjust spell_damage.lua to have MDT + MDTII cap

## Steps to test these changes

Liement elemental absorb:
1. Fight fire elemental with blaze spikes, enfire
2. Use unda rune, use liement to absorb magic damage from nuke, enspell, spike damage. Each time Liement absorbs it should remove the buff from you. 

Liement AoE:
1. get a party
2. use Liement, it should be only activating on you
3. `!setmod LIEMENT_EXTENDS_TO_AREA 1` on yourself
4. use Liement, it should be AoE now

Dread spikes interaction:
1. find a mob with MP
2. force it to cast dread spikes `!exec target:castSpell(277, target)`
3. use Lux, liement
4. Hit mob
5. You should absorb your melee hit as HP, target will not absorb damage, see below:

image

Magic weaponskill absorb:
1. equip staff
2. `!exec target:addStatusEffect(xi.effect.LIEMENT, 25, 0, 999, 0, xi.damageType.EARTH)` on target
3. observe absorb when using earth crusher

Hybrid weaponskill absorb:
1. equip great katana on SAM
2. use Tachi: Jinpu on target a couple times
3. `!exec target:addStatusEffect(xi.effect.LIEMENT, 25, 0, 999, 0, xi.damageType.WIND)` on target
4. use Tachi: Jinpu on target, repeat step 3 to get a few samples
5. Observe consistently lowered damage due to absorb

Blue Magic absorb:
1. change job to BLU
[...]

PR #2050, opened 2022-06-19
[core] change spikes other than dread spikes to do damage again

m241dan:
**_I affirm:_**

Fixes it so spikes other than dread spikes does damage again.

1. Go to MG? Or something with spikes as BLM.
2. Cast Shock spikes on yourself.
3. Fight, notice your spikes actually do damage and MG's spikes do damage to you.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -675,7 +675,7 @@ namespace battleutils
 
         ELEMENT spikeElement = (ELEMENT)((uint8)GetSpikesDamageType(Action->spikesEffect) - (uint8)DAMAGE_TYPE::ELEMENTAL);
 
-        int32 damage = 0;
+        int32 damage = Action->spikesParam;
 
         // int16 intStat = PDefender->INT();
         // int16 mattStat = PDefender->getMod(Mod::MATT);


src/map/utils/battleutils.cpp
```diff
@@ -675,7 +675,7 @@ namespace battleutils
 
         ELEMENT spikeElement = (ELEMENT)((uint8)GetSpikesDamageType(Action->spikesEffect) - (uint8)DAMAGE_TYPE::ELEMENTAL);
 
-        int32 damage = 0;
+        int32 damage = Action->spikesParam;
```
TeoTwawki, line 678: seems this was zero'd in https://github.com/LandSandBoat/server/pull/1878

unsure if by accident or was meant to have an assignment further down that was lost in a rebase or forgotten

TeoTwawki: unless I hear otherwise: putting param back looks good to me

WinterSolstice8:
I think it was an accident. Looking at it my brain tells me I changed it to 0 for a reason, I think it might have had to do with the original layout of the code before I implemented Liement, but glancing around it definitely looks wrong to set it to zero.

That PR was huge for just one feature, I probably just forgot to check normal spikes after making Dread Spikes Liement absorb work.

WinterSolstice8 (approved):

TeoTwawki (approved):

PR #2057, opened 2022-06-19
Remove setHP(0) from Mijin when mob:isInDynamis()

0phie:
**_I affirm:_**

+Removes the global mob:setHP(0) which kills mobs in dynamis upon using Mijin Gakure.

This was tested in multiple zones for Dynamis in retail. Tested all cities, tested Beaucedine and could not find the behavior of killing the mob once it used Mijin Gakure in either normal mobs or NMs within these zones. Thus a global mob:setHP(0) should not be used to potentially fix an issue with a handful of instances which I was unable to prove existed. Without documentation of what these instances were and the lack of additional data located on BG-Wiki, Ffxiclopedia, or other major resources for finding information on the game, I was unable to find any indication that this should happen in current retail. 

If such an instance does crop up, it is suggested to add an additional function to the bg-code for the job_special mixin in order to take care of this rather than have a global solution which is not accurate to every mob. This bg-code should be taken care of in the mob's specific lua as an additional job_special param. Additional methods of adding this in specifically would be to have gates in place for specific mob parameters such as families or IDs in the future, however at this point with the lack of information this global change should be removed. 

+ Bring a normal NIN mob in Dynamis (Vanguard Backstabber or the like) to < 60% and the mob will use Mijin. Prior to the change the mob would immediately have its HP set to 0, now the mob continues living.


TeoTwawki (approved):
holdover from a much older ver of the game possibly, but I don't remember

claywar (approved):

PR #2064, opened 2022-06-21
Rename elemental magic evasion mods to be clearer about what they do

TeoTwawki:
**_I affirm:_**

Renames elemental magic evasion mods to be clearer about what they do.
`FIRE_RES` -> `FIRE_MEVA` etc.
and also adds some very minor prep for future resist rank mods in the sql tables and comments near many of the queries.

Import table, build boot, cast something and note **nothing is different** from before pulling this because only the names changed.


TeoTwawki:
The tabs came in on a commit that wasn't mine but I fixed'em

claywar (approved):

zach2good (approved):

PR #2068, opened 2022-06-21
Avatars Favor Initial Implementation

The-Aerec:
Co-Authored-By: Tiberon <TiberonKalkaz@gmail.com>

**_I affirm:_**

Sources:
https://www.bluegartr.com/threads/112708-Avatar-s-Favor-Testing?p=5400057&viewfull=1#post5400057
https://www.bluegartr.com/threads/112708-Avatar-s-Favor-Testing?p=5400037&viewfull=1#post5400037
https://ffxiclopedia.fandom.com/wiki/Avatar%27s_Favor
https://www.bg-wiki.com/ffxi/Avatar%27s_Favor

I've testing things as is, and all the effects work as is.

I have converted the initial wings implementation to the retail implementation via checks in the settings and increasing Favor power as the smn uses blood pacts.

Implements the Ability: Avatar's Favor & the associated affects.

image
image
image
image
image
image
image
image
image

Summon Avatars -> Use Avatars Favor


The-Aerec:
Need to rework those two CPP functions into perhaps a sql flag for the mobskill? Add a new column to determine special types of mobskills? Then query for that and check instead of those two functions.

The-Aerec:
Waiting for feedback, after talking to Teo today i've cleaned up the cpp and finilized the retail state of things:

Pre-SOA: Avatars have a debuff under favor 
Avatars Favor power increased after using BPs

The-Aerec:
Updated and verified with new settings stuff.

scripts/globals/avatars_favor.lua
```diff
@@ -0,0 +1,149 @@
...
+    then
+        local pet = target:getPet()
+        pet:addMod(xi.mod.MATT, 20)
+        pet:addMod(xi.mod.ATTP, 20)
+        pet:addMod(xi.mod.ACC, 10)
+        pet:addMod(xi.mod.DEFP, 10)
+    end
+end
+
+xi.avatarsFavor.applyAvatarsFavorAuraToPet = function(target, effect)
```
zach2good, line 117: I feel like this logic should live inside TargetFind and not have to be applied everywhere. Aura's should "just work" without having to edit a bunch of different files.

And if the favors are anything like GEO auras, they are just effects with a 3-second tick that will just fall off when the giver is too far away/gone.

scripts/globals/avatars_favor.lua
```diff
@@ -0,0 +1,149 @@
...
+require("scripts/globals/status")
+require("settings/main")
+-----------------------------------
+
+xi = xi or {}
+xi.avatarsFavor = xi.avatarsFavor or {}
+
+xi.avatarsFavor.skill186Index = 6 -- 187 taken from wiki talk page
+xi.avatarsFavor.skillDefaultIndex = 9
+xi.avatarsFavor.skill317Index = 10
```
zach2good: This is voodoo, if we _need_ this, it's going to need a lot more comments to explain what it is, what it does, why it's this way, and why we can't use something more user friendly

scripts/globals/avatars_favor.lua
```diff
@@ -0,0 +1,149 @@
...
+
+-- Avatar Favor buffs scale per tock (~9 seconds) to a max value based on current summoning skill
+-- Index 1 is the initial value
+-- Index 6 is the value for the low skill tier
+-- Index 9 is the value for ~level 75, up to 317 skill
+-- Index 10 is the value for having 318 or more skill
+-- This can be extended for servers who are higher than 75 cap by adding additional values to the "arrays" and new skill tier logic (and Siren)
```
zach2good: 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀

src/map/ai/states/mobskill_state.cpp
```diff
@@ -131,3 +144,9 @@ void CMobSkillState::Cleanup(time_point tick)
         m_PEntity->loc.zone->PushPacket(m_PEntity, CHAR_INRANGE, new CActionPacket(action));
     }
 }
+
+// TODO move this to lua/sql
```
zach2good: Yeah, for sure we can add to/use/reuse `mob_skill_flag` for this in mob_skills.sql

The-Aerec: If someone would be willing to show me the ropes about this, I can definitely do it.

scripts/globals/avatars_favor.lua
The-Aerec: Correct its pretty much the exact same implementation:

If player has ability active then:

when an avatar is summoned it should apply its appropriate effect.
when its dismissed it should automatically fall off.
If someone moves too far from the avatar it should also fall off.

src/map/ai/states/mobskill_state.cpp
TeoTwawki: using c++ enums isn't much different form how you make and use modifiers in lua. you set a value in a list with a name, and then you can check against that named value elsewhere. Numerous examples exist throughout the codebase that can be easily searched out.

https://github.com/LandSandBoat/server/blob/2dbae3a1d7e3de544670f35051620660d5f9205f/src/map/mobskill.h#L28

https://github.com/LandSandBoat/server/blob/70680538d6c2a6fec0e960cd5407fbc0df142d5c/src/map/entities/mobentity.cpp#L634

`==` will check that the mask is exactly that value, `&` will check that the specific bit in the mask is on or off (it's a [bitwise operator](https://www.geeksforgeeks.org/bitwise-operators-in-c-cpp/))

incidentally, the 2hr flag was retired ages ago and its last remnants are in 2 spots. it could be removed and replaced.

https://github.com/LandSandBoat/server/blob/2dbae3a1d7e3de544670f35051620660d5f9205f/src/map/mobskill.cpp#L66

https://github.com/LandSandBoat/server/blob/70680538d6c2a6fec0e960cd5407fbc0df142d5c/src/map/ai/controllers/mob_controller.cpp#L345

TeoTwawki: Just add yourself one with the value of `0x40` for now

The-Aerec: working on this now

src/map/ai/states/mobskill_state.cpp
```diff
@@ -109,6 +109,19 @@ bool CMobSkillState::Update(time_point tick)
             static_cast<CMobEntity*>(PTarget)->PEnmityContainer->UpdateEnmity(m_PEntity, 0, 0);
         }
         m_PEntity->PAI->EventHandler.triggerListener("WEAPONSKILL_STATE_EXIT", CLuaBaseEntity(m_PEntity), m_PSkill->getID());
+
+        if (m_PEntity->objtype == TYPE_PET && m_PEntity->PMaster && m_PEntity->PMaster->objtype == TYPE_PC && (m_PSkill->isBloodPactRage() || m_PSkill->isBloodPactWard()))
+        {
+            CCharEntity* PSummoner = (CCharEntity*)(m_PEntity->PMaster);
```
WinterSolstice8: Can you change this to

`CCharEntity* PSummoner = dynamic_cast<CCharEntity*>(m_PEntity->PMaster);`
It's much, much safer than a direct cast like that. It does voodoo to verify things on the stack to verify the class is infact that class, rather than "forcing" it into that class whether it is that or not.

WinterSolstice8 (changes requested):

scripts/globals/effects/ramuhs_favor.lua
```diff
@@ -1,15 +1,20 @@
...
 
 effect_object.onEffectGain = function(target, effect)
+    target:addMod(xi.mod.CRITHITRATE, effect:getPower())
 end
 
 effect_object.onEffectTick = function(target, effect)
 end
 
 effect_object.onEffectLose = function(target, effect)
+    target:addMod(xi.mod.CRITHITRATE, effect:getPower())
```
WinterSolstice8: delMod

scripts/globals/effects/leviathans_favor.lua
```diff
@@ -1,15 +1,20 @@
...
 
 effect_object.onEffectGain = function(target, effect)
+    target:addMod(xi.mod.MACC, effect:getPower())
 end
 
 effect_object.onEffectTick = function(target, effect)
 end
 
 effect_object.onEffectLose = function(target, effect)
+    target:addMod(xi.mod.MACC, effect:getPower())
```
WinterSolstice8: delMod

scripts/globals/effects/ifrits_favor.lua
```diff
@@ -1,15 +1,20 @@
...
 
 effect_object.onEffectGain = function(target, effect)
+    target:addMod(xi.mod.DOUBLE_ATTACK, effect:getPower())
 end
 
 effect_object.onEffectTick = function(target, effect)
 end
 
 effect_object.onEffectLose = function(target, effect)
+    target:addMod(xi.mod.DOUBLE_ATTACK, effect:getPower())
```
WinterSolstice8: delMod

scripts/globals/effects/garudas_favor.lua
```diff
@@ -1,15 +1,20 @@
...
 
 effect_object.onEffectGain = function(target, effect)
+    target:addMod(xi.mod.EVA, effect:getPower())
 end
 
 effect_object.onEffectTick = function(target, effect)
 end
 
 effect_object.onEffectLose = function(target, effect)
+    target:addMod(xi.mod.EVA, effect:getPower())
```
WinterSolstice8: delMod

scripts/globals/effects/shivas_favor.lua
```diff
@@ -1,15 +1,20 @@
...
 
 effect_object.onEffectGain = function(target, effect)
+    target:addMod(xi.mod.MATT, effect:getPower())
 end
 
 effect_object.onEffectTick = function(target, effect)
 end
 
 effect_object.onEffectLose = function(target, effect)
+    target:addMod(xi.mod.MATT, effect:getPower())
```
WinterSolstice8: delMod

scripts/globals/effects/titans_favor.lua
```diff
@@ -1,15 +1,20 @@
...
 
 effect_object.onEffectGain = function(target, effect)
+    target:addMod(xi.mod.DEF, effect:getPower())
 end
 
 effect_object.onEffectTick = function(target, effect)
 end
 
 effect_object.onEffectLose = function(target, effect)
+    target:addMod(xi.mod.DEF, effect:getPower())
```
WinterSolstice8: delMod

scripts/globals/spells/summoning/carbuncle.lua
```diff
@@ -21,6 +22,14 @@ end
 
 spell_object.onSpellCast = function(caster, target, spell)
     xi.pet.spawnPet(caster, xi.pet.id.CARBUNCLE)
+
+    if caster:hasStatusEffect(xi.effect.AVATARS_FAVOR) then
+        local effect = caster:getStatusEffect(xi.effect.AVATARS_FAVOR)
+        effect:setPower(1) -- resummon resets effect
```
WinterSolstice8, line 28: Does this if condition ever trip? Does setting power mean you could infinitely gain favor buffs since the original power is not deleted? This comment applies to all avatars. It would probably be safer to just delete avatar's favor on summon instead of hack around it. Going to go check on retail how this works because I'm genuinely curious

src/map/status_effect_container.cpp
```diff
@@ -1905,6 +1905,12 @@ void CStatusEffectContainer::TickRegen(time_point tick)
                         CPetEntity* PPet  = (CPetEntity*)m_POwner->PPet;
                         CItem*      hands = PChar->getEquip(SLOT_HANDS);
 
+                        if (PChar->StatusEffectContainer->HasStatusEffect(EFFECT_AVATARS_FAVOR) &&
+                            PPet->m_PetID >= PETID_CARBUNCLE && PPet->m_PetID <= PETID_CAIT_SITH)
+                        {
+                            perpetuation -= 2;
```
WinterSolstice8: Perpetuation costs _more_ with favor up. By 20%,

https://www.bg-wiki.com/ffxi/Avatar%27s_Favor

`perpetuation = static_cast<int16>(perpetuation * 1.2);`

WinterSolstice8 (changes requested):

scripts/globals/spells/summoning/carbuncle.lua
WinterSolstice8: Ah I see how it works. There's no problem here. Just staring at the code didn't get me the whole picture. Avatar's favor is the buff that activates the actual favor aura on your avatar, and you're just using it to track the tick strength for the favor.

You're good!

scripts/globals/avatars_favor.lua
```diff
@@ -0,0 +1,149 @@
...
+    [1] = 316,
+    [2] = 381,
+    [3] = 446,
+    [4] = 511,
+    [5] = 574,
+    [6] = 669,
+    [7] = 670,
+}
+
+-- Based on BGWiki Power 8 -> 10 is only available with +Avatars Favor Gear
```
WinterSolstice8: https://www.bg-wiki.com/ffxi/Avatar%27s_Favor

bg wiki shows 11 tiers. Why are there only 10 here?

scripts/globals/avatars_favor.lua
```diff
@@ -0,0 +1,149 @@
...
+        effect = xi.effect.RAMUHS_FAVOR
+    },
+    [xi.pet.id.DIABOLOS] = -- Refresh
+    {
+        scaling = {3, 4, 4, 5, 5, 5, 6, 7, 7, 8},
+        effect = xi.effect.DIABOLOSS_FAVOR
+    },
+    [xi.pet.id.CAIT_SITH] = -- Magic Defense
+    {
+        scaling = {1, 2, 3, 4, 5, 6, 8, 10, 12, 15},
```
WinterSolstice8: This scaling doesn't match what is known on bg wiki for the last 5 tiers. where did this come from? the known pattern seems to be similar to carbuncles so i'd follow that on a guess, perhaps..

`10, 12, 14, 16, 18, 19, 22, 24, 26, 27, 28`

I don't like guessing, but we at least have something to start with. It will likely be approximate.

WinterSolstice8 (changes requested):

scripts/globals/avatars_favor.lua
The-Aerec: This was just: We don't have a cait sith to use for testing so I didn't update the #s. I've added correct numbers and siren plumbing for future implementations.

scripts/globals/avatars_favor.lua
The-Aerec: final tier was unlocked via job point gifts. And I forgot to add it!

WinterSolstice8 (approved):

PR #2076, opened 2022-06-22
[GEO] Ironing out details for Geomantic Reservoir implementations

CatsEyeXI:
**_I affirm:_**

I realize that this will probably need to be revised in some way, but I wanted to try to get the ball rolling on these as they do not seem to be implemented fully. From what I can gather, we can probably recycle the same event/params used here on all of the other reservoirs as long as we have granular control over the final "unlock spell" text.

image

!changejob GEO 5
!pos 378.875 -39.304 58.313
!addspell 769
Click on the GR.


scripts/zones/East_Ronfaure/npcs/Geomantic_Reservoir.lua
```diff
@@ -0,0 +1,31 @@
...
+-----------------------------------
+local ID = require("scripts/zones/East_Ronfaure/IDs")
+local entity = {}
+
+entity.onTrade = function(player, npc, trade)
+end
+
+entity.onTrigger = function(player, npc)
+    if player:hasSpell(769) then -- Player has Indi-Poison
+        player:startEvent(15000, 277, 3, 380619, -39102, 53620, 163, 539943, 0)
```
zach2good: Most of these args are probably junk and can be omitted, you could try `15000`, `15000, 277` or `15000, 277, 3` and see if it still appears correctly.

scripts/zones/East_Ronfaure/npcs/Geomantic_Reservoir.lua
```diff
@@ -0,0 +1,31 @@
...
+-- !pos 378.875 -39.304 58.313
+-----------------------------------
+local ID = require("scripts/zones/East_Ronfaure/IDs")
+local entity = {}
+
+entity.onTrade = function(player, npc, trade)
+end
+
+entity.onTrigger = function(player, npc)
+    if player:hasSpell(769) then -- Player has Indi-Poison
```
zach2good: Please use our enums instead of naked magic numbers:
https://github.com/LandSandBoat/server/blob/base/scripts/globals/spell_data.lua#L1054

Example here:
https://github.com/LandSandBoat/server/blob/base/scripts/quests/hiddenQuests/Trust_Ingrid.lua#L18

scripts/zones/East_Ronfaure/npcs/Geomantic_Reservoir.lua
```diff
@@ -0,0 +1,31 @@
...
+        player:startEvent(15000, 277, 3, 380619, -39102, 53620, 163, 539943, 0)
+    else
+        player:PrintToPlayer("You are assaulted by an uncanny sensation.", 0xD)
```
zach2good: We don't use `PrintToPlayer` for any messages sent to the client, since they're all baked into the client.
See here:
https://github.com/LandSandBoat/server/blob/base/scripts/quests/hiddenQuests/Trust_Ingrid.lua#L41
`player:messageSpecial(easternAdoulinID.text.YOU_LEARNED_TRUST, ...`

You can use POLUtils to scrape your installed client and it'll show you all of the text IDs for each zone:
https://github.com/Windower/POLUtils

These scrapes ARE A YEAR OLD DO NOT USE! - but they show the output of POLUtils, and the exact text you need to use for our update tooling to continue to work:
https://github.com/zach2good/UpdateExtractor/blob/versions/2021-july/out/scripts/zones/East_Ronfaure/Text.lua.raw

Like I say: THESE ARE OUT OF DATE DO NOT USE - but this is the text in this area:
```
11958 : % learns %!
11960 : You are assaulted by an uncanny sensation.
11961 : The arcane energies begin to course within your veins.
11962 : You feel a mystical warmth welling up inside you!
11963 : The heat seems to have drained from your %.
11964 : Something's not quite right. Head to %, where Arciela awaits you.
```

scripts/zones/East_Ronfaure/npcs/Geomantic_Reservoir.lua
```diff
@@ -0,0 +1,31 @@
...
+    end
+end
+
+entity.onEventUpdate = function(player, csid, option)
+end
+
+entity.onEventFinish = function(player, csid, option)
+    if csid == 15000 then
+        player:addSpell(799) -- TODO: This is not accurate, retail will display message "Player learns Geo-Poison!", with the spell name in green.
```
zach2good: Please use our enums here. 

To get the message, you add another message line, shown here in the trust hidden missions:
https://github.com/LandSandBoat/server/blob/base/scripts/quests/hiddenQuests/Trust_Ingrid.lua#L39-L41

zach2good (changes requested):
+1 cookie for using a retail cap.

Remember that we do a lot of the things you're already trying to do in other zone files/quest files/mission files etc. There are a lot of examples of how to do things.

I would probably break all of this out into a single global like I did here:
https://github.com/LandSandBoat/server/blob/base/scripts/globals/geomagnetic_fount.lua

Then each file will look like:
```lua
    if xi.geomanticReservoir.onTrigger(player, npc) then
        return
    end
end

-- maybe there's an onEventUpdate?

entity.onEventFinish = function(player, csid, option)
    if xi.geomanticReservoir.onEventFinish (player, csid, option) then
        return
    end
end
```

And then you'll have a new global:
```lua
-- SOA Geomantic Reservoirs
require("scripts/globals/keyitems")
require("scripts/globals/missions")

xi = xi or {}
xi.geomanticReservoir = xi.geomanticReservoir or {}

-- A table of each zone, and what spell they give

xi.geomagneticFount.onTrigger = function(player, npc)
    -- Get the zone text ID from the player object
    -- Do your checks
    -- Send your various messageSpecial()'s, or whatever they are
end

xi.geomagneticFount.onEventFinish= function(...)
    etc.
end
```
Good luck 👍

zach2good:
Here are the user-friendly dumps for June-2022: https://github.com/zach2good/UpdateExtractor/tree/versions/2022-june/out

CatsEyeXI:
> ```lua
> xi = xi or {}
> xi.geomanticReservoir = xi.geomanticReservoir or {}
> ```

Thank you for the detailed explanation, I'm slowly trying to chug through the recommendations. One thing I have noticed, is that some zones have multiple reservoirs, such as La Theine. How would you handle such a case? Check for the NPCID to determine which spell to unlock?

TeoTwawki:
> > ```lua
> > xi = xi or {}
> > xi.geomanticReservoir = xi.geomanticReservoir or {}
> > ```
> 
> Thank you for the detailed explanation, I'm slowly trying to chug through the recommendations. One thing I have noticed, is that some zones have multiple reservoirs, such as La Theine. How would you handle such a case? Check for the NPCID to determine which spell to unlock?

ID's shift and would then require an update to the script every time the client version changes, better to modify their name in the sql table so they are separate scripts (can append a number to the name like `Geomantic_Reservoir_2`)

CatsEyeXI:
> > > ```lua
> > > xi = xi or {}
> > > xi.geomanticReservoir = xi.geomanticReservoir or {}
> > > ```
> > 
> > 
> > Thank you for the detailed explanation, I'm slowly trying to chug through the recommendations. One thing I have noticed, is that some zones have multiple reservoirs, such as La Theine. How would you handle such a case? Check for the NPCID to determine which spell to unlock?
> 
> ID's shift and would then require an update to the script every time the client version changes, better to modify their name in the sql table so they are separate scripts (can append a number to the name like `Geomantic_Reservoir_2`)

brilliant!

CatsEyeXI:
if this method is agreeable, i can try to move forward with the rest of these...

CatsEyeXI:
realized i am missing a check here to see if player has required Indi-spell. Ill update soon.

scripts/globals/geomantic_reservoir.lua
```diff
@@ -0,0 +1,63 @@
...
+
+xi.geomanticReservoir.onTrigger = function(player, npc, geoSpell)
+    -- TODO: According to BG-Wiki there is a sequence here that a player can proc '!!' and achieve a Geomancy skill-up.
+    if not player:hasSpell(geoSpell) and player:getMainJob() == xi.job.GEO then
+        player:startEvent(15000, 277)
+    elseif player:hasSpell(geoSpell) and player:getMainJob() == xi.job.GEO then
+        player:messageSpecial(zones[player:getZoneID()].text.UNCANNY_SENSATION)
+    else -- Default text when player is not on GEO
+        player:messageSpecial(zones[player:getZoneID()].text.NOTHING_OUT_OF_ORDINARY)
+    end
```
Xaver-DaRed, line 56:
```
if player:getMainJob() == xi.job.GEO then
    if not player:hasSpell(geoSpell) then
        player:startEvent(15000, 277)
    else
        player:messageSpecial(zones[player:getZoneID()].text.UNCANNY_SENSATION)
    end
else -- Default text when player is not on GEO
    player:messageSpecial(zones[player:getZoneID()].text.NOTHING_OUT_OF_ORDINARY)
end
```

scripts/zones/East_Ronfaure/npcs/Geomantic_Reservoir.lua
```diff
@@ -0,0 +1,27 @@
+-----------------------------------
+-- Area: East Ronfaure
+-- NPC: Geomantic Reservoir
+-- Unlocks: Geo-Poison
+-- !pos 379.572 -39.057 57.502
+-----------------------------------
+require("scripts/globals/geomantic_reservoir")
+require("scripts/globals/spell_data")
+local ID = require("scripts/zones/East_Ronfaure/IDs")
```
Xaver-DaRed: this local is unneeded

CatsEyeXI:
I think this is ready for another review, please let me know if anything else needs changed.

scripts/globals/geomantic_reservoir.lua
```diff
@@ -0,0 +1,65 @@
...
+        end
+    else -- Default text when player is not on GEO
+        player:messageSpecial(zones[player:getZoneID()].text.NOTHING_OUT_OF_ORDINARY)
+    end
+end
+
+xi.geomanticReservoir.onEventFinish = function(player, csid, geoSpell)
+    if csid == 15000 then
+        player:addSpell(geoSpell) -- TODO: addSpell() has a message baked in, but it isn't the correct one.
+        player:messageSpecial(zones[player:getZoneID()].text.LEARNS, geoSpellTable[geoSpell][1])
```
zach2good: You can silence the addSpell message, we do the same thing for trusts:
https://github.com/LandSandBoat/server/blob/base/src/map/lua/lua_baseentity.cpp#L8387-L8389
https://github.com/LandSandBoat/server/blob/base/scripts/quests/hiddenQuests/Trust_Cherukiki.lua#L59-L60

scripts/globals/geomantic_reservoir.lua
```diff
@@ -0,0 +1,65 @@
...
+    [xi.magic.spell.GEO_FRAILTY   ] = { xi.items.GEO_FRAILTY    },
+    [xi.magic.spell.GEO_WILT      ] = { xi.items.GEO_WILT       },
+    [xi.magic.spell.GEO_MALAISE   ] = { xi.items.GEO_MALAISE    },
+    [xi.magic.spell.GEO_GRAVITY   ] = { xi.items.GEO_GRAVITY    },
+    [xi.magic.spell.GEO_HASTE     ] = { xi.items.GEO_HASTE      },
+    [xi.magic.spell.GEO_FADE      ] = { xi.items.GEO_FADE       },
+}
+
+xi.geomanticReservoir.onTrigger = function(player, npc, geoSpell)
+    -- TODO: According to BG-Wiki there is a sequence here that a player can proc '!!' and achieve a Geomancy skill-up.
```
zach2good, line 48: You can leave this as TODO, just leave a link to the source:
https://www.bg-wiki.com/ffxi/Geomantic_Reservoir

scripts/zones/East_Ronfaure/IDs.lua
```diff
@@ -45,6 +45,8 @@ zones[xi.zone.EAST_RONFAURE] =
         VOIDWALKER_UPGRADE_KI_2  = 11063, -- The <keyitem> takes on a deeper, richer hue and becomes <keyitem>!
         VOIDWALKER_BREAK_KI      = 11064, -- The <keyitem> shatters into tiny fragments.
         VOIDWALKER_OBTAIN_KI     = 11065, -- Obtained key item: <keyitem>!
+        LEARNS                   = 11955, -- <name> learns spell!
```
zach2good: LEARNS_SPELL

zach2good (changes requested):

scripts/globals/geomantic_reservoir.lua
CatsEyeXI: neat, thanks for pointing this out.

scripts/zones/East_Ronfaure/IDs.lua
CatsEyeXI: updated

CatsEyeXI:
Hi, is there anything I can do to help move this along?

scripts/globals/geomantic_reservoir.lua
```diff
@@ -0,0 +1,66 @@
...
+    -- https://www.bg-wiki.com/ffxi/Geomantic_Reservoir
+    if player:getMainJob() == xi.job.GEO then
+        if not player:hasSpell(geoSpell) then
+            player:startEvent(15000, 277)
+        else
+            player:messageSpecial(zones[player:getZoneID()].text.UNCANNY_SENSATION)
+        end
+    else -- Default text when player is not on GEO
+        player:messageSpecial(zones[player:getZoneID()].text.NOTHING_OUT_OF_ORDINARY)
+    end
```
Omnione, line 56: this line:
`player:messageSpecial(zones[player:getZoneID()].text.NOTHING_OUT_OF_ORDINARY)`
should be default even on geo, **if** you have the spell,
I'm not 100% on the **UNCANNY_SENSATION** part, as from my own testing it has only come up as part of the CS that starts when acquiring the spell.
ref video here: [https://www.youtube.com/watch?v=GDaXDry-Gsk&list=PLC4LtN6cINj38PQP-gig-YFSe8S_oqeBa&index=4](url)

Also what is the first param used for?, i cant tell what the 277 refers too, and is just a magic number right now.

Other than that the rest seems good

CatsEyeXI: Thanks for the review, I've made the changes as requested per our discussion.

zach2good (approved):

PR #2079, opened 2022-06-22
[Item Mods] Add Double/Triple Attack Damage + mods and functionality

Schaedelschaden:
[Item Mods] Add Double/Triple Attack Damage + mods and functionality

**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

- Creates new Double Attack and Triple Attack mods.
- Adjusts double/triple attack core functionality to ascribe all hits of a double/triple attack the PHYSICAL_ATTACK_TYPE::DOUBLE or PHYSICAL_ATTACK_TYPE::TRIPLE attributes. This allows the core to apply the above mods to all hits of the double/triple attack.

## PR Description

- Addresses the addition and handling of two new item mods.

## Steps to test these changes

!setmod 286 1-100 charName (Sets DOUBLE_ATTACK_DMG to specified value)
!setmod 287 1-100 charName (Sets TRIPLE_ATTACK_DMG to specified value)
!setmod 288 1-100 charName (sets DOUBLE_ATTACK proc rate to specified value)
!setmod 302 1-100 charName (sets TRIPLE_ATTACK proc rate to specified value)

Go to the mob/NM of your choice, set your mods, and start whacking away. Changing the value of the DOUBLE_ATTACK_DMG or TRIPLE_ATTACK_DMG should increase the damage of all double/triple attack white damage hits.

## References
Double Attack Damage - https://www.bg-wiki.com/ffxi/War._Cuisses_%2B2
Double Attack Damage - https://www.bg-wiki.com/ffxi/Agoge_Cuisses_%2B3
Triple Attack Damage - https://www.bg-wiki.com/ffxi/Asn._Poulaines_%2B2

PR #2080, opened 2022-06-22
[Item Mods] Add Double/Triple Attack Damage + mods and functionality

Schaedelschaden:
[Item Mods] Add Double/Triple Attack Damage + mods and functionality

**_I affirm:_**

- Creates new Double Attack and Triple Attack mods.
- Adjusts double/triple attack core functionality to ascribe all hits of a double/triple attack the PHYSICAL_ATTACK_TYPE::DOUBLE or PHYSICAL_ATTACK_TYPE::TRIPLE attributes. This allows the core to apply the above mods to all hits of the double/triple attack.

## PR Description

- Addresses the addition and handling of two new item mods.
- Corrects the deletion of weaponskills.lua from PR #2079

!setmod 1038 1-100 charName (Sets DOUBLE_ATTACK_DMG to specified value)
!setmod 1039 1-100 charName (Sets TRIPLE_ATTACK_DMG to specified value)
!setmod 288 1-100 charName (sets DOUBLE_ATTACK proc rate to specified value)
!setmod 302 1-100 charName (sets TRIPLE_ATTACK proc rate to specified value)

Go to the mob/NM of your choice, set your mods, and start whacking away. Changing the value of the DOUBLE_ATTACK_DMG or TRIPLE_ATTACK_DMG should increase the damage of all double/triple attack white damage hits.

## References
Double Attack Damage - https://www.bg-wiki.com/ffxi/War._Cuisses_%2B2
Double Attack Damage - https://www.bg-wiki.com/ffxi/Agoge_Cuisses_%2B3
Triple Attack Damage - https://www.bg-wiki.com/ffxi/Asn._Poulaines_%2B2

### DIFF
--- a/src/map/attackround.cpp
+++ b/src/map/attackround.cpp
@@ -303,8 +303,7 @@ void CAttackRound::CreateAttacks(CItemWeapon* PWeapon, PHYSICAL_ATTACK_DIRECTION
     // Quad/Triple/Double Attack
     else if (xirand::GetRandomNumber(100) < quadAttack)
     {
-        AddAttackSwing(PHYSICAL_ATTACK_TYPE::QUAD, direction, 4);
-        multiHitOccurred = true;
+        AddAttackSwing(PHYSICAL_ATTACK_TYPE::QUAD, direction, 3);
     }
     else if (xirand::GetRandomNumber(100) < tripleAttack)
     {
@@ -319,13 +318,11 @@ void CAttackRound::CreateAttacks(CItemWeapon* PWeapon, PHYSICAL_ATTACK_DIRECTION
     // Mythic Weapons Aftermath, only main hand
     else if (direction == PHYSICAL_ATTACK_DIRECTION::RIGHTATTACK && xirand::GetRandomNumber(100) < occAttThriceRate)
     {
-        AddAttackSwing(PHYSICAL_ATTACK_TYPE::NORMAL, direction, 3);
-        multiHitOccurred = true;
+        AddAttackSwing(PHYSICAL_ATTACK_TYPE::NORMAL, direction, 2);
     }
     else if (direction == PHYSICAL_ATTACK_DIRECTION::RIGHTATTACK && xirand::GetRandomNumber(100) < occAttTwiceRate)
     {
-        AddAttackSwing(PHYSICAL_ATTACK_TYPE::NORMAL, direction, 2);
-        multiHitOccurred = true;
+        AddAttackSwing(PHYSICAL_ATTACK_TYPE::NORMAL, direction, 1);
     }
     // Iga Garb +2 Set augment: possibility to add another swing while using Dual Wield
     // TODO: Double check correct priority for Empyrian armor modifiers? Outsource? Lua function?
--- a/src/map/attackround.cpp
+++ b/src/map/attackround.cpp
@@ -304,6 +304,7 @@ void CAttackRound::CreateAttacks(CItemWeapon* PWeapon, PHYSICAL_ATTACK_DIRECTION
     else if (xirand::GetRandomNumber(100) < quadAttack)
     {
         AddAttackSwing(PHYSICAL_ATTACK_TYPE::QUAD, direction, 3);
+        multiHitOccurred = true;
     }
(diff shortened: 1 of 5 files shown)


src/map/modifier.h
```diff
@@ -309,6 +309,7 @@ enum class Mod
 
     // Warrior
     DOUBLE_ATTACK      = 288, // Percent chance to proc
+    DOUBLE_ATTACK_DMG  = 286, // Increases "Double Attack" damage/"Double Attack" damage + (in percents, e.g. +20 = +20% damage)
```
WinterSolstice8: 286 is already used by `BARAMNESIA`

See the bottom of the enum:

`    // SPARE = 1038, and onward`

Suggest using 1038, and updating SPARE to 1040 (following using 1039 for the other mod)

src/map/modifier.h
```diff
@@ -356,6 +357,7 @@ enum class Mod
     DESPOIL           = 896, // Increases THF Despoil chance
     PERFECT_DODGE     = 883, // Increases Perfect Dodge duration in seconds
     TRIPLE_ATTACK     = 302, // Percent chance
+    TRIPLE_ATTACK_DMG = 287, // Increases "Triple Attack" damage/"Triple Attack" damage + (in percents, e.g. +20 = +20% damage)
```
WinterSolstice8: 287 is already used by `ATMA`

See the bottom of the enum:

`// SPARE = 1038, and onward`

Suggest using 1039, and updating SPARE to 1040

WinterSolstice8 (changes requested):

WinterSolstice8:
Other than those two minor points, looks good!

src/map/attackround.cpp
```diff
@@ -301,24 +303,29 @@ void CAttackRound::CreateAttacks(CItemWeapon* PWeapon, PHYSICAL_ATTACK_DIRECTION
     // Quad/Triple/Double Attack
     else if (xirand::GetRandomNumber(100) < quadAttack)
     {
-        AddAttackSwing(PHYSICAL_ATTACK_TYPE::QUAD, direction, 3);
+        AddAttackSwing(PHYSICAL_ATTACK_TYPE::QUAD, direction, 4);
```
claywar, line 306: I wonder if this function is misleading.  If we're adding attack swings here, wouldn't that be in addition to the first hit? (So, for quad attack, it would make sense to add 3 to that)

WinterSolstice8: They aren't adding the "first swing" anymore near the bottom, I think this means a call of `AddAttackSwing(PHYSICAL_ATTACK_TYPE::QUAD, direction, 4);` results in 4 hits of QUAD. It currently seems to only give 3 hits of QUAD

Schaedelschaden: I put the multiHitOccurred = true into the Quad, OA3, and OA2 checks anticipating them counting as double/triple attacks but I don't think it's accurate to have them in there. The Quad Attacks assignment was (prematurely) anticipating SE adding similar mods for QA. I'll remove the multiHitsOccurred from those three and leave it as only DA/TA that are affected.

The code at the very bottom of the function adds in the 'default' first hit so it makes sense to leave QA/OA2/OA3 alone.

claywar: Aha, I see now.  Thanks for clarifying!

Schaedelschaden:
Removed multiHitOccurred = true from QA, OA2, and OA3 in attackround.cpp.
Moved DOUBLE_ATTACK_DMG to modID 1038.
Moved TRIPLE_ATTACK_DMG to modID 1039.
Updated SPARE to modID 1040.

src/map/attackround.cpp
WinterSolstice8: Could you add back the Quad attack setting `multiHitOccurred` to true? This way, DA/TA/QA have the same pattern of adding the same amount of hits as their name. It could be confusing otherwise. Thanks!

Schaedelschaden: Restored multiHitOccurred to QA
Restored QA AddAttackSwing value to 4

WinterSolstice8 (approved):

TeoTwawki (approved):

PR #2099, opened 2022-06-24
Reraise menu disappearance fix

mumencoder:
**_I affirm:_**

if the 'accept raise' menu disappears due to zoning/position change it will come back.

cloudy orb
!additem 1551

scroll of instant reraise
!additem 4182

!send testplayer 146
Trade the orb, use the scroll, and then wipe in the battlefield. Wait 3 minutes for removal. Raise menu should reappear afterwards

### DIFF
--- a/src/map/ai/states/death_state.cpp
+++ b/src/map/ai/states/death_state.cpp
@@ -32,6 +32,7 @@ CDeathState::CDeathState(CBattleEntity* PEntity, duration death_time)
 : CState(PEntity, PEntity->targid)
 , m_PEntity(PEntity)
 , m_deathTime(death_time)
+, m_lastRaiseSent(GetEntryTime())
 , m_raiseTime(GetEntryTime())
 {
     m_PEntity->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DEATH, true);
@@ -55,13 +56,15 @@ bool CDeathState::Update(time_point tick)
         Complete();
         m_PEntity->OnDeathTimer();
     }
-    else if (m_PEntity->objtype == TYPE_PC && tick > GetEntryTime() + 8s && !IsCompleted() && !m_raiseSent && m_PEntity->isDead())
+    else if (m_PEntity->objtype == TYPE_PC && tick > m_lastRaiseSent + 8s && !IsCompleted() && m_PEntity->isDead())
     {
         auto* PChar = static_cast<CCharEntity*>(m_PEntity);
         if (PChar->m_hasRaise)
         {
+            // client does not mind receiving redundant raise menu packets
+            // TODO: figure out how the server actually decides to resend these packets
             PChar->pushPacket(new CRaiseTractorMenuPacket(PChar, TYPE_RAISE));
-            m_raiseSent = true;
+            m_lastRaiseSent = tick;
         }
     }
     return false;
--- a/src/map/ai/states/death_state.h
+++ b/src/map/ai/states/death_state.h
@@ -52,7 +52,7 @@ public:
 private:
     CBattleEntity* const m_PEntity;
     duration             m_deathTime;
-    bool                 m_raiseSent{ false };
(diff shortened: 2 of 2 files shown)


TeoTwawki:
> if the 'accept raise' menu disappears due to zoning/position change it will come back.

Disregarding a sec that your draft went all kindsa crazy because branch didn't start clean, what are you trying to fix? Loss of reraise when being zoned is retail accurate and battlefields don't strip it on exit normally because they don't zone you - it's the same zone in the lobby area as it is inside the fight. Last I checked this weas working properly on our base branch.

mumencoder:
I tried to PR from a separate fork, thinking it would be OK since github would only let me have one anyways. Working on fixing that. Also I tested this on retail and the raise menu does come back.

Xaver-DaRed:
Could you share what test did you perform in retail?
From memomry, the only time RR carried over was when expulsed out of a battlefield, as teo suggested, so itwould be useful to know.

TeoTwawki:
> I tried to PR from a separate fork, thinking it would be OK since github would only let me have one anyways. Working on fixing that. Also I tested this on retail and the raise menu does come back.

Tested getting kicked from a battlefield while dead or tested getting ***zoned*** while dead? These are ***not*** the same thing.

mumencoder:
This is not the effect of re-raise carrying over. This is the 'accept raise' menu disappearing and not coming back if the client goes into the ... loading ... screen, which happens on a position/zone change. In retail it will come back.

TeoTwawki:
> This is not the effect of re-raise carrying over. This is the 'accept raise' menu disappearing and not coming back if the client goes into the ... loading ... screen, which happens on a position/zone change. In retail it will come back.

ok look:

you zone into horloise peak, enter bcnm, wipe, get kick out, you are still in horloise peak. You will retain rr menu, on retail. because you still have the effect and the effect will re-open the menu again.

If however, something causes you to get actually zoned - its lost. it will not come back up. This is a rare but annoying circumstance.


We on same page?



If battlefield exit has been stripping it (this shouldn't be zonage) or we've been actually zoning people for that, I am going to be annoyed at the lack of reports since the last time I checked it, coz this used to work accurately.

mumencoder:
Mostly but I maintain that getting kicked out of the battlefield due to the 3 minute timer also removes the menu. I didn't actually test *zoning*, but I assume it would happen as well (getting kicked from a battlefield causes the screen to go black and the ... loading... screen to pop up which is probably what does it)

I should add that I did not test this on LSB base, I did inspect the code and didn't see anywhere it could have been fixed besides deathstate.cpp between the different forks, but I admit I am new to the codebase. I could set up LSB base testing as well for this and will for sure for future PRs.

TeoTwawki:
> Mostly but I maintain that getting kicked out of the battlefield due to the 3 minute timer also removes the menu.

Ok, that is def something for us to fix if that happens, I'm maintaining that that is not a zone (or at least should not be).

I also just had a chat with the team where I said I wouldn't be opposed to going against retail behavior in regard to the zoning part - it's incredibly annoying even if difficult to make happen on purpose. But I know what retail does so when you say "zone" I have to confirm that's actually what you mean or not, since battlefield kick isn't supposed to be a zone at all but both would do a fade to black (so, we could have 2 bugs here, or a diff one than expected, or a simple misunderstanding of what is being adjusted)


Edit: we're doing a little checking in our end as well. So far, seems I am right that you meant just battlefield exit when you said "zoning" but we do have a real problem with bf exit

mumencoder:
Commenting out the lines at https://github.com/LandSandBoat/server/blob/df175711d1f7cd935becd9c091e500a64665b7d1/src/map/lua/luautils.cpp#L2079-L2084 actually seems to fix the problem as well, at least for Charming Trio.

KnowOne134:
just tossing this into here for information:

tested in retail and a zoning while being kicked out of a assault instance (battlefield) RR menu does not persist and you have to homepoint.

PR #2101, opened 2022-06-25
[Interaction] Finish CoP Interaction Format 8-1 -> 8-4

0phie:
**_I affirm:_**

+ Converts CoP 8-1: Garden of Antiquity to interaction format
- Moves default actions and mission completed actions into the mission lua.
- Cleaned up the Rubious Crystal code already present to be more compact.
- Implemented a loop to check tower cs vars and cleanup if requirements to proceed are met.

+ Converts CoP 8-2: A Fate Decided to Interaction format.
- Added additional message text indicating a gate or portal is locked out while the player does not have access to it.
- Added childproof locks on all NPCs which the player should not be able to use while the mission is active. This avoids people cheating the system and shortening the quasilumen run.

+ Converts CoP 8-3: When Angels Fall to interaction format.
- Made all ebon panels the same name so we only reference them once in the lua.
- Rewrote the ebon panel onTrigger to use a for loop rather than an if else chain.
- Added additional dialogue to npcs when they do not have any actions the player can perform on them.
- Closed off any NPCs that should not be able to be interacted with post mission.

+ Converts CoP 8-4: Dawn to interaction format.
- Cleaned up ring provision, check, and update formatting.
- Added needed post mission interactions for ring reprovisioning while Dawn is completed.
- Changed some PromathiaStatus checks in trusts to use the mission status instead.
- Changed the starting requirement of Storms of Fate to use the mission status instead.
- Added additional 'Status' resets to Apocalypse Nigh's reset.

+ Within Quest and Mission scripts PromathiaStatus besides the handful of instances it is used in the shared storyline currently.

+ Tested 8-1 individually by stepping through each step outlined on BG-Wiki as well as following my script.
+ Tested 8-2 individually by stepping through each step outlined on BG-Wiki as well as following my script.
+ Tested 8-3 individually by stepping through each step outlined on BG-Wiki as well as following my script.
+ Tested 8-4 individually by stepping through each step outlined on BG-Wiki as well as following my script.
+ Tested 8-1 through 8-4 in a chain, stepping through each step outlined on BG-Wiki as well as following the scripts.
+ Tested the ring mechanic by talking to the door and changing vars appropriately.
- Tested for the first ring.
- Tested for the second ring. (No Lockout)
- Tested for the third ring. (Lockout worked, artificially set the var to ensure it can be obtained.)
- Tested event lockout if the player has a ring already.


scripts/zones/AlTaieu/DefaultActions.lua
```diff
@@ -1,5 +0,0 @@
-local ID = require("scripts/zones/AlTaieu/IDs")
```
claywar, line 1: Just leaving this here before I have time to review.  This file should _not_ be removed, and that is indeed the default action for _0x0 prior to having access.

0phie: I understand this, but I have moved the default action into the actual mission script based on when the player gains access to Al'Taieu. Is there a specific reason we need this in a separate lua?

PR #2113, opened 2022-06-25
[core] Set TrustEntity and PetEntity to be unclaimable

WinterSolstice8:
**_I affirm:_**

This will prevent assignment of CharEntity's PClaimedMob when it should never be set to a pet or a mob

1. Summon a trust
2. Cure the trust
3. Dismiss the trust
4. Zone (preferably immediately)
5. Don't crash


zach2good (approved):
Looks reasonable

PR #2138, opened 2022-06-28
Comment out call to UpdateServerMessage() until implemented

claywar:
**_I affirm:_**

### DIFF
--- a/scripts/commands/updateservermessage.lua
+++ b/scripts/commands/updateservermessage.lua
@@ -10,5 +10,6 @@ cmdprops =
 }
 
 function onTrigger(player)
-    UpdateServerMessage()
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

scripts/commands/updateservermessage.lua
```diff
@@ -10,5 +10,5 @@ cmdprops =
 }
 
 function onTrigger(player)
-    UpdateServerMessage()
+    -- UpdateServerMessage()
```
TeoTwawki, line 13: we should probably message ppl that its temporary disabled pending rework

TeoTwawki (approved):

Issue #2116, opened 2022-06-26
🐛 [Crash] luautils::OnTrigger  [commit 04cab83]


kaincenteno:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Branch affected by issue 

`base`

## Steps to reproduce

talk to question mark in abby vunkerl

## Expected behavior

no crash

```
(gdb) bt full
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
        set = {__val = {0, 0, 9223372034707292159, 9223372034707292159, 8388608, 0, 1065353216, 0, 1114757840, 0, 4902744558670580787, 0, 4902744560784953698, 0, 1065353216, 0}}
        pid = <optimized out>
        tid = <optimized out>
        ret = <optimized out>
#1  0x00007fe32871e859 in __GI_abort () at abort.c:79
        save_stage = 1
        act = {__sigaction_handler = {sa_handler = 0x7fe3288e95c0 <_IO_2_1_stderr_>, sa_sigaction = 0x7fe3288e95c0 <_IO_2_1_stderr_>}, sa_mask = {__val = {140613617993744, 3432, 140613613303768, 45, 0, 1, 140613614736832, 45, 
              45, 1, 140613617993744, 140613614720160, 140613613253617, 1106284512, 140613312809712, 140613602724784}}, sa_flags = 1106281336, sa_restorer = 0x7fe3168f8af0}
        sigs = {__val = {32, 0 <repeats 15 times>}}
#2  0x00007fe328b1b911 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#3  0x00007fe328b2738c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#4  0x00007fe328b273f7 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
No symbol table info available.
#5  0x000000000040e32b in __clang_call_terminate ()
No symbol table info available.
[...]

PR #2150, opened 2022-06-30
Homepoint the player instead of setting a specific zone

### LINKED ISSUE #2122: 🐛 Invalid Player Position Set Intermittently on Zoning

## Branch affected by issue 

`base`

## Steps to reproduce

**Easy Method** (This is how I usually trigger this issue)
1. Repeat the **Warp** spell repeatedly until you arrive in a position that is outside of your homepoint's map, such as shown the below screenshot.

**Alternate Method** (This just happened to one of xiWeb's new players)
1. Create a brand new character in Sandoria (no warp ring, scroll, warp spell, poison potions, etc)
_--Probably unrelated: Here the player spoke to [Abioleget](https://ffxiclopedia.fandom.com/wiki/Abioleget?so=search) to start the [The Vicasque's Sermon](https://ffxiclopedia.fandom.com/wiki/The_Vicasque%27s_Sermon) quest, then spoke to him again and got stuck in a no-UI bug (I can confirm this happens to me, too.)_
2. Force close your FFXI client by right-clicking the taskbar group and choosing **"Close all windows on this display"**
_-- Repeat step 2 - potentially many times - until step 3 occurs._
3. Reboot FFXI and login to a position outside the map, such as shown in this screenshot.
_-- During my 4 tests it placed me back in the sanctuary 3x and, the other time, it placed me in an invalid position outside of Aht Urgan Whitegate, my previous zone.)_
4. Realize that you have no possible way to leave this scenario and require GM intervention (moderate impact bug for end user)
unknown

## Expected behavior

The player is not placed in an invalid, out-of-map, position.

TeoTwawki:
Closes #2122 but this is really more of a band-aid.

**_I affirm:_**

Since homepointing also takes care of coordinates in the zone I don;' have to set those. Alternatively, I could zero the pos and let the default from the zones lua script move the player to a valid position. I tried to ask before opening the PR but didn't see an answer/ Its a rel;atively trivial change, but I feel homepoiting is actually less of a problem than was thought when the choice of "prevzone or valkurm" was decided. A scroll of instant warp is a mere 7cp, or free daily at the satchel moogle. Not much of an exploit and the repeated log warnings should tip off a server operator.

Honestly, I could not even get myself transported to valkurm on purpose (I tried) without cheating it by editing database to make prevzone invalid, and zooming back to previous zone can be pretty exploitable on servers that don't give everyone the `!return` command. While if you actually land in valkurm at too low a lv your only choice is to die and get homepointed anyway.

**Still, let me know if you'd prefer I just zero the pos.**

Note: while testing this, I discovered _another_ bug which I will open another issue for. It involved new characters only and does not happen once your new character has zoned. This PR does not cause the bug, but it did expose it.

Build, and try to trigger and invalid key message, watch map server move player to homepoint.

Zone normally without tripping it, see player zone as normal.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -274,9 +274,10 @@ void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PCh
 
         if (destination >= MAX_ZONEID)
         {
+            // TODO: work out how to drop player in moghouse that exits them to the zone they were in before this happened, like we used to.
             ShowWarning("packet_system::SmallPacket0x00A player tried to enter zone out of range: %d", destination);
-            ShowWarning("packet_system::SmallPacket0x00A dumping player `%s` to a valid zone!", PChar->GetName());
-            PChar->loc.destination = destination = (uint16)ZONE_SELBINA;
+            ShowWarning("packet_system::SmallPacket0x00A dumping player `%s` to homepoint!", PChar->GetName());
+            charutils::HomePoint(PChar);
         }
 
         zoneutils::GetZone(destination)->IncreaseZoneCounter(PChar);
@@ -320,12 +321,9 @@ void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PCh
         ShowWarning("Client cannot receive packet or key is invalid: %s, Zone: %s (%i)",
                     PChar->GetName(), PChar->loc.zone->GetName(), PChar->loc.zone->GetID());
 
-        // Write a sane location for them
         // TODO: work out how to drop player in moghouse that exits them to the zone they were in before this happened, like we used to.
-        ShowWarning("packet_system::SmallPacket0x00A dumping player `%s` to a valid zone!", PChar->GetName());
-        auto prevZone          = PChar->loc.prevzone ? PChar->loc.prevzone : (uint16)ZONE_VALKURM_DUNES;
-        PChar->loc.destination = prevZone;
-        sql->Query("UPDATE chars SET pos_zone = %u WHERE charid = %u", prevZone, PChar->id);
+        ShowWarning("packet_system::SmallPacket0x00A dumping player `%s` to homepoint!", PChar->GetName());
+        charutils::HomePoint(PChar);
     }
 
     charutils::SaveCharPosition(PChar);


zach2good (approved):
Works for me, the less fiddly logic the better

Roland-J:
As a player, if choose to enter a zone and then, for some reason, I close my client - or whatever - and my position becomes invalid within that zone...
In that situation I'd actually want to stay in the zone - _I chose to go there for a reason, after all_ - and have my position corrected to a valid one, even if it's a zone-line. I might find it annoying to be moved to another zone.

Just wanted to provide a player perspective. As a dev, I don't know if the framework is in place to support that, though. (zone defaults, for every zone)

TeoTwawki:
> As a player, if choose to enter a zone and then, for some reason, I close my client - or whatever - and my position becomes invalid within that zone... In that situation I'd actually want to stay in the zone - _I chose to go there for a reason, after all_ - and have my position corrected to a valid one, even if it's a zone-line. I might find it annoying to be moved to another zone.
> 
> Just wanted to provide a player perspective. As a dev, I don't know if the framework is in place to support that, though. (zone defaults, for every zone)

well surprise, that location data for what zone you were at got corrupted so we MUST do something that moves you. I felt homepointing you was saner than attempting (and failing) to dump you to valkurm. There is a reason I called this a band-aid. We still need to find why players get a corrupted location and fix that.

RAIST5150:
To back Teo, look at what recently happened with my character data (#2116 ).

Screwy crash situation in an Abyssea zone that ultimately resulted in a forced exit on load that has led to corruption on my character that now crashes the server every time it tries to load into the game.

Not just my client... but the SERVER.

If this can prevent similar scenarios in the future... I will take the forced home point over a recurring crash.

TeoTwawki:
Yeah, I do want a better solution eventually, but better to get HP'd than take the whole server down or get a broken char stuck in the ether. :+1: 

I could have rigged previous zone still I think, but its actually more code and its more abusable. (wouldn't be the zoneline - it'd be whatever default pos was)

Dev-Roland:
So, if the player's current position is getting corrupted... and if we can identify a corrupt position...

What if we had two sets of positions for the player?

**Set 1**
Current Zone
Current XYZ

**Set 2**
Last Known Zone (not previous zone)
Last Known XYZ

Where:
A) on every x footsteps of the player set 2 is updated, and where 
B) Set 2 is a metatable with a constraint on the writing operations that rejects corrupt positions

That way, if set 1 gets corrupted, set 2 would have a fallback position that the player wouldn't even realize WAS a fallback position, since it's so close to their current position. The server would have to do the magic to detect the scenario on login and swap set 1 out, and swap set 2 into its place, before the client requests its position.

TeoTwawki:
This is a little to complicated to explain to you in github comments. Having more records of where we knew you were isn't going to help.

> A) on every footstep of the player both are updated, with set 2 being one step behind set 1, and where

There is no "footstep" to track. Movement is client side, and updating your positional data at the server happens on specific code triggers a few of which we do not yet have available that retail does (and trying to discuss that could easily take up my entire weekend)

PR #2153, opened 2022-06-30
Prevent Crashing for New Player LS

### LINKED ISSUE #2145: 🐛 Module: crash on give linkshell on char creation

## Branch affected by issue 

`base`

https://github.com/LandSandBoat/server/blob/base/modules/custom/lua/new_player_linkshell.lua
Appears to be an issue with linkshell being equipped before char creation is fully complete.

## Steps to reproduce

Create new char, something happens causing server crash. 

## Expected behavior

LS is created and equipped on char creation. 
[xi_map.exe_28-6_3-15-0.log](https://github.com/LandSandBoat/server/files/9011886/xi_map.exe_28-6_3-15-0.log)

RunAwayDSP:
**_I affirm:_**

In rare occasions server would crash because player isn't fully recognized by game when it tries sending ls chat to them (null/invalid). This change makes it so it is not equipped but just put into inventory, and chat no longer attempts to go to an invalid player.
Prevents #2145


zach2good:
> Fixes https://github.com/LandSandBoat/server/issues/2145

This doesn't fix the crash, it just does something else so the code that crashes isn't hit

RunAwayDSP:
> > Fixes #2145
> 
> This doesn't fix the crash, it just does something else so the code that crashes isn't hit

true, but it does prevent the crash that is produced due to this module that gives a new character a ls equipped and causes null pointers on rare occasions... No longer equipped = no longer null 👍

PR #2158, opened 2022-06-30
Shen rework + removal of Hakutaku's triple attack mod + remove drop pools from mob bst's pets

dallano:
**_I affirm:_**

Created new branch to organize NM_Mob_Fixes into a single commit as well as shaved off irrelevant and or unnecessary changes.

Changes included:
Shen rework
Removal of Hakutaku's triple attack mod upon request.
Removal of mob beastmaster pet's ability to drop items.


scripts/zones/Bibiki_Bay/mobs/Shen.lua
```diff
@@ -2,14 +2,23 @@
...
+-----------------------------------
 local entity = {}
 
+entity.onMobSpawn = function(mob)
+end
+
 entity.onMobFight = function(mob, target)
-    if (mob:getBattleTime() % 45 == 0) then
-        local mobId = mob:getID()
+    entity.spawnFiltrate(mob, target)
```
claywar: This can just be a local function, and does not need to be attached to the entity table

scripts/zones/Bibiki_Bay/mobs/Shen.lua
```diff
@@ -2,14 +2,23 @@
 -- Area: Bibiki Bay
 --  Mob: Shen
 -----------------------------------
+local ID = require("scripts/zones/Bibiki_Bay/IDs")
+require("scripts/globals/mixins/families/uragnite")
```
Mortalelite: we would point to 

`scripts/mixins/families/uragnite`

Mortalelite:
Looks good

zach2good (approved):
Seems fine, squash on merge

(Please consider keeping the title of your PR, and your commits, somewhat clean looking. Once they become part of the history it's next to impossible for us to change it👍)

scripts/zones/Bibiki_Bay/mobs/Shen.lua
```diff
@@ -2,36 +2,29 @@
...
+entity.onMobSpawn = function(mob)
+end
+
 entity.onMobFight = function(mob, target)
-    if (mob:getBattleTime() % 45 == 0) then
-        local mobId = mob:getID()
-        for i = 1, 2 do
-            if (not GetMobByID(mobId+i):isSpawned()) then
-                SpawnMob(mobId+i):updateEnmity(target)
+    local spawnFiltrate = function()
```
claywar: Is there a reason for defining a function here only to call it right below?  Seems redundant

dallano: I was told to make this a local function. I must have been confused lol. Should I keep it local and just move it outside of onMobFight?

claywar: You're right, and I apologise for my original review.  I meant more that it shouldn't be attached to the object itself, since its unnecessary.  In this case, the function is only used in one spot, so there's really not a need to define a function.

```
    local spawnFiltrate = function()
        if mob:getAnimationSub() == 1 and mob:getHP() > 0 then
            for i = 1, 2 do
                if not GetMobByID(mob:getID()+i):isSpawned() then
                    SpawnMob(mob:getID()+i):updateEnmity(target)
                end
            end
        end
    end

    spawnFiltrate()
```
That's the same as saying:
```
      if mob:getAnimationSub() == 1 and mob:getHP() > 0 then
          for i = 1, 2 do
              if not GetMobByID(mob:getID()+i):isSpawned() then
                  SpawnMob(mob:getID()+i):updateEnmity(target)
              end
          end
      end
```

dallano: Oh! Most definitely. I'll try and employ this more in the future. Thank you

TeoTwawki:
Your one of your previous commits (diff pr? comments appear to have been lost..) used different values than the mixin for Shen's shell mods and I had suggested we (not necessarily you, if you aren't comfortable doing it) modify the mixin to facilitate this. I just want to know if shen needs different values or if using the mixin as is has turned out to be retail-accurate. That is all.

dallano:
I can definitely hop on this! you guys at LSB and over at ASB have been extremely supportive with teaching me how things run and what not. I'll make sure to push some proper tables and mixins soon. Or at least report some findings.

PR #2162, opened 2022-07-01
[core] Adjust party.cpp to allow for party leader removal in "mob" parties safely

WinterSolstice8:
**_I affirm:_**

Previously, mob parties never had deletion/party removal because they just didn't need it. Mobs never destructed, only went dormant. Now with dynamic entities this is no longer the case, so this PR adds in the backend to allow things to clean up safely.

1. !fafnir
2. !hp 0 on fafnir
3. !fafnir before the previous fafnir's nameplate vanishes
4. wait for previous fafnir's nameplate to vanish plus a bit extra for the previous fafnir to destruct
5. don't crash

### DIFF
--- a/src/map/party.cpp
+++ b/src/map/party.cpp
@@ -162,6 +162,13 @@ void CParty::DisbandParty(bool playerInitiated)
             message::send(MSG_PT_DISBAND, data, sizeof data, nullptr);
         }
     }
+    else if (m_PartyType == PARTY_MOBS)
+    {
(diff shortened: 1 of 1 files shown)


RAIST5150:
May actually be correct... just wanted to make sure m_PQuaterMaster is not a typo (vs m_PQuarterMaster).

Have seen similar "misspellings" be legit code before.

WinterSolstice8:
image

TeoTwawki (approved):

zach2good (approved):

PR #2168, opened 2022-07-01
Transport Animation Fixes + Mhaura/Selbina Pirate Event

The-Aerec:
**_I affirm:_**

Co-Authored-By: Seztor + Mattyg

1. Corrects Animations, Animation Directions, Animation Duration for: Airships and Ships

* Animations now properly show for all transports: For example a ship leaving selbina will have the zone animation   timed correctly play during the transport event from selbina -> mhaura

2. Corrects Docking / Departing Animations for Mhaura / Selbina Ships

image

3. Adds Pirate event and scripted ship mobs.
4. Adds safety check for players who logout/crash on Mhaura / Selbina Ships, setting their positions to a safe spot on login.
5. And More~

Ride some boats and airships!


scripts/zones/Ship_bound_for_Mhaura_Pirates/IDs.lua
```diff
@@ -18,20 +18,158 @@ zones[xi.zone.SHIP_BOUND_FOR_MHAURA_PIRATES] =
...
-        ARRIVING_SOON_MHAURA    = 7339, -- We are on our way to Mhaura. We will be arriving soon.
+        ARRIVING_SOON_MHAURA    = 7346, -- We are on our way to Mhaura. We will be arriving soon.
     },
     mob =
     {
-        WIGHT      = 17711120,
-        SILVERHOOK = 17711121,
-        SEA_MONK   = 17682442,
-        SEA_HORROR = 17682446,
+        SEA_CREATURES =
```
zach2good: All of the following new IDs could be built off of a single offset. Better yet, if there is only a single "Wight" in the zone entities, you can just use `zone:queryEntitiesByName(name)` once at zone init (I think things are ready at that point) to get the entity, the id, and then build offsets to all the other mobs that you need.

Having every single mob have it's own entry here is _so_ fragile to ID shifts

zach2good (changes requested):

scripts/zones/Ship_bound_for_Mhaura/Zone.lua
```diff
@@ -4,28 +4,95 @@
...
 local zone_object = {}
 
 zone_object.onInitialize = function(zone)
 end
 
-zone_object.onZoneIn = function(player, prevZone)
+zone_object.onZoneIn = function(player, prevZone, zone)
     local cs = -1
-    if (player:getXPos() == 0 and player:getYPos() == 0 and player:getZPos() == 0) then
+    local zoneID = 221
```
zach2good: No need for magic number, we have `xi.zone.ZONE_NAME`

and below. you only need to get the zone once:

```lua
local mhaura = GetZone(xi.zone.MHAURA)
mhaura:setLocalVar(...)
```
etc.

scripts/zones/Ship_bound_for_Mhaura/Zone.lua
```diff
@@ -4,28 +4,95 @@
...
+
+    if
+        player:getXPos() == 0 and
+        player:getYPos() == 0 and
+        player:getZPos() == 0
+    then
         local position = math.random(-2, 2) + 0.150
         player:setPos(position, -2.100, 3.250, 64)
+        if
+            player:getGMLevel() == 0 and
```
zach2good, line 26: What happens if a player is a GM?

scripts/zones/Ship_bound_for_Mhaura/Zone.lua
```diff
@@ -4,28 +4,95 @@
...
+            GetZone(zoneID):getLocalVar('stateSet') == 0
+        then
+            GetZone(zoneID):setLocalVar('stateSet', 1)
+            GetZone(zoneID):setLocalVar('state', 2)
```
zach2good: If I'm understanding this right: stateSet is for when you have state, and state is the state itself?

Why not just have state start at 1, that was if it's set, it's set. If state is 0, then it isn't set?

scripts/zones/Ship_bound_for_Mhaura/Zone.lua
```diff
@@ -4,28 +4,95 @@
...
+        math.random(0, 100) < 3 and
+        not mob:isSpawned() and
+        os.time() > mob:getLocalVar("respawnTime")
+    then
+        mob:spawn()
+    end
+end
+
+zone_object.onZoneTick = function(zone)
+    if zone:getLocalVar('state') == 1 then
```
zach2good, line 72: You have an enum defined, why not use it?

if zone:getLocalVar('state') == state.PREPARING then

or whatever

scripts/zones/Ship_bound_for_Mhaura_Pirates/IDs.lua
```diff
@@ -18,20 +18,169 @@ zones[xi.zone.SHIP_BOUND_FOR_MHAURA_PIRATES] =
...
+        CHHAYA_SHOP_DIALOG      = 7345, -- May I offer you items to help you on your journey?
+        ARRIVING_SOON_MHAURA    = 7346, -- We are on our way to Mhaura. We will be arriving soon.
     },
     mob =
     {
-        WIGHT      = 17711120,
-        SILVERHOOK = 17711121,
-        SEA_MONK   = 17682442,
-        SEA_HORROR = 17682446,
+        SEA_CREATURES = 17711110, -- Starting Sea Creature offset
```
zach2good, line 28: align these

scripts/zones/Ship_bound_for_Selbina/Zone.lua
```diff
@@ -14,29 +15,101 @@ end
...
+        local players = zone:getPlayers()
+        for _, player in pairs(players) do
+            if player:hasKeyItem(xi.ki.SEANCE_STAFF)
+                and player:getVar("Enagakure_Killed") == 0
+                and not GetMobByID(ID.mob.ENAGAKURE):isSpawned()
+            then
+                GetMobByID(ID.mob.ENAGAKURE):spawn()
+            end
+        end
+        if
```
zach2good, line 52: newlines between these blocks please

scripts/zones/Ship_bound_for_Selbina/Zone.lua
```diff
@@ -14,29 +15,101 @@ end
...
+            math.random() < 0.20 and
+            not GetMobByID(ID.mob.PHANTOM):isSpawned()
+        then
+            GetMobByID(ID.mob.PHANTOM):spawn()
+        end
+    else
+        if GetMobByID(ID.mob.PHANTOM):isSpawned() then
+            DespawnMob(ID.mob.PHANTOM)
+        end
```
zach2good, line 61: newlines between these blocks please

scripts/zones/Ship_bound_for_Selbina/Zone.lua
```diff
@@ -14,29 +15,101 @@ end
...
+        mob:spawn()
+    end
+end
+
+zone_object.onZoneTick = function(zone)
+    if zone:getLocalVar('state') == 1 then
+        if GetMobByID(ID.mob.PHANTOM):isSpawned() then
+            DespawnMob(ID.mob.PHANTOM)
+        end
+        if GetMobByID(ID.mob.SEA_HORROR):isSpawned() then
```
zach2good, line 84: newlines between these blocks please

scripts/zones/Ship_bound_for_Selbina_Pirates/Zone.lua
```diff
@@ -5,25 +5,94 @@
...
+            math.random() < 0.20 and
+            not GetMobByID(ID.mob.PHANTOM):isSpawned()
+        then
+            GetMobByID(ID.mob.PHANTOM):spawn()
+        end
+    else
+        if GetMobByID(ID.mob.PHANTOM):isSpawned() then
+            DespawnMob(ID.mob.PHANTOM)
+        end
+        if GetMobByID(ID.mob.ENAGAKURE):isSpawned() then
```
zach2good, line 66: newlines between these blocks please

scripts/zones/Ship_bound_for_Selbina_Pirates/mobs/Blackbeard.lua
```diff
@@ -0,0 +1,25 @@
...
+local entity = {}
+
+entity.onMobSpawn = function(mob)
+    mob:setMod(xi.mod.SILENCERES, 100)
+    mob:setMod(xi.mod.BINDRES, 100)
+    mob:setMod(xi.mod.MDEF, 50)
+end
+
+entity.onMobDeath = function(mob, player)
+
```
zach2good: no newline in here

scripts/zones/Ship_bound_for_Selbina_Pirates/mobs/Crossbones.lua
```diff
@@ -0,0 +1,19 @@
...
+--  Mob: Crossbones
+-----------------------------------
+local entity = {}
+
+entity.onMobSpawn = function(mob)
+    mob:setMobMod(xi.mobMod.ROAM_COOL, 8)
+end
+
+entity.onMobDeath = function(mob, player)
+
```
zach2good: no newline in here

scripts/zones/Ship_bound_for_Selbina_Pirates/mobs/Ship_Wight.lua
```diff
@@ -0,0 +1,19 @@
...
+--  Mob: Ship Wight
+-----------------------------------
+local entity = {}
+
+entity.onMobSpawn = function(mob)
+    mob:setMobMod(xi.mobMod.ROAM_COOL, 8)
+end
+
+entity.onMobDeath = function(mob, player)
+
```
zach2good: ^

sql/mob_spell_lists.sql
```diff
@@ -4936,7 +4936,30 @@ INSERT INTO `mob_spell_lists` VALUES ('Shens_Filtrate',488,172,1,255); -- water_
...
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 151, 1, 255); -- blizzard iii
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 180, 1, 255); -- blizzaga ii
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 201, 1, 255); -- waterga iii
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 206, 1, 255); -- freeze
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 212, 1, 255); -- burst
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 214, 1, 255); -- flood
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 250, 1, 255); -- ice spikes
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 252, 1, 255); -- stun
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 253, 1, 255); -- sleep
+INSERT INTO `mob_spell_lists` VALUES('Silverhook', 492, 274, 1, 255); -- sleepga ii
```
zach2good: no need for the additional spacing in here, like the rest of the file. Also, new file style needs to be applied

Maybe also PR these separately from this PR in case this PR takes a while to go in - that way you don't have to deal with db conflicts

src/map/lua/lua_baseentity.cpp
```diff
@@ -4789,6 +4789,59 @@ void CLuaBaseEntity::setAnimationSub(uint8 animationsub)
     }
 }
 
+/************************************************************************
+ *  Function: setAnimPath()
```
zach2good: anim -> animation, there's not really any need to truncate names, we have some very long bindings and member names

src/map/packets/entity_update.cpp
```diff
@@ -230,6 +230,27 @@ void CEntityUpdatePacket::updateWith(CBaseEntity* PEntity, ENTITYUPDATE type, ui
             this->setSize(0x48);
             ref<uint16>(0x30) = PEntity->look.size;
             memcpy(data + (0x34), PEntity->GetName(), (PEntity->name.size() > 12 ? 12 : PEntity->name.size()));
+            if (PEntity->manualConfig)
```
zach2good, line 233: Is this only used for ships? I don't really like all of this extra stuff (for ships) being pushed into base entity

src/map/zone.cpp
```diff
@@ -1136,6 +1141,45 @@ void CZone::CheckRegions(CCharEntity* PChar)
...
+//==========================================================
+
+void CZone::SetZoneDirection(uint8 direction)
+{
+    m_ZoneDirection = direction;
+}
+void CZone::SetZoneAnimation(uint8 animation)
+{
+    m_ZoneAnimation = animation;
+}
```
zach2good, line 1155: Need some gaps between these funcs

src/map/zone.h
```diff
@@ -613,6 +613,16 @@ class CZone
 
     time_point m_LoadedAt; // time zone was loaded
 
+    void SetZoneDirection(uint8 direction);
+    void SetZoneAnimation(uint8 animation);
+    void SetZoneAnimStartTime(uint32 startTime);
+    void SetZoneAnimLength(uint16 length);
+
+    uint8           GetZoneDirection();
```
zach2good: These need to be properly aligned, too much space

src/map/zone.h
```diff
@@ -647,6 +657,11 @@ class CZone
 
     time_point m_timeZoneEmpty; // The time_point when the last player left the zone
 
+    uint8           m_ZoneDirection;        // which direction for transport to travel (0 or 4)
```
zach2good: These need to be properly aligned

zach2good (commented):
Seems like it's getting there, my only major concern is the adding of a bunch of different animation state trackers to baseentity. Do all types of entity need to inherit those, or are they just for ships? 

Could we make a new entity type (from NPC I guess?) and store them there?

Or we could keep a global (to the zone) and track the animation states of the few things that need them there

TeoTwawki:
> Could we make a new entity type (from NPC I guess?) and store them there?

we actually have one already, just needs built out farther

https://github.com/LandSandBoat/server/blob/d0212ae023500827cb16d5066ad8449cfdab7231/src/map/entities/baseentity.h#L39

zach2good:
> > Could we make a new entity type (from NPC I guess?) and store them there?
> 
> we actually have one already, just needs built out farther
> 
> https://github.com/LandSandBoat/server/blob/d0212ae023500827cb16d5066ad8449cfdab7231/src/map/entities/baseentity.h#L39

Yeah, this is the way

scripts/zones/Ship_bound_for_Mhaura/Zone.lua
The-Aerec: the event doesn't activate. (This is to stop GMs from using !zone to the zones and breaking the event) 

We don't have anything currently that will let us auto start a transportation event. So this was my: Lets start the pirate event because a normal player is now in the zone.

TeoTwawki: Just want to put out that one of my long term wishlist items is to murder the transports table and make transports self-initialize in script the same way elevators do, and that after that rework it may be worth revisiting to enable an "on demand" transport event after that. Don't anyone hold their breath for it tho, it'll be a while.

scripts/zones/Ship_bound_for_Mhaura/Zone.lua
The-Aerec: I mean could: 

0 is idle
1 is end
2 is start
3 is active

they don't match pirates.lua since the pirates enum only matter for 2/4 ship zones that use state. So I wanted to avoid using those perticular enums. I can make another enum for more descriptiveness in sea_creatures since its used in all four for the boat zone states.

src/map/packets/entity_update.cpp
The-Aerec: Yes, but it should only be used in that case of the ship? I will have to let you all reformat that, as I had to get core assistance in the first place for all the ship timings and animations.

scripts/zones/Ship_bound_for_Mhaura/Zone.lua
The-Aerec: Is this acceptable for the time being then?

TeoTwawki: My comment is unrelated to any mergeability decision, and I haven't personally reviewed, just read comments and decided to give everyone a heads up :+1:

PR #2184, opened 2022-07-04
Provides Docker config defaults

mumencoder:
**_I affirm:_**

Separates docker config defaults from original defaults and provides different network defaults. Merges environment network settings into the cpp settingsMap

In the root folder: 
```
docker-compose up
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,11 +28,14 @@ ADD . /server
 # Configure and build
 RUN mkdir docker_build && cd docker_build && cmake .. && make -j $(nproc)  && cd .. && rm -r /server/docker_build
 
-# Copy the docker config files to the conf folder instead of the default config
-COPY /conf/default/* conf/
-
-# Copy the docker settings files to the settings folder instead of the default settings
-COPY /scripts/settings/default/* scripts/settings/
+# Set SQL_HOST in lua settings 
+RUN echo '\
+xi = xi or {}\n\
+xi.settings = xi.settings or {}\n\
+xi.settings.network =\n\
+{\n\
+    SQL_HOST     = "db",\n\
+}' >> settings/network.lua
 
 # Ensure wait_for_db_then_launch.sh is executable
 RUN chmod +x ./tools/wait_for_db_then_launch.sh


settings/default_docker/network.lua
```diff
@@ -0,0 +1,34 @@
+-----------------------------------
```
zach2good: There has to be a better way of providing defaults to docker that doesn't involve polluting the build for all other non-docker systems?

zach2good: Given that we're on the fence about dumping any in-repo docker files anyway, this doesn't sit well

Sunderous:
So I've spent a bunch of hours over the last few days trying to get this running in Docker on my laptop, so I could connect from my PC. I finally got it working, and this was next on my list figuring out how to put the settings in before I submitted a PR.

After looking at the code, I think the solution is to add these settings:

image

To this piece of code:

image

From the looks of things, if they were in there one could just set those environment variables in their docker-compose file and be good to go they'd override the defaults.

Not sure when I'd be able to test, but looks straightforward.

kaincenteno:
there's no need to add another settings file, the Dockerfile already handles this with

```
ENV XI_DB_HOST=db
ENV XI_DB_PORT=3306
ENV XI_DB_USER=xiuser
ENV XI_DB_USER_PASSWD=xipassword
ENV XI_DB_NAME=xidb
```

the change for `COPY /settings/default/* settings/default/` is needed as settings moved

mumencoder:
There was an issue in settings.cpp which merged in XI_DB_HOST too late. Just that fix alone in this PR would right things but I thought I would add some redundancy and encourage the default settings files to diverge just in case. I have a new alternative Dockerfile that echos a network.lua file to the settings folder, but I could just scrap the Dockerfile changes entirely.

Sunderous:
> 

I'm playing around with this now, any idea why I can override the ZONE_DATA_IP for instance, and print it out at the end very end of common/settings.cpp when we're done loading everything, and it has the correct value.

Yet my connectivity clearly indicates it's not respecting the value I overwrote it to. If it just slap the same value directly into the settings file it all works fine.

Is there some weird thing going on where that lua[] table with all the settings gets used before our overrides are in it?

zach2good:
In order to provide defaults, allow user settings, and publish all of these settings into c++, we have to load things in a very particular order:

https://github.com/LandSandBoat/server/blob/f27717ea0114d360cc1c327444a5c1521603fd13/src/common/settings.cpp#L97

The settings are all loaded as soon as logging and a lua state are built, and they aren't modified after that initial load

Dockerfile
```diff
@@ -28,11 +28,14 @@ ADD . /server
 # Configure and build
 RUN mkdir docker_build && cd docker_build && cmake .. && make -j $(nproc)  && cd .. && rm -r /server/docker_build
 
-# Copy the docker config files to the conf folder instead of the default config
-COPY /conf/default/* conf/
-
-# Copy the docker settings files to the settings folder instead of the default settings
-COPY /scripts/settings/default/* scripts/settings/
+# Set SQL_HOST in lua settings 
+RUN echo '\
```
zach2good, line 32: I think a better way of doing this is to:

- Still COPY the default settings up one level
- Create a new script in tools that replaces the contents of `settings/network.lua` with whatever else it is that you need

If memory serves, `>>` is append and `>` is replace. These files are designed to only have a single table with relevant keys in. If it's working with multiple tables in a file, that's entirely not by design and I wouldn't rely on that.

zach2good (changes requested):

Dockerfile
Sunderous: I've been looking at this in my own branch, and I'm trying to debug some stuff but I think we might be over-complicating this a little bit.

The docker file itself shouldn't have any default settings in it, it's supposed to be configuration agnostic. The settings should be configured in the compose file.

That said, let's say my compose file has env vars: XI_NETWORK_SQL_HOST, XI_NETWORK_SQL_PORT.

In the settings.ccp where it does defaults -> user settings, the last step should be applying any environment variables over the user settings: 

if(getenv(XI_{}_{}), outerKey, innerKey)
  // Apply the environment setting.
  
But for some reason in my branch, even though I can verify the settingsMap is getting updated, there's still some issues.

This seems like the correct approach though because it keeps the docker concerns (build, configure) separated, and it will work with any future config values added so you wouldn't have to touch settings.cpp again ever.

zach2good: > In the settings.ccp where it does defaults -> user settings, the last step should be applying any environment variables over the user settings:
> if(getenv(XI_{}_{}), outerKey, innerKey)
> // Apply the environment setting.

+1 for this approach, then we can get rid of those hard-coded function blocks that try to read the env settings in

zach2good: That would maybe look like this:
```cpp
        // Scrape user settings into cpp's settingsMap
        // This will overwrite the defaults, if user settings exist. Otherwise the
        // defaults will be left intact.
        for (auto [outerKeyObj, outerValObj] : lua["xi"]["settings"].get<sol::table>())
        {
            auto outerKey = outerKeyObj.as<std::string>();

            for (auto [innerKeyObj, innerValObj] : outerValObj.as<sol::table>())
            {
                auto innerKey = innerKeyObj.as<std::string>();
                auto key      = to_upper(fmt::format("{}.{}", outerKey, innerKey));

                if (innerValObj.is<bool>())
                {
                    settingsMap[key] = innerValObj.as<bool>();
                }
                else if (innerValObj.is<double>())
                {
                    settingsMap[key] = innerValObj.as<double>();
                }
                else if (innerValObj.is<std::string>())
                {
                    settingsMap[key] = innerValObj.as<std::string>();
                }

                auto envKey = fmt::format("XI_{}_{}", to_upper(outerKey), to_upper(innerKey));
                if (auto* value = std::getenv(envKey.c_str()))
                {
                    settingsMap[key] = value;
                }
            }
        }
```

zach2good: Actually, we probably want a little logging here too:
```cpp
                auto envKey = fmt::format("XI_{}_{}", to_upper(outerKey), to_upper(innerKey));
                if (auto* value = std::getenv(envKey.c_str()))
                {
                    ShowInfo(fmt::format("Applying setting from environment variable {}: {} -> {}", envKey, key, value).c_str());
                    settingsMap[key] = value;
                }
```

Sunderous: I'll try this in my branch when I get home, but it's more or less what I was doing (I was just doing it in a separate loop, similar to how default/user settings were scraped).

But I get the feeling there's something weird going on with how the settings bubble back up to the lua[] config from the settingsMap that makes it not end up with the right values, so I was planning on just populating the lua again, and putting in the map last. Either way this looks cleaner and I'll try and test it out a bit when I get a chance.

m241dan:
Not to be an absolute barbarian and asshole BUUUUUUT the proper thing to do would be to drop network stuff from settings files all together,  and either read from secret files or environment variables or environment variables that POINT to secret files. Pick your poison. 

I think all this is doing is highlighting the problem that has always been the network / user name / password stuff living in files that could get CM'd.

Xaver-DaRed:
> Not to be an absolute barbarian and asshole BUUUUUUT the proper thing to do would be to drop network stuff from settings files all together, and either read from secret files or environment variables or environment variables that POINT to secret files. Pick your poison.
> 
> I think all this is doing is highlighting the problem that has always been the network / user name / password stuff living in files that could get CM'd.

The proper thing is to just not use docker, imo.

m241dan:
> > Not to be an absolute barbarian and asshole BUUUUUUT the proper thing to do would be to drop network stuff from settings files all together, and either read from secret files or environment variables or environment variables that POINT to secret files. Pick your poison.
> > I think all this is doing is highlighting the problem that has always been the network / user name / password stuff living in files that could get CM'd.
> 
> The proper thing is to just not use docker, imo.

Sure, if that's what you want. Personally, I like docker. But I don't use our docker implementation.

I guess my point was that everything above isn't really a docker issue but a "bad practices" issue.

Sunderous:
> > > Not to be an absolute barbarian and asshole BUUUUUUT the proper thing to do would be to drop network stuff from settings files all together, and either read from secret files or environment variables or environment variables that POINT to secret files. Pick your poison.
> > > I think all this is doing is highlighting the problem that has always been the network / user name / password stuff living in files that could get CM'd.
> > 
> > 
> > The proper thing is to just not use docker, imo.
> 
> Sure, if that's what you want. Personally, I like docker. But I don't use our docker implementation.
> 
> I guess my point was that everything above isn't really a docker issue but a "bad practices" issue.

I think this gets us 90% of the way there. If we implement it the generic way, then you're free to set ANY setting you want through the environment, doesn't matter if you're running in docker or not.

After that it's up to the user if they want to use files or not. Having it in files may not be proper for production when secrets are involved, but having runnable defaults for local development packaged with it is a pattern I feel lots of projects use.

Docker isn't everyones cup of tea, but I think it'll be nice to have it in a state that you can docker-compose and get it running immediately to start playing around. The stuff I'm working on now is just to expand that a bit, so you can override some options if you want to also access it from another machine (I like to code on my laptop, but don't have FFXI installed there). Because the default docker behavior leaves it only accessible from the same box.

TeoTwawki:
how about, someone NOT ME make a docker fork so I can stop hearing about docker in a project that never had any intention to support it

claywar:
Outsider opinion coming in.  While I'm not a fan of the idea of Docker, I can see that some folks are highly interested in it.  That's great; however, anything that hits this repo devs will be expected to support by other parties.  Given that none of us are well-versed in its usage and upkeep, I would suggest that this moves to either a submodule or wiki pages for unsupported vs the mainline branch.

zach2good:
+1 for wiki page

Sunderous:
> how about, someone NOT ME make a docker fork so I can stop hearing about docker in a project that never had any intention to support it

I'm moderately invested in this at this point haha, so if you want to either make a "docker/base" branch I'd volunteer to maintain that. Or maintain it in a totally separate fork if that's more preferrable. Not super open-source-contribution savvy so no idea what the right way to do it is apologies.

TeoTwawki:
> > how about, someone NOT ME make a docker fork so I can stop hearing about docker in a project that never had any intention to support it
> 
> I'm moderately invested in this at this point haha, so if you want to either make a "docker/base" branch I'd volunteer to maintain that. Or maintain it in a totally separate fork if that's more preferrable. Not super open-source-contribution savvy so no idea what the right way to do it is apologies.

IMO, ideally someone very savvy with both git and docker would just fork and keep it updated while handling docker specific issues.

zach2good:
> > how about, someone NOT ME make a docker fork so I can stop hearing about docker in a project that never had any intention to support it
> 
> I'm moderately invested in this at this point haha, so if you want to either make a "docker/base" branch I'd volunteer to maintain that. Or maintain it in a totally separate fork if that's more preferrable. Not super open-source-contribution savvy so no idea what the right way to do it is apologies.

I mentioned in the p.server discord the other day:

```
zach2good — Yesterday at 17:17
I'd rather not have something in the repo that's maintained by someone who isn't in the "core dev" group. We have a long and storied history of people abandoning their stuff and leaving it to us to clean their mess
Not even a targetted statement, it's just happened so many times
```

A fork is probably the least invasive for us 👍

Sunderous:
> > > how about, someone NOT ME make a docker fork so I can stop hearing about docker in a project that never had any intention to support it
> > 
> > 
> > I'm moderately invested in this at this point haha, so if you want to either make a "docker/base" branch I'd volunteer to maintain that. Or maintain it in a totally separate fork if that's more preferrable. Not super open-source-contribution savvy so no idea what the right way to do it is apologies.
> 
> IMO, ideally someone very savvy with both git and docker would just fork and keep it updated while handling docker specific issues.

Sorry I also missed the message about the wiki page. Considering it's just a template for a Dockerfile + docker-compose, I agree it can probably just put it in the wiki without having a dedicated fork (a whole fork seems heavy). And I'd volunteer to keep that page up to date monthly or something once we get the initial template working. But if that also messes with the core dev setup, then can just fork it for now.

I'm mostly just trying to get this setup with the env vars for my own workflow, before I start getting poking at it to see if there's anything I can help implement.

m241dan:
> > > how about, someone NOT ME make a docker fork so I can stop hearing about docker in a project that never had any intention to support it
> > 
> > 
> > I'm moderately invested in this at this point haha, so if you want to either make a "docker/base" branch I'd volunteer to maintain that. Or maintain it in a totally separate fork if that's more preferrable. Not super open-source-contribution savvy so no idea what the right way to do it is apologies.
> 
> IMO, ideally someone very savvy with both git and docker would just fork and keep it updated while handling docker specific issues.

So, seeing as no one at the top is interested in supporting Docker. I vote for dropping it. I use docker, but not the stuff provided because our provided docker is not great, safe, or ideal for my deployment pipeline / development practices.

I'd be willing to try to support a better docker practices in community but doing so in a separate branch / fork / git / whatever seems silly. Either support it or don't. If you do, because it is a useful tool, let's do it right.

Don't make me invoke the ghost of Christmas past's "half hearted, half implemented garbage" that causes tech debt for like literal generations at this point.

TeoTwawki:
> > > > how about, someone NOT ME make a docker fork so I can stop hearing about docker in a project that never had any intention to support it
> > > 
> > > 
> > > I'm moderately invested in this at this point haha, so if you want to either make a "docker/base" branch I'd volunteer to maintain that. Or maintain it in a totally separate fork if that's more preferrable. Not super open-source-contribution savvy so no idea what the right way to do it is apologies.
> > 
> > 
> > IMO, ideally someone very savvy with both git and docker would just fork and keep it updated while handling docker specific issues.
> 
> So, seeing as no one at the top is interested in supporting Docker. I vote for dropping it. I use docker, but not the stuff provided because our provided docker is not great, safe, or ideal for my deployment pipeline / development practices.
> 
> I'd be willing to try to support a better docker practices in community but doing so in a separate branch / fork / git / whatever seems silly. Either support it or don't. If you do, because it is a useful tool, let's do it right.
> 
> Don't make me invoke the ghost of Christmas past's "half hearted, half implemented garbage" that causes tech debt for like literal generations at this point.

***rattles chains***


You can't have my job.


edit: may seem silly to you, but anybody can fork and have any additions they want. We already don't support truly it and that's why it being in the repo is being a problem for us.

Dockerfile
Sunderous: > Actually, we probably want a little logging here too:
> 
> ```c++
>                 auto envKey = fmt::format("XI_{}_{}", to_upper(outerKey), to_upper(innerKey));
>                 if (auto* value = std::getenv(envKey.c_str()))
>                 {
>                     ShowInfo(fmt::format("Applying setting from environment variable {}: {} -> {}", envKey, key, value).c_str());
>                     settingsMap[key] = value;
>                 }
> ```

@zach2good Hey, sorry only got to this now and figured I'd ask here. I'm not a C++ guru, haven't used it in like 10 years so a lot of the 14/17 features are new to me.

After implementing this idea, I'm getting the same behavior as I was when I tried it my way: calling 
`get<std::string>("NETWORK.SQL_HOST)`
for instance, prints "TRUE", so it's storing the values as booleans not strings. Any idea what might be going on there between the auto and variant<bool, double, string> types? Thanks.

Once I get this working, I'll make a base fork + write up a PR for a wiki page after ironing out the other kinks.

Sunderous: Scratch it, got it sorted mostly, but not ideal:
- Trying to set the value in the if statement for whatever reason makes it come out the other side as a boolean.
- Port values need to be converted to ints/doubles before they're sent to the lua config, or they don't get used properly.

Will eventually make a PR with the settings files updates, and can talk about it there if there's a better approach for keeping the config vars that have to be typed for LUA in sync.

zach2good:
This PR ended up kicking off a series of events that lead to us making some internal changes and deciding to remove docker support from the repo, in favour of just having a wiki page explaining what to do:

https://github.com/LandSandBoat/server/pull/2219

So, going to close this PR. Thanks for trying though!

PR #2190, opened 2022-07-04
Fixes Content Restriction Setting

The-Aerec:
**_I affirm:_**

Fixes the ability to use content restriction for server settings

Enable/disable expansions and turn on Restrict_content


WinterSolstice8:
If you meant to PR this here, you can always check "changed files" before publishing it to make sure that it's what you want

can always fix the branch and force push it as well

TeoTwawki:
> can always fix the branch and force push it as well

instead of giving everyone double or triple the notifications by closing and opening new ones in new branches for every mistake made.

PR #2191, opened 2022-07-04
Fix Content Restriction & remove GetSettingsVariable binding

The-Aerec:
Co-Authored-By: Ophelia <heliashelium@gmail.com>

**_I affirm:_**

Fixes Content_restrict setting in main settings

enable/disable content and then turn on content restriction

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -258,7 +258,7 @@ void CBattlefield::ApplyLevelRestrictions(CCharEntity* PChar) const
         // Check if it's a mission and if config setting applies.
         if (!settings::get<bool>("map.LV_CAP_MISSION_BCNM") && m_isMission == 1)
         {
-            cap = luautils::GetSettingsVariable("MAX_LEVEL"); // Cap to server max level to strip buffs - this is the retail diff between uncapped and capped to max lv.
+            cap = settings::get<uint8>("main.MAX_LEVEL"); // Cap to server max level to strip buffs - this is the retail diff between uncapped and capped to max lv.
         }
 
         PChar->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DEATH, true);
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -1510,7 +1510,7 @@ namespace luautils
     {
         TracyZoneScoped;
         TracyZoneCString(variable);
-        return lua["xi"]["settings"][variable].valid() ? lua["xi"]["settings"][variable].get<uint8>() : 0;
+        return lua["xi"]["settings"]["main"][variable].valid() ? lua["xi"]["settings"]["main"][variable].get<uint8>() : 0;
     }
 
     /************************************************************************
@@ -4789,7 +4789,6 @@ namespace luautils
             // 0xFFFF is gil, so we will always return a value less than that as a warning
             id = 0xFFFF - sql->NumRows() + 1;
         }
-        
 
         return id;
     }
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -251,7 +251,7 @@ namespace luautils
         }
 
         // Handle settings
-        contentRestrictionEnabled = GetSettingsVariable("RESTRICT_CONTENT") != 0;
+        contentRestrictionEnabled = settings::get<bool>("main.RESTRICT_CONTENT");
 
         moduleutils::LoadLuaModules();
 
@@ -1502,19 +1502,6 @@ namespace luautils
 
     /************************************************************************
      *                                                                       *
-     *  Get a Variable From Settings.lua                                     *
-     *                                                                       *
-     ************************************************************************/
-
-    uint8 GetSettingsVariable(const char* variable)
-    {
-        TracyZoneScoped;
-        TracyZoneCString(variable);
-        return lua["xi"]["settings"]["main"][variable].valid() ? lua["xi"]["settings"]["main"][variable].get<uint8>() : 0;
-    }
-
-    /************************************************************************
-     *                                                                       *
      *  Check if an something is restricted in Settings.lua                  *
      *  ENABLE_ is subject to RESTRICT_BY_EXPANSION                          *
      *  ALLOW_ is NOT subject to RESTRICT_BY_EXPANSION                       *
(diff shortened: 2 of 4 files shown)


zach2good (commented):
Do we use `GetSettingsVariable` in many places? Now that we have direct access to the settings (and there are more settings areas than just main) we can probably rip out this binding

The-Aerec:
> Do we use GetSettingsVariable in many places? Now that we have direct access to the settings (and there are more settings areas than just main) we can probably rip out this binding

No not really. I didn't find many uses. This was just the quick fix Ophie and I found that fixed the issue.

Based on a quick search:

image

Yea its uh not used pretty much at all =p

zach2good:
Replace then I think, then it's clear what file the setting is being read from, and it's easier to move them around in the future without breaking things

sql/npc_list.sql
```diff
@@ -7628,7 +7628,7 @@ INSERT INTO `npc_list` VALUES (17093444,'Volker','Volker',0,0.000,0.000,0.000,32
 INSERT INTO `npc_list` VALUES (17093445,'Ajido-Marujido','Ajido-Marujido',0,0.000,0.000,0.000,32768,50,50,9,0,0,6,27,0x00004A0000000000000000000000000000000000,32,'TOAU',0);
 INSERT INTO `npc_list` VALUES (17093446,'Curilla','Curilla',0,0.000,0.000,0.000,32768,50,50,0,0,0,6,27,0x0000450000000000000000000000000000000000,32,'TOAU',0);
 INSERT INTO `npc_list` VALUES (17093447,'Ayame','Ayame',0,0.000,0.000,0.000,32768,50,50,0,1,0,6,27,0x0100030211100920003009400950006000700000,32,'TOAU',0);
-INSERT INTO `npc_list` VALUES (17093448,'Shantotto','Shantotto',0,0.000,0.000,0.000,32768,50,50,0,1,0,6,27,0x0100060600101320133013401350006000700000,32,'TOUA',0);
+INSERT INTO `npc_list` VALUES (17093448,'Shantotto','Shantotto',0,0.000,0.000,0.000,32768,50,50,0,1,0,6,27,0x0100060600101320133013401350006000700000,32,'TOAU',0);
```
zach2good, line 7631: Nice catch

src/map/utils/zoneutils.cpp
```diff
@@ -337,8 +337,8 @@ namespace zoneutils
 
     void LoadMOBList()
     {
-        uint8 normalLevelRangeMin = luautils::GetSettingsVariable("NORMAL_MOB_MAX_LEVEL_RANGE_MIN");
-        uint8 normalLevelRangeMax = luautils::GetSettingsVariable("NORMAL_MOB_MAX_LEVEL_RANGE_MAX");
+        uint8 normalLevelRangeMin = settings::get<int>("main.NORMAL_MOB_MAX_LEVEL_RANGE_MIN");
+        uint8 normalLevelRangeMax = settings::get<int>("main.NORMAL_MOB_MAX_LEVEL_RANGE_MAX");
```
zach2good: Keep an eye on what types you're getting, this is a truncation: int -> uint8. 

I think for most of these settings that you've replaced, you should be asking for bool or uint8

src/map/lua/luautils.cpp
```diff
@@ -1538,8 +1525,10 @@ namespace luautils
             }
             else
             {
+                std::string fileName("main.");
+                fileName.append(contentVariable);
                 // Cache contentTag lookups in a map so that we don't re-hit the Lua file every time
-                contentEnabled = (GetSettingsVariable(contentVariable.c_str()) != 0);
+                contentEnabled = (settings::get<int>(fileName)) != 0;
```
zach2good: This section could be:
```cpp
contentEnabled = settings::get<bool>(fmt::format("main.{}", contentVariable));
```

src/map/lua/luautils.cpp
```diff
@@ -251,7 +251,7 @@ namespace luautils
         }
 
         // Handle settings
-        contentRestrictionEnabled = GetSettingsVariable("RESTRICT_CONTENT") != 0;
+        contentRestrictionEnabled = settings::get<int>("main.RESTRICT_CONTENT") != 0;
```
zach2good: contentRestrictionEnabled is a bool, ask settings for a bool: settings::get<bool>

If it's a 1 in the settings, it'll automagically convert it during lookup

src/map/battlefield.cpp
```diff
@@ -258,7 +258,7 @@ void CBattlefield::ApplyLevelRestrictions(CCharEntity* PChar) const
         // Check if it's a mission and if config setting applies.
         if (!settings::get<bool>("map.LV_CAP_MISSION_BCNM") && m_isMission == 1)
         {
-            cap = luautils::GetSettingsVariable("MAX_LEVEL"); // Cap to server max level to strip buffs - this is the retail diff between uncapped and capped to max lv.
+            cap = settings::get<int>("main.MAX_LEVEL"); // Cap to server max level to strip buffs - this is the retail diff between uncapped and capped to max lv.
```
zach2good: cap is uint8, ask settings for uint8

zach2good (changes requested):
Almost there

src/map/lua/luautils.cpp
```diff
@@ -1539,7 +1526,7 @@ namespace luautils
             else
             {
                 // Cache contentTag lookups in a map so that we don't re-hit the Lua file every time
-                contentEnabled = (GetSettingsVariable(contentVariable.c_str()) != 0);
+                contentEnabled = settings::get<bool>(fmt::format("main.{}", contentVariable)) != 0;
```
zach2good: No need for ` != 0;`, this is now a bool of its own and doesn't need to converted into one

zach2good (changes requested):

src/map/lua/luautils.cpp
```diff
@@ -251,7 +251,7 @@ namespace luautils
         }
 
         // Handle settings
-        contentRestrictionEnabled = GetSettingsVariable("RESTRICT_CONTENT") != 0;
+        contentRestrictionEnabled = settings::get<bool>("main.RESTRICT_CONTENT") != 0;
```
zach2good: Same here

zach2good (approved):
🎉

claywar (approved):

sql/npc_list.sql
TeoTwawki: OH HOHO h...oh oh?

TeoTwawki:
> Do we use `GetSettingsVariable` in many places? Now that we have direct access to the settings (and there are more settings areas than just main) we can probably rip out this binding

historical footnote: we used to use it for some more things that no longer exist. old old oooold times you'd have to go way back in history to see.

PR #2192, opened 2022-07-04
ZNM Part 1: Sanraku, Ryo, Soul ***s, and Zeni

m241dan:
**_I affirm:_**

Implements Sanraku, Ryo, Soul plate / Soul trapper, and Zeni. Handles the pop cost dynamic components, Sanraku's interest / fauna, and allows for trading in Soul plates to get Zeni.

The only question mark is when Sanraku's interest / fauna should shift. According to chatter on wiki's, it's more of a "X many have been traded, time to change thing" then a "JST MIDNIGHT" / "VANADIEL DAY" thing.

**Tier 1 ZNM buying is disabled (thus all doing the mobs are actually disabled).** Part 2 will be audited Tier 1 NMs

Big thanks to @zach2good and @dphillips2014 for doing the majority of the work here. 

1. Talk to Sanraku
2. Talk to Ryo
3. Obtain Soultrapper / Plates using Jettons
4. Go take some photos (following Sanraku's interest or fauna if you want)
5. Trade plates to Sanraku and receive Zeni
6. Notice that you cannot buy any Tier 1 pops, frown, but understand that we need to audit the Tier 1 NMs first.

### DIFF
--- a/scripts/globals/znm_data.lua
+++ b/scripts/globals/znm_data.lua
@@ -258,7 +258,8 @@ xi.znm.POP_ITEMS =
 ------------------------------------------------------------
 -- Default: Tier 1 ZNMs + "Don't Ask"
 -- (if bit = 0: add ZNM to Sanraku's Menu)
-xi.znm.DefaultMenu = 0x7F8FE3F8
+-- xi.znm.DefaultMenu = 0x7F8FE3F8
+xi.znm.DefaultMenu = 0x7FFFFFFF -- No Tier 1 NMs available (need to audit their behaviours)
 
 -- Adjusts the bitmask based on owned seals
 xi.znm.MENU_BITMASKS =


scripts/globals/pankration.lua
```diff
@@ -20,71 +20,84 @@ xi.pankration = xi.pankration or {}
...
-    189, -- Orc
-    190, -- Orcish Warmachine
-    196, -- Poroggo
-    199, -- Qiqirn
-    200, 201, 202, -- Quadav
-    213, -- Sahagin
-    243, 244, -- Tonbery
-    246, -- Troll
-    270, -- Yagudo
+    [25] = true, -- Antica
```
zach2good: If these are just going to be used as quick lookup confirmation for having an entry in this table, we have a util for `set` somewhere, you can see it in the rov global I think

zach2good: If none of these are going to be false, this is a set, and we have a util for that (you can see it used in the rov global)

scripts/globals/znm.lua
```diff
@@ -16,27 +17,98 @@ require("scripts/globals/msg")
...
 
 xi = xi or {}
 xi.znm = xi.znm or {}
 
+--------------------------------------------------
+-- Sanraku's Interest and Recommended Fauna
+-- Applies bonuses to soul plate zeni-value
+--------------------------------------------------
+-- Called during JstMidnight tick
+xi.znm.UpdateSanrakusMobs = function()
```
zach2good: Lowr case func name

scripts/globals/znm.lua
```diff
@@ -16,27 +17,98 @@ require("scripts/globals/msg")
...
+    return false
+end
+
 -----------------------------------
 -- Soultrapper
 -----------------------------------
 xi.znm.soultrapper = xi.znm.soultrapper or {}
 
+-----------------------------------
+-- onItemCheck
```
zach2good: No need for big header comments with the name of the function. This doesn't add any additional info. The existing block comments are for different sections

scripts/globals/znm.lua
```diff
@@ -147,61 +251,326 @@ end
...
 
-        player:addCurrency("zeni_point", zeniValue)
+xi.znm.updatePopPrice = function(znm_tier)
+    if not xi.znm.ZNM_STATIC_POP_PRICES then
+        local pop_cost = math.min(xi.znm.getPopPrice(znm_tier) + xi.znm.ZNM_POP_COSTS[znm_tier].addedPrice,
+                xi.znm.ZNM_POP_COSTS[znm_tier].maxPrice)
+        SetServerVariable("[ZNM][T" .. znm_tier .. "]PopCost", pop_cost )
+    end
+end
+
```
zach2good: Multiple line breaks

scripts/globals/znm.lua
```diff
@@ -147,61 +251,326 @@ end
...
+            SetServerVariable("[ZNM][T" .. znm_tier .. "]PopCost", pop_cost)
+        end
     end
 end
+
+-----------------------------------
+---- General Utility Helpers
+-----------------------------------
+
+
```
zach2good: Multiple line breaks

scripts/globals/znm_data.lua
```diff
@@ -0,0 +1,285 @@
...
+-----------------------------------
+
+xi = xi or {}
+xi.znm = xi.znm or {}
+
+-----------------------------------
+--- General Helper Variables
+-----------------------------------
+
+-- Soultrapper Variables
```
zach2good: I wonder if these could be broken out into the settings

src/map/lua/luautils.cpp
```diff
@@ -937,6 +937,37 @@ namespace luautils
         message::send(zoneId, str);
     }
 
+    /***************************************************************
+     * Function calls for Zeni Notorious Monster (ZNMs)
+     ***************************************************************/
+    // Update Sanraku's Interests and Recommended Fauna
```
zach2good: Both of these comments don't add any more context than the function name itself

zach2good (commented):
Generally looks very tidy, I'm sure the ASB peeps would be happy to help you playtest this

The-Aerec:
@TiberonKalkaz Can probably comment, he did all the research for ASB prior

scripts/globals/znm.lua
m241dan: I'll remove, but I find they provide a nice "reading" break in the code.

m241dan:
> Generally looks very tidy, I'm sure the ASB peeps would be happy to help you playtest this

Sure, they're welcome to playtest but we also play testing on my server as well.

m241dan:
> @TiberonKalkaz Can probably comment, he did all the research for ASB prior

Ya, we'd love a little more feedback on when to change his interest and what the actual values that come out when you trade should be. Since I guess it's possible to get over 1k in modern times?

scripts/globals/znm_data.lua
m241dan: Could be. Are Settings the future or are modules?

zach2good: Tweaking numbers: settings
Random custom non-retail bullshit: modules

scripts/globals/znm.lua
```diff
@@ -147,61 +251,326 @@ end
...
+    local plateData = item:getSoulPlateData()
 
-    if npcUtil.tradeHasExactly(trade, xi.items.SOUL_PLATE) then
-        -- Cache the soulplate value on the player
-        local item = trade:getItem(0)
-        local plateData = item:getSoulPlateData()
-        player:setLocalVar("[ZNM][Sanraku]SoulPlateValue", plateData.zeni)
-        player:startEvent(910, plateData.zeni)
+    xi.znm.sanraku.setTradedPlateValue(player, plateData.zeni)
+    player:startEvent(910, plateData.zeni)
```
TiberonKalkaz: I'll assume it was tested but one issue I ran into on wings was trying to align the 75 cap era zeni values per plate with the client side logic for the plate trade in CS.

In 75 cap era 100 points was exceptional.  However the client wings uses actually considers 100 points to be a bad example.  This is something that LSB likely does not have to be concerned with - but ASB will have to address

```
EventEntryPoint_16982568_910.L0B42:
.L0B42: MOV [0003], Param0
        Opcode20 01
        Opcode42
        LOOKAT [Player]
        WAIT16
        Opcode70
        MOV EventResult, #1
        UPDATEEVENT
        UPDATEEVENT WAIT
        NPCSAY #13135 // Oh my, is that a soul plate you have brought for me?\x07If I could be permitted to do the honors...<7F31>
        WAITFORINPUT
        NPCSAY #13136 // Aaah...hmmm...<7F31>
        WAITFORINPUT
        CMP [0003] >= #495, .L0B71
        NPCSAY #13139 // Hmmm... Surely you can do better than this, my friend. But I will accept it nevertheless.<7F31>
        WAITFORINPUT
        JMP .L0B84

.L0B71: CMP [0003] >= #990, .L0B80
        NPCSAY #13138 // A fine piece of photography!\x07You have my thanks!<7F31>
        WAITFORINPUT
        JMP .L0B84

.L0B80: NPCSAY #13137 // Magnificent! Truly a triumph in photography!\x07My faith in you is obviously well placed!<7F31>
        WAITFORINPUT
.L0B84: MOV Param1, [0003]
        PRINT #13087 // You earn {e:1} zeni.<7F31>
        WAITFORINPUT
        Opcode2E
        Opcode20 00
        FINISHEVENT
        END
```

TeoTwawki: > trying to align the 75 cap era zeni values 

the values on retail didn't really change, what happened was newer mobs were added to the game that were just so much higher strength that their base value actually exceeded the values for lower lv things Sanraku's actually wanted, so we all started doing Nauls (mini fafnirs) in abyssea tahrongi instead. This is also why SE blocked soul plating mobs in adoulin or eschan zones even if the mob type was already allowed prior.

TiberonKalkaz:
In 75 cap era, 100 per plate was the top end.
Retail is closer to 1k per plate (I think, I dont have a retail account) and generally easy to hit that from what I understand.

It will be tedious, but someone could spend time capturing to see what the current retail caps.
  
My work on wings targeted era and a known smaller community so the limits used to trigger category changes were based on 100 zeni soul plates and population of 500-1k.  The intended outcome was rapidly changing categories when new tiers were released, and more calm rotations when people relaxed.

Having the limits be configurable is likely the best choice as different size servers will have different wants.  
Capturing will inform the defaults.

scripts/globals/znm.lua
```diff
@@ -16,27 +17,98 @@ require("scripts/globals/msg")
...
+    if fauna == nil or fauna == 0 then
+        fauna = math.random(#xi.znm.SANRAKUS_FAUNA)
+        SetServerVariable('[ZNM][Sanraku]Fauna', fauna)
+    end
+    return fauna
+end
+
+--- Does this mob fall under Sanraku's current "Subject of Interest"?
+-- Some families had multiple famIDs (see Sea Monks), and some superfamIDs were too general (see elementals)
+-- In order to only need to use/store a single ID per mob, a bool was added to check what type of ID is stored
```
TiberonKalkaz: Zeni calculation is a two step process.  Basic parameters must be captured at the time the picture is taken (hpp, distance, etc).  Bonus parameters (matching fauna or interest) can only be determined when the plate is traded.

The boolean for matching fauna or interest was introduced due to the wings code base having limitations on pulling mob information.  We did not have a passive way to generate or look up mob information, and considering instance mob generation required the instance the mob was spawned in - active generation was ruled out.   (read: I wasnt going to implement a full dynamic mob system or mob query system to get a value off a zeni plate)

Additionally, there are space limitations for m_extra, so getting down to just a few bytes of data was ideal.

In LSB, I was considering that we may be able to resolve this and return to storing mobID, which would simplify some use cases like the ASA mission which uses the soul trapper.  However, Dynamic Mobs (if I understand the impl correctly) would add complexity to that endeavor.  This could be solved by not allowing pictures of dynamic mobs but half the fun of soul trappers is seeing them go off in dynamis.

TeoTwawki:
> In 75 cap era, 100 per plate was the top end. Retail is closer to 1k per plate (I think, I dont have a retail account) and generally easy to hit that from what I understand.
> 
> It will be tedious, but someone could spend time capturing to see what the current retail caps.
> 
> My work on wings targeted era and a known smaller community so the limits used to trigger category changes were based on 100 zeni soul plates and population of 500-1k. The intended outcome was rapidly changing categories when new tiers were released, and more calm rotations when people relaxed.
> 
> Having the limits be configurable is likely the best choice as different size servers will have different wants. Capturing will inform the defaults.

I will agree on being configurable. Then anyone can set it however they wish or believe, like what we do for crafting.

But they didn't really cap it at 100 they just weren't worth much plus there were a bunch of rules under the hood: distance from target, targets hp%, matching request of the time from the npc or not etc etc 


I did tons of this system and even capped out my zeni and jettons to the max value at one point. I still have a crazy high amount on one of my characters on retail.  Then along came abyssea, every plate worth 1400+ even if sanraku wanted something else, even more if his wanted thing was dragons that day.  I got over 100 _the day it came out._ 127 to be exact. There weren't stronger mobs to really get more off unless you got really lucky with the combinations the old man asked for.

scripts/globals/znm.lua
```diff
@@ -46,9 +118,33 @@ xi.znm.soultrapper.onItemCheck = function(target, user)
...
+        local data = plate:getSoulPlateData()
+        utils.unused(data)
+        user:messageBasic(xi.msg.basic.SOULTRAPPER_SUCCESS, 0, xi.items.SOUL_PLATE)
+    end
+end
+
+-----------------------------------
+-- onItemUse Helpers
+-----------------------------------
 xi.znm.soultrapper.getZeniValue = function(target, user, item)
```
TiberonKalkaz: At least from 75 era sources, there was a lot of conjecture and partial proof of two more components impacting zeni value,

Mob Size - larger size, more zeni.  Additionally there were therories that the zeni bonuses branched based on mob size.
Instances - Things like Nyzul Isle and Salvage were reported to give better zeni for the same level mobs found outside.  This is why wings has a "beingIrresponsible" zeni modifier that awards a bonus for taking picutues during instanced combat  -when the player should be focused on their objective.

Again, community data is unverified and should be captured

Some sources for Zeni per mob in the 75 era can be found here, as well on differing wiki talk pages
https://www.bluegartr.com/threads/63949-Zeni-per-Image-rates  (has some theories on size based value)
http://ffxiaydel.blogspot.com/2009/03/zeni-point-farming.html
https://wikiwiki.jp/ffxi/NM/%E8%8A%B1%E9%B3%A5%E9%A2%A8%E6%9C%88 (has a table of what they believe is part of zeni calc which includes size)
http://wiki.ffo.jp/html/1155.html ( has more detailed laid out points but did not often fit real world examples like flower nm)

scripts/globals/znm.lua
```diff
@@ -57,86 +153,94 @@ xi.znm.soultrapper.getZeniValue = function(target, user, item)
...
     -- Angle/Facing Component
     if isFacing then
-        zeni = zeni * 1.5
+        zeni = zeni * xi.znm.SOULPLATE_FACING_MULT
     end
 
     -- Bonus for HS Soul Plate
     if user:getEquipID(xi.slot.AMMO) == xi.items.BLANK_HIGH_SPEED_SOUL_PLATE then
-        zeni = zeni * 1.5
+        zeni = zeni * xi.znm.SOULPLATE_HS_MULT
```
TiberonKalkaz: I dont know if this changed from 75 era to now retail - if so, please disregard

I dont see anything that takes into account:
- Mob level -  can take a pic of a level 1 mob as a 75 with no penalty.  In 75 era taking a picture of EP got you garbage points (15ish or worse)
- Mob owner/mob target - in 75 era if you didnt have claim, or the mob was passive or someone else had claim, you got garbage points (5 or less)

scripts/globals/znm.lua
```diff
@@ -57,86 +153,94 @@ xi.znm.soultrapper.getZeniValue = function(target, user, item)
...
+    end
+
+    -- Check for Sanraku's "Subject of Interest"
+    if xi.znm.isCurrentInterest(target:getSuperFamily(), target:getFamily()) then
```
TiberonKalkaz: You cant check check for fauna/interest when the picture is taken.
This must be done at trade in.

There are multiple reports of a player going out to take take pictures of mob X when it was the subOfInterest.
Then trading in 10 plates worth and having the zeni value drop somewhere in the middle of the 10 plates due to change of interest.

TiberonKalkaz:
I assume that the impl on wings has already been shared via another channel but just in case not:
[This was the initial commit on wings to get the ZNM NPC functionality running](https://gitlab.com/ffxiwings/wings/-/merge_requests/1663) on the wings code base (which orginially ported over from LSB by Veyln I think?)

There were more commits over time as the system was tested via beta and live.

scripts/globals/znm.lua
TiberonKalkaz: That makes sense.  When I was trying to wrap my head around zeni calculation factors - Nauls were the bane of my existence.  Once they were found by the player base no one cared about the fine details anymore.  It was Nauls or Nothing™ :joy:

Still, the client side handling of "how good" a picute is a pain and ASB should look into a client side modification.  On wings Sanraku is just perpetually unhappy.  I considered passing zeni*10 to the CS and displaying a custom string to players saying "you actually got 80 zeni, not 800" but in testing - people only read the number and didnt care if the old coot was happy or not.

zach2good:
I've merged this into `feature/znm_improvements` so that other people can pick it up and work on it 👍

PR #2193, opened 2022-07-05
Uragnite mixin adjustments

CriticalXI:
**_I affirm:_**

Adjustments to the Uragnite mixin:
- They will now always use venom shell when forced into their shell with damage.
- Uragnites are forced into their shell by doing a certain % of their HP of damage rather than being a random chance
- They now also can go into and out of their shell on a timer, even when roaming.

All adjustments are from my own time spent poking uragnites on retail.

Watch a uragnite in game to see that they now go into their shell randomly even when out of combat.

### DIFF
--- a/scripts/mixins/families/uragnite.lua
+++ b/scripts/mixins/families/uragnite.lua
@@ -56,15 +56,30 @@ local function exitShell(mob)
     mob:setMobMod(xi.mobMod.NO_MOVE, 0)
 end
 
+local function shellTimer(mob)
+    -- Uragnites go into their shell every 40-60 seconds
+    local timeToShell = mob:getLocalVar("[uragnite]shellTimer")
+
+    -- mob:isMobType check due to Shell We Dance ENM. Those uragnites do not enter their shell except through damage
+    if os.time() > timeToShell and mob:getAnimationSub() == 0 and not mob:isFollowingPath() and not mob:isMobType(xi.mobskills.mobType.BATTLEFIELD) then
+        enterShell(mob)
+
+        local timeInShell = math.random(mob:getLocalVar("[uragnite]timeInShellMin"), mob:getLocalVar("[uragnite]timeInShellMax"))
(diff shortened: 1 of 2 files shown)


scripts/mixins/families/uragnite.lua
```diff
@@ -56,16 +56,31 @@ local function exitShell(mob)
     mob:setMobMod(xi.mobMod.NO_MOVE, 0)
 end
 
+local function shellTimer(mob)
+    -- Uragnites go into their shell every 40-60 seconds
```
claywar, line 60: This may not always be the case.  Anecdotal, but from what I have seen it seems more based on damage taken and/or critical hits.  This from Uragnites in Ceizak.

TeoTwawki: I always thought it was "big hit" like anything that took a large chunk of hp in 1 go be it a crit or a ws to make them go in, then timed for them to come out again.

CriticalXI: It is both. If you see farther down, I still have code that they are forced into their shell from damage. In retail, you can tell the difference between the two, as when they are forced into their shell through damage they react with a venom shell every time. But normal uragnite mobs, even when they're just wandering unclaimed you can watch them go into their shell and out of it.

CriticalXI: I also tested the damage portion, it not a huge hit because I was able to get them forced into their shell by just tickling them with a lvl 1 weapon until I hit 10% health and they would go into their shell.

Xaver-DaRed: From my current test, the timer seems to go from 25s to 50s while engaged and left alone.
Granted, this is a very limited test.

CriticalXI:
I am in the middle of a grind to get things ready for Horizon launch and I don't want these sitting here collecting dust on your repo, so I will go ahead and close this and revisit this at a later time.

PR #2206, opened 2022-07-06
[core] [RUN] Add "modifier" to action, add MB message to Swipe/Lunge, fixup reaction bits

WinterSolstice8:
**_I affirm:_**

Adds a "modifier" to Action to allow modifiers such as"Resist!" "Cover!" , "Immunobreak" swipe/lunge MB in messages to work.
Adds Swipe/Lunge MB message
Fixes up Guard reaction bits and related code to be treated more like SE appears to uses them

New feature:
Switch to SAM/RUN, get 2k tp, use flabra, use sekkanoki, use 2x tachi-jinpu for detonaton, use swipe/lunge and see MB message for swipe/lunge. (does not work on absorb, client is not aware of such a thing but the bit is set)

Regression testing:
Test Zanshin on sam to verify it still works on guard/counter/miss/parry
Test spikes, enspells and verity you only give spike damage when you get hit, and only deal enspell damage when you hit and not get parried
Test parry spikes on RUN with Battuta, verify you only deal parry spikes damage when you parry
Verify Parry/Block/Guard animations still work (mobs don't guard, so test on PLD/MNK yourself)

### DIFF
--- a/scripts/globals/ability.lua
+++ b/scripts/globals/ability.lua
@@ -677,11 +677,21 @@ xi.reaction =
 {
     NONE     = 0x00,
     MISS     = 0x01,
+    GUARDED  = 0x02,
     PARRY    = 0x03,
     BLOCK    = 0x04,
     HIT      = 0x08,
     EVADE    = 0x09,
-    GUARD    = 0x14,
+    ABILITY  = 0x10,
+}
+
+xi.actionModifier =
+{
+    NONE        = 0x00,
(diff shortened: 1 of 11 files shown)


zach2good (approved):

WinterSolstice8:
Force push to fix typo that didn't deserve another commit

Xaver-DaRed (approved):

PR #2207, opened 2022-07-06
change logic for floor objectives, no repeats, NMs can repeat in same run

KnowOne134:
**_I affirm:_**

found in retail that floor objectives will never repeat so builds a list of objectives and excludes the floor that was just completed.
also seen that random NMs can repeat per run and not removed from the list, still not back to back as the mobs do have different lists for odd/even floors


scripts/zones/Nyzul_Isle/instances/nyzul_isle_investigation.lua
```diff
@@ -295,29 +310,30 @@ local function pickMobs(instance)
 
             if spawnedNMs > 0 then
                 local floorSection = math.floor(currentFloor / 20) + 1
+                local floorNMs = {}
+
+                if currentFloor % 2 == 0 then
+                    mobGroup = xi.nyzul.randomNMs.evenFloor[floorSection]
+                else
+                    mobGroup = xi.nyzul.randomNMs.oddFloor[floorSection]
```
Xaver-DaRed, line 319: mobGroup needs local definition
```
local mobGroup = 0
local floorSection = math.floor(currentFloor / 20) + 1
local floorNMs = {}
```
for example.
or just set local this 2.

```
if currentFloor % 2 == 0 then
    mobGroup = xi.nyzul.randomNMs.evenFloor[floorSection]
else
    mobGroup = xi.nyzul.randomNMs.oddFloor[floorSection]
```

Xaver-DaRed (approved):

PR #2216, opened 2022-07-07
Add canUseAbilities() Lua Binding

The-Aerec:
**_I affirm:_**

Adds the canUseAbility lua function to utils. Separating from two current MRs to avoid conflicts

Try this function out!

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -1518,6 +1518,32 @@ uint8 CLuaBaseEntity::getCurrentAction()
 }
 
 /************************************************************************
+ *  Function: canUseAbilities()
+ *  Purpose : Sees if a mob has perticular effects that stop it from doing thing
+ *  Example : mob:canUseAbilities()
+ *  Notes   : Can't use on NPCs
+ ************************************************************************/
+
+bool CLuaBaseEntity::canUseAbilities()
+{
+    if (auto* PEntity = static_cast<CBattleEntity*>(m_PBaseEntity))
+    {
+        return !(PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_SLEEP) ||
+                PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_IMPAIRMENT) ||
+                PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_SLEEP_II) ||
+                PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_STUN) ||
+                PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_AMNESIA) ||
+                PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_LULLABY) ||
+                PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_PETRIFICATION) ||
+                PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_TERROR) ||
+                !(m_PBaseEntity->PAI->CanChangeState()));
+    }
+
+    ShowError("canUseAbilities() : Wrong Entity Type");
+    return false;
+}
+
+/************************************************************************
  *  Function: lookAt()
  *  Purpose : Forces an entity to 'look' at something like it's self-aware
  *  Example : npc:lookAt(player:getPos()) -- Make an NPC look at the PC
@@ -14123,6 +14149,7 @@ void CLuaBaseEntity::Register()
     SOL_REGISTER("getStatus", CLuaBaseEntity::getStatus);
     SOL_REGISTER("setStatus", CLuaBaseEntity::setStatus);
     SOL_REGISTER("getCurrentAction", CLuaBaseEntity::getCurrentAction);
+    SOL_REGISTER("canUseAbilities", CLuaBaseEntity::canUseAbilities);
 
     SOL_REGISTER("lookAt", CLuaBaseEntity::lookAt);
     SOL_REGISTER("clearTargID", CLuaBaseEntity::clearTargID);
--- a/src/map/lua/lua_baseentity.h
+++ b/src/map/lua/lua_baseentity.h
@@ -120,6 +120,7 @@ public:
     uint8 getStatus();
     void  setStatus(uint8 status); // Sets Character's Status
     uint8 getCurrentAction();
+    bool  canUseAbilities();
 
     void lookAt(sol::object const& arg0, sol::object const& arg1, sol::object const& arg2); // look at given position
     void clearTargID();                                                                     // clears target of entity


zach2good (commented):
This looks a whole lot like it should be a binding

WinterSolstice8:
Those checks are very rigid for the states, given that it also ignores stuff like ranged, items, normal weaponskill state, etc. It's not a one size fits all function like it seems like it would be for the outside observer. Perhaps checking `ROAMING or ATTACK`, which are the "idle" and auto attack states.

scripts/globals/utils.lua
```diff
@@ -571,3 +571,25 @@ function utils.lateralTranslateWithOriginRotation(origin, translation)
...
+        act == xi.act.MOBABILITY_USING or
+        act == xi.act.MOBABILITY_FINISH or
+        act == xi.act.MAGIC_START or
+        act == xi.act.MAGIC_CASTING or
+        mob:getStatusEffect(xi.effect.STUN) ~= nil or
+        mob:getStatusEffect(xi.effect.PETRIFICATION) ~= nil or
+        mob:getStatusEffect(xi.effect.TERROR) ~= nil or
+        mob:getStatusEffect(xi.effect.SLEEP_I) ~= nil or
+        mob:getStatusEffect(xi.effect.SLEEP_II) ~= nil or
+        mob:getStatusEffect(xi.effect.AMNESIA) ~= nil or
```
WinterSolstice8: For completeness, you will also want to check `xi.effect.IMPAIRMENT` in case someone decides to call this on a player (or someone who's making a module uses Impairment on a mob for some reason)

WinterSolstice8 (changes requested):

zach2good:
I sat down to look at how this could be turned into a binding, and I think the name `canUseAbility` is kind of misleading. Most other functions we have that look like this take an argument about what you're checking against, rather than checking "am I able to act?".

Consider naming it something else. maybe `canUseAbilities`?

Still think this should be a binding: `if mob:canUseAbilities() then ...`

You can look at the guts of `uint8 CLuaBaseEntity::getCurrentAction()` for how to get started

The-Aerec:
I'll start on converting this now.

The-Aerec:
@zach2good - All Binded up

src/map/lua/lua_baseentity.cpp
```diff
@@ -1517,6 +1517,36 @@ uint8 CLuaBaseEntity::getCurrentAction()
...
+ *  Purpose : Sees if a mob has perticular effects that stop it from doing thing
+ *  Example : mob:canUseAbilities()
+ *  Notes   : Can't use on NPCs
+ ************************************************************************/
+
+bool CLuaBaseEntity::canUseAbilities()
+{
+    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype == TYPE_NPC);
+
+    if (m_PBaseEntity == nullptr)
```
claywar: m_PBaseEntity should never be null in these functions.  @zach2good correct me if I'm wrong, but I believe there is safety in the CLuaBaseEntity class itself.

The-Aerec: Awesome i'll adjust on confirm, I just didn't want to introduce null chances.

src/map/lua/lua_baseentity.cpp
```diff
@@ -1517,6 +1517,36 @@ uint8 CLuaBaseEntity::getCurrentAction()
...
+    {
+        return false;
+    }
+
+    auto* PEntity = static_cast<CBattleEntity*>(m_PBaseEntity);
+    bool actionNotFree = !(m_PBaseEntity->PAI->IsCurrentState<CInactiveState>() || m_PBaseEntity->PAI->IsCurrentState<CRangeState>() || m_PBaseEntity->PAI->IsCurrentState<CAttackState>());
```
WinterSolstice8: This is still ignoring the other states that should prevent an action

After digging a bit it looks like this should be:

`bool actionNotFree = !PEntity->PAI->CanChangeState()`

`!PEntity->PAI->CanChangeState()` returns false during states like job abilities, weaponskills, casting, ranged attacks, etc. It will only return true if the state lets you switch out or your states are completed. It is used in `CPlayerController::Ability`, spell, etc like retail to prevent players from repeatedly using spells or JAs in too short of a window.

WinterSolstice8 (changes requested):

src/map/lua/lua_baseentity.cpp
The-Aerec: This is checking that the mob is either: Auto attacking / Ranged Attacking / Idle.

If its those it returns false (Because we want the double negative for the return statement). if its anything but these states then it will return true making the below return says that its not available to do things.

It also worked successfully when I tested with an print check. Anytime the mob was not those three states or any of the status effects the binding was returning false.

WinterSolstice8: There are more states than just those, such as the item state. The current implementation is not expandable for future states that may block actions. `CanChangeState()` is generic and queries whether or not any state is allowed to be actionable in it and does not add technical debt.

tl;dr: checking that does the same thing you want, but better

The-Aerec: Thanks for the explanation~! Fixed

src/map/lua/lua_baseentity.cpp
The-Aerec: Winter confirmed what you said @claywar - null check removed.

WinterSolstice8 (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -1517,6 +1517,31 @@ uint8 CLuaBaseEntity::getCurrentAction()
...
+
+    return !(PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_SLEEP) ||
+             PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_IMPAIRMENT) ||
+             PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_SLEEP_II) ||
+             PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_STUN) ||
+             PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_AMNESIA) ||
+             PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_LULLABY) ||
+             PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_PETRIFICATION) ||
+             PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_TERROR) ||
+             actionNotFree);
```
zach2good: Something like this: (XI_DEBUG_BREAK_IF brings down servers)

```cpp
bool CLuaBaseEntity::canUseAbilities()
{
    if (auto* PEntity = dynamic_cast<CBattleEntity*>(m_PBaseEntity))
    {
        return !(PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_SLEEP) ||
            PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_IMPAIRMENT) ||
            PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_SLEEP_II) ||
            PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_STUN) ||
            PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_AMNESIA) ||
            PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_LULLABY) ||
            PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_PETRIFICATION) ||
            PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_TERROR) ||
            !(m_PBaseEntity->PAI->CanChangeState());
    }

    // Log about using the wrong entity type
    reuturn false;
}

File: src/map/lua/lua_baseentity.cpp [hunk 5]
```diff
@@ -1517,6 +1517,33 @@ uint8 CLuaBaseEntity::getCurrentAction()
...
+/************************************************************************
+ *  Function: canUseAbilities()
+ *  Purpose : Sees if a mob has perticular effects that stop it from doing thing
+ *  Example : mob:canUseAbilities()
+ *  Notes   : Can't use on NPCs
+ ************************************************************************/
+
+bool CLuaBaseEntity::canUseAbilities()
+{
+
```
zach2good: Whitespace

zach2good (approved):

PR #2223, opened 2022-07-08
Aragoneu Region NM Audit

CriticalXI:
**_I affirm:_**

- Blighting Brand added missing TP move Whispers of Ire. Two versions: Drain attributes from player, or target self to wipe debuffs.
- Citipati can only spawn during night.
- Xolotl rework, now spawns pets and can only spawn at night.
- Roc now has Benediction and Draw In.
- ??? respawn to spawn Feeler Antlion (Rostrum Pumps) is now based off of Alastor Antlion's death and not the Feeler Antlion.
- Several minor trait adjustments to other NMs.


scripts/globals/mobskills/whispers_of_ire.lua
```diff
@@ -0,0 +1,40 @@
...
+
+mobskill_object.onMobWeaponSkill = function(target, mob, skill)
+    local amount = math.random(2, 5)
+    local count = 0
+
+    for i = 1, amount do
+        -- str down - chr down
+        local effectType = math.random(136, 142)
+
+        skill:setMsg(xi.mobskills.mobDrainAttribute(mob, target, effectType, 10, 3, 60))
```
TeoTwawki, line 26: I don't think we should set message inside the for loop. Its set below anyway 100% of the time, so you can just do `xi.mobskills.mobDrainAttribute()` here without placing it inside `setMsd()`

scripts/globals/mobskills/whispers_of_ire.lua
```diff
@@ -0,0 +1,40 @@
...
+    return 0
+end
+
+mobskill_object.onMobWeaponSkill = function(target, mob, skill)
+    local amount = math.random(2, 5)
+    local count = 0
+
+    for i = 1, amount do
+        -- str down - chr down
+        local effectType = math.random(136, 142)
```
TeoTwawki, line 24: We could make this a bit more human readable using the actual effect enum instead of raw numbers+comment.

scripts/globals/mobskills/whispers_of_ire_self.lua
```diff
@@ -0,0 +1,32 @@
...
+require("scripts/globals/settings")
+require("scripts/globals/status")
+require("scripts/globals/msg")
+-----------------------------------
+local mobskill_object = {}
+
+mobskill_object.onMobSkillCheck = function(target, mob, skill)
+    local dispel = target:eraseStatusEffect()
+
+    if dispel ~= xi.effect.NONE then
```
TeoTwawki, line 16: it doesn't even try if it didn't succeed in erasing at least one effect? and by doing this in the skillcheck, its already erasing one status before the mob even finishes. If someone stuns it, one effect still got erased.

scripts/zones/Attohwa_Chasm/Zone.lua
```diff
@@ -85,6 +85,21 @@ zone_object.onGameHour = function(zone)
         starting at ID.npc.MIASMA_OFFSET. some are supposed to toggle open, but need retail test
         to determine which.  for now, they're just statically set per npc_list.animation
     --]]
+
+    -- Don't allow Citipati or Xolotl to spawn outside of night
```
TeoTwawki, line 89: suggestion: 
`-- Don't allow Citipati or Xolotl to spawn outside of night if unclaimed`

if he's claimed he'll stay up until killed or claim is lost. While I believe your code as written allows for that (as disallow doesn't depop mid fight iirc), it's worth mentioning in the comment.

scripts/zones/Attohwa_Chasm/mobs/Xolotl.lua
```diff
@@ -3,20 +3,90 @@
...
+    local timeInterval = mob:getBattleTime() % 6
+
+    -- Spawn skeleton pets
+    for i = 1, 2 do
+        local child = GetMobByID(mob:getID() + i)
+        if child:isSpawned() then
+            if target and child:getCurrentAction() == xi.act.ROAMING then -- doing nothing, make share enmity
+                child:updateEnmity(target)
+            end
+        elseif mob:getCurrentAction() ~= 30 and mob:actionQueueEmpty() and timeInterval == (i-1)*3 then -- not spawned, not casting, not using an ability and should summon
```
TeoTwawki, line 24: other devs, I summon thee. Loops inside onMobFight give me a migraine. One of y'all check it out and I'll trust you.

scripts/zones/Attohwa_Chasm/mobs/Xolotls_Sacrifice.lua
```diff
@@ -0,0 +1,21 @@
...
+-- Note: Pet for Xolotl
+-----------------------------------
+require("scripts/globals/world")
+-----------------------------------
+local entity = {}
+
+entity.onMobRoam = function(mob)
+    local totd = VanadielTOTD()
+
+    if totd ~= xi.time.NIGHT and totd ~= xi.time.MIDNIGHT then
```
TeoTwawki, line 13: check if it or its master are actively in a battle too. Have held him in daytime. Claim loss (including via charm) results in depop.

scripts/zones/Meriphataud_Mountains/mobs/Coo_Keja_the_Unseen.lua
```diff
@@ -2,17 +2,55 @@
...
     })
+
+    -- Takes half damage from all attacks
+    mob:addMod(xi.mod.UDMGPHYS,-5000)
+    mob:addMod(xi.mod.UDMGRANGE,-5000)
+    mob:addMod(xi.mod.UDMGMAGIC,-5000)
+    mob:addMod(xi.mod.UDMGBREATH,-5000)
+
+    -- May spawn in a party with two other Orcs
+    if math.random(1,2) == 1 then
```
TeoTwawki, line 27: `if math.random(3) == 2 then` like you agreed to with the other ones of this type :+1:

sql/mob_droplist.sql
```diff
@@ -17777,14 +17776,14 @@ INSERT INTO `mob_droplist` VALUES (2202,0,0,1000,19213,@UNCOMMON); -- White Tath
 INSERT INTO `mob_droplist` VALUES (2202,0,0,1000,901,20);          -- Venomous Claw (2.0%)
 
 -- ZoneID: 200 - Serket
-INSERT INTO `mob_droplist` VALUES (2203,0,0,1000,4172,@ALWAYS); -- Reraiser (Always, 100%)
-INSERT INTO `mob_droplist` VALUES (2203,0,0,1000,4173,990);     -- Hi-Reraiser (99.0%)
-INSERT INTO `mob_droplist` VALUES (2203,0,0,1000,901,750);      -- Venomous Claw (75.0%)
```
TeoTwawki, line 17782: whoa I dunno how nobody caught that before this.

TeoTwawki (changes requested):
My apologies that it was so long before a review got here for this one.

scripts/globals/mobskills/whispers_of_ire.lua
jmcmorris: Could be this:
```suggestion
        local effectType = math.random(xi.effect.STR_DOWN, xi.effect.CHR_DOWN)
```

Should this be applying the same effect multiple times? If not then this should be randomly selecting from a list and removing them as they're selected.

scripts/globals/mobskills/whispers_of_ire_self.lua
jmcmorris: I believe you can use this instead `mob:hasStatusEffectByFlag(xi.effectFlag.ERASABLE)`

scripts/zones/Attohwa_Chasm/Zone.lua
```diff
@@ -85,6 +85,21 @@ zone_object.onGameHour = function(zone)
...
+    local xolre = GetMobByID(ID.mob.XOLOTL):getLocalVar("xolotlRespawn")
+
+    if VanadielHour() >= 4 and VanadielHour() < 20 then
```
jmcmorris, line 92: Can we use mob_groups `spawntype` instead? I believe `SPAWNTYPE_ATNIGHT` can be set to only allow it to spawn during night.

scripts/zones/Attohwa_Chasm/mobs/Citipati.lua
```diff
@@ -6,6 +6,22 @@ require("scripts/globals/hunts")
 -----------------------------------
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:setMod(xi.mod.TRIPLE_ATTACK, 5)
+end
+
+entity.onMobRoam = function(mob)
```
jmcmorris, line 13: If we can use `spawntype` then we don't need this code!

scripts/zones/Attohwa_Chasm/mobs/Xolotl.lua
jmcmorris: Could use `xi.action.MAGIC_CASTING` instead of `30` but besides that it looks good

scripts/zones/Attohwa_Chasm/Zone.lua
TeoTwawki: I might be wrong, but I believe the nature of its other mechanics precludes the use of the spawn type that would normally handle it and would cause it to then pop every night. worth double checking.

CriticalXI:
I am in the middle of a grind to get things ready for Horizon launch and I don't want these sitting here collecting dust on your repo, so I will go ahead and close this and revisit this at a later time.

PR #2224, opened 2022-07-08
Kolshushu Region NM Audit

CriticalXI:
**_I affirm:_**

- Shen rework. Now properly going into shell on a timer, and pet behavior introduced.
- Added the following Manaclipper NMs: Cyclopean Conch, Harajnite, and Zoredonite.
- Several minor trait adjustments to other NMs.


scripts/zones/Bibiki_Bay/mobs/Shen.lua
```diff
@@ -3,24 +3,111 @@
...
 local ID = require("scripts/zones/Bibiki_Bay/IDs")
-require("scripts/mixins/families/uragnite")
 -----------------------------------
 local entity = {}
 
+local function enterShell(mob)
+    mob:setAnimationSub(1)
+    mob:SetAutoAttackEnabled(false)
+    mob:SetMagicCastingEnabled(false)
+    mob:setMod(xi.mod.UDMGPHYS, -8500)
```
TeoTwawki, line 13: mmmkkkaaaay. This is the [2nd time](https://github.com/LandSandBoat/server/pull/2158#issuecomment-1173310670) recently I've seen this done with slightly diff values from the mixin. I really feel we should adjust the mixin to allow for some parameters to control it, rather than having the same code with diff input in multiple places.

sql/mob_spawn_mods.sql
```diff
@@ -608,6 +608,12 @@ INSERT INTO `mob_spawn_mods` VALUES (17588703,55,180,1);
 INSERT INTO `mob_spawn_mods` VALUES (17428677,1,18000,1); -- min gil 18000
 INSERT INTO `mob_spawn_mods` VALUES (17428677,2,30000,1); -- max gil 30000
 
+-- [Narasimha]
+-- Values taken from wiki on 04/23/2018 (http://ffxiclopedia.wikia.com/wiki/Narasimha)
```
TeoTwawki, line 613: remembered the gil drop mug gil AND cites sources! Is it Christmas?!

scripts/zones/Bibiki_Bay/mobs/Shen.lua
CriticalXI: I guess I personally found it cleaner to just put the logic in Shen's code so I don't have to clog up the mixin with several checks to see if its a certain mob or not. But I can take a look into that.

TeoTwawki: we wouldn't want check if its a particular mob, we'd want input for how potent the dmg reductions will be if diff from defaults. other stuff would still be separate if needed, but reduces down to just checking animation sub

sql/mob_spell_lists.sql
```diff
@@ -4929,7 +4929,6 @@ INSERT INTO `mob_spell_lists` VALUES ('Hakutaku',486,249,1,255); -- blaze_spikes
 -- Shen (487)
 INSERT INTO `mob_spell_lists` VALUES ('Shen',487,172,1,255); -- water_iv (1~255)
 INSERT INTO `mob_spell_lists` VALUES ('Shen',487,201,1,255); -- waterga_iii (1~255)
-INSERT INTO `mob_spell_lists` VALUES ('Shen',487,214,1,255); -- flood (1~255)
```
Mortalelite, line 4934: I thought Shen did cast Flood

https://www.youtube.com/watch?v=L76e2twvwz4 @3:14 shows starts casting from what I found

https://ffxiclopedia.fandom.com/wiki/Shen also shows cast it

CriticalXI: In my research watching videos and getting a capture of Shen, he only casts Flood to summon his water elementals. If you see my Shen.lua code, he instant casts Flood so that he can spawn a pet at the location of the player that he cast flood on. Also that video you linked happens to be my old FFXI retail LS :D

Mortalelite: Ah ok. I see. No worries, was just asking since was recently added and then removed.

scripts/zones/Bibiki_Bay/mobs/Shen.lua
CriticalXI: So I took another look at this, and I still feel like it would be best to not utilize the mixin for Shen. Shen does not follow the same behavior the mixin uses at all other than the enter/exit shell function, which I need to call directly in Shen's lua to behave properly.

TeoTwawki (approved):
While I would still like the urganite mixin to be made more flexible via parameters that is not requisite for this PR to pass.

At its heart, the mixin applies damage taken mods and toggles some behaviors, with some logic for when to change mods that doesn't apply to Shen. All of this and more can be parameterized the same way the mob 2hr mixin handles multiple behaviors - so while Shen may not act exactly the same, there is no reason the parts he does differently / doesn't do at all can't be controlled by params and variables in the future.

```lua
xi.mix.uragnite.config(mob, {
    chanceToShell = 10,
    timeInShellMin = 45,
    timeInShellMin = 60,
})
```
Any param can be omitted, and we can define more for alternate behaviors. I hope to see future work in this direction.

PR #2226, opened 2022-07-08
[core] [packets] Adjust incoming login packet logic to be more safe

### LINKED ISSUE #2157: 🐛 When character is created, the first attempt at zoning fails

## Branch affected by issue 
`base`

## Steps to reproduce
1. Create a new character, give it GM status before it arrives in game worked for the 1st time (I typically do this at the screen the progress bar fills up and says "registered" before hitting ok to get into game)
2. Notice the log giving an invalid key message in the console log, but you do arrive in game properly.
3. once in game (for me this was bastok mines) try to use `!goto <me>` to warp to yourself. Nothing happens. No error, no nothing. Doesn't respond at all. Now try and walk out of town. You will arrive back at your starting point (default homepoint).
4. Try to zone again. works normally. Try the same `!goto <me>` command again, you get zoned and arrive back in same location as expected.
5. Logging out and back in zoning will work as normal after. This ONLY happens if you try to immediately zone after the starting cs movie completes.

## Expected behavior
Zonage should work correctly the 1st time.

WinterSolstice8:
**_I affirm:_**

Adjusts zoning/login logic to be more safe -- specifically it allows for the client to send repeat 0x00A and 0x00C without errors. 
Fixes #2157

Change zones/login and have it work
~~Change zones/login into mog house with mannequin and have it work~~ edit: mannequins aren't implemented?
Breakpoint 0x00D/0x00A, attempt zone/login, wait several seconds, remove breakpoint, continue. Have no errors and login/zone successfully.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -241,10 +241,16 @@ void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PCh
     TracyZoneScoped;
     data.ref<uint32>(0x5C) = 0;
 
-    PChar->clearPacketList();
-
-    if (PChar->status == STATUS_TYPE::DISAPPEAR)
+    if (PSession->blowfish.status == BLOWFISH_ACCEPTED && PChar->status == STATUS_TYPE::NORMAL) // Do nothing if character is zoned in
     {
+        ShowWarning("packet_system::SmallPacket0x00A player '%' attempting to send 0x00A when already logged in", PChar->GetName());
+        return;
+    }
+
+    if (PSession->blowfish.status == BLOWFISH_WAITING) // Generate new blowfish session, call zone in, etc, only once.
+    {
+        PChar->clearPacketList();
+
         if (PChar->loc.zone != nullptr)
         {
             // Remove the char from previous zone, and unset shuttingDown (already in next zone)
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
Marked as draft until this can be tested by more than just myself out of safety precautions.

TeoTwawki:
I'll help test, I was working on a problem in same area (where I had stuffed that homepoint line as a temp measure)

src/map/packet_system.cpp
```diff
@@ -454,6 +449,12 @@ void SmallPacket0x00C(map_session_data_t* const PSession, CCharEntity* const PCh
 void SmallPacket0x00D(map_session_data_t* const PSession, CCharEntity* const PChar, CBasicPacket data)
 {
     TracyZoneScoped;
+
+    if (PChar->status == STATUS_TYPE::DISAPPEAR && (PSession->blowfish.status == BLOWFISH_WAITING || PSession->blowfish.status == BLOWFISH_SENT)) // Character has already requested to zone, do nothing.
```
zach2good, line 453: 👍

zach2good (approved):
I'll mess with this in the next few days and see if I can break it, otherwise, looks pretty good

zach2good:
Tested a little on my lunch break, looks good 👍

WinterSolstice8:
Alright, draft status removed

TeoTwawki (approved):

src/map/packet_system.cpp
```diff
@@ -313,28 +319,20 @@ void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PCh
                 PChar->loc.zoning = true;
             }
         }
-        PChar->status = STATUS_TYPE::NORMAL;
-    }
-    else if (PChar->loc.zone != nullptr)
-    {
-        // TODO: this should only happen ONCE, instead of spamming the log dozens of times..
```
TeoTwawki, line 320: death to log spam and overzealous homepointing

TeoTwawki: next stop, figuring out why intro cs was tripping this on new chars

TeoTwawki:
> I'll help test, I was working on a problem in same area (where I had stuffed that homepoint line as a temp measure)

I had forgotten yesterday, but saw no adverse effect this morning

src/map/packet_system.cpp
WinterSolstice8: the client was probably sending another 0x0A for some reason, despite already being zoned in. I don't think it's necessary for us but SE's backend might be doing something different during character creation (perhaps their new characters aren't visible until they finish the CS?)

TeoTwawki: > (perhaps their new characters aren't visible until they finish the CS?)

it either doesn't actually show the player in game yet or they are a statue there till their cs ends, I'd need to watch from a second account to really know for sure.

I know -other- players shouldn't be visible during your cs, and I am unsure how that is implemented on retail - on private servers you can see people running past

WinterSolstice8: >I know -other- players shouldn't be visible during your cs, and I am unsure how that is implemented on retail - on private servers you can see people running past

I think we could add a flag for "new player" and not set/send the update PCs flag to accomplish this. We want NPCs, but not players.

TeoTwawki: > I think we could add a flag for "new player" and not set/send the update PCs flag to accomplish this. We want NPCs, but not players.

May possibly be some other cs that do this as well (don't show ppl on retail, but do on private server). As far as a flag, we do have one that we aren't checking we could potentially use.

claywar (approved):

PR #2229, opened 2022-07-08
[core] Fishing Heal Fix and Spelling Corrections

The-Aerec:
Co-Authored-By: Setzor <setzor@gmail.com>

**_I affirm:_**

Fixes the CalculateHeal function to properly calculate how much a fish heals on player mistake during the fishing minigame. 
(Fixes Spellings of that function as well)

Fish 
reduce fish stamina some 
purposely mess up the minigame
see fish heal

### DIFF
--- a/src/map/utils/fishingutils.cpp
+++ b/src/map/utils/fishingutils.cpp
@@ -376,14 +376,14 @@ namespace fishingutils
     {
         uint8 bonusAdd = (legendary) ? rod->lgdBonusAtk : 0;
 
-        return (uint16)std::floor(difficulty * ((rod->fishAttack + bonusAdd) / 100)) * 20;
+        return (uint16)std::floor(difficulty * (((static_cast<float>(rod->fishAttack) + bonusAdd) / 100.0f)) * 20.0f);
     }
 
-    uint16 CaculateHeal(bool legendary, uint8 difficulty, rod_t* rod)
+    uint16 CalculateHeal(bool legendary, uint8 difficulty, rod_t* rod)
     {
         uint16 attack = CalculateAttack(legendary, difficulty, rod);
 
-        return (uint16)std::floor((attack / 20) * (rod->fishRecovery / 100)) * 10;
+        return (uint16)std::floor((static_cast<float>(attack) / 20.0f) * (static_cast<float>(rod->fishRecovery) / 100.0f)) * 10.0f;
     }
 
     uint8 CalculateRegen(uint8 fishingSkill, rod_t* rod, FISHINGCATCHTYPE catchType, uint8 sizeType, uint8 catchSkill, bool legendaryCatch, bool NM)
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):

src/map/utils/fishingutils.cpp
```diff
@@ -379,11 +379,11 @@ namespace fishingutils
         return (uint16)std::floor(difficulty * ((rod->fishAttack + bonusAdd) / 100)) * 20;
     }
 
-    uint16 CaculateHeal(bool legendary, uint8 difficulty, rod_t* rod)
+    uint16 CalculateHeal(bool legendary, uint8 difficulty, rod_t* rod)
     {
         uint16 attack = CalculateAttack(legendary, difficulty, rod);
 
-        return (uint16)std::floor((attack / 20) * (rod->fishRecovery / 100)) * 10;
+        return (uint16)std::floor((attack / 20) * ((float)rod->fishRecovery / 100)) * 10;
```
zach2good: You probably want to make all the terms that are doing division or multiplication into floats, so that nothing is getting truncated.

The attack portion of this will still be getting truncated.

https://godbolt.org/z/KEnxc1nnh

zach2good (changes requested):

src/map/utils/fishingutils.cpp
The-Aerec: Roger, now that i'm looking everywhere else standby as I change a few more things.

The-Aerec: Addressed

Xaver-DaRed: Lies

zach2good: If you're going to do conversions between numeric types, please use `static_cast` like I showed in the godbolt example.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast

The-Aerec: > Lies

What? (Why comment this?)

The-Aerec: Pushed with static casts.

zach2good:
```cpp
    uint16 CalculateHeal(bool legendary, uint8 difficulty, rod_t* rod)
    {
        auto attack = CalculateAttack(legendary, difficulty, rod);
        return std::floor<uint16>((static_cast<float>(attack) / 20.0f) * (static_cast<float>(rod->fishRecovery) / 100.0f)) * 10.0f;
    }
```

The-Aerec: 👍

Xaver-DaRed: > > Lies
> 
> What? (Why comment this?)

When that comment was done, the conversation was closed as resolved with a "Addressed" comment on your part, but it really wasn't.

The-Aerec: >When that comment was done, the conversation was closed as resolved with a "Addressed" comment on your part, but it really wasn't.

Then can you comment with what needed to be further addressed? I'm still very basically learning cpp, so instead of commenting something that was completely useless to anyone, you could of instead helped me understand the issue.

zach2good (approved):

PR #2236, opened 2022-07-09
Draw In Adjustments

CriticalXI:
**_I affirm:_**

- Draw In now defaults to drawing the player in to the center of the mob's hit box.
- Bound or stationary mobs will now draw in the moment you leave their attack range.
- Adds new lua binding to force a mob to draw in on command. Primarily used for King Vinegaroon and the ENM Pulling the Plug.
- Added several new mobmods to alter draw in behavior:
   - DRAW_IN_FRONT: Adds the ability to still have draw in place the player slightly in front of the mob (HNMs usually follow this)
   - DRAW_IN_INCLUDE_PARTY: Adjusted draw in to have a new mod to draw in player/alliance
   - DRAW_IN_CUSTOM_RANGE: Allows you to set a custom minimum draw in range.
   - DRAW_IN_MAXIMUM_REACH: Allows you to set a custom maximum draw in range.
   - DRAW_IN_IGNORE_STATIONARY: Mob mod to remove the bound/stationary draw in behavior

A lot of this work is primarily bringing over work done by other devs on Wings to flesh out and correct the draw in system to be more accurate and flexible as draw in behaves differently all over the game.

I will be following up this PR with another of me adding these mob mods to the appropriate monsters.

Place the mob mods on various monsters to see that they work as intended.

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -2443,6 +2443,11 @@ xi.mobMod =
     ALLI_HATE           = 68, -- Range around target to add alliance member to enmity list.
     NO_LINK             = 69, -- If set, mob cannot link until unset.
     NO_REST             = 70, -- Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
+    DRAW_IN_INCLUDE_PARTY     = 71, -- this will cause the mob's draw-in to also affect all party and alliance members
+    DRAW_IN_FRONT             = 72, -- Mob will draw in slightly in front of them instead of the center of their hitbox
+    DRAW_IN_CUSTOM_RANGE      = 73, -- override the default range of MeleeRange*2 of when players start to get drawn-in
+    DRAW_IN_MAXIMUM_REACH     = 74, -- players further than this range (yalms) will be unaffected by the draw-in. default (0) is whole zone
+    DRAW_IN_IGNORE_STATIONARY = 75, -- stationary or bound mobs draw-in the moment they cannot attack you anymore (out of range). Place this mob mod to stop that behavior.
 }
 
 -----------------------------------
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -13552,6 +13552,57 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
 }
 
 /************************************************************************
+ *  Function: triggerDrawIn()
+ *  Purpose : Forces a mob to use DrawIn on the mob's current target
+ *  Example : mob:triggerDrawIn(bool includeParty, CLuaBaseEntity* PEntity)
+ *  Note    : Params can assume a default value by passing nil
+ *          : e.g. triggerDrawIn(true) to pull in a party/alliance
+ ************************************************************************/
+inline int32 CLuaBaseEntity::triggerDrawIn(CLuaBaseEntity* PMobEntity, sol::object const& includePt, sol::object const& drawRange, sol::object const& maxReach, sol::object const& target)
+{
+    XI_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
+    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
+
+    auto*          PMob    = static_cast<CMobEntity*>(PMobEntity->m_PBaseEntity);
+    CBattleEntity* PTarget = PMob->GetBattleTarget();
+
+    // Default values
+    uint8  drawInRange  = PMob->GetMeleeRange() * 2;
+    uint16 maximumReach = 0xFFFF;
+    bool   includeParty = false;
+    float  offset       = PMob->GetMeleeRange() - 0.2f;
+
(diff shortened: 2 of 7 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -13551,6 +13551,57 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
...
 }
 
+/************************************************************************
+ *  Function: triggerDrawIn()
+ *  Purpose : Forces a mob to use DrawIn on the mob's current target
+ *  Example : mob:triggerDrawIn(bool includeParty, CLuaBaseEntity* PEntity)
+ *  Note    : Params can assume a default value by passing nil
+ *          : e.g. triggerDrawIn(true) to pull in a party/alliance
+ ************************************************************************/
+inline int32 CLuaBaseEntity::triggerDrawIn(CLuaBaseEntity* PMobEntity, sol::object const& includePt, sol::object const& drawRange, sol::object const& maxReach, sol::object const& target)
```
claywar, line 13561: No reason to inline any of these functions, as the compiler is smart enough.  Also, given that there is only one retval, this should be a void.

src/map/lua/lua_baseentity.cpp
```diff
@@ -13551,6 +13551,57 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
...
+{
+    XI_DEBUG_BREAK_IF(m_PBaseEntity == nullptr);
```
claywar, line 13563: m_PBaseEntity is guaranteed not null. This is unnecessary

src/map/lua/lua_baseentity.cpp
```diff
@@ -13551,6 +13551,57 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
...
+    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
```
claywar, line 13564: Should be a sane return with logging.
```
if (m_PBaseEntity->objtype != TYPE_MOB)
{
    ShowDebug("Something here, I'm not a mob!);
}
```

src/map/lua/lua_baseentity.cpp
```diff
@@ -13551,6 +13551,57 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
...
+    auto*          PMob    = static_cast<CMobEntity*>(PMobEntity->m_PBaseEntity);
+    CBattleEntity* PTarget = PMob->GetBattleTarget();
+
+    // Default values
+    uint8  drawInRange  = PMob->GetMeleeRange() * 2;
+    uint16 maximumReach = 0xFFFF;
+    bool   includeParty = false;
+    float  offset       = PMob->GetMeleeRange() - 0.2f;
+
+    if ((drawRange != sol::lua_nil) && drawRange.is<uint8>())
```
claywar, line 13575: Would probably be cleaner to use ternaries here and for the below blocks

src/map/utils/battleutils.cpp
```diff
@@ -5762,51 +5767,53 @@ namespace battleutils
...
 
         bool  success        = false;
-        float drawInDistance = (float)(PMob->getMobMod(MOBMOD_DRAW_IN) > 1 ? PMob->getMobMod(MOBMOD_DRAW_IN) : PMob->GetMeleeRange() * 2);
+        // float drawInDistance = (float)(PMob->getMobMod(MOBMOD_DRAW_IN) > 1 ? PMob->getMobMod(MOBMOD_DRAW_IN) : PMob->GetMeleeRange() * 2);
 
-        if (std::chrono::time_point_cast<std::chrono::seconds>(server_clock::now()).time_since_epoch().count() - PMob->GetLocalVar("DrawInTime") < 2)
-        {
-            return false;
-        }
+        // if (std::chrono::time_point_cast<std::chrono::seconds>(server_clock::now()).time_since_epoch().count() - PMob->GetLocalVar("DrawInTime") < 2)
```
CriticalXI, line 5772: Todo for myself to clean up this section

scripts/globals/status.lua
```diff
@@ -2443,6 +2443,11 @@ xi.mobMod =
     ALLI_HATE           = 68, -- Range around target to add alliance member to enmity list.
     NO_LINK             = 69, -- If set, mob cannot link until unset.
     NO_REST             = 70, -- Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
+    DRAW_IN_INCLUDE_PARTY     = 71, -- this will cause the mob's draw-in to also affect all party and alliance members
```
TeoTwawki, line 2446: We want to take a different direction than this. we're planning to move the bulk of draw in to lua via a combination of listener, mixin, and binding. While I appreciate the binding using multiple mob mods to control its behavior is cumbersome. (by it I mean draw in, not the binding - we'll take that binding :+1: )

We're going to handle the behaviors controls on our end and current plan is to use parameters on a mixin to set behavior. Kinda like how 2hrs get controlled by params.

PR #2243, opened 2022-07-09
[cmake] Force download and build LuaJIT on Unix

### LINKED ISSUE #2238: 🐛 LuaJIT forced external unwinding causes crashes on x86_64 linux only

## Branch affected by issue 

`base`

## Steps to reproduce

Build on x86_64 in linux (WSL or bare metal is known to be effected)
Add "delmodnil.lua" to scripts\commands, contents as follows:
```
----------------------------------
-- func: delmodnil
-- desc: Tests passing nil to delMod to see if it crashes linux system

require("scripts/globals/status")

cmdprops =
{
    permission = 1,
    parameters = ""
}


function onTrigger(player)
    player:delMod(xi.mod.REGEN, nil)
end
```

run !delmodnil as GM and you will crash xi_map

## Expected behavior

error message as follows:
```
[07/08/22 23:33:16:322][map][error] cmdhandler::call: (delmodnil) error: stack index 3, expected number, received nil: not a numeric type (bad argument into 'void(unsigned short, short)')
stack traceback:
        [C]: in function 'delMod'
        ./scripts/commands/delmodnil.lua:16: in function <./scripts/commands/delmodnil.lua:15> (CCommandHandler::call:321)
```
        

## Related issues
Root cause of #2115 
Might have been the root cause of #2116

## Known resolution, but not ideal fix

Pull source tree (latest commit) of https://github.com/LuaJIT/LuaJIT (may not need latest src)

edit `src/Makefile` and add `-DLUAJIT_NO_UNWIND` to `CCOPT=`
compile, `sudo make install`
symlink the newly built libluajit.so to the same location as xi_map
export LD_LIBRARY_PATH=`working directory of xi_map`
xi_map can now run the previous command without error

## Research into problem

This is the function that had a differing flow in aarch64 vs x86_64 (taken from gdb)

Source is here: https://github.com/LuaJIT/LuaJIT/blob/8271c643c21d1b2f344e339f559f2de6f3663191/src/lj_err.c#L507
```
/* Throw error. Find catch frame, unwind stack and continue. */
507 LJ_NOINLINE void LJ_FASTCALL lj_err_throw(lua_State *L, int errcode)
508 {
509   global_State *g = G(L);
510   lj_trace_abort(g);
511   setmref(g->jit_base, NULL);
512   L->status = LUA_OK;
513 #if LJ_UNWIND_EXT
514   err_raise_ext(errcode);
515   /*
516   ** A return from this function signals a corrupt C stack that cannot be
517   ** unwound. We have no choice but to call the panic function and exit.
518   **
519   ** Usually this is caused by a C function without unwind information.
520   ** This should never happen on x64, but may happen if you've manually
521   ** enabled LUAJIT_UNWIND_EXTERNAL and forgot to recompile *every*
522   ** non-C++ file with -funwind-tables.
523   */
524   if (G(L)->panic)
525     G(L)->panic(L);
526 #else
527   {
528     void *cf = err_unwind(L, NULL, errcode);
529     if (cframe_unwind_ff(cf))
530       lj_vm_unwind_ff(cframe_raw(cf));
531     else
532       lj_vm_unwind_c(cframe_raw(cf), errcode);
533   }
534 #endif
535   exit(EXIT_FAILURE);
536 }
```

On aarch64 the code flows like this:
```
512       L->status = LUA_OK;
(gdb) step
528         void *cf = err_unwind(L, NULL, errcode);
```

On x86_64 the code flows like this:

```
512       L->status = LUA_OK;
(gdb) step
514       err_raise_ext(errcode);
```

This is because `LJ_UNWIND_EXT` is defined in x86_64 -- see https://github.com/LuaJIT/LuaJIT/blob/8271c643c21d1b2f344e339f559f2de6f3663191/src/lj_err.c#L64 :

```
#if defined(__GNUC__) && (LJ_TARGET_X64 || defined(LUAJIT_UNWIND_EXTERNAL)) && !LJ_NO_UNWIND
#define LJ_UNWIND_EXT	1
#elif LJ_TARGET_WINDOWS
#define LJ_UNWIND_EXT	1
#endif
```

Note the `LJ_TARGET_X64` -- this is forced on. See the previous comment a few lines above `EXT is the default on x64 systems and on Windows` 
The documentation has since been changed since that file was updated, you can read some history here:
https://github.com/LuaJIT/LuaJIT/blob/v2.1/src/lj_err.c#L78

It seems that perhaps not all libs on ubuntu are built with `-funwind-tables` or `-fexceptions`, causing the external unwinding to fail. The documentation in the updated lj_err.c seems to hint that this might be the case:

```
** - EXT requires unwind tables for *all* functions on the C stack between
**   the pcall/catch and the error/throw. C modules used by Lua code can
**   throw errors, so these need to have unwind tables, too. Transitively
**   this applies to all system libraries used by C modules -- at least
**   when they have callbacks which may throw an error.
```

If we define `LJ_NO_UNWIND`, the behavior will be consistent on all *nix systems.

## Longterm fix

@zach2good will hook up LuaJIT to auto build with the `LJ_NO_UNWIND` defined and link in cmake (this will add maybe 30s of compile time)

zach2good:
**_I affirm:_**

Fixes https://github.com/LandSandBoat/server/issues/2238
(I hope)

Build on Linux, you should see LuaJIT get downloaded and built (doesn't take long)
Everything should start as normal, that crash should be gone.

### DIFF
--- a/cmake/FindLuaJIT.cmake
+++ b/cmake/FindLuaJIT.cmake
@@ -22,10 +22,31 @@
 #
 # TEST AND MAKE SURE THAT EVERYTHING STILL WORKS!
 
+if (UNIX)
+    message(STATUS "Downloading LuaJIT src")
+    CPMAddPackage(
+        NAME LuaJIT
+        GITHUB_REPOSITORY LuaJIT/LuaJIT
+        GIT_TAG e3bae12fc0461cfa7e4bef3dfed2dad372e5da8d
+        DOWNLOAD_ONLY YES
+    )
+    if (LuaJIT_ADDED)
+        message(STATUS "Modifying LuaJIT build flags (adding -DLUAJIT_NO_UNWIND=1)")
+        file(READ "${LuaJIT_SOURCE_DIR}/src/Makefile" FILE_CONTENTS)
+        string(REPLACE "CCOPT= -O2 -fomit-frame-pointer\n" "CCOPT= -O2 -fomit-frame-pointer -DLUAJIT_NO_UNWIND=1\n" FILE_CONTENTS "${FILE_CONTENTS}")
+        file(WRITE "${LuaJIT_SOURCE_DIR}/src/Makefile" "${FILE_CONTENTS}")
+
+        # LuaJIT has no CMake support, so we break out to using make on it's own
+        message(STATUS "Building LuaJIT from src")
+        execute_process(COMMAND make WORKING_DIRECTORY ${LuaJIT_SOURCE_DIR})
+    endif()
+endif()
+
 find_library(LuaJIT_LIBRARY
     NAMES
         luajit luajit_64 luajit-5.1 libluajit libluajit_64
     PATHS
+        ${LuaJIT_SOURCE_DIR}/src/
         ${PROJECT_SOURCE_DIR}/ext/luajit/${libpath}
         /usr/
         /usr/bin/
--- a/cmake/FindLuaJIT.cmake
+++ b/cmake/FindLuaJIT.cmake
@@ -44,7 +44,7 @@ endif()
 
 find_library(LuaJIT_LIBRARY
     NAMES
-        luajit luajit_64 luajit-5.1 libluajit libluajit_64
+        libluajit.a luajit luajit_64 luajit-5.1 libluajit libluajit_64
     PATHS
         ${LuaJIT_SOURCE_DIR}/src/
         ${PROJECT_SOURCE_DIR}/ext/luajit/${libpath}
--- a/cmake/FindLuaJIT.cmake
+++ b/cmake/FindLuaJIT.cmake
@@ -38,7 +38,13 @@ if (UNIX)
 
         # LuaJIT has no CMake support, so we break out to using make on it's own
         message(STATUS "Building LuaJIT from src")
-        execute_process(COMMAND make WORKING_DIRECTORY ${LuaJIT_SOURCE_DIR})
+
+        if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+            execute_process(COMMAND export MACOSX_DEPLOYMENT_TARGET=11.6 ; make WORKING_DIRECTORY ${LuaJIT_SOURCE_DIR}) # TODO don't hardcode MACOSX_DEPLOYMENT_TARGET
+        else()
+            execute_process(COMMAND make WORKING_DIRECTORY ${LuaJIT_SOURCE_DIR})
+        endif()
+
     endif()
 endif()
 


claywar (approved):

neuromancerxi:
Applied patch against ec031c23f1d09c3ea958c7713ebe58e7d1cf01f6

```
OK        Successfully built LuaJIT
make[1]: Leaving directory '/home/ffxitest/git/temp/build/_deps/luajit-src/src'
==== Successfully built LuaJIT 2.1.0-beta3 ====
-- Found LuaJIT: /home/ffxitest/git/temp/build/_deps/luajit-src/src/libluajit.so  
-- LuaJIT_FOUND: TRUE
-- LuaJIT_LIBRARY: /home/ffxitest/git/temp/build/_deps/luajit-src/src/libluajit.so
-- LuaJIT_INCLUDE_DIR: /home/ffxitest/git/temp/ext/luajit/include/
```

Launched server, uncommented qm1 in Abyssea Vunk `onTrigger`, crash still occurs with same error as https://github.com/LandSandBoat/server/issues/2116#issuecomment-1179019148

neuromancerxi:
Tested the POC `!delmodnil.lua` suggested in https://github.com/LandSandBoat/server/issues/2238, server crash still occurs.

WinterSolstice8:
We're still working on it but you can workaround the current problem like this:

`ln -s /home/ffxitest/git/temp/build/_deps/luajit-src/src/libluajit.so  /home/ffxitest/git/temp/build/_deps/luajit-src/src/libluajit-5.1.so.2`

you can verify its using it hre:

```
ldd ./xi_search
        linux-vdso.so.1 (0x00007ffd011ba000)
        libluajit-5.1.so.2 => /home/ffxitest/git/temp/build/_deps/luajit-src/src/libluajit-5.1.so.2

```

WinterSolstice8:
Just updated the PR with what should fix it without manual intervention.

neuromancerxi (approved):
```
[07/09/22 14:07:25:396][map][info] [FileWatcher] COMMAND scripts/commands/delmodnil.lua (CacheLuaObjectFromFile:549)
[07/09/22 14:07:31:714][map][error] cmdhandler::call: (delmodnil) error: stack index 3, expected number, received nil: not a numeric type (bad argument into 'void(short unsigned int, short int)')
stack traceback:
        [C]: in function 'delMod'
        ./scripts/commands/delmodnil.lua:16: in function <./scripts/commands/delmodnil.lua:15> (call:321)
[07/09/22 14:08:43:693][map][debug] CZone:: Northern_San_dOria DecreaseZoneCounter <0> Dave (DecreaseZoneCounter:466)
[07/09/22 14:08:46:521][map][debug] CZone:: Abyssea-Vunkerl IncreaseZoneCounter <1> Dave (InsertPC:102)
[07/09/22 14:08:51:797][map][info] [FileWatcher] scripts/zones/Abyssea-Vunkerl/npcs/qm1.lua -> xi.zones.Abyssea-Vunkerl.npcs.qm1 (CacheLuaObjectFromFile:647)
[07/09/22 14:08:59:046][map][error] luautils::onTrigger: stack index 1, expected number, received nil: not a numeric type (bad argument into 'std::optional<CLuaBaseEntity>(unsigned int, const sol::basic_object<sol::basic_reference<false> >&)')
stack traceback:
        [C]: in function 'GetMobByID'
        ./scripts/globals/abyssea.lua:687: in function 'qmOnTrigger'
        scripts/zones/Abyssea-Vunkerl/npcs/qm1.lua:16: in function 'fallbackHandler'
        ./scripts/globals/interaction/interaction_lookup.lua:397: in function <./scripts/globals/interaction/interaction_lookup.lua:361> (OnTrigger:1835)
```

Errors (expected), but no crash!

zach2good:
Merge whenever, I'll fix the osx stuff another time

PR #2245, opened 2022-07-09
Quest: Know One's Onions scroll reward change

mumencoder:
**_I affirm:_**

The Blaze Spikes scroll will be awarded immediately after you turn in the wild onions instead of when the quest completes. Source https://ffxiclopedia.fandom.com/wiki/Know_One%27s_Onions

!setplayerlevel 5
!completequest 2 36
!pos -26.8 -6 190 240 testplayer
!giveitem testplayer 4387 4

### DIFF
--- a/scripts/quests/windurst/SOB2_Know_Ones_Onions.lua
+++ b/scripts/quests/windurst/SOB2_Know_Ones_Onions.lua
@@ -16,7 +16,6 @@ quest.reward =
 {
     fame     = 10,
     fameArea = xi.quest.fame_area.WINDURST,
-    item     = xi.items.SCROLL_OF_BLAZE_SPIKES,
     title    = xi.title.SOB_SUPER_HERO,
 }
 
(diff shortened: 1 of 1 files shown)


claywar:
Could you please provide a cap to support this?  I'm all for updating to ensure accuracy, but have often-found that wiki to be questionable in the past.

mumencoder:
I can't provide a cap but https://ffxi.allakhazam.com/db/quests.html?fquest=30 agrees and https://www.bg-wiki.com/ffxi/Know_One%27s_Onions doesn't. Comment section in allakhazam also corroborates getting the scroll early. I think it makes sense, because there is some special logic for this quest that happens after the onion trade and perhaps SE didn't want to screw up the player receiving the scroll or not.

claywar:
I'm sorry, without accurate captures, this cannot be accepted.  If you feel that this is an issue that needs investigation, please raise one, and it can be verified at a later point.

TeoTwawki:
> I can't provide a cap but https://ffxi.allakhazam.com/db/quests.html?fquest=30 agrees and https://www.bg-wiki.com/ffxi/Know_One%27s_Onions doesn't. Comment section in allakhazam also corroborates getting the scroll early. I think it makes sense, because there is some special logic for this quest that happens after the onion trade and perhaps SE didn't want to screw up the player receiving the scroll or not.

I believe there are multiple branching paths with this, as in 6 diff chars it didn't work for me the way they describe, but on a 7th char it did. It may have to do with how long the kids get left waiting for you before you arrive with onions, or some other task being already done or not when the quest starts. 6/7 didn't happen the way BG describes. 

 We will def need a cap to see all the params the scenes are checking, may shed light on wtf is going on with it.

At one point in the past I was the one to change this because _I ran through and did not get my scroll till later than wiki says,_ and I had been checking because the quest was being exploited for duplicate scrolls to sell on AH at the time.

PR #2247, opened 2022-07-10
[core] Fix mob not rotating towards player at very specific distances, add Petrification handling to CMobController::CheckDetection

WinterSolstice8:
**_I affirm:_**

Adds in a check to determine if the mob should face their target if they are just barely inside melee range and don't need to move
Adds in Petrification to the list of statuses that a mob should check to determine deaggro.

FaceTarget:
Zone to sky, attack a Flamingo at 2.80~ distance and rotate around it. It should consistently rotate towards you, where previously sometimes it would not.

Petrification check:
Cast Break on an enemy, move far from it and put up sneak/invis/deodorize. It should deaggro shorter than before.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -145,7 +145,7 @@ bool CMobController::CheckDetection(CBattleEntity* PTarget)
 {
     TracyZoneScoped;
     if (CanDetectTarget(PTarget) || CanPursueTarget(PTarget) ||
-        PMob->StatusEffectContainer->HasStatusEffect({ EFFECT_BIND, EFFECT_SLEEP, EFFECT_SLEEP_II, EFFECT_LULLABY }))
+        PMob->StatusEffectContainer->HasStatusEffect({ EFFECT_BIND, EFFECT_SLEEP, EFFECT_SLEEP_II, EFFECT_LULLABY, EFFECT_PETRIFICATION }))
     {
         TapDeaggroTime();
     }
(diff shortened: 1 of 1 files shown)


src/map/ai/controllers/mob_controller.cpp
```diff
@@ -145,7 +145,7 @@ bool CMobController::CheckDetection(CBattleEntity* PTarget)
 {
     TracyZoneScoped;
     if (CanDetectTarget(PTarget) || CanPursueTarget(PTarget) ||
-        PMob->StatusEffectContainer->HasStatusEffect({ EFFECT_BIND, EFFECT_SLEEP, EFFECT_SLEEP_II, EFFECT_LULLABY }))
+        PMob->StatusEffectContainer->HasStatusEffect({ EFFECT_BIND, EFFECT_SLEEP, EFFECT_SLEEP_II, EFFECT_LULLABY, EFFECT_PETRIFICATION }))
```
zach2good, line 148: I don't know this system on retail,  can you link me some proof of this?

zach2good (approved):

src/map/ai/controllers/mob_controller.cpp
WinterSolstice8: So after looking into it, that "tapDeaggroTime" seems to update a mob has "seen" or "known" about an
attacker. 

Checked on retail:
image
(sneak and invis wore off because the Bee attacked me)

I was standing at 49.5' so I could still see it, I then repeated the test where I moved hundreds of units away and had a second character follow the bee. There's no log because there's no way to prove without a huge packet dump or video of what happened, but I will describe it:

As above, the mob woke up and then went straight for the character that cast Sleep II on it, but at exactly 30 seconds later it deaggroed, waited a couple seconds, then despawned. The bee remained claimed until it stopped moving just before it went unclaimed. This is consistent with `return PMob->CanDeaggro() && (m_Tick >= m_DeaggroTime + 25s);` provided m_DeaggroTime is 5 seconds.

zach2good (approved):
Thanks for the retail testing 🙏

PR #2250, opened 2022-07-10
Adjust drops for Meteorite Crawlers

Aurasplash:
**_I affirm:_**

Corrects which Crawlers drop Meteorite. Only Crawlers on the Hill in WEST Sarutabaruta should drop Meteorites. Currently the Crawlers on the hill use the Drop ID for EAST Sarutabaruta instead of having their own Drop ID.

Killed Crawlers


zach2good (approved):
Seems reasonable

claywar (changes requested):
Looks like you're still using the branch you created previously for Yngvi Choker (which was merged).  Also, looking at that, 26040 is still defined as yngvi_earring in item_equipment

Issue #1833, opened 2022-05-29
Item Levels failing to apply to pets.


Hinekri:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Branch affected by issue 

`petutils.cpp'

## Steps to reproduce
I dunno where the code is supposed to be at to check it, (I thought it was already in, and thought it was petutils.cpp, but I didn't see it there just now) but this is THE ONLY place we set it. There is no modifier. There should not be one ever because that's not how its intended to work. Avatar still weak means it's probably missing a line of code to apply it to the avatar or that said line of code is broken.

^ Quote from my convo in Discussions related to the subject.

Use !getstats on an Avatar naked.
Use !getstats on an Avatar with Epitaph iLvl119 Ammo.
   No changes in stats at all remains base lv99.
Same results with Pup Automaton & Bst Jugs.

## Expected behavior

Pets should scale with proper Item equipped
![Kirito_2022 05 29_141110](https://user-images.githubusercontent.com/106329891/170887830-95d5abf0-0546-47e4-be1e-a52936717cf1.png)
![Kirito_2022 05 29_141119](https://user-images.githubusercontent.com/106329891/170887834-05b61fb0-c9ec-43ee-935d-8260b82d57f5.png)
.

TeoTwawki:
related discussion: [#1832](https://github.com/LandSandBoat/server/discussions/1832#discussioncomment-2844651)

WinterSolstice8:
avatars, wyverns and automatons have various lvl+ modifier gear. trusts too, apparently.
https://www.ffxiah.com/search/item?&description=Lv.%2B1#adv
also all ilvl 119 Nirvanas have + levels for avatars as well

TeoTwawki:
~~I'd really like us to _stop_ using the word "modifier". we do not use a modifier and calling it that is going to have ppl confusing it with one. We have a column on the item in the database table. In the discussion thread I quoted BG's page for which slot affects the levels of which entities. Item level already exists on the item object itself. We just need to ensure it is checked and applied.~~

Me: HEY! I _said_ its not a...
SE: oh Teo...
[...]

PR #2251, opened 2022-07-10
[DRAFT] Apply ilvl to Trust

neuromancerxi:
This builds out the plumbing to apply additional "levels" to the Trust based on the Master's ilvl.
Reference: https://www.bg-wiki.com/ffxi/Category:Trust#Alter_Ego_Statistics_.26_Levels

**_I affirm:_**

The initial pass at implementing the ilvl bonuses was to just adjust the level of the trust based on the player ilvl, but this created problems with Party EXP. Instead, approached it from the standpoint of faking in levels at the `LoadTrustStatsAndSkills` phase.

Skill values were approximated based on observations from ilvl weapons:

|          |i105|i109|i115|i117|i119|
|---------|---|---|---|---|---|
|Parry     | 54|108|188|215|242|
|Weapon    | 54|108|188|215|242|
|Shield    |   | 48|   |100|112|
|Melee MACC| 42| 85|146|167|188|
|Staff MACC| 51|102|177|203|228|

These have been approximated into tiers in `trustutils`:
100 - 109, 110 - 118, and 119

These boosts apply to the Trust as follows:

STR/DEX/HP/ETC are Modified as Main/Sub Level + iLvl Bonus
EVA/ACC/ATT/RACC/RATT are Modified based on iLvl Parry/Weapon values
MEVA has had a fixed 12.5/ilvl applied since armor varies wildly

I've also added (but commented out) the shield tiering, and also applied the MeleeSkill bonuses to the block where these skills are called out since (currently) trusts don't actually hold weapons (these are applied as ACC/ATT/RACC/RATT instead).

* Log a test character in
* Add a trust
* Summon the Trust without any ilvl Gear
* Review mod values for STR, INT, HP, MP, ATT, ACC, etc
* Add ilvl gear (e.g. [Outrider Gear](https://www.ffxiah.com/search/item?&name=outrider#adv)
* Equip ilvl gear to get to a target ilvl
* Re-Summon the trust
* Review mod values for STR, INT, HP, MP, ATT, ACC, etc

### DIFF
--- a/src/map/entities/trustentity.h
+++ b/src/map/entities/trustentity.h
@@ -61,6 +61,7 @@ public:
     void OnMobSkillFinished(CMobSkillState& state, action_t& action) override;
     void OnWeaponSkillFinished(CWeaponSkillState& state, action_t& action) override;
 
+    uint8               iLvlBonus;
     uint32              m_TrustID{};
     TRUST_MOVEMENT_TYPE m_MovementType;
 };
--- a/src/map/utils/trustutils.cpp
+++ b/src/map/utils/trustutils.cpp
@@ -411,6 +411,7 @@ namespace trustutils
         // assume level matches master
         PTrust->SetMLevel(PMaster->GetMLevel());
         PTrust->SetSLevel(std::floor(PMaster->GetMLevel() / 2));
+        PTrust->iLvlBonus = charutils::getItemLevelDifference(PMaster);
 
         LoadTrustStatsAndSkills(PTrust);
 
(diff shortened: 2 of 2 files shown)


neuromancerxi:
Soliciting feedback before moving out of draft. Thanks in advance.

zach2good:
One thing that needs to come in with iLvl for trusts is that the iLvl is dynamic and you don't need to re-summon trusts if you go from 99 -> i105 -> i119 etc.

EDIT: It's easy to do, there's no reason it shouldn't be done now

Hinekri:
> One thing that needs to come in with iLvl for trusts is that the iLvl is dynamic and you don't need to re-summon trusts if you go from 99 -> i105 -> i119 etc.
> 
> EDIT: It's easy to do, there's no reason it shouldn't be done now

Related current issue: https://github.com/LandSandBoat/server/issues/1833 .   

An iLvl modifier PR was merged 2 days ago that only included Pet/Avatar/Auto and this scaling is just for Trusts. Probably beyond the scope of this PR but we also need the +1Mod for Trusts and ilvl Scale for the Pets. Will this also handle scaling over 119 for the additional modifiers?

PR #2254, opened 2022-07-11
Sarutabaruta Region NM Audit

CriticalXI:
**_I affirm:_**

- Implemented missing NM Brazen Bones
- Duke Decapod no longer stands back and casts water frequently.
- Maltha now has a much higher attack speed.


scripts/zones/Giddeus/mobs/Yagudo_Persecutor.lua
```diff
@@ -11,7 +11,7 @@ entity.onMobDeath = function(mob, player, isKiller)
 end
 
 entity.onMobDespawn = function(mob)
-    xi.mob.phOnDespawn(mob, ID.mob.JUU_DUZU_THE_WHIRLWIND_PH, 5, math.random(3600, 7200)) -- 1 to 2 hours
+    xi.mob.phOnDespawn(mob, ID.mob.JUU_DUZU_THE_WHIRLWIND_PH, 5, 3600) -- 1 to 2 hours
```
TeoTwawki, line 14: missed updating the comment :)

also anecdotal but I'm pretty sure he and mjoo have the same spawn rate as I usually get them back to back when I've been killing their ph's. I run the circuit and boomerang his ph from the ledge on my way to vu puqu the beguiler and usually get all 3 if nobody else messes up my rotation

scripts/zones/Giddeus/mobs/Yagudo_Piper.lua
```diff
@@ -12,7 +12,7 @@ entity.onMobDeath = function(mob, player, isKiller)
 end
 
 entity.onMobDespawn = function(mob)
-    xi.mob.phOnDespawn(mob, ID.mob.VUU_PUQU_THE_BEGUILER_PH, 5, 900) -- 15 minutes
+    xi.mob.phOnDespawn(mob, ID.mob.VUU_PUQU_THE_BEGUILER_PH, 15, 900) -- 15 minutes
```
TeoTwawki, line 15: 15min??

see above, I don't think I've ever seen him that fast myself, but I guess its possible. if it is I'd guess he's lower rate tho

scripts/zones/Giddeus/mobs/Yagudo_Persecutor.lua
TeoTwawki: just did some digging and found some posts indicating all 3 yag NM are actually pure lotto, so my 1hr repop for all 3 is combo of luck and spawn rates I guess

TeoTwawki: ima leave well enough alone due to not knowing which sources I trust on this. you wanna edit that comment tho :wink:

CriticalXI: Would you consider the posts you found good enough evidence to give them all pure lotto cooldown?

TeoTwawki: I'm not sure how much I trust hoo mjuu true lotto. I think I could believe it on the other 2 even tho I never witnessed it any of my own camping sessions, but as camped as mjuu was I just find it hard to believe nobody ever caught it sooner. I used to love ruining the RMT's day there and even now on modern I swill sweep through all 3 NM there when I am bored (and the newer on they added near the BCNM end of the zone)

You can change Duzu if you want or leave as is - bg backs the true lotto claim on his wiki page

Xaver-DaRed (approved):

PR #2258, opened 2022-07-11
[Trust] - Apply Mob Pool Mods to Trust

neuromancerxi:
**_I affirm:_**

* Add Call to load mob_pool_mods before Trust Stats are applied which allows for MPP and others to be applied correctly
* Update lua_base_entity to expand break conditions to exclude Trusts as well as Mobs
* Migrates MODs that don't scale with level from individual scripts to sql/mob_pool_mods, but leaves MODs that scale by level in place (e.g. Shantotto II power scaling)

Summon Various Trust, Compare MOD Value with mob_pool_mods value

e.g. Shikaree Z -> 
`!getmod JUMP_TP_BONUS`
`!getmod MPP`

### DIFF
--- a/scripts/globals/spells/trust/apururu_uc.lua
+++ b/scripts/globals/spells/trust/apururu_uc.lua
@@ -61,7 +61,7 @@ spell_object.onMobSpawn = function(mob)
     mob:addSimpleGambit(ai.t.SELF, ai.c.NOT_STATUS, xi.effect.STONESKIN, ai.r.MA, ai.s.SPECIFIC, xi.magic.spell.STONESKIN)
 
     -- BGwiki states 75/tick regain.  Only used for Nott WS.
-    -- REGAIN MOD migrated to mob_pool_mods
+    mob:addMod(xi.mod.REGAIN, 75)
 
     mob:SetAutoAttackEnabled(false)
 end
--- a/scripts/globals/spells/trust/apururu_uc.lua
+++ b/scripts/globals/spells/trust/apururu_uc.lua
@@ -61,7 +61,7 @@ spell_object.onMobSpawn = function(mob)
     mob:addSimpleGambit(ai.t.SELF, ai.c.NOT_STATUS, xi.effect.STONESKIN, ai.r.MA, ai.s.SPECIFIC, xi.magic.spell.STONESKIN)
 
     -- BGwiki states 75/tick regain.  Only used for Nott WS.
-    mob:addMod(xi.mod.REGAIN, 75)
+    -- REGAIN MOD migrated to mob_pool_mods
 
     mob:SetAutoAttackEnabled(false)
 end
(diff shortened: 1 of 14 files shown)


TeoTwawki:
Is this necessary? is there a reason to avoid having them in the scripts?

neuromancerxi:
> Is this necessary? is there a reason to avoid having them in the scripts?

Currently things like HPP/MPP will not apply correctly when added to the script. When present in the script the Trust will spawn with their base HP/MP, then on engaging a mob for the first time the MOD will apply. This creates a situation where the Regen/Refresh pool will never refill above the base value the mob spawned with as well as potentially kicking the Trust into Yellow HP.

When applied as a mob_pool_mod, the trust will spawn with the mod already present, (128 MP becomes 256 for MPP 100, etc).

TeoTwawki:
> Currently things like HPP/MPP will not apply correctly when added to the script.

@neuromancerxi  Open an issue for that, I want those to function via script - regardless of if its a trust or not.

>  When present in the script the Trust will spawn with their base HP/MP, then on engaging a mob for the first time the MOD will apply. This creates a situation where the Regen/Refresh pool will never refill above the base value the mob spawned with as well as potentially kicking the Trust into Yellow HP.
> 
> When applied as a mob_pool_mod, the trust will spawn with the mod already present, (128 MP becomes 256 for MPP 100, etc).

do we have an onMobInitialize or equivalent for trusts @zach2good ? or just onSpawn and onEngage etc

@neuromancerxi for mods that absolutely _**can not**_ be made to work under script I would support conversion, but please do not convert any that will function under script. Regain in particular should not be having a problem being on spawn.

TeoTwawki:
actually, let me just double down here: even if we do things in mob pools mods, whether temporarily or permanently to work around not being able to script it, should have an issue raised about not being able to be done in script - I think it should ALLL be scriptable and see any inability to do so as a flaw.

That goes for mobmod as well - we should refactor eventually to make it all possible.

neuromancerxi:
https://github.com/LandSandBoat/server/issues/2259 opened.

neuromancerxi:
@TeoTwawki Following up on this one. Minus the conflict that I need to resolve, were there any other concerns with adopting this PR?

TeoTwawki:
> @TeoTwawki Following up on this one. Minus the conflict that I need to resolve, were there any other concerns with adopting this PR?

Got a conflict to resolve and need other dev(s) to read it.

WinterSolstice8:
I don't see anything wrong with this PR technically but I think Teo's point that this could be moved to an event in lua instead makes a lot of sense. I could cook up an event, make a PR for it, then we never have to touch mob pools again for trusts which will make it very user friendly for modules.

The idea I have is simple: the event is called by a function where in said event you can set a trusts mods/mobmods in it, and it applies mobmods just like it would for real mobs. How's that sound?

neuromancerxi:
> I don't see anything wrong with this PR technically but I think Teo's point that this could be moved to an event in lua instead makes a lot of sense. I could cook up an event, make a PR for it, then we never have to touch mob pools again for trusts which will make it very user friendly for modules.
> 
> The idea I have is simple: the event is called by a function where in said event you can set a trusts mods/mobmods in it, and it applies mobmods just like it would for real mobs. How's that sound?

I'm onboard with keeping it in lua, I just didn't see a mechanism available that would apply these mods on initialize correctly.

WinterSolstice8:
dang, forgot all about this. Work got really busy at that time. Needs a rebase

WinterSolstice8 (approved):

PR #2267, opened 2022-07-12
Rename Yngvi Earring to Ygnvi Choker, remove from AH listing

Aurasplash:
This was renamed and as it is RA/EX it should not be on the AH.

**_I affirm:_**

Yngvi Choker has been renamed from Yngvi Earring. It was also RA/EX but on the AH.


TeoTwawki (approved):
https://www.ffxiah.com/item/26040/yngvi-choker

WinterSolstice8 (approved):

PR #2274, opened 2022-07-13
[core] Remove Mobspell list ID limit

The-Aerec:
**_I affirm:_**

Removes the requirement of a spell list limit on loading.

Tested by running around and poking mobs and ensuring they still use their proper spells

make new spell lists over id 500 and have fun.

### DIFF
--- a/src/map/mob_spell_list.cpp
+++ b/src/map/mob_spell_list.cpp
@@ -38,24 +38,22 @@ void CMobSpellList::AddSpell(SpellID spellId, uint16 minLvl, uint16 maxLvl)
 // Implement namespace to work with spells
 namespace mobSpellList
 {
-    CMobSpellList* PMobSpellList[MAX_MOBSPELLLIST_ID];
+    uint16                      MAX_MOBSPELLLIST_ID = 1;
+    std::vector<CMobSpellList*> PMobSpellList;
 
     // Load list of spells
     void LoadMobSpellList()
     {
-        memset(PMobSpellList, 0, sizeof(PMobSpellList));
-        PMobSpellList[0] = new CMobSpellList();
-
+        PMobSpellList.emplace_back(new CMobSpellList());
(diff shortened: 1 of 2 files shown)


The-Aerec:
Once this passes and everyone likes the final method for this i'll go through and do the same for skills etc.

Hinekri:
MAX_MOBSPELLLIST_ID 500  all we did was change this to 5000 and it works.

TeoTwawki:
> MAX_MOBSPELLLIST_ID 500 all we did was change this to 5000 and it works.

I did the same thing.

Due to potential gaps in the list and future changes that may create gaps, I don't think this method is the way to go. some of out other things used to have MAX_WHATEVERS and those became unordered maps for the most part. tried to track down the commits to link last night but had no luck (which lead to a whole brainstorming session about findable tagging in commit messages, but I digress) - an unordered map is prolly the way to go for this one.

The-Aerec:
This accounts for gaps in the list as there are already gaps. Also when I spoke to Zach this is the method he suggested.

TeoTwawki:
@zach2good tag you're it!

zach2good:
An `unordered_map` would handle gaps gracefully, if the stored type is some kind of pointer.

Conversely, how often are we querying into a gap, should we be adding logging for that to find out?

TeoTwawki:
> Conversely, how often are we querying into a gap, should we be adding logging for that to find out?

I have a concern about unneeded warnings causing ppl to think they need to re-sequence the list to eliminate them (gaps and warnings) and I bigtime don't want that to happen. ATM gaps are harmless besides my OCD noticing them and wanting them infilled when new stuff is added later. I Don't think a warning is needed or useful as long as gaps are marked in-file.

zach2good:
In the interest of getting our open PRs down, I'd recommend just reverting what you've done so far (sorry!) and bump the limit up to 5000.

PR #2067, opened 2022-06-21
Scheduled Persistence Plumbing

zach2good:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

Brings in a bunch of goodies from Eden, setting us up to have scheduled/delayed persistence, charvar caching, and charvar change announcements between processes.

Closes https://github.com/LandSandBoat/server/issues/2273

## Steps to test these changes

- Make sure that content that uses charvars (read & write) still work
- Make sure that content that uses servervars (read & write) still work

PR #2278, opened 2022-07-14
[lua] Add '!trustengage' custom helper

zach2good:
**_I affirm:_**

As always, I had assumed that someone else would step in and provide this. I laid it out to be very easy to add. But here we are.

Depends on https://github.com/LandSandBoat/server/pull/2067 to be performant, and not need a strange caching solution.

Allows you to set the engagement type of your trusts

- `!trustengage` to see help menu
- Summon a trust
- Engage as normal, but stay out of melee range
- `!trustengage 1`
- See the trust run in
- Further engagements will act this way
- Will survive zoning and logging in/out
- `!trustengage 0` to go back to normal/retail

### DIFF
--- a/scripts/commands/trustengage.lua
+++ b/scripts/commands/trustengage.lua
@@ -1,0 +1,39 @@
+-----------------------------------
+-- func: trustengage
+-- desc: Sets the engagement type of a players trusts
+-----------------------------------
+require("scripts/globals/settings")
+
+cmdprops =
+{
+    permission = 0,
+    parameters = "i"
+}
+
+local types =
+{
+    [0] = { 0, "Retail: Master engage and melee swing" },
+    [1] = { 1, "Attack: Master engage" },
+}
+
+function error(player)
+    player:PrintToPlayer(string.format("!trustengage <type number>\n" ..
+    "0: %s\n" ..
+    "1: %s", types[0][2], types[1][2]))
+    local type = player:getCharVar("TrustEngageType")
+    player:PrintToPlayer(string.format("Currently set to:\n %i: %s", type, types[type][2]))
+end
+
+function onTrigger(player, type)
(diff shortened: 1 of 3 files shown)


Hinekri:
I thought I remembered them doing this at one point before, became wildly abused when trusts had decently scaled stats on our server as an easy macro bot "tab, engage, wait for trusts to do work". Honestly kind of like the melee requirement personally but quite happy it comes with permissions built in, ty!

zach2good:
Oh good idea, I should add a server setting (default to off) that enables/disables this globally

TeoTwawki:
> Oh good idea, I should add a server setting (default to off) that enables/disables this globally

I thought that was the plan back when you 1st started adding trust ai

zach2good:
It's been so long now, I don't even remember what the initial plans were ;_;

TeoTwawki:
> It's been so long now, I don't even remember what the initial plans were ;_;

I recall discussion of configurable "better than retail" behavior including different engage mechanics

zach2good:
Well then, thats this, it's just globally on/off, and then players can decide if they want to change the conditions or not

zach2good:
The pre-req branch for this has been merged, this is ready for review/merge

WinterSolstice8 (approved):
LGTM

and now you've gotten me thinking about custom overrides for spell/ability use again, which we need for Job Points/Impact/Dispelga etc that could be used to add a similar feature as this PR for DRG wyverns...

TeoTwawki:
> LGTM
> 
> and now you've gotten me thinking about custom overrides for spell/ability use again, which we need for Job Points/Impact/Dispelga etc that could be used to add a similar feature as this PR for DRG wyverns...

-nods-

atm we do impact by adding to the character on equip, and removing on de-equip. like deleting it out of the players sql rows. I think we need a more refined approach to that than what we do now.

PR #2219, opened 2022-07-07
Environment Variable Refactor + Remove Docker files

Sunderous:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

Update settings.cpp to allow for arbitrary configuration of environment variables using the template XI_{outerKey}_{innerKey}.

Updates to the docker related files to configure everything through the environment, as well as add a reverse proxy to make server available from non-localhost boxes by default.

## Steps to test these changes

Ignoring the Docker stuff, if you set any environment variables before you start like XI_NETWORK_SQL_HOST, or XI_MAIN_SERVER_NAME, it should be automatically reflected in the logs and the runtime behavior when starting up the server.

PR #2318, opened 2022-07-17
changed where env variables are loaded

m241dan:
**_I affirm:_**

Load important environment variables again in a way that matters.

Use environment variables for you sensitive data instead of silly-ole -settings scripts.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -157,6 +157,8 @@ namespace settings
             }
         }
 
+        network_settings_from_env();
+
         // Scrape user settings into cpp's settingsMap
         // This will overwrite the defaults, if user settings exist. Otherwise the
         // defaults will be left intact.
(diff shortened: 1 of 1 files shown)


zach2good:
I'm wanting to get the env var reading logic from the following PR merged: https://github.com/LandSandBoat/server/pull/2219

So this PR won't be necessary. Thanks though :)

m241dan:
Okay, but perhaps let's not submit PRs that break other people's deployments?

zach2good:
Doesn't break our build. If you deviate from our build, that's up to you to fix.

m241dan:
See edit, deployment.

TeoTwawki:
Your edit doesn’t really change anything for us. In our stock intended setup, we aren’t getting any major problem.

m241dan:
I'm not trying to be a dick, I know Im coming off as one. But I guess it's stemming from a larger frustration of the fact that we have this little check box thing that says, "I tested my code." And I keep coming across things that are entirely untested.

TeoTwawki:
Yes, you are coming off as one. No acknowledging it doesn’t make it ok.

And yes, we do test our stuff - for the stock setup which is why we have an open discussion about dropping docker off because WE don’t use it

PR #2326, opened 2022-07-17
[core] Update ENV VAR handling to be more universal

Sunderous:
**_I affirm:_**

Update settings.cpp to allow for arbitrary configuration of environment variables using the template XI_{outerKey}_{innerKey}.

Ignoring the Docker stuff, if you set any environment variables before you start like XI_NETWORK_SQL_HOST, or XI_MAIN_SERVER_NAME, it should be automatically reflected in the logs and the runtime behavior when starting up the server.

### DIFF
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -35,16 +35,19 @@ namespace settings
 {
     std::unordered_map<std::string, SettingsVariant_t> settingsMap;
 
-    void network_settings_from_env()
+    // We need this to figure out which environment variables are numbers
+    // so we can pass them to the lua settings properly typed.
+    bool isNumber(const std::string& stringValue)
     {
-        lua["xi"]["settings"]["network"]["SQL_HOST"]     = std::getenv("XI_DB_HOST") ? std::getenv("XI_DB_HOST") : get<std::string>("network.SQL_HOST");
-        lua["xi"]["settings"]["network"]["SQL_PORT"]     = std::getenv("XI_DB_PORT") ? std::stoi(std::getenv("XI_DB_PORT")) : get<uint16>("network.SQL_PORT");
-        lua["xi"]["settings"]["network"]["SQL_LOGIN"]    = std::getenv("XI_DB_USER") ? std::getenv("XI_DB_USER") : get<std::string>("network.SQL_LOGIN");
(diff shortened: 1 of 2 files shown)


Sunderous:
Sorry, is that sanity check failing from the formatting? Hard to tell just from the output.

zach2good:
You can save the output of that check as a file, and then apply it to your code with 'git am filename', and it'll fix your formatting (or you can run clang-format-14).

The patch to save:
```
diff --git a/src/common/settings.cpp b/src/common/settings.cpp
index ee44976fc2..62b69b9fe3 100644
--- a/src/common/settings.cpp
+++ b/src/common/settings.cpp
@@ -37,12 +37,12 @@ namespace settings
     // We need this to figure out which environment variables are numbers
     // so we can pass them to the lua settings properly typed.
-    bool isNumber(const std::string stringValue) 
+    bool isNumber(const std::string stringValue)
     {
-        for (char const c : stringValue) 
+        for (char const c : stringValue)
         {
-            if(std::isdigit(c) == 0) 
-            { 
+            if (std::isdigit(c) == 0)
+            {
                 return false;
             }
         }
@@ -191,7 +191,7 @@ namespace settings
                 // If we try to assign this value in the if() statement, it will
                 // come back as a bool, so we have to check only then assign in the
                 // block.
-                if (std::getenv(envKey.c_str())) 
+                if (std::getenv(envKey.c_str()))
                 {
                     auto value = std::string(std::getenv(envKey.c_str()));
                     ShowInfo(fmt::format("Applying ENV VAR {}: {} -> {}", envKey, key, value));
@@ -199,11 +199,11 @@ namespace settings
                     // If we don't convert the PORTS to doubles (or ints), then the LUA
                     // doesn't interpret them correctly and it breaks everything.
                     // Therefor we need to check if the value is a number.
-                    if (isNumber(value)) 
+                    if (isNumber(value))
                     {
                         settingsMap[key] = std::stod(value);
                     }
-                    else 
+                    else
                     {
                         settingsMap[key] = value;
                     }
```

Sunderous:
Ahhhhh, I was like "this looks a lot like a git diff..." haha. If I end up doing much work in C++ land I'll figure out how to setup clang format to auto-format my code in the editor. For now I'll apply that and try and squash it.

Sunderous:
@zach2good 
Alright, clang issues addressed, should be good to go when the rest of the checks finish.

Thanks for the help.

claywar (approved):

zach2good (approved):

PR #2337, opened 2022-07-18
BCNM Royal Jelly Implementation

dallano:
**_I affirm:_**

Implements BCNM Royal Jelly

Changes:
 - Removed Princess Jellies skill set.
 - Gave Princess Jellies weakness depending on their element
 - Added SCRIPT variable to pathfinding to allow the jellies to path to the center of the room (Can be used for a multitude of other mobs)
 - Slightly increased pre-existing spell cool down on Princes Jellies as they were casting faster than evidence found in youtube videos.
 - Removed some unused functions from code ported.
 - Increased the distance at which the jellies will cast on players to 20 yalms. (Previously set at 15 which would allow players to cast on jellies without any retaliation)
 - Slightly decreased the odds a Princess Jelly would cast bind by 10% to 20% (Review of past videos proved 33% was too high of odds - which was the previous rate on code from Setzor)
 - Implemented the drop tables for Royal Jelly which was taken from ffxiclopedia with a sample rate of ~400+
 - Cleaned up table formatting in Princess_Jelly.lua
 - Fixed Queen Jelly from spawning in the initiation of the battlefield
 - Moved some modifiers being added onto princess and queen luas into mob pools SQL
 - Added spell lists for both Princess and Queen Jelly

Enter BCNM Royal Jelly in Waughroon Shrine to test fight


scripts/zones/Waughroon_Shrine/mobs/Princess_Jelly.lua
```diff
@@ -0,0 +1,185 @@
...
+require("scripts/globals/utils")
+-----------------------------------
+local princesses =
+{
+    {17367174, 17367175, 17367176, 17367177, 17367178, 17367179, 17367180, 17367181},
+    {17367184, 17367185, 17367186, 17367187, 17367188, 17367189, 17367190, 17367191},
+    {17367194, 17367195, 17367196, 17367197, 17367198, 17367199, 17367200, 17367201},
+}
+local elementalSpells =
+{
```
claywar, line 12: Pretty sure we have enums for these, and should be used in that case

scripts/zones/Waughroon_Shrine/mobs/Princess_Jelly.lua
```diff
@@ -0,0 +1,185 @@
+-----------------------------------
+-- Area: Waughroon Shrine
+--  MOB: princess_jelly
+-- Author: Setzor (ported to LSB by dallano)
+-----------------------------------
+require("scripts/globals/pathfind")
+require("scripts/globals/utils")
+-----------------------------------
+local princesses =
+{
```
claywar: While not necessary, I think for readability's sake, it might be beneficial to explicitly key these values for the BCNM ID

scripts/zones/Waughroon_Shrine/mobs/Princess_Jelly.lua
```diff
@@ -0,0 +1,185 @@
...
+        mob:addMod(xi.mod.DARK_MEVA, -250)
+    else
+        mob:addMod(xi.mod.LIGHT_MEVA, -250)
+    end
+end
+
+local function getDistanceFromCenter(bfNum,mob)
+    local pos = mob:getPos()
+
+    local centerX = centers[bfNum][1]
```
claywar: Would suggest just pulling the array as a single value, or assigning this directly in difX/Y/Z.  Seems a bit excessive

scripts/zones/Waughroon_Shrine/npcs/Armoury_Crate.lua
```diff
@@ -116,6 +116,65 @@ local loot =
         },
     },
 
+     -- BCNM Royal Jelly
```
claywar, line 119: Assuming the rest of this script doesn't have a "pretty" formatting pattern, but would be nice to clean this up in the future.

scripts/zones/Waughroon_Shrine/mobs/Princess_Jelly.lua
```diff
@@ -0,0 +1,172 @@
...
+    mob:setMobMod(xi.mobMod.MAGIC_COOL, 40)
+end
+
+entity.onMobSpawn = function(mob)
+    mob:setSpeed(2)
+    mob:setUnkillable(false)
+    mob:setLocalVar('mobElement', math.random(1,8))
+    local element = mob:getLocalVar('mobElement')
+
+    if element == 1 then
```
TeoTwawki: Could use a keyed table instead of all this `elseif` conditionaling.

```lua
local mevaList =
{
    [0] = xi.mod.LIGHT_MEVA,
    [1] = xi.mod.WATER_MEVA,
    [2] = xi.mod.THUNDER_MEVA,
    [3] = xi.mod.EARTH_MEVA,
    [4] = xi.mod.WIND_MEVA,
    [5] = xi.mod.ICE_MEVA,
    [6] = xi.mod.FIRE_MEVA,
    [7] = xi.mod.DARK_MEVA
}
mob:addMod(mevaList[mob:getLocalVar('mobElement')], -250)
```
A variable not set with always return zero, so I just put light at the top as zero instead of an `else`.

edit: realized I don't even need the `element` variable, can get the localVar right there on the spot.

scripts/zones/Waughroon_Shrine/IDs.lua
```diff
@@ -86,6 +86,50 @@ zones[xi.zone.WAUGHROON_SHRINE] =
             17367285,
         },
     },
+
+    princesses =
+    {
+        [1] =
+        {
+            17367174,
```
TeoTwawki: since they are all sequential in the 3 sets, could we possibly use just the 1st of each set and offset from there?

TeoTwawki: actually, could just offset from the queens IDs

TeoTwawki:
needs rebased

PR #2340, opened 2022-07-18
BCNM Petrifying Pair Fixes

dallano:
The basis and evidence for all changes can be found in the following links:
https://www.youtube.com/watch?v=kU6rSHxnm9k&t=85s&ab_channel=HeavyMetalWyrm
https://ffxiclopedia.fandom.com/wiki/Petrifying_Pair?so=search

**_I affirm:_**

Fixed:
 - Increased attack rate of both lizards.
 - Gave lizards ability to use other lizard mob abilities.
 - Corrected Spawn points of both lizards to be retail accurate.
 - Gave both lizards sleep and lullaby immunity.
 - Both lizards will now link upon engagement
 - Both lizards will use secretion upon engagement.
 - Both lizards will favor baleful gaze. Upon extensive review, it appears to be roughly a 50% chance that the lizards use it.
 
 Evidence:
 https://www.youtube.com/watch?v=kU6rSHxnm9k&t=85s&ab_channel=HeavyMetalWyrm
https://ffxiclopedia.fandom.com/wiki/Petrifying_Pair?so=search

Enter Petrifying Pair BCNM in ghelsba outpost to test


scripts/zones/Ghelsba_Outpost/mobs/Kalamainu.lua
```diff
@@ -5,6 +5,24 @@
...
 
+entity.onMobSpawn = function(mob)
+    mob:setMod(xi.mod.SLEEPRES, 100)
+    mob:setMod(xi.mod.LULLABYRES, 100)
+    mob:setMobMod(xi.mobMod.NO_MOVE, 1) -- lock from moving
+end
+
+entity.onMobEngaged = function(mob, target)
+    mob:setMobMod(xi.mobMod.NO_MOVE, 0) -- unlock from moving
+    GetMobByID(mob:getID() + 1):updateEnmity(target) -- link Kilioa
```
TeoTwawki: its actually possible to just set them to "superlink" via mobmods, which is what bcnms do auto in the core - the reason it doesn't happen in ghelsba is it can't be flagged as a bcnm zone type (it's a special snowflake in that non bcnm thinsg happen in the same zone) 

https://github.com/LandSandBoat/server/blob/ff257cf51c312c5b186e203410318b066ccf3f1c/src/map/zone_entities.cpp#L252

https://github.com/LandSandBoat/server/blob/5f6b9bc26e7c0ee9d7f2fc3b306829daf2d03b16/src/map/utils/mobutils.cpp#L854

long term, I think we need to move away from relying on the zone type for that in the first place.

sql/mob_pools.sql
```diff
@@ -2233,7 +2233,7 @@ INSERT INTO `mob_pools` VALUES (2177,'Kadjaya','Kadjaya',149,0x01000B02000087208
 INSERT INTO `mob_pools` VALUES (2178,'Kadraeth_the_Hatespawn','Kadraeth_the_Hatespawn',233,0x0000EF0600000000000000000000000000000000,4,4,12,240,100,0,0,0,0,2,0,32,0,157,0,0,252,0,0,920,233);
 INSERT INTO `mob_pools` VALUES (2179,'Kagetora','Kagetora',149,0x01000E013B10042058305840585085613A711E80,13,13,9,240,100,0,0,0,0,0,0,32,0,153,0,0,7,0,0,149,149);
 INSERT INTO `mob_pools` VALUES (2180,'Kaiser_Behemoth_S','Kaiser_Behemoth',479,0x0000940100000000000000000000000000000000,1,1,6,380,100,1024,1,1,1,18,0,32,307,155,0,0,126,0,0,479,479);
-INSERT INTO `mob_pools` VALUES (2181,'Kalamainu','Kalamainu',174,0x0000490100000000000000000000000000000000,1,1,7,240,100,0,1,0,1,18,1,32,6623,157,0,0,0,0,0,711,174);
+INSERT INTO `mob_pools` VALUES (2181,'Kalamainu','Kalamainu',174,0x0000490100000000000000000000000000000000,1,1,7,150,100,0,1,0,1,18,1,32,6623,157,0,0,0,0,0,711,174);
```
TeoTwawki, line 2236: I will need to see some evidence that they are 150 delay attack speeds beyond eyballing a video. If gleaned from video, I want to know the method 150 was determined. While possible historically it has not been reliable.

dallano: Yeah I simply lined up the attack rate with videos. it was significantly faster than the default delay that was previously set. I can leave it at default if that's prefered.

TeoTwawki: yes. I the vid the lizard was getting approximately 2 attacks to that great axe warriors 1 (double attacks excluded) and since great axes are typically significantly over 300 I can be reasonably sure it's more than 150, even if that doesn't pin down what its precise attack speed is. better to leave alone when we don't have a new number we can prove.

sql/mob_spawn_points.sql
```diff
@@ -50573,8 +50573,8 @@ INSERT INTO `mob_spawn_points` VALUES (17350930,'Furies','Furies',28,-185,-10,48
 INSERT INTO `mob_spawn_points` VALUES (17350931,'Furies','Furies',28,-191,-10,51,127);
 
 -- Petrifying Pair
-INSERT INTO `mob_spawn_points` VALUES (17350933,'Kalamainu','Kalamainu',29,-193,-9,50,95);
-INSERT INTO `mob_spawn_points` VALUES (17350934,'Kilioa','Kilioa',30,-185,-10,42,105);
+INSERT INTO `mob_spawn_points` VALUES (17350933,'Kalamainu','Kalamainu',29,-184,-10,45,95);
```
TeoTwawki, line 50576: these pos come from retail captures? If not and the captures discord doesn't have any for it, we can get some pretty easily :+1:  if they are from a cap, awesome and ty.

dallano: they were from retail youtube videos, yes. They were previously set into a tree that would cause them to not agro at all in my findings. Spawn positions are most likely within 1-2 yalms accuracy.

TeoTwawki: video is not what is meant by captures. usually, a video accompanies pictures, but we get precise positional data down to 0.0001 yalms from _packet captures_

The capture discord has guides on this stuff.

TeoTwawki: apologies for not realizing you didn't know the lingo sooner

dallano: oh no it's worries! I'm trying to learn as fast as I can. sorry if I come off as a little dense lol. just started ~2 weeks ago

dallano: but no live captures were found in the captures discord.

TeoTwawki: kk, we'll hafta to get some then. positions are among the easiest thing to cap, so it shouldn't take long.

dallano: kk! Want me to revert to what was set previously? or keep this PR open and wait for them?

TeoTwawki: if it were going to take a long while I'd say change back, but I don't think it'll be long. we can just wait.

Issue #2346, opened 2022-07-19
🔨 Make zone:queryEntitiesByName() fast


zach2good:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Describe the feature

We're currently getting a lot of use and safety out of `zone:queryEntitiesByName()` but the current implementation is $O(n)$. We need to bring this down to _at worst_ $O(log n)$, but more likely $O(1)$.

We'll probably want to populate a `std::unordered_map<std::string, std::vector<CLuaEntity>>` at startup, and use that for our lookups.

This however won't work for dynamic entities, so we'd also need a way to force the linear search by name.

jmcmorris:
This is more or less resolved in https://github.com/LandSandBoat/server/pull/2754. `zone:queryEntitiesByName()` now uses memoization so subsequent calls are simply a map lookup which I believe is _O(logn)_.

zach2good:
`find` in `unordered_map` is $O(1)$ on average case, so even better :)

https://cplusplus.com/reference/unordered_map/unordered_map/find/

PR #2347, opened 2022-07-19
[Abyssea] Enable Altepa's QM's Triggers and Trades

Xaver-DaRed:
**_I affirm:_**

Enables NM mob spawning for Abyssea-Altepa from QM interactions (Trigger and Trade when applicable)

!zone to Abyssea-Altepa and interact with the QMs with the applicable items or Key Items


claywar (approved):
Looks good to me.  One thing I might suggest is perhaps we could change the NPC names for each of those QMs to be more descriptive (`qm_vadleany`, for example).

For IDs being looked up dynamically, I think that will come at a later day when #2346 is complete.

zach2good:
I'd argue that dynamic lookup for these is fine right now, the performance might only be a problem if it's multiple lookups a second

zach2good:
(as mentioned in discord, I'm planning on doing some plumbing work to make these lookups fast, this PR can go ahead and I'll modify later)

zach2good (approved):

PR #2348, opened 2022-07-19
Update TCP/UDP Settings to Reside in the new network.lua file

jmcdaniel83:
**_I affirm:_**

This utilizes the existing new LUA settings files for the TCP/UDP settings.  The settings from the two previous files `src/common/packet_tcp.conf` and `src/common/packet_udp.conf` have been removed in favor of the new settings setup.  `src/common/socket.cpp` has been updated to leverage the new `settings.h` file to pull specific setting parameters to make this process much cleaner and consistent throughout the code and makes the deployment much easier.  The only potential issue is that currently we no longer would be able to support imports from other TCP/UDP setting files.

I've also added some comments in `src/map/zone.cpp` in order to identify which navmesh files are not properly loading.

I've gone through the newly updated `settings/default/network.lua` file and validated that modifying the options is affecting who can log in to the server (allowed or denied) and have been able to toggle debug messages off and on.  Have also tested turning off the IP rules entirely and made sure that was working as well.


zach2good:
Cool! I was too scared to rip these files apart during the settings refactor. Nice work 👍

Xaver-DaRed:
Gonna need a rebase to solve conflicts

jmcdaniel83:
have merged from `base` branch into my fork to complete pull request

zach2good:
Could you rebase onto the latest `base`  removing the "revert" commit, instead of merging? That way you'll get to keep your commits. Otherwise, we can just squash when we merge.

jmcdaniel83:
Just to make sure that I don't mess up anything, do you want me to do the following on my fork?

Before the `[8d273f5]` commit, perform the following
```
git rebase [what]
```

We can squash it for the moment since I don't want to mess up anything but, just for future reference, just need to know how to properly rebase from a fork into the main `base` branch.

WinterSolstice8:
the basic instructions are:

`git remote add upstream git@github.com:LandSandBoat/server.git` (only need to do this once)
->
`git fetch` (have to do this every time you want to rebase
->
`git rebase upstream/base`
you'll need to do some googling on it to do what you need, i've never removed commits in this fashion before

cocosolos:
If you ever just want to look back at something like the last 9 commits and drop/edit/squash/rearrange/whatever some of them, you can do `git rebase -i HEAD~9` and it'll bring up your text editor and you can change what you need to then force push. It gets a little harder when you have other branches merging like here though. If you want to keep your commits in order and include new commits from upstream you can use `git rebase --onto <new_parent> <old_parent>` and it'll replay your commits on top of the new parent instead of merging them in and creating a mixed history. This is a little more complicated IMO and you'll probably want to look up [better instructions](https://git-scm.com/docs/git-rebase) before attempting that.

zach2good:
If you want to be a filthy casual like me, I find the VSCode git workflow to be pretty easy to use also

settings/default/network.lua
```diff
@@ -31,4 +31,79 @@ xi.settings.network =
     -- Central message server settings (ensure these are the same on both all map servers and the central (lobby) server
     ZMQ_IP   = "127.0.0.1",
     ZMQ_PORT = 54003,
+
+    -- ===========================
```
zach2good, line 35: Would you mind adding an additional comment block around here, saying that the "below settings are advanced options and it is very unlikely that users will need to modify them", or something to that effect?

Once these are in plainer view of regular users, I have no doubt that people will start to fiddle with them and start coming to us with all sorts of evil networking issues

zach2good (approved):

settings/default/network.lua
jmcdaniel83: That's a good call

zach2good (approved):
Great, thanks!

PR #2367, opened 2022-07-21
[IDs] Update messages documentation with missing messages

Schaedelschaden:
**_I affirm:_**

Adds missing messages to documentation/message.log.

### DIFF
--- a/documentation/message.log
+++ b/documentation/message.log
@@ -1,417 +1,421 @@
-001 - The <player> hits <target> for .. points of damage
-002 - a <player> casts (NULL). <target> takes .. points of damage.
-003 - The <player> starts casting Curaga IV.
-004 - <target> is out of range.
-005 - Unable to see <target>.
-006 - The <player> defeats <target>.
-007 - The <player> casts (NULL). <target> recovers .. HP.
-008 - The <player> gains 0 experience points.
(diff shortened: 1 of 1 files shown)


Schaedelschaden:
Not sure if all of the unassigned messages need to go into msg.lua, but at the very least the documentation can be updated and converted from a .log file to a .txt

TeoTwawki:
> Not sure if all of the unassigned messages need to go into msg.lua, but at the very least the documentation can be updated and converted from a .log file to a .txt

So what we've done is to add them to the lua/c++ enums on an add needed basis or as we expect to near future use them. doesn't mean we'd necessarily reject additions is properly named though. And co updates always good.

TeoTwawki:
it looks like you renamed a file outside of git (like renamed in windows) and so it's committed a delete of the file and add of a new file of the same content. that breaks the history record. Also adding to what I said about adding to the lua / enums there's not much point of basically just copying the entire document into it as a gigantic series of comments - essentially an unusable form. Let's leave that to the doc.

claywar (approved):

Schaedelschaden:
Thanks for the clarification Teo!

Xaver-DaRed (approved):

Issue #1198, opened 2022-01-21
Shantotto II not generating her own TP


RAIST5150:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: base branch, on xiweb, client /ver reporting  30211225_1 (but it was synched to retail 01/20/22)

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Xiweb server was updated today to include recent PR for Shantotto II to add regular attacks and WS gambits, as well as January's client shifts (among other things).  

Client was also synched to retail tonight as well.

Shantotto II attacks consistently for damage, cast spells, etc. as expected--but she was not generating any TP on her own.

As such, unable to comment on WS gambits.  Even after 6 fights, she did not grab enough hate to get smacked around any either... was still sitting at 0 TP.

TeoTwawki:
> [5:15 PM] Strong: Ahh maybe a new issue I noticed, August and Shantotto II do not receive TP using the attack LUA's. Setting mob:addTP() to the weaponskill sets TP at a flat rate with no increase. Any ideas?
[5:18 PM] TeoTwawki: bug
[5:18 PM] TeoTwawki: the bug being that their "mobskill" which is their melee, is eating tp
[5:19 PM] TeoTwawki: I thought we had it in tracker but sicne we still did not fix guessing it was in previous github repo

zach2good:
For future reference (for August, Shantotto II, and Ajido-Marujido etc.), marking the skill as "special" (flag 4) will fix this. This is what I did for Monberaux's skill in this PR: https://github.com/LandSandBoat/server/pull/1200

TeoTwawki:
@zach2good that field was deprecated 6 years ago and should not be used. Melee specials belong in an entirely separate skill list from regular skills and the core needs and adjustment.

TeoTwawki:
> ~~enum SKILLFLAG
{
    SKILLFLAG_NONE           = 0x000,
    SKILLFLAG_TWO_HOUR       = 0x002, // Special skill (ranged attack / call beast)
[...]

PR #2378, opened 2022-07-22
Hostile herbivores

dallano:
Basis and evidence are based off the following:
https://www.youtube.com/watch?v=2q93do8wQKQ&ab_channel=WiggoCaptures
https://drive.google.com/open?id=1lAiUP1bGNvEBVpIFZFshmzjonNK89Cvv

**_I affirm:_**

- Added a unique sheep charge ability for the fighting sheep that has no prepare time as seen in the video. mob skill ID based off of live capture.
- Sheep given an auto attack of sheep charge that deals knockback

Enter Hostile Herbivores bcnm in Horlais Peak to test.


scripts/zones/Horlais_Peak/mobs/Fighting_Sheep.lua
```diff
@@ -12,6 +12,10 @@ entity.onMobInitialize = function(mob)
     mob:setMod(xi.mod.ICE_MEVA, 75) -- Todo: Move to mob_resists.sql
 end
 
+entity.onMobSpawn = function(mob)
+    mob:addMod(xi.mod.REGAIN, 3000)
```
CriticalXI: Another way to go about this which I believe would be more accurate would be to replace their auto attacks with this ability. You can do this by passing `mob:SetMobSkillAttack(####)` with the number being the skill list ID.

dallano: oooooo I like that. I'll get on it when I can

TeoTwawki: Note you'll need two skill lists because of this, one for their autoskill-as-melee, and one for their other skills as they actually can use sheep song if they get sufficient TP to do so through being attacked. Probably lamb chop and rage as well but I haven't personally seen it. Unlike most autoskills, SE didn't suppress the "uses" message on these guys so you don't have to create a duplicate skill at least (different params would have been needed to suppress the message).

dallano: As far as I can tell from the capture in discord they don't use any other skills. But in other youtube videos I can find they use all other sheep skills minus regular sheep charge.

dallano: ah nevermind! they seem to also use sheep song.

TeoTwawki: > As far as I can tell from the capture in discord they don't use any other skills. 

Captures will only pick up on what happened at the time a capture was being done. This is why we have to do so much capturing to get everything.

> But in other youtube videos I can find they use all other sheep skills minus regular sheep charge.

:+1: 

> ah nevermind! they seem to also use sheep song.

I'd personally seen sheep song so I knew that one, just didn't know for sure on the others :)

CriticalXI: One issue that comes in with this is that even though its an auto attack, replacing it with a skill uses up the mob's TP every time they auto attack. Encountered this with Geush Urvan never using TP moves because of me doing the same thing. I had to do this to get him to be able to use TP moves: 

```
entity.onMobWeaponSkillPrepare = function(mob, target)
    mob:setLocalVar("skill_tp", mob:getTP())
end

entity.onMobWeaponSkill = function(target, mob, skill)
    if skill:getID() == ### then 
        mob:addTP(mob:getLocalVar("skill_tp"))
        mob:setLocalVar("skill_tp", 0)
    end
end
```

The skillID in there would be the ID of the move the sheep are using to auto attack with.

dallano: Thanks, crit! I'll put this in asap

TeoTwawki: > One issue that comes in with this is that even though its an auto attack, replacing it with a skill uses up the mob's TP every time they auto attack. 

Thought someone pr'd to fix that bug. if not it needs fixed instead of worked around, I'll try and make it higher priority so contributors don't have to worry about it.

TeoTwawki: according to [this comment](url) setting teh "special" value in the mobs_skills param prevents the TP loss, but then we would need either a separate entry of sheep charge in the global skill list (not this specific mobs list) which we can only do if multiple exist in the dat, or a method of flagging it just for this mob.

According to the dat, these are the skills available to us, with the +256 offset we use in the sql table:
```
(0006)  262 Sheep Charge
(0018)  274 Sheep Charge
(1379) 1635 Sheep Charge
(3603) 3603 Sheep Charge
```
do you know which one is used in the captures of this mob?

edit: typo'd 284 there

dallano: 274 was the param of their auto attack (sheep charge) in the capture from discord

scripts/zones/Horlais_Peak/mobs/Fighting_Sheep.lua
```diff
@@ -12,6 +12,21 @@ entity.onMobInitialize = function(mob)
...
+entity.onMobWeaponSkillPrepare = function(mob, target)
+    mob:setLocalVar("skill_tp", mob:getTP())
+end
+
+entity.onMobWeaponSkill = function(target, mob, skill)
+    if skill:getID() == 274 then
+        mob:addTP(mob:getLocalVar("skill_tp"))
+        mob:setLocalVar("skill_tp", 0)
+    end
+end
```
zach2good: I don't understand whats happening here? Is this to preserve the mob's TP while using the special melee mobskills?

If the mob's TP is being erased on these special melee attacks, that's a bug that we need to fix, rather than for you to work around

zach2good: https://github.com/LandSandBoat/server/issues/1198

TeoTwawki: if the previous conversation had not been resolved, you might have seen where I already mentioned this but with more detail. the fact github keeps allowing people other than us to resolve the conversations is maddening.

TeoTwawki: https://github.com/LandSandBoat/server/pull/2378#discussion_r928151545

scripts/zones/Horlais_Peak/mobs/Fighting_Sheep.lua
TeoTwawki: It needs a skill ID, not a param. 274 wouldn't be the right skill ID. I'm not sure what's happening there.

scripts/zones/Horlais_Peak/mobs/Fighting_Sheep.lua
zach2good: If you wanted to get your hands dirty, I think the fix here is something like this:
https://github.com/LandSandBoat/server/blob/base/src/map/ai/controllers/mob_controller.cpp#L349-L352

```cpp

    if (currentDistance <= PMobSkill->getDistance())
    {
        auto mobSkill = MobSkill(PActionTarget->targid, PMobSkill->getID());
        if (wsList == PMob->getMobMod(MOBMOD_ATTACK_SKILL_LIST))
        {
            mobSkill.setTP(0);
        }
        return mobSkill;
    }
```

scripts/zones/Horlais_Peak/mobs/Fighting_Sheep.lua
dallano: I can link the timestamp if you'd like to take a look

TeoTwawki: param isn't 274, action ID is. I'll get back to you after I know wtf is happening because 274 is freaking memento mori not sheep charge.  (╯°□°）╯︵ ┻━┻ _grumbles something about SE spaghetti code and wanders off_
image

TeoTwawki: wondering if packet viewer is adding the 256 offset we use in our ID lists. if it is that'd put it at 18, which is a legit sheep charge that I typo'd as _284_ above


...I expect packets to contain SE's values, not ours, so I had not thought of this.

scripts/zones/Horlais_Peak/mobs/Fighting_Sheep.lua
TeoTwawki: @dallano 
I want you to try 
```sql
INSERT INTO `mob_skills` VALUES (274,18,'sheep_charge',0,7.0,2000,1000,4,16,0,0,0,0,0); -- Sheep charge as melee attack
```
Adding that to its skill list for its melee attack, and tell me if it still strips TP. If it does, then try 20 where the 16 is, and try again. Normally for these, we would zero out that 1000 to silence the message (Iron Giants etc) but per the video, the message is included so this should retain it.

scripts/zones/Horlais_Peak/mobs/Fighting_Sheep.lua
dallano: This does prevent TP loss, yes. However they're not gaining any TP from attacking. Which I guess would be expected. Another side effect is it has caused the damage to go from ~40 to ~400

scripts/zones/Horlais_Peak/mobs/Fighting_Sheep.lua
dallano: This does prevent TP loss, yes. However they're not gaining any TP from attacking. Which I guess would be expected. Another side effect is it has caused the damage to go from ~40 to ~400

TeoTwawki: > This does prevent TP loss, yes. However they're not gaining any TP from attacking. Which I guess would be expected. Another side effect is it has caused the damage to go from ~40 to ~400

we may need to rename this 2nd instance (I'd go with `'sheep_charge_melee'`) and adjust its script a bit. I also suspect the tp gain from attacking with "special" melee like this may need handled on a case by case basis anyway - some might have different gains than usual or lack thereof, but at the script level we can just `mob:addTP()` as needed if we go that route. Other LSB members may have differing opinions and I am partly thinking out loud ..er.. in text. There hasn't been a major effort to study the tp gains on different mobs with this yet. I wouldn't worry about that yet. The dmg drop is more of a concern as nothing here should have changed that at all.

TeoTwawki:
would you like me to rebase this for you after I get out of work today?

dallano:
> would you like me to rebase this for you after I get out of work today?

sure! I've been stumbling through github lmao

TeoTwawki:
github can't do it, a local git client has to. i will try and record

dallano:
Critical has been helping me out a ton. But rebasing and managing dozens of different branches at the same time is difficult

TeoTwawki:
might be worth your while to waste hdd space on full separate clones in diff folders just to keep yourself from accidently mixing the content the way you've been

dallano:
yeah perhaps! I've been trying to push to ASB as well, so I rebase to push for them. But then I forgot there was still a PR open here, so I should have cloned it. And now it's all a mess lol.

WinterSolstice8:
Maybe I'm lacking context here, but you know if you merge something here ASB will get it when they merge upstream (LSB). Is there something different you need to do to cater to ASB? If we solve that problem you won't need to maintain two branches

dallano:
Well I'm trying to get content pushed into the database asap so I can start throwing testers at things for horizon (I'm implementing all the BCNM / KSNMs) So the sooner I can get things in the better. ASB usually merges my stuff within a few hours and can go straight to the testers.
Genuinely curious, and also have very little idea how github works, but can you guys merge my BCNM changes from ASB to LSB using something like cherry-pick? That way the content would be tested and confirmed to be working accurately and it wouldn't require as much work dealing with my terrible code lol

TeoTwawki:
> using something like cherry-pick? 

that results in conflicts and duplicated commits. its a bad workflow.

TeoTwawki:
> So the sooner I can get things in the better. 

it's usually the opposite in my experience. the faster things get rammed through the more disasters a server will have. Best of luck to them. >.>

TeoTwawki:
its a really weird situation we're in, we take consideration of you folks on downstream servers that you never have any idea we're covertly saving people from themselves. there are things we could in theory do to bring in code but it would then give the server operator an aneurism next time they pull, so we don't do that. But since they have no idea about it, they often wonder why we can't just xyz. Its unfortunately difficult to communicate all the potential disasters people almost do to themselves daily. I've seen the worst of the worst happen, seen operators throw their hands up and shut down over things they never knew were their own fault. And many a contributor give up after getting their forks in the exact state yours is now. 

We try our best to prevent tho.

dallano:
yeah, no I get it. And I truly don't intend to step on anyone's toes or be a nuisance here. I'm simply over eager to help out and haven't learned common etiquette working with big databases like this.

TeoTwawki:
> etiquette working with big ~~databases~~ repo's like this.

This actually isn't that big. messy maybe. We've inherited a lot from years of iteration, both good and bad. Check out unreal engine's repo. That thing is HUGE. Scary huge!

Edit: a word

TeoTwawki:
Should be better now

Vid of what I did: https://www.twitch.tv/videos/1550500996

TeoTwawki:
@dallano 
You should be good to remove the bit about the TP  in the mob script now BTW, the melee skill should not consume tp anymore.

TeoTwawki:
> @dallano You should be good to remove the bit about the TP in the mob script now BTW, the melee skill should not consume tp anymore.

nvm, took care of it myself

TeoTwawki (approved):

Xaver-DaRed (approved):

PR #2379, opened 2022-07-22
Final Bout implementation

dallano:
Basis and evidence for changes:
https://www.youtube.com/watch?v=ISRajaJ96b0&ab_channel=BlargedyMcgee
https://www.dropbox.com/s/v8vr3xfp2zwq74n/The%20Final%20Bout.rar?dl=0

**_I affirm:_**

Changes:
- Waughroon Kid will now use blow twice in a row. This behavior was witnessed in the linked live capture
- Added a counterstance ability for Waughroon Kid that behaves the same as the ability for players (increase counter rate and lower defence). These effects can also be witnessed in the video linked.
- Waughroon Kid will only use counterstance after reaching 40% hp. This behavior can be witnessed in the linked live capture
- Updated Waughroon Kid's skill list to include only blow and beatdown

Enter Waughroon Shrine BCNM The Final Bout to test


sql/mob_spawn_points.sql
```diff
@@ -51601,7 +51601,7 @@ INSERT INTO `mob_spawn_points` VALUES (17367232,'Tartaruga_Gigante','Tartaruga G
 INSERT INTO `mob_spawn_points` VALUES (17367234,'Tartaruga_Gigante','Tartaruga Gigante',33,222.866,-59.812,137.969,124);
 
 -- The Final Bout
-INSERT INTO `mob_spawn_points` VALUES (17367236,'The_Waughroon_Kid','The Waughroon Kid',34,-177,60,-141,125);
+INSERT INTO `mob_spawn_points` VALUES (17367236,'The_Waughroon_Kid','The Waughroon Kid',34,-177,60,-142,125);
```
Xaver-DaRed, line 51604: From the capture you provided:
```
x = -177.354
y =   59.696
z = -142.000
r =  128
```
So end result would be:
```
INSERT INTO `mob_spawn_points` VALUES (17367236,'The_Waughroon_Kid','The Waughroon Kid',34,-177.354,59.696,-142,128);
```

Other than that, the rest looks reasonable.

sql/mob_skills.sql
```diff
@@ -1163,7 +1163,7 @@ INSERT INTO `mob_skills` VALUES (1326,988,'final_retribution',1,10.0,2000,1500,4
 -- INSERT INTO `mob_skills` VALUES (1328,1072,'ink_jet',0,7.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (1329,990,'gala_macabre',1,10.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (1330,1074,'hoof_volley',0,7.0,2000,1500,4,0,0,2,0,0,0);
--- INSERT INTO `mob_skills` VALUES (1331,1075,'counterstance',0,7.0,2000,1500,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (1331,437,'counterstance_waughroon_kid',0,7.0,2000,0,4,0,0,0,0,0,0);
```
TeoTwawki, line 1166: Is there any reason this couldn't keep its original name, and then IF any other mobs happen to use counterstance (diff from Orcish counterstance) we wouldn't need multiple scripts to do essentially the same thing.

```
(00448)   704   Counterstance
(01075)  1331  Counterstance (this one)
(01173)  1429  Counterstance
```

TeoTwawki: I don't know of other mobs doing this to compare with, but the dats say theres at least 3 so that's why I'm asking. The way you did it may be better - I don't know.

PR #2384, opened 2022-07-24
Feature/blu scadoo - BLU Spells to 99

phoenik:
**_I affirm:_**

I'm still in the middle of testing, but since this is a large PR I figured I'd open in now for anyone else that wants to check it out.

- [x] Adds all missing BLU spells
- [x] Refactors BLU Traits with values that follow along with retail 
- [x] Creates LUAs for all BLU spells not yet implemented with information from bg-wiki where available
- [x] Implements BLU JP Gift for trait bonuses 
- [x] Audited BLU spells for status effect tick minimum requirement of 3 

## Important notes
- **Fantod** is not properly implemented. It just copies MNK boost at the moment.
- Values not available from bg-wiki are either best guessed based on other spells or defaulted to basic values
- BLU Breath spells need another look at since they are HP based and it looks like the calculations aren't based on that. Most later breath spells are just a copy of Heat Breath.
- Animations for missing BLU spells are all set to 2 seconds and may need auditing.
- When increasing BLU job points, current active traits are not refreshed and require a job change or spell assignment to gain the newly received gift.

- Change BLU spells and verify traits are properly given.
- Cast BLU spells and verify effects, animation, and messaging.
- Check trait values when under the effect of any BLU JP Gift.
- Check spell **Magic Barrier** gives new magic stoneskin type effect.


phoenik:
This does *NOT* include implementing the equivalent mob abilities and connecting them with learning.

zach2good:
I appreciate all the work that's gone into this PR, but it's an unreasonable amount for us to properly review (and test) all in one go (4700+ lines) . I'd recommend submitting this in smaller chunks, something like this:
- C++ changes, and the minimum required sql to support them
- the rest of the sql, commented out if they would enable lua files that don't exist (yet)
- any new lua utils and globals
- drip feeding in the new spells and their sql entries, maybe 10 at a time

TeoTwawki:
> Fantod is not properly implemented. It just copies MNK boost at the moment.

Then please don't include it.

As for everything else, I hope you've read the discussion thread on the subject of new blue scripts before I get around to reviewing. https://github.com/LandSandBoat/server/discussions/204 The state of existing Blue Magic spell scripts is _very bad_ from things just being guessed at and I do not want more of it going in nor the existing wrong stuff being copy pasted to fill in missing ones. There is a little bit of leeway on at least making them too weak instead of too strong, but we are in desperate need of someone to actually get the correct values into existing scripts checking on an actual retail account where needed, and possibly changes to the underlying system in cases where it can't yet be done.

Also, this is too many in one pull request. This should be several pull requests.

phoenik:
I'll close and move some things around.

PR #2385, opened 2022-07-24
[core] [SMN] [lua] Implement "Petskill" system, add basic Siren implementation

WinterSolstice8:
**_I affirm:_**

adds "Petskill" system, which is largely a copy paste of mobskill, but distinctly different in how categories are used for action packets. This allows for proper usage of pet models & pet mobskill IDs in packets.
Implements Siren at a basic level -- welt, favor only.

- Summon siren
- Use Avatar's Favor
- See `!getmod subtle_blow` climbs with time
- Use Welt on a mob and have it work + deal damage

## Notes
petskill interface is subject to change as I am not sure what is needed or not. Ideally the interface looks identical to any ability in lua but core can be more lightweight.


src/map/ai/states/petskill_state.h
```diff
@@ -0,0 +1,67 @@
+﻿/*
```
zach2good, line 1: New files can be attributed to LandSandBoat Dev team 2022

src/map/ai/ai_container.cpp
```diff
@@ -253,6 +265,16 @@ bool CAIContainer::Internal_MobSkill(uint16 targid, uint16 wsid)
     return false;
 }
 
+bool CAIContainer::Internal_PetSkill(uint16 targid, uint16 abilityid)
+{
+    auto* entity{ dynamic_cast<CPetEntity*>(PEntity) };
+    if (entity)
```
zach2good, line 271: If auto* entity = dynamic cast...
...

src/map/lua/lua_ability.h
```diff
@@ -44,9 +44,11 @@ class CLuaAbility
     uint16 getID();
     int16  getMsg();
     uint16 getRecast();
+    uint16 getRecastID();
     uint16 getRange();
     auto   getName() -> const char*;
     uint16 getAnimation();
+    uint16 getAddType();
```
zach2good: What is add type? Additional type? Could use full version of the name

src/map/lua/luautils.cpp
```diff
@@ -3789,6 +3792,39 @@ namespace luautils
         return result.get_type(0) == sol::type::number ? result.get<int32>(0) : 0;
     }
 
+    int32 OnPetAbility(CBaseEntity* PTarget, CPetEntity* PPet, CPetSkill* PPetSkill, CBaseEntity* PMobMaster, action_t* action)
+    {
+        TracyZoneScoped;
+
+        std::string filename = fmt::format("./scripts/globals/abilities/pets/{}.lua", PPetSkill->getName());
```
zach2good, line 3799: Do these hot reload correctly?

src/map/utils/battleutils.cpp
```diff
@@ -290,6 +336,17 @@ namespace battleutils
         }
     }
 
+    /************************************************************************
+     *  Clear Pet Skills List                                                *
+     ************************************************************************/
+    void FreePetSkillList()
+    {
+        for (auto iter = g_PPetSkillList.begin(); iter != g_PPetSkillList.end();)
```
zach2good, line 344: 👍 Like this pattern

zach2good (commented):
Lots of good in here, if this is the only way to fix pet job abilities, then I guess we have to do it (I assume this models retail more correctly than we do now, and Teo signs off on it)

src/map/ai/ai_container.cpp
WinterSolstice8: Looks like copy/paste from Internal_Mobskill:

```
bool CAIContainer::Internal_MobSkill(uint16 targid, uint16 wsid)
{
    auto* entity{ dynamic_cast<CMobEntity*>(PEntity) };
```

will fix both

src/map/lua/lua_ability.h
WinterSolstice8: addType is the same as MobSkill's addType, the intent here was to make the 2 types as close to the same as possible so if someone understands mobskill they should understand petskill pretty easily too. I can add a comment about it

WinterSolstice8:
> Lots of good in here, if this is the only way to fix pet job abilities, then I guess we have to do it (I assume this models retail more correctly than we do now, and Teo signs off on it)

This may not be the "only" way -- the core difference between this and a regular mobskill is you can choose what the action packet category ends with. mobskills right now are hardcoded and that's bad juju for SMN. There is also the fact that there are duplicates of each avatar BP for animations and it may get confusing when the pet ability and the mob ability are in the same sql table. The intent was to make a concrete difference between "these are skills only for mobs" and "these are skills only for pets", right now the lines are blurred

src/map/lua/luautils.cpp
WinterSolstice8: They do, it's sourced the same way as current pet abilities used by SMN are. Can confirm that I can edit Welt in real time and it reloads.

src/map/lua/lua_ability.h
TeoTwawki: it being called addType is really the only thing that snags for me, but don't have a naming suggestion either. good enough I guess.

TeoTwawki (approved):

scripts/globals/ability.lua
```diff
@@ -704,6 +704,22 @@ xi.specEffect =
     CRITICAL_HIT   = 0x22,
 }
 
+-- addType, used in ability:getAddType(). the addType of an ability is defined in sql.
+xi.addType =
```
zach2good, line 708: Happy with this pr, would subtype be a good name for this in the future?

zach2good (approved):

PR #2402, opened 2022-07-26
[JAs] Rename BERSERK_EFFECT mod, sort mods, and add notes

Schaedelschaden:
**_I affirm:_**

Rename BERSERK_EFFECT mod to BERSERK_POTENCY in modifier.h, status.lua, and mods_by_id.txt.
Reorganize Warrior mod positions in modifier.h and status.lua lists.
Update job_utils/warrior.lua with renamed mod.
Update item_mods.sql with renamed mod.
Add note to effects/berserk.lua regarding DEFP latent effect from Warrior's Calligae NQ/+1/+2.
Add note to effects/berserk.lua regarding CRIT_HIT_RATE & DOUBLE_ATTACK latent effects from Conqueror (all forms).

This PR is primarily intended as housekeeping and to clarify the mod used by the Conqueror GAxes.
Conqueror i119 III is the only version which does not refer to it as Augments "Berserk".

**References**

https://www.bg-wiki.com/ffxi/Berserk
https://www.bg-wiki.com/ffxi/War._Calligae_%2B2
https://www.bg-wiki.com/ffxi/Conqueror_(Level_119)
https://www.bg-wiki.com/ffxi/Conqueror_(Level_119_III)

2. Note base ATK and DEF before using Berserk.
3. Use `!setmod 948 8 <me>` to set BERSERK_POTENCY's value to 8 (i119 III Conqueror level).
4. Use Berserk.
5. ATK should be 33% higher and DEF should be 33% lower.


TeoTwawki:
> Rename BERSERK_EFFECT mod to AUGMENT_BERSERK

that exact naming was avoided deliberately, because items that say "augments berserk" might not be the only things to ever use it -and- people often confuse it with being an actual augment instead of a modifier (and I hate SE for using that word for 2 things)

Schaedelschaden:
Aaah, ok, that makes sense. I'm right there with you on how SE decided to name these things, it's insanely confusing.

Should I leave this PR open and reduce it to only clarification in the notes or close it entirely?

TeoTwawki:
> Should I leave this PR open and reduce it to only clarification in the notes or close it entirely?

only objection is that rename, everything else is fine :+1: 

if you can think of a better name that describes what the mods does, that'd be fine to. tbh "`_effect`" isn't much better than "`augments_`"

Schaedelschaden:
The first thing that springs to mind is BERSERK_MYTHIC_BONUS. So far, the only things that provide that bonus are the Mythic GAxes.

TeoTwawki:
Per BG it "gives the following:"

> Attack to +30% (±77/256), Defense -30% (Upgraded from the original Berserk effect of +25%/-25%)
Effect on Attack persists even after Conqueror removal.
Critical Hit Rate +5%
Effect on Critical Hit Rate does not persist when Conqueror is removed.

I dunno if it'd make sense to say "potency" when it also gives crit rate but that was my 1st thought.

¯\\\_(ツ)_/¯

Schaedelschaden:
Potency would make sense to me and it's a lot easier than `_MYTHIC_BONUS`. The crit hit rate bonus is already handled by Berserk latents on the Conquerors, so the only thing that would need to be updated should be the names.

TeoTwawki:
oh. well I guess that works then :+1:

sql/item_mods.sql
```diff
@@ -44161,11 +44161,11 @@ INSERT INTO `item_mods` VALUES (21756,506,130); -- EXTRA_DMG_CHANCE: 130
 INSERT INTO `item_mods` VALUES (21756,507,200); -- OCC_DO_EXTRA_DMG: 200
 INSERT INTO `item_mods` VALUES (21756,659,40);  -- WSD+% (Metatron Torment): 40
 
--- Conqueror (119 III)
```
TeoTwawki, line 44164: I think we should retain the previous verion of the item name comments so that we can tell apart _which_ version of Conqueror it is at a glance

Schaedelschaden: Nice catch! I had undone the mod rename and didn't get those back in there.

I have added the level comments back in the same format that BG Wiki has them here:
https://www.bg-wiki.com/ffxi/Conqueror

TeoTwawki (approved):

WinterSolstice8 (approved):

PR #2405, opened 2022-07-26
Add Inthius script to Eastern_Adoulin

Tom-Neverwinter:
**_I affirm:_**

This is a copy pasta of the other weather NPC's with its appropriate event id added. 

This completes the current wiki listed weather NPC's

Talk to Inthius 
https://ffxiclopedia.fandom.com/wiki/Inthius

### DIFF
--- a/scripts/zones/Eastern_Adoulin/npcs/Inthius.lua
+++ b/scripts/zones/Eastern_Adoulin/npcs/Inthius.lua
@@ -5,19 +5,18 @@
 -- !pos -97.145 0.001 -36.710 257
 -----------------------------------
 local entity = {}
- 
+
 entity.onTrade = function(player, npc, trade)
 end
- 
+
 entity.onTrigger = function(player, npc)
     player:startEvent(4, 0, 0, 0, 0, 0, 0, 0, VanadielTime())
 end
- 
+
 entity.onEventUpdate = function(player, csid, option)
 end
- 
+
 entity.onEventFinish = function(player, csid, option)
 end
- 
+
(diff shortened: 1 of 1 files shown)


scripts/zones/Eastern_Adoulin/npcs/Inthius.lua
```diff
@@ -0,0 +1,22 @@
+-----------------------------------
+-- Area: Eastern Adoulin
+-- NPC: Inthius
+-- Type: Weather Reporter
+-- !pos -97.3680 0.0000 -34.3807
```
zach2good: From `npc_list.sql`: use these values (and the zone number):

`!pos -97.145 0.001 -36.710 257`

The intention for _all_ of these comments is that you can copy/paste them into the game and travel to that NPC from any zone.

zach2good (changes requested):

scripts/zones/Eastern_Adoulin/npcs/Inthius.lua
Tom-Neverwinter: I always set it in front of the NPC, if you want it directly on top of the npc I can do that as well

https://pastebin.com/41EhLYG3

TeoTwawki: I think best practice is just copy paste form the npc list, because otherwise someone might think it's the npc list that's wrong and try to "fix it" since everywhere else these have always matched.

zach2good (approved):

PR #2413, opened 2022-07-27
[Fix] Debug crash when a not PC performs a Jump

Xaver-DaRed:
**_I affirm:_**

As title says

Summon Shikaree Z Trust and wait until she jumps. See no error in log. The other ftp warning is unrelated.

### DIFF
--- a/scripts/globals/job_utils/dragoon.lua
+++ b/scripts/globals/job_utils/dragoon.lua
@@ -75,7 +75,13 @@ end
 
 -- Generic Function for damage-based Jumps
 local function performWSJump(player, target, action, params)
-    local taChar = player:getTrickAttackChar(target)
+    local taChar = nil
+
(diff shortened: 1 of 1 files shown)


scripts/globals/job_utils/dragoon.lua
```diff
@@ -75,7 +75,13 @@ end
 
 -- Generic Function for damage-based Jumps
 local function performWSJump(player, target, action, params)
-    local taChar = player:getTrickAttackChar(target)
+    local taChar = nil
+
+    -- Check if preformer is PC to avoid (HIT DEBUG CONDITION: m_PBaseEntity->objtype != TYPE_PC) in lua_baseentitity.
```
claywar, line 80: Going to fix this in core, as I'm not a huge fan of breaking instead of sanely logging.

Xaver-DaRed:
After a conversation with Teo, and getting to know that some Trusts do Trick Attack, and Claywars PR #2414 this is obsolete and was not the proper fix anyway.

PR #2417, opened 2022-07-28
[Abyssea] Enable Grauberg's QM's Triggers and Trades

Xaver-DaRed:
**_I affirm:_**

Enables NM mob spawning for Abyssea-Grauberg from QM interactions (Trigger and Trade when applicable)

!zone to Abyssea-Grauberg and interact with the QMs with the applicable items or Key Items

### DIFF
--- a/scripts/zones/Abyssea-Grauberg/IDs.lua
+++ b/scripts/zones/Abyssea-Grauberg/IDs.lua
@@ -71,6 +71,28 @@ zones[xi.zone.ABYSSEA_GRAUBERG] =
     },
     mob =
     {
+        IRONCLAD_SUNDERER            = 17818041,
+        NINGISHZIDA                  = 17818042,
+        TEUGGHIA                     = 17818043,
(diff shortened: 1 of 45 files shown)


claywar (approved):

scripts/zones/Abyssea-Grauberg/IDs.lua
```diff
@@ -71,6 +71,28 @@ zones[xi.zone.ABYSSEA_GRAUBERG] =
     },
     mob =
     {
+        IRONCLAD_SUNDERER            = 17818041,
```
claywar, line 74: These could probably be moved to dynamic lookups, but if so, the ID local in those scripts would need to pull data from the zones global table

Xaver-DaRed: The idea was to convert the closed PRs first and then get back to this. (Hopefully I will have gotten the hang of the dynamic lookup by then)

PR #2424, opened 2022-07-29
BCNM30 Petrifying Pair

TeoTwawki:
**_I affirm:_**
- [:x:] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

Revives #2340 
Went on retail and got the pos with a friend and reverted those delay changes till we have confirmed correct delay values.


TeoTwawki:
From the run I just did, if anything the delay looks higher, not lower, than 240. Leaving it alone unless very convincing proof to the contrary.

Xaver-DaRed (approved):
They did seem to resist sleep, without being inmune.
No other debuff resisted (that was tested, minimal testing at that)

This was tested using a lvl 30 rdm, capped skill, no subjob, no extra stats from gear

TeoTwawki:
> They did seem to resist sleep, without being inmune. No other debuff resisted (that was tested, minimal testing at that)
> 
> This was tested using a lvl 30 rdm, capped skill, no subjob, no extra stats from gear

I think pretty much every scripted instance of resist mods need evaluated again once we have status resists working retail accurate. Right now it's just bonus meva and basically worthless (otherwise this would have been more resistant than should be, instead of less than should be)

PR #2425, opened 2022-07-30
Royal jelly ASB implementation

dallano:
**_I affirm:_**

Changes:

Removed Princess Jellies skill set.
Gave Princess Jellies weakness depending on their element
Added SCRIPT variable to pathfinding to allow the jellies to path to the center of the room (Can be used for a multitude of other mobs)
Slightly increased pre-existing spell cool down on Princes Jellies as they were casting faster than evidence found in youtube videos.
Removed some unused functions from code ported.
Increased the distance at which the jellies will cast on players to 20 yalms. (Previously set at 15 which would allow players to cast on jellies without any retaliation)
Slightly decreased the odds a Princess Jelly would cast bind by 10% to 20% (Review of past videos proved 33% was too high of odds - which was the previous rate on code from Setzor)
Implemented the drop tables for Royal Jelly which was taken from ffxiclopedia with a sample rate of ~400+
Cleaned up table formatting in Princess_Jelly.lua
Fixed Queen Jelly from spawning in the initiation of the battlefield
Moved some modifiers being added onto princess and queen luas into mob pools SQL
Added spell lists for both Princess and Queen Jelly


WinterSolstice8:
p.s. if you scroll down when making the PR you can see how many files changed before clicking the make PR button

PR #2431, opened 2022-07-31
Adjust SMN BP to use current TP when using BPs, initialize uninitialized variable in mob/petskill

WinterSolstice8:
**_I affirm:_**

While implementing siren I noticed random/garbage TP values cropping up from this call in lua: `ability:getTP()`, which eventually reads from this variable: `m_spentTP`

That variable is _only_ set here:
```
void CMobSkillState::SpendCost()
{
    if (m_PSkill->isTpSkill())
    {
        m_spentTP  = m_PEntity->health.tp;
        m_PEntity->health.tp = 0;
    }
}
```

And what is `isTpSkill()` you might ask?

```
bool CMobSkill::isTpSkill() const
{
    return !isSpecial() && !isAttackReplacement();
}
```

isSpecial is a flag, 0x04 in `mob_skill_flag` in the db. At a glance it seems that _all_ BPs set the special flag. Therefore, m_spentTP was whatever garbage was in memory when creating MobSkillState/PetSkillState. Aside from that, avatars (and mobs while we're at it) don't use a _stale_ value of when the move was readied for TP, but their current TP value. 

Drafting for comments while I test every BP affected to see if they still work.

### DIFF
--- a/scripts/globals/abilities/pets/aero_ii.lua
+++ b/scripts/globals/abilities/pets/aero_ii.lua
@@ -14,7 +14,7 @@ end
 
 ability_object.onPetAbility = function(target, pet, skill)
     local dINT = math.floor(pet:getStat(xi.mod.INT) - target:getStat(xi.mod.INT))
-    local tp = skill:getTP()
+    local tp   = pet:getTP()
 
     local damage = math.floor(45 + 0.025*(tp))
     damage = damage + (dINT * 1.5)
--- a/scripts/globals/abilities/pets/aero_iv.lua
+++ b/scripts/globals/abilities/pets/aero_iv.lua
@@ -14,7 +14,7 @@ end
 
 ability_object.onPetAbility = function(target, pet, skill)
     local dINT = math.floor(pet:getStat(xi.mod.INT) - target:getStat(xi.mod.INT))
-    local tp = skill:getTP()
+    local tp   = pet:getTP()
 
     local damage = math.floor(325 + 0.025*(tp))
     damage = damage + (dINT * 1.5)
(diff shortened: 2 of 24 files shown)


WinterSolstice8:
No script errors encountered during testing, marking as ready for review

TeoTwawki:
> isSpecial is a flag, 0x04 in `mob_skill_param` in the db. 

wrong field
```sql
  `mob_skill_flag` tinyint(1) unsigned NOT NULL DEFAULT '0',
  `mob_skill_param` smallint(5) NOT NULL DEFAULT '0',
```
`mob_skill_flag` holds flags while `mob_skill_param` is sort of arbitrary, used and misused for things like potency of a skill. It's a leftover from when DSP allowed multiple copies of a skill to exist regardless of what was in the client dat, so you could make a stronger version of pollen by just putting in a 2nd copy with the param set higher, because the script checked the param. The field is still useful for a few things so it never got removed.

> At a glance it seems that _all_ BPs set the special flag. Therefore, m_spentTP was whatever garbage was in memory when creating MobSkillState/PetSkillState. Aside from that, avatars (and mobs while we're at it) don't use a _stale_ value of when the move was readied for TP, but their current TP value.

related footnote: melee subs (deprecated flag of 0x10) behave similarly to "special" flag 0x04. Somewhere in the future both were supposed to die, which is a good reason imo for pet things to use their own flagging _eventually_.

WinterSolstice8:
> > isSpecial is a flag, 0x04 in `mob_skill_param` in the db.
> 
> wrong field

oops. edited.

Xaver-DaRed (approved):

zach2good (approved):

claywar (approved):

PR #2435, opened 2022-08-01
[LUA] Create job_utils/monk and migrate JA functionality

Schaedelschaden:
**_I affirm:_**

Create job_utils/monk.lua and populates necessary ability check and ability use functions.
Moves functionality from all Monk globals/abilities files to globals/job_utils/monk.lua.

2. Load the server.
3. `!changejob MNK 99`

### DIFF
--- a/scripts/globals/job_utils/monk.lua
+++ b/scripts/globals/job_utils/monk.lua
@@ -9,6 +9,15 @@ xi = xi or {}
 xi.job_utils = xi.job_utils or {}
 xi.job_utils.monk = xi.job_utils.monk or {}
 
+local chakraStatusEffects =
+{
+    POISON       = 0, -- Removed by default
+    BLINDNESS    = 0, -- Removed by default
+    PARALYSIS    = 1,
+    DISEASE      = 2,
+    PLAGUE       = 4,
+}
+
 -----------------------------------
 -- Ability Check Functions
 -----------------------------------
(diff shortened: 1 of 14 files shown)


Schaedelschaden:
I have all of the Monk JA functionality moved to the job_utils file and have updated all of the JA files to run through it. Should I use this PR to port all of the functionality to job_utils or would that be too intensive from the testing standpoint?

Xaver-DaRed:
I say go for it.

scripts/globals/job_utils/monk.lua
```diff
@@ -0,0 +1,143 @@
...
+        local effect = player:getStatusEffect(xi.effect.BOOST)
+        effect:setPower(effect:getPower() + power)
+        player:addMod(xi.mod.ATTP, power)
+    else
+        player:addStatusEffect(xi.effect.BOOST, power, 0, 180)
+    end
+end
+
+xi.job_utils.monk.useChakra = function(player, target, ability)
+    local chakraStatusEffects =
```
claywar: Just for clarity, could this be defined outside of the function in this global?

claywar (approved):

scripts/globals/job_utils/monk.lua
Schaedelschaden: Since the variable is only used for Chakra I’d put it into Chakra’s function but can definitely move it out into the global. Should it go above all of the other ability functions or directly above Chakra’s function?

claywar: I'm a fan of above all of the other functions, just in case for some (most likely no) reason in the future, it needs to be changed.

Xaver-DaRed (approved):

PR #2439, opened 2022-08-01
Cleanup mobskill flagging and give more appropriate naming to current usage.

TeoTwawki:
**_I affirm:_**

Renames some flags and functions, strips some flags and functions that were uneeded or unused, updates sql fields accordingly:
- `SKILLFLAG_DRAW_IN` was used nowhere
- `SKILLFLAG_REPLACE_ATTACK` was shifted away from, and is now handled by a function to load the skill list as a melee attack, accordingly `isAttackReplacement()` is no longer needed.
- `isTpSkill()` simply checked for melee replacement -or- the special flag, and was what decided to take TP or not. Both it and `isSpecial()` have been replaced by `isTpFreeSkill()` 

pull build and reimport mobskills table (not skill lists) go to Kons. Highlands and provoke a sheep with godmode up so it can't build tp on you. Use `!tp` to set its TP to some number below the amount it could use but higher than 0. Use `!exec` to force it to use a skill that has been flagged `SKILLFLAG_NO_TP_COST` and check that its TP was not consumed. Now feed the sheep TP and let us is that TP for real. Check it again and its TP is zero right after.

Next, go to qufim and harass some gigas! The ranger ones should throw boulders every few seconds. this should also not cost them any TP. After that, feed them some TP till they use an actual skill like grand slam. You will have to be at a distance for the boulder, but up close for grandslam. Observe their TP was taken as it should be.

### DIFF
--- a/sql/mob_skills.sql
+++ b/sql/mob_skills.sql
@@ -106,7 +106,7 @@ INSERT INTO `mob_skills` VALUES (270,14,'roar',1,10.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (271,15,'razor_fang',0,7.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (272,16,'ranged_attack',0,25.0,2000,0,4,4,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (273,17,'claw_cyclone',4,15.0,2000,1500,4,0,0,0,0,0,0);
--- INSERT INTO `mob_skills` VALUES (274,18,'sheep_charge',0,7.0,2000,1500,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (274,6,'sheep_charge',0,7.0,2000,1000,4,4,0,0,0,0,0); -- Sheep charge as melee attack
 INSERT INTO `mob_skills` VALUES (275,809,'sand_blast',1,8.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (276,810,'sand_pit',0,7.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (277,811,'venom_spray',4,10.0,2000,1500,4,0,0,0,0,0,0);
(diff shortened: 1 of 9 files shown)


sql/mob_skills.sql
```diff
@@ -106,7 +106,7 @@ INSERT INTO `mob_skills` VALUES (270,14,'roar',1,10.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (271,15,'razor_fang',0,7.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (272,16,'ranged_attack',0,25.0,2000,0,4,4,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (273,17,'claw_cyclone',4,15.0,2000,1500,4,0,0,0,0,0,0);
--- INSERT INTO `mob_skills` VALUES (274,18,'sheep_charge',0,7.0,2000,1500,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (274,6,'sheep_charge',0,7.0,2000,1000,4,4,0,0,0,0,0); -- Sheep charge as melee attack
```
TeoTwawki, line 109: This is legit fyi. For the Sheep in Hostile Herbivores BCNM.

claywar (approved):

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -342,7 +342,7 @@ bool CMobController::MobSkill(int wsList)
 
         PActionTarget = luautils::OnMobSkillTarget(PActionTarget, PMob, PMobSkill);
 
-        if (PActionTarget && !PMobSkill->isTwoHour() && luautils::OnMobSkillCheck(PActionTarget, PMob, PMobSkill) == 0) // A script says that the move in question is valid
+        if (PActionTarget && !PMobSkill->isAstralFlow() && luautils::OnMobSkillCheck(PActionTarget, PMob, PMobSkill) == 0) // A script says that the move in question is valid
```
WinterSolstice8, line 345: Is this "isAstralFlow" even needed here? Will anything even attempt to use an AF ability through this method?

TeoTwawki: no idea, I just renamed the bastid for now, revisit to murder the F outa it later coz the answer -should- be no; but you never know what's flagged or being checked for the flag.

WinterSolstice8 (approved):
Looks like CI is complaining because of the commented unused variables, so the longer bp rage/ward isn't "aligned" to the block above it, or so it thinks

zach2good:
TIL You can drop the output of the clang-format job in a code block with `diff` as the language, and it'll show you it nicely

```diff
diff --git a/src/map/mobskill.h b/src/map/mobskill.h
index e1f4ed59e2..b71a0c57c6 100644
--- a/src/map/mobskill.h
+++ b/src/map/mobskill.h
@@ -27,10 +27,10 @@
 
 enum SKILLFLAG
 {
-    SKILLFLAG_NONE           = 0x000,
-    SKILLFLAG_ASTRAL_FLOW    = 0x002, // Player's Avatar Astral Flow blood pacts. TODO: give player pet skills their own separate enum, move avatar stuff there.
-    SKILLFLAG_NO_TP_COST     = 0x004, // Don't auto deduct TP
-    SKILLFLAG_HIT_ALL        = 0x008, // Strike players even if not in party/alliance
+    SKILLFLAG_NONE        = 0x000,
+    SKILLFLAG_ASTRAL_FLOW = 0x002, // Player's Avatar Astral Flow blood pacts. TODO: give player pet skills their own separate enum, move avatar stuff there.
+    SKILLFLAG_NO_TP_COST  = 0x004, // Don't auto deduct TP
+    SKILLFLAG_HIT_ALL     = 0x008, // Strike players even if not in party/alliance
     // unused                = 0x010,
     // unused                = 0x020,
     SKILLFLAG_BLOODPACT_RAGE = 0x040,
```

PR #2462, opened 2022-08-05
[LUA] Create job_utils/black_mage, migrate and update JA functionality

Schaedelschaden:
**_I affirm:_**

- Create job_utils/black_mage.lua and populates necessary ability check and ability use functions.
- Moves functionality from all Black Mage globals/abilities files to globals/job_utils/black_mage.lua.
- Adds functionality to Enmity Douse.
- Updates Manawell to target self and party members.
- Removes -ENMITY mod from Subtle Sorcery effect onEffectTick (Ability's enmity effect will need further supporting work).
- Removes erroring delMod ENMITY from Subtle Sorcery effect onEffectLose (Ability's enmity effect will need further supporting work).
- Updates notes for lua_baseentity.cpp setCE and setVE functions no longer used exclusively by Ventriloquy.

**References**
https://www.bg-wiki.com/ffxi/Enmity_Douse
https://www.bg-wiki.com/ffxi/Manawell
https://www.bg-wiki.com/ffxi/Subtle_Sorcery

2. Load the server.
3. Use the !changejob BLM 99 command.

### DIFF
--- a/scripts/globals/effects/subtle_sorcery.lua
+++ b/scripts/globals/effects/subtle_sorcery.lua
@@ -6,6 +6,7 @@ require("scripts/globals/status")
 -----------------------------------
 local effect_object = {}
 
+-- Spell Cumulative Enmity reduction handled in magic_state.cpp
 effect_object.onEffectGain = function(target, effect)
     local jpValue = target:getJobPointLevel(xi.jp.SUBTLE_SORCERY_EFFECT)
 
--- a/scripts/globals/effects/subtle_sorcery.lua
+++ b/scripts/globals/effects/subtle_sorcery.lua
@@ -14,14 +14,12 @@ effect_object.onEffectGain = function(target, effect)
 end
 
 effect_object.onEffectTick = function(target, effect)
-    target:addMod(xi.mod.ENMITY, -20)
 end
 
 effect_object.onEffectLose = function(target, effect)
     local jpValue = target:getJobPointLevel(xi.jp.SUBTLE_SORCERY_EFFECT)
 
     target:delMod(xi.mod.MACC, 100)
-    target:delMod(xi.mod.ENMITY)
     target:delMod(xi.mod.UFASTCAST, jpValue)
 end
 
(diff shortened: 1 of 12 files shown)


scripts/globals/effects/subtle_sorcery.lua
```diff
@@ -14,14 +14,12 @@ effect_object.onEffectGain = function(target, effect)
 end
 
 effect_object.onEffectTick = function(target, effect)
-    target:addMod(xi.mod.ENMITY, -20)
```
Schaedelschaden, line 17: The tick isn't set up when the effect is added and this would give the user -20 ENMITY per tick. According to https://www.bg-wiki.com/ffxi/Subtle_Sorcery:
> Reduces cumulative enmity from magic spells to 0.

Handling for the enmity would need to set CE to 0 on spell cast. I'm not sure if there's a global way that the effect could be applied to spells.

scripts/globals/effects/subtle_sorcery.lua
```diff
@@ -14,14 +14,12 @@ effect_object.onEffectGain = function(target, effect)
...
 end
 
 effect_object.onEffectLose = function(target, effect)
     local jpValue = target:getJobPointLevel(xi.jp.SUBTLE_SORCERY_EFFECT)
 
     target:delMod(xi.mod.MACC, 100)
-    target:delMod(xi.mod.ENMITY)
```
Schaedelschaden, line 24: Throws an error because no value is specified.

scripts/globals/effects/subtle_sorcery.lua
claywar: This might be something that needs to be handled by `CMagicState::ApplyEnmity()`

scripts/globals/effects/subtle_sorcery.lua
claywar: This the addMod was removed, should have no impact from removing this value here.

scripts/globals/effects/subtle_sorcery.lua
Schaedelschaden: Set up `CMagicState::ApplyEnmity()` to check for Subtle Sorcery and set CE to 0 for all spells. In-game ability text states "Reduces the amount of enmity generated from magic spells" so the status check applies to any magic spell and is not restricted to elemental magic or enemy targeting only spells.

WinterSolstice8 (approved):
LGTM

PR #2499, opened 2022-08-11
[LUA] Correct Flee movement speed increase

N3ckB3ard:
**_I affirm:_**

Updates Flee to reflect retail movement speed increase to +60% rather than the +100% it is set to currently

Setting the modifier FLEE value to 61 within Flee.lua reflects +60% movement increase from 0%

Good Info:
Retail capture with addon to check movement speed
https://www.bg-wiki.com/ffxi/Flee
http://wiki.ffo.jp/html/1682.html

Bad Info:
https://ffxiclopedia.fandom.com/wiki/Flee_(Status_Effect)

Use Flee. Player should reflect a +60% movement speed increase

### DIFF
--- a/scripts/globals/abilities/flee.lua
+++ b/scripts/globals/abilities/flee.lua
@@ -15,7 +15,7 @@ end
 
 ability_object.onUseAbility = function(player, target, ability)
     local duration = 30 + player:getMod(xi.mod.FLEE_DURATION)
-    player:addStatusEffect(xi.effect.FLEE, 100, 0, duration)
+    player:addStatusEffect(xi.effect.FLEE, 60, 0, duration)
 end
 
 return ability_object
--- a/scripts/globals/abilities/flee.lua
+++ b/scripts/globals/abilities/flee.lua
@@ -1,6 +1,6 @@
 -----------------------------------
 -- Ability: Flee
--- Increases movement speed.
+-- Increases movement speed by +60%
 -- Obtained: Thief Level 25
 -- Recast Time: 5:00
 -- Duration: 0:30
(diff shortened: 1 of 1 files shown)


N3ckB3ard:
Time spent researching shows that the initial cap for all movement speed of players is set to +60% outside of chocobo cap of +100%

December 2013 Version Update increased movement speed of all players by +20% and altered the new "cap" increases. Implementation of the Movement speed system would allow us to hard cap movement speed without having to adjust each lua that might be incorrect.

My question is how to consider 1 item modifier for MOVE and not allow other equipped items from further increasing this int value for reflection of speed? Highest value takes priority and stacks with spells, abilities and KIs to a maximum of +60%

TeoTwawki:
sorry but I cannot merge this as is. our our flee is wrong, but your attempt to fix it isn't correct. Our modifier doesn't work this way and our overall speed handling need a rework for additional adjustments and clamping. when you pop flee in retail you aren't given a %boost. You are statically set to a speed value. I have already done work to PROVE this. The value the client holds in memory during flee is 80. which is double the original game speed of 40. When they updated player movement, they did not update flee. 

Mounts and chocobo meanwhile, use 40 but the client knows to "double it" without the server telling it to.

TeoTwawki:
> Time spent researching shows that the initial cap for all movement speed of players is set to +60% outside of chocobo cap of +100%
> 
> December 2013 Version Update increased movement speed of all players by +20% and altered the new "cap" increases. Implementation of the Movement speed system would allow us to hard cap movement speed without having to adjust each lua that might be incorrect.
> 
> My question is how to consider 1 item modifier for MOVE and not allow other equipped items from further increasing this int value for reflection of speed? Highest value takes priority and stacks with spells, abilities and KIs to a maximum of +60%

[Lets take it to a discussion thread,](#2501) and I can info dump you. I've been making my client spit out whats in memory for speed values for about a year and a half now.

PR #2515, opened 2022-08-13
Fix Dismounting sometimes causing players to become invisible

### LINKED ISSUE #2514: 🐛 Dismounting sometimes causes players to become invisible/untargetable to others

## Branch affected by issue 

`base`

## Steps to reproduce

1. Have two characters logged in, in the same zone. We'll call them X and Y. X must have the ability to mount (or access to the !mount command).
2. X mounts up. This should be a non-Chocobo mount.
3. X gets within 45 yalms of Y, such that Y can see X on a mount.
4. X leaves range of Y (more than 50 yalms) such that they stop rendering for Y.
5. X dismounts (while more than 50 yalms from Y).
6. X gets within 45 yalms of Y, such that X can see Y.
7. X will not be visible or targetable (not even the name will be visible) for Y.

## Expected behavior

7. X is visible and targetable to Y, as per normal.

Interestingly, this bug does not occur for Chocobo mounts. Only non-Chocobo mounts are affected. To my testing, the only ways for the bugged client to be able to see the invisible character again is either if the bugged client character zones, or if the invisible character is mounted within update distance of the bugged client. The invisible character leaving the zone and returning did not make them visible.

Tested on /ver 30181205_0, though I'm informed this bug has been present for years, since DSP days.

DiscipleOfEris:
**_I affirm:_**

Fixes #2514

This fix has a side effect of sending the associated dismount packet to all characters in the zone. I consider this low impact, but notable enough that I'd consider this fix temporary. A better solution would be to figure out where the packets the server is sending differ from retail.

1. Have two characters logged in, in the same zone. We'll call them X and Y. X must have the ability to mount (or access to the !mount command).
2. X mounts up. This should be a non-Chocobo mount.
3. X gets within 45 yalms of Y, such that Y can see X on a mount.
4. X leaves range of Y (more than 50 yalms) such that they stop rendering for Y.
5. X dismounts (while more than 50 yalms from Y).
6. X gets within 45 yalms of Y, such that X can see Y.
7. X is visible and targetable to Y, as per normal.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -1054,6 +1054,8 @@ void SmallPacket0x01A(map_session_data_t* const PSession, CCharEntity* const PCh
             PChar->animation = ANIMATION_NONE;
             PChar->updatemask |= UPDATE_HP;
             PChar->StatusEffectContainer->DelStatusEffectSilent(EFFECT_MOUNTED);
+            // Workaround for a bug where dismounting out of update range would cause the character to stop rendering.
+            PChar->loc.zone->PushPacket(PChar, CHAR_INZONE, new CCharPacket(PChar, ENTITY_UPDATE, UPDATE_HP));
         }
         break;
         case 0x13: // tractor menu


WinterSolstice8 (approved):
Thanks for the PR

DiscipleOfEris:
Happy to help. I fixed it on my server after someone said it's been a problem for years. I actually downloaded LSB just to check if this bug was fixed.

src/map/packet_system.cpp
```diff
@@ -1054,6 +1054,8 @@ void SmallPacket0x01A(map_session_data_t* const PSession, CCharEntity* const PCh
             PChar->animation = ANIMATION_NONE;
             PChar->updatemask |= UPDATE_HP;
             PChar->StatusEffectContainer->DelStatusEffectSilent(EFFECT_MOUNTED);
+            // Workaround for a bug where dismounting out of update range would cause the character to stop rendering.
+            PChar->loc.zone->PushPacket(PChar, CHAR_INZONE, new CCharPacket(PChar, ENTITY_UPDATE, UPDATE_HP));
```
TeoTwawki, line 1058: I think there must be a problem with the update mask in packet we send out when chars come into range. The reason your fix works is it's sending it to the entire zone regardless, but those players should receive that data when the char becomes within 50 yalm (we intentionally don't send out entity updates for things beyond that distance). If it works it works, better this than invisible players, but we'll have to figure the rest out later.

TeoTwawki: > This fix has a side effect of sending the associated dismount packet to all characters in the zone. I consider this low impact, but notable enough that I'd consider this fix temporary. A better solution would be to figure out where the packets the server is sending differ from retail.

Oh you already said as much. Well that will learn me to read code before the description. I shall now resume my work at the department of redundancy department.

DiscipleOfEris: I tested this and tried force setting the update mask to always include UPDATE_HP (in Windower it's called Update Vitals). The 0x00D packet with the flag forcefully set would get received (along with the corresponding Mount and Status fields), but the client would just ignore it.

Interestingly, this bug doesn't affect chocobo mounts. Only non-chocobo mounts. This leads me to believe it's not the update mask, but rather there's some element of the packet that is not constructed how the client expects. Either that, or the client requires the packet to include *only* UPDATE_HP and not also UPDATE_POSITION in order to properly register the dismount (but only for non-Chocobos), despite the combined update working for other cases.

In any case, I don't know anyone on retail so I can't test if the packet itself is part of the problem.

zach2good: If you describe what you need I can grab caps for you. This is rate-limited by the client, right? It isn't possible to hop on and off a mount any more than once every 30s? So it isn't really at risk of hurting performance

zach2good (approved):

src/map/packet_system.cpp
DiscipleOfEris: > If you describe what you need I can grab caps for you. This is rate-limited by the client, right? It isn't possible to hop on and off a mount any more than once every 30s? So it isn't really at risk of hurting performance

That would be great. To capture, you'll need a friend/second account with a non-Chocobo mount. I want the captures from the observing/passive character, *not* the mounting/dismount character.

I'd like two captures, with three extras if you're willing to be thorough. The important ones: dismounting out of range with a non-chocobo mount and in range with a non-chocobo mount. The extras: dismounting out of range but with a chocobo and in range with a chocobo, and finally just walking in range without having dismounted out of range.

Have two characters. We'll call the mounting character M, and the capturing character C. They should both be in a zone where mounting is allowed.

Dismounting a non-Chocobo out of range.
1. M mounts a non-Chocobo mount and gets within 45 yalms of C, such that C and M can see eachother.
2. M walks at least 50 yalms away from C, such that neither M nor C can see eachother.
3. M dismounts while out of update range of C.
4. M gets within 45 yalms of C, such that both can see eachother.

Dismounting a non-Chocobo in range.
1. M mounts a non-Chocobo mount and gets within 45 yalms of C, such that C and M can see eachother.
2. M dismounts while within update range of C.

Bonus. Dismounting a Chocobo out of range.
1. M mounts a Chocobo mount and gets within 45 yalms of C, such that C and M can see eachother.
2. M walks at least 50 yalms away from C, such that neither M nor C can see eachother.
3. M dismounts while out of update range of C.
4. M gets within 45 yalms of C, such that both can see eachother.

Bonus. Dismounting a Chocobo in range.
1. M mounts a Chocobo mount and gets within 45 yalms of C, such that C and M can see eachother.
2. M dismounts while within update range of C.

Bonus. Walking in range without dismounting.
1. M gets within 45 yalms of C, such that C and M can see eachother.
2. M walks at least 50 yalms away from C, such that neither M nor C can see eachother.
3. M gets within 45 yalms of C, such that both can see eachother.

The mount has a shared 1 minute cooldown according to the wiki, so this will probably take a minimum of five minutes. These five captures should reveal any differences between retail and LSB w.r.t. this bug, but if this is a bit too much, then just the first two (dismounting both in and out if range) should *hopefully* be sufficient.

DiscipleOfEris: I also posted the capture request in the FFXI Captures discord.

DiscipleOfEris: @zach2good

PR #2532, opened 2022-08-15
Fix the MAGIC_TAKE event always triggering twice

### LINKED ISSUE #2531: 🐛 Logic using MAGIC_TAKE listener is executed twice

## Branch affected by issue 

`base`

## Steps to reproduce

Spoke about this briefly in discord, here's what @DiscipleOfEris came up with:
```
There are two points in the code when it gets triggers
CMagicState::Update() and luautils::OnMagicHit()
luautils::OnMagicHit() gets called in CBattleEntity::OnCastFinished()
Yeah, looks like MAGIC_TAKE is getting called twice, in two different places at the same time.
MAGIC_TAKE is the only triggerListener in luautils 
```
## Expected behavior

The logic inside this function should only be executed once

DiscipleOfEris:
**_I affirm:_**

The `MAGIC_TAKE` event is emitted in `luautils::OnMagicHit()` *only* after `MAGIC_TAKE` is emitted in `CMagicState::Update()`. There's no reason to have the listener trigger in `luautils::OnMagicHit()` since it's always guaranteed to have already just triggered.

This removes the duplicate `MAGIC_TAKE` trigger.

Fixes #2531

1. In a lua script for a mob, do `mob:addListener("MAGIC_TAKE", "whatever", function(target, caster, spell) print("MAGIC_TAKE") end)`
2. Cast a spell on the mob.
3. Check the console for xi_map.exe
4. Notice only one "MAGIC_TAKE" printed.

Before the fix, there would be two "MAGIC_TAKE" printed.

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -2659,8 +2659,6 @@ namespace luautils
             return -1;
         }
 
-        PTarget->PAI->EventHandler.triggerListener("MAGIC_TAKE", CLuaBaseEntity(PTarget), CLuaBaseEntity(PCaster), CLuaSpell(PSpell));
-
         sol::function onMagicHit = getEntityCachedFunction(PTarget, "onMagicHit");
         if (!onMagicHit.valid())
         {


WinterSolstice8:
I can't check right this second, but my only concern with this is things like enspells & spikes maybe not triggering the event because it's not part of magic state.

DiscipleOfEris:
If that's the case, then perhaps the trigger in the MagicState should be removed. I'll test what you suggested. That said, is it the expected behavior that the damage from an enspell or spikes should trigger `MAGIC_TAKE`? Cause at the moment, `luautils::OnMagicHit()` is only called from `battleutils::OnCastFinished()`. So I don't think the enspells and spikes are getting a `MAGIC_TAKE` trigger anyway.

TeoTwawki:
> I can't check right this second, but my only concern with this is things like enspells & spikes maybe not triggering the event because it's not part of magic state.

the onMagicHit binding was specifically for spells only, and if we want triggers for on spikes or on add effects, we'd make additional separate listeners for those (an all in one is not desirable)

TeoTwawki (approved):

PR #2534, opened 2022-08-15
Fix Skillchains sometimes triggering before the 3 second window

### LINKED ISSUE #2533: 🐛 Skillchains can sometimes be accomplished without waiting

## Branch affected by issue 

`base`

## Steps to reproduce

You'll likely need two characters to see how "instant" it can be because of delay between one character's actions.

1. Use a Weaponskill on a valid target.
2. Wait 3 seconds.
3. Use a second Weaponskill that can SkillChain with the first.
4. Results in a SkillChain.
5. Wait a *maximum* of 2 seconds (the less wait, the more obvious).
6. Use a third Weaponskill that can SkillChain with the second.
7. Results in a SkillChain.

## Expected behavior

7. No SkillChain should happen because the delay was too short (should be a 3 to 10 second window).

DiscipleOfEris:
**_I affirm:_**

The code was not checking for the delay if it was a multi skillchain. That is, the 3rd or later WS in the skillchain.

Fixes #2533

See #2533

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -3752,7 +3752,6 @@ namespace battleutils
                 }
             }
             
-
             if (skillchain != SC_NONE)
             {
                 PSCEffect->SetStartTime(server_clock::now());
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -3729,28 +3729,29 @@ namespace battleutils
                 PSCEffect = PDefender->StatusEffectContainer->GetStatusEffect(EFFECT_SKILLCHAIN, 0);
             }
             // Previous effect exists
-            else if (PSCEffect && PSCEffect->GetTier() == 0)
+            else if (PSCEffect->GetStartTime() + 3s < server_clock::now())
             {
-                XI_DEBUG_BREAK_IF(!PSCEffect->GetPower());
-                // Previous effect is an opening effect, meaning the power is
-                // actually the ID of the opening weaponskill.  We need all 3
-                // of the possible skillchain properties on the initial link.
-                if (PSCEffect->GetStartTime() + 3s < server_clock::now())
+                if (PSCEffect && PSCEffect->GetTier() == 0)
                 {
+                    XI_DEBUG_BREAK_IF(!PSCEffect->GetPower());
+                    // Previous effect is an opening effect, meaning the power is
+                    // actually the ID of the opening weaponskill.  We need all 3
+                    // of the possible skillchain properties on the initial link.
                     auto properties = PSCEffect->GetPower();
                     resonanceProperties.push_back((SKILLCHAIN_ELEMENT)(properties & 0b1111));
                     resonanceProperties.push_back((SKILLCHAIN_ELEMENT)((properties >> 4) & 0b1111));
                     resonanceProperties.push_back((SKILLCHAIN_ELEMENT)((properties >> 8) & 0b1111));
                     skillchain = FormSkillchain(resonanceProperties, skillProperties);
                 }
+                else
+                {
+                    // Previous effect is not an opening effect, meaning the power is
+                    // The skill chain ID resonating.
+                    resonanceProperties.push_back((SKILLCHAIN_ELEMENT)PSCEffect->GetPower());
+                    skillchain = FormSkillchain(resonanceProperties, skillProperties);
+                }
             }
-            else
-            {
-                // Previous effect is not an opening effect, meaning the power is
-                // The skill chain ID resonating.
-                resonanceProperties.push_back((SKILLCHAIN_ELEMENT)PSCEffect->GetPower());
-                skillchain = FormSkillchain(resonanceProperties, skillProperties);
-            }
+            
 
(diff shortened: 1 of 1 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -3729,28 +3729,29 @@ namespace battleutils
                 PSCEffect = PDefender->StatusEffectContainer->GetStatusEffect(EFFECT_SKILLCHAIN, 0);
             }
             // Previous effect exists
-            else if (PSCEffect && PSCEffect->GetTier() == 0)
+            else if (PSCEffect->GetStartTime() + 3s < server_clock::now())
```
zach2good: Can SCEffect be nullptr here?

TeoTwawki: there's something else to consider here. we _perceive_ a 3 second delay on retail, but there's lag time between the player and Japan. Actually, is it 3 seconds? I've not timed it. I'm unsure if BG accounts for this but gives a very vague timeframe on its skillchains page near the top that def sounds like nobody actually documented real world tests. Potentially animation times of the involved weaponskills matter. I'd have never thought something so fundamental to the game would make me say "I don't know" but here I am, doubting myself.

zach2good: Sounds like a good candidate for a setting, but also, 2.4s is very suspiciously nearby

DiscipleOfEris: This is in an else of an else branch. To reach the first else branch, at least one of `PSCEffect` and `PCBEffect` must not be `nullptr`. To reach the second branch, `PCBEffect` must be `nullptr`. As such, to reach this branch `PSCEffect` cannot be `nullptr`. That said, the code might look nicer with a `nullptr` check anyway.

TeoTwawki: small clarification: it was 3 before the contributor got here, so I don't mean we need to check it before merging this. they are just making the timing it already said actually work when it wasn't. but we should in the future confirm the timing was even right.

zach2good: I'm on my phone, I can see the upper check now, thanks!

DiscipleOfEris: The SC delay should be relatively easy to test (for someone with a retail account), including if it's affected by the weaponskill and/or skillchain duration.

WinterSolstice8: you can't trust the unmodified client for the timing based on the log message for skillchains. some of them (like Transfixion/fragmentation) emit the message early, and others a bit later than most and you can screw up the skillchain timing. Better testing would need an addon to see exactly when the packet came through

this will get checked in the Great Rewrite:tm:

TeoTwawki: > Sounds like a good candidate for a setting, but also, 2.4s is very suspiciously nearby

as it happens time to japan is about 300ms. 300 there, 300 back. 600ms +2.4 seconds = 3 seconds!

DiscipleOfEris: @WinterSolstice8 Well yeah. I had exactly that in mind, since I do a lot of packet sniffing. That said, I suppose my definition of "easy to test" isn't exactly standard, so point.

DiscipleOfEris: I'd basically just use a script to control the input timings of the WSs involved, choosing some that have a low animation time. Then adjust the wait period down until the SC triggers about half the time.

So like, `//input /ws "Shining Strike" <t>; wait 2.9; send myalt input /ws "Gust Slash" <t>;`. Where I keep adjusting that 2.9 to find the sweet spot. Then switch to a different set of WSs to see if the window differs.

And perform the same test but for double SCs (using a SAM or three alts or something), to see if the animation duration of the SC affects the timing window of the 3rd WS.

DiscipleOfEris:
The CPP Formatting sanity check failed again... Anyone have a better approach to running these sanity checks locally before committing?

TeoTwawki:
> The CPP Formatting sanity check failed again... Anyone have a better approach to running these sanity checks locally before committing?

I have legit never tried to do that ever. ~~But your stuff looks fine and I am gonna say the checker here is wrong not you : its griping about some removed whitespace, and I _want_ whitespace to be removed so the checker can stuff it :stuck_out_tongue:~~

claywar:
That's what the error is showing.  CI puts out a diff for what was there, and what it expects it to be replaced with.  It's not saying add whitespace.

TeoTwawki:
> That's what the error is showing. CI puts out a diff for what was there, and what it expects it to be replaced with. It's not saying add whitespace.

Then the display is backwards from what is expected. The diff shows whitespace being removed, and we _want_ whitespace removed. When I look at CI things my expectation is to be shown "where's where you are wrong". Not "here is what to do", Thats just fuckin confusing.

TeoTwawki:
incidentally, just always trimming trailing when you save is a good practice to be in anyway

DiscipleOfEris:
I guess I understand how to read the sanity check now.

TeoTwawki:
> I guess I understand how to read the sanity check now.

doin better than me then, apparently! _mutter grumble_

zach2good (approved):

PR #2537, opened 2022-08-16
Fix wyverns disappearing on certain types of zoning

### LINKED ISSUE #2535: Player DRG: Wyvern is despawned upon zoning

## Branch affected by issue 

`base`

## Steps to reproduce

1. Be DRG
2. Call Wyvern
3. Zone
4. Wyvern is gone.

## Expected behavior

The wyvern is supposed to stay with the player until desummoned/logged out/KOd or the player is KOd.  1 exception to this (the only one I know of) is when zoning into a town, the mob is "put away" and comes back out when they leave the town into a field zone.

### LINKED ISSUE #2536: 🐛 Wyverns and other pets lost on certain types of zoning

## Branch affected by issue 

`base`

## Steps to reproduce

In retail, wyverns are saved when you go to town and returned when you leave town. I believe they're also maintained when transferring zones via NPC interactions (events), like the Crystalline Field in Al'Taieu. But in LSB wyverns and other pets are lost when transferring zones via an event and do not return when leaving town.

Part 1. Towns.

1. Be in a zone next to town with your pet out.
2. Zone into town.
3. Leave town.
4. Observe your pet not returning.

Part 2. Events.

1. Be next to the Crystalline Field in Al'Taieu with your pet out.
2. Use the Crystalline Field to zone.
3. Observe your pet abandoning you.

## Expected behavior

Part 1.

4. For wyverns, your wyvern returns.

Part 2.

3. For wyverns, your wyvern stays with you.

We need more information for different types of zoning with different pets to properly match retail. I would like to know how retail handles:

 * Zoning normally with each type of pet.
 * Going into town and leaving town each type of pet.
 * Zoning to non-level restricted zones with an NPC interaction with each type of pet.
 * Zoning into level restricted zones with each type of pet.
 * Teleporting with spell with each type of pet.
 * Teleporting with home point with each type of pet.
 * Zoning into battlefields with each type of pet.

I believe Wyverns are kept through everything. I believe avatars are not kept through anything. I think automatons are kept. I don't know about jug pets. Obviously charmed mobs aren't kept. I don't know anything about Trusts, Fellows, or Luopans.

DiscipleOfEris:
**_I affirm:_**

Wyverns, Automatons, and Avatars weren't being respawned when leaving town (if you entered with it) or when zoning via the `luautils::setPos()` function, which includes NPC interactions and the `!zone` command. They should now be retained in those instances.

Also, Jug Pets were being kept through normal zoning, but I've been informed that is not retail behavior so it will be removed by this PR as well.

Fixes #2536 
Fixes #2535 

See #2536 

More information on how this is handled by retail for different types of pets and different types of zoning would be valuable.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -2660,6 +2660,8 @@ void CLuaBaseEntity::setPos(sol::variadic_args va)
 
     if (m_PBaseEntity->objtype == TYPE_PC)
     {
+        auto* PChar = static_cast<CCharEntity*>(m_PBaseEntity);
+
         if (va[4].is<double>())
         {
             auto zoneid = va[4].as<uint16>();
@@ -2668,22 +2670,22 @@ void CLuaBaseEntity::setPos(sol::variadic_args va)
                 return;
             }
 
-            if (static_cast<CCharEntity*>(m_PBaseEntity)->PPet != nullptr)
+            if (PChar->PPet != nullptr)
             {
-                static_cast<CCharEntity*>(m_PBaseEntity)->setPetZoningInfo();
+                PChar->setPetZoningInfo();
             }
 
-            static_cast<CCharEntity*>(m_PBaseEntity)->loc.destination = zoneid;
-            static_cast<CCharEntity*>(m_PBaseEntity)->status          = STATUS_TYPE::DISAPPEAR;
-            static_cast<CCharEntity*>(m_PBaseEntity)->loc.boundary    = 0;
-            static_cast<CCharEntity*>(m_PBaseEntity)->m_moghouseID    = 0;
-            static_cast<CCharEntity*>(m_PBaseEntity)->clearPacketList();
-            charutils::SendToZone(static_cast<CCharEntity*>(m_PBaseEntity), 2, zoneutils::GetZoneIPP(m_PBaseEntity->loc.destination));
-            //static_cast<CCharEntity*>(m_PBaseEntity)->loc.zone->DecreaseZoneCounter(static_cast<CCharEntity*>(m_PBaseEntity));
+            PChar->loc.destination = zoneid;
+            PChar->status          = STATUS_TYPE::DISAPPEAR;
+            PChar->loc.boundary    = 0;
+            PChar->m_moghouseID    = 0;
+            PChar->clearPacketList();
+            charutils::SendToZone(PChar, 2, zoneutils::GetZoneIPP(PChar->loc.destination));
+            //PChar->loc.zone->DecreaseZoneCounter(PChar);
         }
-        else if (static_cast<CCharEntity*>(m_PBaseEntity)->status != STATUS_TYPE::DISAPPEAR)
+        else if (PChar->status != STATUS_TYPE::DISAPPEAR)
         {
-            static_cast<CCharEntity*>(m_PBaseEntity)->pushPacket(new CPositionPacket(static_cast<CCharEntity*>(m_PBaseEntity)));
+            PChar->pushPacket(new CPositionPacket(PChar));
         }
     }
     m_PBaseEntity->updatemask |= UPDATE_POS;
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -2681,7 +2681,7 @@ void CLuaBaseEntity::setPos(sol::variadic_args va)
             PChar->m_moghouseID    = 0;
             PChar->clearPacketList();
             charutils::SendToZone(PChar, 2, zoneutils::GetZoneIPP(PChar->loc.destination));
-            //PChar->loc.zone->DecreaseZoneCounter(PChar);
+            // PChar->loc.zone->DecreaseZoneCounter(PChar);
         }
         else if (PChar->status != STATUS_TYPE::DISAPPEAR)
         {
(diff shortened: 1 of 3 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -2668,6 +2668,11 @@ void CLuaBaseEntity::setPos(sol::variadic_args va)
                 return;
             }
 
+            if (((CCharEntity*)m_PBaseEntity)->PPet != nullptr)
```
claywar: Just for consistency, can the C-style casts please be converted to static_cast?

`if (static_cast<CCharEntity*>(m_PBaseEntity)->PPet != nullptr)`

DiscipleOfEris: Done. Touched up the code in the pet respawning section while at it, just so the block all matches.

src/map/lua/lua_baseentity.cpp
```diff
@@ -2668,17 +2668,22 @@ void CLuaBaseEntity::setPos(sol::variadic_args va)
...
             }
 
-            ((CCharEntity*)m_PBaseEntity)->loc.destination = zoneid;
-            ((CCharEntity*)m_PBaseEntity)->status          = STATUS_TYPE::DISAPPEAR;
-            ((CCharEntity*)m_PBaseEntity)->loc.boundary    = 0;
-            ((CCharEntity*)m_PBaseEntity)->m_moghouseID    = 0;
-            ((CCharEntity*)m_PBaseEntity)->clearPacketList();
-            charutils::SendToZone((CCharEntity*)m_PBaseEntity, 2, zoneutils::GetZoneIPP(m_PBaseEntity->loc.destination));
-            //((CCharEntity*)m_PBaseEntity)->loc.zone->DecreaseZoneCounter(((CCharEntity*)m_PBaseEntity));
+            if (static_cast<CCharEntity*>(m_PBaseEntity)->PPet != nullptr)
```
claywar: Looking at this now, we should probably follow the example used in the upper block, and not cast each time, but instead declare PChar in this one as well (`auto* PChar = static_cast<CCharEntity*>(m_PBaseEntity);`)  That'll reduce the bulk of those repeated casts, and just replace with the new var.

WinterSolstice8: Should be `dynamic_cast<CPetEntity*>(m_PBaseEntity)->PPet != nullptr`
PPet will never be a character but could be a mob or a pet, and we only care if pets can zone. To the best of my knowledge this applies to all "real" pets, e.g. avatar/wyvern/automaton but unsure about call beast. I can test this real quick and will report back on if we need to check pet type for jug pets

dynamic_cast is done at runtime, which is important if there's a distinction between pets and mobs like this instance

WinterSolstice8 (dismissed):

src/map/lua/lua_baseentity.cpp
WinterSolstice8: Wait, my brain isn't working from just getting home. I think Clay's comment is right.

WinterSolstice8:
Jug pets don't persist between zone, reverifying Avatars

WinterSolstice8:
Avatars persist on zone too

WinterSolstice8 (approved):
Looks good, if this still works after your edits this should be fine to merge

DiscipleOfEris:
> Avatars persist on zone too

Is there a difference in pet persistence between the different methods of zoning? Zoning normally, zoning via an NPC event, zoning into town and then out, etc.

DiscipleOfEris:
@WinterSolstice8 See #2536 where I suggest the types of things to check retail.

WinterSolstice8:
test was: zone to south gustaberg, summon carby, go back in bastok, zone back out, carby still there. Job change (or subjob) should remove the pet, but not simply zoning into MH. HP warping elsewhere (to sky) carby still showed up. Same behavior as DRG wyvern/automaton. Essentially once they're out they persist until job/subjob change or death

edit: the master's death OR their death, to be clear.

DiscipleOfEris:
Doesn't seem to be working now. Let me test.

EDIT: Nevermind, only avatars aren't working. Automatons and wyverns still persist through zoning. I'm guessing the avatar is getting deleted when the mp goes to zero during the zone process.

TeoTwawki:
> Jug pets don't persist between zone, reverifying Avatars

_cries in beastmaster's gil_

DiscipleOfEris:
How many ticks of perpetuation cost are paid while zoning with an avatar? And is this the same as the number of regen/refresh ticks?

WinterSolstice8:
None is accumulated when zoning back in, cost appears to only be deducted on tick when it's out in the zone

TeoTwawki:
> How many ticks of perpetuation cost are paid while zoning with an avatar? And is this the same as the number of regen/refresh ticks?

That will depend on how long the zoneout processing takes. Your character is still "active" on retail for a few milliseconds on the black screen before the zone in starts. At the zone you are arriving in nothing should happen/count till the fade in, where char is already loaded and stats exist. To do this right we won't want to try any kind of timer - this is going to be variable lag depending, so we really just want to make sure we're not checking till after the char is loaded again

DiscipleOfEris:
Okay. This PR also has the (intended) side effect of stopping Jug Pets from zoning with you. Avatars should be maintained through zoning now.

zach2good (approved):

PR #2549, opened 2022-08-18
Fix Caederva Mire Logging Point IDs

ozuwozu:
**_I affirm:_**

The Caedarva Mire Logging table has incorrect IDs which causes uninteractable NPCs to appear in trees. This fixes the IDs to start at the Logging Point IDs nearby in the npc_list.sql.
image

1. Go to the coordinates listed for the logging point entries.
2. Verify that the trees contain logging points and not Nashmeira.


zach2good (approved):
Thanks for the fix 👍

claywar (approved):

PR #2044, opened 2022-06-18
[COP] Sea NM System T2 + JOL

The-Aerec:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

Note this PR requires: canUseAbility from -> https://github.com/LandSandBoat/server/pull/1773

(Has been tested with this addition, and then removed so conflicts don't occur)

Adds mechanics for the following mobs and their associated pets:

Jailer of Hope
Jailer of Prudence
Jailer of Justice
Jailer of Love

Mechanics, immunities, etc come from the following sources: (As before, primarily captures from the FFXICaptures disc. And filled in unknowns via bg-wiki and other videos.)

Jailer of Hope:
https://youtu.be/iYBs2tzEzmo
https://www.youtube.com/watch?v=V9xRnhdGdgA
https://www.youtube.com/watch?v=oRO2WSEhE9M

Jailer of Prudence:
https://youtu.be/xzo8C3LI2BE
https://www.youtube.com/watch?v=fvQeahXGI3k
https://www.youtube.com/watch?v=QIFoJwpEBIM

Jailer of Justice:
https://youtu.be/aAtqdiBGp5w
https://www.youtube.com/watch?v=7E4qWnPS5G8
https://www.youtube.com/watch?v=wgkf1uToypE

Jailer of Love:
https://youtu.be/zxZNmitQp2s
https://www.youtube.com/watch?v=F0Ewb96Yg2Y
https://www.youtube.com/watch?v=PqZ69gIkVC4
https://www.youtube.com/watch?v=9rrIDx4nigw

## Steps to test these changes

Spawn any of the mobs and fight!

PR #2578, opened 2022-08-22
[core] Improve CLuaBaseEntity::injectActionPacket()

WinterSolstice8:
**_I affirm:_**

Improves functionality of CLuaBaseEntity::injectActionPacket to be "limitless" in what you can do on a single target.

Needed for #2044  and possibly other future NMs like Plouton who use actions to update the client's model or simply play an animation via action packet with no side effects.

see attached lua command, [yoinkitem.zip](https://github.com/LandSandBoat/server/files/9437305/yoinkitem.zip)

use !yoinkitem on a target to "steal" a cool item, either a nice ring or a player's main weapon, which they then unequip and it looks like it worked.

image
Yoink!

### DIFF
--- a/scripts/commands/injectaction.lua
+++ b/scripts/commands/injectaction.lua
@@ -21,6 +21,18 @@ function onTrigger(player, actionId, animationId, speceffect, reaction, message)
         return
     end
 
+    if actionId == 0 or actionId > 15 then
+        error(player, "<action ID> is out of range. Current valid and tested action IDs are 1-15.")
+        return
+    end
+
+    local target = player:getCursorTarget()
+
+    if not target then
+        error(player, "No valid target found")
+        return
+    end
+
     -- validate animationId
     if animationId == nil then
         error(player, "You must provide an animation ID.")
@@ -41,5 +53,5 @@ function onTrigger(player, actionId, animationId, speceffect, reaction, message)
     end
 
     -- inject action packet
-    player:injectActionPacket(actionId, animationId, speceffect, reaction, message)
+    player:injectActionPacket(target:getID(), actionId, animationId, speceffect, reaction, message, 10, 1)
 end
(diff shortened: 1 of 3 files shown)


TeoTwawki (approved):

claywar (approved):

zach2good:
Forge is a very interesting name for this, why not create?

TeoTwawki:
> Forge is a very interesting name for this, why not create?

I read it as forge like a signature, not forge as in what a blacksmith uses. Makes sense given we're faking things with it.

zach2good:
Ultimately, this is the same sort of thing as `injectActionPacket`, should this replace it and take its name?

TeoTwawki:
> Ultimately, this is the same sort of thing as `injectActionPacket`, should this replace it and take its name?

Possibly. It eliminates some of the shortcomings from it. where inject was made for test purposes only this is meant for more and may evolve over time if we add safeties/defaults/fallbacks to it later. Inject has a few hardcoded things and more than you can't specify, where as with forge you _must_ specify all the things. maybe combine them another day and update calls?

WinterSolstice8:
my intent was "forge" as in "to counterfeit", because these packets aren't tied behind any system and could do anything. I don't really mind what name it takes, but indeed, this is essentially an "upgraded" version of injectActionPacket that lets you do anything. Several parameters in this PR's new function are hardcoded in injectActionPacket

WinterSolstice8:
I think I'm just going to replace injectActionPacket with this and rewrite the `!injectaction` command to have the safeties instead. No real reason to have both when safety can be accomplished on the command.

WinterSolstice8:
Alright, give it a review again. I've replaced the commit with a new one that simply improves injectActionPacket instead.

claywar (approved):

zach2good (approved):

TeoTwawki (approved):

zach2good:
Will this be able to fix issue with Trust: Nanaa Mihgo where she steals but the message comes out as garbage because "reasons"?

WinterSolstice8:
Is there some specific instance where that bugs out? random test shows it looks fine
image

zach2good:
Huh, looks like it got fixed somewhere along the way, used to be borked :+1:

PR #2585, opened 2022-08-22
[LUA] Create job_utils/thief, migrate JA functionality, and add Larceny

Schaedelschaden:
**_I affirm:_**

- Create job_utils/thief.lua and populates necessary ability check and ability use functions.
- Moves functionality from all Thief globals/abilities files to globals/job_utils/thief.lua.
- Sets up Larceny to steal status effects, prioritizing SP Abilities used by enemies.
- Adjusts the Larceny ability to use the correct actions and messages when successful or failing to steal a buff.

## References

https://www.bg-wiki.com/ffxi/Larceny

2. Load the server.
3. Use the !changejob THF 99 command.
5. Use Larceny against a mob that uses SP Abilities and has other buffs to test stealing priority.

### DIFF
--- a/scripts/globals/job_utils/thief.lua
+++ b/scripts/globals/job_utils/thief.lua
@@ -27,7 +27,7 @@ local despoilDebuffs =
     xi.effect.SLOW
 }
 
-local stealableSPEffects = 
+local stealableSPEffects =
 {
     xi.effect.MIGHTY_STRIKES,   xi.effect.HUNDRED_FISTS, xi.effect.MANAFONT,     xi.effect.CHAINSPELL,
     xi.effect.PERFECT_DODGE,    xi.effect.INVINCIBLE,    xi.effect.BLOOD_WEAPON, xi.effect.SOUL_VOICE,
--- a/scripts/globals/job_utils/thief.lua
+++ b/scripts/globals/job_utils/thief.lua
@@ -35,43 +35,9 @@ local stealableSPEffects =
     xi.effect.ELEMENTAL_SFORZO
 }
 
--- Quadavs Group ID's that the THF AF1 item can be stolen from:
--- Bronze Quadav groupid = 7949
--- Garnet Quadav groupid = 7960
--- Silver Quadav groupid = 7977
--- Zircon Quadav groupid = 7985
-local validThfQuestMobs =
-{
-    17379367, 17379368, 17379459, 17379470, 17379477, 17379489, 17379493, 17379495, 17379501, 17379505, 17379509,
-    17379513, 17379517, 17379521, 17379525, 17379529, 17379533, 17379538, 17379543, 17379547, 17379552, 17379556,
-    17379560, 17379565, 17379569, 17379573, 17379577, 17379581, 17379585, 17379597, 17379363, 17379364, 17379462,
-    17379473, 17379480, 17379496, 17379498, 17379500, 17379504, 17379508, 17379512, 17379516, 17379520, 17379524,
-    17379528, 17379532, 17379536, 17379541, 17379546, 17379550, 17379555, 17379559, 17379563, 17379568, 17379572,
-    17379576, 17379580, 17379584, 17379588, 17379600, 17379361, 17379362, 17379460, 17379471, 17379478, 17379490,
-    17379491, 17379497, 17379502, 17379506, 17379510, 17379514, 17379518, 17379522, 17379526, 17379530, 17379534,
-    17379539, 17379544, 17379548, 17379553, 17379557, 17379561, 17379566, 17379570, 17379574, 17379578, 17379582,
-    17379586, 17379598, 17379365, 17379366, 17379461, 17379472, 17379479, 17379492, 17379494, 17379499, 17379503,
-    17379507, 17379511, 17379515, 17379519, 17379523, 17379527, 17379531, 17379535, 17379540, 17379545, 17379549,
-    17379554, 17379558, 17379562, 17379567, 17379571, 17379575, 17379579, 17379583, 17379587, 17379599
-}
-
 -----------------------------------
 -- Local Functions
 -----------------------------------
-
-local function checkThfAfQuest(player, target)
-    local targid = target:getID()
-
-    if player:getQuestStatus(xi.quest.log_id.WINDURST, xi.quest.id.windurst.THE_TENSHODO_SHOWDOWN) == QUEST_ACCEPTED then
-        for key, value in pairs(validThfQuestMobs) do
-            if value == targid then
-                return true
-            end
-        end
-
-        return false
-    end
-end
 
 local function processDebuff(player, target, ability, debuff)
     local power = 10
@@ -413,7 +379,7 @@ end
 
 xi.job_utils.thief.useSteal = function(player, target, ability, action)
     local thfLevel
-    local stolen = 0
+    local stolen = action:getParam(target:getID())
 
     if player:getMainJob() == xi.job.THF then
         thfLevel = player:getMainLvl()
@@ -425,14 +391,11 @@ xi.job_utils.thief.useSteal = function(player, target, ability, action)
 
     local stealChance = 50 + stealMod * 2 + thfLevel - target:getMainLvl()
 
-    stolen = target:getStealItem()
+    if stolen == 0 then
+        stolen = target:getStealItem()
+    end
 
     if target:isMob() and math.random(100) < stealChance and stolen ~= 0 then
-        -- This guarantees item if quest is flagged
-        if checkThfAfQuest(player, target) == true then
-            stolen = xi.items.BOWL_OF_QUADAV_STEW
-        end
-
         player:addItem(stolen)
         target:itemStolen()
         ability:setMsg(xi.msg.basic.STEAL_SUCCESS) -- Item stolen successfully
@@ -455,6 +418,7 @@ xi.job_utils.thief.useSteal = function(player, target, ability, action)
                 stolen = player:stealStatusEffect(target)
                 if stolen ~= 0 then
                     ability:setMsg(xi.msg.basic.STEAL_EFFECT)
+                    action:setAnimation(target:getID(), 181)
                 end
             -- else
             --     effect = target:dispelStatusEffect()
--- a/scripts/globals/job_utils/thief.lua
+++ b/scripts/globals/job_utils/thief.lua
@@ -114,7 +114,7 @@ end
 -- Ability Check Functions
 -----------------------------------
 xi.job_utils.thief.checkAccomplice = function(player, target, ability)
-    if (target == nil or target:getID() == player:getID() or not target:isPC()) then
+    if target == nil or target:getID() == player:getID() or not target:isPC() then
         return xi.msg.basic.CANNOT_ON_THAT_TARG, 0
     else
         return 0, 0
@@ -122,7 +122,7 @@ xi.job_utils.thief.checkAccomplice = function(player, target, ability)
 end
 
 xi.job_utils.thief.checkCollaborator = function(player, target, ability)
-    if (target == nil or target:getID() == player:getID() or not target:isPC()) then
+    if target == nil or target:getID() == player:getID() or not target:isPC() then
         return xi.msg.basic.CANNOT_ON_THAT_TARG, 0
     else
         return 0, 0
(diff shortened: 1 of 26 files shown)


scripts/globals/job_utils/thief.lua
```diff
@@ -0,0 +1,486 @@
...
+}
+
+-- Quadavs Group ID's that the THF AF1 item can be stolen from:
+-- Bronze Quadav groupid = 7949
+-- Garnet Quadav groupid = 7960
+-- Silver Quadav groupid = 7977
+-- Zircon Quadav groupid = 7985
+local validThfQuestMobs =
+{
+    17379367, 17379368, 17379459, 17379470, 17379477, 17379489, 17379493, 17379495, 17379501, 17379505, 17379509,
```
claywar: This is going to need some love, and a new method in general.  Wall of magic numbers is a show-stopper in my book, even if that's how we used to do things.

scripts/globals/job_utils/thief.lua
```diff
@@ -0,0 +1,486 @@
...
+xi.job_utils.thief.checkAccomplice = function(player, target, ability)
+    if (target == nil or target:getID() == player:getID() or not target:isPC()) then
+        return xi.msg.basic.CANNOT_ON_THAT_TARG, 0
+    else
+        return 0, 0
+    end
+end
+
+xi.job_utils.thief.checkCollaborator = function(player, target, ability)
+    if (target == nil or target:getID() == player:getID() or not target:isPC()) then
```
claywar: outer parens not necessary

scripts/globals/job_utils/thief.lua
```diff
@@ -0,0 +1,486 @@
...
+            hpSteal = mobHP
+        end
+
+        target:addHP(-hpSteal)
+        player:addHP(hpSteal)
+    end
+
+    local mugChance = 90 + thfLevel - target:getMainLvl()
+
+    if (target:isMob() and math.random(100) < mugChance and target:getMobMod(xi.mobMod.MUG_GIL) > 0) then
```
claywar: outer parens not necessary

scripts/globals/job_utils/thief.lua
```diff
@@ -0,0 +1,486 @@
...
+        local purse = target:getMobMod(xi.mobMod.MUG_GIL)
+        local fatpurse = target:getGil()
+        gil = fatpurse / (8 + math.random(0, 8))
+
+        if (gil == 0) then
```
claywar: outer parens not necessary

scripts/globals/job_utils/thief.lua
```diff
@@ -0,0 +1,486 @@
...
+            gil = fatpurse / 2
+        end
+
+        if (gil == 0) then
+            gil = fatpurse
+        end
+
+        if (gil > purse) then
```
claywar: outer parens not necessary

scripts/globals/job_utils/thief.lua
```diff
@@ -0,0 +1,486 @@
...
+            gil = purse
+        end
+
+        if (gil <= 0) then
```
claywar: outer parens not necessary

claywar (commented):
Biggest thing is that wall of magic numbers for mob IDs.  There has to be a better way.  Outside of that, minor formatting.

scripts/globals/job_utils/thief.lua
Schaedelschaden: All of these mobs have the Quadav Stew (item ID 4569) in their drop lists as a stealable item. It looks like the function checks if the mob is one of those particular ID's before it forces a guaranteed steal. My thought is if the `validThfQuestMobs` function checks the drop ID on the mob, and if it matches one of the four drop ID's for the valid mobs, it returns true.

claywar: Perhaps a binding that checks if a certain item is in the droplist for steal?  Seems a bit niche for that, but I think working backwards might be a bit cleaner. 

If I potentially have stew: (Stew!)

Schaedelschaden:
```
local validThfQuestDropPools =
{
    362, 936, 2254, 2903
}
```

and

```
local function checkThfAfQuest(player, target)
     if player:getQuestStatus(xi.quest.log_id.WINDURST, xi.quest.id.windurst.THE_TENSHODO_SHOWDOWN) == QUEST_ACCEPTED then
        for i = 1, #validThfQuestDropPools do
            if target:getObjType() == xi.objType.MOB and target:getDropID() == validThfQuestDropPools[i] then
                return true
            end
        end

        return false
    end
end
```

Adjusting the functions to the above reduces the number of ID's that have to be tracked. I found only the Beadeaux Bronze, Garnet, Silver, and Zirvon Quadav had those drop lists. I haven't been able to find any core functions that check a mob's drop list for what items are contained in the list.

claywar: Pulling the elder into this one @TeoTwawki .  Seeing as pools are internal, this seems much more reasonable, but would like his feedback.

Schaedelschaden: Oh, wait. `getStealItem` might work, hang on.

Schaedelschaden: `getStealItem` does work but the subject Quadav can also have beastcoins stolen from them. Using `getStealItem` randomly pulls from the mob's stealable item pool on their drop list. We might get the beastcoin returned by the function, or we might get the stew returned.

TeoTwawki: or, hear me out here, we make an onSteal for these the same way we have onCriticalHit and onMagicHit, and stick that in these mobs scripts.

there is actually a lot I'd like to do on steal attempt / on landing anyway

Schaedelschaden: I'm all ears and will happily give it a shot!

WinterSolstice8: check `uint16 OnMobWeaponSkillPrepare(CBattleEntity* PMob, CBattleEntity* PTarget)` in luautils.cpp for a decent template to add an "onSteal" event. It looks close enough to what you want.

Schaedelschaden:
- Removed the job_utils/thief.lua `validThfQuestDropPools` local variable
- Removed the job_utils/thief.lua `checkThfAfQuest` local function
- Created the luautils.cpp `onSteal` function
- Created the lua_action.cpp `getParam` function
- Added a charentity.cpp call to check the target's `entity.onSteal` function when Steal is used
- Returns 0 with no `entity.onSteal` function present. Can be used to force a specific, integer result (Item ID)
- `onSteal` results are injected into the action's param value
- Adjusted job/utils.lua `useSteal` function to pull the action's param value with `getParam`. If no previous value has been provided it will use the `getStealItem` function to determine a stealable item
- Set up Bronze, Garnet, Silver, and Zircon Quadav to force the Quadav Stew item to be stolen if The Tenshodo Showdown quest has been accepted and the stealing player does not already have the item

claywar:
I really like where you're going with this solution, and started thinking about ways we could make this more generic to potentially apply to other abilities as opposed to depending on specific cases in core.  Something like `onPlayerAbility(mob, player, abilityId)`
could be used to catch all abilities from a player on the mob's enmity list.  Would also most likely need other params (override vs fallthrough, etc)

Again, outside of scope for this PR, just dumping thoughts.

Schaedelschaden:
`onPlayerAbility(mob, player, ability, action)` might be the ideal setup for the function so that everything relating to the user, target, and ability can be adjusted before it hits. I'm not sure what other abilities might need these sorts of modifications specifically from the individual mobs but am definitely interested in some other cases to look at.

At the moment `onSteal` is only set up to provide an override for the item stolen. In thinking about it, some mobs might have specific statuses that can't be stolen and if the function is used to check for a status and returns the status ID the Steal will assume it's an item. Assuming it's only used for items right now then it should be fine, but depending on future demands on it then it would probably be better set up generically anyway.

TeoTwawki:
> In thinking about it, some mobs might have specific statuses that can't be stolen

I'd been thinking about flagging to handle that type of thing - currently we check if its dispellable, but their may be things that don't match up perfectly with dispellable in retail - I know there are in our core, since we have to work around that in some cases by manually setting an effect in a mob script to be dispellable or not when it doesn't match the effects standard behavior.

But the current effectflags enum is already to large in my opinion. If I could think up good naming conventions and a good logical division I'd probably try and break it into multiple fields. Maybe "steal flags" could be its own thing entirely tho for steal, aura steal, and larceny. 

Just thinking ~~out loud~~ _in text_ :)

TeoTwawki:
I was envisioning a per-mob way to make logic things for steal regardless of success or not, override optional, similar to spell scripting logic. We have scripts that change the spell being cast, and others that force a spell to become aoe or change its mp cost (king behemoths meteor was 1 mp last I looked)

What you've done here does what it needs to do and can be expanded on/updated as we need to keeping us from bloating the steal script or utils global. :+1: 

if I am understanding correctly if I wanted to do something but keep whatever already would have been stolen or not, I can make the return be the original getStealitem (example: make a mob complain about your sticky fingers in chat) so a lot of things are already possible future forward before we need them (maybe there will be mobs that "proc" via steal - who knows).

If in the future this can be expanded into a generic for all JA as clay suggests that would be wonderful. Or even if it winds up having to be 2 functions. As for the status support, script assuming item - aura steal only kicks in if item steal fails, so if it falls through back to the main steal scripting logic, I think we're covered for now till we have a reason to do modified status logic.

TeoTwawki (dismissed):
obligatory "lgtm"

scripts/zones/Beadeaux/mobs/Bronze_Quadav.lua
```diff
@@ -0,0 +1,28 @@
...
+--  Quest THE_TENSHODO_SHOWDOWN active
+-----------------------------------
+local ID = require("scripts/zones/Beadeaux/IDs")
+require('scripts/globals/items')
+require("scripts/globals/mobs")
+-----------------------------------
+local entity = {}
+
+entity.onSteal = function(player, target, ability, action)
+    if player:getQuestStatus(xi.quest.log_id.WINDURST, xi.quest.id.windurst.THE_TENSHODO_SHOWDOWN) == QUEST_ACCEPTED and
```
claywar: Multiline conditional formatting should follow style guide

scripts/zones/Beadeaux/mobs/Garnet_Quadav.lua
```diff
@@ -0,0 +1,28 @@
...
+--  Quest THE_TENSHODO_SHOWDOWN active
+-----------------------------------
+local ID = require("scripts/zones/Beadeaux/IDs")
+require('scripts/globals/items')
+require("scripts/globals/mobs")
+-----------------------------------
+local entity = {}
+
+entity.onSteal = function(player, target, ability, action)
+    if player:getQuestStatus(xi.quest.log_id.WINDURST, xi.quest.id.windurst.THE_TENSHODO_SHOWDOWN) == QUEST_ACCEPTED and
```
claywar: Multiline conditional formatting should follow style guide

scripts/zones/Beadeaux/mobs/Silver_Quadav.lua
```diff
@@ -0,0 +1,28 @@
...
+--  Quest THE_TENSHODO_SHOWDOWN active
+-----------------------------------
+local ID = require("scripts/zones/Beadeaux/IDs")
+require('scripts/globals/items')
+require("scripts/globals/mobs")
+-----------------------------------
+local entity = {}
+
+entity.onSteal = function(player, target, ability, action)
+    if player:getQuestStatus(xi.quest.log_id.WINDURST, xi.quest.id.windurst.THE_TENSHODO_SHOWDOWN) == QUEST_ACCEPTED and
```
claywar: Multiline conditional formatting should follow style guide

scripts/zones/Beadeaux/mobs/Zircon_Quadav.lua
```diff
@@ -1,13 +1,25 @@
...
+--  Quest THE_TENSHODO_SHOWDOWN active
 -----------------------------------
 local ID = require("scripts/zones/Beadeaux/IDs")
+require('scripts/globals/items')
 require("scripts/globals/mobs")
 -----------------------------------
 local entity = {}
 
+entity.onSteal = function(player, target, ability, action)
+    if player:getQuestStatus(xi.quest.log_id.WINDURST, xi.quest.id.windurst.THE_TENSHODO_SHOWDOWN) == QUEST_ACCEPTED and
```
claywar: Multiline conditional formatting should follow style guide

claywar (commented):
Just some additional styling changes needed from what I've seen so far

WinterSolstice8 (approved):

Schaedelschaden:
Do I need to make any further adjustments for this PR?

Issue #1250, opened 2022-02-05
Unable to access Satiator island in Promy-Dem


hooksta4:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: BASE

Promy Dem has two 3rd level islands. One of which has NM Satiator spawn. After countless attempts, can never get the MR portal to send to that island. 
```

MEMORY_STREAMS =
        {
            [11]        = { 157, -4,  -82,  161, 4,  -77, {46}},    -- floor 1 return
            [21]        = {-383, -4,   -2, -278, 4,    2, {41}},    -- floor 2 return
            [31]        = {-160, -4,  437, -157, 4,  441, {30}},    -- floor 3 return
            [32]        = {  -2, -4, -322,    2, 4, -317, {30}},    -- floor 3 return
            [41]        = { 357, -4,  237,  361, 4,  242, {34}},    -- floor 4 return
            [16851277]  = { 117, -4, -283,  122, 4, -277, {30}},    -- floor 1 MR1
            [16851281]  = { -83, -4,  -83,  -77, 4,  -76, {34, 35}}, -- floor 2 MR1
            [16851282]  = { -82, -4,   76,  -77, 4,   80, {34, 35}}, -- floor 2 MR2
            [16851283]  = {-282, -4, -202, -277, 4, -196, {34, 35}}, -- floor 2 MR3
            [16851284]  = {-361, -4,   36, -356, 4,   42, {34, 35}}, -- floor 2 MR4
            [16851278]  = {  37, -4, -203,   43, 4, -198, {32}},    -- floor 3 MR1
            [16851279]  = {-122, -4, -242, -116, 4, -237, {32}},    -- floor 3 MR2
            [16851280]  = {-122, -4, -402, -116, 4, -396, {32}},    -- floor 3 MR3
            [16851285]  = {-322, -4,  156, -316, 4,  162, {32}},    -- floor 3 MR4
            [16851286]  = { -42, -4,  317,  -37, 4,  322, {32}},    -- floor 3 MR5
[...]

PR #2589, opened 2022-08-23
Move spawn locations for Promy 3rd floor NM's #1250 (alternate location commented out)

hooksta4:
There are two spawn locations on floor 3 in Promy-Dem.

**_I affirm:_**

Adds second spawn location for NM Satiator in Promy Dem. Currently, the elevators are wonky at best and will only take to one island, which coincidently isnt the island the NM is on. Solves issue #1250 

Progress through Dem by defeating Memory Receptacles. Also I spawn/despawned a bunch of times.


sql/mob_spawn_points.sql
```diff
@@ -3760,7 +3760,7 @@ INSERT INTO `mob_spawn_points` VALUES (16851248,'Gorger','Gorger',20,15.474,-0.5
 
 -- TODO: Implement Apex Mobs
 
-INSERT INTO `mob_spawn_points` VALUES (16851267,'Satiator','Satiator',21,-205.880,-0.500,268.850,80);
+INSERT INTO `mob_spawn_points` VALUES (16851267,'Satiator','Satiator',21,-44.3170,0,-242.987,80);
```
TeoTwawki, line 3763: Didn't really matter which one was the "default". on despawn as long as the function `UpdateNMSpawnPoint()` is called it'll select a new one from the other table.

You will need to edit [the mobs script](https://github.com/LandSandBoat/server/blob/base/scripts/zones/Promyvion-Dem/mobs/Satiator.lua) or your extra row in NM Spawn Points will do nothing, as I see the script currently does not call the function. If you want the original spawn at server start to pick a random place, you'll have to do that in the zone script ~~using `setSpawn()` to set its "home" then `setPos()` to move it there (otherwise it will walk back to the 1st position)~~ _scratch that, call  `UpdateNMSpawnPoint()`on init, that's before the 1st spawn_

TeoTwawki (changes requested):

sql/mob_spawn_points.sql
hooksta4: So the bigger issue is the MR portals will NOT send players to other island (where the NM is located at). By changing it to the later, the NM spawns on the island where players are transported. Perhaps comment out the other one until the portals are fixed?

TeoTwawki: yes, commented out is fine, but honestly, I would have focused on why the portal isn't random-ing before touching the NM. The problem isn't the NMs problem, and this is working around an issue without really fixing it. So if #1250 is being closed by this I expect a new issue for the teleport problem. A whole island is inaccessible atm and we're kind of removing the incentive to fix it now. I say kind of because I dunno anyone that seriously camps this thing >.>;

hooksta4: I understand. This is a temporary fix for a larger problem as stated. Either way, the NM does have 2 spawn points and until the portals are fixed, this would at least make the NM available. I suppose the NM is trivial at post 75, but its pretty useful nonetheless.

TeoTwawki: which problem is larger is debatable and honestly not important right now as it's already agreed that it's fine to comment it out - just make a new tracker issue for the portal please. Otherwise it'll be forgotten.

hooksta4: Sounds good.

hooksta4: Well hold on, out of curiosity, I went and looked at the other 2 3rd floor NM's. Appears they too, spawn on the inaccessible island. let me update the other two to put on the accessible island and then can create a more specific issue for teleporters.

sql/nm_spawn_points.sql
```diff
@@ -252,6 +252,8 @@ INSERT INTO `nm_spawn_points` VALUES (16806227,49,-549.802,-8.944,-24.848);
 INSERT INTO `nm_spawn_points` VALUES (16806249,0,-490.195,-4.503,145.451); -- Ambusher Antlion
 INSERT INTO `nm_spawn_points` VALUES (16822423,0,4.609,11.472,-101.235); -- Bugbear Strongman (west)
 INSERT INTO `nm_spawn_points` VALUES (16822427,0,60.728,15.487,-118.886); -- Bugbear Strongman (east)
+INSERT INTO `nm_spawn_points` VALUES (16851267,0,-44.3170,0,-242.987); -- Satiator
+-- INSERT INTO `nm_spawn_points` VALUES (16851267,1,-205.880,-0.500,268.850); -- Satiator on inaccessable island.
```
TeoTwawki: This wasn't the part that needs to be put in commented out. As it stands now, neither sql row in this table is at all being used at all. Add the function calls, and leave them in commented out, then this PR is ready to pass.

TeoTwawki (approved):
good enough

EpicTaru (approved):

EpicTaru:
https://github.com/LandSandBoat/server/issues/2591 since I didn't see it referenced yet in this PR.

PR #2590, opened 2022-08-24
Fix halving of Phantom Roll recast when rolling 11

CatsEyeXI:
**_I affirm:_**

image
Fixes what I believe is a typo, preventing the normal halving of the phantom roll cooldown when a corsair rolls an 11.

roll an 11, first recast should be instant, subsequent recasts should be 30 seconds (without merits) as long as an XI roll is up.


WinterSolstice8 (approved):
LGTM

zach2good (approved):

Issue #2564, opened 2022-08-20
🐛 BRD March effect is not being properly calculated by spells and attacks for cd/delay reduction


SirGouki:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Branch affected by issue 

`base`

## Steps to reproduce

I used 75 BRD/37 NIN to test spells (Utsusemi Ichi) and 75 WAR/37 BRD with all skills capped to trigger.
Equipment for Spells:  Gjallarhorn (no other bonuses)

Casted Utsusemi Ichi with no haste effect, got 30 seconds CD each time (3 times)
Repeated the test with marches:
Utsusemi: Ichi - BRD75/NIN37 - Advancing March - 29 seconds
Utsusemi: Ichi - BRD75/NIN37 - Victory March - 29 seconds
Utsusemi: Ichi - BRD75/NIN37 - Both - 29 seconds

For comparison, I also tested with 75 RDM / 37 NIN:
Ustsuemi: Ichi - RDM75/NIN37- Haste - 23 seconds
Utsusemi: Ichi - RDM75/NIN37- no haste - 27 seconds 

For melee I used the Chaosbringer (delay:666, approx 11 seconds per swing.  Note: Chat window timestamps were used to measure delay)
75 WAR / 37 BRD - no haste effects - ~11 seconds per swing
75 WAR / 37 BRD - Advancing March - ~11 seconds per swing
99 WAR / 49 RDM (level diff for haste) - Haste - ~8 seconds

## Expected behavior

Marches should reduce the delay between melee attacks.  Per BGwiki (https://www.bg-wiki.com/ffxi/Category:March)
Advancing March should be at 10.55% potency, and Victory March should be at 15.92% potency when skill capped (though the skill caps for these 2 need more info/verification).  This means around a 3-4.5 (I don't know if both marches = 26.47% haste or something else) second reduction in cd for Utsusemi Ichi, depending on the march effect(s), and around 1-1.5 second reduction between swings for Chaosbringer, depending on the march effect(s), as opposed to what the testing I did returned.

kaincenteno:
[...]

PR #2601, opened 2022-08-25
[lua] Adjust Skill "if" check in enhancing_song and change division/multiplication order for March power

WinterSolstice8:
**_I affirm:_**

Partially helps #2564

Use Marches, !getmod haste_magic and see it not be a super low number like 4. Closer to 10% per march: 1,000


TeoTwawki (approved):
lgtm

zach2good (approved):

Issue #2597, opened 2022-08-24
🔨 More control over additional effect proc chance


CriticalXI:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Describe the feature

I was hoping to see if we could have the additional effect system adjusted to allow for more control over how a weapon's additional effect procs. I could be missing or wrong and there is already a way to do this, but as of right now I haven't been able to find a way. As of right now it seems that the system is built out that either it procs or it doesn't purely based just off of the ITEM_ADDEFFECT_CHANCE mod.

This causes issues for a couple items that I've ran into so far trying to get additional effects functioning for some items.

https://ffxiclopedia.fandom.com/wiki/Chicken_Knife
This is one example, the chance to proc should vary depending on the level difference between you and the mob.

https://ffxiclopedia.fandom.com/wiki/Treat_Staff
Another example that the proc chance should be blocked depending on time and moon, or an event is active.

As far as I understand it currently, these are impossible to get working properly in the current system.

TeoTwawki:
I guess you missed my response in discord, these would happen via item_latents

CriticalXI:
I still haven't been able to find a way to make Chicken Knife even looking into the item_latents as there isn't any that proc depending on a check on the mob. Do you have an example of one doing this? I was also curious how to make the visible proc animation happen with a latent too.

TeoTwawki:
I've been held up while diagnosing some issues with the latent system, but I have several that will be there soon, chicken knife included. Additionally the current handling of ranged additional effects is going to soon change from just checking if the item is ranged to looking for a modifier, so that dStat and level checks can optionally be applied to any weapon. Since Add effect Flee is the only one to allegedly use the `/check` value and the latent system is proving untrustworthy, I am doing the chicken knife (and also Shijo) myself in an upcoming pr with some script logic to optionally accommodate it.

TeoTwawki:
I've assigned myself this, and will get back to you soon

TeoTwawki:
> I've assigned myself this, and will get back to you soon

And Maury meme determined that was a lie (unintentional)

😢 

[...]

PR #2602, opened 2022-08-25
Latent effects: Set/check all the modifiers related to additional effect modifier application, not just the addType.

TeoTwawki:
**_I affirm:_**

So I have been trying to add some additional effects that need to be latent triggered, and was scratching my head at how not all the sql rows seemed to be working. Changing equipment I started getting unexpected values for proc chance and such.

Turned out only the addEffect Type modifier was being correctly set the items - they go on the item, not the player. 

This is important because we do not want the proc rates and such to stack up when multiple items get equipped.

~~Additionally attempt to cleanup old legacy code from back when addEffects were hard coded in battleutils many years ago.~~

~~Some weird loops and such were used back then and oddities like "Adding" a negative mod or zero'ing it out rather than just delMod to reverse it.~~

Lessons Learned: that loop and setting 0 as well as "adding" negative mod amounts are all because normal functionality for mod handling on item objects is largely crap where it isn't outright nonexistent. I can't fully fix this without rewriting parts of item_equipment.cpp either.

1. rebuild
2. make latent add effects in the latents table
3. gm yourself the item and go pick on some mobs
4. do the same with some existing latent items unrelated to additional effect, just to be sure I didn't break them!

### DIFF
--- a/src/map/latent_effect.cpp
+++ b/src/map/latent_effect.cpp
@@ -104,25 +104,47 @@ void CLatentEffect::SetModPower(int16 power)
     m_ModPower = power;
 }
 
+bool CLatentEffect::ModOnItemOnly(Mod modID)
+{
+    if (modID == Mod::DMG ||
+        modID == Mod::ITEM_ADDEFFECT_TYPE ||
+        modID == Mod::ITEM_SUBEFFECT ||
+        modID == Mod::ITEM_ADDEFFECT_DMG ||
+        modID == Mod::ITEM_ADDEFFECT_CHANCE ||
+        modID == Mod::ITEM_ADDEFFECT_ELEMENT ||
+        modID == Mod::ITEM_ADDEFFECT_STATUS ||
+        modID == Mod::ITEM_ADDEFFECT_POWER ||
+        modID == Mod::ITEM_ADDEFFECT_DURATION)
+    {
+        return true;
+    }
+    return false;
+}
+
 bool CLatentEffect::Activate()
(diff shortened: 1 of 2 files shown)


TeoTwawki:
~~on hold pending some extended testing, leaving in draft status till then~~

EpicTaru:
Possibly related to https://github.com/LandSandBoat/server/issues/2597.  Please let me know if not and I can delete this comment.

TeoTwawki:
ok, the bug I was experiencing is not related to the changes here in this pr and has in fact always been a problem - I will dive into this separately. it has to do with us checking condition changes on the player (for example you delevel and are now below the lv the latent should work at)

TeoTwawki:
open to naming suggestions for `IsAddEffect()` for expanding it into something for "these mods are on the item not the player" because thats the main difference there. Will look AGAIN at simplifying this code after we work the bugs out of latent effect conditions.

EpicTaru:
I'll cherry pick this again tomorrow and see what happens

TeoTwawki:
> I'll cherry pick this again tomorrow and see what happens

don't bother, I already went and checked out what I needed to.

which resulted in my rolling back everything except for the breaking out some additional mod checks.

TeoTwawki:
renamed the function and re-introduced a small part of my cleanup (death to c-style cast)

claywar (approved):

WinterSolstice8 (approved):

PR #2605, opened 2022-08-25
Implement Ul'xzomit mother and baby distinction

### LINKED ISSUE #2573: 🐛 [cpp] [lua] Entity may occasionally face a random angle after `pathTo()`

## Branch affected by issue 

`base`

## Steps to reproduce

This is mostly caused by a mixture of floating point precision and [lua] `lookAt()` or [cpp] `LookAt()` not checking for invalid inputs before calling/calculating `worldAngle()`. If the `x` and `z` coordinates of the two points are too close, it can result in an unexpected angle. Worse, if the `x` and `z` coordinates are identical, this results in a `NaN` which gets turned into a `0` when converted to `uint8`.

This can result in entities pathing using `pathTo()` (doesn't happen with `pathAround()`) occasionally looking in seemingly random directions.

### Test 1. `pathTo()`
1. Have an entity intermittently path using `pathTo()`.
2. Watch it carefully.
3. When the entity reaches its destination, it will occasionally look in a random direction.

### Test 2. [cpp] `LookAt()` or [lua] `lookAt()`

1. Have an entity look at its own position but with a different `y` coordinate.
2. The entity will face `0`.

### Test 1.

3. When the entity reaches its destination, it will keep looking in the direction it was moving.

### Test 2.

2. The entity won't change its facing (because the target was too close).

### LINKED ISSUE #2604: 🔨 Make baby Ul'xzomit follow mother Ulxzomit

## Describe the feature

In Al'Taieu, each Ul'xzomit is supposed to be followed by two smaller Ul'xzomit. The mother and baby Ul'xzomit also have different drop rates.

See:
 * https://www.youtube.com/watch?v=QVLQUynPjvs
 * https://ffxiclopedia.fandom.com/wiki/Ul%27xzomit

The babies are supposed to spawn after the mother, and they will link with the mother but not vice versa (nor with eachother). The wiki page suggests there are 45 mothers and 90 babies, but the SQL tables have only 45 Ul'xzomit entries, so it's not clear what the mob ids of the babies would be. Considering the Ul'xzomit ids always appear in groups of three, I suggest having 15 mother Ul'xzomit and making the two subsequent Ul'xzomit in the table be babies.

DiscipleOfEris:
**_I affirm:_**

 * Two thirds of the current Ul'xzomit mobs in Al'Taieu are now babies following the remaining one third Ul'xzomits.
 * The baby Ul'xzomit are smaller and the mother Ul'xzomit are larger.
 * Increase roam distance of mother Ul'xzomit.
 * Mother and baby Ul'xzomit have appropriate (different) drop tables, as in: https://ffxiclopedia.fandom.com/wiki/Ul%27xzomit
    * Make elemental clusters slotted drops (10% for mothers, 1% for babies, with equal chance for each cluster).

### Under the hood

 * Create a new lua function in `scripts/globals/follow.lua` that allows you to call `xi.follow.follow(follower, target[, options])` to have the follower mob persistently follow the target entity.
 * Rename `ROAMFLAG_EVENT` to `ROAMFLAG_SCRIPTED` for a more intuitive name.
 * Add a new `MOBMOD_ROAM_RESET_FACING`. This implemented previous behavior that was hard-coded into `ROAMFLAG_EVENT`. Specifically, it makes mobs reface their original spawn facing when they roamed back home (like after the players died).
 * Add a new `MOBMOD_DONT_ROAM_HOME`. If used, it disables the leash that causes mobs to eventually roam back home.
 * Fix some bugs involving `LookAt()` that cropped up while working on this.
 * Add several utility functions in lua and cpp, and modify the `distance` functions to add an optional `ignoreVertical` flag.
 * Rework the lua `setMobFlags()` function to work from a mob as well, like `mob:setMobFlags(flags)`.
 * Adds a few more `triggerListener()` calls for pathing and roaming.

### Not implemented/Need more research

 * Ul'xzomit roam patterns. At the moment they roam normally but with a larger range. Need more retail research.
 * Babies respawning. At the moment they only spawn when the mother spawns. Need more retail research.
 * Baby roam behavior if mother is dead. At the moment they don't roam at all. Need more retail research.
 * Difference in HP/stats of mother vs. baby. Need more retail research.

Fixes #2604 
Fixes #2573 

~~We might consider making the elemental clusters be slotted drops. At the moment they're implemented as 10% and 1% each for the Ul'xzomit mothers and babies, but I think it's supposed to be 10% and 1% that a cluster drops at all. As is, the mothers have a 57.0% chance for at least one cluster dropping and the babies have a 7.7% chance for at least one cluster.~~

Also, is there interest in a gm command for making a mob follow an entity? I'm not sure if that belongs in LSB core.

1. Go to Al'Taieu.
2. Watch the Ul'xzomit roam about. The babies follow the mothers.
3. Attack a baby Ul'xzomit. Nothing should link.
4. Attack a mother Ul'xzomit. The two babies should link, but no others.
5. Kill a few mother and baby Ul'xzomit. Notice the different drop rates.


scripts/globals/status.lua
```diff
@@ -2468,6 +2468,7 @@ xi.mobMod =
     ALLI_HATE           = 68, -- Range around target to add alliance member to enmity list.
     NO_LINK             = 69, -- If set, mob cannot link until unset.
     NO_REST             = 70, -- Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
+    LEADER              = 71, -- Indicate this mob is a leader. Used for Ul'xzomit mobs.
```
TeoTwawki: `Used for mobs that follow a defined "leader"` would be preferred description.

src/map/mob_modifier.h
```diff
@@ -99,6 +99,7 @@ enum MOBMODIFIER : int
     MOBMOD_ALLI_HATE          = 68, // Range around target to add alliance member to enmity list.
     MOBMOD_NO_LINK            = 69, // If set, mob cannot link until unset.
     MOBMOD_NO_REST            = 70, // Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
+    MOBMOD_LEADER             = 71, // Indicate this mob is a leader. Used for Ul'xzomit mobs.
```
TeoTwawki: `Used for mobs that follow a defined "leader"` would be preferred description.

src/map/lua/lua_baseentity.cpp
```diff
@@ -12884,7 +12888,7 @@ uint8 CLuaBaseEntity::getModelSize()
 
 void CLuaBaseEntity::setMobFlags(uint32 flags, uint32 mobid)
 {
-    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_MOB);
```
TeoTwawki: m_PBaseEntity here is the player, we get their cursor target in this binding and the mob is passed in from that or an optional ID input.

src/map/entities/mobentity.h
```diff
@@ -57,7 +57,7 @@ enum ROAMFLAG : uint16
     ROAMFLAG_NONE2   = 0x04,  //
     ROAMFLAG_NONE3   = 0x08,  //
     ROAMFLAG_NONE4   = 0x10,  //
-    ROAMFLAG_NONE5   = 0x20,  //
+    ROAMFLAG_SCRIPT  = 0x20,  // Will be implemented in a lua script (e.g. entity.onMobRoamAction())
```
TeoTwawki: > ROAMFLAG_EVENT   = 0x100, // calls lua method for roaming logic

:thinking: whatever "EVENT" was intended to do its poorly named and described and sounds suspiciously like it was meant to be what you are adding.

Also antlion ambush roamflag value and worm values are suspiciously present in another flag field in retail captures.

Frankly our entire enum is suspect, and I am against adding to it just to do things in script.

----

How pathing normally works is:

Do things in a script say to do things? then do them, else fallback to standard mob behavior. Turning off the script side is itself a scriptable action as well.  If the purpose was to make it toggle-able there shouldn't be a need for more core flags to do that - we have scripts that already do it.

```lua

entity.onPath = function(mob)
    --[[
        path logic here i.e. tell it where it needs to go
        as long as the mob is pathing this will continue to exec each "step" along its path
        example:
            xi.path.patrol(mob, pathNodes, xi.path.flag.RUN)
        tells the pathing logic to grab its points from a table named pathNodes and have "mob" do this 
        with the running flag to "hustle". 

        Hypothetical: Where was momma last tick? add that to our table for next tick,
        and remove where we were last tick.
    ]]
end

entity.onMobRoam = function(mob)
    -- move to start position if not moving
    if not mob:isFollowingPath() then
        --[[
        kick off pathing logic here, in the 1st tick of its roaming.
        once pathing begins standard roam logic shuts off until pathing id manually ended
        or the mobs battle AI begins. Example:
            mob:pathThrough(table input of places to go, any flags needed)
        ]]
    end
end
```

src/map/lua/lua_baseentity.cpp
claywar: Just following up on this as well, we really should move away from these breaks in Lua Land and move towards logging output in debug.

src/map/entities/mobentity.h
DiscipleOfEris: > ```lua
> entity.onMobRoam = function(mob)
> ```

I didn't use this one because it's hard coded to run once every three seconds. The `onPath()` one can never avoid the code that forces a mob to roam home, which would make `isFollowingPath()` true while roaming home.

src/map/lua/lua_baseentity.cpp
DiscipleOfEris: Ah, that got left in there when I was trying to figure out why it wasn't working the way I was expected at the time. I should have removed the break line entirely, since I need this function for the mob to change its own flag on spawn. Instead, it should just be a check in the function rather than a break.

src/map/entities/mobentity.h
TeoTwawki: > I didn't use this one because it's hard coded to run once every three seconds. The `onPath()` one can never avoid the code that forces a mob to roam home, which would make `isFollowingPath()` true while roaming home.

If thats all then we can use a different condition to not re-kick off that function till after it is within a range from home again. For example, a `mob:getLocalVar()` can be read and only kick off pathing if zero, and it within 5 yalms of its homepoint, set it back to zero again. in onPath could set it to one because one in there its pathing.

We could also store the last position it had before it was in battle mode, and path functions do include having it go back to its last node - we just make that pos its last node. Those I do not know how well that works - we haven't been using it except in big tabled NMs.

----

edit:

per discord discussion, a new mobmod for `DONT_ROAM_HOME` is acceptable.

src/map/lua/lua_baseentity.cpp
TeoTwawki: lil trivia:

I had actually been intending for setMobFlags and player nameflag functions and npc's entity flags related functions to all be combined (its the same stuff on diff entities, so its a lot of duplication!) but each entity type is slightly off from the others in where the bits land. so they either need bit shifting or more preferable, rework the masks. There are more bits that belong in those masks that we aren't using - far to much for one single mask field (retail is using EIGHT mask fields for all this crap!!) so its on the back burner till thats done.

src/map/entities/mobentity.h
DiscipleOfEris: As we discussed in DMs, I'll make a MobMod that disables the behavior that forces a mob to roam home when it gets a set distance from its spawn location. Then I could refactor the mobs currently using ROAMFLAG_EVENT (dynamis and battlefield mobs) so they use lua to face their spawn location after roaming home (e.g. `onMobRoam() to check distance from spawn and reface its spawn facing when it's 0). Then claim ROAMFLAG_EVENT for my purposes.

I added the `triggerListener()` stuff because I wanted `xi.follow.follow()` to be a fire and forget function, so it could be used in, say, a GM command for example.

DiscipleOfEris: In addition to `DONT_ROAM_HOME` I created one more MobMod: `ROAM_RESET_FACING` which automatically gets set for all the dynamis and battlefield mobs that were relying on `ROAMFLAG_EVENT` to do it before.

DiscipleOfEris:
Damn you, Clang. I'll get you one day.

DiscipleOfEris:
I've updated the text of the PR to have a more up-to-date and thorough summary of the changes.

EDIT: I would like feedback on whether we believe the elemental clusters should be slotted drops. The wiki says 10% for clusters. My guess is that means 10% for a kill to result in a cluster (which could be any type), not 10% for each cluster, which means the cluster should be a slotted drop. At the moment, it's 10% per cluster, which means about a 57% chance for at least one cluster to drop.

TeoTwawki:
> I've updated the text of the PR to have a more up-to-date and thorough summary of the changes.
> 
> EDIT: I would like feedback on whether we believe the elemental clusters should be slotted drops. The wiki says 10% for clusters. My guess is that means 10% for a kill to result in a cluster (which could be any type), not 10% for each cluster, which means the cluster should be a slotted drop. At the moment, it's 10% per cluster, which means about a 57% chance for at least one cluster to drop.

correct, random clusters sharing a slot on retail. do not trust wiki drop rates - they combine things that shouldn't be and it's all user reported BS. FFXIDB isn't perfect but it's at least a little better.

sql/mob_spawn_mods.sql
```diff
@@ -609,6 +609,40 @@ INSERT INTO `mob_spawn_mods` VALUES (17588703,55,180,1);
...
+INSERT INTO `mob_spawn_mods` VALUES (16912452,71,2,1);
+INSERT INTO `mob_spawn_mods` VALUES (16912455,71,2,1);
+INSERT INTO `mob_spawn_mods` VALUES (16912468,71,2,1);
+INSERT INTO `mob_spawn_mods` VALUES (16912545,71,2,1);
+INSERT INTO `mob_spawn_mods` VALUES (16912548,71,2,1);
+
+-- Ul'xzomit mother roam mods
+INSERT INTO `mob_spawn_mods` VALUES (16912394,31,25,1);
+INSERT INTO `mob_spawn_mods` VALUES (16912409,31,25,1);
+INSERT INTO `mob_spawn_mods` VALUES (16912422,31,25,1);
```
TeoTwawki, line 632: ~~could you add `-- mobmod: modnamehere` to at least the 1st row on these 2 groups of mods, so future readers don't have to jump over to the header file?~~ nvm, automation to the rescue

scripts/zones/AlTaieu/mobs/Ulxzomit.lua
```diff
@@ -0,0 +1,71 @@
...
+    end
+end
+
+entity.onMobRoam = function(mob, player)
+    -- TODO: Respawn followers?
+    -- They might all stay dead until the entire group is killed, then respawn as a group.
+end
+
+-- Do the following and linking for the babies.
+entity.onMobRoamAction = function(mob)
```
TeoTwawki, line 54: see #2629 and comment there, because any change done for that issue will affect the future usage of roaming scripts and I don't want to break your work.

sql/mob_spawn_mods.sql
TeoTwawki: other than that, I think you are set here if nobody spots something I missed

claywar (approved):

sql/mob_spawn_mods.sql
claywar: Apologies, but going to pretty much guarantee a merge conflict here after adding standardized comments.

TeoTwawki (approved):

PR #2607, opened 2022-08-25
[sql] Drop Group 834 Audit

EpicTaru:
**_I affirm:_**

Audits Drop Group 834 (fishtraps fished up)
* Audit entails analyzing the drops listed for this group in mob_droplist.sql and verifying the drop rates of all drops listed for this group.
  * Drops were checked primarily against the TH0 column for an item on FFXIDB.  If an item wasn't found for a particular mob in a particular zone on FFXIDB then FFXIclopedia was checked (which was a drop rate of 0.00% in all instances). Sample sizes/confidence ratings were taken into account for both sites.
    * Drop rates for both sites were then converted to usage of the ``@variables``.
    * Drop rates of all mobs listed included in this Drop Group were then averaged (2 mobs total).  This average value was then converted to usage of the ``@variables``.

Due to this audit, overall drop rates did change.  This will be moreso noticed if there isn't a character applying Treasure Hunter to modify the drop chances.

The 1% rate for item Flytrap Leaf was used due to a count of 29 kills done (across both sites) and only the TH3+ rate was the one that recorded a 20% rate for 5 kills and a confidence rating of around 40%.  Because of this and the unknown amount of TH actually used for this, the very rare drop rate (1%) was used.  Please let me know if this drop rate needs to be adjusted.

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -7007,8 +7007,8 @@ INSERT INTO `mob_droplist` VALUES (833,0,0,1000,2375,@UNCOMMON); -- Zhayolm Card
 
 -- ZoneID:   1 - Fishtrap
 -- ZoneID:   2 - Fishtrap
-INSERT INTO `mob_droplist` VALUES (834,0,0,1000,1727,80); -- Piece Of Garhada Teak Lumber (8.0%)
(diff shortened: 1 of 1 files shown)


EpicTaru:
~~No idea if there is any way for me to resolve the thing the 1 failing check is reporting, please let me know if it is something I can resolve.~~

Disregard, failing check seems to have unborked itself.

claywar (approved):

PR #2351, opened 2022-07-19
[core] Add routine for dynamic mob/npc ID lookups on startup

zach2good:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

Behold! Black magic!

This is a step towards not having to keep mob and npc IDs in the Lua scripts, where they rot and get left behind after version updates - breaking things.

This will take any simple entry in ID.mob or ID.npc that has it's value set to DYNAMIC_LOOKUP (-1) and go through the mobs/npcs for that zone and match on the first entry with the same name (forced to upper case).

## Steps to test these changes

Go to a zone where I've enabled the dynamic lookup, and make sure the mobs still work.

You could also print out the ID values and make sure they're not -1.

Issue #2429, opened 2022-07-30
🐛 npcUtil.popFromQM doesn't work with DYNAMIC_LOOKUP definition.


Xaver-DaRed:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Branch affected by issue 

`base`

## Steps to reproduce

As an example

!pos 488 -1 166 54 (Arrapago reef qm1 pos)
!additem 2601 (pop item for Lil'Apkallu)

Trade the item we just added to the ??? we just warped to.
The NM won't pop. Map log will return the following:

![imagen](https://user-images.githubusercontent.com/60053999/181936219-1579cc28-3967-4d9b-95b5-e6f2c7b6eb08.png)

## Expected behavior

I have no expectations, as in, I dont know if we should adjust the function or use a different (existing) method completely and deprecate the existing popFromQM.

claywar:
Betting money you're setting your ID local to the scripts/zone/IDs.lua

That doesn't pull the cached value, so need to require zones[<zone_ID>]

Xaver-DaRed:
Well, not like "I" am. This is on a fresh untouched build.

claywar:
This is more of a shift that needs to be made with dynamic lookups.  We can't look at the file, because the values are changed at runtime.  Instead, have to look at the resulting cached value.  Songbirds in a Snowstorm was the first time I ran into this.  Resolved by:

`local pastBeaucedineID = zones[xi.zone.BEAUCEDINE_GLACIER_S]`

as opposed to requiring the IDs.lua

This doesn't look like a problem with the function, but rather what is being passed into the function.

N3ckB3ard:
@claywar I have tested your fix related to this issue and it does correct this problem but creates another.

If you are to restart the server after making direct LUA change, this will again repeat the same error if the local is outside of the function for spawning a ZNM (Referencing a QM NPC)

If you place the local within the entity.onTrade = function this will correct the issue and return the proper Zone ID

zach2good:
DYNAMIC_LOOKUP no longer exists, closing

PR #2614, opened 2022-08-29
[LUA] Update ID Shift to DYNAMIC_LOOKUP in Bhaflau Thickets

HaxPrime:
**_I affirm:_**

A recent retail update in Bhaflau Thickets shifted the ID's of two ZNM ID's. Setting these ID's to DYNAMIC_LOOKUP will help correct this shift. I used #2351 as a reference for my update.

Using `!spawnmob 16990473` to spawn Lividroot Amooshah returns an error for invalid mobid. Updating to dynamic lookup to bypass invalid mobid.

Using `!spawnmob 16990474` to spawn Dea returns an error for invalid mobid. Updating to dynamic lookup to bypass invalid mobid.

### DIFF
--- a/scripts/zones/Bhaflau_Thickets/IDs.lua
+++ b/scripts/zones/Bhaflau_Thickets/IDs.lua
@@ -61,8 +61,8 @@ zones[xi.zone.BHAFLAU_THICKETS] =
             [16990398] = 16990403, -- -119 -15 -651
         },
         HARVESTMAN         = 16990252,
-        LIVIDROOT_AMOOSHAH = 16990473,
-        DEA                = 16990474,
+        LIVIDROOT_AMOOSHAH = DYNAMIC_LOOKUP,
+        DEA                = DYNAMIC_LOOKUP,
     },
     npc =
     {
--- a/scripts/zones/Bhaflau_Thickets/npcs/qm1.lua
+++ b/scripts/zones/Bhaflau_Thickets/npcs/qm1.lua
@@ -3,9 +3,10 @@
 --  NPC: ??? (Spawn Lividroot Amooshah(ZNM T2))
 -- !pos 334 -10 184 52
 -----------------------------------
-local ID = require("scripts/zones/Bhaflau_Thickets/IDs")
+require("scripts/zones/Bhaflau_Thickets/IDs")
 require("scripts/globals/npc_util")
 -----------------------------------
+local ID = zones[xi.zone.BHAFLAU_THICKETS]
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
--- a/scripts/zones/Bhaflau_Thickets/npcs/qm1.lua
+++ b/scripts/zones/Bhaflau_Thickets/npcs/qm1.lua
@@ -1,12 +1,10 @@
------------------------------------
 -- Area: Bhaflau Thickets
 --  NPC: ??? (Spawn Lividroot Amooshah(ZNM T2))
 -- !pos 334 -10 184 52
 -----------------------------------
-require("scripts/zones/Bhaflau_Thickets/IDs")
+local ID = zones[xi.zone.BHAFLAU_THICKETS]
 require("scripts/globals/npc_util")
 -----------------------------------
-local ID = zones[xi.zone.BHAFLAU_THICKETS]
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
--- a/scripts/zones/Bhaflau_Thickets/npcs/qm2.lua
+++ b/scripts/zones/Bhaflau_Thickets/npcs/qm2.lua
@@ -3,9 +3,10 @@
 --  NPC: ??? (Spawn Dea(ZNM T3))
 -- !pos -34 -32 481 52
 -----------------------------------
-local ID = require("scripts/zones/Bhaflau_Thickets/IDs")
+require("scripts/zones/Bhaflau_Thickets/IDs")
 require("scripts/globals/npc_util")
 -----------------------------------
+local ID = zones[xi.zone.BHAFLAU_THICKETS]
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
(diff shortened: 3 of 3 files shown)


claywar (changes requested):
Additional work needs to be done other than just changing those two definitions to dynamic.  In this case, any reference back needs to be updated as well.

```
-- Area: Bhaflau Thickets
--  NPC: ??? (Spawn Lividroot Amooshah(ZNM T2))
-- !pos 334 -10 184 52
local ID = require("scripts/zones/Bhaflau_Thickets/IDs")
require("scripts/globals/npc_util")
local entity = {}

    if npcUtil.tradeHas(trade, 2578) and npcUtil.popFromQM(player, npc, ID.mob.LIVIDROOT_AMOOSHAH) then -- Trade Oily Blood
        player:confirmTrade()
        player:messageSpecial(ID.text.DRAWS_NEAR)
    end
end
```
The require in the above situation will return nil for `ID.mob.LIVIDROOT_AMOOSHAH` since it is not loading the updated copy of the IDs table, but rather the original in the zone.  This should instead use the zones[] table like:

`local ID = zones[xi.zone.BHAFLAU_THICKETS]`

N3ckB3ard:
> Additional work needs to be done other than just changing those two definitions to dynamic. In this case, any reference back needs to be updated as well.
> 
> ```
> -----------------------------------
> -- Area: Bhaflau Thickets
> --  NPC: ??? (Spawn Lividroot Amooshah(ZNM T2))
> -- !pos 334 -10 184 52
> -----------------------------------
> local ID = require("scripts/zones/Bhaflau_Thickets/IDs")
> require("scripts/globals/npc_util")
> -----------------------------------
> local entity = {}
> 
> entity.onTrade = function(player, npc, trade)
>     if npcUtil.tradeHas(trade, 2578) and npcUtil.popFromQM(player, npc, ID.mob.LIVIDROOT_AMOOSHAH) then -- Trade Oily Blood
>         player:confirmTrade()
>         player:messageSpecial(ID.text.DRAWS_NEAR)
>     end
> end
> ```
> 
> The require in the above situation will return nil for `ID.mob.LIVIDROOT_AMOOSHAH` since it is not loading the updated copy of the IDs table, but rather the original in the zone. This should instead use the zones[] table like:
> 
> `local ID = zones[xi.zone.BHAFLAU_THICKETS]`

Doing this kind of change will require the 

local ID = zones[xi.zone.BHAFLAU_THICKETS]

To be placed within each function to trigger correctly. This can be tested making direct changes to a lua and restarting the map server if you wish to have this local placed outside of the function and not called multiple times. I believe the purpose of this update is to currently create a module due to LSB standards of calling the local in every function that is pulling incorrect zone IDs.

As this issue has been pending for some time now related to the map log errors it would seem more efficient to update the IDs to dynamic and create modules to handle this. If you have a solution to fix this I can assure you the current method of initiation of that local fails outside of a function. We reference the IDs lua twice per ZNM

#2429 was the topic used as a support for this fix and your mentioned method does infact correct this. A adjustment to the direct local ID = was not used

N3ckB3ard:
> 

Unless you mean entirely removing 
local ID = require("scripts/zones/Bhaflau_Thickets/IDs")

And replacing this with
local ID = zones[xi.zone.BHAFLAU_THICKETS]

In 2351 this adjustment was not made to the direct luas of the ZNM's so having a reference or understanding of this will help support future use of this modification to the LUA files. A PR can also be made to fix Arrapago Reef as these are also missing this local.

claywar:
You will need to replace those variable definitions in any scripts that currently use those mobs that are now dynamically looked up.  I believe theres two scripts currently, one that I posted above.

N3ckB3ard:
> You will need to replace those variable definitions in any scripts that currently use those mobs that are now dynamically looked up. I believe theres two scripts currently, one that I posted above.

unknown-18

qm2.lua 
```lua
-- Area: Bhaflau Thickets
--  NPC: ??? (Spawn Dea(ZNM T3))
-- !pos -34 -32 481 52
local ID = zones[xi.zone.BHAFLAU_THICKETS]
require("scripts/globals/npc_util")
local entity = {}

    if npcUtil.tradeHas(trade, 2576) and npcUtil.popFromQM(player, npc, ID.mob.DEA) then -- Trade Olzhiryan Cactus
        player:confirmTrade()
        player:messageSpecial(ID.text.DRAWS_NEAR)
    end
end

    player:messageSpecial(ID.text.SHED_LEAVES)
end

entity.onEventUpdate = function(player, csid, option)
end

entity.onEventFinish = function(player, csid, option)
end

return entity
```

It seems this error is persistent on reboot. If file watcher sees a modification to this LUA it will then correct the ZoneID. Could this potentially be a initialize related problem? I do know placing the local within the entity.onTrade = function will fix this and work on reboot.

EpicTaru:
Isn't the better resolution to find the new MobIDs of those that shifted and just update the IDs.la file with those new MobIDs? Just trying to understand why we're converting mobs to DYNAMIC_LOOKUPS.

claywar:
> It seems this error is persistent on reboot. If file watcher sees a modification to this LUA it will then correct the ZoneID. Could this potentially be a initialize related problem? I do know placing the local within the entity.onTrade = function will fix this and work on reboot.

Try requiring 'scripts/globals/zones' prior to the ID definition, not sure of the load order.

claywar:
> Isn't the better resolution to find the new MobIDs of those that shifted and just update the IDs.la file with those new MobIDs? Just trying to understand why we're converting mobs to DYNAMIC_LOOKUPS.

This is a do it once and never have to deal with it again operation, as opposed to having to search through all of those lua files each month (and no current automation in place to do it).  Its a small up front cost to converting to save headaches in the future.

EpicTaru:
> This is a do it once and never have to deal with it again operation, as opposed to having to search through all of those lua files each month (and no current automation in place to do it). Its a small up front cost to converting to save headaches in the future.

Gotcha, understood, thanks for the clarification, Claywar. 👍🏼

claywar (approved):

PR #2610, opened 2022-08-26
Add Commitment Effect Handling, Associated Bands, Update Corsairs Roll

neuromancerxi:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

Adds handling for COMMITMENT effect in charutils.cpp
Adds handling for DEDICATION / COMMITMENT items to item_utils.lua
Creates/Updates Scripts for DEDICATION / COMMITMENT items
Adds/Updates SQL for item_equipment and item_usable
Updates Corsair's Roll to handle CAPACITY_BONUS
Adds TODO note in Corsair's Roll about Exemplar Points (Not Implemented)

References:

https://www.bg-wiki.com/ffxi/Corsair's%20Roll
https://ffxiclopedia.fandom.com/wiki/Commitment
https://ffxiclopedia.fandom.com/wiki/Dedication

Retail Observation used for Effect Application: Chariot -> Capacity -> Echad -> Trizek

Closes #1856 

## Steps to test these changes

Create a fresh Test GM Character
`!changejob COR 99`
`!addkeyitem limit_breaker`
`!addkeyitem job_breaker`
`!additem 5493`

**Testing Corsair's Roll Changes**

Use Corsair Die
Use Corsair's Roll

Confirm Corsair's Roll provides EXP_BONUS and CAPACITY_BONUS at the same levels.
Double Up (and hopefully don't bust) Confirm new values.
`!getmod EXP_BONUS`
`!getmod CAPACITY_BONUS`

**Testing Commitment / Dedication Effects**
`!additem 27556` (Echad - Dedication)
`!additem 27557` (Trizek - Commitment)
`!additem 28569` (Expertise - Dedication)
`!additem 28546` (Capacity - Commitment)
`!additem 15198` (Sprout Beret - Dedication)
`!additem 15199` (Guide Beret - Commitment)

**Gaining / Losing Effects**

Equip and Use Sprout Beret. Note Dedication Effect.
Equip and use Guide Beret. Note Dedication Effect drops silently and is replaced with Commitment Effect.
Equip and Use Expertise Ring. Note Commitment Effect drops silently and is replaced with Dedication Effect.

*Note*: Icons look very similar, but the status effects are different:

[...]

PR #2623, opened 2022-08-30
Create capacity_ring.lua

Daywhite83:
**_I affirm:_**
- [x ] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)


RAIST5150:
Handled with #2610... the SQL usable entries are also created for the rings in that PR as well.

PR #2647, opened 2022-09-04
Fix "/heal on" command so that it does not stop healing

JamesMcMinn:
**_I affirm:_**

This fixes an issue with the `/heal` command where the `on` subcommand is ignored, meaning that `/heal on` acts as a toggle for the healing state, rather than ensuring that the PC was healing.

The PR simply adds a guard (similar to that used above in the `ANIMATION_NONE` case for `/heal off`) which checks if `on` was specified and does nothing if the PC is already healing.

1. Have a PC healing (e.g. by using `/heal`)
2. Use `/heal on` 

Expected behaviour: 
 - The PC stays healing 

Current behaviour:
 - The PC stops healing

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -6156,6 +6156,11 @@ void SmallPacket0x0E8(map_session_data_t* const PSession, CCharEntity* const PCh
         break;
         case ANIMATION_HEALING:
         {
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):
Seems reasonable

claywar (approved):

zach2good (approved):

PR #2656, opened 2022-09-05
Fix for Chocobo Jig II targeting issue

SirGouki:
**_I affirm:_**

Fixes Chocobo Jig II not targeting party members, only the user and their trusts by changing the range from 0.0f to 10.0f.
This value was obtained using the in client AoE ring that shows up when trying to use Chocobo Jig II.

Prior to the change:
Be DNC level 70+ and the party leader.
Have a trust and a party member, with the party member not at the same exact location as the leader.
Use Chocobo Jig II:
- Party Leader and trust will gain the effect
- Party member player will NOT gain the effect.
After the change:
- Party Leader, trust, and party member player will all gain the effect.


SirGouki:
PR is trying to include where I fixed the white space in my CPP from a previous PR, not sure why because that should have already been merged from the previous PR.  I'm going to try to resolve it.

TeoTwawki:
rebase, force push onto same branch name and it will auto update the pr.

SirGouki:
I dropped my fork and redid it, cause I had a different issue.  Is that going to effect that?

TeoTwawki:
~~If you do a force push to same branch name, I can still reopen this and it will show that content~~ not if the entire fork is deleted

SirGouki:
it's pushed

TeoTwawki:
god dammit shithub.

PR #2663, opened 2022-09-06
Despawn pets when owner dies while engaged

austinclprojects:
**_I affirm:_**

Right now, if a player with a pet dies while engaged the pet will continue to fight the mob until one of them dies. This despawns pets when player that controls them dies. 

Engage a mob with a pet and then die, pet should despawn immediately now.


src/map/entities/charentity.cpp
```diff
@@ -2007,6 +2008,12 @@ void CCharEntity::Die()
     }
 
     battleutils::RelinquishClaim(this);
+
+    if (this->PPet)
```
WinterSolstice8, line 2012: Thanks for the PR,

This isn't the right behavior for charmed pets, please add this && into your if check:

`&& this->PPet->objtype == TYPE_PET`

Charmed pets will need to have their allegiance changed back to mob where they should despawn if out of their roaming range, but that will probably require a more in depth check. I can look into that behavior more closely later.

WinterSolstice8 (changes requested):

src/map/entities/charentity.cpp
austinclprojects: I based the logic off of [this](https://github.com/LandSandBoat/server/blob/0aabdea4a3ca9bfb68136e9c6f54dc3ade3cd364/src/map/utils/battleutils.cpp#L4856) which handles when a PC gets charmed. Should that be updated as well?

WinterSolstice8: Ok so I dug deeper into the code. Your code is fine as is. Apparently "despawnPet" actually checks for types and handles it accordingly.

WinterSolstice8 (approved):

src/map/entities/charentity.cpp
TeoTwawki: piggy back note: at some point we're going to have to go over every function name and make sure they are appropriately named, because so many are decidedly _not_  as you just saw

PR #2673, opened 2022-09-07
Start porting outdated Garrison impl (part 1)

zach2good:
**_I affirm:_**

The old branch for trying to merge in the partial Garrison impl is wildly out of date and nearly impossible to rebase, so I've ripped out a lot of the guts to try and get them in.

I ran update extractor against the new text IDs so they're all correct

Nothing testable yet


claywar (approved):

scripts/globals/conquest.lua
```diff
@@ -994,6 +995,11 @@ end
 -----------------------------------
 
 xi.conquest.overseerOnTrade = function(player, npc, trade, guardNation, guardType)
+
+    if xi.garrison.onTrade(player, npc, trade, guardNation) then
```
zach2good, line 999: Since garrison and conquest are so tightly coupled, I've opted to put the calls to the garrison stuff in these global helpers for the overseers

Happy to hear other solutions

claywar: Reasonable, and I wouldn't even complain if they were just baked into conquest.

soapytoast: wait.. wasn't this exactly what you argued against when you spoke with Runaway about it 2 months ago? 
image

UynGH:
So what? If it gets in we good. Move along.

scripts/globals/conquest.lua
zach2good: Ah, I wasn't aware I wasn't allowed to change my mind in the face of a massively out-of-date and difficult to rebase piece of work, which I'm going to have to go through by hand to get over the finish line. I'll revert these commits, sorry!

PR #2678, opened 2022-09-08
Add ordering numbers to dbtool migrations

zach2good:
**_I affirm:_**

Adds an ordering index to the migrations for dbtool, so they're executed in the correct order, not alphabetically.

Run dbtool, see the migrations appear in their numerical order


Xaver-DaRed (approved):

cocosolos:
Does this order the migrations correctly? Docs say `os.scandir` [yields results in arbitrary order](https://docs.python.org/3/library/os.html#os.scandir). Does the call to this need to be changed to `for file in sorted(os.scandir("migrations"), key=lambda e: e.name):`?

zach2good:
I tested it on Windows and it came up ordered, but you're right, the order
isn't guaranteed

PR #1478, opened 2022-03-29
[WIP] Nyzul Isle Investigation

Xaver-DaRed:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] that I agree to LandSandBoat's [Limited Contributor License Agreement](https://github.com/LandSandBoat/server/blob/base/.github/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [x] that I have **read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)**
- [x] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

PR #2684, opened 2022-09-08
[nyzul] Zone global, instance, and NPC scripts

claywar:
**_I affirm:_**

Imports zone global, instance script, and NPCs from #1478 
Not exposed to player at this time


Xaver-DaRed (approved):

scripts/globals/conquest.lua
```diff
@@ -1471,8 +1471,6 @@ xi.conquest.onConquestUpdate = function(zone, updatetype)
             end
         end
     end
-
-    xi.garrison.buildNpcTable(zone)
```
cocosolos, line 1475: Looks like this accidentally reverted a garrison change.

claywar: After speaking with Zach, this was intentional.  That buildNpcTable() function was not implemented in base

PR #2687, opened 2022-09-09
Ensure migrations are always run in order

zach2good:
Co-authored-by: Corey <cocosolos@pm.me>

**_I affirm:_**

Ensure migrations are always run in order

Run migrations, see them in the correct order (they were before, but it wasn't guaranteed, maybe by platform?)

### DIFF
--- a/tools/dbtool.py
+++ b/tools/dbtool.py
@@ -46,7 +46,7 @@ except Exception as e:
 
 def populate_migrations():
     migration_list = []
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):

TeoTwawki:
> maybe by platform?

yes. it gets left up to the os, and in the case of your previous version the os you tested with normally does alphabetic by default

PR #2693, opened 2022-09-10
[core] [lua] Add functionality to enable use of job point granted spells

WinterSolstice8:
**_I affirm:_**

Adds a hook in spell.cpp to check out in lua land if a spell is castable or not
Use said hook to check if the spell is a job point gift spell or not and allow casting if you have enough job points and are on the right job
Various sql tweaks to spells (naming, targeting, levels zeroed to disable use without job points)

!changejob RUN 99
!setjobpoints 110
spend >=100 job points
be able to cast `temper`


EpicTaru:
Pure random question (because it's been asked multiple times before on many servers I've lurked on), but is there a possibility to be able to also use this to force the client to allow the use of JP-restricted spells with the ``!addallspells`` command?

WinterSolstice8:
> Pure random question (because it's been asked multiple times before on many servers I've lurked on), but is there a possibility to be able to also use this to force the client to allow the use of JP-restricted spells with the `!addallspells` command?

you'll need an addon to inject an 0x01A with a spell ID that the client normally thinks you can't cast. You could write a module that overrides this functionality to always allow certain spells for certain jobs, but even if you manually type out the spell the client won't send the packet until you meet the criteria. (such as 99 job point RDM still won't send the packet to cast tier 5 nukes)

The same method in theory should also allow casting of "unobtainable" spells for custom stuff if they were implemented.

EpicTaru:
Very very good information to know, tyvm for the explanation!

claywar (approved):

PR #2698, opened 2022-09-11
Restore lost ACP mission checks in bcnm.lua

TeoTwawki:
**_I affirm:_**

Restore lost ACP mission - it's default disabled (commented out) until mob behaviors is complete (Goblin needs teleport+sneak attack mechanic)

1. enable ACP
2. uncomment out the battlefield
3. start your server and see that characters on the appropriate mission can enter said battlefield

### DIFF
--- a/scripts/globals/bcnm.lua
+++ b/scripts/globals/bcnm.lua
@@ -547,6 +547,7 @@ local function checkReqs(player, npc, bfid, registrant)
     local cop      = player:getCurrentMission(xi.mission.log_id.COP)
     local toau     = player:getCurrentMission(xi.mission.log_id.TOAU)
     local wotg     = player:getCurrentMission(xi.mission.log_id.WOTG)
+    local acp      = player:getCurrentMission(xi.mission.log_id.ACP)
     local asa      = player:getCurrentMission(xi.mission.log_id.ASA)
     local natStat  = player:getMissionStatus(player:getNation())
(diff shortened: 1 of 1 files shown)


TeoTwawki:
The CI failure is simply because the variable is unused when the mission is disabled, and in my opinion should be ignored rather than commenting out both lines and potentially confusing people.

WinterSolstice8:
> The CI failure is simply because the variable is unused when the mission is disabled, and in my opinion should be ignored rather than commenting out both lines and potentially confusing people.

I believe there is a function designed just for for this in utils.lua -- `utils.unused(acp)`

TeoTwawki:
> > The CI failure is simply because the variable is unused when the mission is disabled, and in my opinion should be ignored rather than commenting out both lines and potentially confusing people.
> 
> I believe there is a function designed just for for this in utils.lua -- `utils.unused(acp)`

This made my head hurt, and my soul hurt. A whole function devoted to tricking the CI by technically "using" something that was defined so it'll shaddap.

claywar (approved):

WinterSolstice8:
> This made my head hurt, and my soul hurt. A whole function devoted to tricking the CI by technically "using" something that was defined so it'll shaddap.

C++ does it better. I wonder if there's an equivalent we could use for our linter in lua 
https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

TeoTwawki:
back when the ci was travis there actually used to be something I could embed in the commit message to tell it to skip a check.

PR #2721, opened 2022-09-15
Convert 'Greater Colibri' and 'Wivre's to their 'Locus' successors

CatsEyeXI:
**_I affirm:_**

Begin conversion of Wivres and Greater Colibris in Bhaflau Thickets to their 'Locus' successors. 
Captures attached below:
[2022-07-15_11-53.zip](https://github.com/LandSandBoat/server/files/9577400/2022-07-15_11-53.zip)

Visit Bhaflau Thickets and notice that Wivres are no longer called Colibri and vice versa. 

**NOTE: I have two outstanding issues with this work.**

~~1. I cannot get the Locus Wivres to use their skill lists. The list ID's and the skills contained in them haven't changed. Also I can force the mobs to use these abilities using !exec target:useMobAbility(x), stumped on this.~~
~~2. I'm getting an error on startup that doesn't make sense to me. I'm not sure where mob ID 16990487 is being referenced. It's not anywhere in my captures, nor is it in any of our SQLs.~~
Placeholder was added to DB for 16990487 to prevent startup errors.

image


neuromancerxi (changes requested):
It looks like the mob skill list [257](https://github.com/LandSandBoat/server/blob/2e02cec1b4bce72f8359eb53cca3f7baed096267/sql/mob_skill_lists.sql#L1151) has skills [listed](https://github.com/LandSandBoat/server/blob/2e02cec1b4bce72f8359eb53cca3f7baed096267/sql/mob_skills.sql#L1935) that don't have any corresponding [scripts](https://github.com/LandSandBoat/server/tree/base/scripts/globals/mobskills), which might explain why the mobs aren't using any skills.

Please add any missing mobskill scripts.

CatsEyeXI:
> It looks like the mob skill list [257](https://github.com/LandSandBoat/server/blob/2e02cec1b4bce72f8359eb53cca3f7baed096267/sql/mob_skill_lists.sql#L1151) has skills [listed](https://github.com/LandSandBoat/server/blob/2e02cec1b4bce72f8359eb53cca3f7baed096267/sql/mob_skills.sql#L1935) that don't have any corresponding [scripts](https://github.com/LandSandBoat/server/tree/base/scripts/globals/mobskills), which might explain why the mobs aren't using any skills.
> 
> Please add any missing mobskill scripts.

Interesting, so is it possible these have *never* worked properly? I was under the impression they 'broke' when SE added the Locus mobs...

EDIT: yeah creating some dummy scripts caused the mob to start using the abilities.

scripts/globals/mobskills/batterhorn.lua
```diff
@@ -0,0 +1,30 @@
...
+require("scripts/globals/mobskills")
+-----------------------------------
+local mobskill_object = {}
+
+mobskill_object.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskill_object.onMobWeaponSkill = function(target, mob, skill)
+    local numhits = math.random(2, 3)
```
WinterSolstice8: "absorb 2-3 shadows" doesn't necessarily mean it's 2-3 hits, because its a conal ability it will naturally wipe 1+ shadows depending on ninjutsu skill. This is probably just one hit. Need retail verification. For now, 1 would suffice, leave in a TODO for it.

scripts/globals/mobskills/granite_skin.lua
```diff
@@ -0,0 +1,23 @@
...
+require("scripts/globals/status")
+-----------------------------------
+local mobskill_object = {}
+
+mobskill_object.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskill_object.onMobWeaponSkill = function(target, mob, skill)
+    xi.mobskills.mobBuffMove(mob, xi.effect.PHYSICAL_SHIELD, 3, 0, 30)
```
WinterSolstice8, line 19: Granite Skin is a buff that makes targets in front of it deal 0 physical damage. This will need a new effect and some kind of check to zero it properly. I think this is beyond the scope of this PR but it does need to be noted.

scripts/globals/mobskills/demoralizing_roar.lua
```diff
@@ -0,0 +1,31 @@
...
+    local typeEffect = xi.effect.ATTACK_DOWN
+    local power = 500
+    local duration = 60
+
+    xi.mobskills.mobStatusEffectMove(mob, target, typeEffect, power, 0, duration)
+
+    local dmgmod = 1
+    local info = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg() * 3.2, xi.magic.ele.WATER, dmgmod, xi.mobskills.magicalTpBonus.MAB_BONUS, 1)
+    local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.WATER, xi.mobskills.shadowBehavior.WIPE_SHADOWS)
+    target:takeDamage(dmg, mob, xi.attackType.MAGICAL, xi.damageType.WATER)
```
WinterSolstice8: I don't think this is supposed to deal damage. It's a debuff.

WinterSolstice8 (changes requested):

WinterSolstice8 (approved):

WinterSolstice8:
Force push from me was to rebase to clear up conflict

Issue #1155, opened 2022-01-02
[Battlefields] Player DRG Wyvern not autoattacking in CoP battles/Limbus


Skerxan:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: EDITME

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

I noticed through various battlefields that my wyvern wouldnt auto attack my target. Atleast in every Promy boss fight and in Temenos. Can test other battlefields if needed.

PR #2743, opened 2022-09-17
Fix issue with called Wyvern pets not being added to battlefields

jmcmorris:
**_I affirm:_**

Fixes an issue with Wyvern pets not being added to a battlefield when its master is. This causes mobs to behavior incorrectly with the Wyvern.

```
!zone 139
!additem 1177
!changejob drg 75
```
Call your Wyvern and trade the Comet Orb to initiate Eye of the Tiger. Engage with Gerjis and allow your Wyvern to take hate. Once Gerjis is targeting your Wyvern then use `!tp 3000` while targeting Gerjis to see Gerjis correctly perform a TP attack on your Wyvern. Without this fix it would fail.

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -302,6 +302,12 @@ bool CBattlefield::InsertEntity(CBaseEntity* PEntity, bool enter, BATTLEFIELDMOB
                 PChar->ClearTrusts();
                 luautils::OnBattlefieldEnter(PChar, this);
                 charutils::SendTimerPacket(PChar, GetRemainingTime());
+
+                // Try to add the player's pet in case they have one that can
+                if (PChar->PPet != nullptr)
+                {
+                    InsertEntity(PChar->PPet, true);
+                }
             }
             else if (!IsRegistered(PChar))
             {
@@ -330,7 +336,12 @@ bool CBattlefield::InsertEntity(CBaseEntity* PEntity, bool enter, BATTLEFIELDMOB
 
             if (pet && pet->PMaster && pet->PMaster->objtype == TYPE_PC)
             {
-                // dont enter player pet
+                // Only allow Wyvern pets to enter
+                if (pet->getPetType() == PET_TYPE::WYVERN)
+                {
+                    pet->m_bcnmID        = GetID();
+                    pet->m_battlefieldID = GetArea();
+                }
             }
             else
             {
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -336,12 +336,9 @@ bool CBattlefield::InsertEntity(CBaseEntity* PEntity, bool enter, BATTLEFIELDMOB
 
             if (pet && pet->PMaster && pet->PMaster->objtype == TYPE_PC)
             {
-                // Only allow Wyvern pets to enter
-                if (pet->getPetType() == PET_TYPE::WYVERN)
-                {
-                    pet->m_bcnmID        = GetID();
-                    pet->m_battlefieldID = GetArea();
-                }
+                // Properly set the existing pet to exist within this battlefield
+                pet->m_bcnmID        = GetID();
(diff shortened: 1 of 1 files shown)


claywar (approved):

zach2good:
I'm on my phone travelling at the moment, but I have a question:

If you are SMN and you have an avatar out before you enter a BCNM and it's then dismissed, is it still on the registered entities list after this change?

zach2good:
Or BST, or Pup, or anything else that might have a pet disappear on entry

jmcmorris:
> Or BST, or Pup, or anything else that might have a pet disappear on entry

I just tested this and the pet is being despawned when the level restriction is set. Are pets removed when there is no level restriction for a battlefield or do all battlefields have a "level restriction" even if it is set to max level? I believe it is the latter but I never really did battlefields with pet jobs.

WinterSolstice8:
I can do a quick check of this behavior with battlefields tomorrow on retail regarding pup/smn/drg with pets out already.

The wyvern is never despawned when joining any battlefield or "confrontation" (unity/voidwatch etc) to memory, _but_ it's level is changed to the new cap (if any) and any accumulated EXP for additional "wyvern levels" is erased when that happens. Super annoying to have to waste a spirit link every time you start a unity fight...

WinterSolstice8:
I went to a BC30 with lvl 99 SMN + PUP + DRG with pets out and the SMN + DRG kept their pets (but their levels got reduced)
The PUP got this message:
image
It seems that maybe the pets are silently deleted and replaced with lower level versions during the cutscenes, or perhaps SE's function to change automaton levels calls some Activate code and prints a message that's good enough. 

Will track IDs of the pet to see how it goes

claywar:
Can confirm with PUP.  Elemental capacity from job points is removed (bit me when I first did Up in Arms without planning it)

WinterSolstice8:
I just noticed that my Activate timer was reset. Interesting, I guess that means Deactivate's reset of Activate timer at full HP was triggered when that message hit.

WinterSolstice8:
I took off all attachments and went into a BC30:

The IDs stay the same, but I did notice the current/max HP via addons of the Automaton change mid cutscene. Safe to say that SE tries to preserve your pet when possible, which means we have a bug (at least for entering BCs)

WinterSolstice8:
Changed to BST, summoned courier carrie as lvl 99 bst before going into a BC30, the crab stayed with me and entered the BC.

I did also confirm that Activate got reset as soon as the PUP entered with elemental values set too high. Quite forgiving for SE.

src/map/battlefield.cpp
```diff
@@ -330,7 +336,12 @@ bool CBattlefield::InsertEntity(CBaseEntity* PEntity, bool enter, BATTLEFIELDMOB
 
             if (pet && pet->PMaster && pet->PMaster->objtype == TYPE_PC)
             {
-                // dont enter player pet
+                // Only allow Wyvern pets to enter
+                if (pet->getPetType() == PET_TYPE::WYVERN)
```
WinterSolstice8: Based on retail observations, it seems _all_ pets are allowed, even if we erroneously despawn some of them on level sync right now, but that's out of scope for this PR.

WinterSolstice8 (changes requested):

src/map/battlefield.cpp
jmcmorris: I was just writing out a comment stating exactly this. I will remove the check for wyverns and we can deal with not despawning the pets during level sync in another PR in the future. Is this something that SE changed at some point?

WinterSolstice8: I only played DRG throughout retail, so I can only comment that it always worked for DRG based on memory. Can't say for anything else.

WinterSolstice8 (approved):

jmcmorris:
Fixed https://github.com/LandSandBoat/server/issues/1155

PR #2754, opened 2022-09-19
Battlefield Groups

jmcmorris:
**_I affirm:_**

Adding Battlefield groups to support logic associated with a group of mobs. The initial goal for this was to make Limbus mob groups easier to implement but this can easily be used to implement other battlefields. One of the main advantages of this is that it reduces a great deal of repeated logic spread through mob files and localizes it in a single script.

Below is an example of how this groups will look in script. @zach2good suggested I get the mob ids through another means which I'm going to look into doing for the real thing.

```lua
local groups =
{
    -- Floor 1
    {
        -- Boss
        mobs = { "Ghost_Clot" },
        stationary = true,
        mods =
        {
            [xi.mod.IMPACT_SDT] = 0,
            [xi.mod.HTH_SDT] = 0,
            [xi.mod.SLASH_SDT] = 1250,
        },
        death = function(mob, count)
            xi.limbus.openDoor(mob:getBattlefield(), ID.npc.APOLLYON_SE_PORTAL[1])
        end,
    },
    {
        mobs = { "Metalloid_Amoeba" },
        stationary = true,
        mods =
        {
            [xi.mod.IMPACT_SDT] = 0,
            [xi.mod.HTH_SDT] = 0,
            [xi.mod.SLASH_SDT] = 1250,
        },
        setup = function(mobs)
            print("Setup Limbus")
            print(mobs)
            for _, mob in pairs(mobs) do
                print("setup mob"..mob:getID())
            end
        end,
        death = function(mob, count)
            print("Metalloid Amoeba "..mob:getID().." : Death Count: "..count);
            if count == 2 then
                -- xi.limbus.spawnFrom(mob, ID.npc.APOLLYON_SE_TIME_CRATES[1])
            elseif count == 4 then
                -- xi.limbus.spawnRecoverFrom(mob, ID.npc.APOLLYON_SE_RECOVER_CRATES[1])
            elseif count == 8 then
                -- xi.limbus.spawnFrom(mob, ID.npc.APOLLYON_SE_ITEM_CRATES[1])
            end
        end,
        allDeath = function(mob)
            print("all dead")
        end,
        randomDeath = function(mob)
            print("random dead"..mob:getID())
        end,
    },
    {
        mobs = { "Metalloid_Amoeba", "Ghost_Clot" },
        allDeath = function(mob)
            print("all dead including boss")
        end,
    }
}

battlefield:addGroups(groups)
```

Take the above example and place it into `se_apollyon.lua` and place the addGroups function into `onBattlefieldInitialise`.
```
!zone Apollyon
!addkeyitem cosmo_cleanse
!addkeyitem black_card
```
Enter into SE Apollyon and kill some mobs to see print statements pop up in xi_map console.

### DIFF
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -244,6 +244,40 @@ void CZone::SetBackgroundMusicNight(uint8 music)
     m_zoneMusic.m_songNight = music;
 }
 
+const QueryByNameResult_t& CZone::queryEntitiesByName(std::string const& name)
+{
+    TracyZoneScoped;
+
+    // Use memoization since lookups are typically for the same mob names
+    auto result = m_queryByNameResults.find(name);
+    if (result != m_queryByNameResults.end())
+    {
+        return result->second;
+    }
+
+    std::vector<CBaseEntity*> entities;
+
+    // TODO: Make work for instances
+    ForEachNpc([&](CNpcEntity* PNpc)
+               {
+                   if (std::string((const char*)PNpc->GetName()) == name)
+                   {
+                       entities.push_back(PNpc);
+                   }
+               });
+
+    ForEachMob([&](CMobEntity* PMob)
+               {
+                   if (std::string((const char*)PMob->GetName()) == name)
+                   {
+                       entities.push_back(PMob);
+                   }
+               });
+
+    m_queryByNameResults[name] = std::move(entities);
+    return m_queryByNameResults[name];
+}
+
 uint32 CZone::GetLocalVar(const char* var)
 {
     return m_LocalVars[var];
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -258,6 +258,7 @@ const QueryByNameResult_t& CZone::queryEntitiesByName(std::string const& name)
     std::vector<CBaseEntity*> entities;
 
     // TODO: Make work for instances
+    // clang-format off
     ForEachNpc([&](CNpcEntity* PNpc)
     {
         if (std::string((const char*)PNpc->GetName()) == name)
@@ -273,6 +274,7 @@ const QueryByNameResult_t& CZone::queryEntitiesByName(std::string const& name)
             entities.push_back(PMob);
         }
      });
+    // clang-format on
 
     m_queryByNameResults[name] = std::move(entities);
     return m_queryByNameResults[name];
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -259,18 +259,20 @@ const QueryByNameResult_t& CZone::queryEntitiesByName(std::string const& name)
 
     // TODO: Make work for instances
     ForEachNpc([&](CNpcEntity* PNpc)
-               {
-                   if (std::string((const char*)PNpc->GetName()) == name)
-                   {
-                       entities.push_back(PNpc);
-                   } });
+    {
+        if (std::string((const char*)PNpc->GetName()) == name)
+        {
+            entities.push_back(PNpc);
+        }
+    });
 
     ForEachMob([&](CMobEntity* PMob)
-               {
-                   if (std::string((const char*)PMob->GetName()) == name)
-                   {
-                       entities.push_back(PMob);
-                   } });
+    {
+        if (std::string((const char*)PMob->GetName()) == name)
+        {
+            entities.push_back(PMob);
+        }
+     });
 
     m_queryByNameResults[name] = std::move(entities);
     return m_queryByNameResults[name];
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -263,16 +263,14 @@ const QueryByNameResult_t& CZone::queryEntitiesByName(std::string const& name)
                    if (std::string((const char*)PNpc->GetName()) == name)
                    {
                        entities.push_back(PNpc);
-                   }
-               });
+                   } });
 
     ForEachMob([&](CMobEntity* PMob)
                {
                    if (std::string((const char*)PMob->GetName()) == name)
                    {
                        entities.push_back(PMob);
-                   }
-               });
+                   } });
 
     m_queryByNameResults[name] = std::move(entities);
     return m_queryByNameResults[name];
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -851,3 +851,61 @@ void CBattlefield::ForEachAlly(const std::function<void(CMobEntity*)>& func)
         func(ally);
     }
 }
+
+void CBattlefield::addGroup(BattlefieldGroup group)
+{
+    if (group.randomDeathCallback.valid())
+    {
+        group.randomMobId = xirand::GetRandomElement(group.mobIds);
+    }
+    m_groups.push_back(group);
+}
+
+void CBattlefield::handleDeath(CBaseEntity* PEntity)
+{
+    if (PEntity->objtype != TYPE_MOB || m_groups.empty())
+    {
+        return;
+    }
+
+    for (auto& group : m_groups)
+    {
+        for (uint32 mobId : group.mobIds)
+        {
+            if (mobId == PEntity->id)
+            {
+                ++group.deathCount;
+
+                if (group.deathCallback.valid())
+                {
+                    group.deathCallback(CLuaBaseEntity(PEntity), group.deathCount);
+                }
+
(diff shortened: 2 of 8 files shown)


WinterSolstice8:
This is a really good representation of how Limbus works. Does this setup allow to pick a single gigas ahead of time of the 3 on floor 4 of NE Apollyon to open the door to the last level, or just random chance only with a fallback of all dead? I don't think there's a functional difference there though.

jmcmorris:
> This is a really good representation of how Limbus works. Does this setup allow to pick a single gigas ahead of time of the 3 on floor 4 of NE Apollyon to open the door to the last level, or just random chance only with a fallback of all dead? I don't think there's a functional difference there though.

The `randomDeath` works by picking a random mob in the group and when that dies it'll call `randomDeath` passing in the mob.

When the group is added:
https://github.com/LandSandBoat/server/pull/2754/files#diff-b4814a73e4a8ec23d87cecfbfb6a495807b65486a4a7b5a1251259284d9c2bbcR859

When a mob dies in the group:
https://github.com/LandSandBoat/server/pull/2754/files#diff-b4814a73e4a8ec23d87cecfbfb6a495807b65486a4a7b5a1251259284d9c2bbcR903

jmcmorris:
I actually already had this implemented on Eden.
```lua
-- Floor 4
{
    -- Gigas Bosses
    mobs =
    {
        16933099,
        16933100,
        16933101,
    },
    setup = function(mobs)
        GetMobByID(16933099):setMod(dsp.mod.MAGIC_NULL, 100)
        GetMobByID(16933100):setMod(dsp.mod.NULL_RANGED_DAMAGE, 100)
        GetMobByID(16933101):setMod(dsp.mod.NULL_PHYSICAL_DAMAGE, 100)
    end,
    random_death = function(mob)
        dsp.limbus.openDoor(mob:getBattlefield(), ID.npc.APOLLYON_NE_PORTAL[4])
    end,
},
```

WinterSolstice8:
Oh yeah, that does do it. excellent!

WinterSolstice8 (approved):
LGTM

TeoTwawki:
As they IDs tend to be sequential, we can get the 1st ID of the battlefield and then offset from there to get the others. I would wager the ability to get an entity by name was mentioned, but we shouldn't even really need to do that since we've already loaded them into memory from the battlefield data so with a minor edit we could have the battlefield object itself spit out any info that was already tabled (I think the only use of this so far is "getting" the battlefield ID?). 

Unless the plan is to eventually get rid of the sql table data entirely... which I would not be against. In that case get by name is _probably_ the path of least resistance.

jmcmorris:
Initially I was thinking of that the mob id retrieval would be in the script but now that I see how it works I'm going to change the code to have a list of mob names rather than mob ids and in `CLuaBattlefield::addGroups()` we will convert the names to mob ids.

jmcmorris:
> As they IDs tend to be sequential, we can get the 1st ID of the battlefield and then offset from there to get the others. I would wager the ability to get an entity by name was mentioned, but we shouldn't even really need to do that since we've already loaded them into memory from the battlefield data so with a minor edit we could have the battlefield object itself spit out any info that was already tabled (I think the only use of this so far is "getting" the battlefield ID?).
> 
> Unless the plan is to eventually get rid of the sql table data entirely... which I would not be against. In that case get by name is _probably_ the path of least resistance.

It is true that typically the ids are sequential but I don't want to code that as a requirement. I do believe this could be useful in other battlefield content and those may have a group of various mobs. Such as you might want to say "when all slimes and crabs die do this thing" and it'll work. If mobs in the zone share a name but we don't want them to be all included in a group then we can change the name of some of those mobs.

TeoTwawki:
> It is true that typically the ids are sequential but I don't want to code that as a requirement. I do believe this could be useful in other battlefield content and those may have a group of various mobs. Such as you might want to say "when all slimes and crabs die do this thing" and it'll work. If mobs in the zone share a name but we don't want them to be all included in a group then we can change the name of some of those mobs.

even if the mobs we specifically want to call out aren't all together, everything in the battlefield will be. from there is a matter of knowing how far apart this group of crabs if from the starting point, and this other group of bats form the same starting point, etc. I'd really dislike if we did a lookup of every single one by name. I was going to explain why but I can't seem to locate the function I was about to criticize.

jmcmorris:
> > It is true that typically the ids are sequential but I don't want to code that as a requirement. I do believe this could be useful in other battlefield content and those may have a group of various mobs. Such as you might want to say "when all slimes and crabs die do this thing" and it'll work. If mobs in the zone share a name but we don't want them to be all included in a group then we can change the name of some of those mobs.
> 
> even if the mobs we specifically want to call out aren't all together, everything in the battlefield will be. from there is a matter of knowing how far apart this group of crabs if from the starting point, and this other group of bats form the same starting point, etc. I'd really dislike if we did a lookup of every single one by name. I was going to explain why but I can't seem to locate the function I was about to criticize.

`queryEntitiesByName`?

TeoTwawki:
I swear to god I went through the entire thing search for any instances of "byname" (case insesnsitive) and it missed it.

TeoTwawki:
ok yeah, it's like I remember looks through the entire zone comparing everything. I think we could do lookup for battlefield instance (not instance instance, you know what I mean..english is hard) and not have to check all of them but still get all the ids into your tables.

jmcmorris:
> ok yeah, it's like I remember looks through the entire zone comparing everything. I think we could do lookup for battlefield instance (not instance instance, you know what I mean..english is hard) and not have to check all of them but still get all the ids into your tables.

Yeah, it isn't great. I'm refactoring the code some moving `CLuaZone::queryEntitiesByName` over to `CZone::::queryEntitiesByName` and my plan was to add in memoization for this. Once we do that then the cost of calling this function will be negligible as we almost always call it with the same names. How does that sound?

TeoTwawki:
> > ok yeah, it's like I remember looks through the entire zone comparing everything. I think we could do lookup for battlefield instance (not instance instance, you know what I mean..english is hard) and not have to check all of them but still get all the ids into your tables.
> 
> Yeah, it isn't great. I'm refactoring the code some moving `CLuaZone::queryEntitiesByName` over to `CZone::::queryEntitiesByName` and my plan was to add in memoization for this. Once we do that then the cost of calling this function will be negligible as we almost always call it with the same names. How does that sound?

that sound much better really. See also: thing I just said on discord. might also appeal to you. Depends on future plans really. either way is fine.

jmcmorris:
Alright! Now we are using mob names to define a group rather than raw ids. This makes things much more readable and less error prone. I've updated the example to show how they are used and added a third group that combines both the normal mobs and the floor boss.

src/map/zone.cpp
```diff
@@ -244,6 +244,38 @@ void CZone::SetBackgroundMusicNight(uint8 music)
...
+    auto result = m_queryByNameResults.find(name);
+    if (result != m_queryByNameResults.end())
+    {
+        return result->second;
+    }
+
+    std::vector<CBaseEntity*> entities;
+
+    // TODO: Make work for instances
+    ForEachNpc([&](CNpcEntity* PNpc)
```
claywar, line 262: Minor formatting here (clang-format always seems to trip up on lambdas)

claywar (approved):
👍

TeoTwawki (approved):

src/map/zone.cpp
jmcmorris: What's the formatting issue? I now know why there was a clang-format off above the lambdas.

TeoTwawki: bracing/alignment
https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md#lambdas

jmcmorris: Alright, I believe that is the correct formatting

jmcmorris: The change I made look correct to me according to the second example in contributing; however, the sanity checker is complaining that I need to change it back to what I had before. Is this a discrepancy between clang formatter and contributing guidelines or am I missing something?

TeoTwawki: looks correct to me now, I think ci end or formatter is making the mistake

jmcmorris:
I think that failure isn't my fault. Can someone kick it to make it try again?

TeoTwawki:
> I think that failure isn't my fault. Can someone kick it to make it try again?

yah it's the random startup crash I've been seeing. Hamsters have been threatened, wheel spinning once more.

PR #2765, opened 2022-09-20
Fixing various issues with Sagheera

### LINKED ISSUE #2648: 🐛 [NPC] Sagheera Port Jeuno, incorrect behavior and missing options

## Branch affected by issue 

`base`

## Steps to reproduce
The NPC Sagheera in Port Jeuno will sell the key item Cosmo-Cleanse to players with zero COP mission progression. 
The NPC Sagheera will accept ancient beastcoins and exchange them for items to the player with zero COP mission progress. 
The NPC Sagheera will accept ancient beastcoins but, there is no options to return the coins to the player.
The NPC Sagheera will not accept unused limbus chips in exchange for ancient beastcoins.

## Expected behavior
Buying the key item Cosmo-Cleanse and exchanging ancient beastcoins should be locked behind completing COP mission 7-5.
The NPC Sagheera should have an option to return stored ancient beastcoins to the player.
The NPC Sagheera should exchange unused limbus chips for ancient beastcoins.

## Sources of fact
https://www.bg-wiki.com/ffxi/Sagheera
https://ffxiclopedia.fandom.com/wiki/Sagheera
http://wiki.ffo.jp/html/3758.html
http://wiki.ffo.jp/html/27098.html
https://forum.square-enix.com/ffxi/threads/24154

ffxi_2022 09 03_23 18 52

jmcmorris:
**_I affirm:_**

Fixes #2648
Adds CoP check, proper menu display, ability to retrieve stored ABCs and ability to trade chips for ABCs
In order to implement trading chips properly I added in the ability to get a player's trade container outside of `onTrade`. The client does send the quantity of ABCs back to the server so we could technically just use that but then we open ourselves up to an exploit so instead I verify the trade.

```
!pos -3 0.1 -9 246
!additem 1875 10
!additem 1904
!additem 1988
```

Interact with Sagheera in all the various ways


scripts/zones/Port_Jeuno/npcs/Sagheera.lua
```diff
@@ -332,15 +362,53 @@ entity.onTrigger = function(player, npc)
     then
         player:startEvent(313)
 
-    -- DEFAULT DIALOG (menu)
+    -- Prevent interaction until player has progressed through COP enough
+    elseif player:getCurrentMission(xi.mission.log_id.COP) < xi.mission.id.cop.GARDEN_OF_ANTIQUITY then
+        player:showText(npc, ID.text.SAGHEERA_NO_LIMBUS_ACESS)
```
WinterSolstice8: `ACESS` -> `ACCESS`

scripts/zones/Port_Jeuno/IDs.lua
```diff
@@ -43,6 +43,7 @@ zones[xi.zone.PORT_JEUNO] =
         CONQUEST                      = 7404, -- You've earned conquest points!
         SAGHEERA_MAX_ABCS             = 8045, -- I apologize, but I cannot take any more <item> from you.
         SAGHEERA_LACK_ABCS            = 8046, -- You have collected the proper materials, but unfortunately you lack the required amount of <item> for payment.
+        SAGHEERA_NO_LIMBUS_ACESS      = 8054, -- I came from the Near East to peddle my wares. I heard that there are simply hordes of gullib--err, that is, I have heard of the great adventurers that roam this land, and wish to be of service to them...
```
WinterSolstice8: `ACESS` -> `ACCESS`

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
```diff
@@ -332,15 +362,53 @@ entity.onTrigger = function(player, npc)
...
+
+        -- DEFAULT DIALOG (menu)
     else
         -- event parameters
+
+        local storedABCs = player:getCurrency("ancient_beastcoin")
+        local playerSagheera = player:getCharVar("Sagheera")
```
TeoTwawki: is it essential this be a charvar and not a localvar, are we needing this to persist beyond the event? sort of skimming atm from my dayjob

WinterSolstice8 (changes requested):

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
jmcmorris: I do believe it needs to be a charvar as it handles the first time you talk to her she has special dialog to sell you a cosmo cleanse and then she gives you a long explanation in other places that I'm fairly certain only happen once ever and not every time you zone.

claywar: To avoid a forever charvar, we could set a value when completing the requisite mission, and then reset to 0

TeoTwawki: normally what we do for "1st time dialog" vars if have chars born with a var set to 1 on create, then delete it on talking (see moghouse explanation event). server ops would then have to back fill it if they care to fort old players.

we've had some internal discussion about moving all those things to their own sperate storage mask (and the pain of then having an enum for them).

is that the only reason tho?

jmcmorris: Ok, so I'll set this value with all important bits when the player finishes the requisite mission and then remove the bits as the player performs the necessary interactions. Sound right?

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
```diff
@@ -414,4 +497,30 @@ entity.onEventFinish = function(player, csid, option)
     end
 end
 
+local handleTradeChipEvent = function(player, option)
+    local trade = player:getTrade()
+    if (npcUtil.tradeSetInList(trade, tier1Chips) and npcUtil.giveItem(player, { { xi.items.ANCIENT_BEASTCOIN, tier1ChipValue } })) then
```
claywar: No need for outer parens here

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
claywar: Yep, sounds good, that way we don't clutter the db after the player has progressed far enough

TeoTwawki: does she do multiple of those one offs? it may be justified as a forever var if so

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
```diff
@@ -251,9 +251,33 @@ local abcShop =
     [20] = { item =  2127, abc =  75 }, -- metal_chip
 }
 
+-----------------------------------
+-- trades chips for ancient beastcoins
+-----------------------------------
+local tier1Chips =
```
claywar, line 257: Since the same logic seems to happen twice for both of these tables, could combine into one with the item as key, with its chipValue as value

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
TeoTwawki: > Ok, so I'll set this value with all important bits when the player finishes the requisite mission and then remove the bits as the player performs the necessary interactions. Sound right?

yeah :+1:

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
jmcmorris: That's what I initially did but I use these tables for checking the trade.
https://github.com/LandSandBoat/server/pull/2765/files#diff-0bab3dbe20460986828316156b5e730a1831f9237a333abc455b87e8a97101e5R316

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
jmcmorris: Whoops, I was starting to merge these two conditions with an or but they're so lengthy ; ;

claywar: For length, might want to consider our multi-line style as well:
```
if
    condition1 and
    not condition2
then

end
```

WinterSolstice8 (approved):

WinterSolstice8:
LGTM, but pending approval from others for quest/db stuff I don't know much about

claywar (approved):

scripts/zones/Port_Jeuno/npcs/Sagheera.lua
```diff
@@ -332,15 +366,52 @@ entity.onTrigger = function(player, npc)
...
+        player:showText(npc, ID.text.SAGHEERA_NO_LIMBUS_ACCESS)
+
+        -- DEFAULT DIALOG (menu)
     else
         -- event parameters
+
+        local storedABCs = player:getCurrency("ancient_beastcoin")
+        local playerSagheera = player:getCharVar("SagheeraInteractions")
+        -- bitfield of menu options
+        local menu = 0
```
claywar, line 380: Mulling over a way to simplify this logic for menu param, but works for now.  Could possible reduce this by setting the initial value of the charvar to the correct bit positions for those 3 bits, and then apply some and-logic for the rest

jmcmorris: That's clever. Before inverting it we could have just done a bit.band but I don't think we can anymore.

TeoTwawki: from previous convo, I had assumed you were going to split the 1 time use flags off from your main mask in this pr. they don't have to be together really, and I thought that would allow you to use charvars for the one time use and get/setLocalVar on the rest

jmcmorris: There are 3 one time uses for the `SagheeraInteractions` charvar.

1. First time buying a cosmo cleanse (bit 0)
2. First time the player selects "Just wanted to Chat" (bit 1)
3. First time the player selects the Relic Restoration option (bit 2)

I don't think any of these can be made into a localvar as it'd make the player repeat the action that usually involves a lengthy explanation.

TeoTwawki: wasn't suggesting those become localvar, was suggesting the parts that are not those, not be charvar and not be in the same mask with them

unless I am misreading everything is all in one charvar atm

TeoTwawki: talked to JMC in discord, I misread something and my comments can be disregarded here

claywar: If I read correctly, the least significant nibble is the one you're using for those 3 bits (0b1011), where bit 2 is always set to true in the script.

Therefore, I think you could xor that nibble, and then tack on the other items in the upper nibble

jmcmorris: How does the latest 2* commits look? Is that what you're thinking?

claywar: Using a web version right now to test out ideas, is bit 2 inverted with an 11 param?  Might need to provide it 0xF

jmcmorris: bit 2 is now conditionally set if you have stored abcs and will be set after use do the xor to set the initial menu state.

PR #2771, opened 2022-09-21
Adjustments to provide persistence to Hydra time-of-death

CatsEyeXI:
**_I affirm:_**

Currently, Hydra will not persist even if added to the `persist_nm_time_of_deaths` module. This is because mobs get initialized and spawned before zones' onInitialize gets called, so the module isn't having the intended effect (in this case) because the mob is already spawned. 

By setting the respawntime to 0 in the database, and handing this process off to Lua, we are preventing the mob from automatically spawning and allowing the module to do the work.

1. kill mob, wait for despawn, watch log for 'writing server variable'.
2. check '[Respawn]Hydra', confirm it has incremented. convert it to earth time
3. restart the map server, notice the mob is not automatically spawned
4. wait for [Respawn]Hydra to == earth time and mob should spawn.

For testing, i decreased the spawn times to 1 - 3 minutes in the Hydra.lua

### DIFF
--- a/scripts/zones/Wajaom_Woodlands/Zone.lua
+++ b/scripts/zones/Wajaom_Woodlands/Zone.lua
@@ -16,8 +16,6 @@ end
 zone_object.onInitialize = function(zone)
     xi.helm.initZone(zone, xi.helm.type.HARVESTING)
     xi.chocobo.initZone(zone)
-    UpdateNMSpawnPoint(ID.mob.HYDRA)
-    GetMobByID(ID.mob.HYDRA):setRespawnTime(math.random(48, 72) * 3600)
 end
 
 zone_object.onZoneIn = function(player, prevZone)
--- a/scripts/zones/Wajaom_Woodlands/Zone.lua
+++ b/scripts/zones/Wajaom_Woodlands/Zone.lua
@@ -16,6 +16,8 @@ end
 zone_object.onInitialize = function(zone)
     xi.helm.initZone(zone, xi.helm.type.HARVESTING)
     xi.chocobo.initZone(zone)
+    UpdateNMSpawnPoint(ID.mob.HYDRA)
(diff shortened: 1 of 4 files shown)


scripts/zones/Wajaom_Woodlands/Zone.lua
```diff
@@ -16,6 +16,8 @@ end
 zone_object.onInitialize = function(zone)
     xi.helm.initZone(zone, xi.helm.type.HARVESTING)
     xi.chocobo.initZone(zone)
+    UpdateNMSpawnPoint(ID.mob.HYDRA)
+    GetMobByID(ID.mob.HYDRA):setRespawnTime(math.random(48, 72) * 3600)
```
KnowOne134, line 20: would you even need this line to spawn it on startup + time unless after a mait to reset times, shouldnt the persist data handle the spawn from a restart?

cocosolos: How is the initial spawn supposed to happen if respawn time is set to 0 in groups and the zone init doesn't manually set the time?

CatsEyeXI: i committed removal of the two lines referenced by @KnowOne134 . Admittedly im traveling starting tomorrow and won't have much time to look at this until next week, if they need to be re-added let me know, just trying to get my affairs in order before leaving ^^;

cocosolos: The persist data will overwrite this respawn time if it exists, but it still needs this for the first time initial spawn (and if you're not using the module).

KnowOne134: yea sorry mine was a question because im unfamiliar with lsb persist code and thinking will this overwrite the persist on everty restart

PR #2775, opened 2022-09-22
Fixed Haste Augments to work correctly 

GravesXI:
**_I affirm:_**
- [x] I have read and understood the [Contributing Guide](ht ] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

was having issue where haste augment wasn't reading like it should it was returning a value of 105 on augmented gear made changes 

i equipped a piece of augmented gear did !getmod 384 it returned 105 value then i changed the value to 1 and the multiplier to 100 and did !getmod 384 again and got the 600 value it should be 

i did the same steps for slow as well with the same results changing the value to -1 and the multiplier to 100 fixed the issue for slow 

Screenshot 2022-09-22 012358

Screenshot 2022-09-22 012407

Screenshot 2022-09-22 014102

Screenshot 2022-09-22 014354

Screenshot 2022-09-22 014411

Screenshot 2022-09-22 014907

### DIFF
--- a/sql/augments.sql
+++ b/sql/augments.sql
@@ -75,8 +75,8 @@ INSERT INTO `augments` VALUES (45,0,366,1,0,0); -- DMG:+1 (melee,not ranged...Ma
 INSERT INTO `augments` VALUES (46,0,366,-1,0,0); -- DMG:-1 (melee,not ranged...Mainhand only?)
 INSERT INTO `augments` VALUES (47,0,380,1,0,0); -- Delay:+1% (melee,not ranged)
 INSERT INTO `augments` VALUES (48,0,380,-1,0,0); -- Delay:-1% (melee,not ranged)
-INSERT INTO `augments` VALUES (49,0,384,100,0,0); -- Haste+1
-INSERT INTO `augments` VALUES (50,0,384,-100,0,0); -- Slow+1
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

TeoTwawki:
same thing happening with damage taken mods, and I don't see why

and this bothers me a lot. we were telling it to use 100 of the mod per point of augment _both ways_ so why is that only working with 1 way and not the other? It should not require counter intuitive multiplier to achieve this, and makes me wonder what other modifiers aren't being applied as intended.

TeoTwawki:
> same thing happening with damage taken mods, and I don't see why
> 
> and this bothers me a lot. we were telling it to use 100 of the mod per point of augment _both ways_ so why is that only working with 1 way and not the other? It should not require counter intuitive multiplier to achieve this, and makes me wonder what other modifiers aren't being applied as intended.

@GravesXI I'll stamp approve on this because it fixes the in game problem, but could you raise an issue about that behavior? in the long run this is going to be a problem for us with other modifiers I am certain. I don't want this getting forgotten and you should not have had to do this to get correct values - should have been the same both ways and this needs looked into.

TeoTwawki (approved):
https://github.com/LandSandBoat/server/pull/2775#issuecomment-1256834510

PR #2781, opened 2022-09-23
Implement Save and Load of Equipment on Job Change

0phie:
**_I affirm:_**

+ Implements the saving of the last equipped gear on a job prior to a job change.
+ Implements the loading of the last equipped gear when changing to a different job.

Notes:
+ This will equip items of the same itemID in different slots by looking for unique pointers.
+ The load function is further down in the call stack as to allow for the propagation of the dual wield trait.
+ The system does allow for the moving of items between containers then re-equipping of them.
+ The system only allows for re-equipping of gear that is field accessible.

+ Tested by swapping between jobs in a Moghouse with varying levels of equipment (sometimes blank slots, sometimes a full equipment screen). 
+ Tested by equipping items from other inventories besides the inventory itself.
+ Tested by moving an item from one inventory to another and changing jobs.
+ Tested equipping gear from Mog Safe which failed as intended.
+ Tested equipping a deleted item which did not equip anything as intended.
+ Tested equipping items with the same itemID in both the same container and in separate containers, which resulted in both items being equipped.
+ Tested with 2 Kunai on a NIN was reequipped when changing from NIN -> another job -> NIN.


src/map/utils/charutils.cpp
```diff
@@ -2302,6 +2302,89 @@ namespace charutils
...
     }
 
+    void SaveJobChangeGear(CCharEntity* PChar)
+    {
+        if (PChar == nullptr)
+        {
+            return;
+        }
+
+        const char* Query  = "REPLACE INTO char_equip_saved VALUES (%u, %u, %u, %u, %u, %u, %u, %u, %u, %u, %u, %u, %u, %u, %u, %u, %u, %u);";
```
claywar: In the spirit of other queries, we may want to explicitly assign these values when inserting

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,89 @@ namespace charutils
...
+        uint16      waist  = (PChar->getEquip(SLOT_WAIST) != nullptr) ? PChar->getEquip(SLOT_WAIST)->getID() : 0;
+        uint16      ear1   = (PChar->getEquip(SLOT_EAR1) != nullptr) ? PChar->getEquip(SLOT_EAR1)->getID() : 0;
+        uint16      ear2   = (PChar->getEquip(SLOT_EAR2) != nullptr) ? PChar->getEquip(SLOT_EAR2)->getID() : 0;
+        uint16      ring1  = (PChar->getEquip(SLOT_RING1) != nullptr) ? PChar->getEquip(SLOT_RING1)->getID() : 0;
+        uint16      ring2  = (PChar->getEquip(SLOT_RING2) != nullptr) ? PChar->getEquip(SLOT_RING2)->getID() : 0;
+        uint16      back   = (PChar->getEquip(SLOT_BACK) != nullptr) ? PChar->getEquip(SLOT_BACK)->getID() : 0;
+
+        sql->Query(Query, PChar->id, PChar->GetMJob(), main, sub, ranged, ammo,
+                   head, body, hands, legs, feet, neck, waist, ear1, ear2, ring1,
+                   +ring2, back);
```
claywar: Errant `+` here?

0phie: Oop sorry that was due to me copying over from another branch.

0phie: Resolved now.

TeoTwawki:
Next time your branch is so fubar you feel you have to start over, tag me to help you get your redo onto the same PR, coz I will. every time we wind up with one closed a replacement pr opened, the comment history is lost from the new one (and we all get extra notifications).

PR #2782, opened 2022-09-23
Implement Save and Load of Equipment on Job Change

0phie:
**_I affirm:_**

+ Implements the saving of the last equipped gear on a job prior to a job change.
+ Implements the loading of the last equipped gear when changing to a different job.

Notes:
+ This will equip items of the same itemID in different slots by looking for unique pointers.
+ The load function is further down in the call stack as to allow for the propagation of the dual wield trait.
+ The system does allow for the moving of items between containers then re-equipping of them.
+ The system only allows for re-equipping of gear that is field accessible.

+ Tested by swapping between jobs in a Moghouse with varying levels of equipment (sometimes blank slots, sometimes a full equipment screen).
+ Tested by equipping items from other inventories besides the inventory itself.
+ Tested by moving an item from one inventory to another and changing jobs.
+ Tested equipping gear from Mog Safe which failed as intended.
+ Tested equipping a deleted item which did not equip anything as intended.
+ Tested equipping items with the same itemID in both the same container and in separate containers, which resulted in both items being equipped.
+ Tested with 2 Kunai on a NIN was reequipped when changing from NIN -> another job -> NIN.

### DIFF
--- a/sql/char_equip_saved.sql
+++ b/sql/char_equip_saved.sql
@@ -1,0 +1,26 @@
+SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";
+-- ----------------------------
+-- Table structure for char_equip_saved
+-- ----------------------------
+DROP TABLE IF EXISTS `char_equip_saved`;
+CREATE TABLE IF NOT EXISTS `char_equip_saved` (
+    `charid` int(10) unsigned NOT NULL,
+    `jobid` tinyint(2) unsigned NOT NULL,
+    `main` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `sub` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `ranged` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `ammo` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `head` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `body` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `hands` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `legs` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `feet` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `neck` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `waist` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `ear1` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `ear2` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `ring1` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `ring2` smallint(5) unsigned NOT NULL DEFAULT '0',
+    `back` smallint(5) unsigned NOT NULL DEFAULT '0',
+    PRIMARY KEY (`charid`, `jobid`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 AVG_ROW_LENGTH=20;
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -2310,6 +2310,109 @@ namespace charutils
         }
     }
 
+    void SaveJobChangeGear(CCharEntity* PChar)
+    {
+        if (PChar == nullptr)
+        {
+            return;
+        }
+
+        const char* Query = "REPLACE INTO char_equip_saved SET \
+                                    charid = %u, \
+                                    jobid = %u, \
+                                    main = %u, \
+                                    sub = %u, \
+                                    ranged = %u, \
+                                    ammo = %u, \
+                                    head = %u, \
+                                    body = %u, \
+                                    hands = %u, \
+                                    legs = %u, \
+                                    feet = %u, \
+                                    neck = %u, \
+                                    waist = %u, \
+                                    ear1 = %u, \
+                                    ear2 = %u, \
+                                    ring1 = %u, \
+                                    ring2 = %u, \
+                                    back = %u;";
+
+        auto getEquipIdFromSlot = [](CCharEntity* PChar, SLOTTYPE slot) -> uint16
+        {
+            return (PChar->getEquip(slot) != nullptr) ? PChar->getEquip(slot)->getID() : 0;
+        };
+
+        uint16 main   = getEquipIdFromSlot(PChar, SLOT_MAIN);
+        uint16 sub    = getEquipIdFromSlot(PChar, SLOT_SUB);
+        uint16 ranged = getEquipIdFromSlot(PChar, SLOT_RANGED);
+        uint16 ammo   = getEquipIdFromSlot(PChar, SLOT_AMMO);
+        uint16 head   = getEquipIdFromSlot(PChar, SLOT_HEAD);
+        uint16 body   = getEquipIdFromSlot(PChar, SLOT_BODY);
+        uint16 hands  = getEquipIdFromSlot(PChar, SLOT_HANDS);
+        uint16 legs   = getEquipIdFromSlot(PChar, SLOT_LEGS);
+        uint16 feet   = getEquipIdFromSlot(PChar, SLOT_FEET);
+        uint16 neck   = getEquipIdFromSlot(PChar, SLOT_NECK);
+        uint16 waist  = getEquipIdFromSlot(PChar, SLOT_WAIST);
+        uint16 ear1   = getEquipIdFromSlot(PChar, SLOT_EAR1);
+        uint16 ear2   = getEquipIdFromSlot(PChar, SLOT_EAR2);
+        uint16 ring1  = getEquipIdFromSlot(PChar, SLOT_RING1);
+        uint16 ring2  = getEquipIdFromSlot(PChar, SLOT_RING2);
+        uint16 back   = getEquipIdFromSlot(PChar, SLOT_BACK);
+
+        sql->Query(Query, PChar->id, PChar->GetMJob(), main, sub, ranged, ammo,
+                   head, body, hands, legs, feet, neck, waist, ear1, ear2, ring1,
+                   ring2, back);
+    }
+
+    void LoadJobChangeGear(CCharEntity* PChar)
+    {
+        if (PChar == nullptr)
+        {
+            return;
+        }
+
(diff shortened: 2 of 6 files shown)


sql/char_equip_saved.sql
```diff
@@ -0,0 +1,26 @@
+SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";
+-- ----------------------------
+-- Table structure for char_equip_saved
+-- ----------------------------
+DROP TABLE IF EXISTS `char_equip_saved`;
```
claywar, line 5: This should probably be a protected table in dbtool, and not sure if we want to have the drop here (would suggest following whatever is used in other protected tables)

0phie: Added to dbtool.py as a protected table.

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,108 @@ namespace charutils
...
+        uint16 neck   = (PChar->getEquip(SLOT_NECK) != nullptr) ? PChar->getEquip(SLOT_NECK)->getID() : 0;
+        uint16 waist  = (PChar->getEquip(SLOT_WAIST) != nullptr) ? PChar->getEquip(SLOT_WAIST)->getID() : 0;
+        uint16 ear1   = (PChar->getEquip(SLOT_EAR1) != nullptr) ? PChar->getEquip(SLOT_EAR1)->getID() : 0;
+        uint16 ear2   = (PChar->getEquip(SLOT_EAR2) != nullptr) ? PChar->getEquip(SLOT_EAR2)->getID() : 0;
+        uint16 ring1  = (PChar->getEquip(SLOT_RING1) != nullptr) ? PChar->getEquip(SLOT_RING1)->getID() : 0;
+        uint16 ring2  = (PChar->getEquip(SLOT_RING2) != nullptr) ? PChar->getEquip(SLOT_RING2)->getID() : 0;
+        uint16 back   = (PChar->getEquip(SLOT_BACK) != nullptr) ? PChar->getEquip(SLOT_BACK)->getID() : 0;
+
+        sql->Query(Query, PChar->id, PChar->GetMJob(), main, sub, ranged, ammo,
+                   head, body, hands, legs, feet, neck, waist, ear1, ear2, ring1,
```
zach2good, line 2363: Stray + sign?

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,108 @@ namespace charutils
...
+                   +ring2, back);
+
+        return;
```
zach2good: No need to return at the end of a function with a void return type

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,108 @@ namespace charutils
...
+                            {
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        return;
```
zach2good: Same here

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,108 @@ namespace charutils
...
+            return;
+        }
+
+        const char* Query = "SELECT main, sub, ranged, ammo, head, body, hands, legs, feet, neck, waist, ear1, ear2, ring1, ring2, back FROM char_equip_saved AS equip WHERE charid = %u AND jobid = %u;";
+
+        if (sql->Query(Query, PChar->id, PChar->GetMJob()) == SQL_SUCCESS && sql->NumRows() != 0 && sql->NextRow() == SQL_SUCCESS)
+        {
+            for (uint8 i = 0; i < SLOT_LINK1; i++)
+            {
+                if (sql->GetUIntData(i) > 0)
```
zach2good: Don't know if this caches the value or not, consider caching

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,108 @@ namespace charutils
...
+        if (PChar == nullptr)
+        {
+            return;
+        }
+
+        const char* Query = "SELECT main, sub, ranged, ammo, head, body, hands, legs, feet, neck, waist, ear1, ear2, ring1, ring2, back FROM char_equip_saved AS equip WHERE charid = %u AND jobid = %u;";
+
+        if (sql->Query(Query, PChar->id, PChar->GetMJob()) == SQL_SUCCESS && sql->NumRows() != 0 && sql->NextRow() == SQL_SUCCESS)
+        {
+            for (uint8 i = 0; i < SLOT_LINK1; i++)
```
zach2good: Since there are multiple levels of iteration happening here, please consider using more descriptive names than i and x, opt swap to using range-for loops.

Also, use less-than-equals here for the end condition, and the starting slot enum instead of zero, for clarity that you're iterating over equippable inventory slots

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,108 @@ namespace charutils
...
+                                    feet = %u, \
+                                    neck = %u, \
+                                    waist = %u, \
+                                    ear1 = %u, \
+                                    ear2 = %u, \
+                                    ring1 = %u, \
+                                    ring2 = %u, \
+                                    back = %u;";
+
+        uint16 main   = (PChar->getEquip(SLOT_MAIN) != nullptr) ? PChar->getEquip(SLOT_MAIN)->getID() : 0;
```
zach2good: Consider writing a short lambda for this, look in the mannequin stuff in packet system for an example

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,108 @@ namespace charutils
...
+        if (sql->Query(Query, PChar->id, PChar->GetMJob()) == SQL_SUCCESS && sql->NumRows() != 0 && sql->NextRow() == SQL_SUCCESS)
+        {
+            for (uint8 i = 0; i < SLOT_LINK1; i++)
+            {
+                if (sql->GetUIntData(i) > 0)
+                {
+                    for (int x = 0; x < CONTAINER_ID::MAX_CONTAINER_ID; x++)
+                    {
+                        bool found = false;
+                        if (x == LOC_INVENTORY || (x > LOC_MOGLOCKER && x < LOC_RECYCLEBIN))
```
zach2good: Maybe make a function near wherever these location enums are defined for 'Is Valid Storage Location'

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,108 @@ namespace charutils
...
+                        {
+                            for (uint8 slot = 0; slot < PChar->getStorage(x)->GetSize(); slot++)
+                            {
+                                CItem* PItem = PChar->getStorage(x)->GetItem(slot);
+                                if (PItem != nullptr && PItem->getID() == sql->GetUIntData(i))
+                                {
+                                    auto* PEquip   = static_cast<CItemEquipment*>(PItem);
+                                    auto  prevSlot = static_cast<SLOTTYPE>(std::clamp(i - 1, 0, MAX_SLOTTYPE - 1));
```
zach2good: There's got to be a way to roll these safety checks into the conditions above, or those safety checks might already exist in the player equip helper?

sql/char_equip_saved.sql
```diff
@@ -0,0 +1,26 @@
+SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";
```
zach2good, line 1: If you're adding new char sql, then you need to update the char entries in trigger.sql, so this new data will automatically get removed on char deletion

zach2good (changes requested):

src/map/utils/charutils.cpp
0phie: I considered this, but the enums live in item_container.h which would require a PItem. So simply defining this here was meant to save on potentially wasted performance due to the loop not skipping past more code. If you have any ideas of how we could do this or if we want it somewhere like charutils where we don't have to have an item associated with the check that is an option.

sql/char_equip_saved.sql
0phie: Added to trigger.sql, thanks for pointing that out! I was lowkey wondering where the deletion process would take care of it considering the comment in packet_system for char deletion is extremely vague.

src/map/utils/charutils.cpp
0phie: Resolved, was due to a clang fix.

src/map/utils/charutils.cpp
0phie: Removed

src/map/utils/charutils.cpp
0phie: Removed

src/map/utils/charutils.cpp
0phie: Consider it cached now. However, I believe that the value is cached once the query is made, however doesn't hurt to just assign it somewhere in memory temporarily to ensure it is.

src/map/utils/charutils.cpp
0phie: Clarified and utilized less-than-equals to ensure it is understandable.

src/map/utils/charutils.cpp
0phie: Lambda written and applied same as mannequins.

src/map/utils/charutils.cpp
0phie: Rolled up and buttoned up nicely. PItem and PEquip would both return a nullptr if something went wrong, so left those two checks, but all SLOTTYPE checks are already handled within the equip helper as you pointed out.

src/map/utils/charutils.cpp
```diff
@@ -2310,6 +2310,106 @@ namespace charutils
...
+                {
+                    for (int container = LOC_INVENTORY; container <= LOC_WARDROBE8; container++)
+                    {
+                        bool found = false;
+                        if (container == LOC_INVENTORY || (container >= LOC_WARDROBE && container <= LOC_WARDROBE8))
+                        {
+                            for (uint8 slot = 0; slot < PChar->getStorage(container)->GetSize(); slot++)
+                            {
+                                CItem* PItem  = PChar->getStorage(container)->GetItem(slot);
+                                auto*  PEquip = static_cast<CItemEquipment*>(PItem);
```
zach2good: For this to do what you expect here, it needs to be a dynamic cast. Static cast won't return a nullptr unless it is given a nullptr

0phie: Thanks updated it in the most recent build.

zach2good (approved):

PR #2784, opened 2022-09-23
[lua] Update Reactor Cool Blue Magic to match Retail Behavior

neuromancerxi:
**_I affirm:_**

Updates Reactor Cool Script:
Reorder effects to give correct "gains the effect of" message. 
Simplify/Remove conditional effect gain checks as Reactor Cool will explicitly (and silently) overwrite DEF BOOST and ICE SPIKES effects 
Add DEF BOOST and ICE SPIKES

Source: Retail observations as BLU99/BLM49 (steps taken to verify below)

Cast Ice Spikes
Cast Cocoon
`/checkparam`
Note Increased Defense +50%
Cast Reactor Cool
Note Event Log "... gains effect of Defense Boost" 
Note Event Log no lose effect messages
`/checkparam`
Note Defense now only increased by +12% instead of +50% 
Note that SE Status Timers reset to higher duration

### DIFF
--- a/scripts/globals/spells/blue/reactor_cool.lua
+++ b/scripts/globals/spells/blue/reactor_cool.lua
@@ -25,10 +25,10 @@ spell_object.onMagicCastingCheck = function(caster, target, spell)
 end
 
 spell_object.onSpellCast = function(caster, target, spell)
-    local typeEffectOne = xi.effect.ICE_SPIKES
-    local typeEffectTwo = xi.effect.DEFENSE_BOOST
-    local powerOne = 5
-    local powerTwo = 12
+    local typeEffectOne = xi.effect.DEFENSE_BOOST
+    local typeEffectTwo = xi.effect.ICE_SPIKES
+    local powerOne = 12
+    local powerTwo = 5
     local duration = 120
     local returnEffect = typeEffectOne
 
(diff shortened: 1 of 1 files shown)


zach2good (approved):

scripts/globals/spells/blue/reactor_cool.lua
```diff
@@ -25,10 +25,10 @@ spell_object.onMagicCastingCheck = function(caster, target, spell)
 end
 
 spell_object.onSpellCast = function(caster, target, spell)
-    local typeEffectOne = xi.effect.ICE_SPIKES
-    local typeEffectTwo = xi.effect.DEFENSE_BOOST
-    local powerOne = 5
-    local powerTwo = 12
+    local typeEffectOne = xi.effect.DEFENSE_BOOST
```
claywar, line 28: Perhaps these variables could be renamed to reflect the effect type?

neuromancerxi: There are Five other Blue Magic scripts which follow this convention. Seems pretty straight forward and easy to read as-is, imo. 

Can you give me an example of how you're envisioning the variable names?

claywar: The same you say? I take it back, that smells like we should be moving spells like this to the global and tabling data

neuromancerxi: Sounds like subsequent PR territory, that concept is outside of the scope of this one.

claywar: Yup. I'll take a look this weekend for some ideas to clean things up

PR #2785, opened 2022-09-23
Limbus SE Apollyon Area

jmcmorris:
**_I affirm:_**

On a very high level, this refactors the SE Apollyon Limbus area and fixes some missing mechanics within the area. In order to achieve this though there have been many changes necessary. I managed to put many changes into other PRs that have already been merged but this one is still quite beefy. I will do my best to breakdown the changes; however, due to the large number of changes I fully anticipate this PR requiring some iteration.

### [Allow mobs to be triggered while they are "normal" status. Right now …](https://github.com/LandSandBoat/server/commit/68a0c38e2d5404c400e4d46f52d25b488b6944e6)

This change is simply to make it possible to interact with mobs. My understanding is that the client will only send a trigger event to the server when the entity has a normal status which is what gives them a green name. Unfortunately making it possible to interact with a mob via a "onTrigger" in the mob's file would require changing more. We may or may not want to do this?

### [Add support for setting up superlinking within a battlefield group](https://github.com/LandSandBoat/server/commit/67b6e0d74c9b1f844a3bac67251417eead9187ac)

I didn't realize until after the fact that this change isn't needed until a future Limbus area but here it is now!

### [Fixing critical issue with CBaseEntity::IsTargetable()](https://github.com/LandSandBoat/server/commit/c9280a5b55dbbfc6dcf307900f5fc7cbeb32bc39)

I needed a way to prevent a mob that has had its status changed to NORMAL from being hit with AOE spells and discovered that this was coded improperly.

### [Support for recovery crate. It is actually a MOB that looks like an NPC…](https://github.com/LandSandBoat/server/commit/aba7b881d5d2bc4e9352fb1a7f00ecbd7c47d2f1)

It is so beautiful!

### Refactor of SE Apollyon

One of my main objectives with these changes is to reduce duplicated code and bring as much Limbus area related content into a single file as I could. I introduced `LimbusArea` based off of `Mission` that assists with setting everything up. In addition it will create the battlefield object which contains a significant amount of duplicated code that exists in all Limbus area files.

All of this new code exists alongside the majority of the old Limbus code that is still being used for the other areas. Once all areas have been converted over to this new implementation then I will cleanup the old code.

### Regarding Loot

Loot right now is very much WIP as I'm still figuring out how best to balance the relic materials. I'm testing the waters of using the distribution method discussed over at https://github.com/LandSandBoat/server/discussions/2767. My current plan is to finish the main Apollyon areas and then submit a separate PR specifically for reworking the loot in those areas.

```
!zone 38
!addkeyitem cosmo_cleanse
!addkeyitem black_card
```

Enter into SE Apollyon and progress through the area. The following captures were used and can be referenced to verify mechanics are working properly.

https://discord.com/channels/443544205206355968/653844344088035328/704860303141371954
https://discord.com/channels/443544205206355968/653844344088035328/704877361757159494
https://discord.com/channels/443544205206355968/653844344088035328/668861099029954562

Other useful links:
https://ffxiclopedia.fandom.com/wiki/SE_Apollyon
https://www.bg-wiki.com/ffxi/South_East_Apollyon


jmcmorris:
Sanity check is failing due to `scripts/globals/limbus.lua:542:1: (W111) setting non-standard global variable LimbusArea`

How do you go about defining something like this such as for `Mission`? From what I can tell they are the same?

claywar:
Would most likely need an exception in tools/ci/lua.sh

jmcmorris:
> Would most likely need an exception in tools/ci/lua.sh

Ah yes, I see it there now. https://github.com/LandSandBoat/server/blob/base/tools/ci/lua.sh#L53

Thanks! Do I have permission to add that in?

scripts/globals/battlefield.lua
```diff
@@ -57,6 +57,17 @@ xi.battlefield.leaveCode =
...
+{
+    EXTREMELY_LOW  = 2,
+    VERY_LOW       = 10,
+    LOW            = 30,
+    NORMAL         = 50,
+    HIGH           = 70,
+    VERY_HIGH      = 100,
+    EXTREMELY_HIGH = 140,
+}
+
```
Xaver-DaRed, line 70: Where is this used? I cant see it ><

jmcmorris: https://github.com/LandSandBoat/server/pull/2785/files#diff-a986f244ec3d08b169692e68724208e36214c7c5307dcf4b48cb4ee22bfaa4f1R270

src/map/packet_system.cpp
```diff
@@ -842,7 +842,7 @@ void SmallPacket0x01A(map_session_data_t* const PSession, CCharEntity* const PCh
             }
 
             CBaseEntity* PNpc = nullptr;
-            PNpc              = PChar->GetEntity(TargID, TYPE_NPC);
+            PNpc              = PChar->GetEntity(TargID, TYPE_NPC | TYPE_MOB);
```
claywar, line 845: Definitely a fan of this, as I believe it will also help with golden salvage

src/map/entities/npcentity.cpp
```diff
@@ -78,17 +78,17 @@ void CNpcEntity::SetUntargetable(bool untargetable)
 {
     if (untargetable)
     {
-        m_flags |= 0x800;
+        m_flags |= FLAG_UNTARGETABLE;
```
claywar, line 81: 👍

scripts/zones/Apollyon/IDs.lua
```diff
@@ -188,6 +168,69 @@ zones[xi.zone.APOLLYON] =
         APOLLYON_CS_CRATE = 16933126,
         APOLLYON_CENTRAL_CRATE = 16933123,
     },
+    SE_APOLLYON =
+    {
+        mob =
```
claywar, line 173: The mob side of these could probably be moved to dynamic lookups (I'd like to see npcs as well, but there's an open issue that needs to be resolved with load order)

scripts/globals/limbus.lua
```diff
@@ -493,3 +482,189 @@ function xi.limbus.spawnRandomCrate(npc, battlefield, var, mask, canMimic)
...
+end
+
+LimbusArea.name = ""
+LimbusArea.serverVar = ""
+LimbusArea.requiredCard = 0
+LimbusArea.groups = {}
+LimbusArea.paths = {}
+LimbusArea.loot = {}
+
+function LimbusArea:new(name, requiredCard)
```
claywar, line 557: I like the direction this is going, and would like to see this eventually move towards a tabled format similar to how interaction framework is handled.  Perhaps as well, since this seems to wrap battlefields, a more generic base class could be made to implement this design, and then have limbus extend it (similar to the Container class which is utilized by Mission and Quest)

scripts/zones/Apollyon/IDs.lua
jmcmorris: From what I can see it has to be in the format of `ID.mob.<mob_name> = DYNAMIC_LOOKUP` and then it'll find it. It doesn't appear like it will work with mobs that have the same name, is that right? It would be nice to use this but I think the feature needs to be expanded upon first.

scripts/globals/limbus.lua
jmcmorris: I'd be happy to chat with you about this more. Hit me up on discord when you have time. What we have now could work for this PR and then I can expand upon it for a separate PR.

claywar (approved):
I'm good with this.  After discussion, we talked about some ways to improve and generalize this implementation in the future, and think there could be a good direction for applying this structure to battlefields in general.

scripts/zones/Apollyon/IDs.lua
WinterSolstice8: The only realistic way you could do this is by looking up Tieholtsodi then using that ID plus an offset for the adamantshells. There might be something similar before or after the flying spears

WinterSolstice8 (approved):
Lots of good here. Thanks for the PR.

scripts/zones/Apollyon/IDs.lua
jmcmorris: Yeah, I talked to @claywar about how this could be improved. I may put together a PR for it. Got lots to do!

jmcmorris:
I rebased and resolve the merge conflict. I hope I did that right...

PR #2792, opened 2022-09-24
[core] [cmake] [linux] Add "backward" trace stack for Linux

WinterSolstice8:
**_I affirm:_**

Adds a traceback print to linux upon crash, see pic

image

force crash applications, see cool traceback
May need debug mode and not just release with symbols.

Drafted for OSX CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Install Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y software-properties-common cppcheck luajit-5.1-dev luarocks mariadb-server mariadb-client libmariadb-dev-compat
+        sudo apt-get install -y software-properties-common cppcheck luajit-5.1-dev luarocks mariadb-server mariadb-client libmariadb-dev-compat binutils-dev
         pip install -r tools/requirements.txt
         luarocks install luacheck --local
     - id: changed-files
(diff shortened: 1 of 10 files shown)


WinterSolstice8:
Apparently I need to add deps for linux CI, makes sense. Ubuntu wants `binutils-dev`

WinterSolstice8:
`brew install binutils` may work for MacOS depdendencies

EDIT: binutils is sorta part of clang for MacOS and it doesn't need an extra dependency or link

WinterSolstice8:
Ok, I think this is ready for review.

P.S. according to their docs, "backward" supports Windows but I hadn't tested how it compares to our current implementation on Windows.

WinterSolstice8:
One thing I should note: The traceback generation on full debug mode on Linux takes quite some time for my poor 4 core ARM VPS,
```
[09/24/22 18:21:17:784][map][info] Console input thread exiting... (operator():116)
[09/24/22 18:23:16:129][map][critical] Stack trace (most recent call last):
```

Just under 2 minutes.

zach2good (approved):
Looking good 👍

If you're not entirely sick of this, can you move the Linux crash/Debug logic out of kernel and into debug_linux? Trying to kill kernel

WinterSolstice8:
> If you're not entirely sick of this, can you move the Linux crash/Debug logic out of kernel and into debug_linux? Trying to kill kernel

Will look into this, I think "backward" might hook into signals in ways that might not be compatible with the current setup unless you're in a crash state already (which is what is currently being used)

WinterSolstice8:
Latest changes seem to preserve Windows crash handler as I had hoped:

```
[09/25/22 11:08:30:513][map][info] Current State Top: 0, Total Memory Used: 67393kb (luautils::garbageCollectStep:290)
crash
=====================================================
!!! CRASH !!!
Exception code: C0000005 (ACCESS_VIOLATION)
Fault address: 00007FF60EF8DE32 01:000000000009CE32
Process Name: G:\dev\server\xi_map.exe
Full crash report: G:\dev\server\dmp\xi_map.exe_25-9_11-8-42.log
Memory dump: G:\dev\server\dmp\xi_map.exe_25-9_11-8-42.dmp
Time of crash: 2022/09/25 11:08:42
Process Uptime: 84 seconds
```

PR #2799, opened 2022-09-25
Add !skipcs helper

zach2good:
**_I affirm:_**

Adds `!skipcs` and skipEvent() to do as the name implies, there are notes about the limitations in the implementations

- Make a new character
- Zone into Port Jeuno
- Get the long CS
- !skipcs
- Get zoned, and have the CS marked as finished, won't fire again like if you had used `!release`


zach2good:
Gonna pause this until I've tested it with more complex cs's

PR #2805, opened 2022-09-26
Imps regain horns after some time

KnowOne134:
**_I affirm:_**

retail testing shows that Imps regain their horns at varying times of 30, 35, 40, 45, and 60 seconds which is very rare to see
uses the same random that causes the horn to break to apply the time that it will reapply it  (1,19)

fight an imp, with godmode to hit the crits
once horn breaks turn back to it and withon 60s it will get a new one

### DIFF
--- a/scripts/mixins/families/imp.lua
+++ b/scripts/mixins/families/imp.lua
@@ -6,9 +6,24 @@ g_mixins.families = g_mixins.families or {}
 g_mixins.families.imp = function(impMob)
 
     -- 20% chance to break horn on critical hit
+    -- Reacquires horn in stages of 30, 35, 40, 45, and a very rare 60 seconds
     impMob:addListener("CRITICAL_TAKE", "IMP_CRITICAL_TAKE", function(mob)
-        if math.random(100) < 20 and mob:getAnimationSub() == 0 then
+        local random = math.random(100)
+        if random < 20 and mob:getAnimationSub() == 0 then
             mob:setAnimationSub(1)
+            if mob:getLocalVar("hornDisabled") ~= 1 then
+                local time = 0
+                if random <= 2 then
+                    time = 60
+                else
+                    time = 25 + (math.ceil(random / 5) * 5)
+                end
+                mob:timer(time * 1000, function(mobArg)
+                    if mob:isAlive() then
+                        mob:setAnimationSub(0)
+                    end
+                end)
+            end
         end
     end)
 


Xaver-DaRed (approved):

scripts/mixins/families/imp.lua
```diff
@@ -6,9 +6,24 @@ g_mixins.families = g_mixins.families or {}
 g_mixins.families.imp = function(impMob)
 
     -- 20% chance to break horn on critical hit
+    -- Reacquires horn in stages of 30, 35, 40, 45, and a very rare 60 seconds
     impMob:addListener("CRITICAL_TAKE", "IMP_CRITICAL_TAKE", function(mob)
-        if math.random(100) < 20 and mob:getAnimationSub() == 0 then
+        local random = math.random(100)
+        if random < 20 and mob:getAnimationSub() == 0 then
             mob:setAnimationSub(1)
+            if not mob:getLocalVar("hornDisabled") == 1 then
```
Xaver-DaRed: `not` only works with true/false comparations.

I think `if mob:getLocalVar("hornDisabled") ~= 1` would work.

scripts/mixins/families/imp.lua
```diff
@@ -6,9 +6,24 @@ g_mixins.families = g_mixins.families or {}
...
+                local time = 0
+                if random <= 2 then
+                    time = 60
+                else
+                    time = 25 + (math.ceil(random / 5) * 5)
+                end
+                mob:timer(time * 1000, function(mob)
```
Xaver-DaRed: `function(mob)` should be changed to `function(something_else_not_mob)`

scripts/mixins/families/imp.lua
```diff
@@ -6,9 +6,24 @@ g_mixins.families = g_mixins.families or {}
...
             mob:setAnimationSub(1)
+            if mob:getLocalVar("hornDisabled") ~= 1 then
+                local time = 0
+                if random <= 2 then
+                    time = 60
+                else
+                    time = 25 + (math.ceil(random / 5) * 5)
+                end
+                mob:timer(time * 1000, function(mobArg)
+                    if mob:isAlive() then
```
jmcmorris, line 22:
```suggestion
                    if mobArg:isAlive() then
```

scripts/mixins/families/imp.lua
```diff
@@ -6,9 +6,24 @@ g_mixins.families = g_mixins.families or {}
...
+                        mob:setAnimationSub(0)
```
jmcmorris, line 23:
```suggestion
                        mobArg:setAnimationSub(0)
```

jmcmorris (commented):
I'm not 100% sure this is necessary but you don't need to use the upvalue `mob` here since you have `mobArg`.

scripts/mixins/families/imp.lua
cocosolos: Is `mobArg` even valid? It's my understanding that `mob` is already available to the inner function without passing anything as an arg, and nothing is actually passed as an arg anyway so `mobArg` is just nil.

jmcmorris: Yes, mobArg is valid. mob:timer() queues an action and here's the code that handles that action. https://github.com/LandSandBoat/server/blob/base/src/map/ai/helpers/action_queue.cpp#L77

It is true that mob is available inside the anonymous function as an upvalue. My comment was mostly coming from my own experience of avoiding using pointers as upvalues since it could potentially be deleted between creation and invocation, but the reality is that mobs (and most things) are created in memory and stay there so it isn't a real problem here.
TLDR: it kind of doesn't matter :)

cocosolos: Ah I was looking at the struct and somehow missed the handle function, thanks for pointing that out!

WinterSolstice8: essentially, when timer is invoked it calls the function you made directly

```
function(mobArg)
    if mob:isAlive() then
        ...
        end
        ...
end
```

Note that `mob` isn't in scope here, this function _technically_ shouldn't know about `mob`, but it does know about `mobArg`, which will be equivalent to using `mob`  (as that's what's passed in to this function when the timer is invoked)

claywar (approved):

PR #2806, opened 2022-09-26
reset mobs ability list on a full roam heal

KnowOne134:
**_I affirm:_**

resets the ability list a mob generates while fighting on a full heal roam cycle.

fixes scenario:
player A fights mob and either dies or runs away after it did the TP move a bluemage could learn
later down the line player B (a bluemage) comes and kills the same mob and it never did the TP move that they wanted to learn yet they still learned it because it was retained in memory

detailed in above


Xaver-DaRed (approved):

WinterSolstice8:
Does this reset timers or just a list of what was used?

KnowOne134:
Abilities stored that were used for bluemage learning

zach2good (approved):

WinterSolstice8 (approved):

PR #2815, opened 2022-09-27
Pathos effects changes (salvage and nyzul usage)

KnowOne134:
**_I affirm:_**

1. Adds a new effect flag that will only allow wearing of the debuff to only happen from zoning into a non instanced zone.
with this people can D/C in salvage and nyzul and the buff will never wear off, only by leaving the instance will it, since every instanced zone is connected to a non instanced zone.

2. Applies pathos a player may have on to all pets when called upon through core.
3. Removes all pathos from pets as they are removed from them either by cell usage-salvage or by floor progress-nyzul
4. Adds timing to pathos inflicted on entering salvage so the effect cant be removed from logging out. 
5. Adds powers to Omerta to differentiate between white magic, black magic, blue magic, ect - nyzul usage 
6. Adds powers to Impairment to differentiate between weaponskill and job ability usage - nyzule usage

!instance 7400
!changejob pup, smn, drg, or bst
!deleffect 262 to call a smn
!deleffect 261 for all other jobs

pup is easiest since you see all stats.
add cells and use and see that cells effect pets when applied to master
logout and in, dismiss pet and recall should all be how it was.
use fireflies to escape and when out of instance all effects were removed


WinterSolstice8:
could you add some documentation in comments near the omerta/impairment effect enum definitions with their bit values and meanings?

KnowOne134:
Here
https://github.com/LandSandBoat/server/blob/c8741fcd31d4121d4a838cc69978ac3d4732e89c/scripts/globals/nyzul.lua#L75

WinterSolstice8:
Ah, I didn't know it already existed elsewhere

KnowOne134:
Yea from previous work. Porting stuff from eden. This is to hook up missing pieces

scripts/globals/status.lua
```diff
@@ -893,6 +893,7 @@ xi.effectFlag =
     OFFLINE_TICK    = 0x2000000,
     AURA            = 0x4000000,
     HIDE_TIMER      = 0x8000000,
+    ON_ZONE_PATHOS  = 0x10000000,
```
WinterSolstice8, line 896: Might want to front pad the zeroes on the previous flags so they're aligned

src/map/status_effect.h
```diff
@@ -64,12 +64,13 @@ enum EFFECTFLAG
...
-    EFFECTFLAG_OFFLINE_TICK    = 0x2000000, // Duration elapses while offline
-    EFFECTFLAG_AURA            = 0x4000000, // Is an aura type effect
-    EFFECTFLAG_HIDE_TIMER      = 0x8000000, // Sends "Always" in the packet, even though timer is tracked
+    EFFECTFLAG_ON_JOBCHANGE    = 0x0400000,  // Removes effect when you change jobs
+    EFFECTFLAG_NO_CANCEL       = 0x0800000,  // CAN NOT CLICK IT OFF IN CLIENT
+    EFFECTFLAG_INFLUENCE       = 0x1000000,  // Influence effects - e.g. Signet, Sanction, Sigil, Ionis
+    EFFECTFLAG_OFFLINE_TICK    = 0x2000000,  // Duration elapses while offline
+    EFFECTFLAG_AURA            = 0x4000000,  // Is an aura type effect
+    EFFECTFLAG_HIDE_TIMER      = 0x8000000,  // Sends "Always" in the packet, even though timer is tracked
+    EFFECTFLAG_ON_ZONE_PATHOS  = 0x10000000, // removes the effect zoning into a non instanced zone
```
WinterSolstice8, line 73: Same here with the front padding on zeroes

WinterSolstice8 (changes requested):
Content wise looks good to me, but I think we want to pad those zeroes for alignment and maybe saving someone a headache down the line

scripts/globals/status.lua
zach2good: +1

WinterSolstice8 (approved):
LGTM

PR #2824, opened 2022-09-28
Battlefield rework

jmcmorris:
**_I affirm:_**

### Introduction
This is a rework of the battlefield system. The goal here is to simplify the creation of battlefield content. This new solution attempts to achieve this by reducing repeated code and unifying a battlefield details into a single file. This new system works side-by-side with the existing battlefield solution allowing for a gradual migration.

### How it works
How the new solution works is using the interaction system introducing `Battlefield` as a new `Container`. This `Battlefield` brings all common aspects of a battlefield into a single object. It contains common functionality such as battlefield entry npc interactions (see any `Burning_Circle.lua`) and the basic battlefield handlers found in `**\bcnms\*.lua` files. Data such as the required items/key items, menu options bit and unique area number are stored within it as well.

`Battlefield` can be extended upon such as `BattlefieldMission` which helps with common logic related to battlefields in missions.

All new battlefields are contained within `scripts/battlefields/*/*.lua`.

### Looking for Feedback
There are several major changes that are in place that I think we could have a discussion on. First off I will say that there is a good amount of semi-duplicated code in place. All of the prior solution is still in place but once the battlefields are converted over then a substantial amount of code will be cleaned up. This will almost certainly result in a major net negative lines of code which is always a win. Here are the major changes!

- Extended Interaction system to look for battlefields
- Functions such as `OnBattlefieldEnter` now try to find the function within the interaction system
- All battlefields that are created are added to `xi.battlefield.contents` and `xi.battlefield.contentsByZone`. These are used to lookup the battlefields in various places.
- Removes the need for `bcnm_battlefield.sql`
- Removes the need for `bcnm_treasure_chests.sql`
- Removes the need for many values within `bcnm_info.sql`
- Fixes several issues with battlefields along the way such as better messaging

### Known Issues
There are two known issues with this new battlefield solution.
1. When porting over Battlefields they need to be done for an entire zone otherwise events are being handled twice causing menus to pop up multiple times and the likes.
2. SE Apollyon has been ported over but not the other areas in the zone causing issues due to the first issue. I plan to address this soon as I'm going to start tackling the Limbus areas after this is merged in.

Go to any of the new battlefield scripts and there are instructions at the top on how to quickly get to that battlefield and enter. Missions will obviously require you to setup your character mission progress appropriately. Those requirements are detailed fully in the parameters passed to `BattlefieldMission:new()`.

## Post Full Conversion Work
Once all battlefields have been converted over there is a great deal of code (cruft) that can be cleaned up. Here's a fairly extensive list but not necessarily exhaustive.
- Remove m_RequiredEnemyList from Battlefield
- Remove Battlefield::SpawnLoot
- Remove bcnm_battlefield.sql
- Remove bcnm_treasure_chests.sql
- Remove globals/bcnm.lua
- Remove xi.battlefield.* lua functions
- Remove bcnm_info usage from CBattlefieldHandler::LoadBattlefield
- Remove CLuaBattlefield::loadMobs
- Remove Battlefield::getMobs
- Simplify Battlefield::rules (only using loseexp and allowsubjob)
- Remove BATTLEFIELDMOBCONDITION and just spawn when inserting if set to spawn
- Cleanup CLuaBaseEntity::registerBattlefield
- Remove ZoneType::BATTLEFIELD_OLD
- Remove CBattlefield::isInteraction()
- Remove xi.battlefield.HandleLootRolls
- Remove xi.battlefield.HandleWipe


scripts/globals/interaction/interaction_global.lua
```diff
@@ -40,7 +40,7 @@ function InteractionGlobal.loadContainers(shouldReloadRequires)
         package.loaded[interactionContainersPath] = nil
     end
 
-    local containerFiles = GetQuestAndMissionFilenamesList()
+    local containerFiles = GetContainerFilenamesList()
```
TeoTwawki, line 43: :+1:

claywar:
Just throwing this out there before it gets forgotten.  Since this is going to turn into a conversion effort (which I have absolutely no problem doing), could you also provide an example for a mission-based and regular BCNM in this PR?  Just want to provide the proper foundations and testing opportunities to ensure that followup events are being hit

scripts/globals/limbus.lua
```diff
@@ -530,141 +531,177 @@ function xi.limbus.openDoor(battlefield, doorID)
...
+    end
+end
 
-        -- Setup Paths
-        for mobId, path in pairs(self.paths) do
-            GetMobByID(mobId):pathThrough(path, xi.path.flag.PATROL)
-        end
+function Limbus:onEntryTrade(player, npc, trade)
+    if self:checkRequirements(player, npc, true) then
+        player:setCharVar("ApollyonEntrance", 1)
```
cocosolos: Just a note on this `ApollyonEntrance` var: 1 signifies that they entered one of the eastern zones and is used in the zones onEventFinish to set their pos to the eastern side when they exit since the cutscene always spits them out on the western side. There's likely some option that needs to be set for the cutscene instead but I couldn't figure it out at the time.

cocosolos:
> Just throwing this out there before it gets forgotten. Since this is going to turn into a conversion effort (which I have absolutely no problem doing), could you also provide an example for a mission-based and regular BCNM in this PR? Just want to provide the proper foundations and testing opportunities to ensure that followup events are being hit

Something to consider would be the battlefields that set the `phaseChange` local var on initialization so that win conditions don't get triggered when all mobs are "dead" between phases.

scripts/globals/limbus.lua
jmcmorris: Thanks! I noticed that after the player zones out they get shifted over. I'll be sure to look into that.

jmcmorris:
> > Just throwing this out there before it gets forgotten. Since this is going to turn into a conversion effort (which I have absolutely no problem doing), could you also provide an example for a mission-based and regular BCNM in this PR? Just want to provide the proper foundations and testing opportunities to ensure that followup events are being hit
> 
> Something to consider would be the battlefields that set the `phaseChange` local var on initialization so that win conditions don't get triggered when all mobs are "dead" between phases.

I think that'd be best to use the battlefield groups. The first group of mobs can have their own group and when all mobs die the then they can spawn the next set of mobs. I think the ideal scenario is to provide a simple way to setup a mob group that spawns the Armoury Crate and get rid of the default behavior of having the battlefield end when all victory condition mobs die. This will make phase changes much simpler.

cocosolos:
> I think the ideal scenario is to provide a simple way to setup a mob group that spawns the Armoury Crate and get rid of the default behavior of having the battlefield end when all victory condition mobs die. This will make phase changes much simpler.

Agreed. Core side of battlefields needs a lot of love unfortunately.

scripts/globals/limbus.lua
jmcmorris: Got that fixed! https://github.com/LandSandBoat/server/pull/2824/commits/67bcd5b2544c010e245834b72c375a014f693b6d

scripts/battlefields/Apollyon/se_apollyon.lua
```diff
@@ -123,7 +137,7 @@ area.groups =
             [xi.mod.HTH_SDT] = 0,
             [xi.mod.SLASH_SDT] = 1250,
         },
-        death = function(mob, count)
+        death = function(battlefield, mob, count)
```
claywar, line 157: General comment, and not a requirement, but would like to see a newline after any multiline-block table definition

scripts/battlefields/Bearclaw_Pinnacle/flames_for_the_dead.lua
```diff
@@ -0,0 +1,34 @@
...
+    exitNpc       = "Wind_Pillar_Exit",
+
+    missionArea           = xi.mission.log_id.COP,
+    mission               = xi.mission.id.cop.THREE_PATHS,
+    missionStatusArea     = xi.mission.log_id.COP,
+    missionStatus         = xi.mission.status.ULMIA,
+    requiredMissionStatus = 8,
+    title                 = xi.title.TRUE_COMPANION_OF_LOUVERANCE,
+    grantXP               = 1000,
+})
```
claywar, line 30: 👍

claywar: Are there any examples where missionArea and missionStatusArea are different?  This might be redundant

scripts/globals/battlefield.lua
```diff
@@ -68,10 +70,788 @@ xi.battlefield.dropChance =
...
+--  - hasWipeGrace: Grants players a 3 minute grace period on a full wipe before ejecting them. Defaults to true. (optional)
+--  - canLoseExp: Determines if a character loses experience points upon death while inside the battlefield. Defaults to true. (optional)
+--  - delayToExit: Amount of time to wait before exiting the battlefield. Defaults to 5 seconds. (optional)
+--  - requiredItems: Items required to be traded to enter the battlefield.
+--                   Needs to be in the format of { itemid, quantity, useMessage = ID.text.*, wearMessage = ID.text.*, wornMessage = ID.text.* }. (optional)
+--  - requiredKeyItems: Key items required to be able to enter the battlefield - these are removed upon entry (optional)
+--  - title: Title given to players upon victory (optional)
+--  - grantXP: Amount of XP to grant upon victory (optional)
+--  - lossEventParams: Parameters given to the loss event (32002). Defaults to none. (optional)
+function Battlefield:new(data)
```
claywar, line 110: There seems to be a lot of redundant info between the block above and below.  Could this be consolidated and formatted to make it easy to read?

scripts/globals/battlefield.lua
```diff
@@ -68,10 +70,788 @@ xi.battlefield.dropChance =
...
+    if setupEvents then
+        utils.append(zoneSection, {
+            onEventUpdate =
+            {
+                [32000] = Battlefield.redirectEventUpdate,
+                [32003] = utils.bind(Battlefield.redirectEventCall, "onExitEventUpdate"),
+            },
+            onEventFinish =
+            {
+                [32000] = utils.bind(Battlefield.redirectEventCall, "onEventFinishEnter"),
```
claywar, line 185: We might need to verify and/or move some of the hooks in quests and missions to use this code.  Also, need to verify that other BC's in the same area don't process (I don't think they will looking at code so far, but a definite test case)

scripts/globals/battlefield.lua
```diff
@@ -68,10 +70,788 @@ xi.battlefield.dropChance =
...
+    if utils.hasKey(self.zoneId, xi.battlefield.contentsByZone) then
+        table.insert(xi.battlefield.contentsByZone[self.zoneId], self)
+    else
+        xi.battlefield.contentsByZone[self.zoneId] = { self }
+    end
+    return self
+end
+
+function Battlefield:checkRequirements(player, npc, registrant, trade)
+    if (self.entryNpc ~= npc:getName()) then
```
claywar: Unnecessary outer parens

scripts/globals/battlefield.lua
```diff
@@ -68,10 +70,788 @@ xi.battlefield.dropChance =
...
+        if noEntryMessage then
+            player:messageSpecial(noEntryMessage)
+        end
+        return false
+    end
+
+    -- Ensure that the traded item(s) are not worn out
+    local contents = xi.battlefield.contentsByZone[zoneId]
+    for _, content in ipairs(contents) do
+        if #content.requiredItems > 0 and content.requiredItems[1].wornMessage and player:hasWornItem(content.requiredItems[1]) then
```
claywar: This could probably be easier to read if broken out into multiline conditional style:
```
if
    condition1 and
    not condition2
then

end
```

scripts/globals/battlefield.lua
```diff
@@ -68,10 +70,788 @@ xi.battlefield.dropChance =
...
+        player:startEvent(32000, 0, 0, 0, content.index, 0, 0, 0, 0)
+        return true
+    end
+
+    -- Player doesn't have battlefield status effect. That means player wants to register a new battlefield OR is attempting to enter a closed one.
+    -- Check if another party member has battlefield status effect. If so, that means the player is trying to enter a closed battlefield.
+    local alliance = player:getAlliance()
+    for _, member in pairs(alliance) do
+        if member:hasStatusEffect(xi.effect.BATTLEFIELD) then
+            -- player:messageSpecial() -- You are eligible but cannot enter.
```
claywar: Is this commented code intentional?

scripts/globals/battlefield.lua
```diff
@@ -68,10 +70,788 @@ xi.battlefield.dropChance =
...
+-- Static function to lookup the correct battlefield to handle this event update
+function Battlefield.redirectEventUpdate(player, csid, option, extras)
+    if option == 0 or option == 255 then
+        return false
+    end
+
+    local contents = xi.battlefield.contentsByZone[player:getZoneID()]
+    local value = bit.rshift(option, 4)
+    for _, content in pairs(contents) do
+        if (value == content.index) then
```
claywar: Unnecessary outer parens

scripts/globals/battlefield.lua
```diff
@@ -68,10 +70,788 @@ xi.battlefield.dropChance =
...
+    end
+
+    local ID = zones[self.zoneId]
+    player:messageSpecial(ID.text.ENTERING_THE_BATTLEFIELD_FOR, 0, self.index)
+    if self.maxPlayers > 6 then
+        player:messageSpecial(ID.text.MEMBERS_OF_YOUR_ALLIANCE, 0, 0, 0, self.maxPlayers)
+    elseif self.maxPlayers > 1 then
+        player:messageSpecial(ID.text.MEMBERS_OF_YOUR_PARTY, 0, 0, 0, self.maxPlayers)
+    else
+        -- TODO: Need capture of message to know the proper format
```
claywar: We should get an issue raised for this TODO to cover the proper case

scripts/globals/battlefield.lua
```diff
@@ -68,10 +70,788 @@ xi.battlefield.dropChance =
...
+-- Creates a new Limbus Battlefield interaction
+-- Data takes the additional following keys:
+--  - missionArea: The mission area this battlefield is associated with (optional)
+--  - mission: The mission this battlefield is associated with (optional)
+--  - missionStatusArea: The mission area to retrieve the mission status from. Will default to using the player's nation (optional)
+--  - missionStatus: The optional extra status information xi.mission.status (optional)
+--  - requiredMissionStatus: The required mission status to enter
+--  - skipMissionStatus: The required mission status to skip the cutscene. Defaults to the required mission status.
+--  - canLoseExp: Determines if a character loses experience points upon death while inside the battlefield. Defaults to false. (optional)
+function BattlefieldMission:new(data)
```
claywar, line 795: Similar to a previous example.  There's duplicated comments above and below; is there a way we can consolidate this info?

claywar (approved):
Seems reasonable, minor comments added

scripts/battlefields/Apollyon/se_apollyon.lua
jmcmorris: Can be easily done

scripts/battlefields/Bearclaw_Pinnacle/flames_for_the_dead.lua
jmcmorris: That's something I'm not 100% sure about. I believe they are different for the shared starter city missions such as The Shadow Lord Fight but perhaps I'm wrong. I haven't really worked on that code so I'm a little uncertain. If they are redundant then I can remove missionStatusArea.

scripts/globals/battlefield.lua
jmcmorris: Yeah, I was going to remove the now inlined comments but then you don't get variable tooltips. Any idea on how to consolidate that without losing the tooltips or perhaps the tooltips aren't deemed worthy?

scripts/globals/battlefield.lua
jmcmorris: None of these events are blocking and it will call into other hooks including interactions, zones, and npc files. Consequently other BCs in the same area that are not part of this new system will be using the npc file (such as Burning_Circle.lua) and it'll trigger things twice. This is noted in the "known issues". Easiest solution is to just convert over an entire zone of battlefields at a time. If this is not acceptable then we can come up with another solution.

scripts/globals/battlefield.lua
jmcmorris: Kind of? I meant to put a TODO on it. Basically there needs to be some work done on fixing that messaging. The old system was missing a lot of messages.

scripts/globals/battlefield.lua
jmcmorris: I'll ask my capture buddy to try this. I imagine it happens with Maat fights and other solo battlefields.

scripts/globals/battlefield.lua
jmcmorris: I'll address this once we figure out the primary one

scripts/globals/battlefield.lua
jmcmorris: Cannot find use case so I'm simplifying it for now. I think SE leaves a lot of text hanging around so it might not be used anymore or is used in a very specific use case. https://github.com/LandSandBoat/server/pull/2824/commits/e592f93bcf3dd42153c76583ae14d87cd1c2b556

src/map/battlefield_handler.cpp
```diff
@@ -73,94 +73,131 @@ void CBattlefieldHandler::HandleBattlefields(time_point tick)
...
+
+    auto ret = sql->Query(fmtQuery, registration.id);
+
+    if (ret == SQL_ERROR || sql->NumRows() == 0 || sql->NextRow() != SQL_SUCCESS)
+    {
+        ShowError("Cannot load battlefield : %u ", registration.id);
+        return BATTLEFIELD_RETURN_CODE_REQS_NOT_MET;
+    }
+
+    auto* name            = sql->GetData(0);
```
WinterSolstice8: you can use `sql->GetStringData(N)` to avoid casts here

src/map/battlefield_handler.cpp
```diff
@@ -73,94 +73,131 @@ void CBattlefieldHandler::HandleBattlefields(time_point tick)
...
 
-            if (lootid != 0)
-            {
-                PBattlefield->SetLocalVar("loot", lootid);
-            }
+        auto* PBattlefield = new CBattlefield(registration.id, m_PZone, registration.area, PChar, false);
 
-            luautils::OnBattlefieldInitialise(PBattlefield);
-            PBattlefield->InsertEntity(PChar, true);
+        auto* name                = sql->GetData(0);
```
WinterSolstice8: you can use `sql->GetStringData(N)` to avoid casts here

src/map/ai/states/death_state.cpp
```diff
@@ -32,7 +32,7 @@ CDeathState::CDeathState(CBattleEntity* PEntity, duration death_time)
 : CState(PEntity, PEntity->targid)
 , m_PEntity(PEntity)
 , m_deathTime(death_time)
-, m_raiseTime(GetEntryTime())
+, m_raiseTime(GetEntryTime() + 8s)
```
WinterSolstice8: magic number here probably warrants a comment

scripts/zones/Apollyon/IDs.lua
```diff
@@ -18,16 +18,22 @@ zones[xi.zone.APOLLYON] =
...
-        PARTY_WILL_BE_REMOVED_KICK    = 7358, -- All party members in Limbus have fallen in battle. Exiting now.
         THE_PARTY_WILL_BE_REMOVED     = 7366, -- All party members in Limbus have fallen in battle. Exiting in <number> [minute/minutes].
+        MEMBERS_OF_YOUR_PARTY         = 7332, -- Currently, <amount> members of your party (including yourself) have clearance to enter Limbus.
+        MEMBERS_OF_YOUR_ALLIANCE      = 7333, -- Currently, <amount> members of your alliance (including yourself) have clearance to enter Limbus.
+        TIME_LIMIT_FOR_THIS_BATTLE_IS = 7335, -- You may stay in Limbus for <amount> minutes.
+        PARTY_MEMBERS_HAVE_FALLEN     = 7358, -- All party members in Limbus have fallen in battle. Exiting now.
+        YOU_INSERT_THE_CARD_POLISHED  = 7371, -- You insert the <card> polished with <cosmo cleanse> into the slot!
         CHIP_TRADE                    = 7372, -- The light in the <item> has grown dim.
-        TIME_EXTENDED                 = 7374, -- our time in Limbus has been extended <number> [minute/minutes].
```
WinterSolstice8, line 28: Despite it looking wrong, update extractor actually needs this truncated
I believe the reason is that sometimes the first character of the message has bits for color and corrupts the character, so it's skipped. This means if there is an ID shift update extractor will not update `TIME_EXTENDED` and `TIME_LEFT` correctly

You can check these yourself by running the [update extractor ](https://github.com/zach2good/UpdateExtractor) yourself and looking at the output at `out\scripts\zones\Apollyon\Text.lua.raw.txt`

there might be a way for us to mask out the color bits so this isn't a problem in the future

scripts/zones/Apollyon/IDs.lua
```diff
@@ -18,16 +18,22 @@ zones[xi.zone.APOLLYON] =
         LOGIN_CAMPAIGN_UNDERWAY       = 7002, -- The [/January/February/March/April/May/June/July/August/September/October/November/December] <number> Login Campaign is currently underway!
         LOGIN_NUMBER                  = 7003, -- In celebration of your most recent login (login no. <number>), we have provided you with <number> points! You currently have a total of <number> points.
         MEMBERS_LEVELS_ARE_RESTRICTED = 7023, -- Your party is unable to participate because certain members' levels are restricted.
-        TIME_EXCEEDED                 = 7065, -- You have exceeded the time limit. Exiting Limbus now.
```
WinterSolstice8, line 21: Was this unused?

WinterSolstice8 (changes requested):

scripts/zones/Apollyon/IDs.lua
jmcmorris: This is replaced with `ID.text.TIME_IN_THE_BATTLEFIELD_IS_UP` which is shared between all battlefield zones.

scripts/zones/Apollyon/IDs.lua
jmcmorris: Ah ok, thanks for explaining that to me. I undid that "fix" I made so I don't break those.

WinterSolstice8 (approved):

jmcmorris:
Can we get this merged please? :)

PR #2827, opened 2022-09-29
Add xi.itemFlag enum

zach2good:
**_I affirm:_**
- [N/A] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

Add xi.itemFlag enum

N/A

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -3047,3 +3047,25 @@ xi.pathflag =
     SCRIPT   = 0x08, -- don't overwrite this path before completion (except via another script)
     SLIDE    = 0x10,  -- Slide to end point if close enough (so no over shoot)
 };
+
+-- Check Lua item with:
+-- local isEx = bit.band(item:getFlag(), xi.itemFlag.EX) ~= 0
+xi.itemFlag =
+{
+    WALLHANGING  = 0x0001,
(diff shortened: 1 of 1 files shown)


src/map/lua/lua_item.cpp
```diff
@@ -149,6 +149,16 @@ bool CLuaItem::isSubType(uint8 subtype)
     return m_PLuaItem->isSubType(static_cast<ITEM_SUBTYPE>(subtype));
 }
 
+bool CLuaItem::isRare()
+{
+    return m_PLuaItem->getFlag() & ITEM_FLAG_RARE;
```
TeoTwawki: suggestion: collapse these to one function and pass the enum or values in as input so one can check either flag or even multiple as mask bits.

See canUseMisc for an example of what I mean.

zach2good: +1 for this, will change when I have time

zach2good:
I don't think bindings for something as simple as checking flags is really needed anymore, provided we have more human readable enums:

```lua
local isEx = bit.band(item:getFlag(), xi.itemFlag.EX) ~= 0
```

WinterSolstice8 (approved):
+1 for built in tutorial

claywar (approved):

PR #2843, opened 2022-10-01
[module] Add custom module to make mazurka generate additional enmity

zach2good:
**_I affirm:_**

Makes Mazurka generate additional enmity, depending on how many party members it lands on.

This holds a special place in my heart because when I found out about this in early 2007 I sold a lot of my good gear to buy whatever the Mazurka+ instrument of the day was - at a very heavily inflated price. A few days later this behaviour was patched out.

Go be a sick BRD/NIN tank.

### DIFF
--- a/modules/custom/lua/mazurka_generates_enmity.lua
+++ b/modules/custom/lua/mazurka_generates_enmity.lua
@@ -1,0 +1,50 @@
+-----------------------------------
+-- Make Mazurka Provoke!
+--
+-- https://www.bg-wiki.com/ffxi/The_History_of_Final_Fantasy_XI/2007
+-- Mar. 8th, 2007: The enmity gained when using Chocobo Mazurka and Raptor Mazurka was reduced.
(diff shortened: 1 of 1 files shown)


claywar (approved):

modules/custom/lua/mazurka_generates_enmity.lua
```diff
@@ -0,0 +1,46 @@
...
+    super(caster, target, spell)
+
+    -- If not Mazurka, bail out
+    if spell:getID() ~= xi.magic.spell.CHOCOBO_MAZURKA and spell:getID() ~= xi.magic.spell.RAPTOR_MAZURKA then
+        return
+    end
+
+    -- Count party members in range
+    local party = caster:getParty()
+    local hitCount = 0
```
Xaver-DaRed: hitCount = spell:getTotalTargets()

That way you don't need distance check and loop below to count targets

zach2good:
Properly tested now

Xaver-DaRed (approved):

PR #2854, opened 2022-10-02
[core] Various improvements for xi_map, xi_connect, xi_search

WinterSolstice8:
**_I affirm:_**

See commit messages for details, but essentially I found that a certain type of crash was hanging Linux, I investigated and found that none of the executable had paths to cleanup themselves and exit properly.

This entire PR does the following for the 3 applications:

- Sockets now are non blocking
- Console input is now non blocking
- "exit" commands are moved into each application to allow decoupling from Kernel
- Sockets that weren't being closed on exit are now being closed
- Added use of "request exit" type booleans to properly exit normal-runtime infinite loops

Linux only:
This PR also reduces the stack traceback checking in backward from 32 to 10. 32 was taking literally 15 minutes on some crashes which seems unacceptable given 32 layers back usually is useless in the call stack.
Prevent recursive handing of exceptions causing an infinite loop printing stack traces

Windows only:
Changed input method to non blocking on console using some win32 api specific code which looks terrifyingly ugly.  I could not find a good and simple library to do this
Search.cpp now uses wepoll to implement epoll on Windows to check if a socket is ready to connect. map/login don't use this because they don't have issues with blocking, but once decoupled from kernel.cpp they could move over to that to be consistent.

- start all 3 applications, connect, send tells, search, login/logout, no issues
- run linux xi_map, login, `!exec player:messageText(nil, 1)`, see a stack trace print after a while and the application exit
- start all 3 applications, type "exit", have it work and not hang with no segfaults
- on Linux, start xi_map, let it start loading, hit control + C, which will cause a crash, have a stack trace and have the application exit
- Play around with the console on Windows and have it work well enough
- some other stuff I probably forgot

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.16)
-project(server CXX)
+project(server C CXX)
 
 set_property(GLOBAL PROPERTY CXX_STANDARD 17)
 set_property(GLOBAL PROPERTY CXX_STANDARD_REQUIRED ON)
--- a/ext/CMakeLists.txt
+++ b/ext/CMakeLists.txt
@@ -37,8 +37,15 @@ set(EXTERNAL_LIBS
     efsw
 )
 
+if (WIN32)
+    #add wepoll for epoll support on windows
+    add_subdirectory(wepoll)
+
+    list(APPEND EXTERNAL_LIBS
+        wepoll
+    )
 # backwards needs these to add to linker
-if (APPLE)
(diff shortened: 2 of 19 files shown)


WinterSolstice8:
Apparently, macos has no epoll so I had to add an #IFDEF for __APPLE__ in search.cpp. The code I implemented for that #IFDEF was tested on linux, it's POSIX standard so it should also work properly for MacOS but I can't exactly test it.

zach2good (approved):
Great work here 👍

PR #2862, opened 2022-10-02
add H_Quest Invitation_Cards

KnowOne134:
**_I affirm:_**

adds the hidden quest for turning in the lure of the wildcat invitation cards.
they can be turned in anytime and skipped, you can turn in as many as you want until all 4 cards are done then its marked as complete.
you only get bonus coins on 1st turn in, after that its a static reward

add 1-4 kis on player and trigger Naja until you get right prompt to show cards
yes/no option
yes: collect reward
no: zone and come back to re engage convo

### DIFF
--- a/scripts/quests/hiddenQuests/Invitation_Cards.lua
+++ b/scripts/quests/hiddenQuests/Invitation_Cards.lua
@@ -1,0 +1,134 @@
+-----------------------------------
+-- Lure of the Wildcat Invitation Cards
+
+-- Naja: !pos 22.700 -8.804 -45.591 50
+-----------------------------------
+require('scripts/globals/items')
+require('scripts/globals/keyitems')
+require('scripts/globals/npc_util')
+require('scripts/globals/zone')
+require('scripts/globals/interaction/hidden_quest')
+-----------------------------------
+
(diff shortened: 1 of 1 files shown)


KnowOne134:
not sure what to do about that error

scripts/quests/hiddenQuests/Invitation_Cards.lua
```diff
@@ -0,0 +1,122 @@
...
+            {
+                [74] = function(player, csid, option, npc)
+                    if option == 1 then
+                        local cards, coin, amount = quest.getCards(player)
+                        player:updateEvent({ [0] = cards, [1] = coin, [2] = amount, text_table = 0 })
+                    end
+                end,
+                [75] = function(player, csid, option, npc)
+                    if option == 1 then
+                        local cards, coin, amount = quest.getCards(player)
```
claywar: `local cards, _, _ = quest.getCards(player)`

claywar: If that doesn't work, just try assigning cards by itself since its the first returned value

KnowOne134:
Says variables aren't used but they are, tested and returned cs data is correct

claywar (approved):

PR #2869, opened 2022-10-03
[packets] Add setting main.EQUIP_FROM_OTHER_CONTAINERS

zach2good:
**_I affirm:_**

Flattens the logic inside SmallPacket0x050

Adds a setting to enable/disable to behaviour defined in: https://git.ashitaxi.com/Plugins/AshitaCast/commit/8fcdc3c8aa6ca09eb34b04d7002bb6bae950467f

Only tested the happy path so far, need to use addons to test the bad path

- Equip items, it still works
- Use addons to force-equip from non-standard places
- Use the new setting to enable/disable this
- Linkshells should still work also

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -3257,18 +3257,32 @@ void SmallPacket0x050(map_session_data_t* const PSession, CCharEntity* const PCh
     uint8 equipSlotID = data.ref<uint8>(0x05); // charequip slot
     uint8 containerID = data.ref<uint8>(0x06); // container id
 
-    if (containerID != LOC_INVENTORY && containerID != LOC_WARDROBE && containerID != LOC_WARDROBE2 && containerID != LOC_WARDROBE3 &&
-        containerID != LOC_WARDROBE4 && containerID != LOC_WARDROBE5 && containerID != LOC_WARDROBE6 && containerID != LOC_WARDROBE7 &&
-        containerID != LOC_WARDROBE8)
+    bool isAdditionalContainer =
+        settings::get<bool>("main.EQUIP_FROM_OTHER_CONTAINERS") &&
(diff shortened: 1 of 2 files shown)


claywar (approved):

src/map/packet_system.cpp
```diff
@@ -3257,18 +3257,32 @@ void SmallPacket0x050(map_session_data_t* const PSession, CCharEntity* const PCh
     uint8 equipSlotID = data.ref<uint8>(0x05); // charequip slot
     uint8 containerID = data.ref<uint8>(0x06); // container id
 
-    if (containerID != LOC_INVENTORY && containerID != LOC_WARDROBE && containerID != LOC_WARDROBE2 && containerID != LOC_WARDROBE3 &&
-        containerID != LOC_WARDROBE4 && containerID != LOC_WARDROBE5 && containerID != LOC_WARDROBE6 && containerID != LOC_WARDROBE7 &&
-        containerID != LOC_WARDROBE8)
+    bool isAdditionalContainer =
+        settings::get<bool>("main.EQUIP_FROM_OTHER_CONTAINERS") &&
```
TeoTwawki, line 3261: you goofed, the setting still should have been in isEquippableInventory

zach2good: Its equivalent, no? 

- If non of the regular containers
-If setting is enabled, and one of the other other containers

TeoTwawki: no, because now these are not additional containers unless setting is true.

zach2good: that was the intention: You shouldn't be able to equip directly from these containers (using addons) unless the setting is true (default is false)

TeoTwawki: o you aren't understanding this isn't about if you can equip from them lol

PR #2880, opened 2022-10-04
Implement Hundredfaced Hapool Ja (Mamook)

CatsEyeXI:
**_I affirm:_**

As it stands, on server init, the NM will be up along with his 4 'clones' by default. However
according to FFXIClopedia, and confirmed by me research in retail, the clones spawn only when NM uses Utsusemi.

engage mob, wait until it uses utsusemi, and clones should spawn around him, rather than being pre-spawned

### DIFF
--- a/scripts/zones/Mamook/mobs/Hundredfaced_Hapool_Ja.lua
+++ b/scripts/zones/Mamook/mobs/Hundredfaced_Hapool_Ja.lua
@@ -1,0 +1,61 @@
+-----------------------------------
+-- Area: Mamook
+--  Mob: Hundredfaced Hapool Ja
+-----------------------------------
+require("scripts/globals/spell_data")
+require("scripts/globals/titles")
+require("scripts/globals/status")
+mixins = { require("scripts/mixins/job_special") }
+-----------------------------------
+local ID = require("scripts/zones/Mamook/IDs")
+local entity = {}
+
+entity.onMobSpawn = function(mob)
+    local hundredfacedHapoolJa = mob:getID()
+    mob:addListener("MAGIC_USE", "SPAWN_CLONES", function(mobArg, target, spell, action)
+        local spellId = spell:getID()
+        local hateTarget = GetMobByID(hundredfacedHapoolJa):getTarget()
+
+        -- Utsusemi: Ichi (3 clones)
+        if spellId == xi.magic.spell.UTSUSEMI_ICHI then
+            for clone = hundredfacedHapoolJa + 1, hundredfacedHapoolJa + 3 do
+                if not GetMobByID(clone):isSpawned() then
+                    GetMobByID(clone):setSpawn(mob:getXPos()+math.random(1, 5), mob:getYPos(), mob:getZPos()+math.random(1, 5))
+                    SpawnMob(clone):updateEnmity(hateTarget)
+                end
+            end
+
+        -- Utsusemi: Ni/San (4 clones)
+        elseif spellId == xi.magic.spell.UTSUSEMI_NI or spellId == xi.magic.spell.UTSUSEMI_SAN then
+            for clone = hundredfacedHapoolJa + 1, hundredfacedHapoolJa + 4 do
+                if not GetMobByID(clone):isSpawned() then
+                    GetMobByID(clone):setSpawn(mob:getXPos()+math.random(1, 5), mob:getYPos(), mob:getZPos()+math.random(1, 5))
+                    SpawnMob(clone):updateEnmity(hateTarget)
+                end
+            end
+        end
+
+        for i = hundredfacedHapoolJa + 1, hundredfacedHapoolJa + 4 do
+            local pet = GetMobByID(i)
+            if pet:getCurrentAction() == xi.act.ROAMING then
+                pet:updateEnmity(target)
+            end
+        end
+    end)
+end
+
+entity.onMobEngaged = function(mob, target)
+    mob:castSpell(xi.magic.spell.UTSUSEMI_SAN, mob)
+end
+
+entity.onMobDeath = function(mob, player, isKiller)
+    local hundredfacedHapoolJa = mob:getID()
+    for i = 1, 4 do DespawnMob(hundredfacedHapoolJa + i) end
+end
+
+entity.onMobDespawn = function(mob)
+    local hundredfacedHapoolJa = mob:getID()
+    for i = 1, 4 do DespawnMob(hundredfacedHapoolJa + i) end
+end
+
+return entity
--- a/sql/mob_groups.sql
+++ b/sql/mob_groups.sql
@@ -3185,6 +3185,7 @@ INSERT INTO `mob_groups` VALUES (68,6754,65,'Venomfang',0,128,0,0,0,78,78,0);
 INSERT INTO `mob_groups` VALUES (69,6755,65,'Yalungur',0,128,0,0,0,99,99,0);
 INSERT INTO `mob_groups` VALUES (70,6756,65,'Predatory_Colibri',0,128,0,0,0,99,99,0);
 INSERT INTO `mob_groups` VALUES (71,3807,65,'Suhur_Mas',960,0,2359,0,0,68,71,0);
+INSERT INTO `mob_groups` VALUES (72,7056,65,'Hundredfaced_clone',0,128,0,0,0,75,90,0);
 
 -- ------------------------------------------------------------
 -- Mamool_Ja_Training_Grounds (Zone 66)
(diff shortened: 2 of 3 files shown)


scripts/zones/Mamook/Zone.lua
```diff
@@ -8,6 +8,9 @@ local zone_object = {}
 
 zone_object.onInitialize = function(zone)
     xi.helm.initZone(zone, xi.helm.type.LOGGING)
+	-- Clones should not automatically spawn on server init
+	-- TODO: There's probably a better way to handle this in the db
+    for i = 1, 4 do DespawnMob(ID.mob.HUNDREDFACED_HAPOOL_JA + i) end
```
zach2good: You can change the spawn flags (I think in the groups or spawn points) to have them not spawn by default

scripts/zones/Mamook/mobs/Hundredfaced_Hapool_Ja.lua
```diff
@@ -0,0 +1,39 @@
...
+mixins = {require("scripts/mixins/job_special")}
+-----------------------------------
+local entity = {}
+
+entity.onMobFight = function(mob, target)
+    mob:addListener("MAGIC_USE", "SPAWN_CLONES", function(mobArg, target, spell, action)
+    local spellId = spell:getID()
+        -- Utsusemi: Ichi (3 clones)
+        if spellId == xi.magic.spell.UTSUSEMI_ICHI then
+            for i = ID.mob.HUNDREDFACED_HAPOOL_JA + 1, ID.mob.HUNDREDFACED_HAPOOL_JA + 3 do
```
zach2good: You don't need to use this id here, you can get it with mob:getID and then remove the new entry from the zone lua. Then it'll be shift proof

scripts/zones/Mamook/Zone.lua
TeoTwawki: spawnType, set them to scripted spawns.

scripts/zones/Mamook/mobs/Hundredfaced_Hapool_Ja.lua
```diff
@@ -0,0 +1,39 @@
...
+require("scripts/globals/spell_data")
+require("scripts/globals/titles")
+require("scripts/globals/status")
+local ID = require("scripts/zones/Mamook/IDs")
+mixins = {require("scripts/mixins/job_special")}
+-----------------------------------
+local entity = {}
+
+entity.onMobFight = function(mob, target)
+    mob:addListener("MAGIC_USE", "SPAWN_CLONES", function(mobArg, target, spell, action)
```
TeoTwawki, line 15: the listener should be added at spawn time instead of on every tick of onMobFight

scripts/zones/Mamook/mobs/Hundredfaced_Hapool_Ja.lua
TeoTwawki:
```lua
local spellId = spell:getID()
local hundredfaced_hapool_ja = mob:getID()
-- Utsusemi: Ichi (3 clones)
if spellId == xi.magic.spell.UTSUSEMI_ICHI then
    for i = hundredfaced_hapool_ja + 1, hundredfaced_hapool_ja + 3 do
```
so that its only called once while being shift-proof per Zach's suggestion

CatsEyeXI: Thanks for all the feedback. The only other issue I can't seem to move past, is making the clones update enmity to the target. (Seems like target is being lost on the new mob spawn) But I'll push a commit a bit later with the suggestions proposed and we can go from there.

TeoTwawki: we'd need to get target from the main NM rather than relying on input the function arg, because in the context of this cast the target is itself (its where the spell is being cast onto)

`local hateTarget = GetMobByID(hundredfaced_hapool_ja):getTarget()`

`SpawnMob(i):updateEnmity(hateTarget)`

CatsEyeXI:
Everything relating to the mob's actions appears to be working properly now. 
My issue is converting this to a scripted spawn now. I've updated the spawntype value to 128 - "Scripted", and added the logic to onMobDespawn... but now the mob won't pop unless done manually... what am i missing?

CatsEyeXI:
everything is working in this current state, however i did find the need for the static ID entry. 
if this is not correct or if any additional changes required let me know. thanks!

https://media.tenor.com/re_Txs4OUToAAAAd/tab-tab-key.gif

scripts/zones/Mamook/mobs/Hundredfaced_Hapool_Ja.lua
zach2good: I did some similar "target-finding" here: <https://github.com/LandSandBoat/server/blob/base/modules/custom/lua/mazurka_generates_enmity.lua#L24-L29>

zach2good:
We're trying to move away from hard-coded IDs, since shifts are a major pain for us, downstream, and players.
You can get the mob-object for a mob by asking the zone for it, you can see me doing it here for FSA:
https://github.com/LandSandBoat/server/blob/base/scripts/zones/Batallia_Downs/Zone.lua#L32-L44

and here in a module:
<https://github.com/LandSandBoat/server/blob/base/modules/custom/lua/conquest_regional_npcs_always_up.lua>

It might look something like this:

```lua
    xi.helm.initZone(zone, xi.helm.type.LOGGING)

    local hundredFacedHapoolJaID = zone:queryEntitiesByName("Hundredfaced_Hapool_Ja")[1]:getID()
    UpdateNMSpawnPoint(hundredFacedHapoolJaID)
    SpawnMob(hundredFacedHapoolJaID)
end
```

This is also the idea behind `DYNAMIC_LOOKUP`. This works especially well for NMs because they tend to have unique names in a zone that we don't have to disambiguate for any reason.

scripts/zones/Mamook/mobs/Hundredfaced_Hapool_Ja.lua
zach2good: Teo fed you a very quick example here, but keep an eye on styling: <https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md#lua-misc--naming>

TeoTwawki:
> Everything relating to the mob's actions appears to be working properly now. My issue is converting this to a scripted spawn now. I've updated the spawntype value to 128 - "Scripted", and added the logic to onMobDespawn... but now the mob won't pop unless done manually... what am i missing?

His helpers spawned by his shadows would be type scripted, he himself would not be since he's a timed pop. If initial pop is randomized (like HNM timers tend to be for example) we'd init his time in the zone lua and then re-random every time his corpse despawns. The clones aren't the same ID and shouldn't have the same group so that they can be different spawn type and diff loot pool than the main NM.

zach2good:
Once this PR lands: https://github.com/LandSandBoat/server/pull/2882
it'll be safe for you to just use the table entry:`         HUNDREDFACED_HAPOOL_JA   = DYNAMIC_LOOKUP,`

CatsEyeXI:
Oooh. Very fancy. I'll give it a shot.

CatsEyeXI:
> > Everything relating to the mob's actions appears to be working properly now. My issue is converting this to a scripted spawn now. I've updated the spawntype value to 128 - "Scripted", and added the logic to onMobDespawn... but now the mob won't pop unless done manually... what am i missing?
> 
> His helpers spawned by his shadows would be type scripted, he himself would not be since he's a timed pop. If initial pop is randomized (like HNM timers tend to be for example) we'd init his time in the zone lua and then re-random every time his corpse despawns. The clones aren't the same ID and shouldn't have the same group so that they can be different spawn type and diff loot pool than the main NM.

I honestly don't know if his initial spawn is randomized, I guess that would need retail verification at some point? Additionally, i couldn't find any information anywhere regarding his windows, so i used the value of 72 hrs already in the DB. If it's ok with you, I think I'll set him to pop on server init, then set with his corpse despawns.

CatsEyeXI:
everything tested working with dynamic lookup, anything need to be done regarding these debug messages?
i believe these only print when there is no entry present in nm_spawn_points... 

image

scripts/zones/Mamook/mobs/Hundredfaced_Hapool_Ja.lua
```diff
@@ -0,0 +1,64 @@
...
+    local hundredfaced_hapool_ja = mob:getID()
+    for i = 1, 4 do DespawnMob(hundredfaced_hapool_ja + i) end
+end
+
+entity.onMobDespawn = function(mob)
+    local hundredfaced_hapool_ja = mob:getID()
+    for i = 1, 4 do DespawnMob(hundredfaced_hapool_ja + i) end
+    -- TODO: 72 hours taken from DB, need retail verification.
+    UpdateNMSpawnPoint(hundredfaced_hapool_ja)
+    mob:setRespawnTime(259200)
```
cocosolos: Since there's nothing special about his spawn mechanics, you can just rely on the respawn time set in mob_groups and remove this and the changes to zone.lua and IDs.lua. Since it's a normal spawntype and has a respawn time set in the DB, all of this is handled automatically.

zach2good:
We have a new CI rule, you'll need to rename `hundredfaced_hapool_ja` => `hundredfacedHapoolJa`, or else CI will start to fail when this eventually merges

https://github.com/LandSandBoat/server/pull/2894

Xaver-DaRed (approved):

PR #2884, opened 2022-10-05
change load order setting gets checked in

TeoTwawki:
**_I affirm:_**

I know its not being -used- her at the moment, but the additional containers still -are- additional containers, regardless of the settings state. So this just re-orders things to reflect that. Otherwise, when the setting is false, the entire bool there was also false. This didn't impact ability to equip. Nitpicky of me maybe.

print true/false status of each bool before and after.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -3258,10 +3258,9 @@ void SmallPacket0x050(map_session_data_t* const PSession, CCharEntity* const PCh
     uint8 containerID = data.ref<uint8>(0x06); // container id
 
     bool isAdditionalContainer =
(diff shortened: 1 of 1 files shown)


zach2good (approved):

zach2good:
Sanity checks were skipped here

TeoTwawki:
> Sanity checks were skipped here

sanity check it tripped was bad, extra space under a parenthesis line for no good reason causing it to mis-align with the rest of the same block, so I chose to ignore it. Teo was a naughty boy kids, don't be like Teo: https://github.com/LandSandBoat/server/pull/2886#issuecomment-1268558259

PR #2867, opened 2022-10-03
Add mobMod ENCROACH_TARGET for NMs like Claret and Dyinyinga

TiberonKalkaz:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

This PR adds a new mob mod, ENCROACH_TARGET, which when enabled on a mob allows the mob to "close in" on its target closer than is normally allowed.  This functionality is used by NMs like Claret and Dyinyinga where they apply a status to the player if they are able to get close enough.  These NMs attempt to make model to model contact with players.

## Steps to test these changes

1. Get to Claret  
> !zone {Mount Zhayolm}   
> !pos 501 -9 53)
2. Turn off GodMode if on.  Enable !Immortal if not on
3. Get a pectin
>!additem 2591 
4. !hp 0 the nearby clot or toggle GM
5. Trade the pectin to the ??? to spawn Claret
6. Note that Claret will get very close to the player, much closer than normal mobs, and a poison is applied.
7. Move away from Claret slowly, note that Claret continues to close in tightly as soon as the player is a little bit away
8. Run from Claret (with speed boost/flee), note that when Claret finally catches up, it encroaches on the player
9. Aggro/pull a clot nearby, note that the clot does not encroach the player even though Claret still does.

PR #2888, opened 2022-10-05
Add mobMod ENCROACH_TARGET for NMs like Claret and Dyinyinga

TiberonKalkaz:
**_I affirm:_**

This PR adds a new mob mod, ENCROACH_TARGET, which when enabled on a mob allows the mob to "close in" on its target closer than is normally allowed.  This functionality is used by NMs like Claret and Dyinyinga where they apply a status to the player if they are able to get close enough.  These NMs attempt to make model to model contact with players.

1. Get to Claret  
> !zone {Mount Zhayolm}   
> !pos 501 -9 53)
2. Turn off GodMode if on.  Enable !Immortal if not on
3. Get a pectin
>!additem 2591 
4. !hp 0 the nearby clot or toggle GM
5. Trade the pectin to the ??? to spawn Claret
6. Note that Claret will get very close to the player, much closer than normal mobs, and a poison is applied.
7. Move away from Claret slowly, note that Claret continues to close in tightly as soon as the player is a little bit away
8. Run from Claret (with speed boost/flee), note that when Claret finally catches up, it encroaches on the player
9. Aggro/pull a clot nearby, note that the clot does not encroach the player even though Claret still does.

Note: This replaces the [closed PR ](https://github.com/LandSandBoat/server/pull/2867)

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -620,8 +620,9 @@ void CMobController::Move()
         }
     }
 
-    bool  move         = PMob->PAI->PathFind->IsFollowingPath();
-    float attack_range = PMob->GetMeleeRange();
+    bool  move            = PMob->PAI->PathFind->IsFollowingPath();
+    float attack_range    = PMob->GetMeleeRange();
+    float closureDistance = attack_range - 0.2f - (static_cast<float>(PMob->getMobMod(MOBMOD_ENCROACH_TARGET)) / 10.0);
 
     if (PMob->getMobMod(MOBMOD_ATTACK_SKILL_LIST) > 0)
     {
@@ -643,7 +644,7 @@ void CMobController::Move()
         CMobEntity* posShare = (CMobEntity*)PMob->GetEntity(PMob->getMobMod(MOBMOD_SHARE_POS) + PMob->targid, TYPE_MOB);
         PMob->loc            = posShare->loc;
     }
-    else if (((currentDistance > attack_range - 0.2f) || move) && PMob->PAI->CanFollowPath())
+    else if (((currentDistance > closureDistance) || move) && PMob->PAI->CanFollowPath())
     {
         //#TODO: can this be moved to scripts entirely?
         if (PMob->getMobMod(MOBMOD_DRAW_IN) > 0)
@@ -672,7 +673,7 @@ void CMobController::Move()
                 if (!PMob->PAI->PathFind->IsFollowingPath() || distanceSquared(PMob->PAI->PathFind->GetDestination(), PTarget->loc.p) > 10)
                 {
                     // path to the target if we don't have a path already
-                    PMob->PAI->PathFind->PathInRange(PTarget->loc.p, attack_range - 0.2f, PATHFLAG_WALLHACK | PATHFLAG_RUN);
+                    PMob->PAI->PathFind->PathInRange(PTarget->loc.p, closureDistance, PATHFLAG_WALLHACK | PATHFLAG_RUN);
                 }
                 PMob->PAI->PathFind->FollowPath(m_Tick);
                 if (!PMob->PAI->PathFind->IsFollowingPath())
--- a/src/map/mob_modifier.h
+++ b/src/map/mob_modifier.h
@@ -100,6 +100,7 @@ enum MOBMODIFIER : int
     MOBMOD_NO_LINK            = 69, // If set, mob cannot link until unset.
     MOBMOD_NO_REST            = 70, // Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
     MOBMOD_LEADER             = 71, // Used for mobs that follow a defined "leader", such as Ul'xzomit mobs.
+    MOBMOD_ENCROACH_TARGET    = 72, // How close a mob will encroach on it's target, attempting to make model to model contact. Encroach distance * 10
 };
 
 #endif
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -2496,6 +2496,7 @@ xi.mobMod =
     NO_LINK             = 69, -- If set, mob cannot link until unset.
     NO_REST             = 70, -- Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
     LEADER              = 71, -- Used for mobs that follow a defined "leader", such as Ul'xzomit mobs.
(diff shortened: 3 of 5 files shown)


src/map/ai/controllers/mob_controller.cpp
```diff
@@ -620,8 +620,9 @@ void CMobController::Move()
         }
     }
 
-    bool  move         = PMob->PAI->PathFind->IsFollowingPath();
-    float attack_range = PMob->GetMeleeRange();
+    bool  move            = PMob->PAI->PathFind->IsFollowingPath();
+    float attack_range    = PMob->GetMeleeRange();
+    float closureDistance = PMob->getMobMod(MOBMOD_ENCROACH_TARGET) ? attack_range - 3.7f : attack_range - 0.2f;
```
TeoTwawki: why not use the value of the mod so that this mod can be reused at variable distances? And call it `MOBMOD_CLOSE_DISTANCE` 

if zero, use the default, else take mobmod and convert to float (that 0.2f is in yalms btw)

I'd say just modify clarets attack range at the db level, but that would adjust attack range players can hit back as well.

TiberonKalkaz: Two reasons I can think of.
1. mobMods are integer values, and via testing on wings 3.5 was the ideal reduction for Claret and Dyinga (the only wings-era mobs who used the value)
2. The ideal way to make this variable would be to consider mob size (model size) into the equation, as larger mobs would need less of a reduction - unless there is a need for mob-specific-configuration

Are there other mobs which encroach the player in retail era which can make use of a configurable value?  If so I can test with those.

TeoTwawki: > mobMods are integer values

conversion is a trivial casting (if it wasn't clear, you would set mod to `37`, not `3.7` then divide after)

> The ideal way to make this variable would be to consider mob size (model size) into the equation 

we already do this in the attack range, which is based on the family in the mobs family currently tho I would like to move this to someplace that can be more flexible (and check the bit flags for alternate sizes on the same model, as abysea mobs do to spawn in fixed sizes like most others)

```cpp
    int myInterger = 37;
    float myFloat = static_cast<float>(myInterger) / 10.0f;
    cout<<myFloat;

    return 0;
```
output: `3.7`

https://www.onlinegdb.com/online_c++_compiler

TiberonKalkaz: True, conversion is cheap.
I can add the configurablility - I assume somone will make use of it someday, hopefully not a YAGNI

On the topic of mob sizes - one of the biggest pains for this on wings was blue magic - havent looked into that on LSB yet.  I created a custom hack which would consider mob size into the distance calc, otherwise blu's had to clip into larger mobs to be in range.

TeoTwawki: really the hold up for us is that the size bits don't have universal behavior among all entities that have that mask so we already know we're going to refactor the mask the size bits sit in, so anything we do there would wind up getting redone :disappointed: 

as far as YAGNI, even if we never find another use case, I think it's a good method for the use here while opening up future use even if future use turns out to be some random server's custom :+1:

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -620,8 +620,9 @@ void CMobController::Move()
         }
     }
 
-    bool  move         = PMob->PAI->PathFind->IsFollowingPath();
-    float attack_range = PMob->GetMeleeRange();
+    bool  move            = PMob->PAI->PathFind->IsFollowingPath();
+    float attack_range    = PMob->GetMeleeRange();
+    float closureDistance = attack_range - 0.2f - (static_cast<float>(PMob->getMobMod(MOBMOD_ENCROACH_TARGET)) / 10.0);
```
TeoTwawki, line 625: I think it's less intuitive when the `-0.2` stacks with the mod, which as why I suggested:
> if zero, use the default, else take mobmod and convert to float

and I believe you want that `10.0` to be `10.0f` but the compiler might have automatic figured out your intent anyway.

src/map/mob_modifier.h
```diff
@@ -100,6 +100,7 @@ enum MOBMODIFIER : int
     MOBMOD_NO_LINK            = 69, // If set, mob cannot link until unset.
     MOBMOD_NO_REST            = 70, // Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
     MOBMOD_LEADER             = 71, // Used for mobs that follow a defined "leader", such as Ul'xzomit mobs.
+    MOBMOD_ENCROACH_TARGET    = 72, // How close a mob will encroach on it's target, attempting to make model to model contact. Encroach distance * 10
```
TeoTwawki, line 103: I don't think "encroach" is a good name for this.

I'm open to suggestions other than my previous "close-distance" / other opinions

zach2good: https://www.thesaurus.com/browse/encroach

I'm in for MOVE_CLOSE, MOVE_TO_MODEL, etc. more descriptive words vs a more uncommon word that _technically_ describes whats going on

TeoTwawki: `MOBMOD_CLOSE_TO_TARGET`    ?

zach2good: I might err on the side of `MOBMOD_MOVE_CLOSE_TO_TARGET` so it sounds more like an instruction, rather than a question

TeoTwawki: > I might err on the side of `MOBMOD_MOVE_CLOSE_TO_TARGET` so it sounds more like an instruction, rather than a question

kinda long.

but also brings me back to including "distance" in the phrasing.

zach2good: I'd rather have a longer name where its extremely clear what it does, rather than a shorter name where there's any doubt

TeoTwawki: `TARGET_DISTANCE_OFFSET` was suggested on discord chat. I think that's clear.

UmeboshiXI: Bumping this. Was there ever a mod name decided upon so that this could be implemented if that was the reason it was not merged?

TeoTwawki: `TARGET_DISTANCE_OFFSET` since nobody objected to that and the other comment needed addressed https://github.com/LandSandBoat/server/pull/2888#discussion_r988431034

The PR appears to have been abandoned so someone will need to adopt it and preserve the originators authorship for the work they already did.

TeoTwawki:
Continued in #3751 where I have resolved conflicts

PR #2896, opened 2022-10-06
FIX: DMG mods were always applied to mainhand

DiscipleOfEris:
**_I affirm:_**

Before, DMG+X mods (e.g. augments) or DMG:Y latents (see Blau Dolch or Erlking's Blade) would only apply their mods to the weapon current equipped in the main hand, even if they were equipped in the offhand. That is, you could equip an item with a DMG+10 augment in the offhand, and it would apply the DMG+10 to the mainhand instead. This is now fixed.

A new modifier `DMG_RATING` has been added. This will apply the DMG+X modifier to the current weapon rather than to a specific slot like the `MAIN_DMG_RATING` modifiers and its friends.

An `EQUIPPED_IN_SLOT` latent is added, along with implementing this latent for all the Erlking's Blade and similar items (which have Mainhand: DMG:X latents).

The Coffinmaker item had a latent of Crit Rate+6% which is not listed in the wiki, this was removed.

The Sirius Axe was adding the `DMG` modifier, which is actually a `Damage Taken` modifier. I'll submit a rename PR in a bit. Anyway, this has been removed.

Furthermore, all the augments and latents which were using a `MAIN_DMG_RATING` or similar modifier are now using the new `DMG_RATING` modifier.

### `DMG_RATING` changes.

1. Give yourself an item augmented with DMG+X.
2. Equip it in offhand.
3. Note that the DMG+X mod is NO LONGER being applied to the mainhand. It is correctly applied to the offhand.

### `EQUIPPED_IN_SLOT` latent.

1. Give yourself an Erlking's Blade
2. Equip in offhand. Note damage.
3. Equip in mainhand. Note the damage is higher.

Some other items that were previously using `MAIN_DMG_RATING` already had some code handling to make them apply to offhand. These have all been adjusted to use `DMG_RATING` and that superfluous code was removed.

## Notes

All latents are currently being applied *after* equipment scaling, when according to the wiki they should be applied *before* equipment scaling. This should be addressed.

Some research on how augments are treated with equipment scaling should be done. The wiki says all augments are removed with equipment scaling, but some people have suggested the wiki is wrong on this.

The Destroyers and other similar weapons that increase damage when unlocked (with WS points) should have their latents modified. The wiki says that items with DMG+X modifiers do not change their damage rank, despite their weapon damage changing. However, the wiki says the weapon rank of the Destroyers and similar items is always at the unlocked rank, suggesting they should have a latent of "*Not* unlocked": "DMG-X".

The `MAIN_DMG_RATING` and similar modifiers are completely unnecessary now. Only reason to keep them would be for other private servers to apply a non-retail modifier where, for example, an earring gives a DMG+X modifier or something.

Finally, a future PR should rename the `DMG` modifier to make it more obvious that this is a `Reduce all DMG by X %` modifier.

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -1790,7 +1790,7 @@ xi.latent =
     MOON_PHASE               = 37, -- PARAM: 0: New Moon, 1: Waxing Crescent, 2: First Quarter, 3: Waxing Gibbous, 4: Full Moon, 5: Waning Gibbous, 6: Last Quarter, 7: Waning Crescent
     JOB_MULTIPLE             = 38, -- PARAM: 0: ODD, 2: EVEN, 3-99: DIVISOR
     JOB_MULTIPLE_AT_NIGHT    = 39, -- PARAM: 0: ODD, 2: EVEN, 3-99: DIVISOR
-    -- 40 free to use
+    EQUIPPED_IN_SLOT         = 40, -- When item is equipped in the specified slot (e.g. Dweomer Knife, Erlking's Sword, etc.) PARAM: slotID
     -- 41 free to use
     -- 42 free to use
     WEAPON_DRAWN_HP_UNDER    = 43, -- PARAM: HP PERCENT
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -1333,8 +1333,15 @@ xi.mod =
     HIGH_JUMP_ENMITY_REDUCTION      = 363,
     REWARD_HP_BONUS                 = 364,
     SNAP_SHOT                       = 365,
-    MAIN_DMG_RATING                 = 366,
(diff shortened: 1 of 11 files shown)


TeoTwawki:
needs update/rebase, but otherwise looks good to me.

WinterSolstice8:
> However, the wiki says the weapon rank of the Destroyers and similar items is always at the unlocked rank, suggesting they should have a latent of "Not unlocked": "DMG-X".

This seems like a good way to fix wrank for those style weapons, the get damage functions will return their "real" dmg for rank calculation purposes 👍

WinterSolstice8 (approved):

DiscipleOfEris:
> needs update/rebase, but otherwise looks good to me.

Should be good now.

TeoTwawki (approved):

PR #2902, opened 2022-10-07
[chore] Utilize dynamic lookup for Mobs and NPCs with same naming convention, apply styling

claywar:
**_I affirm:_**

* Converts a large chunk of NPCs and Mobs to use `DYNAMIC_LOOKUP`
* Applys consistent styling to all IDs.lua files

* Run DB query for each name in IDs.lua that was converted
* See no startup errors when looking up those IDs


claywar:
Looks like we're going to need some handling in modules as well, will revisit this once a pattern has been established

PR #2907, opened 2022-10-08
Prevent crystals, seals, and geodes from dropping in battlefields and Dynamis

jmcmorris:
**_I affirm:_**

I looked and all zone ids we were excluding were Dynamis zones. This method simply checks the type and excludes Dynamis and now battlefields as they both do not yield crystals, seals, geodes and the likes.

Go into any battlefield or Dynamis and kill mobs. No common items should drop.

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -48,6 +48,7 @@
 #include "../utils/itemutils.h"
 #include "../utils/mobutils.h"
 #include "../utils/petutils.h"
+#include "../utils/zoneutils.h"
 #include "../weapon_skill.h"
 #include "common/timer.h"
 #include "common/utils.h"
@@ -1006,9 +1007,8 @@ void CMobEntity::DropItems(CCharEntity* PChar)
         }
     }
 
-    uint16 Pzone = PChar->getZone();
-
-    bool validZone = ((Pzone > 0 && Pzone < 39) || (Pzone > 42 && Pzone < 134) || (Pzone > 135 && Pzone < 185) || (Pzone > 188 && Pzone < 255));
+    ZONE_TYPE zoneType  = zoneutils::GetZone(PChar->getZone())->GetType();
+    bool      validZone = zoneType != ZONE_TYPE::BATTLEFIELD && zoneType != ZONE_TYPE::DYNAMIS;
 
     if (!getMobMod(MOBMOD_NO_DROPS) && validZone && charutils::CheckMob(m_HiPCLvl, GetMLevel()) > EMobDifficulty::TooWeak)
     {


claywar (approved):

zach2good (approved):

cocosolos:
Should this check Limbus zonetype as well? Also Ghelsba Outpost, Riverne A01, and Riverne B01 have battlefields but aren't battlefield zonetype. I'm not sure if some of this is handled elsewhere e.g. with mobmods.

Also maybe worth noting that Riverne B01 is included in battlefield.lua (L14) as having 1 battlefield area but Riverne A01 is not (probably because Ouryu isn't implemented yet).

TeoTwawki:
> Should this check Limbus zonetype as well? Also Ghelsba Outpost, Riverne A01, and Riverne B01 have battlefields but aren't battlefield zonetype. I'm not sure if some of this is handled elsewhere e.g. with mobmods.
> 
> Also maybe worth noting that Riverne B01 is included in battlefield.lua (L14) as having 1 battlefield area but Riverne A01 is not (probably because Ouryu isn't implemented yet).

there was a `MOBMOD_NO_DROPS` to handle this very thing, but it may have been forgotten when newer things were added
https://github.com/LandSandBoat/server/search?q=mobmod_no_drops

(we can't mark the whole zone because non battlefield mobs would then not drop)

TeoTwawki:
(we can't mark the whole zone because non battlefield mobs would then no drop)

cocosolos:
I don't think that would work for Limbus anyway because they can direct drop beastcoins, though this would probably work for the other 3 zones I mentioned and may actually already be being set. I think I recall KnowOne making some changes to at least the Ghelsba mobs but I'm not sure if no drops was a part of that and don't see anything in the mob scripts.

TeoTwawki:
was referring to that mention of Riverne having battlefield (isn't it just bahamhut? the rest in monarch lin yeah?), sorry I wasn't clear. I theory we could move the coins script side to accommodate that but I don't think it'd be optimal and I'd rather not.

cocosolos:
Limbus is easy because it has it's own zonetype like Dynamis. Riverne A01 has Ouryu, I completely forgot about this until recently as well.

cocosolos:
Sorry I should specify, I mean the Cloud Evoker version of the fight, Ouryu Cometh. The mission with with Ouryu is in Monarch Linn.

TeoTwawki:
I had forgotten they weren't in same place, so its fine

jmcmorris:
I could add Limbus to this but that zone type is going to be phased out very soon and reverted back a battlefield. This will happen once I get the Limbus zones reworked and brought over into the new battlefield system. I hope to have that done within the next few weeks.

Issue #886, opened 2021-10-27
[Limbus] logged out players do not get kicked out on logging back in after the battle end


TeoTwawki:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: All

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
1. enter any Limbus battlefield
2. log out.
3. log back in long after your "bcnm" has ended
4. be amazed as you are trapped forever in a bcnm with no mobs and no exit till a GM moves you out.

zach2good:
This might also be an issue for people who DC in instances

PR #2910, opened 2022-10-08
Battlefield Interactions: Reconnect, kick, and reentry

jmcmorris:
**_I affirm:_**

Adds proper handling with battlefield interactions

1. When a player disconnects inside a battlefield and logs with the battlefield having been destroyed then they should be kicked from the battlefield.
2. When a player disconnects inside a battlefield but other party members are in the battlefield then they should rejoin the battlefield.
3. A player should be able to leave a battlefield and reenter for some battlefield fights (those that don't require a KI from all players to enter)

The way I went about solving this is to set the battlefield effect subpower to 1 when the player enters the battlefield and to 0 when exiting the battlefield. When reconnecting we check if the subpower is 1 (inside battlefield) and eject if they cant rejoin the battlefield.

These changes are based off of the following captures:
https://www.youtube.com/watch?v=V0Qp99O1AkA
https://www.youtube.com/watch?v=SeKiF3qFiAQ

Attempt the aforementioned interactions and observe they are correct.

Useful commands for testing with Brothers fight.
```
!addkeyitem ZEPHYR_FAN
!pos 121 -171 758 6
```

### DIFF
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -1022,6 +1022,26 @@ void CZone::CharZoneIn(CCharEntity* PChar)
         {
             PBattlefield->InsertEntity(PChar, true);
         }
+        else if (PChar->StatusEffectContainer->HasStatusEffectByFlag(EFFECTFLAG_CONFRONTATION))
+        {
+            // Player is in a zone with a battlefield but they are not part of one.
+            if (PChar->StatusEffectContainer->GetStatusEffect(EFFECT_BATTLEFIELD)->GetSubPower() == 1)
+            {
+                // If inside of the battlefield arena then kick them out
+                m_BattlefieldHandler->addOrphanedPlayer(PChar);
+            }
+            else
+            {
+                // Is not inside of a battlefield arena so remove the battlefield effect
+                PChar->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_CONFRONTATION, true);
+                PChar->StatusEffectContainer->DelStatusEffect(EFFECT_LEVEL_RESTRICTION);
+            }
+        }
+    }
+    else if (PChar->StatusEffectContainer->HasStatusEffectByFlag(EFFECTFLAG_CONFRONTATION))
+    {
+        // Player is zoning into a zone that does not have a battlefield but the player has a confrontation effect - remove it
+        PChar->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_CONFRONTATION, true);
     }
 
     PChar->PLatentEffectContainer->CheckLatentsZone();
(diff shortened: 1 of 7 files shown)


WinterSolstice8 (approved):

src/map/zone.cpp
```diff
@@ -1022,6 +1022,26 @@ void CZone::CharZoneIn(CCharEntity* PChar)
         {
             PBattlefield->InsertEntity(PChar, true);
         }
+        else if (PChar->StatusEffectContainer->HasStatusEffectByFlag(EFFECTFLAG_CONFRONTATION))
```
claywar, line 1025: Should this block be dependent on confrontation?  I'm not as familiar with this side of the codebase, but I thought that effect was only applied for open-world fights (like Unity, etc)

jmcmorris: EFFECT_BATTLEFIELD is flagged as a confrontation so this does work.

I will admit I'm not entirely sure about other confrontation effects and how they should be handled particularly when zoning into an area. This is something I'll look into and get back to you.

TeoTwawki: we should probably rename the effect flag - it will be an ongoing source of confusion to both have a status effect and an effect flag using that name, when we inevitably (as here) expand to using it for things other than the actual confrontation status effect.

jmcmorris: So all of the status effects with the confrontation tag include:

- battlefield
- allied_tags (https://ffxiclopedia.fandom.com/wiki/Allied_Tags)
- confrontation (https://ffxiclopedia.fandom.com/wiki/Confrontation)
- pennant (https://ffxiclopedia.fandom.com/wiki/Pennant)
- reive_mark (https://ffxiclopedia.fandom.com/wiki/Reive_Mark)

My understanding is that all of these are removed upon zoning and likely follow the same logic. Remove effect upon zoning. From the looks of it Battlefield seems to be an exception where it will not remove the effect when zoning as long as you are in the Battlefield zone.

jmcmorris: Is renaming the confrontation flag something you'd like me to do within this PR? I can do that or one of us can make an issue to follow up on.

claywar: Outside of my wheelhouse.  IIRC, Zach was the last person do do work for confrontation in general, and might need him to weigh in (or someone else familiar with our flag usage)

WinterSolstice8: > From the looks of it Battlefield seems to be an exception where it will not remove the effect when zoning as long as you are in the Battlefield zone.

I confirmed this just now in retail, intentionally d/ced and waited for time out in real time, logged back in, I had the battlefield status, but then the battlefield kicked me out and during/after the kick-out CS i lost it.

jmcmorris:
Fixed https://github.com/LandSandBoat/server/issues/886

PR #2914, opened 2022-10-10
Update rocks array to match elements in battleutils.

austinclprojects:
**_I affirm:_**

`HELM.lua` was using the old order of elements in `battleutils.h` to determine what rock to give. This corrects the rocs array to match the new order.

Compare array to values in battleutils.

### DIFF
--- a/scripts/globals/helm.lua
+++ b/scripts/globals/helm.lua
@@ -5,12 +5,14 @@
 -- https://ffxiclopedia.wikia.com/wiki/Logging
 -- https://ffxiclopedia.wikia.com/wiki/Mining
 -----------------------------------
+require("scripts/globals/items")
 require("scripts/globals/keyitems")
 require("scripts/globals/missions")
 require("scripts/globals/npc_util")
 require("scripts/globals/quests")
 require("scripts/globals/roe")
 require("scripts/globals/settings")
+require("scripts/globals/spell_data")
 require("scripts/globals/status")
 require("scripts/globals/zone")
 require("scripts/missions/amk/helpers")
@@ -1389,10 +1391,20 @@ local helmInfo =
 }
 
 -----------------------------------
--- colored rocks. do not change this order!
+-- colored rocks array
 -----------------------------------
 
-local rocks = { 769, 771, 770, 772, 773, 774, 776, 775 }
+local rocks =
+{
+    [xi.magic.element.FIRE] = xi.items.RED_ROCK,
+    [xi.magic.element.ICE] = xi.items.TRANSLUCENT_ROCK,
+    [xi.magic.element.WIND] = xi.items.GREEN_ROCK,
+    [xi.magic.element.EARTH] = xi.items.YELLOW_ROCK,
+    [xi.magic.element.THUNDER] = xi.items.PURPLE_ROCK,
+    [xi.magic.element.WATER] = xi.items.BLUE_ROCK,
+    [xi.magic.element.LIGHT] = xi.items.WHITE_ROCK,
+    [xi.magic.element.DARK] = xi.items.BLACK_ROCK,
+}
 
 -----------------------------------
 -- local functions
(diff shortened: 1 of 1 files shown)


claywar (approved):

scripts/globals/helm.lua
```diff
@@ -1392,7 +1392,7 @@ local helmInfo =
 -- colored rocks. do not change this order!
 -----------------------------------
 
-local rocks = { 769, 771, 770, 772, 773, 774, 776, 775 }
+local rocks = { 769, 773, 772, 771, 774, 770, 776, 775 }
```
jmcmorris: Two thoughts here:
1. It concerns me that there is a comment above that says `do not change this order!`. If is truly ok for us to change then we should remove that comment.
2. If we do want these to match the elements then we should do so explicitly. Using the constant values makes things much more readable too!

```lua
-- At top of file
require("scripts/globals/items")

local rocks =
{
    [xi.magic.element.FIRE] = xi.items.RED_ROCK,
    [xi.magic.element.ICE] = xi.items.TRANSLUCENT_ROCK,
    [xi.magic.element.WIND] = xi.items.GREEN_ROCK,
    [xi.magic.element.EARTH] = xi.items.YELLOW_ROCK,
    [xi.magic.element.THUNDER] = xi.items.PURPLE_ROCK,
    [xi.magic.element.WATER] = xi.items.BLUE_ROCK,
    [xi.magic.element.LIGHT] = xi.items.WHITE_ROCK,
    [xi.magic.element.DARK] = xi.items.BLACK_ROCK,
}
```

jmcmorris (changes requested):

scripts/globals/helm.lua
zach2good: +1 for concern about the comment: do not change this order!

I remember there being a big refactor to try and bring weather's and elements and all that sort of stuff in line a long time ago and these things can be very sneaky if handled incorrectly. Proceed with caution!

austinclprojects: @zach2good 

Yea, I think colored rocks got missed in that refactor which is why they are specifically out of whack now. The use `VanadielDayElement()` which gets the element from battleutils.h. I checked and that does match the `xi.magic.element`  array in `spell_data.lua` so the suggestion to change to global constants is a good one. I intentionally left it as close to what it was as possible as I'm not trying to change yalls flow. 

Regarding the "Do not Change" message, it is ~4 yrs old and the refactor that broke this was ~2 years ago, so I think it's safe to delete it.

zach2good: Thanks for looking into that, sounds good 🙏👍

austinclprojects: Updated per recommendations.

scripts/globals/helm.lua
```diff
@@ -1389,10 +1391,20 @@ local helmInfo =
...
+{
+    [xi.magic.element.FIRE] = xi.items.RED_ROCK,
+    [xi.magic.element.ICE] = xi.items.TRANSLUCENT_ROCK,
+    [xi.magic.element.WIND] = xi.items.GREEN_ROCK,
+    [xi.magic.element.EARTH] = xi.items.YELLOW_ROCK,
+    [xi.magic.element.THUNDER] = xi.items.PURPLE_ROCK,
+    [xi.magic.element.WATER] = xi.items.BLUE_ROCK,
+    [xi.magic.element.LIGHT] = xi.items.WHITE_ROCK,
+    [xi.magic.element.DARK] = xi.items.BLACK_ROCK,
+}
```
claywar, line 1407: Dug a little deeper, and I think there's an error in the logic as well.
```
    -- if we picked a colored rock, change it to the day's element
    if item == 769 then
        item = rocks[VanadielDayElement() + 1]
    end
```

This shouldn't be + 1, since we should never get a none-type daily element

TeoTwawki: probably leftover from the old methods of aligning element and day since their orders didn't match back then and they didn't start at the same value

austinclprojects: That was actually me screwing up on git. I had originally fixed that but when I squashed I lost the change somehow.

claywar (approved):

TeoTwawki (approved):

jmcmorris (approved):

PR #2927, opened 2022-10-12
Modifier To Nullify Ranged Damage Only

jmcmorris:
**_I affirm:_**

Adding `Mod::NULL_RANGED_DAMAGE` so we can distinguish between this and `Mod::NULL_PHYSICAL_DAMAGE`. I added usage of this in all of the places we were using `NULL_PHYSICAL_DAMAGE` so that removing it from the ranged damage check doesn't affect anything. This change is needed for NE Apollyon.

Give a monster the new `Mod::NULL_RANGED_DAMAGE` and attempt to do ranged damage to it.


claywar (approved):

WinterSolstice8 (approved):

TeoTwawki (approved):
we should probably also have a mod for ranged damage taken as well as % chance to null it. there are going to be future mobs that will take increased or reduced ranged damage (not just "piercing") for sure.

jmcmorris:
> we should probably also have a mod for ranged damage taken as well as % chance to null it. there are going to be future mobs that will take increased or reduced ranged damage for sure.

This new mod is a % change to nullify it. We can definitely add the Ranged DT mod. I haven't looked into that too much yet but I'll toss it onto my growing backlog.

PR #2932, opened 2022-10-14
CS Apollyon Rework

jmcmorris:
**_I affirm:_**

Migrate and rework CS Apollyon Limbus area to be in the new battlefield framework. All of this work is heavily based off of the following captures and sources:
https://ffxiclopedia.fandom.com/wiki/CS_Apollyon
https://www.bg-wiki.com/ffxi/CS_Apollyon
https://www.youtube.com/watch?v=NHDH5Nz7MAE
https://www.youtube.com/watch?v=0E4At8VB4nQ
https://www.youtube.com/watch?v=iJgt77S_Af4
https://www.youtube.com/watch?v=qnEGCFWnOns
https://www.youtube.com/watch?v=80n5J0yh7ng

There are some supporting changes that I committed separately so they can easily be viewed distinctly from the core CS Apollyon work.

```
!addkeyitem red_card
!addkeyitem cosmo_cleanse
!additem 2127
!pos -600 -0.5 -600 38
```


claywar (approved):

scripts/battlefields/Apollyon/cs_apollyon.lua
```diff
@@ -0,0 +1,312 @@
...
+
+        local group = boss:getLocalVar("supportGroup")
+        local offsets = group == 0 and supportOffsets or otherSupportOffsets
+        for _, offset in ipairs(offsets) do
+            if GetMobByID(bossID + offset):getStatus() ~= xi.status.DISAPPEAR then
+                return
+            end
+        end
+
+        boss:useMobAbility(603)
```
WinterSolstice8: Surely, they don't actually use lateral slash which is a tonberry TP move

Consider using `boss:injectActionPacket` to send the action packet to give the client what it needs and a timer to replace the weaponskill listener

If injectActionPacket doesn't do what we need then we can add functionality to do it

jmcmorris: I didn't know about `injectActionPacket` I'll give that a try. Here's what the `0x028 Action By NPC` packet looks like.
image

sql/mob_skills.sql
```diff
@@ -436,7 +436,7 @@ INSERT INTO `mob_skills` VALUES (599,885,'arctic_impact',1,15.0,2000,2000,4,0,0,
 INSERT INTO `mob_skills` VALUES (600,886,'cold_wave',1,15,2000,2000,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (601,887,'hiemal_storm',1,15.0,2000,2000,4,0,0,0,0,0,0);
 -- INSERT INTO `mob_skills` VALUES (602,346,'hypothermal_combustion',0,7.0,2000,1500,4,0,0,0,0,0,0);
--- INSERT INTO `mob_skills` VALUES (603,432,'lateral_slash',0,7.0,2000,1500,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (603,432,'lateral_slash_nm',0,7.0,2000,0,1,0,0,0,0,0,0);
```
TeoTwawki: Strongly suggest a more decretive term appended than `_nm`

jmcmorris: Removed this entirely as I'm now using `injectActionPacket`

jmcmorris:
@TeoTwawki How do you feel about this new iteration? The mob scripts call into the battlefield function defined in `cs_apollyon.lua`.

TeoTwawki:
> @TeoTwawki How do you feel about this new iteration? The mob scripts call into the battlefield function defined in `cs_apollyon.lua`.

I think that looks good :+1:

TeoTwawki (approved):

WinterSolstice8 (approved):

PR #2940, opened 2022-10-15
Adjusting Omega skills used in One to be Feared

jmcmorris:
**_I affirm:_**

This is in part a change from Eden made by @slassen.

Change Omega (One To Be Feared) skills to use proper skill ids and add logic to switch skill list when below 25% HP.

I'm primarily making this change in preparation for Proto-Omega in Central Apollyon.

```
!addmission 6 638
!zone 32
```

Get to the Omega portion of the fight. Above 25% Omega will have one skillset (most abilities) and then change to just Pile Pitch and Discharger at 25%.


scripts/zones/Sealions_Den/mobs/Omega.lua
```diff
@@ -15,21 +15,22 @@ end
...
+    local stage = mob:getLocalVar("stage")
+
+    if mob:getHPP() < 60 and stage == 0 then
         mob:setDelay(3000)
-    elseif mob:getHPP() < 25 then
+        mob:setLocalVar("stage", 1)
+    elseif mob:getHPP() < 25 and stage < 2 then
         mob:setDelay(2500)
+        mob:addStatusEffect(xi.effect.REGAIN, 5, 3, 0)
+        mob:getStatusEffect(xi.effect.REGAIN):setFlag(xi.effectFlag.DEATH)
```
TeoTwawki: I don't think retail uses the actual status effect for this does it? Can a thf nab it with aurasteal procs? some pretty crazy things can be aura-stole including the darkrider's invincibility (which immediately wears off and self-reapplies to him, but it's still pretty funny to witness) but I'd wager what we probably need ot do is setMod the regain modifier to turn this on and off.

p.s. we probably want to start a pattern of bitwise op-ing the flags on status effects, rather than wholesale overwriting the mask like has been unfortunately done in the past (and that seems like a good thing to maybe make a helper for if we keep the status).

jmcmorris: Honestly not sure and I can change it if desired. My guess is that it is likely a modifier. I was just relocating that chunk of code to this section. On my backlog I have it to go through all of the CoP battlefields and fix them up to have proper mechanics.

WinterSolstice8: +1 to move to a setMod instead of a status effect, would be pretty cool if retail could let you yoink a regain effect though!

jmcmorris: Ok, I changed it to be a mod and to increase over time which is what Eden has and is based off of this awesome capture by Jimmayus: https://discord.com/channels/443544205206355968/443552700064333835/644089353278193665

claywar (approved):
Seems reasonable to me, just need Teo to weigh in

WinterSolstice8 (approved):

TeoTwawki (approved):

PR #2941, opened 2022-10-15
Central Apollyon Rework

jmcmorris:
**_I affirm:_**

Migrate and rework Central Apollyon Limbus area to be in the new battlefield framework. All of this work is heavily based off of the following captures and sources:
https://www.youtube.com/watch?v=lJV2juzuDQQ
https://www.youtube.com/watch?v=qk-f6a1kBBg
https://ffxiclopedia.fandom.com/wiki/Central_Apollyon
https://www.bg-wiki.com/ffxi/Central_Apollyon

A huge chunk of these changes is dealing with mobskills. Previously Omega skills had lots of conditional checks happening to determine when it could use a skill. That logic has been stripped out and it's now simply changing the skill list when changing forms.

```
!addkeyitem black_card
!addkeyitem cosmo_cleanse
!additem 1909
!additem 1910
!additem 1987
!additem 1988
!pos 600 -0.5 -600 38
```


jmcmorris:
This is mostly just waiting for #2940 to make it through since they both touch mobskills.

claywar (approved):

scripts/globals/battlefield.lua
```diff
@@ -82,6 +82,7 @@ xi.battlefield.id =
     NE_APOLLYON = 1292,
     SE_APOLLYON = 1293,
     CS_APOLLYON = 1294,
+    CENTRAL_APOLLYON = 1296,
```
WinterSolstice8: CI didn't trip for alignment here? interesting.

claywar: Would be something nice to have added in the future, but I think that check is only on the cpp side for now.

jmcmorris: I can fix that alignment in just a few

jmcmorris: Actually, would you guys be opposed if I just filled out the entire thing with all existing battlefields? I can easily do this and would be much easier than adding them one at a time.

WinterSolstice8: works for me

WinterSolstice8 (approved):

PR #2946, opened 2022-10-16
[sql] Use CHAR instead of VARCHAR in PopulateIDLookups conversion

### LINKED ISSUE #2939: 🐛 SQL Syntax Error around UPPER(REPLACE(CAST(name as VARCHAR(64)), '-', ''))
[map][error] Query: SELECT npcid FROM npc_list WHERE ((npcid >> 12) & 0xFFF) = 235 AND UPPER(REPLACE(CAST(`name` as VARCHAR(64)), '-', '_')) = 'AQUILLINA' LIMIT 1; (SqlConnection::QueryStr:322)
[10/15/22 07:43:40:030][map][error] mysql_real_query: SQL_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'VARCHAR(64)), '-', '_')) = 'AQUILLINA' LIMIT 1' at line 1 (1064) (SqlConnection::QueryStr:323)
[10/15/22 07:43:40:030][map][error] Could not complete id lookup for Bastok_Markets.ID.npc.AQUILLINA (luautils::PopulateIDLookups::<lambda_1>::operator ():832)

zach2good:
**_I affirm:_**

I wasn't able to repro locally, but this query still works while using CHAR, so it's worth a go to try and fix the issue.

Might fix: https://github.com/LandSandBoat/server/issues/2939

https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#function_convert

CAST might not accept VARCHAR

See no MySQL errors on startup

### DIFF
--- a/sql/pet_list.sql
+++ b/sql/pet_list.sql
@@ -11,7 +11,7 @@ SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";
 
 DROP TABLE IF EXISTS `pet_list`;
 CREATE TABLE IF NOT EXISTS `pet_list` (
-  `petid` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `petid` int(10) unsigned NOT NULL,
   `name` char(15) NOT NULL,
   `poolid` int(10) unsigned NOT NULL DEFAULT '0',
   `minLevel` tinyint(2) unsigned NOT NULL DEFAULT '0',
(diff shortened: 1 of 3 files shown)


claywar (approved):

zach2good:
Also fixes up a small warning about auto-increment stuff, from `PS C:\Program Files\MariaDB 10.6\bin> .\mysqlcheck.exe --check xidb --user root -proot`

PR #2958, opened 2022-10-17
[lua] Update !additem command to intitalize quantity to 1 if nil, check retval of player:addItem

WinterSolstice8:
**_I affirm:_**

The !additem command was not initializing quantity, so if you had a full inventory and did not specify quantity the script would error.
The additem command was also incorrectly informing you both got and didn't get rare items if you had that item already (revtal of addItem was not being checked)

Get a full inventory
`!additem onion_sword` and receive a full inventory message
`!additem onion_sword 1` and receive a full inventory message
image

`!additem sky_orb` x2
The first time you should get sky orb, second time it should only tell you that you can't get the item

### DIFF
--- a/scripts/commands/additem.lua
+++ b/scripts/commands/additem.lua
@@ -41,6 +41,9 @@ function onTrigger(player, item, quantity, aug0, aug0val, aug1, aug1val, aug2, a
         itemToGet = tonumber(item)
     end
 
+    -- if quantity is nil, assume 1 qty
+    quantity = quantity or 1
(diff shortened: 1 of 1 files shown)


zach2good (approved):

WinterSolstice8:
Description updated

WinterSolstice8:
Out of scope for this PR: Currently if you try to add qty 2 of rare item it tells you that you got both, the system isn't really set up to handle that.

PR #2966, opened 2022-10-18
[core] Add binding to remove all or specific Gambits

zach2good:
**_I affirm:_**

- Makes AddGambit return a string
- Adds RemoveGambit, which will remove a gambit based on the given string
- Adds RemoveAllGambits, which, removes all gambits

Man, `fmt::format` is great.

Remove all the gambits from Kupipi, replace with:
```lua
    local id = mob:addSimpleGambit(ai.t.TARGET, ai.c.NOT_STATUS, xi.effect.FLASH, ai.r.MA, ai.s.SPECIFIC, xi.magic.spell.FLASH, 60)
    mob:removeSimpleGambit(id)
```

She should do nothing but melee, as she has no gambits.

### DIFF
--- a/src/map/ai/helpers/gambits_container.cpp
+++ b/src/map/ai/helpers/gambits_container.cpp
@@ -22,7 +22,7 @@ namespace gambits
 {
     // Validate gambit before it's inserted into the gambit list
     // Check levels, etc.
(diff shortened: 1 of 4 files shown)


zach2good:
Ready for review

claywar (approved):

PR #2971, opened 2022-10-19
Fixing Throne Room Battlefields

jmcmorris:
**_I affirm:_**

Apparently the client uses the entity names for some cutscenes such as opening doors in the Throne Room. I changed this awhile back and it broke things and now I'm reverting that. In order to keep the name we needed to support multiple exits in the battlefield setup.

An additional change is to fix an issue where winning some mission battlefields would not advance the mission status properly.

Thanks to @N3ckB3ard to bringing this to my attention.

```
!addmission 1 15
!setmissionstatus <t> 2 1
!pos -111 -6 0.1 165
```

Go through the cutscenes, go into the battlefield, Run away from the battlefield, go back and defeat the Shadow Lord and everything should progress nicely.

### DIFF
--- a/scripts/battlefields/Throne_Room/shadow_lord_battle.lua
+++ b/scripts/battlefields/Throne_Room/shadow_lord_battle.lua
@@ -17,8 +17,8 @@ local content = BattlefieldMission:new({
     levelCap      = 75,
     timeLimit     = utils.minutes(30),
     index         = 0,
-    entryNpc      = "Throne_Room",
-    exitNpc       = "Throne_Room_Exit",
+    entryNpc      = "_4l1",
+    exitNpcs      = { "_4l2", "_4l3", "_4l4" },
 
     mission               = xi.mission.id.nation.SHADOW_LORD,
     requiredMissionStatus = 3,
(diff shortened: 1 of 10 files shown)


Xaver-DaRed (approved):
Pre-ventive  aproval.

Having said that, I don't think the empty npc scripts are necessary.
EDIT: They are, it seems. Disregard.

Also, a newline is needed at the end of the files if we are keeping them.

TeoTwawki:
```
    entryNpc      = "_4l1",
    exitNpcs      = { "_4l2", "_4l3", "_4l4" },
```
ah yeah, anytime you see names like this they are special. Last time we had this happen that I can recall, we broke the entrance to the ark angels and divine might

jmcmorris:
> Pre-ventive aproval.
> 
> Having said that, I don't think the empty npc scripts are necessary.
> 
> Also, a newline is needed at the end of the files if we are keeping them.

Without the scripts the doors open when you trigger then and leave the menu. Is there another way to prevent that from happening?

TeoTwawki:
> Without the scripts the doors open when you trigger then and leave the menu. Is there another way to prevent that from happening?

no you have it right, I think xaver didn't realize

Xaver-DaRed:
> > Without the scripts the doors open when you trigger then and leave the menu. Is there another way to prevent that from happening?
> 
> no you have it right, I think xaver didn't realize

Pretty much.

claywar (approved):

TeoTwawki (approved):

PR #2092, opened 2022-06-24
[core] Update zoneutils::GetEntity to match dynamic entity index change

WinterSolstice8:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

Updates zoneutils::GetEntity to match the ID change logic for Dynamic Entities
Fixes relating to !gotoid causing a debug break on pet IDs
Removes 2 warnings about no initializers in lua_baseentity.cpp
Closes #2047

## Steps to test these changes

Before PR:
Use !fafnir in a zone, use targetinfo to try to !gotoid to the new fafnir, the command will silently fail.

After PR
Use !fafnir in a zone, use targetinfo to try to !gotoid to it, the command will work.

- [x]     spawn fafnir
- [x]     still be able to target and punch him in the face
- [x]     cast spells on him
- [x]     Same with other regular mobs
- [x]     make sure instances still work
- [x]     make sure bcnms still work

PR #2979, opened 2022-10-20
[core] Merge pet/trust/dynamic entities into 0x700 targid range

WinterSolstice8:
* Various cleanup along the way, such as removing several unnecessary re-iteratons of std::maps

**_I affirm:_**

Merges pet/trust/dynamic entity into 0x700+ range
Cleanups along the way, notably removal of several needless re-searches of lists or redundant looping of the same list

targid sharing:
Summon trust, note targid 0x700
Summon a pet, note targid 0x701
`!fafnir`, note targid 0x702

desummon pet and trust, wait for nameplates to disappear
`!fafnir` x2
note new targids of 0x700 and 0x701
`!hp 0` on the fafnir targid 0x702, wait for nameplate to disappear
summon pet or trust
note targid of 0x702 on trust or pet

basic entity checks:
run in/out of range of pets and mobs and see them appear/disappear
you can use `!setmod move -100` to "bind" your pet in place to do this. Doesn't work on trusts because they teleport.

Extra credit:
breakpoint in `CZoneEntities::ZoneServer` and verify that `m_petList` and `m_trustList` are set back to size 0 after trusts and pets are despawned from the zone

Old checklist for dynamic entities:
- [x]     spawn fafnir
- [x]     still be able to target and punch him in the face
- [x]     cast spells on him
- [x]     Same with other regular mobs
- [x]     make sure instances still work
- [x]     make sure bcnms still work
- [x]     make sure !gotoid works on a dynamic entities ID
- [x]     make sure !gotoid works on a dynamic entities ID when the entities targid is above 0x800

### DIFF
--- a/src/map/entities/petentity.cpp
+++ b/src/map/entities/petentity.cpp
@@ -44,12 +44,13 @@ CPetEntity::CPetEntity(PET_TYPE petType)
 : CMobEntity()
 , m_PetType(petType)
 {
-    objtype        = TYPE_PET;
-    m_EcoSystem    = ECOSYSTEM::UNCLASSIFIED;
-    allegiance     = ALLEGIANCE_TYPE::PLAYER;
-    m_MobSkillList = 0;
-    m_PetID        = 0;
-    m_IsClaimable  = false;
+    objtype                 = TYPE_PET;
+    m_EcoSystem             = ECOSYSTEM::UNCLASSIFIED;
+    allegiance              = ALLEGIANCE_TYPE::PLAYER;
+    m_MobSkillList          = 0;
+    m_PetID                 = 0;
+    m_IsClaimable           = false;
+    m_bReleaseTargIDOnDeath = true;
 
     PAI = std::make_unique<CAIContainer>(this, std::make_unique<CPathFind>(this), std::make_unique<CPetController>(this), std::make_unique<CTargetFind>(this));
 }
--- a/src/map/lua/lua_zone.cpp
+++ b/src/map/lua/lua_zone.cpp
@@ -246,6 +246,11 @@ std::optional<CLuaBaseEntity> CLuaZone::insertDynamicEntity(sol::table table)
     m_pLuaZone->GetZoneEntities()->dynamicTargIds.insert(PEntity->targid);
 
     PEntity->id = 0x1000000 + (ZoneID << 12) + PEntity->targid;
+    // Add 0x100 if targid is >= 0x800 -- observed on retail.
+    if (PEntity->targid >= 0x800)
+    {
+        PEntity->id += 0x100;
+    }
 
     PEntity->loc.zone       = m_pLuaZone;
     PEntity->loc.p.rotation = table.get_or<uint8>("rotation", 0);
--- a/src/map/lua/lua_zone.cpp
+++ b/src/map/lua/lua_zone.cpp
@@ -245,7 +245,7 @@ std::optional<CLuaBaseEntity> CLuaZone::insertDynamicEntity(sol::table table)
 
     m_pLuaZone->GetZoneEntities()->dynamicTargIds.insert(PEntity->targid);
 
-    PEntity->id = 0x1000100 + (ZoneID << 12) + PEntity->targid;
+    PEntity->id = 0x1000000 + (ZoneID << 12) + PEntity->targid;
 
     PEntity->loc.zone       = m_pLuaZone;
     PEntity->loc.p.rotation = table.get_or<uint8>("rotation", 0);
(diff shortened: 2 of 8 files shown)


src/map/entities/petentity.cpp
```diff
@@ -44,12 +44,13 @@ CPetEntity::CPetEntity(PET_TYPE petType)
...
-    m_MobSkillList = 0;
-    m_PetID        = 0;
-    m_IsClaimable  = false;
+    objtype                 = TYPE_PET;
+    m_EcoSystem             = ECOSYSTEM::UNCLASSIFIED;
+    allegiance              = ALLEGIANCE_TYPE::PLAYER;
+    m_MobSkillList          = 0;
+    m_PetID                 = 0;
+    m_IsClaimable           = false;
+    m_bReleaseTargIDOnDeath = true;
```
zach2good, line 53: Can we move this down into battle entity now?

src/map/lua/lua_zone.cpp
```diff
@@ -245,7 +245,7 @@ std::optional<CLuaBaseEntity> CLuaZone::insertDynamicEntity(sol::table table)
 
     m_pLuaZone->GetZoneEntities()->dynamicTargIds.insert(PEntity->targid);
 
-    PEntity->id = 0x1000100 + (ZoneID << 12) + PEntity->targid;
+    PEntity->id = 0x1000000 + (ZoneID << 12) + PEntity->targid;
```
zach2good, line 248: Why is the freaky little offset not needed anymore?

zach2good (approved):
This looks like a good first step of cleanup in here, and towards collapsing those different spawn lists into one

src/map/entities/petentity.cpp
WinterSolstice8: the variable declaration itself? I don't see why not.

src/map/lua/lua_zone.cpp
WinterSolstice8: I think the answer is that was a hack that was needed before #2092 to prevent crashing on windower but wasn't removed until now. Without removing that, the only thing that doesn't work is battlemod because windower's ID system actually checks targid vs ID. The client doesn't seem to care and displayed the messages anyway.

It has to do with the fact that targIDs previously started at 0x800 but were being calculated wrong when transformed to IDs.

zach2good: So, for all these entities, once we get past 0x800 (2048), we don't need that transformation? Battlemod will work for everything? What about up to 0x900 (4096)

WinterSolstice8: It should be fine, but I'll test it by spawning a ton of dynamic entities.

WinterSolstice8: Stops working at exactly 0x800. 0x7FF is fine. very interesting.

WinterSolstice8: Based on retail, it looks like they add 0x100 to ID if targid >= 0x800
Pic from retail, an automaton in Sih Gates (zone id 268)
image
`0x1000000 + (268 << 12) + 0x819` = `17877017`
`0x1000100 + (268 << 12) + 0x819` = `17877273`

WinterSolstice8: Once I added that change in to check for 0x800 everything worked fine.

src/map/utils/charutils.cpp
```diff
@@ -6616,26 +6616,23 @@ namespace charutils
...
-                spawnlist = &PChar->SpawnNPCList;
-            }
+            spawnlist = &PChar->SpawnMOBList;
+        }
+        else if (entity->objtype == TYPE_NPC)
+        {
+            spawnlist = &PChar->SpawnNPCList;
         }
-        else if (entity->targid < 0x700)
+        else if (entity->targid == TYPE_PC)
```
TeoTwawki: I think you meant to change these from `targid` to `objtype` like the one above

WinterSolstice8: Oops. fixed. thanks for catching that

src/map/lua/lua_zone.cpp
```diff
@@ -245,7 +245,12 @@ std::optional<CLuaBaseEntity> CLuaZone::insertDynamicEntity(sol::table table)
...
+    // Add 0x100 if targid is 0x800 -- observed on retail.
```
zach2good: Good comment, thanks!

src/map/lua/lua_zone.cpp
zach2good: Nice

zach2good (approved):

PR #2980, opened 2022-10-20
[cpp] Add better logging for missing event NPCs

zach2good:
**_I affirm:_**

image

- !zone {Mog Garden}
- !cs 1
- Look at ur logs

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -726,6 +726,10 @@ void SmallPacket0x016(map_session_data_t* const PSession, CCharEntity* const PCh
             if (!PEntity)
             {
                 PEntity = zoneutils::GetTrigger(targid, PChar->getZone());
+
+                // PEntity->id will now be the full id of the entity we could not find
+                ShowWarning(fmt::format("Server missing npc_list.sql entry <{}> in zone <{} ({})>",
+                                        PEntity->id, zoneutils::GetZone(PChar->getZone())->GetName(), PChar->getZone()));
             }
             PChar->updateEntityPacket(PEntity, ENTITY_SPAWN, UPDATE_ALL_MOB);
         }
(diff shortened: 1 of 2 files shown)


src/map/zone.cpp
```diff
@@ -1197,6 +1197,315 @@ void CZone::CheckRegions(CCharEntity* PChar)
     PChar->m_InsideRegionID = RegionID;
 }
 
+std::string CZone::ZoneIDToString(uint16 zoneID)
```
TeoTwawki: this ginormous switch shouldn't be needed when we can get zone from its ID and getName() on the zone already in other places.

zach2good: On multi process when a zone isn't loaded, we won't be able to query it?

TeoTwawki: how are we requesting an npc for a zone that isn't loaded? that simply should not be happening, ever. Its a larger problem than missing an npc and we should bail out and say the zone isn't there for the message in such a case.

TeoTwawki: I forgot but the zone is already baked into the id, we can get it from there with some bit magic

TeoTwawki: 17924154 is zone 24 and 58th entity out of its 0-4095 block

TeoTwawki: I'm not gonna stand in the way of merge, but I really feel we should handle that situation (missing/wrong zone) differently so that naming it becomes a non-issue

claywar (approved):

src/map/zone.cpp
zach2good: Yeah you were right, I wasn't paying proper attention. We have full access to PChar and PZone one level up from where the logging happens, so I'll just move the logging up there

TeoTwawki (approved):

PR #2982, opened 2022-10-20
Core rework to add Trust iLvl scaling to emulate retail-like conditio…

GodModeXI:
…ns for HP/MP as well as Stat calculation modifiers

**_I affirm:_**

- This PR is designed to add an iLvl scaling emulation to Trusts considering Ilvl of the master for mLvl/sLvl adjustments.
- this PR scales HP/MP and modifies Stat levels to aid in trust efficiency from level 99-119. 
- this was taken from retail observations via youtube of the trust Valaineral.
- Valaineral sample data.
val 99
Val 101
val 103
val 105
val 107
val 109
val 114
val 119
ilvl rework table

1.  load in core change files
2.  Clean and rebuild Server to apply core changes
3.  Summon trusts with various levels of iLVL gear on and use !getstats to observe stat adjustments.


Xaver-DaRed:
Valaineral retail HP MP values
lvl 099 - 1813 | 492
lvl 103 - 1857 | 676
lvl 105 - 1881 | 685
lvl 107 - 1903 | 693
lvl 109 - 1926 | 702
lvl 111 - 1948 | 710
lvl 113 - 1970 | 718
lvl 115 - 1993 | 727
lvl 117 - 2015 | 735
lvl 119 - 2038 | 744

GodModeXI:
> Valaineral retail HP MP values
> 
> lvl 099 - 1813 | 492
> 
> lvl 103 - 1857 | 676
> 
> lvl 105 - 1881 | 685
> 
> lvl 107 - 1903 | 693
> 
> lvl 109 - 1926 | 702
> 
> lvl 111 - 1948 | 710
> 
> lvl 113 - 1970 | 718
> 
> lvl 115 - 1993 | 727
> 
> lvl 117 - 2015 | 735
> 
> lvl 119 - 2038 | 744

It was hard to get them exact since his 99 base wasn't exactly the same to build off of. Thank you for adding the additional information for reference.

TeoTwawki:
we have similar issues for other entity types right now. even our base grades for players veer off after 75. Mobs need extended out to at least 150. There are a few locus mobs I have seen spawn at lv145 and several NM are 150.

But we lack data points to determine how much STR a mob with `rank A` STR gets up to post 99, for example. If we had those, we could fix trust stats to be exact by figuring out the rankings (they are technically mobs as well, but their mob family is "humanoid" for _most_ of them). 

Right now even player stats aren't fully accurate post 75. We just get as close as we can and err on the side of not making anybody too strong.

GodModeXI:
> we have similar issues for other entity types right now. even our base grades for players veer off after 75. Mobs need extended out to at least 150. There are a few locus mobs I have seen spawn at lv145 and several NM are 150.
> 
> 
> 
> But we lack data points to determine how much STR a mob with `rank A` STR gets up to post 99, for example. If we had those, we could fix trust stats to be exact by figuring out the rankings (they are technically mobs as well, but their mob family is "humanoid" for _most_ of them). 
> 
> 
> 
> Right now even player stats aren't fully accurate post 75. We just get as close as we can and err on the side of not making anybody too strong.

Yes, I essentially watched an 119 val versus Apex mobs in Cirdas to gather top shelf values, then tried to reasonably bridge the difference in 5 level steps, leaning a little heavier on the high end since even on retail trusts tend miss frequently on 119+ mobs and lack in damage. The total value change in the chart from bottom to top seems healthy while still landing way below what a Player would be producing.

TeoTwawki:
> Yes, I essentially watched an 119 val versus Apex mobs in Cirdas to gather top shelf values, then tried to reasonably bridge the difference in 5 level steps, leaning a little heavier on the high end since even on retail trusts tend miss frequently on 119+ mobs and lack in damage. The total value change in the chart from bottom to top seems healthy while still landing way below what a Player would be producing.

I think what our mobs currently do tho, is take the last known +stat of our array and simply repeat that same gain on all subsequent levels. so the point gain from 74>75 is identical to you point gain at 75>76 and at 98>99. If there's a big enough difference between that and what you did here with trusts, the trusts could either be far stronger or far weaker than the mobs they'll be fighting. I'm not super concerned tho, because everything above 75 is a crap shoot, item levels even more so, and by now server operators ought to realize they play with 119 stuff at their own risk.

N3ckB3ard:
Do you have this RoE quest completed on retail?
It might help see a slight increase in the trusts to offer a idea of scaling based on party size

https://www.bg-wiki.com/ffxi/Marjory

TeoTwawki:
> Do you have this RoE quest completed on retail? It might help see a slight increase in the trusts to offer a idea of scaling based on party size
> 
> https://www.bg-wiki.com/ffxi/Marjory

there is also domain invasion gear that increases the trusts level. if stacked this means they get well beyond 119 (this can bring them up another 5, with it being 1 lv per each equip slot)

GodModeXI:
> Do you have this RoE quest completed on retail?
> 
> It might help see a slight increase in the trusts to offer a idea of scaling based on party size
> 
> 
> 
> https://www.bg-wiki.com/ffxi/Marjory

I didn't even know this existed to be honest. That could certainly be an exception. Wondering how to implement that.

GodModeXI:
@N3ckB3ard 
@TeoTwawki 

I'll try to figure out how to add into the level function a check for the key item for majory and gear checks for DI. I'll have to hunt around for similar logic to reference.

zach2good:
(misclicked close from mobile)

For the gear and quest adjustments to trust levels, remember that trust stats change _in real time_ when you're using iLevel, so these should all be implemented using item mods.

TeoTwawki:
> For the gear and quest adjustments to trust levels, remember that trust stats change _in real time_ when you're using iLevel, so these should all be implemented using item mods.

which is fun coz we don't want to accidently heal them when you gear swap. I suspect SE doesn't actually set the level but just dynamic add the stats equivilently.

we can see mobs lv in packet..wondering if we see trust lv there as well.

GodModeXI:
> > For the gear and quest adjustments to trust levels, remember that trust stats change _in real time_ when you're using iLevel, so these should all be implemented using item mods.
> 
> 
> 
> which is fun coz we don't want to accidently heal them when you gear swap. I suspect SE doesn't actually set the level but just dynamic add the stats equivilently.
> 
> 
> 
> we can see mobs lv in packet..wondering if we see trust lv there as well.

I thought that iLvl only applied onMobSpawn and requires you to release and recast in order to apply new levels?

neuromancerxi:
> I thought that iLvl only applied onMobSpawn and requires you to release and recast in order to apply new levels?

It's actually a real time adjustment based on your current ilvl, not just on spawn. You can change gear and see HP values adjust in real time in Retail.

zach2good:
https://www.bg-wiki.com/ffxi/Category:Trust

> Alter Egos will not automatically level up with you; this includes Alter
Egos summoned whilst under the effects of Level Sync
<https://www.bg-wiki.com/ffxi/Level_Sync>. As such, Alter Egos should be
released and re-summoned after gaining a level. This does not apply,
however, to aforementioned iLvl equipment; Alter Ego stats will
automatically change if your overall iLvl changes.

GodModeXI:
Apparently, I have some studying to do. XD thank you for the clarifications on those details.

TeoTwawki:
> I thought that iLvl only applied onMobSpawn and requires you to release and recast in order to apply new levels?

ironically their regular pre 99 levels seem to be this way, but not bonus levels which item levels in general seem to basically be

GodModeXI:
> (misclicked close from mobile)
> 
> 
> 
> For the gear and quest adjustments to trust levels, remember that trust stats change _in real time_ when you're using iLevel, so these should all be implemented using item mods. 

- bonus levels
  - gear should apply through item mods
  - KI could be applied as a bonus similarly to how the RoV KI apply exp bonuses in charutils.cpp?

src/map/utils/trustutils.cpp
```diff
@@ -565,13 +570,34 @@ namespace trustutils
             sJobStat = grade::GetHPScale(grade, baseValueColumn) + (grade::GetHPScale(grade, scaleTo60Column) * (sLvl - 1)) +
                        (grade::GetHPScale(grade, scaleOver30Column) * subLevelOver30) + subLevelOver30 + subLevelOver10;
         }
+        
+        if (mLvl >= 99 && mLvl < 105)
```
claywar: All of these conditions are exclusive, and should not be in separate statements, but instead if/else if.  Also, after the first check if using that pattern, the lower bound becomes unnecessary.

src/map/utils/trustutils.cpp
```diff
@@ -601,8 +627,28 @@ namespace trustutils
             grade    = grade::GetJobGrade(sJob, 1);
             sJobStat = grade::GetMPScale(grade, 0) + grade::GetMPScale(grade, scaleTo60Column);
         }
-
-        PTrust->health.maxmp = (int16)(settings::get<float>("map.ALTER_EGO_MP_MULTIPLIER") * (raceStat + jobStat + sJobStat));
+        
+        if (mLvl >= 99 && mLvl < 105)
```
claywar: Same as above

src/map/utils/trustutils.cpp
```diff
@@ -654,6 +700,61 @@ namespace trustutils
...
+            mSTR = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 2), mLvl) * 1.5;
+            mDEX = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 3), mLvl) * 1.5;
+            mVIT = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 4), mLvl) * 1.5;
+            mAGI = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 5), mLvl) * 1.5;
+            mINT = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 6), mLvl) * 1.5;
+            mMND = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 7), mLvl) * 1.5;
+            mCHR = mobutils::GetBaseToRank(grade::GetJobGrade(PTrust->GetMJob(), 8), mLvl) * 1.5;
+        }        
+        
+        if (mLvl >= 110 && mLvl < 115)
```
claywar: This condition duplicates the above, should be combined

src/map/utils/trustutils.cpp
GodModeXI: If (mlvl >= base)
{
    Do
}
else if (mlvl >= higher value)
{
    Do
}
Etc?

claywar: Actually, now that I think about it, invert the conditions.

if level >= upper condition

elseif level >= condition below that

etc

GodModeXI: This has been restacked upper conditions to lower conditions.

src/map/utils/trustutils.cpp
GodModeXI: corrected.

src/map/utils/trustutils.cpp
GodModeXI: adjusted the upper range to not be duplicitous.

GodModeXI:
I fixed the formatting issues then all the other checks started failing for this. I didn't make any changes to these particular lines. the variables are being used in the sJobStats calculations, so I am usure how to correct these.

image
image

GodModeXI:
> I fixed the formatting issues then all the other checks started failing for this. I didn't make any changes to these particular lines. the variables are being used in the sJobStats calculations, so I am usure how to correct these.
> 
> image image

Must have mistakenly removed the section with sLvl calculations when making edits from the reviews. Added back in what was missing and all checks passed.

GodModeXI:
Due to not having a good process to work in dynamic adjustments, I am converting this to a draft in hopes to figure out a method. The PR is becoming dated and It may be beyond my skill set at the moment. Hopefully some support could be added or someone could adopt it prior to tabling it for a revisit at a later time.

PR #2983, opened 2022-10-20
[codeql] Remove or update commented code

claywar:
**_I affirm:_**

Addresses a chunk of codeql warnings around commented code.  

No impact should be observed outside of some Trace-level logs being added

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -747,7 +747,7 @@ void CLuaBaseEntity::entityVisualPacket(std::string const& command, sol::object 
     CBaseEntity* PNpc = nullptr;
     if (entity != sol::lua_nil)
     {
-        CLuaBaseEntity* PLuaBaseEntity = entity.as<CLuaBaseEntity*>(); // nullptr;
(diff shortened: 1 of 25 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -2089,17 +2089,10 @@ bool CLuaBaseEntity::sendGuild(uint16 guildID, uint8 open, uint8 close, uint8 ho
...
     uint8 VanadielHour = (uint8)CVanaTime::getInstance()->getHour();
-    // uint8 VanadielDay = (uint8)CVanaTime::getInstance()->getWeekday();
 
     GUILDSTATUS status = GUILD_OPEN;
 
-    /*
-     * No more guild holidays since 2014
-    if (VanadielDay == holiday)
-    {
-        status = GUILD_HOLYDAY;
```
TeoTwawki, line 2100: on the fence about this commented out block - it's possible it's been saving us from having to explain how to do it with every new "oldschool" server that pops up :laughing:

claywar: Makes sense.  Restored the information that was removed in that function

TeoTwawki (approved):

PR #1752, opened 2022-05-19
[NPC] Purakoko pathing

kaincenteno:
adds 3 points that Purakoko uses for pathing

**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I've _**tested my code and the things my code has changed**_ since the last commit in the PR, and will test after any later commits

## What does this pull request do?

Purakoko is currently static. 
This PR added the pathing AI and her 3 points that she uses to move around.
pathing currently doesnt support waiting at point and/or rotating so added notes of her behaviour when this is implemented

PR #2988, opened 2022-10-20
Purakoko Pathing with BONUS!

jmcmorris:
**_I affirm:_**

Working version of this PR: https://github.com/LandSandBoat/server/pull/1752

- Added Purakoko pathing observed on retail: https://www.youtube.com/watch?v=jsz7TQjPp4Q
- Patrols use previous `x, y, z` but if none had been set before a point such as waiting with a fixed rotation at current position then the NPC would be warped to `-1, -1, -1`. Now these coords are defaulted to the NPCs position.
- Added `onPathPoint` which is very similar to `onPath` but waits until after the specified wait time.
- Added `onPathComplete` which is invoked when the pathing is completed after after the wait time.

`!pos -72.580 -10.500 115.877`

Watch Purakoko path around~

### DIFF
--- a/scripts/zones/Windurst_Walls/npcs/Purakoko.lua
+++ b/scripts/zones/Windurst_Walls/npcs/Purakoko.lua
@@ -5,6 +5,25 @@
 -----------------------------------
 local entity = {}
 
+local paths =
+{
+    { { x = -72.580, y = -10.500, z = 115.877, wait = 2000 } },
+    { { x = -65.567, y = -11.000, z = 120.000, wait = 2000 } },
+    { { x = -59.242, y = -12.500, z = 123.703, wait = 2000 } },
+    { { wait = 2000, rotation = 0 } },
+}
+
+entity.onSpawn = function(npc)
+    npc:initNpcAi()
+    npc:setPos(xi.path.first(paths[1]))
+    npc:pathThrough(paths[1], xi.path.flag.PATROL)
+end
+
+entity.onPathComplete = function(npc)
+    local index = math.random(1, #paths)
+    npc:pathThrough(paths[index], xi.path.flag.PATROL)
+end
+
 entity.onTrade = function(player, npc, trade)
 end
 
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
 ===========================================================================
 
   Copyright (c) 2010-2015 Darkstar Dev Teams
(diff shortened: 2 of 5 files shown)


claywar (approved):
Love the additional options

src/map/lua/luautils.cpp
```diff
@@ -1,4 +1,4 @@
-﻿/*
```
WinterSolstice8, line 1: did the BOM on this file get changed?

zach2good: They tend to flip flop all the time, I tend to ignore it now

jmcmorris: Honestly not sure what happened there.

zach2good: (I have a tool to detect this in tools/, but we aren't using it)

scripts/zones/Windurst_Walls/npcs/Purakoko.lua
```diff
@@ -5,6 +5,25 @@
...
+
+entity.onSpawn = function(npc)
+    npc:initNpcAi()
+    npc:setPos(xi.path.first(paths[1]))
+    npc:pathThrough(paths[1], xi.path.flag.PATROL)
+end
+
+entity.onPathComplete = function(npc)
+    local index = math.random(1, #paths)
+    npc:pathThrough(paths[index], xi.path.flag.PATROL)
```
zach2good, line 24: Nice way of handling this!

zach2good (approved):

src/map/lua/luautils.cpp
WinterSolstice8: looks like its already UTF-8-BOM on my copy so says notepad++ so this might be changing it back to UTF-8. barely matters for 99% of users anyway

scripts/zones/Windurst_Walls/npcs/Purakoko.lua
jmcmorris: Thanks! I ended up discovered that it was mostly "falling through" due to the -1 coord issue but it did have a little bit of fall through going from `paths[3]` to `paths[1]` but if I slightly moved the y up then it stopped.

WinterSolstice8 (approved):

src/map/lua/luautils.cpp
TeoTwawki: editor config as we have it set currently is _supposed_ to auto re-add the bom for cpp files but hell if I know if it actually is working. Even git itself reads and uses that so what editor was used shouldn't matter either, I think...?

oh well. We'll find out if some random person using an east Asian version of windows or visual studio pops in with the error.

PR #2991, opened 2022-10-21
[core] Add inactivity watchdog to kernel

zach2good:
**_I affirm:_**

If for some reason the main tick of the kernel (currently used in all processes) takes longer than 2000ms, something is very very wrong. You're either debugging something, or you've manufactured an infinite loop and are starving the process of the opportunity to do logic, service sockets, accept connections and keep client connections alive.

I was originally going to have this kill the process so that operators could have their stuff auto-restart, but that was clashing with attached debuggers etc.

So I've opted to just fire a one-off critical message. It'll fire when debugging, but that's not a big deal right?

Insert a `std::this_thread::sleep(n)` somewhere in native code, or attach a debugger, or make a fat loop somewhere in Lua and see the message get printed.


zach2good:
Updated ready for review

Xaver-DaRed (approved):

PR #3002, opened 2022-10-22
Aggroable Mobs

jmcmorris:
**_I affirm:_**

This adds support for mobs on the player allegiance to be aggroed by other mobs. This is needed for CoP 8-2 and very likely any other escort mission/quest.

How it works is that it checks every tick to see if a mob is aggroable and if so then looks to see if it should be aggroed. This should have very little performance impact but in the future this and a great deal of other similar checks could be optimized with using spatial hashing.

image

Go to any area with mobs that aggro and target a mob and input the following commands.
```
!exec target:setStatus(xi.status.NORMAL)
!exec target:setAllegiance(xi.allegiance.PLAYER)
!exec target:setIsAggroable(true)
```

This mob should now be aggroed by other mobs.
Siege_2022 10 21_193447

### DIFF
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -1310,7 +1310,8 @@ void CZoneEntities::ZoneServer(time_point tick, bool check_regions)
 
     luautils::OnZoneTick(this->m_zone);
 
-    EntityList_t::iterator it = m_mobList.begin();
+    std::vector<CMobEntity*> aggroableMobs;
+    EntityList_t::iterator   it = m_mobList.begin();
     while (it != m_mobList.end())
     {
         CMobEntity* PMob = dynamic_cast<CMobEntity*>(it->second);
@@ -1364,7 +1365,30 @@ void CZoneEntities::ZoneServer(time_point tick, bool check_regions)
             delete PMob;
             continue;
         }
+
+        if (PMob->allegiance == ALLEGIANCE_TYPE::PLAYER && PMob->m_isAggroable)
+        {
+            aggroableMobs.push_back(PMob);
+        }
+
         it++;
+    }
+
+    // Check to see if any aggroable mobs should be aggroed by other mobs
+    for (CMobEntity* PMob : aggroableMobs)
+    {
+        for (auto PMobCurrentIter : m_mobList)
+        {
+            CMobEntity* PCurrentMob = dynamic_cast<CMobEntity*>(PMobCurrentIter.second);
+            if (PCurrentMob != nullptr && PCurrentMob->isAlive() && PMob->allegiance != PCurrentMob->allegiance && distance(PMob->loc.p, PCurrentMob->loc.p) <= 50)
+            {
+                CMobController* PController = static_cast<CMobController*>(PCurrentMob->PAI->GetController());
+                if (PController != nullptr && PController->CanAggroTarget(PMob))
+                {
+                    PCurrentMob->PEnmityContainer->AddBaseEnmity(PMob);
+                }
+            }
+        }
     }
 
     for (EntityList_t::const_iterator it = m_npcList.begin(); it != m_npcList.end(); ++it)
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -119,6 +119,7 @@ CMobEntity::CMobEntity()
     m_TrueDetection = false;
     m_Detects       = DETECT_NONE;
     m_Link          = 0;
+    m_isAggroable   = false;
     m_battlefieldID = 0;
     m_bcnmID        = 0;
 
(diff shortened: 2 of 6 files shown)


src/map/zone_entities.cpp
```diff
@@ -1367,6 +1367,29 @@ void CZoneEntities::ZoneServer(time_point tick, bool check_regions)
...
+    // Check to see if any aggroable mobs should be aggroed by other mobs
+    for (auto PMobIter : m_mobList)
+    {
+        CMobEntity* PMob = dynamic_cast<CMobEntity*>(PMobIter.second);
+        if (PMob == nullptr || PMob->allegiance != ALLEGIANCE_TYPE::PLAYER || !PMob->m_isAggroable)
+        {
+            continue;
+        }
+
+        for (auto PMobCurrentIter : m_mobList)
```
claywar, line 1380: Do we really need to iterate over m_moblist twice, or could we apply the continue checks in one?

jmcmorris: I'm not sure how we could avoid iterating over it twice unless we did something like I mentioned in the PR where we would introduce spatial hashing to quickly retrieve nearby mobs and only iterate over those. All of the conditional checks in the inner for loop are necessary as they use `PCurrentMob`.

claywar: I guess I'm not understanding, as the outer and inner loop appear to be iterating over the same thing, except for the fact that the outer is continuing on some instances

jmcmorris: I'll break it down and if desired I can add in comments into the code. The outer loop iterates over all mobs looking for any mobs that are aggroable. If we find such a mob then we iterate over all other mobs looking to see if one of those inner loop mobs can aggro the mob from the outer loop.

KnowOne134:
Able to add ot to pets spawned? There is 1 assault where this Is needed

jmcmorris:
> Able to add ot to pets spawned? There is 1 assault where this Is needed

Which assault is that? I'd like to take a look to see how it works.

claywar (approved):

KnowOne134:
Escort professor something. T3 assault. Gimmick is so call out carby, cast sneak on you and all the mobs agro to carby as you walk by to clear out way for the escort

jmcmorris:
> Escort professor something. T3 assault. Gimmick is so call out carby, cast sneak on you and all the mobs agro to carby as you walk by to clear out way for the escort

We can definitely work on figuring that out. I should really do some more research on exactly which mobs (and pets) can be aggroed and under what conditions. Do mobs always aggro summons? is it all pets? do they aggro fellows? Do they follow the same /check aggro rules as players?

KnowOne134:
No. It's just in this 1 instance that I know of

src/map/zone_entities.cpp
```diff
@@ -1367,6 +1367,29 @@ void CZoneEntities::ZoneServer(time_point tick, bool check_regions)
...
+        {
+            CMobEntity* PCurrentMob = dynamic_cast<CMobEntity*>(PMobCurrentIter.second);
+            if (PCurrentMob->isAlive() && PMob->allegiance != PCurrentMob->allegiance && distance(PMob->loc.p, PCurrentMob->loc.p) <= 50)
```
zach2good: Should be checking if this mob is null also, since the outer loop could have only iterated to 10, but the inner loop might be at 100

src/map/zone_entities.cpp
zach2good: One way we can avoid this is to work in reverse

When we iterate all the mobs for updates on the main loops above, we can build a list of all the mobs who are aggro able. Then down here we can iterate over them and then ask the zone mob list to try and aggro them, unrolling this loop

zach2good: (but also +1 for some kind of spacial hashing, I know Eden split up a zone into 50" squares and only checks current and adjacent squares, but I'm also partial to quadtrees, which I'd love to implement)

src/map/zone_entities.cpp
jmcmorris: Good catch!

src/map/zone_entities.cpp
jmcmorris: Collecting aggroable mobs into a vector is a good idea. I went ahead and implemented that.

You're welcome to do the quadtree or whatever anytime you want. I've implemented spatial hashes before but haven't done it with quadtrees. I'm not sure what advantages you'd get over quadtrees as I think most if not all cells will be filled. Maybe if we are doing something like collision with them as well and wanted something smaller than 50".

src/map/zone_entities.cpp
```diff
@@ -1310,7 +1310,8 @@ void CZoneEntities::ZoneServer(time_point tick, bool check_regions)
 
     luautils::OnZoneTick(this->m_zone);
 
-    EntityList_t::iterator it = m_mobList.begin();
+    std::vector<CMobEntity*> aggroableMobs;
```
zach2good, line 1313: 👍

src/map/zone_entities.cpp
zach2good: I guess you're right, most of the saving will be splitting things out from the 50' draw distance. You can go ahead with the spatial hashing if you'd like 👍

zach2good (approved):

PR #3009, opened 2022-10-22
Quasilumin Escort for CoP 8-2

jmcmorris:
**_I affirm:_**

This implements the logic behind escorting the Quasilumin spawned at the Cermet Alcove's in Grand Palace of Hu'Xzoi. This was reasonably tricky to get right due as there are many moving parts including several that were not properly supported. Such as Aggroable Mobs in #3002 and mobs being treated more like NPCs (triggerable).

https://ffxiclopedia.fandom.com/wiki/A_Fate_Decided

```
!gotoid 16916927
!gotoid 16916928
!gotoid 16916929
!gotoid 16916930
```

Go to any of those Cermet Alcove's and interact with it to spawn the Quasilumin from it. You can trigger the Quasilumin to stop it and trigger it again to resume. Once it gets to the destination you can interact with it to open the final door.


scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
```diff
@@ -0,0 +1,307 @@
...
+
+local escortProgress =
+{
+    NONE     = 0, -- Have not started pathing at all yet
+    ENROUTE  = 1, -- Started pathing and is continuing along path
+    PAUSED   = 2, -- Has paused pathing and awaiting orders
+    COMPLETE = 3, -- Has reached the end of the escort mission
+}
+
+local this = {}
```
zach2good: `entity`

scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
```diff
@@ -0,0 +1,307 @@
+-----------------------------------
```
zach2good, line 1: Can we move the contents of this file into a helper in this zones folder, or into the triggering zone or npc lua?

scripts/zones/Grand_Palace_of_HuXzoi/IDs.lua
```diff
@@ -20,26 +20,33 @@ zones[xi.zone.GRAND_PALACE_OF_HUXZOI] =
...
-        [16916898] = 7370,
-        [16916899] = 7371,
-        [16916900] = 7372,
-        [16916901] = 7373,
-        [16916902] = 7374,
-        [16916903] = 7375,
-        [16916904] = 7376,
-        [16916905] = 7377,
-        [16916906] = 7378,
+        [16916897] = 7380,
```
zach2good: Eventually, we'll need/want to replace all of these hardcoded ids with some lookups. But in the in interest of getting COP 'complete' we can ignore it for now

scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
```diff
@@ -0,0 +1,307 @@
...
+            mob:showText(mob, ID.text.PATROL_SUSPENDED)
+            mob:setLocalVar("progress", escortProgress.PAUSED)
+        elseif progress == escortProgress.PAUSED then
+            mob:showText(mob, ID.text.RECOMMENCING_PATROL)
+            mob:setLocalVar("progress", escortProgress.ENROUTE)
+            mob:pathThrough(data.path[point], xi.path.flag.WALK)
+        elseif progress == escortProgress.COMPLETE then
+            mob:showText(mob, ID.text.DUTY_COMPLETE)
+            player:messageSpecial(data.direction)
+            -- TODO: display animation and NPC is not pushing update packet to players in range
```
zach2good, line 285: Is this important?

scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
zach2good: Failing that, some comments explaining this is a dynamic entity and this is the naming convention would be welcome. Its a little arcane (my bad)

zach2good (approved):
Looking good 👍

scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
jmcmorris: I _think_ I was the original person who wrote that comment but it was nearly 3 years ago and I honestly don't remember what it is talking about. I thought about deleting it but thought maybe someone else might know haha

scripts/zones/Grand_Palace_of_HuXzoi/IDs.lua
jmcmorris: They're sequential so we could technically replace it with just one "offset" and calculate it from there. `textID = textOffset + (mob:getID() - quasiluminOffset)`

scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
jmcmorris: Sure I don't mind doing either of these. Can you explain the motivation so I can understand better? Thanks!

src/map/packets/entity_update.cpp
```diff
@@ -82,7 +82,7 @@ void CEntityUpdatePacket::updateWith(CBaseEntity* PEntity, ENTITYUPDATE type, ui
             }
             if (PEntity->spawnAnimation == SPAWN_ANIMATION::SPECIAL)
             {
-                ref<uint8>(0x28) |= 0x45;
+                ref<uint8>(0x28) |= 0x04;
```
WinterSolstice8, line 85: 0x04 for Quasilumins I presume?

This is an interesting thing that needs to be configurable. Currently, Dragoon Wyverns always use the same "come from the sky" animation upon spawning in range (at 50'+) even if it's not their initial spawn. The intent of this "SPAWN_ANIMATION::SPECIAL" currently was to _specifically_ override the spawn animation for dynamic entities.

I know it's out of scope of this PR, but it might be good to convert this to a mobmod so we can use any arbitrary uint16 value for spawns to accomplish animations that may change.

I could make a separate PR to add that functionality to make Dragoon's Wyverns perform properly which will also allow you to select the proper spawn animation for what you need here, too.

TeoTwawki: many entities do this with different animations. even npcs - the synergy furnaces have a spawn in animation that is different from the standard fade in even, tho you usually will not get close enough to them in time to notice it unless you lag a bit.

There are possibly enough entities needing a non default one that it may eventually need to be stored in the db columns, (let us hope not) but a mobmob would be a good start for now.

jmcmorris: The entire `0x28` bitfield needs to be figured out but I'm fairly sure that `0x04` bit is more or less "play special animation on spawn".

TeoTwawki: oh ok so it's not that multiple animations can fill the byte (so far as we know) it's a flag that uses whatever alternate the model has baked in. that makes sense.

jmcmorris: I was able to figure out a few of the bits such:

- `0x01` seems to prevent entity trigger even on green name (maybe something else but this was a side-effect)
- `0x04` play special spawn (and other?) animations
- `0x40` make the entity triggerable on client

jmcmorris: We have a few others in the code already too:

- ref<uint8>(0x28) |= PMob->StatusEffectContainer->HasStatusEffect(EFFECT_TERROR) ? 0x10 : 0x00;
- ref<uint8>(0x28) |= PMob->health.hp > 0 && PMob->animation == ANIMATION_DEATH ? 0x08 : 0;

WinterSolstice8: It doesn't seem that the change here affects dynamic entities in any way. Tested with NPC module + !fafnir

WinterSolstice8 (approved):

jmcmorris:
Still planning on cleaning things up per @zach2good feedback but was waiting for his response first. I know he's busy so I'm just chilling for now. :)

scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
jmcmorris: @zach2good think you could go into some details? :)

zach2good: I don't think the average person poking around in scripting is going to know that these Quasis are DE's, or necessarily what a DE is or does. The name is defined here:

https://github.com/LandSandBoat/server/blob/base/src/map/lua/lua_zone.cpp#L272

I think attaching the logic at the point of creation, or adding some very clear comments about why this file exists and is named the way it is would help out anyone whos poking around in here

scripts/zones/Grand_Palace_of_HuXzoi/IDs.lua
zach2good: +1 for offset

scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
zach2good: Worth leaving it then 👍

scripts/zones/Grand_Palace_of_HuXzoi/IDs.lua
zach2good: The main fear is: If SE decides to put a JP or exemplar camp in Grand Palace, or does _anything else_ to the entity list, these IDs might shift and all of this might break. So looking up using names or offsets from other mobs is a safer way of handling this.

scripts/zones/Grand_Palace_of_HuXzoi/mobs/DE_Quasilumin.lua
jmcmorris: Thanks! 👍

jmcmorris:
Much like earlier, this PR seems to be stuck in some weird limbo. I'm going to try converting it into a draft and then back and see if that'll force it to detect the changes on my branch.

jmcmorris:
No luck. I'm not sure if it is Github or this PR. I'll give it until tomorrow morning and if it is still not updating then I'll remake the PR.

TeoTwawki:
> No luck. I'm not sure if it is Github or this PR. I'll give it until tomorrow morning and if it is still not updating then I'll remake the PR.

Githubs having issues, hang tight

PR's are technically also issues you see - its the same backend

Even if that weren't the case, whenever one service is acknowledged as having a problem, a bunch of others are also having problems without a word on the status page about it.

TeoTwawki:
PR's are technically also issues you see - its the same backend

jmcmorris:
Closing this as github seems to have screwed it up

PR #3019, opened 2022-10-23
[ci] Require padding around comparators and operators

claywar:
**_I affirm:_**

* Adds CI check to require spaces around all operators except for `-` (This will require future effort to determine if it is a negative number or operator)
* Updates all operators checked above to comply with new CI rule

No impact should be observed, CI should pass.


WinterSolstice8 (approved):
🤞🤞🤞

Three times for good measure...

jmcmorris (approved):

PR #3023, opened 2022-10-23
Kaiser Behemoth cast Meteor in Apollyon NW

jmcmorris:
**_I affirm:_**

Fixes issues with Kaiser Behemoth in Apollyon NW not casting Meteor correctly

```
!addkeyitem red_card
!addkeyitem cosmo_cleanse
!pos -600 -0.5 -600 38
```

Once inside Apollyon NW use `gotoid 16932985` and engage the Kaiser Behemoth and it'll cast Meteor now.

### DIFF
--- a/sql/mob_groups.sql
+++ b/sql/mob_groups.sql
@@ -1525,7 +1525,7 @@ INSERT INTO `mob_groups` VALUES (23,2661,38,'Millenary_Mossback',0,128,2858,1900
 INSERT INTO `mob_groups` VALUES (24,202,38,'Apollyon_Scavenger',0,128,2858,4750,0,75,77,0);
 INSERT INTO `mob_groups` VALUES (25,883,38,'Cynoprosopi',0,128,2858,15000,0,85,85,0);
 INSERT INTO `mob_groups` VALUES (26,1771,38,'Gorynich',0,128,2862,5000,0,79,79,0);
-INSERT INTO `mob_groups` VALUES (27,6732,38,'Kaiser_Behemoth',0,128,2860,15000,0,85,85,0);
+INSERT INTO `mob_groups` VALUES (27,6732,38,'Kaiser_Behemoth',0,128,2860,15000,1000,85,85,0);
 INSERT INTO `mob_groups` VALUES (28,2289,38,'Kronprinz_Behemoth',0,128,2863,7000,1000,78,78,0);
 INSERT INTO `mob_groups` VALUES (29,1515,38,'Ghost_Clot',0,128,2856,10000,0,82,83,0);
 INSERT INTO `mob_groups` VALUES (30,2634,38,'Metalloid_Amoeba',0,128,2856,7500,0,80,80,0);
--- a/scripts/battlefields/Apollyon/nw_apollyon.lua
+++ b/scripts/battlefields/Apollyon/nw_apollyon.lua
@@ -370,6 +370,7 @@ content.groups =
         {
             [xi.mobMod.ALLI_HATE] = 50,
             [xi.mobMod.MAGIC_COOL] = 30,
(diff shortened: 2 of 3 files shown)


claywar (approved):

sql/mob_groups.sql
```diff
@@ -1525,7 +1525,7 @@ INSERT INTO `mob_groups` VALUES (23,2661,38,'Millenary_Mossback',0,128,2858,1900
 INSERT INTO `mob_groups` VALUES (24,202,38,'Apollyon_Scavenger',0,128,2858,4750,0,75,77,0);
 INSERT INTO `mob_groups` VALUES (25,883,38,'Cynoprosopi',0,128,2858,15000,0,85,85,0);
 INSERT INTO `mob_groups` VALUES (26,1771,38,'Gorynich',0,128,2862,5000,0,79,79,0);
-INSERT INTO `mob_groups` VALUES (27,6732,38,'Kaiser_Behemoth',0,128,2860,15000,0,85,85,0);
+INSERT INTO `mob_groups` VALUES (27,6732,38,'Kaiser_Behemoth',0,128,2860,15000,1000,85,85,0);
```
TeoTwawki, line 1528: I've actually been wondering why old DSP has mutliple meteor casting behemoths in mob_pools as war/war anyway. I kinda doubt any research was done on if they have any magic attack bonus or their total int math'd out to war/war and not war/blm etc

_(its def not blm/war, because that would behave differently)_ 


This block of script:
```lua
entity.onSpellPrecast = function(mob, spell)
    if spell:getID() == 218 then
        spell:setAoE(xi.magic.aoe.RADIAL)
        spell:setFlag(xi.magic.spellFlag.HIT_ALL)
        spell:setRadius(30)
        spell:setAnimation(280)
        spell:setMPCost(1)
    end
end
```
exists in both Kaiser and King behemoth. 

I think we should merge this PR, but then open an issue requesting additional research.

TeoTwawki (approved):

sql/mob_groups.sql
jmcmorris: I looked for their jobs and it is unknown EXCEPT that the monstrosity jobs are known.

https://www.bg-wiki.com/ffxi/Behemoth_(MON)

I thought about making it DRK/THF but it doesn't seem to have the Elasmoth specific abilities and I don't know if monstrosity jobs are the same as normal mob jobs (I think they are?).

The unfortunate reality is that a HUGE amount of monster jobs/subjobs are wrong and need to be combed through.

TeoTwawki: Monstrosity we have to be careful about fooling ourselves: mobs are _built_ similarly but we can easily trick ourselves into thinking every mob is like one example we see, or that all rules are the same when we don't know that.

But it's a useful datapoint often.

TeoTwawki: example: remember when SE updated the power of direct damage spells? Mobs still use the old formula. if we compare a player who happens to be in monstrosity mode, which formula is it? if we assume and don't know we're incorrect and then try to figure out a mobs nuke damage based on that, we're at high risk of getting that mobs data incorrect.

WinterSolstice8: SE also didn't update mob TP formula. They appear to use the very original formula adjusted for base 1000. We could use that to verify if SE did shenanigans for monstrosity

TeoTwawki: > SE also didn't update mob TP formula. They appear to use the very original formula adjusted for base 1000. We could use that to verify if SE did shenanigans for monstrosity

that would only tell us about tp - it would hint that maybe other systems matched, but it would still be _possible_ that one system treated us as mobs and another treated us as players with a mobs stats (and bolt-on equipped additions).

ya never really know till ya know, ya know? :wink:

PR #3028, opened 2022-10-24
Mob modifier for detection

jmcmorris:
**_I affirm:_**

Add in a mobmod for detection as the only way to set detection right now is on a mob family in the database. In order to avoid creating a new mob family I opted to add a mobmod that enables this detection override.

Along with this change is an example of where it can be used. I've also included a tiny semi-related change with the same mobs making them immune to sleep.

Related capture: https://youtu.be/CNJNBhTBse8?t=701

Fixes: https://github.com/HorizonFFXI/HorizonXI-Issues/issues/295

```
!addkeyitem black_card
!addkeyitem cosmo_cleanse
!pos 600 -0.5 -600 38
```

Go into Apollyon NE and once in type `!cs 214` to skip to the second floor. On the second floor approach the giant birds from behind and it should aggro via true hearing.


scripts/globals/status.lua
```diff
@@ -2470,7 +2470,7 @@ xi.mobMod =
     SEVERE_SPELL_CHANCE = 13, -- % chance to use a severe spell like death or impact
     SKILL_LIST          = 14, -- uses given mob skill list
     MUG_GIL             = 15, -- amount gil carried for mugging
-    -- 16 Available for use
+    DETECTION           = 16, -- Overrides mob family's detection method. In order to set to override to none an unused bit must be set such as DETECT_NONE1.
```
WinterSolstice8, line 2473: +1 for backfilling old mods

WinterSolstice8 (approved):
Super not sure why this wasn't a thing before!

PR #3036, opened 2022-10-25
Battle ID to isolate fights

jmcmorris:
**_I affirm:_**

This introduces a battle ID on the entities which only allows battle interactions to occur between entities with the same battle ID. This can be used for an assortment of content such as Garrison, Pirate's Chart Quest, and anywhere we don't want battle interactions to occur.

Target any mob and input `!exec target:setBattleID(1)` and that mob will no longer be interactable. You can do the same with yourself and then you will be able to interact with only that mob and no other mobs. Use `!exec target:setBattleID(0)` to reset to default.

### DIFF
--- a/sql/mob_skills.sql
+++ b/sql/mob_skills.sql
@@ -1364,7 +1364,7 @@ INSERT INTO `mob_skills` VALUES (1527,1086,'laser_shower',4,10.0,2000,1500,4,0,0
 INSERT INTO `mob_skills` VALUES (1528,1087,'floodlight',2,15.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (1529,1089,'hyper_pulse',1,10.0,2000,1500,4,0,0,0,0,0,0); -- Proto-Omega
 INSERT INTO `mob_skills` VALUES (1530,1088,'stun_cannon',4,20.0,2000,1500,4,0,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (1531,772,'wz_recover_all',1,20.0,3000,0,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (1531,772,'wz_recover_all',1,20.0,0,0,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (1532,1124,'pod_ejection',0,7.0,4500,1,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (1533,1117,'pile_pitch',0,10.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (1534,1118,'guided_missile',2,5.0,2000,1500,4,0,0,0,0,0,0);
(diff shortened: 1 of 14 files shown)


kaincenteno:
ooo nice! this is like they do with those unity nms right? :D

jmcmorris:
I'm wondering if this is something that should be persisted if you logout. If so then in what scenarios.

jmcmorris:
> ooo nice! this is like they do with those unity nms right? :D

That's right!

jmcmorris:
Moving to a draft as I'm investigating the exact desired behavior with logging in/out and zoning in/out.

sql/mob_skills.sql
```diff
@@ -1364,7 +1364,7 @@ INSERT INTO `mob_skills` VALUES (1527,1086,'laser_shower',4,10.0,2000,1500,4,0,0
 INSERT INTO `mob_skills` VALUES (1528,1087,'floodlight',2,15.0,2000,1500,4,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (1529,1089,'hyper_pulse',1,10.0,2000,1500,4,0,0,0,0,0,0); -- Proto-Omega
 INSERT INTO `mob_skills` VALUES (1530,1088,'stun_cannon',4,20.0,2000,1500,4,0,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (1531,772,'wz_recover_all',1,20.0,3000,0,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (1531,772,'wz_recover_all',1,20.0,0,0,4,0,0,0,0,0,0);
```
Xaver-DaRed, line 1367: What is this change?

jmcmorris: Limbus recover Armoury Crates are mobs on the mob allegiance and that skill, `wz_recover_all`, targets the players and heals them. I noticed there was a huge delay and looked on retail and noticed that it shouldn't have any delay. The reason I included it here is because in order for recover Armoury Crates to target the player they have to be in the same battle ID which means that technically they can be hit by AoE spells when activated. I reduced the delay in hopes to reduce the likelihood of this happening.

There is quite possibly a better solution. One thought is to add a special targeting type on skills that ignore battleID. Another would be to add a new flag that says the entity cannot be targeted at all.

TeoTwawki: I'd go with editing the target flags to have one for it, but there might be a flag already to prevent targetting. I wanna say there is and I'd seen it used for that one abyssea NM (big clot type slime) but I am unsure what it was enum'd as to direct you to it.

TeoTwawki: actually iirc that flag was bugged and ppl coudl cheat with aoes anyway..fixing it may be the answer

jmcmorris:
Meloetta verified on retail that the level sync status effect does not persist on logout. Upon logging back in the effect is gone and the player can no longer participate in Garrison. Consequently it makes sense that the battle ID is not persisted.

Xaver-DaRed (approved):

Issue #2926, opened 2022-10-12
🔨 Trust: Create Shield Block function


GodModeXI:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Describe the feature

- Trust shield block enhancement.
- Observations through discussion:
- Through retail observation and captures by Strong/Neckbeard. Specific trusts lack the logic to perform shield block procs. 
![image](https://user-images.githubusercontent.com/99501957/195351403-f3128145-3cfb-46e4-b878-31b66760355c.png)
![image](https://user-images.githubusercontent.com/99501957/195351445-98792d7a-b71e-45c4-8f5b-9c8e61d7203d.png)
![image](https://user-images.githubusercontent.com/99501957/195351475-ee83aca9-b224-4829-a951-3c07d94208c9.gif)
- Battleutils: isBlocked, as observed by @TeoTwawki has logic for PC object types, and a messy copy for PET objects, but lacks logic for TRUST objects all together.
- (https://github.com/LandSandBoat/server/blob/e7ef57e59458fec605840b21c52636f4072f731a/src/map/utils/battleutils.cpp#L2070)
- Battleultis: GetBlockRate, as observed by @TeoTwawki contains logic and formula calculations for shield block rates for PC / PET / MOB type objects but lacks logic for TRUST objects all together.
- (https://github.com/LandSandBoat/server/blob/e7ef57e59458fec605840b21c52636f4072f731a/src/map/utils/battleutils.cpp#L1786)

- Request for enhancement:
- clean up logic for non PC object types for shield block rates that do not consider shield mastery, or shield item slot. A separate exception formula should be generated to override shield size / mastery considerations and create a new calculation to emulate a formula that applies to non PC object types in a similar way.

TeoTwawki:
> Battleutils: isBlocked, as observed by @TeoTwawki has logic for PC object types, 

isBlocked is ok, places its called not so much

the pet stuff was inside the shield mastery block of takeDamage, and I am not sure any pet ever actually has that trait

start of discord convo in case reference is needed: [link](https://discord.com/channels/392903136336936960/832349765718900827/1029461068969422879)

PR #3037, opened 2022-10-25
[core] {trust} add shield base block exception / add reprisal sub eff…

GodModeXI:
…ect exception / add shield skill calculation / adjust reprisal effect to generate spike damage

@Schaedelschaden supported with printf audit and debugging 

**_I affirm:_**

- This PR corrects issue #2926 
- Adds a shield base rate to trusts bypassing shield size calculation (since trusts have no gear)
- Adds in Shield skill to assist in stat calculations.
- Adds subeffect functionality for Reprisal for trusts that bypasses the shield check logic
- Adds a power calculation to the Reprisal effect lua to not return 0 damage. 
- Audit Data
blockrates
    - Block Rates
        - 50 = guaranteed 5% + 45% base applied in Core
        - 97 = base 50 + Reprisal calculations
        - 80 = base 50 + Palisade calculations

reprisal no dmg
    - Reprisal : No spike damage

reprisal with damage
    - Reprisal : with effect power added

1. Load in core change files and reprisal effect lua
2. Clean and rebuild server to apply changes.
3. load a PLD trust with palisade / reprisal logic applied.
4. add a printf() into battleutils.cpp before the return to show block rates.


src/map/utils/trustutils.cpp
```diff
@@ -693,6 +693,15 @@ namespace trustutils
             }
         }
 
+        if (mJob == JOB_PLD)
+        {
+            uint16 maxSkill = battleutils::GetMaxSkill((SKILLTYPE)30, mJob, mLvl > 99 ? 99 : mLvl);
```
WinterSolstice8: SKILLTYPE::SKILL_SHIELD ?

src/map/utils/trustutils.cpp
```diff
@@ -693,6 +693,15 @@ namespace trustutils
...
+            if (maxSkill != 0)
+            {
+                PTrust->WorkingSkills.skill[30] = static_cast<uint16>(maxSkill * settings::get<float>("map.ALTER_EGO_SKILL_MULTIPLIER"));
```
WinterSolstice8: `PTrust->WorkingSkills.skill[SKILLTYPE::SKILL_SHIELD] =` ?

scripts/globals/effects/reprisal.lua
```diff
@@ -8,7 +8,7 @@ local effectObject = {}
 effectObject.onEffectGain = function(target, effect)
     target:addMod(xi.mod.SPIKES, 6)
     -- Spike damage is calculated on hit in battleutils::TakePhysicalDamage
-    target:setMod(xi.mod.SPIKES_DMG, 0)
+    target:setMod(xi.mod.SPIKES_DMG, effect:getPower())
```
WinterSolstice8: why was this needed?

src/map/utils/battleutils.cpp
```diff
@@ -898,6 +898,10 @@ namespace battleutils
                     {
                         PAttacker->takeDamage(spikesDamage, PDefender, ATTACK_TYPE::MAGICAL, DAMAGE_TYPE::LIGHT);
                     }
+                    else if (PDefender->objtype == TYPE_TRUST && PDefender->GetMJob() == JOB_PLD)
+                    {
+                        Action->spikesEffect = (SUBEFFECT)6;
```
WinterSolstice8: SUBEFFECT::SUBEFFECT_REPRISAL

src/map/utils/battleutils.cpp
```diff
@@ -1825,6 +1829,10 @@ namespace battleutils
                 return base + skillModifier;
             }
         }
+        else if (PDefender->objtype == TYPE_TRUST && PDefender->GetMJob() == JOB_PLD)
+        {
+            base = 45;
```
WinterSolstice8: can we put a TODO here to verify if all PLD trusts have the same base block rate
They probably do, though.

WinterSolstice8 (changes requested):

src/map/utils/trustutils.cpp
GodModeXI: Yes. 30 is SKILL_SHIELD

scripts/globals/effects/reprisal.lua
GodModeXI: Spikes returned 0 damage regardless of what was being fought. Adding in the effect power returned damage values.

src/map/utils/battleutils.cpp
GodModeXI: Sure can

scripts/globals/effects/reprisal.lua
WinterSolstice8: Hmm. Smells like there was a PC specific check, perhaps. It should be calculated elsewhere. Can't look into it right now but will check it out

src/map/utils/trustutils.cpp
GodModeXI: This is adjusted.

src/map/utils/trustutils.cpp
GodModeXI: Adjusted

src/map/utils/battleutils.cpp
GodModeXI: Adjusted

src/map/utils/battleutils.cpp
GodModeXI: Added comment for follow up on retail block rate behaviors across the different trusts.

scripts/globals/effects/reprisal.lua
GodModeXI: I'm not sure, I went through the damage calculations for spike damage in Battleutils, tried a few variations of the subeffect exception but no luck with any of the adjustments. They are blocking and mitigation of incoming damage is there, even the reprisal animations but there's no spike damage processing.

TeoTwawki: `Action->spikesParam;` is where the spikes damage normally lived. this feeds into `CalculateSpikeDamage()` where modifications happen.

TeoTwawki: no param value, no dmg

GodModeXI: https://github.com/LandSandBoat/server/blob/dcb1b856a7166a99960a4efbcca7b9b4c8b36c14/src/map/utils/battleutils.cpp#L792

Looks like the math is pulling calculations off a gear slot... may need a trust exception for the main slot. Or I'm reading this entirely wrong. 😜

src/map/utils/battleutils.cpp
```diff
@@ -898,6 +898,10 @@ namespace battleutils
                     {
                         PAttacker->takeDamage(spikesDamage, PDefender, ATTACK_TYPE::MAGICAL, DAMAGE_TYPE::LIGHT);
                     }
+                    else if (PDefender->objtype == TYPE_TRUST && PDefender->GetMJob() == JOB_PLD)
```
TeoTwawki: Both the blocks checking type trust and job should likely instead just check for a mod we set on the trust.

I'm sure there will be cases on non player entities doing block animations that won't always be reprisal paladins, and while we may not have a `reaction` to check against surely there is something we can when they animate a block.

src/map/utils/trustutils.cpp
```diff
@@ -693,6 +693,15 @@ namespace trustutils
             }
         }
 
+        if (mJob == JOB_PLD)
```
TeoTwawki, line 696: No good. we have several pld trust that do not block (no shield in model). this isn't going to fly unless you prove to me the spear wielding paladins (Halver, Exen) and Monberaux (also a pld!) are secretly blocking invisibly.

GodModeXI: That is a very valid and completely overlooked complication.

TeoTwawki: Suggestion: instead of setting their blockrate statically to 45, use a modifier (and can throw the 45 to it by default), then check that the modifier is non-zero here.

There may already exist one intended for a bonus on players that might be usable this way just on trusts.

GodModeXI: Mob:setMod(xi.mod.SHIELDBLOCKRATE) is an existing mod we could use.

TeoTwawki: and only used in 2 spots both in battleutils. one in GetBlockRate() and one in a check specifically for pets (automaton). the auto has a similar problem: its shield is fake. So this seems perfect.

GodModeXI: Ok, I'll test using the mod instead this evening and adjust the PR based on results.

scripts/globals/effects/reprisal.lua
WinterSolstice8: https://github.com/LandSandBoat/server/blob/082fcf03de3992125475c9bd1af35a1047ea909e/src/map/utils/battleutils.cpp#L2065
probably need a trust exception for a trust with shield, this is also where reprisal spikes dmg is being calculated

GodModeXI: I can take a swing at revising this to allow PET and TRUSTs to bypass the equipment check which should make the mod work for them and apply phantom blocks.

src/map/utils/trustutils.cpp
GodModeXI: what I messed with yesterday. applies block, applies spikes. open to ideas / changes

is blocked, adjusted whole block to bypass the gear check and then fall through reprisal

```
if (isBlocked)
            {
                int16 blockRateMod = PDefender->getMod(Mod::SHIELDBLOCKRATE);
                uint8 absorb = 100;
                if (PDefender->m_Weapons[SLOT_SUB]->IsShield())
                {
                    if (PDefender->objtype == TYPE_PC)
                    {
                        absorb = std::clamp(100 - PDefender->m_Weapons[SLOT_SUB]->getShieldAbsorption(), 0, 100);
                        absorb -= PDefender->getMod(Mod::SHIELD_DEF_BONUS); // Include Shield Defense Bonus in absorb amount

                        // Shield Mastery
                        if ((std::max(damage - (PDefender->getMod(Mod::PHALANX) + PDefender->getMod(Mod::STONESKIN)), 0) > 0) &&
                            charutils::hasTrait((CCharEntity*)PDefender, TRAIT_SHIELD_MASTERY))
                        {
                            // If the player blocked with a shield and has shield mastery, add shield mastery TP bonus
                            // unblocked damage (before block but as if affected by stoneskin/phalanx) must be greater than zero
                            PDefender->addTP(PDefender->getMod(Mod::SHIELD_MASTERY_TP));
                        }
                    }
                }
                else if (PDefender->objtype == TYPE_PET && blockRateMod > 0)
                {
                    absorb = 50;

                    // Shield Mastery
                    if ((std::max(damage - (PDefender->getMod(Mod::PHALANX) + PDefender->getMod(Mod::STONESKIN)), 0) > 0) &&
                        (PDefender->getMod(Mod::SHIELD_MASTERY_TP)))
                    {
                        // If the player blocked with a shield and has shield mastery, add shield mastery TP bonus
                        // unblocked damage (before block but as if affected by stoneskin/phalanx) must be greater than zero
                        PDefender->addTP(PDefender->getMod(Mod::SHIELD_MASTERY_TP));
                    }
                }
                else if (PDefender->objtype == TYPE_TRUST && blockRateMod > 0)
                {
                    absorb = 50;

                    // Shield Mastery
                    if ((std::max(damage - (PDefender->getMod(Mod::PHALANX) + PDefender->getMod(Mod::STONESKIN)), 0) > 0) &&
                        (PDefender->getMod(Mod::SHIELD_MASTERY_TP)))
                    {
                        // If the player blocked with a shield and has shield mastery, add shield mastery TP bonus
                        // unblocked damage (before block but as if affected by stoneskin/phalanx) must be greater than zero
                        PDefender->addTP(PDefender->getMod(Mod::SHIELD_MASTERY_TP));
                    }
                }
                else
                {
                    absorb = 50;
                }

                // Reprisal
                if (damage > 0 && PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_REPRISAL))
                {
                    // Reflect a portion of the blocked damage back. This is calculated before Stoneskin, Phalanx, Sentinel or Invincible
                    CStatusEffect* reprisalEffect = PDefender->StatusEffectContainer->GetStatusEffect(EFFECT_REPRISAL);

                    if (reprisalEffect != nullptr)
                    {
                        float spikesBonus   = 1.f + (PDefender->getMod(Mod::REPRISAL_SPIKES_BONUS) / 100.f);
                        int16 effectPower   = (int16)(reprisalEffect->GetPower() * spikesBonus);
                        int32 blockedDamage = (damage * (100 - absorb)) / 100;
                        int32 spikesDamage  = 0;

                        if (PDefender->StatusEffectContainer->HasStatusEffect({ EFFECT_INVINCIBLE, EFFECT_SENTINEL }))
                        {
                            blockedDamage = (baseDamage * (100.f - absorb)) / 100.f;
                        }

                        spikesDamage = blockedDamage * (effectPower / 100.f);

                        // Set Reprisal spike damage
                        PDefender->setModifier(Mod::SPIKES_DMG, spikesDamage);
                    }
                }
                damage = (damage * absorb) / 100;
            }
```
for get block rate. only falls through block calculations if the mod is on.

```
else if (PDefender->objtype == TYPE_TRUST && PDefender->getMod(Mod::SHIELDBLOCKRATE) > 0)
        {
            base = PDefender->getMod(Mod::SHIELDBLOCKRATE);
        }
```

for reprisal effect, only applies reprisal spikes if mod is present

```
else if (PDefender->objtype == TYPE_TRUST)
        {
             if (PDefender->getMod(Mod::SHIELDBLOCKRATE) > 0)
             {
                  Action->spikesEffect = SUBEFFECT::SUBEFFECT_REPRISAL;
             }
```

GodModeXI: New changes apply flat block rate controlled by mob:setMod(xi.mod.SHIELDBLOCKRATE, num)

If no mod, no block
If no mod, no reprisal

image

WinterSolstice8: I think we should not re-use the block rate mod for this and add 2 new ones:

a mod to indicate that the entity has a shield
a mod to indicate the base block rate when the above mod is non-zero

This way we dont have to touch it again in the future if we figure out that certain trusts have different base block rates

The block rate mod is used to indicate how much more block rate you should get from gear/buffs, and it modifies block rate in a way that is non-linear

GodModeXI: So we need to create a whole new mobmod that toggles true/false for having a shield. 

Would we want to set block rate using the modifier value as well like I have here so that base rates can be manually adjusted in the Lua in the event they are variable by trust?

TeoTwawki: I nominate backfilling currently unused mobmod ID 19 (since the mod won't even be on players, it makes sense as a mobmod rather than a regular mod_

GodModeXI: Making sure I have correct thinking,

Call it something like phantomshield then explain it in comments?

Mobmodifier.h - add 19, phantomshield
Status.lua global- add 19, phantomshield

xi.mobmod.PHANTOMSHIELD

Then swap out SHIELDBLOCKRATE with that modifier in Battleutils.cpp?

TeoTwawki: no to "phantom" anything. `MOBMOD_CAN_SHIELD_BLOCK` works

GodModeXI: Lol ok sorry XD
Will adjust.

TeoTwawki: a separate regular modifier would determine base black rate.

@WinterSolstice8 I don't think he has to add 2 mods for this, just the one for the "it has a shield/can block" part. The existing one is only used by pup automatons currently and appears to simply be its base blockrate since there is no shield "size" available.

I think we want that mobmod tho just to make it clear whats happening

WinterSolstice8: Sounds good to me if that's the case, I was probably conflating it with block rate +% gear base on its name alone

GodModeXI: So we would still use the SHIELDBLOCKRATE to set the base rate via lua and the mobmod to toggle shield exists or not?

TeoTwawki: > Sounds good to me if that's the case, I was probably conflating it with block rate +% gear base on its name alone

it is on some gear, however that gear can't affect non-players at all. so the automaton is using it to give it a non-zero block rate. in the existing code it's getting 45 of that mod when it valoredge form. its not a percent on top of base - it is the base percent rate blocking happens for its nonexistent "shield" (and why bart keeps wanting to say phantom xD).

src/map/utils/battleutils.cpp
```diff
@@ -898,6 +898,13 @@ namespace battleutils
                     {
                         PAttacker->takeDamage(spikesDamage, PDefender, ATTACK_TYPE::MAGICAL, DAMAGE_TYPE::LIGHT);
                     }
+                    else if (PDefender->objtype == TYPE_TRUST)
```
WinterSolstice8: This block shouldn't be necessary anymore due to the change in `TakePhysicalDamage` go ahead and try without it

GodModeXI: Confirmed. Staging reversion.

image

src/map/utils/battleutils.cpp
```diff
@@ -1825,6 +1826,22 @@ namespace battleutils
...
+                if (PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_REPRISAL))
+                {
+                    base = base * 1.5f;
+                }
+                return base + palisadeMod;
+            }
+            else
+            {
+                return 0;
+            }
```
WinterSolstice8, line 1852: CI is losing its mind from a missing close bracket here

GodModeXI: I got it XD I have no idea how I missed that, I literally just tested this before pushing.

src/map/utils/battleutils.cpp
```diff
@@ -1825,6 +1826,23 @@ namespace battleutils
...
         }
+        else if (PDefender->objtype == TYPE_TRUST)
+        {
+            CMobEntity* PTrust = (CMobEntity*)PDefender;
+            if (PTrust->getMobMod(MOBMOD_CAN_SHIELD_BLOCK) > 0)
+            {
+                base = PDefender->getMod(Mod::SHIELDBLOCKRATE);
+                if (PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_REPRISAL))
+                {
+                    base = base * 1.5f;
```
WinterSolstice8: I don't think checking Reprisal and returning base block rate is needed, reprisal and the combat skill check for block rate is calculated normally, otherwise you'd have to copy paste in all the missing stuff

reducing it to 

```
            if (PTrust->getMobMod(MOBMOD_CAN_SHIELD_BLOCK) > 0)
            {
                base = PDefender->getMod(Mod::SHIELDBLOCKRATE);
            }
```
should accomplish everything we need, which was to establish base block rate.... but that would get `SHIELDBLOCKRATE` added a second time, boosting it higher than it should. @TeoTwawki thoughts? new mobmod/mod for base block rate or copy paste in the skill calculations?

This block would add back in the skill calc:
```
            if (PTrust->getMobMod(MOBMOD_CAN_SHIELD_BLOCK) > 0)
            {
                base = PDefender->getMod(Mod::SHIELDBLOCKRATE);
                
                // Check for Reprisal and adjust skill and block rate bonus multiplier
                if (PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_REPRISAL))
                {
                    blockSkill   = blockSkill * 1.15f;
                    reprisalMult = 1.5f; // Default is 1.5x

                    // Adamas and Priwen set the multiplier to 3.0x while equipped
                    if (PDefender->getMod(Mod::REPRISAL_BLOCK_BONUS) > 0)
                    {
                        reprisalMult = 3.0f;
                    }
                }
                
                skillModifier = (blockSkill - attackSkill) * 0.2325f;

                // Add skill and Palisade bonuses
                base += skillModifier + palisadeMod;
                // Multiply by Reprisal's bonus
                base = base * reprisalMult;

                // Apply the lower and upper caps
                blockRate = (base < 5) ? 5 : base;
                blockRate = (base > 100) ? 100 : base;
                
                return blockRate;
            }
```

which is leaving in REPRISAL_BLOCK_BONUS check in case someone wanted custom content with Priwen/Adamas trust or something

TeoTwawki: I was thinking, there are mobs that should also be able to block and I am 99% sure they don't right now.

I thought we we could do something like:
```cpp
if (PDefender->objtype != TYPE_PC)
{
    if (PDefender->getMobMod(MOBMOD_CAN_SHIELD_BLOCK) > 0)
    {
        base = PDefender->getMod(Mod::SHIELDBLOCKRATE);

        if (PDefender->objtype == TYPE_PET)
        {
            // Automaton specific skillcheck
        }
        else
        {
            // trust and mob specific skillcheck 
            // (it's the same code of checking its skillcap for its job's rank regardless which it is)
        }

       // any needed clampage here
    }
}
```
SHIELDBLOCKRATE is just standing in for the shield-size they don't actually have (despite appearances on the model).

Using the skill is the already established pattern for extant code. (I don't know if that was right, I just know it's what was there before we started - it's hard to tell if automaton blocks were ever right or wrong when I've never played it)

Reprisal and optional custom bonuses would then do its thing after the skill checking per your description above.

I figured we'd be working out of this spot: https://github.com/LandSandBoat/server/blob/9bcbddb2a94358bfbf1a1400ce6d506dba84a8aa/src/map/utils/battleutils.cpp#L1814

GodModeXI: I can adjust this, but CBattleEntity wouldn't set getMobMob, so I had to reassign PDefender to CMobEntity.
```
CMobEntity* PTrust = (CMobEntity*)PDefender;
if (PTrust->getMobMod(MOBMOD_CAN_SHIELD_BLOCK) > 0)
```

TeoTwawki: above was kinda pseudo code, but we'd start off just converting the pet block over to a "non player" block with the other entity checks within it, and the skill-calc for mob and trust would be identical, entity casting aside was what I was getting at.

WinterSolstice8: seems reasonable to me

GodModeXI: the mobMod will need to be added to the Automatons also?

TeoTwawki: > the mobMod will need to be added to the Automatons also?

I was thinking it'd be added to all non-players we want to do blocks. So where valoredge currently is getting its "hey you are a tank now bro" we'd give it the mobmod. Looks like that would be over here, and [gosh look at that](https://github.com/LandSandBoat/server/blob/ff257cf51c312c5b186e203410318b066ccf3f1c/src/map/utils/petutils.cpp#L712), it really is a "phantom shield". I hate this so much. Non Players don't have gear but we're cramming a value into an imaginary slot just to have it in memory. _and then proceeded to not use the "size" set there and hardcode a value **anyway**_   Thanks past devs, I hate it.

GodModeXI:
```
        if (PDefender->objtype == TYPE_PC)
        {
            CCharEntity*    PChar = (CCharEntity*)PDefender;
            CItemEquipment* PItem = PChar->getEquip(SLOT_SUB);

            if (PItem)
            {
                shieldSize = PItem->getShieldSize();
            }
            else
            {
                return 0;
            }
        }
        else if (PDefender->objtype != TYPE_PC)
        {
            CMobEntity* PEntity = (CMobEntity*)PDefender;
            if (PEntity->getMobMod(MOBMOD_CAN_SHIELD_BLOCK) > 0)
            {
                base = PDefender->getMod(Mod::SHIELDBLOCKRATE);
                if (PDefender->objtype == TYPE_PET)
                {
                    skillModifier = (int8)((PDefender->GetSkill(SKILL_AUTOMATON_MELEE) - attackSkill) * 0.215f);
                    if (base <= 0)
                    {
                        return 0;
                    }
                    else
                    {
                        return base + skillModifier;
                    }
                }
                else if (PDefender->objtype == TYPE_TRUST)
                {
                    // Check for Reprisal and adjust skill and block rate bonus multiplier
                    if (PDefender->StatusEffectContainer->HasStatusEffect(EFFECT_REPRISAL))
                    {
                        blockSkill   = blockSkill * 1.15f;
                        reprisalMult = 1.5f; // Default is 1.5x

                        // Adamas and Priwen set the multiplier to 3.0x while equipped
                        if (PDefender->getMod(Mod::REPRISAL_BLOCK_BONUS) > 0)
                        {
                            reprisalMult = 3.0f;
                        }
                    }

                    skillModifier = (blockSkill - attackSkill) * 0.2325f;

                    // Add skill and Palisade bonuses
                    base += skillModifier + palisadeMod;
                    // Multiply by Reprisal's bonus
                    base = base * reprisalMult;

                    // Apply the lower and upper caps
                    blockRate = (base < 5) ? 5 : base;
                    blockRate = (base > 100) ? 100 : base;

                    if (PDefender->objtype == TYPE_TRUST)
                    {
                        printf("battleutils.cpp GetBlockRate TRUST: [%s] BLOCKRATE: [%3.0f]\n", PDefender->GetName(), blockRate);
                    }

                    return blockRate;
                }
                else if (PDefender->objtype == TYPE_MOB)
                {
                    if (PMob->m_EcoSystem != ECOSYSTEM::UNDEAD && PMob->m_EcoSystem != ECOSYSTEM::BEASTMAN)
                    {
                        return 0;
                    }
                }
            }
        }
```
Mob would have same calc as trusts? do they use reprisal and palisade? (honestly don't know)

WinterSolstice8: It's not so much that they "do" but they "could" so we don't have to rework it later

GodModeXI: ahh, got it. 👍

WinterSolstice8: Change this `else if (PDefender->objtype == TYPE_TRUST)` to just an `else`

and then

Delete this block, as it's no longer necessary
```                
                else if (PDefender->objtype == TYPE_MOB)
                {
                    if (PMob->m_EcoSystem != ECOSYSTEM::UNDEAD && PMob->m_EcoSystem != ECOSYSTEM::BEASTMAN)
                    {
                        return 0;
                    }
                }
```

This way we can add mobmods to allow things to block when they should be able to

For automatons we can remove `PPet->m_Weapons[SLOT_SUB]->setShieldSize(3);` and replace with `setMobMod()` that mod that was created in this PR. Don't know if that shieldSize is even used at all but we need to set the mod anyway

GodModeXI: @TeoTwawki @WinterSolstice8 

Blockskill pulls Skill_Shield for calculations. Will need to add that back into trustutils for the blocking calculations. image

TeoTwawki: > @TeoTwawki @WinterSolstice8
> 
> Blockskill pulls Skill_Shield for calculations. Will need to add that back into trustutils for the blocking calculations. !

The calculating can be done anywhere because we don't need to get skill they actually "have" like a player would, instead we're getting what the cap _would be_ if they _were_ a player with maxed out skill for the level, make sense?

we fake skills for mobs all over the core this way for various things.

GodModeXI: Adding SKILL_SHIELD for trust

image

Adding base block rate to valor edge 

image

GodModeXI: > > @TeoTwawki @WinterSolstice8
> 
> > 
> 
> > Blockskill pulls Skill_Shield for calculations. Will need to add that back into trustutils for the blocking calculations. !
> 
> 
> 
> The calculating can be done anywhere because we don't need to get skill they actually "have" like a player would, instead we're getting what the cap _would be_ if they _were_ a player with maxed out skill for the level, make sense?
> 
> 
> 
> we fake skills for mobs all over the core this way for various things.
> 
> 

Wellllll I just submitted it right as you commented 🙄 it was flooring rates to -96 with the blockskill pulling a 0 into the calculation. Adding the skill hit the multiplier like a player and can still be cranked to cap with the mod if an admin so chooses.

TeoTwawki: what we do on mobs is instead of checking skill of the entity, we check skill of its job(s) at that level:


```cpp
battleutils::GetMaxSkill((SKILLTYPE)i, PMob->GetSJob(), mLvl > 99 ? 99 : mLvl); 
```
This example is checking a subjob because it was a fallback for mobs that cast spells from a sub when main has no magic - it was just 1st example I found

TeoTwawki: > Wellllll I just submitted it right as you commented roll_eyes it was flooring rates to -96 with the blockskill pulling a 0 into the calculation. Adding the skill hit the multiplier like a player and can still be cranked to cap with the mod if an admin so chooses.

I just woke up and checked my notices so I could not catch you sooner, sorry

P,S.
That skill setting angers me because it should not exist. Trust aren't bloody supposed to have real skills, they aren't like players in that respect on retail, they are like mobs. We don't have a skill booster for MOBS now do we? -mutter grumble-

GodModeXI: > what we do on mobs is instead of checking skill of the entity, we check skill of its job(s) at that level:
> 
> ```c++
> battleutils::GetMaxSkill((SKILLTYPE)i, PMob->GetSJob(), mLvl > 99 ? 99 : mLvl); 
> ```
> 
> This example is checking a subjob because it was a fallback for mobs that cast spells from a sub when main has no magic - it was just 1st example I found

thats what i thought this was?
`battleutils::GetMaxSkill(SKILLTYPE::SKILL_SHIELD, mJob, mLvl > 99 ? 99 : mLvl);
            if (maxSkill != 0)`

TeoTwawki: if it helps you understand me, I can't actually see your code anymore, only your comments/notifications that describe what you said you did. github here is totally broken for me on this pr. anytime I try to look at the file changes the page goes "not responding".

So I am relying on what you said you did and winter's suggestions that were among them before I got here.

I saw you say you were giving them skill in trust utils and made the logical conclusions. And the setting is still bunk I tell you, bunk! _sulks_

GodModeXI: Should I close and reopen this as a new PR how it stands for clarity? I really want to do this right for continuity.

TeoTwawki:
> Should I close and reopen this as a new PR how it stands for clarity? I really want to do this right for continuity.

I'd normally say no, but github seems to have broken something during their outage the other night. Lets see if we can open again from same branch. It'll lose the comments here but I think all of us are pretty well caught up atm.

WinterSolstice8:
No problem with reopening here

PR #3039, opened 2022-10-26
[core] Make mobs pivot when in melee range rather than move around

### LINKED ISSUE #2268: 🐛 Mobs should not have a required minimum distance between themselves and their target

## Branch affected by issue 

`base`

## Steps to reproduce

1. Have a mob actively attacking you
2. Move around within that mob's melee attack range
3. Watch it completely reposition itself to ensure there is a minimum distance between itself and you

## Expected behavior

Mobs should only reposition themselves if they are too close to another mob or you move outside of their attack range. If you move around within a mob's melee range, it should just spin to face you rather than move its body entirely. 

An example of this on retail is from this KV capture: https://www.youtube.com/watch?v=UZZDwYNVgUo&t=251s
If you see Commencal gets drawn in and KV just spins to look at him instead of completely repositioning itself.

This issue has an added side effect in game in connection with draw in. If you trigger draw in repeatedly on a mob, you can cause the mob to back itself up over and over off the bounds of the map. Which then will cause the mob to draw you in out of the map.

WinterSolstice8:
**_I affirm:_**

Fixes #2268 by way of allowing mobs to pivot when their target is within melee range
Adds an early return on draw-in to prevent any movement just in case

See #2268

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -652,6 +652,7 @@ void CMobController::Move()
             if (currentDistance >= PMob->GetMeleeRange() * 2 && battleutils::DrawIn(PTarget, PMob, PMob->GetMeleeRange() - 0.2f))
             {
                 FaceTarget();
+                return;
             }
         }
         if (PMob->speed != 0 && PMob->getMobMod(MOBMOD_NO_MOVE) == 0 && m_Tick >= m_LastSpecialTime)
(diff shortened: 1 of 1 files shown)


TeoTwawki (approved):
Fine for now but bear in mind on retail mobs often DO move when they didn't need to, creating a very annoying near constant flanking in some cases (so we're better-than-retail as of this change)

its typically smaller mobs that do that, where as larger ones like behemoths, marids, dhalmels, scorpians etc usually will pivot.

WinterSolstice8:
if I had to guess, mobs on retail might base distance on path nodes/navmesh and that's why sometimes they shift when they don't need to. Pure speculation of course.

zach2good:
From observation, I'd agree with the pathing node idea

zach2good (approved):

zach2good (approved):

TeoTwawki:
@WinterSolstice8 @zach2good kind of but also not purely: mobs path to nearest node _when they are chasing down a point out of melee range_ with larger mobs (or those with crazy melee range - few smaller mobs can hit you from like 10 yalm away -_- and thus don't "dance") you are still in reach but smaller ones when you move in that arc around their pos, for the briefest of moments on retail I think for the briefest of moments it perceives you out of range and tries to adjust to nearest node again, resulting in that irritating dance around each other. Since we don't do pathing in exact same way nor do we have all sizes/melee ranges nor even handle that the same as retail for targets, we were getting dances where we shouldn't. Better to have none than extra, not exactly a feature to retail anyone liked anyway. /endTeoTriviaArchive

PR #3045, opened 2022-10-26
[Fix][Interaction] Correct variable check in Crawlers Nest treasure chest.

Xaver-DaRed:
**_I affirm:_**

Makes crawlers blood KI obtainable again.

Run the quest.

### DIFF
--- a/scripts/globals/treasure.lua
+++ b/scripts/globals/treasure.lua
@@ -603,10 +603,13 @@ xi.treasure.treasureInfo =
                 misc =
                 {
                     {
-                        test = function(player) return player:getCharVar("needs_crawler_blood") == 1 end,
+                        test = function(player)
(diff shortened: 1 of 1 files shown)


scripts/globals/treasure.lua
```diff
@@ -603,10 +603,13 @@ xi.treasure.treasureInfo =
                 misc =
                 {
                     {
-                        test = function(player) return player:getCharVar("needs_crawler_blood") == 1 end,
+                        test = function(player)
+                            return player:getCharVar("Quest[0][85]prog") >= 2 and
```
claywar: Since we're not space-constrained (like in bcnm.lua), could you please use xi.quest.getVar and its parameters?

Xaver-DaRed: Done

claywar (approved):

PR #3061, opened 2022-10-28
[core] Update pet/trust despawn animation logic to match retail, fix pets having green names

WinterSolstice8:
**_I affirm:_**

From my commit:

* This implements the following about pet/trust spawn/despawn animations:
    Only the first spawn animation of a pet/trust plays the "special" animation.
    Zoning (pets only) doesn't play the "die" animation. Trusts still "die" as they are not stored on zone.
    This means pets only disappear on zoning now, as expected.

    * "Special animations" include:

    Wyvern "from the sky" spawn in animation
    Avatar glow + crest effects on the ground
    Automaton "appear" animation with the effect
    Trust summon effect

    * GEO bubbles are not effected as they only have one spawn in animation.

In addition, fixes pets having green names.

2box, have one character summon an avatar, have the other run to 50' of the avatar and see it despawn. walk back in range and see if not have the "special" spawn animation and simply appear.

Have the summoner zone, and see the avatar despawn instead of die.
Repeat for Wyverns, automatons.

For trusts the above is true except the "die" animation is used on zone.

### DIFF
--- a/src/map/packets/entity_update.cpp
+++ b/src/map/packets/entity_update.cpp
@@ -68,10 +68,7 @@ void CEntityUpdatePacket::updateWith(CBaseEntity* PEntity, ENTITYUPDATE type, ui
         case ENTITY_SPAWN:
         {
             updatemask = UPDATE_ALL_MOB;
-            if (PEntity->objtype == TYPE_PET)
-            {
-                ref<uint8>(0x28) = 0x04;
-            }
+
             if (PEntity->look.size == MODEL_EQUIPPED || PEntity->look.size == MODEL_CHOCOBO)
             {
                 updatemask = 0x57;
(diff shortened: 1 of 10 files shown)


claywar (approved):

src/map/packets/entity_update.cpp
```diff
@@ -68,10 +68,7 @@ void CEntityUpdatePacket::updateWith(CBaseEntity* PEntity, ENTITYUPDATE type, ui
         case ENTITY_SPAWN:
         {
             updatemask = UPDATE_ALL_MOB;
-            if (PEntity->objtype == TYPE_PET)
-            {
-                ref<uint8>(0x28) = 0x04;
-            }
+
```
jmcmorris, line 71: So happy to see this go away. I looked at the history and this code goes back 11 years. Someone added it in not knowing what it actually was or needed to do and it has just stayed around that entire time!

jmcmorris (approved):

kaincenteno:
awesome! always was annoying to have to see that animation when moving close to SMN pets
great work 😄

claywar (approved):

PR #3072, opened 2022-10-30
Add mob modifier `MAGIC_RANGE` which allows for changing the range mo…

jmcmorris:
**_I affirm:_**

Add mob modifier `MAGIC_RANGE` which allows for changing the range mob will aggro magic

Go to any mob that aggros magic and use the following command while targeting it. Try casting magic 15' away and it shouldn't aggro. Get within 10' and cast again and you should get aggro.
```
!setmobmod MAGIC_RANGE 10
```

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -2528,6 +2528,7 @@ xi.mobMod =
     NO_LINK             = 69, -- If set, mob cannot link until unset.
     NO_REST             = 70, -- Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
     LEADER              = 71, -- Used for mobs that follow a defined "leader", such as Ul'xzomit mobs.
+    MAGIC_RANGE         = 72, -- magic aggro range
 }
 
 -----------------------------------
@@ -2912,7 +2913,7 @@ xi.animation =
 {
     NONE                    = 0,
     ATTACK                  = 1,
-    -- Death 2              = 2,
+    DESPAWN                 = 2,
     DEATH                   = 3,
     CHOCOBO                 = 5,
     FISHING                 = 6,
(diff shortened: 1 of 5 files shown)


scripts/globals/status.lua
```diff
@@ -2912,7 +2913,7 @@ xi.animation =
 {
     NONE                    = 0,
     ATTACK                  = 1,
-    -- Death 2              = 2,
+    DESPAWN                 = 2,
```
jmcmorris, line 2916: I didn't mean to commit this but it doesn't hurt anything. Just a missing entry for some reason.

TeoTwawki: I'm not sure the naming is correct: when I try this on things they do their death animation, and don't despawn/fade out at all

TeoTwawki: it probably expects to be paired with something or happen in order at the client

jmcmorris: This is what it is set to on the C++ side. This brings them in line; however, I'm happy to just revert this line change.

TeoTwawki: _shrug_

core one prolly shouldn't have been done without some notation AND adding the same here. It really angers me when ppl desync the damned enumeration

kaincenteno: > I'm not sure the naming is correct: when I try this on things they do their death animation, and don't despawn/fade out at all

can be named it PLAY_DEAD like what dogs and cats do since they dont despawn lol

TeoTwawki: > > I'm not sure the naming is correct: when I try this on things they do their death animation, and don't despawn/fade out at all
> 
> can be named it PLAY_DEAD like what dogs and cats do since they dont despawn lol

image

claywar (approved):

WinterSolstice8 (approved):

Issue #2240, opened 2022-07-09
🔨 dbtool improvements 


zach2good:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Describe the feature

Dbtool is hot shit. It's one of the best things to happen to the project in the last 5 years that isn't a massive internal refactoring. 

That being said, I think it needs some love:
- https://github.com/LandSandBoat/server/issues/86 (not sure if this is still valid) 
- not hard coding it to only run from the tools folder 
- add some more functionality (settings ips etc) 
- make the ui dynamic, and make sure everything you can do from the ui can be done from command line
- break out into:
```
tools/
    dbtool/
        ... dbtool internal scripts ...
    dbtool.py
    ...
```

PR #3075, opened 2022-10-30
[dbtool] Refactor to allow running from other directories like repo root

zach2good:
**_I affirm:_**

As part of https://github.com/LandSandBoat/server/issues/2240, one of the issues is to remove the requirement for dbtool to always be run from <root>/tools. This does that.

Use dbtool as normal, but launch it from `<root>/tools`, or `<root>`, or `<root>/scripts/globals` and make sure everything works as expected.

### DIFF
--- a/tools/dbtool.py
+++ b/tools/dbtool.py
@@ -23,13 +23,18 @@ except: # lgtm [py/catch-base-exception]
     print("ERROR: Make sure git is installed and available on your system's PATH environment variable.")
     preflight_exit()
 
-# - dbtool.py is designed to be run from <root>/tools folder, not <root>
-if not os.path.isfile("./dbtool.py"):
-    print("ERROR: dbtool.py is designed to be run from the <root>/tools folder, not <root>. Please run from the tools folder.")
-    preflight_exit()
+# If launching from somewhere that isnt <root>/tools/, figure out where dbtool.py is located, and the server root.
+dbtool_dir_path = os.path.normpath(os.path.realpath(os.path.dirname(__file__)))
+server_dir_path = os.path.normpath(os.path.realpath(os.path.dirname(dbtool_dir_path)))
+
+def from_dbtool_path(path):
+    return os.path.normpath(os.path.join(dbtool_dir_path, path))
+
+def from_server_path(path):
+    return os.path.normpath(os.path.join(server_dir_path, path))
 
 # - Repo should be checked out as a git repo, not as plain files
-if not subprocess.call(['git', '-C', "../", 'status'], stderr=subprocess.STDOUT, stdout = open(os.devnull, 'w')) == 0:
+if not subprocess.call(['git', '-C', server_dir_path, 'status'], stderr=subprocess.STDOUT, stdout = open(os.devnull, 'w')) == 0:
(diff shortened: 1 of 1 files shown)


tools/dbtool.py
```diff
@@ -263,24 +276,25 @@ def fetch_files(express=False):
     if express:
         try:
             global express_enabled
-            sql_diffs = repo.commit(db_ver).diff(release_version,paths='sql/')
+            sql_diffs = repo.commit(db_ver).diff(release_version, paths=from_server_path('sql/'))
             if len(sql_diffs) > 0:
                 for diff in sql_diffs:
                     import_files.append(diff.a_path[4:])
```
cocosolos: Need to add path here for express update imports. It reads them correctly but fails when it tries to actually import them.

tools/dbtool.py
```diff
@@ -263,24 +276,25 @@ def fetch_files(express=False):
...
-            import_files.extend(filenames)
+        for (_, _, filenames) in os.walk(from_server_path('sql/')):
+            for filename in filenames:
+                import_files.append(from_server_path('sql/' + filename))
             break
     check_protected()
     backups.clear()
-    for (_, _, filenames) in os.walk('../sql/backups/'):
+    for (_, _, filenames) in os.walk(from_server_path('sql/backups/')):
         backups.extend(filenames)
```
cocosolos: Same here, needs path.

tools/dbtool.py
```diff
@@ -579,96 +596,135 @@ def menu():
...
+def tasks():
+    if input('Make server public-facing? [y/N] ').lower() == 'y':
+        import urllib.request
+        external_ip = urllib.request.urlopen('https://ident.me').read().decode('utf8')
+        print_green(f'Detected your public-facing IP as: {external_ip}')
+        if input('Is this correct? [y/N] ').lower() == 'y':
+            set_external_ip(external_ip)
+        else:
+            external_ip = input('Please enter your public-facing IP: ')
+            set_external_ip(external_ip)
```
cocosolos: Should maybe validate that that `external_ip` isn't blank and skip setting if it is in case they decide to bail.

tools/dbtool.py
```diff
@@ -226,15 +239,15 @@ def fetch_configs():
...
 def fetch_module_files():
-    with open('../modules/init.txt', 'r') as file:
+    with open(from_server_path('modules/init.txt'), 'r') as file:
         for line in file.readlines():
             if not line.startswith('#') and line.strip() and not line in ['\n', '\r\n']:
-                line = "../modules/" + line.strip()
+                line = from_server_path("modules/" + line.strip())
                 if pathlib.Path(line).is_dir():
                     for filename in pathlib.Path(line).glob('**/*.sql'):
                         import_files.append(str(filename).replace('\\', '/'))
```
cocosolos: Think this needs path added too.

zach2good: `str(filename)` on a file object provides an absolute path, which is handy 👍. Confirmed with debugger

zach2good:
Tested initial import, backup, nuking the db, restoring from backup, and the new IP task. All working as expected 👍

claywar (approved):

PR #3076, opened 2022-10-30
Update Glenne.lua

Daywhite83:
fixed the weird thing he was doing in the middle. pretty close to retail accurate

**_I affirm:_**

Fixes the weird walking Glenne was doing, Removed redundant paths. he only cycles between the 2 points.
Looks pretty close to how he walks in retail now. Will link video from retail in next comment


Daywhite83:
Retail video: https://www.youtube.com/watch?v=688_wdqkyS8&list=PLY8c2lFcRz1XAXg5CvjTND6m5tUpERTmS&index=8

TeoTwawki:
image

"update filename" doesn't help anyone understand what you did.

----

### From the last time I posted on this subject:

normally, both the commit message AND the pr title should be like a summary of what you've actually done (commit messages are suggested to be like you are giving git orders `cook fries before putting them in the basket`) and then "`fixes #number` would be tacked on the next line of the commit message and in the description body of the pr for best results. We don't enforce a particular style, but we do ask for descriptive naming in both PR titles and in commit messages so we can look in the history of both local copies and github merges and see what it actually was at a glance.


Links that may help with writing meaningful messages, and a short sample from them:

https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/


> To come up with thoughtful commits, consider the following:
> 
> - Why have I made these changes?
> - What effect have my changes made?
> - Why was the change needed?
> - What are the changes in reference to?
> 
> Assume the reader does not understand what the commit is addressing. They may not have access to the story addressing the detailed background of the change.

https://initialcommit.com/blog/git-commit-messages-best-practices

> As a general rule, your messages should start with a single line that’s no more than about 50 characters and that describes the changeset concisely, followed by a blank line, followed by a more detailed explanation.

And finally, this is from the author of git himself:

https://github.com/torvalds/subsurface-for-dirk/blob/a48494d2fbed58c751e9b7e8fbff88582f9b2d02/README#L88

> Header line: explain the commit in one line (use the imperative)
> 
> Body of commit message is a few lines of text, explaining things
> in more detail, possibly giving some background about the issue
> being fixed, etc etc.
> 
> The body of the commit message can be several paragraphs, and
> please do proper word-wrap and keep columns shorter than about
> 74 characters or so. That way "git log" will show things
> nicely even when it's indented.
> 
> Make sure you explain your solution and why you're doing what you're
> doing, as opposed to describing what you're doing. Reviewers and your
> future self can read the patch, but might not understand why a
> particular solution was implemented.

TeoTwawki:
deletion isn't needed, nor is closing the pr. git is capable of overwriting the branch and github automatically updates the PR.

TeoTwawki:
git rebase can edit any/all commits, while git amend lets you edit the most recent one. Then, one would use a force-push to overwrite the version on github.

PR #3078, opened 2022-10-31
Rework Temenos Eastern Tower

jmcmorris:
**_I affirm:_**

Migrate and rework Temenos Eastern Tower Limbus area to be in the new battlefield framework. All of this work is heavily based off of the following captures and sources:
https://www.bg-wiki.com/ffxi/Eastern_Temenos
https://ffxiclopedia.fandom.com/wiki/Temenos_-_Eastern_Tower
https://youtube.com/watch?v=oCDo9jrk9_o
https://youtu.be/3sqAx8hwvtY

```
!addkeyitem white_card
!addkeyitem cosmo_cleanse
!pos 580.000 -2.375 104.000 37
```


scripts/globals/utils.lua
```diff
@@ -43,6 +43,22 @@ function utils.bind(func, ...)
...
+-- Creates a slice of an input table and returns a new table
+function utils.slice(inputTable, first, last, step)
+    local slicedTable = {}
+    first = first or 1
+    last = last or #inputTable
+    step = step or 1
+    local position = 1
+
+    for i = first, last, step do
+        slicedTable[position] = inputTable[i]
```
zach2good, line 55: Is it possible for this to step over the end and place a nil in the new table?

zach2good (approved):

scripts/globals/utils.lua
jmcmorris: Not from what I tested. If step > 1 then it will properly stop not place a nil. Even if it did though it wouldn't cause problems since it'd be nil.

jmcmorris:
Build is failing due to needing `MAGIC_RANGE` from #3072

PR #3087, opened 2022-11-01
Fix Jorm and Tiamat logic being broken when slept mid-flight

Xaver-DaRed:
**_I affirm:_**

Becouse of certain factors (mainly, mobs not being able to target specific positions instead of entities) when sleeping Tiamat or Jorm when in subanimation == 1 (Flight mode) and then retreating enough to get away from skill range, its logic would break and it would forever stay flying.

This happens becouse its the skill itself that handles the "landing" and change to subanimation 2.

In retail, Tiamat and Jorm will just use the skill anyway no matter if theres some1 in range.

This PR attempts to mittigate this problem by adding both distance and sleep checks to ensure the skill is only put on qeue when they are awake and used when the main target is in range.

It also attempts to clean their scripts.

Fight both mobs.


scripts/zones/Uleguerand_Range/mobs/Jormungand.lua
```diff
@@ -1,61 +1,80 @@
 -----------------------------------
 -- Area: Uleguerand Range
---  HNM: Jormungand
```
WinterSolstice8, line 3: Poor jormy getting downgraded

WinterSolstice8 (approved):

scripts/zones/Uleguerand_Range/mobs/Jormungand.lua
Xaver-DaRed: A mob is a mob XD

PR #3104, opened 2022-11-02
Clean nomad_cap script and make it use a table

Xaver-DaRed:
**_I affirm:_**

Changes the elseif chain with a table. No actual functionality changes.

Use nomad cap

### DIFF
--- a/scripts/globals/items/nomad_cap.lua
+++ b/scripts/globals/items/nomad_cap.lua
@@ -1,9 +1,20 @@
 -----------------------------------
---  ID: 16119
---  Nomad Cap
---  Transports the user to their Home Nation
+-- ID: 16119
+-- Nomad Cap
+-- Transports the user to their Home Nation
+-- TODO: Confirm wiki claims of random zone destinations among a same nation.
+-----------------------------------
+require("scripts/globals/zone")
 -----------------------------------
 local itemObject = {}
+
+local pos =
+{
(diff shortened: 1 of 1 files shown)


scripts/globals/items/nomad_cap.lua
```diff
@@ -1,29 +1,29 @@
...
+-- Transports the user to their Home Nation
+-- TODO: Confirm wiki claims of random zone destinations among a same nation.
+-----------------------------------
+require("scripts/globals/zone")
 -----------------------------------
 local itemObject = {}
 
+local pos =
+{
+    -- Content  [Nation] = {   x,   y,   z, rot, zone},
```
zach2good: CI even looks in the comments! :D

```
Table closed without an appropriate preceding space or newline: scripts/globals/items/nomad_cap.lua:13:54
-- Content  [Nation] = {   x,   y,   z, rot, zone},                              <-- HERE
```

zach2good (approved):

jmcmorris (approved):

scripts/globals/items/nomad_cap.lua
claywar: I have comments ignored for things in a separate branch I've been working on, if we want all comments ignored, can bump that up in the list of checks

zach2good: In this specific case, I think its useful. But in general I'm sure we'll hit lots of false positives

PR #3122, opened 2022-11-03
[fix] Weaponskill Damage Message Param Update

Dukilles:
**_I affirm:_**

-Properly display Overwhelm/Phalanx damage modifications to weaponskills in chat log by moving the damage message parameter update to after takeWeaponskillDamage()

- Take a 75+ SAM to any mob.
- Ensure SAM has KI 606 (Limit Breaker)
- !setmerits 30 for the SAM
- In Mog House, put 5/5 points in Overwhelm 
- Choose a GK weaponskill (I used Tachi: Gekko, scripts\globals\weaponskills\tachi_gekko.lua) 
- - temporarily add the following to its lua, just above ```return tpHits, extraHits, criticalHit, damage```:
- - print("Damage:" ..damage)
- - print("Target HP:" ..target:getHP())
- Use a second character (preferably PLD) that is able to hold hate.
- Engage a mob, hold hate with PLD.
- !tp 3000 SAM
- !hp 1000 mob
- Tachi: Gekko from behind mob
- Note that damage displayed in chat and in print are the same.
- Note that HP remaining in print is correct, mob took expected damage.
- !hp 1000 mob
- Tachi: Gekko in front of mob
- Note that damage displayed in chat and in print are NOT the same.
- Note that HP remaining demonstrates that chat log is incorrect.
- Apply PR changes to scripts/globals/weaponskills.lua
- Start again from behind mob and repeat the process.
- Note that damage displayed in chat is now accurate.
- Remove temporary prints from scripts\globals\weaponskills\tachi_gekko.lua

Overwhelm and Phalanx are calculated in battleutils TakeWeaponskillDamage, so the message param update needs to be after takeWeaponskillDamage() to accurately reflect these damage modifications.

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -890,7 +890,6 @@ function takeWeaponskillDamage(defender, attacker, wsParams, primaryMsg, attack,
             end
         end
 
-        action:param(defender:getID(), math.abs(finaldmg))
     elseif wsResults.shadowsAbsorbed > 0 then
         action:messageID(defender:getID(), xi.msg.basic.SHADOW_ABSORB)
         action:param(defender:getID(), wsResults.shadowsAbsorbed)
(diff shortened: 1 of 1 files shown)


TeoTwawki (approved):
makes sense to me

detetivepro15:
> makes sense to me

ok

PR #3124, opened 2022-11-03
Fix getstats treasure hunter and battletime

xkoredev:
**_I affirm:_**

- Fixes conditional checks in getstats that would cause th level and battle time to not be printed

Target a mob and use !getstats. Should show th level and battle time


TeoTwawki (approved):
good catch, somehow typo went unnoticed

PR #3127, opened 2022-11-03
Fix Weaponskill message returning 0 shadows when target is under Blink/Utsusemi effect

Dukilles:
**_I affirm:_**

-I realized there was a problem with ```[fix] Weaponskill Damage Message Param Update #3122``` where shadows/blink message would also be updated and show 0. This pull request resolved that issue. Apologies for the oversight.

-Follow steps in ```[fix] Weaponskill Damage Message Param Update #3122```
-In addition, !setmod 307 10 and note that shadow absorb message is accurate, as well as weaponskill damage messages.

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -869,11 +869,14 @@ end
 -- handles displaying the appropriate action/message, delivering the damage to the mob, and any enmity from it
 function takeWeaponskillDamage(defender, attacker, wsParams, primaryMsg, attack, wsResults, action)
     local finaldmg = wsResults.finalDmg
+    local targetTPMult = wsParams.targetTPMult or 1
+    finaldmg = defender:takeWeaponskillDamage(attacker, finaldmg, attack.type, attack.damageType, attack.slot, primaryMsg, wsResults.tpHitsLanded, (wsResults.extraHitsLanded * 10) + wsResults.bonusTP, targetTPMult)
 
     if wsResults.tpHitsLanded + wsResults.extraHitsLanded > 0 then
         if finaldmg >= 0 then
             if primaryMsg then
                 action:messageID(defender:getID(), xi.msg.basic.DAMAGE)
+                action:param(defender:getID(), math.abs(finaldmg))
             else
                 action:messageID(defender:getID(), xi.msg.basic.DAMAGE_SECONDARY)
             end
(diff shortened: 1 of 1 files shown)


scripts/globals/weaponskills.lua
```diff
@@ -869,11 +869,14 @@ end
...
 function takeWeaponskillDamage(defender, attacker, wsParams, primaryMsg, attack, wsResults, action)
     local finaldmg = wsResults.finalDmg
+    local targetTPMult = wsParams.targetTPMult or 1
+    finaldmg = defender:takeWeaponskillDamage(attacker, finaldmg, attack.type, attack.damageType, attack.slot, primaryMsg, wsResults.tpHitsLanded, (wsResults.extraHitsLanded * 10) + wsResults.bonusTP, targetTPMult)
 
     if wsResults.tpHitsLanded + wsResults.extraHitsLanded > 0 then
         if finaldmg >= 0 then
             if primaryMsg then
                 action:messageID(defender:getID(), xi.msg.basic.DAMAGE)
+                action:param(defender:getID(), math.abs(finaldmg))
```
TeoTwawki: possibly also needed in the block at line 881 ? (I am legit not sure)

claywar:
Could you also please amend the commit name to be a bit more descriptive?

Dukilles:
> Could you also please amend the commit name to be a bit more descriptive?

Of course!

scripts/globals/weaponskills.lua
Dukilles: I'll test thoroughly to be 100% sure, but I think that both of those messages use the one param?

Dukilles: Testing with circle blade, DAMAGE_SECONDARY does also need the finaldmg param. Thanks!

Dukilles:
Ive got a different idea for this to try out, so marked as draft until i can test out my latest theory.

Dukilles:
Apologies for the mess of commits. :/
I have tested this way as many ways as I can think of.
This is closest to the original while moving only the damage params to after takeWeaponskillDamage fires.

Also will be looking into this, but I think its Phalanx is possibly being done twice. Both in the lua and the battleutil

Dukilles:
Put things back together the way they were, just with a param update after takeWeaponskillDamage for damage >= 0. This will update the message to be accurate with how much HP target actually loses, taking into account all of the damage modifications that happen in battleutils::takeWeaponskillDamage.

Last commit. Truly sorry for the mess here, I can close this and submit a new branch if needed.

I've tested quite a bit and it does seem to be the case that Phalanx is being applied twice for weaponskills: once in weaponskills.lua and again in battleutils::takeWeaponskillDamage. This PR only addresses the message being accurate and does not address double Phalanx.

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #3141, opened 2022-11-05
Remove SKILLFLAG_NO_TP_COST from avatar skill flags

xkoredev:
**_I affirm:_**

- Removes the unwanted `SKILLFLAG_NO_TP_COST` from a set of avatar tp moves

Closes https://github.com/AirSkyBoat/AirSkyBoat/issues/472

Fight all avatars and observe no repeated tp moves one after another.


Xaver-DaRed (approved):

zach2good (approved):

TeoTwawki:
wow, nice catch. I looked in history going back multiple years and that 0x04 has been included on most of these since the early days of darkstar even, when the 0x04 flag was just named "special" and mainly used for mobs using a skill as melee or ranged attack, and somehow every time flags were updated to add or remove other stuff, the 0x04 stayed. In more recent times it went totally unused (as setting mobskills-as-attacks is in lua now, and ranged ones are in the core with a mob family check) until I renamed and merged it with the also unused mob 2hr flag, which resulted in 0x04 being used again. It should have been stripped out of every row back when DSP axed it, but instead it was left dangling.

PR #3143, opened 2022-11-05
[ci] Refactor Lua stylechecks

claywar:
**_I affirm:_**

When written originally, the python usage in `tools/ci/lua.sh` was focused on tables only, which has quickly grown to other uses as well.  This will separate lua style checks into its own python file called by Lua CI, and define unique check functions for each set of rules to be called in various parts of scripts.

### Intent
* Move Lua style checks to its own python script to allow for easier readability in IDEs
* Create unique functions for each style check performed
* Create basic check function for each file, and apply style checks based on what structure is used (tables, conditions, functions, etc)
* Add docstring references to Wiki for each failure type

### TODO:
* Update Wiki for all CI requirements

NOTE: This adds `regex` dependency to python requirements.txt

CI should run successfully both locally and as an Action


claywar:
Ready for review.  Expecting CI to fail out, then will fix the remaining 19 errors and squash into the final commit

zach2good (approved):

PR #3144, opened 2022-11-05
[chore] Remove explicit boolean comparisons from Lua conditions

claywar:
**_I affirm:_**

Removes instances of comparing variables and functions to `true` or `false` in preparation for a CI requirement.  Functions that return bool, and boolean variables should be named appropriately to be readable using `not` or, in the case of true, by itself.

Example:
`if not player:hasKeyItem(...)` is preferred over `if player:hasKeyItem(...) == true`

Search pattern: `== true|== false|~= true|~= false`

No impact should be observed

### DIFF
--- a/scripts/commands/exec.lua
+++ b/scripts/commands/exec.lua
@@ -39,9 +39,9 @@ function onTrigger(player, str)
     end
 
     -- Execute the string..
-    local status, err1 = pcall(scriptObj)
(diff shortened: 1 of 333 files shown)


zach2good (approved):

jmcmorris (approved):
Good job and good riddance!

PR #3151, opened 2022-11-06
[core] [lua] [sql] DRG Jump audit + a few other DRG related things

WinterSolstice8:
**_I affirm:_**

dragoon.lua:
   * Implement Soul/Spirit jump force crit, attack bonus, extra TP when Wyvern is out
    * Update specEffect to match retail bits with retail testing
    * Move Jump TP bonus to performWSJump, instead of feeding it through the weaponskill system
    * Fly High now properly resets Jump timers
    * When jumps are used with Fly High active, their timers are set to 10 seconds.
    * Correct Spirit Surge high jump to reduce TP of target by damage * 2, not damage * .02
    * Remove enmity removal from Soul/Spirit jump (they are just 1 CE & 0 VE actions)
    * Fix jump fTP VIT bonus not applying correctly
    * Add various mods:
    ** Spirit Jump TP bonus
    ** Soul & Spirit Jump Attack Bonus
    ** Jump accuracy bonus
    ** Jump Double Attack bonus
    
weaponskills.lua:
* Add better support for Jumps
    * Gives full TP return to every hit of a jump
    * Add Jump Double attack support
    * Add Jump Accuracy support

A few message updates for Ancient Circle, fix AoE messages on JA that depended on CAbility::getAoEMsg()
Fix logspam when Jump is used
Correct Angon message when angon isn't equipped to match retail, remove magic number for Angon
Add AoE message case for Ancient Circle
Fix extra TP gain from hits that doesn't match retail during weaponskills

Test Jump TP bonus:
!setmod STORETP 34
!setmod JUMP_TP_BONUS 71
!setmod JUMP_SPIRIT_TP_BONUS 70

with a 492 delay weapon:
* Spirit Jump should return exactly 555 tp on a single hit with your wyvern out, and 371 without
* Soul Jump should return exactly 645 tp on a single hit with your wyvern out, and 278 without
* Jump and High Jump should return 278 tp on a single hit

Test Jump Double Attack:
set mods above
!setmod JUMP_DOUBLE_ATTACK 100
* Spirit Jump should return exactly 922 tp on a DA with your wyvern out, and 555 without
* Soul Jump should return exactly 1195 tp on a DA with your wyvern out, and 426 without
* Jump and High jump should return exactly 426 TP

Test Fly High:
use all jump timers, use fly high, see them all reset
use jumps, see all jumps have a timer of 10s during duration of fly high

Test Jump VIT bonus:
!setmod VIT 999
use normal jump, see it do ludicrous damage compared to high jump which does not have the fTP bonus

Ancient Circle aoe message should look as follows:
image

### DIFF
--- a/scripts/globals/items.lua
+++ b/scripts/globals/items.lua
@@ -3671,6 +3671,7 @@ xi.items =
     ORPHIC_EGG                      = 18256,
     BIBIKI_SEASHELL                 = 18257,
     THROWING_TOMAHAWK               = 18258,
+    ANGON                           = 18259,
     CAESTUS                         = 18263,
     SPHARAI                         = 18264,
     BATARDEAU                       = 18269,
--- a/scripts/globals/job_utils/dragoon.lua
+++ b/scripts/globals/job_utils/dragoon.lua
@@ -1,12 +1,13 @@
 -----------------------------------
 -- Dragoon Job Utilities
 -----------------------------------
-require("scripts/globals/settings")
 require("scripts/globals/ability")
-require("scripts/globals/status")
-require("scripts/globals/msg")
-require("scripts/globals/weaponskills")
+require("scripts/globals/items")
 require("scripts/globals/jobpoints")
+require("scripts/globals/msg")
+require("scripts/globals/settings")
+require("scripts/globals/status")
+require("scripts/globals/weaponskills")
 -----------------------------------
 xi = xi or {}
 xi.job_utils = xi.job_utils or {}
@@ -165,10 +166,10 @@ end
 xi.job_utils.dragoon.abilityCheckAngon = function(player, target, ability)
     local id = player:getEquipID(xi.slot.AMMO)
 
-    if id == 18259 then
+    if id == xi.items.ANGON then
         return 0, 0
     else
-        return xi.msg.basic.UNABLE_TO_USE_JA, 0
+        return xi.msg.basic.CANNOT_PERFORM, 0
     end
 end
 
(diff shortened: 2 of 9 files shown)


scripts/globals/items.lua
```diff
@@ -3671,6 +3671,7 @@ xi.items =
     ORPHIC_EGG                      = 18256,
     BIBIKI_SEASHELL                 = 18257,
     THROWING_TOMAHAWK               = 18258,
+    ANGON                           = 18258,
```
Xaver-DaRed: Wrong ID. Should be 18259

scripts/globals/job_utils/dragoon.lua
```diff
@@ -165,10 +193,10 @@ end
 xi.job_utils.dragoon.abilityCheckAngon = function(player, target, ability)
     local id = player:getEquipID(xi.slot.AMMO)
 
-    if id == 18259 then
+    if id == xi.items.ANGON then
```
Xaver-DaRed, line 197: Missing the associated require. `require("scripts/globals/items")`

scripts/globals/items.lua
WinterSolstice8: Oops. fixed.

scripts/globals/job_utils/dragoon.lua
WinterSolstice8: Done and sorted the requires while I was there

Xaver-DaRed (approved):

claywar (approved):

PR #3153, opened 2022-11-07
Dark Knight Update Consume Mana & Dark Seal

Schaedelschaden:
**_I affirm:_**

**Job Abilities**

- Sets up Consume Mana to apply during weaponskills.
- Adds Dark Seal merit and augment handling to globals/job_utils/dark_knight.lua.

**Effects**

- Sets up Dark Seal effect DARK_MAGIC_CAST and DARK_MAGIC_DURATION modifiers.

**Mods**

- Creates DARK_MAGIC_CAST and adds it to Dark Seal effect.
- Creates DARK_MAGIC_DURATION and adds it to Dark Seal effect.
- Creates ENHANCES_DARK_SEAL and adds it to Dark Seal subPower.

**Core/LUA Functions**

- Adds DARK_MAGIC_CAST handling to battleutils.cpp CalculateSpellCastTime function.

**Database**

- Adds ENHANCES_DARK_SEAL to augment 1370 in augments.sql.
- Adjusts Dark Seal merit value in merits.sql.

## References

Dark Knight - https://www.bg-wiki.com/ffxi/Dark_Knight
Consume Mana - https://www.bg-wiki.com/ffxi/Consume_Mana
Dark Seal - https://www.bg-wiki.com/ffxi/Dark_Seal
Dark Seal - http://wiki.ffo.jp/html/3247.html

2. Load the server.
3. Use the `!changejob DRK 99` command.
5. Use the !setmod command to add the updated ability modifiers.
6. Observe changes in job ability behavior with modifiers set.

### DIFF
--- a/scripts/globals/job_utils/dark_knight.lua
+++ b/scripts/globals/job_utils/dark_knight.lua
@@ -85,8 +85,12 @@ xi.job_utils.dark_knight.useConsumeMana = function(player, target, ability)
 end
 
 xi.job_utils.dark_knight.useDarkSeal = function(player, target, ability)
-    local merits = player:getMerit(xi.merit.DARK_SEAL)
-    player:addStatusEffect(xi.effect.DARK_SEAL, merits, 0, 60)
+    -- Power: Each merit level after the first reduces Dark Magic casting time by -10% (total of -40% bonus).
+    -- Sub Power: Enhances Dark Seal effect by increasing duration of Dark Magic by 10% per merit level (total of 50% bonus).
+    local power    = player:getMerit(xi.merit.DARK_SEAL) - 10
+    local subPower = (player:getMerit(xi.merit.DARK_SEAL) / 10) * player:getMod(xi.mod.ENHANCES_DARK_SEAL)
+
+    player:addStatusEffect(xi.effect.DARK_SEAL, power, 0, 60, 0, subPower)
 end
 
 xi.job_utils.dark_knight.useDiabolicEye = function(player, target, ability)
--- a/scripts/globals/effects/dark_seal.lua
+++ b/scripts/globals/effects/dark_seal.lua
@@ -4,12 +4,16 @@
 local effectObject = {}
 
 effectObject.onEffectGain = function(target, effect)
+    target:addMod(xi.mod.DARK_MAGIC_CAST, -effect:getPower())
+    target:addMod(xi.mod.DARK_MAGIC_DURATION, effect:getSubPower())
 end
 
 effectObject.onEffectTick = function(target, effect)
 end
 
 effectObject.onEffectLose = function(target, effect)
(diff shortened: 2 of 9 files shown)


scripts/globals/job_utils/dark_knight.lua
```diff
@@ -85,8 +85,10 @@ xi.job_utils.dark_knight.useConsumeMana = function(player, target, ability)
 end
 
 xi.job_utils.dark_knight.useDarkSeal = function(player, target, ability)
-    local merits = player:getMerit(xi.merit.DARK_SEAL)
-    player:addStatusEffect(xi.effect.DARK_SEAL, merits, 0, 60)
+    local power    = player:getMerit(xi.merit.DARK_SEAL) - 10
```
WinterSolstice8, line 90: Might want to leave a comment here about how the -10 is because the initial merit provides no cast time reduction and it's otherwise unusable without it, plus that each merit is 10% so 5/5 = -40%

scripts/globals/spells/damage_spell.lua
```diff
@@ -472,11 +472,6 @@ xi.spells.damage.calculateResist = function(caster, target, spell, skillType, sp
         magicAcc = magicAcc + 100 -- TODO: Confirm this value in retail
     end
 
-    -- Dark Seal
-    if casterJob == xi.job.DRK and skillType == xi.skill.DARK_MAGIC and caster:hasStatusEffect(xi.effect.DARK_SEAL) then
```
WinterSolstice8: Why was this removed? JP wiki seems to agree that nobody knows what it is, but they also agree that it's powerful like elemental seal. I can confirm Dark Seal more or less guarantees a land of an unresisted Drain 3 as long as its not dark resistant/immune. As long as it uses Elemental Seal value or less (there's data to show dispel is around +200 macc on it's own) I don't think it's all _that_ unreasonable

http://wiki.ffo.jp/html/3247.html

Schaedelschaden: As of the moment all dark magic is using the globals/magic.lua applyResistance function and never even hit this global. Should these be moved to use the xi.spells.damage.calculateResist function?

Alternatively, I had missed moving this check over into globals/magic.lua

Xaver-DaRed: They will, eventually, be moved to use a "not magic.lua" function. But that day is not today.

For now though, I would suggest leaving it there.

Schaedelschaden: Restored Dark Seal check to globals/spells/damage_spell.lua and added a check to globals/magic.lua.

scripts/globals/job_utils/dark_knight.lua
Schaedelschaden: Comment added.

WinterSolstice8 (approved):

PR #3154, opened 2022-11-07
Fix petrify not breaking on dmg taken

xkoredev:
**_I affirm:_**

Fixes petrification not breaking on damage taken.
Closes https://github.com/HorizonFFXI/HorizonXI-Issues/issues/345

Go to dunes lizzard camp. Spam mob with 3000 tp until it uses gaze. Get petrified and observe damage does not break petrification.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -7767,7 +7767,7 @@ void CLuaBaseEntity::takeDamage(int32 damage, sol::object const& attacker, sol::
     // Check for special flags which may prevent damage from waking up the target
     bool wakeUp        = true;
     bool breakBind     = true;
-    bool removePetrify = true;
+    bool removePetrify = false;
 
     // TODO: Unused in current code, needs testing; change type to sol::table?
     //       Find a way to make this better! (Optional keys as well?)


WinterSolstice8:
Player casted break [does remove on auto attacks](https://www.bg-wiki.com/ffxi/Break), though I don't know the history of this boolean at all. I don't really see any reason to not change this to default not breaking petrify, but we do need a way to detect that.

WinterSolstice8 (approved):

claywar (approved):

TeoTwawki:
```
    bool removePetrify = false;
```
```
        if (removePetrify)
        {
            PDefender->StatusEffectContainer->DelStatusEffect(EFFECT_PETRIFICATION);
        }
```

I'm confused how setting that false allows petrify to break when the code is checking for it to be true, bind and sleep checks work same way. 

Additionally, I expect we need something to calculate the rate the same as we do bind (though bind is wonky atm) but that's a future us problem (maths!)

xkoredev:
>     bool removePetrify = false;
        if (removePetrify)
        {
            PDefender->StatusEffectContainer->DelStatusEffect(EFFECT_PETRIFICATION);
        }
> I'm confused how setting that false allows petrify to break when the code is checking for it to be true, bind and sleep checks work same way.

> Additionally, I expect we need something to calculate the rate the same as we do bind (though bind is wonky atm) but that's a future us problem (maths!)

There is an override block down the line:

```
    if (flags != sol::lua_nil)
    {
        auto flag_map = flags.as<std::map<std::string, bool>>();

        wakeUp        = flag_map["wakeUp"];
        removePetrify = flag_map["removePetrify"];
        breakBind     = flag_map["breakBind"];
    }
```

What i changed is the default behavior when these are not set.

PR #3160, opened 2022-11-08
Fix players not standing up when damaged while /sit or /sitchair

Xaver-DaRed:
**_I affirm:_**

Players were not standing up/breakinf from their sitting animation when attacked. This makes it so they do.

Get hit while /sit or /shitchair (0-10)

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -4040,7 +4040,7 @@ namespace battleutils
         {
             case TYPE_PC:
             {
-                if (PDefender->animation == ANIMATION_SIT)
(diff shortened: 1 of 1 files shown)


claywar (approved):

src/map/utils/battleutils.cpp
```diff
@@ -4040,7 +4040,7 @@ namespace battleutils
         {
             case TYPE_PC:
             {
-                if (PDefender->animation == ANIMATION_SIT)
+                if (PDefender->animation == ANIMATION_SIT || (PDefender->animation >= ANIMATION_SITCHAIR_0 && PDefender->animation <= ANIMATION_SITCHAIR_10))
```
WinterSolstice8, line 4043: wait, this check was in `TakeSkillchainDamage`? I am struggling to comprehend why it was here to begin with, but the change is fine

WinterSolstice8 (approved):

PR #3163, opened 2022-11-09
Battlefield improvements

jmcmorris:
**_I affirm:_**

@hookstar is migrating battlefields on ASB and discovered a few things that needed to be fixed.
- Armoury Crates are not always at the end of a battlefield area mob group.
- Sometimes required key items need to be kept


WinterSolstice8 (approved):

scripts/globals/battlefield.lua
```diff
@@ -900,12 +900,16 @@ function Battlefield:onBattlefieldEnter(player, battlefield)
             if type(item) == 'table' then
                 for _, subitem in ipairs(item) do
                     if player:hasKeyItem(subitem) then
-                        player:delKeyItem(subitem)
+                        if (not self.requiredKeyItems.keep) then
```
claywar: No outer parens here (and below)

claywar (approved):
Looks good to me, just one styling issue on the lua side

scripts/globals/battlefield.lua
jmcmorris: haha these language switches sometimes get you. thanks!

PR #3165, opened 2022-11-09
[core] Fix mutex race condition in watchdog on cleanup

WinterSolstice8:
**_I affirm:_**

Fix mutex race condition where a thread join will never join because it's waiting for a lock that will never unlock via going out of scope

run xi_connect/xi_map, issue "exit" on the console and see it exit normally

### DIFF
--- a/src/common/watchdog.cpp
+++ b/src/common/watchdog.cpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
 ===========================================================================
 
   Copyright (c) 2022 LandSandBoat Dev Teams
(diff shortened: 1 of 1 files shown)


src/common/watchdog.cpp
```diff
@@ -34,11 +34,12 @@ Watchdog::~Watchdog()
 {
     if (m_running)
     {
-        std::unique_lock<std::mutex> lock(m_bottlneck);
+        m_bottlneck.lock();
```
zach2good: Can we achieve this same thing with some careful scoping? Manual lock/unlock is bad juju in my experience

zach2good (approved):

PR #3166, opened 2022-11-09
[core] Fix exit issues with Watchdog and ConsoleService

zach2good:
**_I affirm:_**

* Introduces `nonstd::jthread`; a C++11 shim for C++20's `std::jthread`. This is a thread that joins on its own destruction. Very handy.
* Replaces the `this_thread::sleep` calls inside ConsoleService with a `condition_variable`
* Refactors getLine's impls to constantly poll and retrieve individual letters the command line in a non-blocking way
* Introduced DEBUG_SOCKETS flag, to hide the incorrectly marked "only log on debug" messages about fds in sockets

On Windows and Linux:
- Start xi_connect
- CTRL+C -> You should exit gracefully

- Start xi_connect
- Type exit -> You should exit gracefully

### DIFF
--- a/src/common/console_service.cpp
+++ b/src/common/console_service.cpp
@@ -40,24 +40,11 @@ bool stdinHasData()
     // this works by harnessing Windows' black magic:
     return _kbhit();
 #else
-    // using a timeout of 0 so we aren't waiting:
-    struct timespec timeout
-    {
-        0l, 0l
-    };
+    struct pollfd pollFileDescriptor = { STDIN_FILENO, POLLIN, 0 };
 
-    // create a file descriptor set
-    fd_set fds{};
-
-    // initialize the fd_set to 0
-    FD_ZERO(&fds);
-    // set the fd_set to target file descriptor 0 (STDIN)
-    FD_SET(0, &fds);
-
-    // pselect the number of file descriptors that are ready, since
-    // we're only passing in 1 file descriptor, it will return either
-    // a 0 if STDIN isn't ready, or a 1 if it is.
-    return pselect(0 + 1, &fds, nullptr, nullptr, &timeout, nullptr) == 1;
+    // poll the stdin file descriptor until it tells us it's ready
+    // poll returns a 0 if STDIN isn't ready, or a 1 if it is.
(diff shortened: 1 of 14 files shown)


src/login/lobby.cpp
```diff
@@ -93,7 +93,7 @@ int32 lobbydata_parse(int32 fd)
         char* buff = &sessions[fd]->rdata[0];
         if (ref<uint8>(buff, 0) == 0x0d)
         {
-            ShowDebug("Posible Crash Attempt from IP: <%s>", ip2str(sessions[fd]->client_addr));
+            ShowWarning("Posible Crash Attempt from IP: <%s>", ip2str(sessions[fd]->client_addr));
```
WinterSolstice8: While we are here: `Possible`

zach2good:
Ok, I'm happy with the layout of this now, and it works fine on Windows, but on Linux it requires you to hit enter x2 for it to register the command. Would you mind taking a look @WinterSolstice8? I'm out for the next day or so, so feel free to do that force-push-merge thing 👍

src/common/console_service.cpp
```diff
@@ -120,91 +194,20 @@ ConsoleService::ConsoleService()
...
-
-            if (std::isprint(keyCharacter))
-            {
-                line += keyCharacter;
-            }
-        }
-    }
-// Linux has a proper polling system for stdin
-#else
-    struct pollfd pollFileDescriptor = { STDIN_FILENO, POLLIN, 0 };
```
zach2good, line 194: Maybe bringing back your impl with a 0 wait might have the same effect as what I'm using?

WinterSolstice8: I changed it to poll, but it had the same problem. I believe this is because windows returns stdin is ready for a character and linux returns stdin is ready on a newline via stdin file descriptors. I used std::getline on std::cin within an #IFDEF for non-windows and tested it, then added a commit.

zach2good:
LGTM 👍

zach2good:
Retested on Windows and Linux, looks good 👍

Xaver-DaRed (approved):

PR #3173, opened 2022-11-10
[lua] Remove hardcoded ids from Bibiki Bay

zach2good:
**_I affirm:_**

After a recent shift, most hardcoded mob (and npc) ids in Bibiki bay would have been made useless. This moves them all over to a pattern where we ask the zone for the entities by name and their relative index in the zone compared to other entities of the same name in that zone. These lookups protect these PH's, qm's, etc. from breaking during future shifts.

See comments


zach2good:
lol, update hasn't even landed yet...

scripts/zones/Bibiki_Bay/mobs/Jagil.lua
```diff
@@ -12,7 +11,14 @@ entity.onMobDeath = function(mob, player, optParams)
 end
 
 entity.onMobDespawn = function(mob)
-    xi.mob.phOnDespawn(mob, ID.mob.SERRA_PH, 10, 3600) -- 1 hour
+    -- PH is the 5th Jagil entry in the zone
+    local jagilId  = mob:getZone():queryEntitiesByName('Jagil')[5]:getID()
+    local shenId   = mob:getZone():queryEntitiesByName('Shen')[1]:getID()
```
TeoTwawki: Shen is not related to this PH or its NM

zach2good: Good catch 👍

kaincenteno:
nice so now we are quering by name instead of placing the npc ID inside the zone IDs?

TeoTwawki:
> Remove hardcoded ids from Bibiki Bay 

That is what the man said, Kain. 

Lookup by name was added awhile back and is gradually getting used more to avoid shifting IDs.

zach2good:
Tested by:
- Killing Shen and his filtrates: no errors
- Kill each of the PH mobs: no errors

WinterSolstice8 (approved):

PR #3178, opened 2022-11-11
Use MOBMOD_DETECTION for all detection logic

jmcmorris:

I previously added in `MOBMOD_DETECTION` to support changing the detection programatically. After encountering an issue with linking due to not using this mobmod I decided it would likely be best to have a single source for detection. Now detection is read into the mobmod from db and is only used that way.

Go get aggro


jmcmorris:
Just noticed an issue with this that I have no idea why we would see it now and not before. Any one got any ideas?
image

jmcmorris:
This ends up being an issue with mob initialization which I've made a new PR for: https://github.com/LandSandBoat/server/pull/3180

jmcmorris:
Aforementioned PR must be merged in order for this to pass CI

claywar (approved):

PR #3209, opened 2022-11-16
[lua] Change trade conditions for Zaldon

zach2good:
**_I affirm:_**

See changes for details

N/A

### DIFF
--- a/scripts/zones/Selbina/npcs/Zaldon.lua
+++ b/scripts/zones/Selbina/npcs/Zaldon.lua
@@ -512,6 +512,11 @@ local function tradeFish(player, fishId)
     local found   = false
     local sum     = 0
 
+    -- NOTE: We confirm the trade now, and not at the end of the cutscene as normal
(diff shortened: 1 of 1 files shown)


scripts/zones/Selbina/npcs/Zaldon.lua
```diff
@@ -512,6 +512,11 @@ local function tradeFish(player, fishId)
     local found   = false
     local sum     = 0
 
+    -- NOTE: We confirm the trade now, and not at the end of the cutscene as normal
+    --     : because the cutscene gives away whether or not the trade was successful
+    --     : or not, and it's possible for players to cheese this trade by force-dc-ing.
+    player:confirmTrade()
```
Xaver-DaRed, line 518: I would recommend moving this to line 519/524

This way, we would make sure the trade is actually valid before nuking it, but before the CS starts

claywar (approved):

PR #3211, opened 2022-11-16
[dbtool] Exception handling

cocosolos:
**_I affirm:_**

Fixes an issue with dbtool not falling back to default settings if user settings don't exist. Improve exception handling to be more useful by printing exception. Shouldn't need the lgtm comments anymore so removed those.

Delete `settings/network.lua`, run dbtool, get useful error message.

Update `settings/default/network.lua` credentials, run dbtool without error.


zach2good (approved):

zach2good:
Don't worry about `build / Linux_GCC11_64bit`, a bad commit got in and is now fixed in latest base

PR #3212, opened 2022-11-16
Fix packet overflow in /random caused by long player names of 15 chars

paladindamarus:
**_I affirm:_**

Addresses a buffer overflow issue when players with 15-character names use /random.  These packets are normally 0x30 bytes in length, but with a 15-character name, 0x31 bytes are required.  Retail captures indicate that an additional 4 bytes are added to the end of the packet to capture this condition.

Create a character with 15 chars for a name, and use /random.  Nearby players (this does not affect self) should see normal messages for /random rolls.  If not, they will see additional data past the end of the sentence.

### DIFF
--- a/src/map/packets/message_standard.cpp
+++ b/src/map/packets/message_standard.cpp
@@ -98,7 +98,7 @@ CMessageStandardPacket::CMessageStandardPacket(uint32 param0, uint32 param1, uin
 CMessageStandardPacket::CMessageStandardPacket(CCharEntity* PChar, uint32 param0, MsgStd MessageID)
 {
     this->setType(0x09);
-    this->setSize(0x30);
+    this->setSize(0x34);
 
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
`this->setSize(0x34);` and call it a day IMO, the client doesn't particularly care if size is too big, only too small
will let others chime in

paladindamarus:
I had thought of that, but I wasn't sure if checks would be made by the client otherwise so just went with "this is what retail does". :)

zach2good:
Id be on board for just setting it to the larger size, as long as it tests OK

paladindamarus:
I just tested locally with using a 0x34 size packet universally, with no crashing or issues that I can tell.  Happy to leave it like this, but if someone finds an issue with it down the road, it's an easy fix.

WinterSolstice8 (approved):

zach2good (approved):

PR #3219, opened 2022-11-17
Add "Inside the Belly" quest items to lua item list

Xaver-DaRed:
**_I affirm:_**

In preparation to the conversion of "Inside the Belly" quest to interaction system (and its mandatory cleanup) I've decided to add the items separately, to avoid possible conflicts as much as possible.

None


scripts/globals/items.lua
```diff
@@ -1280,16 +1310,31 @@ xi.items =
...
     SLICE_OF_COCKATRICE_MEAT        = 4435,
     BAKED_POPOTO                    = 4436,
     COBALT_JELLYFISH                = 4443,
     RARAB_TAIL                      = 4444,
     YAGUDO_CHERRY                   = 4445,
     SCREAM_FUNGUS                   = 4447,
     PUFFBALL                        = 4448,
+    SILVER_SHARK                    = 4451,
+    EMPEROR_FISH                    = 4454,
+    AHTAPOT                         = 4455,
```
TeoTwawki: Here we run into a fish naming problem:
we have `AHTAPOT                         = 5455,` at line 1466, there are basically 2 of every fish, one for what's in our inventory and another for special purposes. We need to document said purposes and adjust naming convention to account for it.

Xaver-DaRed: As discussed, for the time being, we will go with _ALT for the copies as a temporal measure, until we get the actual IDs used for fishing and the IDs used in mog garden

claywar: Just a heads up, this is going to really muck with tooling.  If we have two copies of each in the DB, please make sure they're named appropriately as well.

TeoTwawki: noted. we'll need to settle on a convention for duplicate fishes in general and update both places

Xaver-DaRed: I went through all the items again and it turns out that, while we do have dulicates in DB, this concrete case was a matter of me being dumb and screwing one number.

I've removed the duplicate entry. This should be good to go.

claywar (approved):

PR #3221, opened 2022-11-17
Fix pets lost when warping

### LINKED ISSUE #3220: 🐛 Pets Lost when Warping
**_I affirm:_**

Windows 10

## Branch affected by issue 

`base`

## Steps to reproduce

1. Be on Dragoon in a zone you can summon pets.
2. Call Wyvern.
3. Teleport through some method (e.g. Scroll of Instant Warp).
4. Move to a zone where pets are allowed.
5. Observe the wyvern did not return.

## Expected behavior

5. Wyvern should return.

DiscipleOfEris:
**_I affirm:_**

Make pets persist through warping in addition to other types of zoning.

Also fix console [critical] messages in debug mode related to jug pet code.

Note: I moved the `shouldPersistThroughZone()` pet helper function to `CCharEntity` rather than `CPetEntity`.

Fixes #3220

See #3220

Note: I am noticing an issue where the pet's stats are not preserved across zoning. I imagine this happens when `SaveCharStats()` is called *before* `setPetZoningInfo()`.

EDIT: Actually, I'm not able to reproduce the "pet's stats are not preserved across zoning". Maybe the stat update was delayed in one of my tests and that's why I thought it wasn't updating.

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -437,11 +437,11 @@ void CCharEntity::setPetZoningInfo()
         case PET_TYPE::AVATAR:
         case PET_TYPE::AUTOMATON:
         case PET_TYPE::WYVERN:
-            petZoningInfo.petLevel     = PPetEntity->getSpawnLevel();
-            petZoningInfo.petHP        = PPet->health.hp;
-            petZoningInfo.petTP        = PPet->health.tp;
-            petZoningInfo.petMP        = PPet->health.mp;
-            petZoningInfo.petType      = PPetEntity->getPetType();
+            petZoningInfo.petLevel = PPetEntity->getSpawnLevel();
+            petZoningInfo.petHP    = PPet->health.hp;
+            petZoningInfo.petTP    = PPet->health.tp;
+            petZoningInfo.petMP    = PPet->health.mp;
+            petZoningInfo.petType  = PPetEntity->getPetType();
             break;
         default:
             break;
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -431,6 +431,8 @@ void CCharEntity::setPetZoningInfo()
             {
                 break;
             }
+            petZoningInfo.jugSpawnTime = PPetEntity->getJugSpawnTime();
+            petZoningInfo.jugDuration  = PPetEntity->getJugDuration();
             [[fallthrough]];
         case PET_TYPE::AVATAR:
         case PET_TYPE::AUTOMATON:
@@ -440,8 +442,6 @@ void CCharEntity::setPetZoningInfo()
             petZoningInfo.petTP        = PPet->health.tp;
             petZoningInfo.petMP        = PPet->health.mp;
             petZoningInfo.petType      = PPetEntity->getPetType();
-            petZoningInfo.jugSpawnTime = ((CPetEntity*)PPet)->getJugSpawnTime();
-            petZoningInfo.jugDuration  = ((CPetEntity*)PPet)->getJugDuration();
             break;
         default:
             break;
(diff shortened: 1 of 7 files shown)


src/map/entities/charentity.cpp
```diff
@@ -462,6 +462,32 @@ void CCharEntity::resetPetZoningInfo()
     petZoningInfo.jugSpawnTime = 0;
     petZoningInfo.jugDuration  = 0;
 }
+
+bool CCharEntity::shouldPetPersistThroughZoning()
+{
+    PET_TYPE petType;
+    auto     PPetEntity = dynamic_cast<CPetEntity*>(PPet);
+
+    if (PPetEntity == nullptr && !petZoningInfo.respawnPet)
```
WinterSolstice8, line 471: Should this be an `||`?

src/map/entities/petentity.cpp
```diff
@@ -286,8 +278,12 @@ void CPetEntity::loadPetZoningInfo()
         health.tp = static_cast<uint16>(master->petZoningInfo.petTP);
         health.hp = master->petZoningInfo.petHP;
         health.mp = master->petZoningInfo.petMP;
-        setJugDuration(master->petZoningInfo.jugDuration);
-        setJugSpawnTime(master->petZoningInfo.jugSpawnTime);
+
+        if (m_PetType == PET_TYPE::JUG_PET)
```
xkoredev, line 282: nit: this check is not necessary as `petZoningInfo.jugDuration` will just contain default values for non jug pets

src/map/entities/charentity.cpp
```diff
@@ -462,6 +462,32 @@ void CCharEntity::resetPetZoningInfo()
...
+    }
+    else // petZoningInfo.respawnPet == true
+    {
+        petType = petZoningInfo.petType;
+    }
+
+    return petType == PET_TYPE::WYVERN ||
+           petType == PET_TYPE::AVATAR ||
+           petType == PET_TYPE::AUTOMATON ||
+           (petType == PET_TYPE::JUG_PET && settings::get<bool>("map.KEEP_JUGPET_THROUGH_ZONING"));
```
xkoredev, line 488: Thanks for adding the check. Should we add the `KEEP_JUGPET_THROUGH_ZONING` default value to settings/default/map.lua?

src/map/entities/charentity.cpp
```diff
@@ -431,17 +431,17 @@ void CCharEntity::setPetZoningInfo()
             {
                 break;
             }
+            petZoningInfo.jugSpawnTime = PPetEntity->getJugSpawnTime();
```
xkoredev, line 434: good catch. In theory this should be harmless for all pets, but I did add some xi_debug breaks so that was probably triggering it.

DiscipleOfEris: Yes it was those debug breaks that I moved the jugpet code around for.

src/map/entities/charentity.cpp
DiscipleOfEris: This check will exit with `false` early if we have no pet *and* there's no stored pet (`respawnPet == true`).

If we have a pet, then we want to check it. If we have a stored pet, then again we want to check it. It's only when we have neither that we assume there's nothing to persist.

Hypothetically, if `respawnPet` is set then we already did `setPetZoningInf()` which should always be after `shouldPetPersistThroughZoning()` returned true, but I didn't want to assume that was always the case.

src/map/entities/charentity.cpp
DiscipleOfEris: I already added one a while ago. [settings/default/map.lua#L254](https://github.com/LandSandBoat/server/blob/base/settings/default/map.lua#L254)

src/map/entities/charentity.cpp
WinterSolstice8: Gotcha, thanks.

WinterSolstice8 (approved):

src/map/entities/petentity.cpp
DiscipleOfEris: I put that check there because of debug breaks throwing `[critical]` errors to console if the pet wasn't a jug pet. I agree the functions are harmless, but so long as we have the debug check inside those functions, we should keep the check here.

src/map/entities/charentity.cpp
xkoredev: Ah nice. I didn't know this or I would have used it. Thanks.

claywar (approved):

PR #3225, opened 2022-11-17
DRAFT: Eschan_Portal_#1 (Escha-Ru'Aun)

Daywhite83:
Gets the portals working, the catch is that all portals are open by default, this will need some input from others to fix. Once this one gets approved i can upload the others. (have them all made for all 3 zones)

**_I affirm:_**

Gets the portals working, the catch is that all portals are open by default, this will need some input from others to fix.
Once this one gets approved i can upload the others. (have them all made for all 3 zones)

I added some comments into the file explaining some of the things i have observed, which i can remove if needed.

Go to Escha - Ru'Aun and use Eschan Portal No.1. (make sure you have at least 50 silt)

### DIFF
--- a/scripts/zones/Escha_RuAun/npcs/Eschan_Portal_#1
+++ b/scripts/zones/Escha_RuAun/npcs/Eschan_Portal_#1
@@ -25,7 +25,9 @@ entity.onEventUpdate = function(player, csid, option)
 end
 
 entity.onEventFinish = function(player, csid, option)
-    player:delCurrency("escha_silt", siltCost)
+    if option == 1 then
+        player:delCurrency("escha_silt", siltCost)
+    end
 end
 
 return entity
--- a/scripts/zones/Escha_RuAun/npcs/Eschan_Portal_#1
+++ b/scripts/zones/Escha_RuAun/npcs/Eschan_Portal_#1
@@ -1,0 +1,31 @@
+-----------------------------------
+-- Area: Escha - Ru'Aun
+--  NPC: Eschan Portal #1
+-----------------------------------
+require("scripts/zones/Escha_RuAun/IDs")
+require("scripts/globals/npc_util")
+require("scripts/globals/items")
+require("scripts/globals/utils")
+-----------------------------------
+local entity = {}
+local siltCost = 50
+
+entity.onTrade = function(player, npc, trade)
+end
+
+entity.onTrigger = function(player, npc)
+    -- (Event ID, ???, portal array, zoneID, currentPortalIndex, ???, playersSilt, siltCost, ???)
+    -- The portal array seems to effect what portals are displayed -1 shows all, anything else i couldnt get anything to pop.
+    -- The silt cost by default is 50 silt per port, it can be reduced somehow but im not sure how, perhaps a KI
+    -- currentPortalIndex 0-7 are used in Escha-Zi'Tah, 8 = portal 1 in escha-ru'aun 9 = portal 2 etc.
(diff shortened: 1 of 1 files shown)


scripts/zones/Escha_RuAun/npcs/Eschan_Portal_#1
```diff
@@ -0,0 +1,31 @@
...
+    -- The silt cost by default is 50 silt per port, it can be reduced somehow but im not sure how, perhaps a KI
+    -- currentPortalIndex 0-7 are used in Escha-Zi'Tah, 8 = portal 1 in escha-ru'aun 9 = portal 2 etc.
+    player:startEvent(9100, 0, -1, 289 , 8, 0, player:getCurrency("escha_silt") , siltCost, 0)
+end
+
+entity.onEventUpdate = function(player, csid, option)
+end
+
+entity.onEventFinish = function(player, csid, option)
+    player:delCurrency("escha_silt", siltCost)
```
zach2good: If you bail out of the event, this cost will still be taken. You need to check against option

cocosolos:
It looks like the portal array is a bit field where bits 0-7 are Zi'Tah, 8-22 are Ru'Aun, and the rest are Reisenjima. e.g. 11111101 = 253 = Zi'Tah all portals unlocked except portal 2. Probably needs a new column in char_unlocks and everything that comes with that. Silt cost looks like its effected by zone blessings from domain invasion.

cocosolos:
Also it looks like option 5 after `currentPortalIndex` is a bool to trigger the unlock message (Light begins to shimmer and twist as you pass your hand over it.)

Edit: Not a bool, another bitfield where bit 0 is the unlock message, bit 1 gives the option for teleport with ethereal droplet, bit 2 gives option for scintillating rhapsody.

Daywhite83:
option 5 seems to be the option for eschan droplets 3 = have it, 0 = no, 1 seem to have no effect here or it also means no. 2 also seems like a yes, but from captured data it used bit 3 so thats what i did as well

the portals feel retail accurate on my end now, fixed being able to see all portals by default, and they seem to unlock as you open them by retail methods. here are the portals 2-15 if u want to test for yourself.
[EschanPortals2-15.zip](https://github.com/LandSandBoat/server/files/10044838/EschanPortals2-15.zip)

cocosolos:
All 3 Escha zones share the same portal unlock variable. It is an [array of bits](https://en.wikipedia.org/wiki/Bit_array) where each bit gets flipped as you unlock the corresponding portal. Option 5 is also comprised of bits. If none of the bits are set then the value is 0 and you get the normal menu. If the first bit is set the value is 1 and you get the portal unlock message. If the first bit is not set and the second bit is, the value is 2 and you get the eschan droplets option without the unlock message. If both the first and second bits are set the value is 3 and you get the unlock message and the eschan droplets option.
image

To implement this properly you'll probably need to make use of some of the Lua bitwise operator functions like `bit.bor`, `bit.band`, `bit.lshift`, `bit.rshift`, etc.

The eschan droplets option should only be enabled if the player _has_ eschan droplets from the mission, and care needs to be taken to give and take them properly as the player progresses.

The portals should be handled the same way homepoints, survival guides, abyssea confluxes, waypoints and other unlockables are handled in the char_unlocks table of the database and using the `addTeleport` and `getTeleport` functions, not with a charvar. This will require a little C++ work to add escha portal fields as well as adding it to the SQL file and a Python migration.

You'll also want to check and follow the [project Lua style guide](https://github.com/LandSandBoat/server/wiki/Development-Guide-Lua).

Xaver-DaRed:
This seems abandoned. It's also obsolete by now.

PR #3228, opened 2022-11-18
Keep (gm) speed when zoning / relogging

xkoredev:
**_I affirm:_**

Preserves the value of the last saved speed via !speed command, so that it persists after zoning or relogging

!speed 255. Zone or warp.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -329,6 +329,12 @@ void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PCh
             PChar->m_charHistory.mhEntrances++;
             gardenutils::UpdateGardening(PChar, false);
         }
+
+        // Restore speed saved by !speed command (if set and gm char)
+        auto savedSpeed = PChar->getCharVar("speed");
+        if (savedSpeed > 0 && PChar->m_GMlevel > 0) {
+            PChar->speed = std::clamp<uint8>(savedSpeed, 0, 255);
+        }
     }
 
     // Only release client from "Downloading Data" if the packet sequence came in without a drop on 0x00D
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5167,6 +5167,7 @@ void CLuaBaseEntity::setSpeed(uint8 speedVal)
         if (m_PBaseEntity->objtype == TYPE_PC)
         {
             auto* PChar = static_cast<CCharEntity*>(m_PBaseEntity);
+            PChar->setCharVar("speed", speed);
             PChar->pushPacket(new CCharUpdatePacket(PChar));
         }
         else


src/map/packet_system.cpp
```diff
@@ -329,6 +329,12 @@ void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PCh
             PChar->m_charHistory.mhEntrances++;
             gardenutils::UpdateGardening(PChar, false);
         }
+
+        // Restore speed saved by !speed command (if set and gm char)
+        auto savedSpeed = PChar->getCharVar("speed");
```
zach2good, line 334: While I think this idea is useful, I don't like the idea of introducing a somewhat temporary charvar just for the benefit of GMs.

In this current layout, it will be adding a DB read for every character who zones who isn't a GM.

You could get around this by checking for GM status first, and then doing the read/compare/set of speed.

But that still leaves us with a charvar, only used by GMs, only while zoning, which I'm not super keen on.

I wonder if it's possible to just add a mod for speed to the GM inside !speed, and if that will persist while zoning

xkoredev: That sounds like a cleaner idea. I'll look into the mod.

TeoTwawki: Unfortunately, modifiers are lost and then reapplied in status effects on zoning, which was why godmode used a charVar.

TeoTwawki: Worth noting that GMs could just give themselves flee is they just "gotta go fast" and don't want to retype command every time they zone. There's a command to add effects by name and infinite duration is possible.

zach2good: Thats a great idea, does flee have a strength? We can apply `!speed 100`: 100 as the strength, and duration for infinite

TeoTwawki: sort of: we do flee wrong compared ot retail, and we sort of have to in order to support retail+oldschool+custom base speeds.

Retail flee just statically sets a speed of 80 (it was 80 before they changed things from 40 to 50, and it is still 80 now, so contrary to popular beliefs and #shitWikiSays, it is not simply "double speed" or a %)

Presently here on LSB we do `target:addMod(xi.mod.MOVE, effect:getPower())` in our flee script, and:
```lua
    local duration = 30 + player:getMod(xi.mod.FLEE_DURATION)
    player:addStatusEffect(xi.effect.FLEE, 100, 0, duration)
```
in the job ability. with 100 being the wrong value for both era and modern. Considering how many servers modifier base speed, we should probably check and involve it there. Maybe just add 40 and call it close enough. A PR tried to make a a % and I super don't like that (and it's still wrong as a %).

...So anyway yes you can set the power of flee. This might be worth just having a separate command for flee for GMs to use.

xkoredev: Clever. So new command and leave speed alone. Maybe just hardcode the 40?

TeoTwawki: > Clever. So new command and leave speed alone. Maybe just hardcode the 40?

`!flee <optional param defaulting to 240 maybe>` -> `player:addStatusEffect(xi.effect.FLEE, input, 0, 0)` should do it

xkoredev: Struggling to find the right power here. Seems like speed caps at certain point, way under 255

TeoTwawki: > Struggling to find the right power here. Seems like speed caps at certain point, way under 255

I thought it stopped at 240. it's an unsigned int8 that's bitpacked with a couple flags. Try 200 even I know I've used value higher than that.

xkoredev: 200 seems like !speed 100 or so.
Someone on discord suggested a GM earring that adds a speed mod. I actually thought adding a charvar would be simpler so I opened this PR, but earring is starting to sound like the better approach now haha

TeoTwawki: > 200 seems like !speed 100 or so. 

Something very wrong there on my own server I see a clear diff between 100 and 200 and I typically lazy tap out 222 when I `!speed` command and things work as I expect. I've done some pretty extensive testing on how the client uses speed and speedsub too.

I wonder if we broke clampage of mod move somewhere at some point. speed command straight overwrites base speed, mob move is added into it (base + mod = new speed)

TeoTwawki: I wonder if we broke clampage of mob move soemwhere at some point. speed command straight overwrites base speed, mob move is added into it (base+mob= new speed)

TeoTwawki: *moD

I keep typo'ing that B in there

TeoTwawki: poke me on discord in the reddit servers discord if you can @xkoredev

TeoTwawki: so after having kore do a few test for me, the behavior seen lead me to a little digging and I found I missed or forget a change and it does NOT do the addition I thought. Someone adjusted the behavior I don't know how long ago to be correct for gear mods: as a percentage. `float modAmount     = (100.0f + static_cast<float>(getMod(mod))) / 100.0f;`

Unfortunately, that's not correct for flee. But that's why he wasn't able to push flee power higher, it was capping out at 100%

TeoTwawki: since I sent everyone down this rabbit hole, I will look into adding what's needed later this week.

TeoTwawki: @xkoredev I made several attempts at sorting out speed, failed, and then Xaver pulled off most of the major corrections I was aiming for recently. looking over still open pr's this morning and re-reading the conversation here reminded me.

PR #3231, opened 2022-11-19
Fix imagery types so they cannot be applied more than once globally

xkoredev:
**_I affirm:_**

- Fixes synth imagery statuses so they cannot be applied more than once.
- Closes https://github.com/AirSkyBoat/AirSkyBoat/issues/1209

See ticket


TeoTwawki (approved):
somebody used "type" correctly. how did that happen? /s

PR #3234, opened 2022-11-19
[lua] Fix Steam Clock not getting consumed on trade during quest

WinterSolstice8:
**_I affirm:_**

Adds a line to consume trade items to The Cold Light of Day interaction

Trade steam clock to Malene, see steam clock get consumed after getting 500 gil


claywar (approved):
No reward that could prevent the completion, so looks good to me.  If there's an item reward, or something that could fail, keep in mind that you'd want to compare the complete first (ex below)

```
    player:confirmTrade()
end
```

PR #3238, opened 2022-11-20
Fix Godmode crashing mordion

xkoredev:
**_I affirm:_**

- Fixes Hundred Fists onLose callback crashing mordion
- Adds a NoopController to be used instead of setting controller to null
- Jailed players use a NoopController rather than a null one
- AutomatonEntity use a NoopController rather than a null one

!godmode
!jail <me>

### DIFF
--- a/src/map/status_effect_container.cpp
+++ b/src/map/status_effect_container.cpp
@@ -488,6 +488,12 @@ bool CStatusEffectContainer::AddStatusEffect(CStatusEffect* PStatusEffect, bool 
         return false;
     }
 
+    if (m_POwner->PAI->GetController() == NULL)
+    {
+        ShowWarning("status_effect_container::AddStatusEffect Owner PAI has null controller");
+        return false;
+    }
+
     uint16 statusId = PStatusEffect->GetStatusID();
 
     if (statusId >= MAX_EFFECTID)
--- a/src/map/utils/jailutils.cpp
+++ b/src/map/utils/jailutils.cpp
@@ -19,19 +19,16 @@
 ===========================================================================
 */
 
+#include <memory>
+
 #include "jailutils.h"
 
 #include "../conquest_system.h"
 #include "../entities/charentity.h"
 
 #include "../ai/ai_container.h"
+#include "../ai/controllers/noop_controller.h"
 #include "../ai/controllers/player_controller.h"
-
-/************************************************************************
- *                                                                       *
- *                                                                       *
- *                                                                       *
- ************************************************************************/
 
 namespace jailutils
 {
@@ -43,15 +40,11 @@ namespace jailutils
 
     void Add(CCharEntity* PChar)
     {
-        PChar->PAI->SetController(nullptr);
-
-        // TODO:
+        PChar->PAI->SetController(std::make_unique<NoopController>(PChar));
     }
 
     void Del(CCharEntity* PChar)
     {
         PChar->PAI->SetController(std::make_unique<CPlayerController>(PChar));
-
-        // TODO:
     }
 }; // namespace jailutils
--- a/src/map/ai/controllers/CMakeLists.txt
+++ b/src/map/ai/controllers/CMakeLists.txt
@@ -5,6 +5,8 @@ set(AI_CONTROLLER_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/controller.h
     ${CMAKE_CURRENT_SOURCE_DIR}/mob_controller.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/mob_controller.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/noop_controller.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/noop_controller.h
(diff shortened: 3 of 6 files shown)


src/map/status_effect_container.cpp
```diff
@@ -488,6 +488,11 @@ bool CStatusEffectContainer::AddStatusEffect(CStatusEffect* PStatusEffect, bool
         return false;
     }
 
+    if (m_POwner->PAI->GetController() == NULL) {
```
xkoredev: This is where Controller was null when zoning into jail, which caused the newly added onEffectLose side effect of hundred fists to crash.
This should not happen anymore with the noop controller, but just to avoid a future crash, good to have the check here.

src/map/utils/jailutils.cpp
```diff
@@ -43,15 +40,11 @@ namespace jailutils
 
     void Add(CCharEntity* PChar)
     {
-        PChar->PAI->SetController(nullptr);
```
xkoredev, line 46: This was the crux of the problem. Setting controller to null was harmless.... until it wasn't

xkoredev:
@zach2good @TeoTwawki

TeoTwawki:
I'm unable to produce this crash on stock.

I tried it with my char both jailed and not jail (just in the zone). I tried toggling hundred fists before the zone in, and after while in zone. Nada. no kabooms. There was supposed to be a vanadiel shattering kaboom!

What are your reproduction steps for the bug you are trying to fix? 

Did you successfully reproduce this on a clean build of LSB or just on a fork? 

I have been made aware of a change on ASB that I can all but guarantee will produce a crash, and suspect the crash you are attempting to fix _**could**_ be that exact crash.

TeoTwawki:
> @zach2good @TeoTwawki

@xkoredev its not unusual for things to sit for random looking durations anywhere from 30 seconds to weeks, we all have other things to do outside of this game and project, which is a hobby. Sometimes even our own work can longer for a long time before we can help each other out / read it.

TeoTwawki:
p.s. I also tried jailing a separate char from my gm while they had hundred fists up. no crash.

Xaver-DaRed:
https://github.com/AirSkyBoat/AirSkyBoat/blob/staging/scripts/globals/effects/hundred_fists.lua

Those flags shouldn't be there. They don't account for the target being a player, for starters.

I would place them in a mixin or in the mobskill script IF I wanted it to handle it in lua

TeoTwawki:
> https://github.com/AirSkyBoat/AirSkyBoat/blob/staging/scripts/globals/effects/hundred_fists.lua
> 
> Those flags shouldn't be there. They don't account for the target being a player, for starters.
> 
> I would place them in a mixin or in the mobskill script IF I wanted it to handle it in lua

I wouldn't lua it, but if I did that's where it'd happen, as that is already mob-specific eliminating the need to go "is this a mob?" every time the effect script runs. The core already has similar checks in mob AI directly for chainspell iirc, but there probably aren't many examples of mobs that have both casting and hundred fists. I've for sure seen them mobskill during\* it ~~so that's not even a retail accurate change.~~

\*Edit: it's a low occurrence possibly where I've seen it is the exception not the rule, or possibly they hold for max tp. Because now that I think about it the ones in dyna seem to wait till after but lolMimas will just go if he has the tp, but he usually doesn't have tp

TeoTwawki:
All I know for sure on this pr is I could not repro the crash on my testbed.

xkoredev:
> All I know for sure on this pr is I could not repro the crash on my testbed.

It's possible that the change in https://github.com/AirSkyBoat/AirSkyBoat/blob/staging/scripts/globals/effects/hundred_fists.lua was only in ASB. It just happened to expose the fact that we do have a null controller, which if not surfaced by this usage (albeit incorrect), would potentially be surfaced by something else in the future.

xkoredev:
> @xkoredev its not unusual for things to sit for random looking durations anywhere from 30 seconds to weeks, we all have other things to do outside of this game and project, which is a hobby. Sometimes even our own work can longer for a long time before we can help each other out / read it.

I hope the ping wasn't too much. Coincidentally, when I started working with ASB / LSB, the review cadence in LSB had been really fast. This was probably just due to the time you all wonderful people were able to dedicate to the project at that time. In my brain, this was just the cadence year round since that's all I had seen. 

I have no problem waiting for things to be done right. Just need to know "at which point do I ping code owners?".

TeoTwawki:
That change is only in asb, and its a situation I would ***expect*** to crash if it makes its way to the core and should not be patched around. Thats not supposed to work. if anything, the bindings should silently bail out with error text rather than even get there. (spoiler: they just pass it on without checking anything)

TeoTwawki:
> the review cadence in LSB had been really fast.

blind luck :)

some days are slow workdays that give me more time to look at things. can't speak to anyone else's schedule so I dunno details, but sometimes we're plowing through PRs like nobody's business and other times we're away for a couple weeks.

xkoredev:
From discussion here and in the ASB PR, and since https://github.com/LandSandBoat/server/pull/3288 should fix any potential crashes from calling those methods, we can close this now.

Thanks for going through it!

PR #3243, opened 2022-11-21
Mac build improvements

xkoredev:
**_I affirm:_**

- Disables LTO in mac builds
- Adds binary configurations for linux in vscode.

Build on mac


WinterSolstice8 (approved):
lgtm but I don't use vscode so the approval is contingent on someone who does chiming in on that part

zach2good (approved):

zach2good:
> "MIMode": "lldb",

What voodoo is this? I'd wager that most non-apple Unix users are using gdb instead of lldb. So in this case I'd either rename teh categories to "OSX" or "Unix/clang" to make sure that gdb users still have room to operate

xkoredev:
> > "MIMode": "lldb",
> 
> What voodoo is this? I'd wager that most non-apple Unix users are using gdb instead of lldb. So in this case I'd either rename teh categories to "OSX" or "Unix/clang" to make sure that gdb users still have room to operate

Good point. I'd rather not reuse the unix tasks for osx anyways so that makes sense.

Edit: Renamed

zach2good (approved):

PR #3253, opened 2022-11-22
Correct "Errant Houppelande" recipe skills

TonzoffunCE:
JP-wiki has this recipe as Goldsmithing, not Smithing.

http://wiki.ffo.jp/html/7275.html

**_I affirm:_**

Fixes the crafting recipe for the Errant/Mahatma Houppelande to reflect the recipe on jp-wiki.

Edit this sql to have the lvl "50" in the correct skill slot (Goldsmithing not Smithing).


TeoTwawki (approved):
https://www.bg-wiki.com/index.php?title=Errant_Hpl.&type=revision&diff=663820&oldid=451897

old 'clopedia agrees with JP wiki to (and it's kind of obvious just looking at the material) so I've corrected the same mistake at BG wiki. The database error likely stems from someone miscounting fields or using BG as their primary source. BG was used as primary source for many of these via a semi-automated scrape.

Xaver-DaRed (approved):

PR #3264, opened 2022-11-25
Revert "Initial Framework for Cutscene Opt Params"

zach2good:
Reverts LandSandBoat/server#3227

Unmerging this, it doesn't appear to be properly tested against the interaction framework :/


zach2good:
Clay has offered to step in and fix the things 🙏

PR #3270, opened 2022-11-25
Adds Ecphoria Ring Functionality

TonzoffunCE:
This item cures Amnesia status.

**_I affirm:_**

This ring cures Amnesia status effect, this adds it's functionality.

### DIFF
--- a/scripts/globals/items/ecphoria_ring.lua
+++ b/scripts/globals/items/ecphoria_ring.lua
@@ -1,0 +1,23 @@
+-----------------------------------------
+-- ID: 15817
+-- Item: Ecphoria Ring
+-- Item Effect: This item remedies amnesia.
+-----------------------------------------
+require("scripts/globals/status")
+-----------------------------------------
+
+local itemObject = {}
+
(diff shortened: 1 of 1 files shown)


TonzoffunCE:
inb4: After posting this, I realized that this ring should not clear ALL amnesia status effects. It should not clear amnesia from Ladybugs or Rafflesia.

TeoTwawki:
Don’t actually have to check “has” before deletion - delStatusEffect was made smart enough to just do nothing if the effect isn’t there.

If you need to omit some sources, those sources need to use the tier field and then the item script can do getStatusEffect():getTier() to check it

TeoTwawki:
Don’t actually have to check “has” before deletion - delStatusEffect was made smart enough to just do nothing if the effect isn’t there.

If you need to omit some sources, tgose sources need to use the tier field and then the item script can do getStTusEffect()getTeir() to check it

TeoTwawki:
Ugh..posting from mobile and can’t edit to spell that correctly atm

TeoTwawki:
superseded by #4731

PR #3274, opened 2022-11-26
Adds lua binding for checking difficulty of a mob

Frankie-hz:
**_I affirm:_**

- Adds a lua entity so you can check the difficulty of a mob

This will return the value of the following enum:
image

This can be used in several cases where you need to check the difficulty of the mob in order to do something specific.

As an example add the line:
`local difficulty = player:checkDifficulty(target)`
`if difficulty > 0 then`
`    *do something special if mob is not too weak*`
`end`

If the mob is IEP is will return 1, EP 2, etc.

### DIFF
--- a/scripts/globals/mobs.lua
+++ b/scripts/globals/mobs.lua
@@ -531,10 +531,10 @@ xi.mob.difficulty =
     INCREDIBLY_EASY_PREY = 1,
     EASY_PREY            = 2,
     DECENT_CHALLENGE     = 3,
-    EVENMATCH            = 4,
+    EVEN_MATCH           = 4,
     TOUGH                = 5,
-    VERYTOUGH            = 6,
-    INCREDIBLYTOUGH      = 7,
+    VERY_TOUGH           = 6,
+    INCREDIBLY_TOUGH     = 7,
     MAX                  = 8,
 }
 xi.mob.diff = xi.mob.difficulty
--- a/scripts/globals/mobs.lua
+++ b/scripts/globals/mobs.lua
@@ -520,3 +520,21 @@ xi.mob.onAddEffect = function(mob, target, damage, effect, params)
 
     return 0, 0, 0
 end
+
+-----------------------------------
+-- mob difficulty enums for checkDifficulty()
+-----------------------------------
+
+xi.mob.difficulty =
+{
+    TOO_WEAK             = 0,
+    INCREDIBLY_EASY_PREY = 1,
+    EASY_PREY            = 2,
+    DECENT_CHALLENGE     = 3,
+    EVENMATCH            = 4,
+    TOUGH                = 5,
+    VERYTOUGH            = 6,
+    INCREDIBLYTOUGH      = 7,
+    MAX                  = 8,
+}
+xi.mob.diff = xi.mob.difficulty
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -9150,15 +9150,15 @@ bool CLuaBaseEntity::checkKillCredit(CLuaBaseEntity* PLuaBaseEntity, sol::object
  ************************************************************************/
 uint8 CLuaBaseEntity::checkDifficulty(CLuaBaseEntity* PLuaBaseEntity)
 {
-    CMobEntity*  PMob  = static_cast<CMobEntity*>(PLuaBaseEntity->GetBaseEntity());
-    CCharEntity* PChar = static_cast<CCharEntity*>(m_PBaseEntity);
+    CMobEntity*  PMob  = dynamic_cast<CMobEntity*>(PLuaBaseEntity->GetBaseEntity());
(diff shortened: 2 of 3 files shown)


jmcmorris (changes requested):
Looks good!

I'd like to see the enum `xi.mob.difficulty` added to Lua so that checks can use something like `if player:checkDifficulty(mob) == xi.mob.difficulty.EASY_PREY then`. This would go in `mob.lua` global.

src/map/lua/lua_baseentity.cpp
```diff
@@ -9142,6 +9142,26 @@ bool CLuaBaseEntity::checkKillCredit(CLuaBaseEntity* PLuaBaseEntity, sol::object
...
 
+/************************************************************************
+ *  Function: checkDifficulty()
+ *  Purpose : Checks the mobs difficulty
+ *  Example : local difficulty = player:checkDifficulty(mob)
+ *  Notes   : Returns a value based on the enum EMobDifficulty
+ ************************************************************************/
+uint8 CLuaBaseEntity::checkDifficulty(CLuaBaseEntity* PLuaBaseEntity)
+{
+    CMobEntity*  PMob  = static_cast<CMobEntity*>(PLuaBaseEntity->GetBaseEntity());
```
zach2good: These need to be dynamic casts, so that if the cast is invalid it'll end with nullptr and the next check will fail

src/map/lua/lua_baseentity.cpp
```diff
@@ -9142,6 +9142,26 @@ bool CLuaBaseEntity::checkKillCredit(CLuaBaseEntity* PLuaBaseEntity, sol::object
...
+    CCharEntity* PChar = static_cast<CCharEntity*>(m_PBaseEntity);
+
+    if (PChar && PMob)
+    {
+        return (uint8)charutils::CheckMob((PChar->GetMLevel()), (PMob->GetMLevel()));
+    }
+
+    ShowError("lua::checkDifficulty value is not valid");
```
zach2good: This should read CLuaBaseEntity, not just lua

zach2good:
+1 on what jmcm said

src/map/lua/lua_baseentity.cpp
claywar: We shouldn't need to prefix at all, as logging will automatically append the function its coming from

zach2good: Oh true!

Frankie-hz:
Not sure why the linux build is failing? Any idea?

Xaver-DaRed:
This is only meant for player to mob interaction, correct?

For instance, what if a trust, a summon, an automaton or a pet or a charmed mob check against a mob?

Frankie-hz:
> 

The intention is to be able to call this for a summon, pet, trust, or any entity against the mob its attacking. So if a pet that is summoned by a 75 bst is only 45, it should have the ability to check the pets level against the mob its attacking.

TeoTwawki:
but do you want the actual level difference or to know what expcheck would be?

TeoTwawki (approved):
regardless of the intended use, no objection to the enum + binding from me

Xaver-DaRed:
> > 
> 
> The intention is to be able to call this for a summon, pet, trust, or any entity against the mob its attacking. So if a pet that is summoned by a 75 bst is only 45, it should have the ability to check the pets level against the mob its attacking.

But the way it's written, it expects the user to be a player.

`CCharEntity* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity);`

I guess you could feed the owner on the lua side, but some entities, like npc allies in some battlefields, as an example, do not have an owner

TeoTwawki:
@Xaver-DaRed presumably you would use the pet's owner though, which would be the player

scripts/globals/mobs.lua
```diff
@@ -520,3 +520,21 @@ xi.mob.onAddEffect = function(mob, target, damage, effect, params)
...
+xi.mob.difficulty =
+{
+    TOO_WEAK             = 0,
+    INCREDIBLY_EASY_PREY = 1,
+    EASY_PREY            = 2,
+    DECENT_CHALLENGE     = 3,
+    EVENMATCH            = 4,
+    TOUGH                = 5,
+    VERYTOUGH            = 6,
+    INCREDIBLYTOUGH      = 7,
```
jmcmorris:
```suggestion
    EVEN_MATCH            = 4,
    TOUGH                = 5,
    VERY_TOUGH            = 6,
    INCREDIBLY_TOUGH      = 7,
```

jmcmorris (commented):
Just need a few underscores and I'm happy!

scripts/globals/mobs.lua
TeoTwawki: good catch, I missed that entirely in my read through

zach2good:
Looks good to me, just needs what jmcmorris said

jmcmorris (approved):

PR #3289, opened 2022-11-29
Save essential pet modifiers in case it's affected by level restriction

jmcmorris:
**_I affirm:_**

Fixes pets taking 0 melee damage after being affected by level restriction.
When a pet undergoes level restriction its modifiers are cleared and then rebuilt; however, the melee SDT mods are set in CBattleEntity ctor and are not reset elsewhere. Consequently they end up being 0 after level restriction. The easiest fix here is to save the modifiers immediately after the pet object is created.

Change to a pet job, get a pet out and then do `!addeffect level_restriction 40` and then check your pet's `!getmod SLASH_SDT`. It should be 1000.

### DIFF
--- a/scripts/globals/pets/wyvern.lua
+++ b/scripts/globals/pets/wyvern.lua
@@ -110,14 +110,6 @@ end
 entity.onMobSpawn = function(mob)
     local master = mob:getMaster()
 
-    -- https://www.bg-wiki.com/ffxi/Wyvern_(Dragoon_Pet)#Combat_Stats
-    -- innate -40% DT, which does not contribute to the -50% cap (this is a unique attribute to pets having a "higher" DT cap)
-    -- TODO: need "UDMG" modifier or equivalent
-    mob:addMod(xi.mod.DMG, -4000)
-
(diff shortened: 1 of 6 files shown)


WinterSolstice8:
Does this also carry over the wyvern's innate -40% DT? That's added in wyvern.lua

jmcmorris:
Nope! I was unaware of these so good catch! These, and any other similarly set mods, should be set in Calculate*Stats function in petutils and then it'll be reapplied properly. I'll add these to the PR along with any others that I can find. Let me know if you know of any more.

WinterSolstice8:
``` 
    mob:addMod(xi.mod.DMG, -4000)

    -- innate +40 subtle blow
    mob:addMod(xi.mod.SUBTLE_BLOW, 40)
```
To the best of my knowledge, 40 subtle blow is also present at all levels. There's been no explicit testing, but patch notes didn't indicate any level requirement for subtle blow.

WinterSolstice8 (approved):

claywar (approved):

PR #3294, opened 2022-11-29
ignore this

TeoTwawki:
FU github, I wasn't ready

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5752,7 +5752,11 @@ bool CLuaBaseEntity::hasTitle(uint16 titleID)
 
 void CLuaBaseEntity::addTitle(uint16 titleID)
 {
-    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    if (!(m_PBaseEntity->objtype & TYPE_PC))
+    {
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -5752,7 +5752,11 @@ bool CLuaBaseEntity::hasTitle(uint16 titleID)
 
 void CLuaBaseEntity::addTitle(uint16 titleID)
 {
-    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    if (!(m_PBaseEntity->objtype & TYPE_PC))
+    {
+        ShowError("CLuaBaseEntity::addTitle invalid entity (name: %s type: %d)", m_PBaseEntity->name, m_PBaseEntity->objtype);
```
claywar, line 5757: Logging will automatically append the function name to the log line, so not really necessary to prefix it in these

claywar (approved):

TeoTwawki:
I was still editing, and didn't mean to open this yet

src/map/lua/lua_baseentity.cpp
TeoTwawki: it also shows the namespace the function was in

PR #3295, opened 2022-11-29
Convert some debug breaks into log error+bailout

TeoTwawki:
**_I affirm:_**

swaps out some debug breaks for error messaging and returns out of the function that was misused.

Deliberately call one or more of the edited bindings on the wrong entity type, observe error message instead of crash

## Other details
I've not been super thorough with testing, but there isn't much TO test - I've also gone with bitwise for the checks, which we had a mix of both bitwise and explicit throughout our core. I think bitwise makes sense, because the enum show it was meant to be a mask, and pets/trusts are also still technically mobs and we should work toward making use of that instead of checking mob OR pet OR trust all the time.  To really reach that goal doubtless other changes will likely be needed but remains to be seen.

The methodology of checkage isn't fully consistent for the sake of brevity. For example, addTitle checks that this is "not player" before bailing out. if it's a player it proceeds past that, whereas addMobMod check if the object is either an NPC or a Player, since all of mobs, pets, and trusts are mobMod capable 

```cpp
enum ENTITYTYPE : uint8
{
    TYPE_NONE   = 0x00,
    TYPE_PC     = 0x01,
    TYPE_NPC    = 0x02,
    TYPE_MOB    = 0x04,
    TYPE_PET    = 0x08,
    TYPE_SHIP   = 0x10,
    TYPE_TRUST  = 0x20,
    TYPE_FELLOW = 0x40,
};
```
TYPE_SHIP is [used in exactly 2 places](https://github.com/LandSandBoat/server/search?q=TYPE_SHIP) on things that are also already NPCs, and TYPE_FELLOW is [used in exactly one place](https://github.com/LandSandBoat/server/blob/0be845eddc7e605437b91131c081c8c858a6d94a/src/map/entities/battleentity.cpp#L496) for a placeholder setting that doesn't work yet because the entity doesn't really exist.

(and if you are adding mobmods to a ship you deserve to crash anyway - how would you even do that? I'm not adding a type check for ships all over just in case the user is ***insane*** !)

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5755,7 +5755,11 @@ bool CLuaBaseEntity::hasTitle(uint16 titleID)
 
 void CLuaBaseEntity::addTitle(uint16 titleID)
 {
-    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    if (!(m_PBaseEntity->objtype & TYPE_PC))
+    {
+        ShowError("function call on invalid entity! (name: %s type: %d)", m_PBaseEntity->name, m_PBaseEntity->objtype);
+        return;
+    }
 
     auto* PChar = static_cast<CCharEntity*>(m_PBaseEntity);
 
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

TeoTwawki:
force pushed to kill a magic number along the way. its ready when CI is now.

PR #3297, opened 2022-11-30
Allow into mission battlefield if player has completed mission.

austinclprojects:
**_I affirm:_**

Currently the battlefield framework only allows party members into a mission battlefield if they are on the fight. This allows them to join the battlefield if they have completed the mission and someone on the fight has initialized the battlefield.

Have party member on 5-2 fight enter shadow lord battlefield and then enter as a character that has completed the fight previously.

### DIFF
--- a/scripts/globals/battlefield.lua
+++ b/scripts/globals/battlefield.lua
@@ -1212,7 +1212,9 @@ function BattlefieldMission:checkRequirements(player, npc, isRegistrant, trade)
     local current = player:getCurrentMission(missionArea)
     local missionStatusArea = self.missionStatusArea or player:getNation()
     local status = player:getMissionStatus(missionStatusArea, self.missionStatus)
-    return current == self.mission and status == self.requiredMissionStatus
+    return (not isRegistrant and (player:hasCompletedMission(missionArea, self.mission) or (current == self.mission and status >= self.requiredMissionStatus))) or
+        (current == self.mission and status == self.requiredMissionStatus)
+
 end
 
 function BattlefieldMission:checkSkipCutscene(player)


scripts/globals/battlefield.lua
```diff
@@ -1212,7 +1212,9 @@ function BattlefieldMission:checkRequirements(player, npc, isRegistrant, trade)
     local current = player:getCurrentMission(missionArea)
     local missionStatusArea = self.missionStatusArea or player:getNation()
     local status = player:getMissionStatus(missionStatusArea, self.missionStatus)
-    return current == self.mission and status == self.requiredMissionStatus
+    return (not isRegistrant and player:hasCompletedMission(missionArea, self.mission)) or
+        (current == self.mission and status == self.requiredMissionStatus)
```
jmcmorris, line 1216: While we're at it. Should we have it be `status >= self.requiredMissionStatus`? If you're still progressing through a mission but have completed the battlefield are you able to help other people or do you have to complete the mission first?

jmcmorris (commented):
Good catch!

scripts/globals/battlefield.lua
austinclprojects: That makes sense, added that as a check for non registrants.

austinclprojects: Added a check for that.

jmcmorris: Looks good to me! I'll let @claywar to chime in on the formatting as I'm still learning how that's preferred.

jmcmorris (approved):

scripts/globals/battlefield.lua
claywar: Formatting looks fine (no outer parens, reasonable length, and indent after the return)

Xaver-DaRed (approved):

PR #3299, opened 2022-11-30
Despawn Ally NPCs After BCNM

panicstevenson:
**_I affirm:_**

Cleans up allies after a battlefield. Handles the unique edge case of a player waiting a very long time (allowing for battlefield cleanup) before fully exiting the battlefield.

Enter a BCNM such as Bastok's 9-2 Where Two Paths Converge
Finish the BCNM successfully and skip the event
Finish the BCNM successfully and don't skip the event
Fail the BCNM by letting the ally die
Fail the BCNM by dying and take the homepoint immediately
Fail the BCNM by dying and let it zone you out
Use GM commands to warp out of the BCNM

Ally should despawn. Can redo the BCNM, a different BCNM in the same zone, or use wallhack to find out.

See also: https://github.com/AirSkyBoat/AirSkyBoat/pull/1798

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -597,9 +597,14 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
                 {
                     if (std::find(m_AllyList.begin(), m_AllyList.end(), PMobEntity) != m_AllyList.end())
                     {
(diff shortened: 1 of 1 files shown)


jmcmorris (approved):
Looks good! Thanks for bringing this upstream.

Xaver-DaRed (approved):

PR #3310, opened 2022-12-01
[chore] remove whitespace between function and param declaration

claywar:
**_I affirm:_**

Removes any whitespace between lua keyword function and opening paren

No impact should be observed

### DIFF
--- a/scripts/globals/abilities/reward.lua
+++ b/scripts/globals/abilities/reward.lua
@@ -53,43 +53,43 @@ abilityObject.onUseAbility = function(player, target, ability, action)
             totalHealing   = math.floor(minimumHealing + 2 * (playerMnd - 10))
         end,
 
-        [17017] = function (x) -- pet food beta biscuit
+        [17017] = function() -- pet food beta biscuit
             minimumHealing = 180
             regenAmount    = 3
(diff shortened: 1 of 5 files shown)


Xaver-DaRed (approved):
Didn't reward get cleaned this way already? oh well

scripts/globals/abilities/reward.lua
```diff
@@ -115,14 +115,14 @@ abilityObject.onUseAbility = function(player, target, ability, action)
...
-        [15095] = function (x) -- monster jackcoat
+        [15095] = function() -- monster jackcoat
             -- This will remove Weight, Slow and Silence from the pet.
             pet:delStatusEffect(xi.effect.WEIGHT)
             pet:delStatusEffect(xi.effect.SLOW)
             pet:delStatusEffect(xi.effect.SILENCE)
         end,
 
-        [14508] = function (x) -- monster jackcoat +1
+        [14508] = function() -- monster jackcoat +1
```
TeoTwawki, line 125: I always wondered why the x was even there

TeoTwawki (approved):

scripts/globals/abilities/reward.lua
claywar: Assume that future developers will always use the ugliest code first as a template for additional work.

PR #3317, opened 2022-12-03
[core] Logging improvements

zach2good:
**_I affirm:_**

- Introduce new setting: logging.PATTERN; allowing the user to set whatever logging pattern they'd like. Adds examples for "classic" and "new".
- Adds custom options for logging patterns.
- Make core logging macros exception safe. Example:
```
[12/03/22 20:50:18:666][login][error] Encountered logging exception!: invalid type specifier (do_init:63)
[12/03/22 20:50:18:666][login][error] C:\ffxi\server\src\login\login.cpp:63 (do_init:63)
```
- Remove logging that may happen before the logging pattern is loaded
- Removes another little chunk of kernel

### DIFF
--- a/documentation/model_ids.txt
+++ b/documentation/model_ids.txt
@@ -2460,6 +2460,24 @@ ID Number    Costume Name
 2909    FF14 Spriggan
 2910    FF14 Spriggan
 2928    Porxie
+2933    Mumor II
+2934    Gargoyle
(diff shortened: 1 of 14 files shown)


claywar (approved):

zach2good:
Not changing the default logging pattern in this PR

PR #3321, opened 2022-12-04
Fix SQL query for updating unity accolades

CatsEyeXI:
**_I affirm:_**

Corrects an error issue where accolades are failing to update due to a malformed query.
image

Watch map server log while gaining accolades and you shouldn't encounter errors.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -6028,7 +6028,7 @@ namespace charutils
     {
         TracyZoneScoped;
 
-        const char* Query = "UPDATE char_points SET %s = GREATEST(LEAST(%s+%d, %d), 0) WHERE charid = %u;";
+        const char* Query = "UPDATE char_points SET %s = GREATEST(LEAST(%s+(%d), %d), 0) WHERE charid = %u;";
 
         sql->Query(Query, type, type, amount, max, PChar->id);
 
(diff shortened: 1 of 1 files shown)


src/map/utils/charutils.cpp
```diff
@@ -6035,7 +6035,7 @@ namespace charutils
         if (strcmp(type, "unity_accolades") == 0 && amount > 0)
         {
             float       evalPoints   = static_cast<float>(amount) / 1000;
-            const char* rankingQuery = "UPDATE unity_system SET points_current = points_current+%f WHERE leader=%d;";
+            const char* rankingQuery = "UPDATE unity_system SET points_current = points_current+(%f) WHERE leader=%d;";
```
claywar, line 6038: From your error, I'm going to assume that your code does not match LSB.

```
        const char* Query = "UPDATE char_points SET %s = GREATEST(LEAST(%s+(%d), %d), 0) WHERE charid = %u;";

        sql->Query(Query, type, type, amount, max, PChar->id);

        if (strcmp(type, "unity_accolades") == 0 && amount > 0)
        {
```
The amount value in your logs is <0, and the condition provided requires amount to be >0.  Please confirm that you have checked your local codebase.

CatsEyeXI: while our file is a bit out of date (Nov. 18), the functions that handle the accolades are the same.

image

CatsEyeXI: i originally drafted this PR thinking i had a fix, then noticed what you are stating (num < 0), but i can't account for why that is happening... I guess this can be closed, and re-opened as an issue? though i think i already have one out there for something similar.

claywar: The only thing that comes to mind at the moment is if you had an _insanely_ high modifier for xp gain, and somehow over 2 billion points were allocated to a single unity leader

PR #3322, opened 2022-12-05
Changing Ultima in One to be Feared to use stages and skill lists

jmcmorris:
**_I affirm:_**

Reworking Ultima from One to be Feared to use combat stages and skill lists instead of sticking the logic inside the mob skill scripts. This is being done primarily in preparation for Proto-Ultima.

```
!zone 32
!addmission 6 638
!setplayervar <me> 'Mission[6][638]Status' 3
```

Go through the battle and get to Ultima and then keep doing `!tp 3000` to make it use TP moves at the different stages (69%, 39%, 19%).

### DIFF
--- a/scripts/zones/Sealions_Den/mobs/Ultima.lua
+++ b/scripts/zones/Sealions_Den/mobs/Ultima.lua
@@ -15,8 +15,13 @@ entity.onMobInitialize = function(mob)
     mob:setMobMod(xi.mobMod.GIL_MAX, -1)
 end
 
+entity.onMobSpawn = function(mob)
+    mob:setMobMod(xi.mobMod.SKILL_LIST, 728)
+end
+
 entity.onMobWeaponSkillPrepare = function(mob, target)
-    if mob:getHPP() > 75 then
+    local stage = mob:getLocalVar("stage")
+    if stage == 0 then
         local checker = math.random()
         if checker < 0.50 then
             return abilities[1]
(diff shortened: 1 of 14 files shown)


Xaver-DaRed (approved):

scripts/zones/Sealions_Den/mobs/Ultima.lua
```diff
@@ -15,8 +15,14 @@ entity.onMobInitialize = function(mob)
     mob:setMobMod(xi.mobMod.GIL_MAX, -1)
 end
 
+entity.onMobSpawn = function(mob)
+    mob:setLocalVar("stage", 1)
```
TeoTwawki: if we call the 1st stage `0` instead of `1` you can omit setting it on spawn (things not set yet equal 0 when checked, and localVars are lost on spawn so this would always start at zero)

jmcmorris: That's what I had at first but then I was thinking that reading stage 1, 2, etc. is easier than it being 0 based. I'm fine it being either way.

TeoTwawki: I think enough programmers and even amateurs are familiar enough with starting at 0

zach2good (approved):

claywar (approved):

TeoTwawki (approved):

PR #3324, opened 2022-12-05
Reassign Players to Battlefield on Disconnect

panicstevenson:
**_I affirm:_**

Change the logic for RemoveEntity to only check if the player is leaving the battlefield. If the player disconnects, retain the battlefield subpower so players can reengage in the battle. If the player teleports or warps they will lose the battlefield debuff through various other checks.

Run battlefields with 2 players. Have one player !goto themselves to simulate a disconnect with and without this change.

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -527,7 +527,7 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
 
         m_Zone->updateCharLevelRestriction(PChar);
 
-        if (PChar->StatusEffectContainer->HasStatusEffectByFlag(EFFECTFLAG_CONFRONTATION))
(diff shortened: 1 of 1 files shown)


TeoTwawki:
From what I recall, retail will let you rejoin only if battlefield didn't go empty. So if solo, disconnecting does end it. If in a party, you can log back in and rejoin the group.

However, Teo's swiss cheese memory should not be trusted in a vacuum. I want to say captures with accompanying video likely exist for these situations already, but I haven't checked.

jmcmorris:
There is a great capture of this by Wiggo that tests all scenarios out: https://discord.com/channels/443544205206355968/443594542613004308/697309989223661579

I worked with @panicstevenson on this to come to this solution. It was working until I broke things in #3136.

jmcmorris (approved):
Thanks for making the PR!

zach2good (approved):

claywar (approved):

TeoTwawki (approved):

PR #3333, opened 2022-12-06
Fixing some battlefield interactions

### LINKED ISSUE #3332: 🐛 Entering a Battlefield After Party Engages Teleports Player to Empty Battlefield
**_I affirm:_**

Windows 10

## Branch affected by issue 

`base`

## Steps to reproduce

Have two players, have the first player start a battlefield and the other player waits outside. Begin the battlefield with the first player. The second player then interacts with the entrance NPC and is teleported inside. The player will receive this message:

image

The second player will be unable to use the exit NPC to leave.

## Expected behavior

Print the message but leave the player outside of the battlefield.

jmcmorris:
**_I affirm:_**

This fixes a few issues:

**If the player's party enters a battlefield with area 1 but never goes inside but logs out with the effect then when they log in they'll be incorrectly kicked out of the battlefield.**

Fix #3332 
The fix is to better track the player's battlefield entered status with effect's subpower. The first bit of the subpower is the entered status and everything after that is used for which battlefield area the player is in. This is done in order to know if the player needs to be ejected when logging in with a battlefield status effect.

I was partially doing this before but I failed to realize that the subpower was being set to the battlefield area when the status effect is created. This stored area value it not used from what I can see but I figured I'd preserve it just in case we do ever need it. 

**Character is able to reenter a locked battlefield if they leave after it has been locked**

This is easily fixed by removing the battlefield status effect when they leave a locked battlefield. This is how the character is prevented from entering a locked battlefield normally or at least should have...

**Character's battlefield status effect is not removed when a battlefield is locked**

This was simply an issue with `Battlefield:onBattlefieldStatusChange` trying to take an extra parameter. I had forgotten I had even fixed this otherwise I would have put it into its own commit. Whoops!

Go to a battlefield with two characters and try the scenarios described above.

### DIFF
--- a/scripts/globals/battlefield.lua
+++ b/scripts/globals/battlefield.lua
@@ -796,6 +796,7 @@ function Battlefield.redirectEventCall(eventName, player, csid, option)
 end
 
 function Battlefield:onEventFinishEnter(player, csid, option)
+    player:setEnteredBattlefield(true)
     player:setLocalVar("[battlefield]area", 0)
     self:setLocalVar(player, "CS", 1)
 end
@@ -892,13 +893,14 @@ end
 function Battlefield:onBattlefieldRegister(player, battlefield)
 end
 
-function Battlefield:onBattlefieldStatusChange(battlefield, players, status)
+function Battlefield:onBattlefieldStatusChange(battlefield, status)
     -- Remove battlefield effect for players in alliance not inside battlefield once the battlefield gets locked. Do this only once.
     if
         status == xi.battlefield.status.LOCKED and
         battlefield:getLocalVar("statusRemoval") == 0
     then
         battlefield:setLocalVar("statusRemoval", 1)
+        local players = battlefield:getPlayers()
 
         for _, player in pairs(players) do
             local alliance = player:getAlliance()
@@ -981,11 +983,6 @@ function Battlefield:onBattlefieldEnter(player, battlefield)
     end
 
     player:messageSpecial(ID.text.TIME_LIMIT_FOR_THIS_BATTLE_IS, 0, 0, 0, math.floor(self.timeLimit / 60))
-
-    if player:hasStatusEffect(xi.effect.BATTLEFIELD) then
-        local status = player:getStatusEffect(xi.effect.BATTLEFIELD)
-        status:setSubPower(1)
-    end
 end
 
 function Battlefield:onBattlefieldDestroy(battlefield)
@@ -996,9 +993,6 @@ function Battlefield:onBattlefieldLeave(player, battlefield, leavecode)
         self:onBattlefieldWin(player, battlefield)
     elseif leavecode == xi.battlefield.leaveCode.LOST then
         self:onBattlefieldLoss(player, battlefield)
-    elseif player:hasStatusEffect(xi.effect.BATTLEFIELD) then
-        local status = player:getStatusEffect(xi.effect.BATTLEFIELD)
-        status:setSubPower(0)
     end
 end
 
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -529,7 +529,7 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
 
         if (leavecode == BATTLEFIELD_LEAVE_CODE_EXIT && PChar->StatusEffectContainer->HasStatusEffectByFlag(EFFECTFLAG_CONFRONTATION))
         {
-            PChar->StatusEffectContainer->GetStatusEffect(EFFECT_BATTLEFIELD)->SetSubPower(0);
+            setPlayerEntered(PChar, false);
         }
 
         if (PChar->isDead())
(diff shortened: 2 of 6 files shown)


src/map/battlefield_handler.cpp
```diff
@@ -223,8 +225,10 @@ uint8 CBattlefieldHandler::LoadBattlefield(CCharEntity* PChar, const Battlefield
 
     if (!PChar->StatusEffectContainer->GetStatusEffect(EFFECT_BATTLEFIELD))
     {
+        // The first bit of the subpower is reserved for if the player has entered the battlefield or not
+        uint16 subpower = PBattlefield->GetArea() << 1;
```
TeoTwawki: I seriously dislike setting this precedent, or using this pattern at all especially for this. Bitpacking/shifting to handle things in an effect field don't sit well with me.

I'd rather do something far messier tbh. 3 (4?) rewrites ago, players would get "registered" in the core to a battlefield copy on entry for its duration. I forget the exact code involved in that version of things.

TeoTwawki: I'm genuinely curious: How terribad would it be to add new things to the battlefield object to toggle per copy+person?

jmcmorris: What we need this for is when the battlefield doesn't exist anymore and we need to kick the player out. Some alternatives include:
- Scrap the area since we aren't actively using it and allow the subpower to simply be entered or not (1 or 0).
- Use a charvar that we set when they enter and clear when they leave. This honestly wouldn't be too bad since it'll be cleared out when they log back in.

claywar: Instead of storing this on the player in the effect, I wonder if it might be better to track this in the battlefield object itself, since _I think_ we already know they're registered?

claywar: The other option is that we could potentially use status effect Tier as well

TeoTwawki: > What we need this for is when the battlefield doesn't exist anymore and we need to kick the player out. Some alternatives include:
> 
> * Scrap the area since we aren't actively using it and allow the subpower to simply be entered or not (1 or 0).
> * Use a charvar that we set when they enter and clear when they leave. This honestly wouldn't be too bad since it'll be cleared out when they log back in.

one or more of the previous incarnations of bcnm systems used to do the charvar. to me it'd be better than bitpacking this into the effect. as for the battlefield not existing anymore, couldn't we use that in itself as a stand in - use the value in the object when the object does exist, use the fact it doesn't exist when it doesn't to mean "false" as well?

using tier as an in/out  (0 or 1) also is more appealing than using subpower here, if we go a 1 or 0 route

jmcmorris: The reason we cant store this on the battlefield object itself is because the character is logging out for an unknown amount of time. We wouldn't want to keep the battlefield instance around and even if we tried doing that there is always the scenario that the server goes down.

Using the tier looks easy enough and is a great idea! That's what I'm going to roll with.

TeoTwawki: > The reason we cant store this on the battlefield object itself is because the character is logging out for an unknown amount of time. We wouldn't want to keep the battlefield instance around and even if we tried doing that there is always the scenario that the server goes down.

Yes but if the battlefield ceased to exist, we wouldn't need to know if a player can enter it anymore - we can treat that as the same as "no you can't" if we check for that condition (nil/null) see what i am saying?

TeoTwawki: Tier is fine, as long as we ain't bitpacking I'm content

jmcmorris: > 
> Yes but if the battlefield ceased to exist, we wouldn't need to know if a player can enter it anymore - we can treat that as the same as "no you can't" if we check for that condition (nil/null) see what i am saying?

This is for the scenario that the player has logged out/disconnected while **inside** the battlefield. In this case we need to kick the player out. However, if they didn't enter but have the battlefield effect then we don't want to kick them out.

jmcmorris:
I reworked the entered state tracking to be on the battlefield status effect tier.

src/map/battlefield_handler.cpp
TeoTwawki: > This is for the scenario that the player has logged out/disconnected while inside the battlefield. In this case we need to kick the player out. 

this is still "battlefield ceased to exist, you don't belong here /kick"

> However, if they didn't enter but have the battlefield effect then we don't want to kick them out.

????

how exactly do you have the effect and be inside a battlefield that ceased to exist

jmcmorris: Log out inside a battlefield and when you log back in you'll still have the effect

TeoTwawki: > Log out inside a battlefield and when you log back in you'll still have the effect

which would be an invalid one (because empty = ends) and kick them if solo. if they weren't solo, it'd still exist with someone inside it.

panicstevenson:
Tested this out and it works great! I have two things and both are extremely minor.

1. Is there any way to retrofit this to battlefields not currently on the framework as a stopgap? I think it's fine if the answer is "convert it to the new framework".
2. If a player in a party is outside of the battlefield but still in the selection window when the rest of the party engages and the player loses the battlefield buff the player can still get teleported into the zone but will receive the message that they were unable to zone in, leaving the player stranded. The player in this scenario is obviously playing with fire, but I wonder if some of https://github.com/AirSkyBoat/AirSkyBoat/pull/2061 can land to add an Event Skipped when a player loses their confrontation flag while in a menu in the future for a GM-free experience.

panicstevenson (approved):

jmcmorris:
> Tested this out and it works great! I have two things and both are extremely minor.
> 
> 1. Is there any way to retrofit this to battlefields not currently on the framework as a stopgap? I think it's fine if the answer is "convert it to the new framework".
> 2. If a player in a party is outside of the battlefield but still in the selection window when the rest of the party engages and the player loses the battlefield buff the player can still get teleported into the zone but will receive the message that they were unable to zone in, leaving the player stranded. The player in this scenario is obviously playing with fire, but I wonder if some of [Stop Players from Interacting with Event Menus and Teleporters when Damaged AirSkyBoat/AirSkyBoat#2061](https://github.com/AirSkyBoat/AirSkyBoat/pull/2061) can land to add an Event Skipped when a player loses their confrontation flag while in a menu in the future for a GM-free experience.

1. Should be possible to go through the entry npcs such as Burning Circles and add to the event finished `player:setEnteredBattlefield(true)` for the right event, `32000`.
2. Thanks! I'm fully aware of this. I made #3334 yesterday to track this as I'm hoping to get a capture to see exactly how retail handles it.

claywar (approved):

PR #3350, opened 2022-12-11
Implement Status Effect Magic Evasion modifiers

Xaver-DaRed:
**_I affirm:_**

- Creates STATUS_MEVA family of modifiers that will be used in the upcoming rewrite of enfeebling spells.
- Corrects Bar-Effect spells potency, since they seemingly arent affected by enhancing skill.
- Substitutes STATUSRES with STATUS_MEVA when applicable. Hopefully I got them all.

For the time being? No steps.


scripts/globals/effects/asylum.lua
```diff
@@ -6,42 +6,42 @@ require("scripts/globals/status")
 local effectObject = {}
 
 effectObject.onEffectGain = function(target, effect)
```
WinterSolstice8, line 8: This is probably a massive nerf to how Asylum works, but I don't think we know the specifics either. ${STATUS}RES is probably _more_ accurate than pure meva. Thoughts?

scripts/globals/effects/perfect_defense.lua
```diff
@@ -10,16 +10,16 @@ effectObject.onEffectGain = function(target, effect)
     target:addMod(xi.mod.UDMGBREATH, -effect:getPower())
     target:addMod(xi.mod.UDMGMAGIC, -effect:getPower())
     target:addMod(xi.mod.UDMGRANGE, -effect:getPower())
-    target:addMod(xi.mod.SLEEPRES, effect:getPower())
```
WinterSolstice8, line 13: This one's also a massive nerf to how it works like Asylum, but details aren't particularly known either.

WinterSolstice8:
I know you had asked where "Resist!" procs come from but I legitimately forgot about bard songs

JP wiki says bard songs are resist proc 
http://wiki.ffo.jp/html/note16731.html

scripts/globals/effects/asylum.lua
Xaver-DaRed: Job points increase MEVA by 4 points each to asylum, so i would say it definately uses STATUS_MEVA modifier

The numbers, however can be increased. I didnt touch any values

Xaver-DaRed:
Reverted bard song modifiers.

As for "Asylum" and "Perfect Defense", they do not trigger Resist! message, meaning, they probably give a massive boost to MEVA (specific to the status) akin to the MACC that Elemental Seal and such JAs give (256)

I would propose raising the potency of this modifiers instead.

Another note: STATUSRES is supposed to cap at 80%, meaning they would get nerfed even if left with STATUSRES.
And I have honest doubts they were having any actual effect anyway.

WinterSolstice8 (approved):

claywar (approved):

PR #3362, opened 2022-12-12
[Rewrite] Convert Enfeebling spells to new system

Xaver-DaRed:
**_I affirm:_**

 - Removes logic from enfeebling spells (White, Black and Ninjutsu) and uses global file to handle parameters and calculations.
 - Makes enfeebling spells return magic burst! message.
 - Makes STATUSRES modifiers work as they should, not affecting MEVA, but allowing for a separate check for spell nulification.
 - Makes potency resistance separate from duration resistance.
 - Added check for mob status immunities.

Cast ninjutsu, black and white enfeebling spells. Make sure server doesnt return any errors. Make sure the spells are aplied, or rsisted, etc...


jmcmorris:
Looking good! Are you going to handle immunity with the "completely resists" message in this?

Xaver-DaRed:
Yes. The very first function in the main file handles it. I did have to make separate modifiers to divide resist! chance from meva against specific status. That was done in separate PRs

EDIT: I completely missread what you wrote. To answer your question, yes, I plan to.

claywar (approved):

WinterSolstice8 (approved):

PR #3370, opened 2022-12-13
[login] Switch from Select to epoll on Linux/Windows

WinterSolstice8:
**_I affirm:_**

Switch from Select to epoll on Linux/Windows for Login only (Map only uses 1 fd and doesn't necessarily need this)
Increase max concurrent connection count by tenfold to 10240

Login, abuse it

Drafting for comments (and if I missed anything)


zach2good (approved):
🙏

claywar (approved):

PR #3375, opened 2022-12-13
Redesign Limbus Armoury Crate Loot

jmcmorris:
**_I affirm:_**

I had been holding off on the Limbus Armoury Crate loot tables until I finished all areas. This is that rework I had promised.
I reviewed the wiki and all other available data and decided it simply was not good enough and unreasonable to achieve that:
1. Very small sample size
2. Likely biased data entry
3. Data collected before and after potentially major changes with the release of new jobs

Consequently I decided to design my own loot distribution based off of the general likelihood to get an AF+1 material. All of these materials are evenly distributed with this solution as far as likelihood to get one AF+1 material to another as well as on which floor the material can drop from. I have also evenly distributed BLU, COR, PUP, DNC and SCH so they can easily be commented out and have little to no impact on other material drop rates.


Xaver-DaRed (approved):
Looks reasonable to me

PR #3377, opened 2022-12-14
[lua] Implement energy drain and energy steal, add return values to some lua functions

WinterSolstice8:
**_I affirm:_**

Implements Energy Drain & Energy Steal

Use Energy Drain & Energy Steal

Thanks to Virtuosus for the formulas & verification help


Xaver-DaRed (approved):
Good to go once the styling is fixed.

I will suggest `mpRestored` as an alternative for the variable name

claywar (approved):

PR #3379, opened 2022-12-14
Loot Reformation Part 1

jmcmorris:
**_I affirm:_**

This replaces `addDropListModification` with ITEM_DROPS event that provides scripts with a `CLuaLootContainer` object that can have items and item groups added to it. I have provided two example use cases for it but there are countless more. Both `SetDropID` and `SetDropRate` could primarily be replaced with these; however, they are used in modules so we may want to leave the functions around for overriding data.

Go into Central Temenos Basement or Central Apollyon and kill Aern or Gunpod respectively and they should drop different loot based off of the logic in the scripts.

### DIFF
--- a/scripts/battlefields/Temenos/central_temenos_basement.lua
+++ b/scripts/battlefields/Temenos/central_temenos_basement.lua
@@ -10,6 +10,7 @@ require("scripts/globals/battlefield")
 require("scripts/globals/limbus")
 require("scripts/globals/items")
 require("scripts/globals/keyitems")
+require("scripts/globals/loot")
 -----------------------------------
 
 local content = Limbus:new({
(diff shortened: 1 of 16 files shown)


scripts/battlefields/Temenos/central_temenos_basement.lua
```diff
@@ -54,7 +55,10 @@ content.groups =
                     end
                 end)
 
-                -- TODO: Aern should drop ABC = 1 + math.min(3, mob:getLocalVar("AERN_RERAISES")) (before the reraise)
+                mob:addListener("ITEM_DROPS", "ITEM_DROPS_AERN", function(mobArg, loot)
+                    local quantity = math.min(3, mob:getLocalVar("AERN_RERAISES"))
```
WinterSolstice8, line 59: Comment said `1 + math.min(3, mob:getLocalVar("AERN_RERAISES"))`
was that wrong?

WinterSolstice8 (approved):

scripts/battlefields/Temenos/central_temenos_basement.lua
jmcmorris: Correct. I wasn't sure if this would run before or after the bar is incremented. One is also coming from the drop id set on the mob pool

PR #3382, opened 2022-12-15
[core] Add message modifier to Spells

WinterSolstice8:
**_I affirm:_**

Adds the ability to set the message modifier on spells from lua

An example of how to use it, though only for demonstration purposes:
```
spellObject.onSpellCast = function(caster, target, spell)
    spell:setModifier(xi.actionModifier.RESIST)
    return xi.spells.damage.useDamageSpell(caster, target, spell)
end
```

after that your modified spell should land with "Resist! {normal message here}"

### DIFF
--- a/src/map/spell.h
+++ b/src/map/spell.h
@@ -1110,6 +1110,9 @@ public:
     void setMultiplier(float multiplier);
     void setMessage(uint16 message);
     void setMagicBurstMessage(uint16 message);
+    auto getModifier() -> MODIFIER;
+    void setModifier(MODIFIER modifier); // set Spell modifier message, MUST reset the modifier on use otherwise it will be stale
+
     void setCE(uint16 ce);
     void setVE(uint16 ve);
     void setRequirements(uint8 requirements);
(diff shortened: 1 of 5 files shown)


src/map/spell.h
```diff
@@ -1090,26 +1090,29 @@ class CSpell
...
+    void     setSpellFamily(SPELLFAMILY SpellFamily);
+    void     setSkillType(uint8 SkillType);
+    void     setZoneMisc(uint16 Misc);
+    void     setAOE(uint8 AOE);
+    void     setBase(uint16 base);
+    void     setElement(uint16 element);
+    void     setMultiplier(float multiplier);
+    void     setMessage(uint16 message);
+    void     setMagicBurstMessage(uint16 message);
+    MODIFIER getModifier();
```
zach2good: No need to shift all all this alignment;

auto getModifier() -> MODIFIER;

zach2good (approved):

src/map/spell.h
WinterSolstice8: done

claywar (approved):

PR #3383, opened 2022-12-15
[ci] Colorize cpp diff output

zach2good:
**_I affirm:_**

image

I have seen that it should be possible to colorize the diff output inside the github action output. This would make it much clearer to people trying to format their code, if they aren't comfortable reading plaintext diffs

CI DANCE!


zach2good:
image

zach2good:
In the interest of getting this in quickly, I'll squash all these commits when I merge

claywar (approved):

PR #3400, opened 2022-12-19
[core] Replace DYNAMIC_LOOKUP with GetFirstID

zach2good:
**_I affirm:_**

`DYNAMIC_LOOKUP` had very limited use, and wasn't super clear about how it worked under the hood. I've replaced with with `GetFirstID`, which uses the same logic, in the same order, but allows you to look up anything regardless of the key you're applying it to. 

Will be good for building offsets, etc.

Server starts up fine, without any errors, zone into `Arrapago_Reef`, still no errors

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -449,7 +449,7 @@
         "ForceCrash",
         "BuildString",
         "SendLuaFuncStringToZone",
-        "DYNAMIC_LOOKUP",
+        "GetFirstID",
     ],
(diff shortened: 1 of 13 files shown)


claywar (approved):

TeoTwawki (approved):

WinterSolstice8 (approved):
👍 Very much like this, it's a lot more obvious what's happening to the casual reader

Xaver-DaRed (approved):

PR #3401, opened 2022-12-19
[core] Add debug::isRunningUnderDebugger

zach2good:
**_I affirm:_**

When I wrote the watchdog, I was scared of having to write this bit of code. This was my original intention: If you've got something going on where the main tick has been delayed massively (like an infinite loop or whatever) the server process needs to die. This generates a crash dump and allows any service things to kick in and restart the process.

### DIFF
--- a/src/common/debug.h
+++ b/src/common/debug.h
@@ -3,4 +3,7 @@
 namespace debug
 {
     void init();
+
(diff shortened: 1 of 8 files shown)


claywar (approved):
Pre-emptive stamp for when MacOS CI doesn't complain

Xaver-DaRed (approved):

PR #3405, opened 2022-12-19
[modules] Make ah_announcement.cpp more stable

zach2good:
**_I affirm:_**

The two queries that comprise the logic in here weren't guaranteed to be handling the same rows, so in a highly contested (multi-process) environment messages may have been harmlessly sent to the wrong recipients.

Now, we store the seller id in LAST_INSERT_ID, which is guaranteed to be correct per-connection.

Each thread has its own connection, so no worries across threads.

- Enable the module
- Log in with 2 chars
- Sell something on char 1
- Buy it with char 2
- char 1 gets the notification

### DIFF
--- a/modules/custom/cpp/ah_announcement.cpp
+++ b/modules/custom/cpp/ah_announcement.cpp
@@ -1,9 +1,9 @@
 ﻿/************************************************************************
-* Auction House Announcements
-*
-* This will send a message to the seller of an item when it is bought,
-* informing them that their item sold, to who, and for how much.
-* It will only send this message if the seller is online.
+ * Auction House Announcements
+ *
+ * This will send a message to the seller of an item when it is bought,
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

WinterSolstice8 (approved):

modules/custom/cpp/ah_announcement.cpp
```diff
@@ -1,9 +1,9 @@
 ﻿/************************************************************************
-* Auction House Announcements
-*
-* This will send a message to the seller of an item when it is bought,
-* informing them that their item sold, to who, and for how much.
-* It will only send this message if the seller is online.
+ * Auction House Announcements
+ *
```
TeoTwawki, line 3: brick throwing

PR #3406, opened 2022-12-20
Don't allow enmity across zone boundaries

zach2good:
**_I affirm:_**

Makes sense to me!


Xaver-DaRed:
What is this doing exactly?

A player should have his enmity persist through zoning, so long as they are not the main target. Then the enmity should be cleared if the mob tries to target said player and he (the player) isnt in the zone

zach2good:
This doesn't clear enmity, this stops you adding enmity to a player in a different zone to the enmity holder

Xaver-DaRed (approved):

PR #3420, opened 2022-12-21
[Nyzul] Change raw Ids to GetFirstID() and fix some npcs not loading

Xaver-DaRed:
**_I affirm:_**

 - Changes some raw IDs to use GetFirstID()
 - Changed some NPCs pos values to be "not 0" so they actually load
 - Changed RUNIC_LAMP_1 name to RUNIC_LAMP_OFFSET and removed unused npc definitions in IDs file

? Nyzul is disabled

### DIFF
--- a/scripts/zones/Nyzul_Isle/IDs.lua
+++ b/scripts/zones/Nyzul_Isle/IDs.lua
@@ -187,15 +187,11 @@ zones[xi.zone.NYZUL_ISLE] =
         },
 
         -- Nyzul Isle Investigation
-        RUNIC_LAMP_1        = 17093332,
-        RUNIC_LAMP_2        = 17093333,
-        RUNIC_LAMP_3        = 17093334,
-        RUNIC_LAMP_4        = 17093335,
-        RUNIC_LAMP_5        = 17093336,
-        _257                = 17093359,
-        _259                = 17093361,
-        RUNE_TRANSFER_START = 17093429,
-        VENDING_BOX         = 17093430,
+        RUNIC_LAMP_OFFSET   = GetFirstID("Runic_Lamp"),
+        _257                = GetFirstID("_257"),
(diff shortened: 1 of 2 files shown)


claywar (approved):

scripts/zones/Nyzul_Isle/IDs.lua
```diff
@@ -173,29 +173,25 @@ zones[xi.zone.NYZUL_ISLE] =
...
         TREASURE_COFFER =
         {
-            17092611, 17092612, 17092614
+            GetFirstID("Armoury_Crate") + 2, GetFirstID("Armoury_Crate") + 3, GetFirstID("Armoury_Crate") + 5
         },
 
         TREASURE_CASKET =
         {
-            17092609, 17092610, 17092613, 17092620
+            GetFirstID("Armoury_Crate"), GetFirstID("Armoury_Crate") + 1, GetFirstID("Armoury_Crate") + 4, GetFirstID("Armoury_Crate") + 11
```
zach2good: I'm not certain this will work;

https://github.com/LandSandBoat/server/blob/base/src/map/lua/luautils.cpp#L790

We don't iterate down multiple levels of the npc and mob tables in here, only the base level

zach2good (changes requested):

scripts/zones/Nyzul_Isle/IDs.lua
Xaver-DaRed: I dont see why it wouldnt.  GetFirstID() returns a numeric value, to which I add a number. Or do you mean we cant have math in here?

Xaver-DaRed: For now, reverted that set of changes.

zach2good (approved):

PR #3428, opened 2022-12-23
healing.lua fix for Dances with Luopans Geo quest

ffxijuggalo:
**_I affirm:_**

This fix is for completing the "Dances with Luopans" GEO job unlock quest.

During the quest, you /heal at the Ergon Locust. The one I chose is near the frontier station in Ceizak Battlegrounds.
While healing, I immediately received the message: "The arcane energies begin to course within your veins."
According to the quest's script, the maximum wait time while healing is 8 minutes. I ended up waiting for over 30 for the next message to display, which never did. At the same time received this error msg in the map server:

[12/21/22 16:28:06:540][map][error] luautils::onEffectGain: ./scripts/globals/effects/healing.lua:41: attempt to index field 'map' (a nil value)
stack traceback:
	./scripts/globals/effects/healing.lua:41: in function <./scripts/globals/effects/healing.lua:16> (luautils::OnEffectGain:2375)

Now I'm by no means an expert, it just seemed odd to me the "main" in xi.settings.main.map.HEALING_TICK_DELAY was there. So I deleted it. Rebooted the server. Retested /heal near the locust and sure enough, 4 and a half minutes later I received the second message: "You feel a mystical warmth welling up inside you!" Also there were no errors in the map server after this change.

### DIFF
--- a/scripts/globals/effects/healing.lua
+++ b/scripts/globals/effects/healing.lua
@@ -38,7 +38,7 @@ effectObject.onEffectGain = function(target, effect)
         target:messageSpecial(ID.text.ENERGIES_COURSE)
 
         local maxWaitTime = 480  -- Max wait of 8 minutes
-        local secondsPerTick = xi.settings.main.map.HEALING_TICK_DELAY
+        local secondsPerTick = xi.settings.map.HEALING_TICK_DELAY
         local minWaitTime = math.min(3 * secondsPerTick, maxWaitTime)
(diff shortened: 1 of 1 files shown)


TeoTwawki:
yeah we have a `settings.main` and a `settings.map` but they are never in the same line as `settings.main.map`- some copy paste or find/replace mistake happened there methinks

claywar (approved):

PR #3430, opened 2022-12-23
Loot Reform Part 2

jmcmorris:
**_I affirm:_**

Moves nearly all uses of SetDropRate and SetDropID to use ITEM_DROPS event instead.
Added in support for fixed drop rate (not affected by TH) as well since we did not have a method to support.

Go fight any of the changed mobs and drops should work the same.


scripts/mixins/animated_weapon.lua
```diff
@@ -0,0 +1,22 @@
+-- Animated Weapon (Dynamis) mixin
+
+require("scripts/globals/mixins")
+
+g_mixins = g_mixins or {}
+g_mixins.families = g_mixins.families or {}
+
+g_mixins.families.animated_weapon = function(weaponMob, options)
+    weaponMob:addListener("ENGAGE", "ENGAGE_ANIMATED", function(mob, target)
+        if mob:getAnimationSub() == 3 then
```
WinterSolstice8, line 10: huh, is that how that works? interesting

src/map/lua/lua_baseentity.cpp
```diff
@@ -14332,6 +14332,7 @@ uint32 CLuaBaseEntity::getDropID()
 }
 
 /************************************************************************
+ *  DEPRICATED: Use target:addListener("ITEM_DROPS", ...) instead.
```
WinterSolstice8: deprecated

src/map/lua/lua_loot.cpp
```diff
@@ -69,7 +65,47 @@ void CLuaLootContainer::addItem(uint16 item, uint16 rate, sol::variadic_args va)
...
+
+void CLuaLootContainer::addGroupFixed(uint16 groupRate, sol::table items)
+{
+    addGroupToContainer(groupRate, items, true);
+}
+
+void CLuaLootContainer::addItemToContainer(uint16 item, uint16 rate, sol::variadic_args va, bool hasFixedRate)
+{
+    const auto quantity = va.get_type(0) == sol::type::number ? va.get<uint16>(0) : 1;
+    for (uint16 i = 0; i < quantity; ++i)
```
WinterSolstice8, line 100: Should this limit to the max size of a treasure pool? Or is that handled elsewhere? I suppose most of the time that's a moot point because mobs never have enough drops the flood the pool from one mob anyway

jmcmorris: That'll be handled when the drops are calculated. Mob drop lists come from SQL, `ITEM_DROPS` can temporarily add to the drop list, and then the drops are calculated.

src/map/lua/lua_baseentity.cpp
jmcmorris: I sware I speel gud

scripts/mixins/animated_weapon.lua
jmcmorris: Honestly not sure how this works I just replicated the existing logic

scripts/commands/fafnir.lua
```diff
@@ -60,7 +60,7 @@ function onTrigger(player)
     -- Use the mob object as you normally would
     mob:setSpawn(player:getXPos(), player:getYPos(), player:getZPos(), player:getRotPos())
 
-    mob:setDropID(0) -- No loot!
```
TeoTwawki, line 63: I think the reason this used setDropID was to serve as an example of how to give a dynamic mob a drop list via script, since it would not have any sql to do so with. so this should stay for that reason.

scripts/mixins/families/aern.lua
```diff
@@ -44,7 +43,6 @@ g_mixins.families.aern = function(aernMob)
 
         mob:timer(12000, function(mobArg)
             mobArg:setHP(mob:getMaxHP())
-            mobArg:setDropID(dropid)
```
TeoTwawki, line 47: not sure why this was done here in the 1st place.

scripts/zones/Dynamis-Xarcabard/mobs/Animated_Claymore.lua
```diff
@@ -2,18 +2,14 @@
...
+mixins = { require("scripts/mixins/animated_weapon") }
+mixinOptions = { item = xi.items.INTRICATE_FRAGMENT }
 -----------------------------------
 local entity = {}
 
 entity.onMobEngaged = function(mob, target)
-    if mob:getAnimationSub() == 3 then
-        SetDropRate(102, 1574, 1000)
-    else
-        SetDropRate(102, 1574, 0)
```
TeoTwawki, line 14: death to this pattern. DEATH!

scripts/zones/Lufaise_Meadows/mobs/Padfoot.lua
```diff
@@ -11,12 +11,19 @@ local ID = require("scripts/zones/Lufaise_Meadows/IDs")
...
+entity.onMobInitialize = function(mob)
+    mob:addListener("ITEM_DROPS", "ITEM_DROPS_PADFOOD", function(mobArg, loot)
+        if mob:getID() == ID.mob.PADFOOT[GetServerVariable("realPadfoot")] then
+            loot:addGroup(xi.loot.rate.GUARANTEED,
+            {
+                { item = xi.items.ASSAILANTS_RING, weight = 75 },
+                { item = xi.items.ASTRAL_EARRING, weight = 25 },
+            })
+        else
+            loot:addItem(xi.items.SHEEPSKIN, xi.loot.rate.GUARANTEED)
```
TeoTwawki: I've actually recently killed this guy without getting the sheepskin on his fakes. just the cube.

scripts/zones/Lufaise_Meadows/mobs/Padfoot.lua
```diff
@@ -11,12 +11,19 @@ local ID = require("scripts/zones/Lufaise_Meadows/IDs")
...
-        mob:setDropID(2911)
-    else
-        mob:setDropID(1972)
-    end
+entity.onMobInitialize = function(mob)
+    mob:addListener("ITEM_DROPS", "ITEM_DROPS_PADFOOD", function(mobArg, loot)
+        if mob:getID() == ID.mob.PADFOOT[GetServerVariable("realPadfoot")] then
+            loot:addGroup(xi.loot.rate.GUARANTEED,
+            {
+                { item = xi.items.ASSAILANTS_RING, weight = 75 },
```
TeoTwawki: we're only using 0-100 for the % rates? I'd like to be consistent with the sql pattern and have the extra decimal point precision if only to match it.

sql/mob_droplist.sql
```diff
@@ -1041,53 +1041,7 @@ INSERT INTO `mob_droplist` VALUES (94,0,0,1000,2893,280); -- Gargantuan Black Ti
...
-
--- ZoneID: 135 - Animated Staff
-INSERT INTO `mob_droplist` VALUES (108,0,0,1000,1582,@ALWAYS); -- Celestial Fragment (Always, 100%)
-
--- ZoneID: 135 - Animated Tabar
-INSERT INTO `mob_droplist` VALUES (109,0,0,1000,1575,@ALWAYS); -- Runaeic Fragment (Always, 100%)
-
--- ZoneID: 135 - Animated Tachi
-INSERT INTO `mob_droplist` VALUES (110,0,0,1000,1580,@ALWAYS); -- Divine Fragment (Always, 100%)
+-- 95-110 Available
```
TeoTwawki, line 1044: :+1:

src/map/lua/lua_loot.cpp
TeoTwawki: @WinterSolstice8 I'd say no, there are certain cases where it would but the handling for those situations lies elsewhere and we'd a hard limit in treasure pool for our current drops, which I can actually make a reasonable argument is actually not retail in the 1st place (if it overflows and items are dropped, thats not necessarily wrong, tho in 99% of cases retail uses groupings to avoid that causing loss of rare loot)

scripts/commands/fafnir.lua
jmcmorris: Added it back in

scripts/zones/Lufaise_Meadows/mobs/Padfoot.lua
jmcmorris: That makes sense. I changed this to be 750 and 250 to match sql 1000 weights.

scripts/zones/Lufaise_Meadows/mobs/Padfoot.lua
jmcmorris: Teo confirmed! Reduced to very common.

Xaver-DaRed (approved):

TeoTwawki (approved):
lgtm

PR #3432, opened 2022-12-24
[Fix] Quest Past Perfect

bkesecker:
**_I affirm:_**

Fixed quest Past Perfect having too many parameters to quest:keyItem call.  Does not close a known issue.

Set Bastok rank to 2, Bastok fame to 1000, zone to Port Bastok and start the quest by talking to Evi, to Rafaela, and then Evi again.  Zone to Konschtat and travel southwest from the Crag to the base of the windmill having the ???, and then interact with it.  Instead of an error being reported on the map server about invalid parameters, the key item should be obtained.  Return to Port Bastok and talk to Evi to complete the quest.

### DIFF
--- a/scripts/quests/bastok/Past_Perfect.lua
+++ b/scripts/quests/bastok/Past_Perfect.lua
@@ -73,7 +73,7 @@ quest.sections =
             {
                 onTrigger = function(player, npc)
                     if not player:hasKeyItem(xi.ki.TATTERED_MISSION_ORDERS) then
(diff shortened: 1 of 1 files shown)


claywar (approved):
Good catch,  Absolutely an issue

KnowOne134 (approved):

jmcmorris (approved):
Happy to see you contributing @bkesecker!

PR #3443, opened 2022-12-27
Tweak `!goto` and `!bring`

TeoTwawki:
**_I affirm:_**

1. Fix misplaced `)` in `!goto`
2. Escape the input name like we do elsewhere.
3. Remove unneeded `XI_DEBUG_BREAK_IF`
1. Build it
2. use `!gfoto` and `!bring`  in the following scenarios:
- The target player is online and in the same map server instance (should work)
- The target player is on but in a different instance (should work)
- The target player is offline (this should print an error message in game)
- The target player **does not exist** (this should print an error message in game)
- The target has a completely invalid name with symbols or jank in it (this should print an error message in game)

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -3209,11 +3209,13 @@ void CLuaBaseEntity::goToEntity(uint32 targetID, sol::object const& option)
 
 bool CLuaBaseEntity::gotoPlayer(std::string const& playerName)
 {
-    XI_DEBUG_BREAK_IF(m_PBaseEntity->objtype != TYPE_PC);
+    bool found = false;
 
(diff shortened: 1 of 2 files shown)


WinterSolstice8 (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -3239,10 +3243,14 @@ bool CLuaBaseEntity::gotoPlayer(std::string const& playerName)
 
 bool CLuaBaseEntity::bringPlayer(std::string const& playerName)
 {
+
```
TeoTwawki: bah. brb while I amend that extra blank line out.

claywar (approved):

PR #3445, opened 2022-12-29
[codeql] Add null checks for several warnings

claywar:
**_I affirm:_**

Adds null checks for several codeql warnings and logs output

No impact should be observed (unless null, then get logs)

### DIFF
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -582,7 +582,7 @@ void CZone::updateCharLevelRestriction(CCharEntity* PChar)
     {
         // If the level restriction is already the same then no need to change it
         CStatusEffect* statusEffect = PChar->StatusEffectContainer->GetStatusEffect(EFFECT_LEVEL_RESTRICTION);
-        if (statusEffect->GetPower() == m_levelRestriction)
+        if (statusEffect == nullptr || statusEffect->GetPower() == m_levelRestriction)
         {
             return;
         }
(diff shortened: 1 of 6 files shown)


Xaver-DaRed (approved):

src/map/zone.cpp
```diff
@@ -582,7 +582,7 @@ void CZone::updateCharLevelRestriction(CCharEntity* PChar)
     {
         // If the level restriction is already the same then no need to change it
         CStatusEffect* statusEffect = PChar->StatusEffectContainer->GetStatusEffect(EFFECT_LEVEL_RESTRICTION);
-        if (statusEffect->GetPower() == m_levelRestriction)
+        if (statusEffect == nullptr || statusEffect->GetPower() == m_levelRestriction)
```
TeoTwawki, line 585: shouldn't be physically possible to get here and have that be null, since it was just used in the conditional above

`if (PChar->StatusEffectContainer->HasStatusEffect(EFFECT_LEVEL_RESTRICTION))`

claywar: Agreed, but 3330 was also raised in a case where I thought it would be impossible to hit.  This is just adding additional safety per the warnings raised.

TeoTwawki: I think something else has to be going on, `HasStatusEffect` would have returned false, and I can't believe it was somehow be destroyed after passing that check but before the `GetStatusEffect` would execute.

Edit: per discord discussion, this wasn't expecting to fix anything, except to silence a warning from automation

WinterSolstice8 (approved):

PR #3456, opened 2023-01-01
[Nyzul] Fix missing settings and references

Xaver-DaRed:
**_I affirm:_**

- Fixes 4 settings that were lost in translation from original feature branch.
- Fixes certain npc references that were not converted when changing to using an offset.

Nyzul is, or should be, disabled.

### DIFF
--- a/scripts/globals/nyzul.lua
+++ b/scripts/globals/nyzul.lua
@@ -2,10 +2,11 @@
 -- Nyzul Isle Global
 -----------------------------------
 local ID = require("scripts/zones/Nyzul_Isle/IDs")
+require("scripts/globals/appraisal")
 require("scripts/globals/items")
+require("scripts/globals/settings")
 require("scripts/globals/status")
 require("scripts/globals/utils")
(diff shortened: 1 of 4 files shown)


scripts/globals/nyzul.lua
```diff
@@ -535,7 +536,7 @@ xi.nyzul.handleAppraisalItem = function(player, npc)
     for _, cofferID in ipairs(ID.npc.TREASURE_COFFER) do
         if npc:getID() == cofferID and npc:getLocalVar("opened") == 0 then
             -- Appraisal Items
-            local mobOffset = npc:getLocalVar("appraisalItem") - 17092723
+            local mobOffset = npc:getLocalVar("appraisalItem") - 17092723 -- TODO: What's this and how do we not have this here?
```
KnowOne134: Its a mobid, to offset the apprasial id to the mob that dropped the chest

claywar (approved):
Almost had me at the comparisons for settings, but saw moved to bool.  👍

PR #3460, opened 2023-01-01
[core] Add broader filters for logging

zach2good:
**_I affirm:_**

- Adds settings to filter out whole categories of logging levels, not just specific debug ones.
- Adds a bunch of comments
- Adds new console command: `reload_settings`
- EVEN MORE MACROS IN `logging.h`!!!!

- Copy the new settings across from `settings/default/logging.lua` to `settings/logging.lua`
- Change the settings however you like
- Either restart your server or use the new console command `reload_settings` to see changes in logging

Example:
- Don't alter the default settings
- Log in with a character
- See debug logging for ZoneCounterIncrease etc.
- Disable `LOG_DEBUG` and use `reload_settings`
- Zone your character, that debug logging is gone

### DIFF
--- a/settings/default/logging.lua
+++ b/settings/default/logging.lua
@@ -37,12 +37,20 @@ xi.settings.logging =
     --]]
     PATTERN = "[%D %T:%e][%&]%^[%n]%$ %v (%!:%#)",
 
-    DEBUG_SOCKETS        = false,
-    DEBUG_NAVMESH        = false,
-    DEBUG_PACKETS        = false,
-    DEBUG_ACTIONS        = false,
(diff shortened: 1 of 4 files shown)


claywar (approved):

zach2good:
Also, the nice thing about the settings system is that if people don't update their own settings, the defaults will fill in and the logging will continue to work as expected, rather than going dark

WinterSolstice8 (approved):

TeoTwawki (approved):

Xaver-DaRed (approved):

PR #3463, opened 2023-01-02
Implement Moghouse 2F

zach2good:
**_I affirm:_**

- [ ] Make sure "Change Floors" option isn't available if you've not completed enough exit quests
- [x] Make sure all exit quests still work
  - [x] Sandy
  - [x] Bastok
  - [x] Windy
  - [x] Jeuno
  - [x] AhtUrghan
- [x] Make sure you can't get to 2F through packet abuse
- [x] Log and reject any packet abuse around Moghouse zoneline and 2F
- [x] Changing floors (1F)
- [x] Changing floors (2F)
- [x] Exit from 2F, re-enter into 1F
- [x] Exit from 1F, re-enter into 1F
- [x] Log out in 1F, log in to 1F
- [x] Log out in 2F, log in to 2F
- [x] DC in 1F, log in to 1F
- [x] DC in 2F, log in to 2F
- [x] Furniture placement (2F)
- [x] Hide Moogle in 2F
- [x] Cutscene telling you you've unlocked 2F
  - [x] Cutscene works, but Moogle is stuck in DISAPPEAR state... 
- [x] Mog patio item + usage
- [x] Mog patio KI and checking when remodelling
- Remodelling
  - [x] Sandoria
  - [x] Bastok
  - [x] Windurst
  - [x] Patio (+ enable with KI)
  - [x] Remodelling (changing) while you're on 2F zones you so the change can take effect
- [x] Ensure 2F furniture doesn't increase storage (is that right?)
- [x] Ensure 2F furniture doesn't affect Moghancement or Moglification (is that right?)
- [x] Ensure it isn't possible to do Gardening on 2F
  - It looks like its possible to do gardening on your existing pots in your STORAGE_1, like when you're in a rent-a-room
  - If you have pots available to place in your storage, they're filtered out when you open the menu
  - Should still be checking, because, packet abuse

NOTE: Change Floors option is greyed out when you're in a rent-a-room, that could be a good place to look for how to grey out this option when you're at home.

### Out of scope

- Invite to Moghouse
- Exit to Mog Garden

Go through the checklist above

### DIFF
--- a/scripts/quests/bastok/A_Ladys_Heart.lua
+++ b/scripts/quests/bastok/A_Ladys_Heart.lua
@@ -94,7 +94,7 @@ quest.sections =
                         if quest:complete(player) then
                             player:confirmTrade()
                             local mhflag = player:getMoghouseFlag()
-                            player:setMoghouseFlag(bit.band(mhflag, 0x0002))
+                            player:setMoghouseFlag(mhflag + 0x0002)
                             player:messageSpecial(portBastokID.text.MOGHOUSE_EXIT)
                         end
 
--- a/scripts/quests/bastok/A_Ladys_Heart.lua
+++ b/scripts/quests/bastok/A_Ladys_Heart.lua
@@ -93,7 +93,8 @@ quest.sections =
                     if option == 2002 then
                         if quest:complete(player) then
                             player:confirmTrade()
-                            player:setMoghouseFlag(2)
+                            local mhflag = player:getMoghouseFlag()
+                            player:setMoghouseFlag(bit.band(mhflag, 0x0002))
                             player:messageSpecial(portBastokID.text.MOGHOUSE_EXIT)
                         end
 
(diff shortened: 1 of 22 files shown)


scripts/quests/bastok/A_Ladys_Heart.lua
```diff
@@ -92,8 +92,8 @@ quest.sections =
                     -- Correct trade option (Amaryllis).
                     if option == 2002 then
                         if quest:complete(player) then
-                            player:confirmTrade()
```
Xaver-DaRed, line 95: Don't remove the confirmTrade, Sir. It's needed.

zach2good: Whoops, good catch 👍

claywar (approved):

src/map/zone.h
```diff
@@ -232,7 +232,6 @@ enum ZONEID : uint16
     ZONE_DYNAMIS_BASTOK                 = 186,
     ZONE_DYNAMIS_WINDURST               = 187,
     ZONE_DYNAMIS_JEUNO                  = 188,
-    ZONE_189                            = 189, // Southern San d'Oria [S] Residential Area
```
TeoTwawki, line 235: we actually want to keep the placeholder, and just kill the false statement about what it is.

TeoTwawki: else we potentially have a nullptr in other places, even tho we aren't actively using that zone

zach2good: Yessir 👍

PR #3466, opened 2023-01-03
[Nyzul] Typo correction and fix negative pathos logic

Xaver-DaRed:
**_I affirm:_**

 - Corrects a typo in a variable name.
 - Corrects logic when applying a negative pathos, so we guarantee that:
     - A negative pathos is always set when possible (and when aplicable)
     - The pathos set is random

Nyzul is disabled?

### DIFF
--- a/scripts/globals/nyzul.lua
+++ b/scripts/globals/nyzul.lua
@@ -1014,46 +1014,52 @@ xi.nyzul.addPenalty = function(mob)
 
     -- Status effect penalty
     else
+        -- Create table with negative pathos that are not currently set.
+        local availablePathos = {}
+
         for i = 1, 17 do
-            local randomEffect = math.random(1, 17) -- TODO: Ask KO why we are looping 17 times for 17 random rolls. Shouldn't we be checking for "i" down below?
-                                                    -- NOTE: "i" is unused. 17 is the amount of different floor layouts - 1 (We start at 0, boss layout)
+            if not utils.mask.getBit(pathos, i) then
+                table.insert(availablePathos, i)
+            end
+        end
 
-            if not utils.mask.getBit(pathos, randomEffect) then
-                instance:setLocalVar("floorPathos", utils.mask.setBit(pathos, randomEffect, true))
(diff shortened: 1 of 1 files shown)


scripts/globals/nyzul.lua
```diff
@@ -654,7 +654,7 @@ xi.nyzul.tempBoxPickItems = function(npc)
 end
 
 xi.nyzul.tempBoxFinish = function(player, csid, option, npc)
-    local ID = require("scripts/zones/"..player:getZoneName().."/IDs")
+    local ID = require("scripts/zones/"..player:getZoneName().."/IDs") -- ?
```
claywar: This could be replaced with `local ID = zones[player:getZoneID()]`

scripts/globals/nyzul.lua
```diff
@@ -753,16 +753,19 @@ xi.nyzul.handleRunicKey = function(mob)
         for _, entity in pairs(chars) do
             -- Does players Runic Disk have data saved to a floor of entering or higher
             if
-                entity:getVar("NyzulFloorProgress") + 1 >= startFloor and
+                entity:getVar("NyzulFloorProgress") > startFloor and
```
claywar: Are we sure a 0 index isn't used for something else here?

scripts/globals/nyzul.lua
Xaver-DaRed: Players WILL be in Nyzul though.
This function is only used once, in \scripts\zones\Nyzul_Isle\npcs\Armoury_Crate.lua

We already have ID defined at the top of the file
`local ID = require("scripts/zones/Nyzul_Isle/IDs")`

So I'm thinking this `ID` re-definition is unneccesary

I was waiting, in case KO had some knowledge on this (And a comment down below)

EDIT: I've decided to remove it, in the end

scripts/globals/nyzul.lua
Xaver-DaRed: Good catch. Reverted it.

claywar (approved):

PR #3481, opened 2023-01-07
[map] Fix long names returning NULL PChar for certain zmq message types

WinterSolstice8:
**_I affirm:_**

Fixes long named characters to have the following work on a different process:
* tell
* linkshell promote to sack
* linkshell kick

The effective change expands out to

`char characterName[15] = {};`
->
`char characterName[16] = {};`
but only memcpy 15 characters to guarantee the null terminator

have longest name character, attempt to /tell, kick from ls, or promote to sack from a different process and have it work

### DIFF
--- a/src/map/message.cpp
+++ b/src/map/message.cpp
@@ -109,8 +109,8 @@ namespace message
             }
             case MSG_CHAT_TELL:
             {
-                char characterName[15] = {};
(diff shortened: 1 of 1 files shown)


claywar (approved):

zach2good (approved):
Thanks for using the `PacketNameLength` definition!

PR #3484, opened 2023-01-10
[Quest] Test My Mettle Implementation

bkesecker:
**_I affirm:_**

It sure has been easy to get a pair of Power Sandals before this work.  Not really sure if anyone will be happy this will start working.

Does not close a known issue.

- Meet level, rank, and fame requirements
- Zone to Selbina, !gotoname Devean
- Bet him you can find the sandals
- Zone to Davoi, !gotoname Jar and interact
- Zone to Selbina and either talk or trade with Devean
- Verify bet amount and sandal status (removed or still in storage)
- Repeat (there are a few permutations)

Unrelated to the quest logic
- Go to Davoi and find the pot, wait for it to move on a random game hour


scripts/quests/otherAreas/Test_My_Mettle.lua
```diff
@@ -0,0 +1,438 @@
...
+    -- option = 1073741824 value for first, second, third, and fourth decline
+    -- binary = 0100 0000 0000 0000 0000 0000 0000 0000
+    -- option = 124  100 gil  24 hours
+    -- option = 286  250 gil  36 hours
+    -- option = 322  250 gil  72 hours
+    -- option = 548  500 gil  48 hours
+    -- option = 1060 1000 gil 60 hours
+    -- option = 1072 1000 gil 72 hours
+    
+    if option ~= 1073741824 then
```
claywar: This really should never be a direct compare, and if there is the option to cancel a cs, we should only be checking the relevant bits.

scripts/quests/otherAreas/Test_My_Mettle.lua
```diff
@@ -0,0 +1,438 @@
...
+        -- if the sandals were traded, Devean says too bad
+        return quest:progressEvent(npcEventID.devean.FAILURE)
+    end
+end
+
+local function recordOption(player, option)
+    -- option examples:
+    -- option = 1073741824 value for first, second, third, and fourth decline
+    -- binary = 0100 0000 0000 0000 0000 0000 0000 0000
+    -- option = 124  100 gil  24 hours
```
claywar: These options are a bitfield, and should be parsed as such

scripts/quests/otherAreas/Test_My_Mettle.lua
```diff
@@ -0,0 +1,438 @@
...
+    end
+
+    return false
+end
+
+local function removeSandals(player)
+    -- confirm trade if it was traded...
+    player:confirmTrade()
+
+    if player:hasItem(xi.items.POWER_SANDALS, xi.inv.INVENTORY) then
```
claywar: Not understanding the point here.  If the player traded, there's no need for this logic

scripts/quests/otherAreas/Test_My_Mettle.lua
```diff
@@ -0,0 +1,438 @@
...
+
+    -- must clear bet to prevent a player from being penalized more than once
+    quest:setVar(player, 'bet', 0)
+end
+
+local function success(player)
+    removeSandals(player)
+
+    local gil = quest:getVar(player, 'bet')
+    local reward = xi.settings.main.GIL_RATE * gil
```
claywar: npcUtil.giveCurrency should be used here, remove reward * settings multiplier and the direct addGil call along with messageSpecial

scripts/quests/otherAreas/Test_My_Mettle.lua
```diff
@@ -0,0 +1,438 @@
...
+
+    npcUtil.queueMove(npc, newPosition)
+end
+
+local function interactJar(player, npc)
+    if player:hasItem(xi.items.POWER_SANDALS) or not canRepeat(player) then
+        return -- use default action
+    end
+
+    local questFlagged = player:getQuestStatus(xi.quest.log_id.OTHER_AREAS, xi.quest.id.otherAreas.TEST_MY_METTLE) ~= QUEST_AVAILABLE
```
claywar: Why?  We have quest sections for this, and this script is dedicated to that specific quest.  There is absolutely no reason you should need to pull this value in a function.

scripts/quests/otherAreas/Test_My_Mettle.lua
```diff
@@ -0,0 +1,438 @@
...
+local quest = Quest:new(xi.quest.log_id.OTHER_AREAS, xi.quest.id.otherAreas.TEST_MY_METTLE)
+
+-- do not set gil, it is not a static amount
+quest.reward =
+{
+    fame = 30,
+    fameArea = xi.quest.fame_area.SELBINA_RABAO,
+}
+
+local npcEventID =
```
claywar, line 29: While I love the idea of reducing magic numbers, this actually makes things more obtuse.  It's fairly easy to track event processes in the trigger/finish flow, and this forces the next maintainer to look here.

Also, why a subtable with only one entry?  There's no point to this.

claywar (commented):
There's really almost too much abstraction here for a simple quest.  This really should be simplified in order to create better readability and maintainability.

scripts/zones/Selbina/IDs.lua
```diff
@@ -44,7 +44,8 @@ zones[xi.zone.SELBINA] =
     },
     npc =
     {
-        EXPLORER_MOOGLE = 17793131,
+        DEVEAN_DEFAULT                = 124,
```
Xaver-DaRed: DEVEAN_DEFAULT is an event.
Events do not shift on version updates. so we do not define them in the IDs table.

So this should be removed

scripts/zones/Selbina/DefaultActions.lua
```diff
@@ -3,6 +3,7 @@ local ID = require("scripts/zones/Selbina/IDs")
 return {
     ['Abelard']      = { text = ID.text.ABELARD_DIALOG },
     ['Elfriede']     = { event = 25 },
+    ['Devean']       = { event = ID.npc.DEVEAN_DEFAULT },
```
Xaver-DaRed: EDIT: I am wrong. You defined an event, just put the event number here (124)

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: Talking to the NPC can result in successful completion, thus there is no trade.  No errors/issues calling `confirmTrade` when there is none.

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: This value is used in a few other cancelled dialog scenarios, such as Lest We Forget, Rune of Transfer, and Splintery Chest.  Can you reference an example, so that I can follow a better example?

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: Correct.  Missed during conversion to the interaction framework.

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: I saw an opportunity to make the code more readable.  Instead of using short-term memory remembering which values matched, use words.  Is it likely for the IDs to change with subsequent releases of the game?

For the subtable, thought it might be better practice to show that it could be used in situations where there were more than one NPC.

scripts/quests/otherAreas/Test_My_Mettle.lua
TeoTwawki: Retail specifically does not allow you to complete the quest and keep the sandals, I had to redo the quest and then intentionally not turn them in to keep them and the quest specifically would not end without my trading them. Talking to him while wearing them doesn't end the quest, only a trade will. 

I have a mule that is now the proud owner of vit shoes I'll never use.

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: Fixed.

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: Fixed.

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: They are not bit fields.  For example, a bet of 100 and 24 hours results in 124 and 250 with 36 hours results in 286.  I see no way to validly `bit.band` these option values.  Does anyone else?

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: @claywar I found a potentially better example in `besieged.lua` and `atma.lua` that uses `option < 0x40000000`.  Perhaps I can open another PR after this one to standardize the usage.  What is your take?

scripts/zones/Selbina/IDs.lua
bkesecker: Since this was more confusing than maintainable, I removed it.

scripts/zones/Selbina/DefaultActions.lua
bkesecker: Changed to use number.

scripts/quests/otherAreas/helpers.lua
```diff
@@ -0,0 +1,42 @@
...
+        {   25.080,  0.440, -107.300 }, -- [5] H-8 Jar near campfire
+        { -106.770,  2.580,   26.060 }, -- [6] F-6 Jar north of Hut Flap
+        { -128.720,  3.250, -237.600 }, -- [7] F-9 Jar near stump
+        {  -36.720,  3.300, -218.270 }, -- [8] G-9 Jar near well
+        {   54.800,  0.720, -202.950 }, -- [9] H-9 Jar near campfire
+    }
+
+    local newPosition = npcUtil.pickNewPosition(davoiID.npc.JAR, positions, false)
+
+    npcUtil.queueMove(npc, newPosition)
```
Xaver-DaRed, line 41: All this logic, in general should not live in a quest helper file, imo. It should live in the zone.lua file.

The resoning behind this, is that the NPC exists and moves independently from the quest.

bkesecker: @Xaver-DaRed  Given other existing implementations, I agree, but it is being called by the quest Lua too.  I suspect it would need duplicated in `Zone.lua` and the quest Lua if not in a helper.  Thus, much like what was applied to the Gobbie Bag questline, used that setup as a compromise (and also satisfy sanity checks).

scripts/quests/otherAreas/Test_My_Mettle.lua
bkesecker: Thanks for the retail account @TeoTwawki, I have addressed the following points:

- Retail specifically does not allow you to complete the quest and keep the sandals (player can fail when time runs out without the sandals),
- the quest specifically would not end without trading them, and
- only a trade will end the quest.

These details allowed the logic to be simplified, so all the easier to review.

bkesecker:
> There's really almost too much abstraction here for a simple quest. This really should be simplified in order to create better readability and maintainability.

With the changes Teo pointed out, many of the previously defined methods were eliminated.  I have fully retested all permutations and will squash.  Unless there are further comments, I am considering this effort complete.

Issue #3412, opened 2022-12-20
🐛 Overrides not applying correctly


CatsEyeXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Windows 10

## Branch affected by issue 

`base`

## Steps to reproduce
Create the sample module below, none of the prints will fire.

```
-----------------------------------
-- Common Requires
-----------------------------------
require("modules/module_utils")
require("scripts/globals/npc_util")
require("scripts/globals/status")
require("scripts/globals/zone")
-----------------------------------
local ID = require("scripts/zones/Caedarva_Mire/IDs")

local m = Module:new("wild_karakul")

m:addOverride("xi.zones.Caedarva_Mire.mobs.Wild_Karakul.onMobSpawn", function(mob)
    print("WILD KARAKUL SPAWNED!")
end)

m:addOverride("xi.zones.Caedarva_Mire.mobs.Wild_Karakul.onMobFight", function(mob, target)
    print("WILD KARAKUL FIGHTING!")
end)

m:addOverride("xi.zones.Caedarva_Mire.mobs.Wild_Karakul.onMobDeath", function(mob, player, optParams)
    print("WILD KARAKUL DIED!")
end)

m:addOverride("xi.zones.Caedarva_Mire.mobs.Wild_Karakul.onMobDespawn", function(mob)
    print("WILD KARAKUL DESPAWNED!")
end)

return m
```

## Expected behavior

Prints should fire in each function

Per Zach:
`Presumably luautils::init(); // Also calls moduleutils::LoadLuaModules(); is being fired before zoneutils::LoadZoneList(); and the cache entries are getting clobbered
`

TonzoffunCE:
Hey @zach2good, do you have any pointers or direction as to where we could look in the code, so we can try to fix this issue? Any insight would be greatly appreciated.

zach2good:
The issue here is how the following functions interact, and the order in which they interact:
[...]

PR #3495, opened 2023-01-14
[Modules] Simplify HNM system logic

Xaver-DaRed:
**_I affirm:_**

Simplifies logic of both hnm system modules and makes sure they work.

Unrelated to the actual issue in Issue #3412 but it was mentioned.

Enable ONE of them.
Kill a Land King
Make sure server variables get set/updated when it despawns


modules/custom/lua/custom_HNM_system.lua
```diff
@@ -61,8 +67,14 @@ hnmSystem:addOverride("xi.zones.Dragons_Aery.mobs.Fafnir.onMobDespawn", function
...
 
 hnmSystem:addOverride("xi.zones.Dragons_Aery.npcs.qm0.onTrade", function(player, npc, trade)
-    if not GetMobByID(dragonsAeryID.mob.FAFNIR):isSpawned() and not GetMobByID(dragonsAeryID.mob.NIDHOGG):isSpawned() then
-        if npcUtil.tradeHasExactly(trade, 3340) and npcUtil.popFromQM(player, npc, dragonsAeryID.mob.NIDHOGG) then
+    if
+        not GetMobByID(dragonsAeryID.mob.FAFNIR):isSpawned() and
+        not GetMobByID(dragonsAeryID.mob.NIDHOGG):isSpawned()
+    then
+        if
+            npcUtil.tradeHasExactly(trade, 3340) and
```
claywar: Can these trade items get added to items.lua?

Also, the nested condition could be simplified to just one, since there's only one single path

modules/custom/lua/custom_HNM_system.lua
```diff
@@ -106,8 +120,14 @@ hnmSystem:addOverride("xi.zones.Valley_of_Sorrows.mobs.Adamantoise.onMobDespawn"
...
 
 hnmSystem:addOverride("xi.zones.Valley_of_Sorrows.npcs.qm1.onTrade", function(player, npc, trade)
-    if not GetMobByID(valleySorrowsID.mob.ADAMANTOISE):isSpawned() and not GetMobByID(valleySorrowsID.mob.ASPIDOCHELONE):isSpawned() then
-        if npcUtil.tradeHasExactly(trade, 3344) and npcUtil.popFromQM(player, npc, valleySorrowsID.mob.ASPIDOCHELONE) then
+    if
+        not GetMobByID(valleySorrowsID.mob.ADAMANTOISE):isSpawned() and
+        not GetMobByID(valleySorrowsID.mob.ASPIDOCHELONE):isSpawned()
+    then
+        if
+            npcUtil.tradeHasExactly(trade, 3344) and
```
claywar: Same as above

modules/custom/lua/custom_HNM_system.lua
```diff
@@ -151,8 +173,14 @@ hnmSystem:addOverride("xi.zones.Behemoths_Dominion.mobs.Behemoth.onMobDespawn",
...
 
 hnmSystem:addOverride("xi.zones.Behemoths_Dominion.npcs.qm2.onTrade", function(player, npc, trade)
-    if not GetMobByID(behemothDomID.mob.BEHEMOTH):isSpawned() and not GetMobByID(behemothDomID.mob.KING_BEHEMOTH):isSpawned() then
-        if npcUtil.tradeHasExactly(trade, 3342) and npcUtil.popFromQM(player, npc, behemothDomID.mob.KING_BEHEMOTH) then
+    if
+        not GetMobByID(behemothDomID.mob.BEHEMOTH):isSpawned() and
+        not GetMobByID(behemothDomID.mob.KING_BEHEMOTH):isSpawned()
+    then
+        if
+            npcUtil.tradeHasExactly(trade, 3342) and
```
claywar: Same as above

modules/era/lua/era_HNM_system.lua
```diff
@@ -15,249 +14,250 @@ local behemothDomID   = require("scripts/zones/Behemoths_Dominion/IDs")
...
+    super(zone)
+
     local hnmPopTime   = GetServerVariable("[HNM]Fafnir")   -- Time the NM will spawn at.
     local hnmKillCount = GetServerVariable("[HNM]Fafnir_C") -- Number of times NQ King has been slain in a row.
+    local currentTime  = os.time()
 
     -- First-time setup.
     if hnmPopTime == 0 then
-        hnmPopTime = os.time() + math.random(1, 48) * 1800
+        hnmPopTime = currentTime + math.random(1, 48) * 1800
```
cocosolos, line 39: For the era module (maybe the custom too, idk) I believe the window is only supposed to be 3 hours instead of 24 from when servers come up, cutting the random roll down to 6 from 48.

modules/era/lua/era_HNM_system.lua
```diff
@@ -15,249 +14,250 @@ local behemothDomID   = require("scripts/zones/Behemoths_Dominion/IDs")
...
-    -- HQ King.
-    if hnmKillCount > 3 and (math.random(1, 5) == 3 or hnmKillCount > 6) then
-        UpdateNMSpawnPoint(dragonsAeryID.mob.NIDHOGG)
+    -- Calculate monster to pop.
+    local monster = dragonsAeryID.mob.FAFNIR
+
+    if
+        hnmKillCount > 3 and
+        (math.random(1, 5) == 3 or hnmKillCount > 6)
+    then
```
cocosolos, line 50: Not sure how accurate this is, but [this 2009 wiki revision](https://ffxiclopedia.fandom.com/wiki/Adamantoise?oldid=1044368) says that high tier is available to pop after server restart. I assume this applies for Aspid, KB, and Nid if true.

modules/era/lua/era_HNM_system.lua
Xaver-DaRed: Correct.

This section is the "first time" setup. It's there in case there is no server variable, either becouse the system has never been used or becouse it has been deleted on server operator error. This makes it so they dont immediately pop in said cases and randomizes when it pops, to prevent player abuse.

The onMobDespawn sections of each nm contain the regular windows

```
local randomPopTime = 75600 + math.random(0, 6) * 1800 -- 21 hours minimum pop time. 7 windows. 1 every 30 mins.

SetServerVariable("[HNM]Fafnir", os.time() + randomPopTime) -- Save next pop time.

GetMobByID(dragonsAeryID.mob.FAFNIR):setRespawnTime(randomPopTime) -- Sets the respawn countdown
```

modules/era/lua/era_HNM_system.lua
Xaver-DaRed: Since this system is supposed to perpetuate/save the status of HNMs in case of accidental crash, which didnt happen in retail, I skip the reseting of the "kill count" variable to day 3. Again, to prevent player abuse by forcefully/puposely crashing the server to exploit the system.

modules/era/lua/era_HNM_system.lua
cocosolos: I was thinking server ops could delete the vars during scheduled downtime to give the old retail experience of logging out in aery before downtime to catch the early windows, while also catching unexpected crashes and persisting the ToD. The catching unexpected crashes looks like it works as intended but there isn't really a way to simulate that old maintenance camping with this setup. It would still result in a minimum 30 min downtime but only a max 3 hours instead of 24 hours of possible windows.

TeoTwawki: Ultimately, this is an optional module and server ops are gonna edit as needed but preventing abuse of crashed was a highly desired feature that lead to lots of help requests from people wanting to know how to do it (which means lead to lots of answering the same questions).

Myself on my own server got a step the other direction and instead of trying to retain the prior time remaining from before a crash, I punish the entire world by resetting everything to its full cooldown so that server crashing makes _nobody_ a winner (seen to many people actively seek new ways to crash a server just to make themselves get the claim over someone else, even when the ToD wasn't an issue - and that made me a leeeetle angry).

claywar (approved):

PR #3497, opened 2023-01-15
[core] Make sure cache entries aren't overwritten on multiple file reads

zach2good:
**_I affirm:_**

I kept a bunch of notes of my findings in https://github.com/LandSandBoat/server/issues/3412, this moves some things around and makes sure that cache entries aren't overwritten unless you pass `true` into the caching function.

Closes: https://github.com/LandSandBoat/server/issues/3412

**THIS WILL REQUIRE A LOT OF TESTING TO MAKE SURE EVERYTHING STILL WORKS AS EXPECTED: Hot-reloading, modules, etc.**

See my notes in https://github.com/LandSandBoat/server/issues/3412

### DIFF
--- a/modules/custom/lua/claim_shield.lua
+++ b/modules/custom/lua/claim_shield.lua
@@ -75,7 +75,7 @@ for _, entry in pairs(nmsToShield) do
 
             mobArg:setClaimable(true)
             mobArg:setUnkillable(false)
-            mob:setCallForHelpBlocked(false)
+            mobArg:setCallForHelpBlocked(false)
 
             mobArg:resetAI()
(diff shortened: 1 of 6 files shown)


zach2good:
For testing:

I _believe_ it should be sufficient to temporarilty swap out
```cpp
 table[sol::override_value][part] = file_result;
```
with 
```cpp
table[sol::update_if_empty][part] = file_result;
```

This will disable hot-reloading while you're testing, but it will be sufficient to make sure cache entries aren't getting clobbered

zach2good:
Tested by Xaver and Carver, looks like its stable 👍

claywar (approved):

PR #3501, opened 2023-01-17
[core] Better handling of dropped 0x00A packets on zoning in

WinterSolstice8:
**_I affirm:_**

This changes 0x00A handling to copy the syncCount to CZoneInPacket,

The client expects CZoneInPacket to match its sent sync count and otherwise will not accept the data. This means that CZoneInPacket (0x00A) *cannot* be cached.

On ashitav4, install this addon and load:
[test.zip](https://github.com/LandSandBoat/server/files/10434571/test.zip)
Block server to client 0x00A, try to zone. wait for counters to go up, unblock, and be able to zone in.

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -783,6 +783,15 @@ int32 parse(int8* buff, size_t* buffsize, sockaddr_in* from, map_session_data_t*
 
     if (ref<uint16>(buff, 2) != map_session_data->server_packet_id)
     {
+        /*
+         * If the client and server have become out of sync, then caching takes place. However, caching
+         * zone packets will result in the client never properly connecting. Ignore those specifically.
+         */
+        if (SmallPD_Type == 0x0A)
+        {
+            return 0;
+        }
+
         ref<uint16>(map_session_data->server_packet_data, 2) = SmallPD_Code;
         ref<uint16>(map_session_data->server_packet_data, 8) = (uint32)time(nullptr);
 
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -248,6 +248,22 @@ void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PCh
         return;
     }
 
+    /*
+     * Handle out of sync zone correction..
+     */
+    if (data.ref<uint16_t>(0x02) > 1)
+    {
+        PSession->server_packet_id = data.ref<uint16_t>(0x02);
+
+        // Clear all pending packets for this character.
+        // This incoming 0x00A from the client wants us to set the the starting sync count for all new packets to the sync count from 0x02.
+        // If we do not do this, all further packets may be ignored by the client and will result in disconnection from the server.
+        if (PChar)
+        {
+            PChar->clearPacketList();
+        }
+    }
+
     if (PSession->blowfish.status == BLOWFISH_WAITING) // Generate new blowfish session, call zone in, etc, only once.
     {
         PChar->clearPacketList();


WinterSolstice8:
drafting because its 3 am and I forgot co authors

zach2good:
Snippet from atom0s:

```
atom0s — 17/01/2023 11:50
To simplify on that approach, you can correct it with just 4 lines (not including comments).
I don't have updated lsb to make patch against but, in map.cpp line 794 atm: https://github.com/LandSandBoat/server/blob/base/src/map/map.cpp#L794
Inside of this if at the start of it, you can add:
```
```cpp
        /**
         * If the client and server have become out of sync, then caching takes place. However, caching
         * zone packets will result in the client never properly connecting. Ignore those specifically.
         */
        if (SmallPD_Type == 0x0A)
            return 0;
```
```
Inside of SmallPacket0x00A at the top after the first few lines, before the blowfish status check add:
```
```cpp
    /**
     * Handle out of sync zone correction..
     */
    if (data.ref<uint16_t>(0x02) > 1)
        PSession->server_packet_id = data.ref<uint16_t>(0x02);
```
```
And it'll do the same thing.
```

I dunno how well these different styles of patches are going on Horizon; I'd imagine we're going to wait for them to stabilise, and then we can take whatever code changes and configuration tips (for the wiki) they used.

zach2good (approved):

atom0s:
I wanted to expand on the convo bits of mine that Zach shared above in regards to this pull and what it's for / what it does for those that may read this and are unaware. 

The current socket implementation inside of LSB (and all previous forks of the project) does not properly implement FFXI's socket setup for the main game traffic, which uses UDP. The current socket implementation that is used inside of LSB is taken from rAthena and tweaked to work with FFXI, but does not properly recreate the actual expected setup of XI's. Instead, as-is the socket is being treated like TCP where all traffic is being constantly resent to clients if they fail to receive something, and will continue to happen until the client either does receive it or disconnects.

On retail, this is partially correct, where the server will try to resend data to the client (and vice-versa) until a certain distance of data has been reached, which will cause you to R0. However, with this setup, retail also implements a packet priority system which is used to help further reduce the flow of unneeded traffic when congestion happens or in specific zones / under certain conditions. (ie. like Besieged.) _(In zones like Besieged, there is a second 'layer' of control used which implements a grid-like system to further reduce the amount of traffic you get.)_

Due to how things are setup on LSB currently, the caching of dropped packets is causing multiple issues. The one this pull request is aimed to fix is that you cannot successfully enter a zone if your client dropped the initial response from the server.

A quick rundown of what happens is, when your client is entering a new zone, you send a 0x000A (`GP_CLI_COMMAND_LOGIN`) request to the server. This tells the server that your client is ready to begin loading into the zone and is requesting information on what it needs to properly load in. The server will respond with a 0x000A packet (`GP_SERV_COMMAND_LOGIN`) when this request is successful. However, if this packet is dropped and never reaches your client, you will never be able to complete the zone-in process and will just time out and have to reconnect and try again. 

This is happening because the original 0x000A response from the server is being queued and resent each time the client requests it, however, the server has a special case with the sync count on this packet request/response chain. When the client fails to receive the response, it will continue to increment its packet sync count and request it again. This causes the client to expect that the server will respond with the sync count realigned to the clients. (This is done to drop and ignore all previous data that is sent with the zone-in response that the client may have gotten out of order / without the 0x000A response.)

As-is, LSB is queueing the packet and resending it with the old/original sync count of 0x01, when it should be realigned to the client with the other data resent as well. 

This pull is a temporary fix/bandaid to fix the blackscreen / failing to zone issue for now until a more complete overhaul can be done to the socket setup.

---

In regards to the patches to be done, this is the updated adjustments to be made to the code:

The first patch is in `map.cpp`, which tells the parser to not queue zone responses:
https://github.com/LandSandBoat/server/blob/1ce6689502ca3027fbbbdca1c074df8974680127/src/map/map.cpp#L805

```cpp
    if (ref<uint16>(buff, 2) != map_session_data->server_packet_id)
    {
        /**
         * If the client and server have become out of sync, then caching takes place. However, caching
         * zone packets will result in the client never properly connecting. Ignore those specifically.
         */
        if (SmallPD_Type == 0x0A)
            return 0;

        ref<uint16>(map_session_data->server_packet_data, 2) = SmallPD_Code;
        ref<uint16>(map_session_data->server_packet_data, 8) = (uint32)time(nullptr);

        g_PBuff   = map_session_data->server_packet_data;
        *buffsize = map_session_data->server_packet_size;

        map_session_data->server_packet_data = buff;
        return -1;
    }
```

The second patch is in the 0x000A request handler to check for out-of-sync issues and realign the response:
https://github.com/LandSandBoat/server/blob/1ce6689502ca3027fbbbdca1c074df8974680127/src/map/packet_system.cpp#L243

```cpp
void SmallPacket0x00A(map_session_data_t* const PSession, CCharEntity* const PChar, CBasicPacket data)
{
    TracyZoneScoped;
    data.ref<uint32>(0x5C) = 0;

    /**
     * Handle out-of-sync zone correction..
     */
    if (data.ref<uint16_t>(0x02) > 1)
        PSession->server_packet_id = data.ref<uint16_t>(0x02);

    if (PSession->blowfish.status == BLOWFISH_ACCEPTED && PChar->status == STATUS_TYPE::NORMAL) // Do nothing if character is zoned in
    {
        ShowWarning("packet_system::SmallPacket0x00A player '%s' attempting to send 0x00A when already logged in", PChar->GetName());
        return;
    }

    if (PSession->blowfish.status == BLOWFISH_WAITING) // Generate new blowfish session, call zone in, etc, only once.
    {
```

_Again, this is not a full fix to the actual underlying issue(s) at hand, this is just a temp fix for the time being to reduce the number of black screens while zoning.)_

WinterSolstice8:
Undrafting, tested locally with the blocker addon and works as expected. This is also nearly identical to what Horizon's using (I added a nullptr check to PChar)

zach2good (approved):

Xaver-DaRed (approved):

PR #3508, opened 2023-01-19
Re-align region IDs with what the client actually uses.

TeoTwawki:
(as seen in ROM/165/78.dat)

**_I affirm:_**

https://gist.github.com/TeoTwawki/d46e51e761a95c6dffe11f0dc999f4d1

Tested as far as build and boot, Xaver was going to do some further tests for me. Potential issues anywhere the Lua does a greater than or less than check on regions, and where regions equal a bit in a mask.


claywar (dismissed):

TeoTwawki:
Note: I am expecting issues related to Lua side region checks involved with bitmasks for teleports. Namely the expected bits might not be in the same expected order as the regions anymore. And I don't really know how to correct it should that prove to be the case.

zach2good (dismissed):
LGTM, but we don't need git history like: "Edits purely to satisfy CI"

TeoTwawki:
so far, from xavers testing:
- I broke supplies
- I broke conquest NPC selection for spawning
- I broke conquest ranking display
- I broke outpost warps
- I broke unlocking them

I had this in draft mode I dunno what switched it, putting it back till issues resolved and dismissing current aprovals.

TeoTwawki:
well this is gonna be a fun rebase! and then I still have to actually fix it.

TeoTwawki:
I think I'll split off the enum renaming parts, so that this deals solely with re-ordering.

TeoTwawki:
I'm pretty much never going to have the time to return to this and fix its issues, and the rebase its needs will be straight up hellish. So no point in keeping this open.

PR #3509, opened 2023-01-19
Fix merit exploit

### LINKED ISSUE #3507: 🐛 Potential exploit for free merit upgrades and ignoring merit cap
**_I affirm:_**

Windows10

## Branch affected by issue 

`base`

## Steps to reproduce

1. Login to a character with access to merits.
2. Have plenty of available merit points (perhaps use `!givexp 10000` a few times).
3. Have merit window open to show the merit you want to cheat on.
4. Inject a packet to raise a merit (must be done with an addon. See below for code example).
5. Repeat step 4 until you reach cap, and then continue repeating step 4.
6. Observe that you're able to get more merits than the cap, *and* the merits after cap don't cost any merit points either.

Note that there is separate code that caps the merit count returned in `GetMeritValue()` to 15 (with a lower cap at levels lower than 80), so this exploit only really allows you to reach 15 merits, even if it will display more.

## Expected behavior

6. Merits do not increase beyond the cap, and no free merit upgrades.

This is caused by lack of validation in the code processing the raise merit packet and the default value of a static array being 0 (see the `upgrade[10][45]` array [here](https://github.com/LandSandBoat/server/blob/base/src/map/merit.cpp#L38). When you hit ~45 merits or so it'll overflow to the next row in that array and start costing merits again. Eventually you can hit out of bounds on the `upgrade` array.

Here's an example addon for Windower to test this:

To use, do `//exploit raisemerit <merit_id>`. For example, `//exploit rm 2252` for the BLM group 2 merit Ancient Magic Magic Attack Bonus.

```Lua
-- Removed
```

DiscipleOfEris:
**_I affirm:_**

Add some validation to `RaiseMerit()`, both to check the max allowable upgrades (specified in the `upgrade` column of `merits.sql`) and the max allowable merits in that category (specified in [`meritCatInfo`](https://github.com/LandSandBoat/server/blob/base/src/map/merit.cpp#L93)).

Closes #3507

See #3507

### DIFF
--- a/src/map/merit.cpp
+++ b/src/map/merit.cpp
@@ -274,6 +274,32 @@ uint8 CMeritPoints::GetMeritPoints() const
 
 /************************************************************************
  *                                                                       *
+ *  Return the total merits in the same category as the given merit.     *
+ *                                                                       *
+ ************************************************************************/
+
+uint16 CMeritPoints::GetMeritCountInSameCategory(MERIT_TYPE merit)
+{
+    if (!this->IsMeritExist(merit))
+    {
+        return 0;
+    }
+
+    Merit_t* PMerit = Categories[GetMeritCategory(merit)];
+
+    uint16 total = 0;
+
+    for (int i = 0; i < meritCatInfo[GetMeritCategory(merit)].MeritsInCat; ++i)
+    {
+        total += PMerit->count;
+        PMerit++;
+    }
+
+    return total;
+}
+
+/************************************************************************
+ *                                                                       *
  *  Добавляем персонажу limit points                                     *
  *                                                                       *
  ************************************************************************/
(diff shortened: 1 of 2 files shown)


DiscipleOfEris:
Forgot to include `merit.h` in initial commit... Well, don't forget to squash.

claywar (approved):
Seems reasonable to me, and nice catch on checking category that was overlooked in the past.

DiscipleOfEris:
I wasn't sure if LSB style guidelines would've wanted that long-ish `if` statement broken to multiple lines, but Clang doesn't seem to be complaining.

claywar:
> I wasn't sure if LSB style guidelines would've wanted that long `if` statement broken to multiple lines, but Clang doesn't seem to be complaining.

Looked at that, but there's really only two long conditions.  If you wanted to beautify, I would suggest creating a bool prior to that check, but I'll let others weigh in.

zach2good (approved):
I don't mind about the styling as long as clang-format isn't complaining, `merit.cpp` is a garbage fire anyway 👍

PR #3519, opened 2023-01-22
Fixed Ayame SC, order was reversed.

TonzoffunCE:
**_I affirm:_**

Fixed Ayame's SC calculation issue. Ayame is an OPENER, so the skillchain order should be reversed so that when it calculates the battleutil 'FormSkillchain', it uses her weaponskill as the 'resonanceProperties', and the player's weaponskill as the 'skillProperties'. 

Tested on local server and now it properly opens with the right SC.

### DIFF
--- a/src/map/ai/helpers/gambits_container.cpp
+++ b/src/map/ai/helpers/gambits_container.cpp
@@ -977,14 +977,14 @@ namespace gambits
                         for (auto& skill : tp_skills)
                         {
                             std::list<SKILLCHAIN_ELEMENT> resonanceProperties;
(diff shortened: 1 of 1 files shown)


TonzoffunCE:
In 'src/map/utils/battleutils.cpp', the function 'FormSkillchain' calculates the skillchain based off the "first" and "second" weaponskill used, hence when calculating which weaponskill the OPENER should use, it should be supplying that function with Ayame's usable weaponskills as the "first" parameter (opening weaponskill) and the player's used weaponskill as the "second" parameter (closing weaponskill). This way Ayame uses the correct weaponskill to "open" that will make the weaponskill that the player uses "close" the skillchain.

zach2good (approved):

PR #3520, opened 2023-01-23
[Proposal] Set account ID to 0 instead of deleting character. (For Login Rewrite branch)

Xaver-DaRed:
**_I affirm:_**

As discussed a few weeks ago, retail doesn't immediately delete characters.
Setting a character account ID to 0 will make that character Inaccessible but will keep the character data, allowing for a fast recovery.

What this does:
- Sets account to 0 instead of deleting char ID
- Adds `original_accid` to chars table, to keep track of who the character belonged to/if some1 is abusing the system

Create a character
Delete a character
- Check in sql that data is still there
- Check that character cannot be accessed.

### DIFF
--- a/sql/chars.sql
+++ b/sql/chars.sql
@@ -6,6 +6,7 @@ DROP TABLE IF EXISTS `chars`;
 CREATE TABLE `chars` (
   `charid` int(10) unsigned NOT NULL,
   `accid` int(10) unsigned NOT NULL,
+  `original_accid` int(10) unsigned NOT NULL DEFAULT '0',
   `charname` varchar(15) NOT NULL,
   `nation` tinyint(1) unsigned NOT NULL DEFAULT '0',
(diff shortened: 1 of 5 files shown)


sql/chars.sql
```diff
@@ -6,6 +6,7 @@ DROP TABLE IF EXISTS `chars`;
 CREATE TABLE `chars` (
   `charid` int(10) unsigned NOT NULL,
   `accid` int(10) unsigned NOT NULL,
+  `original_accid` int(10) unsigned NOT NULL DEFAULT '0',
```
TeoTwawki, line 9: Since its a char table, we'll want a migration in dbtool to add the new column

Xaver-DaRed: Added the migration script

WinterSolstice8 (approved):
approved; contingent on Teo's

TeoTwawki (approved):

PR #3521, opened 2023-01-23
[core] MOVE MESSAGE SERVER TO XI_WORLD PROCESS

zach2good:
**_I affirm:_**

- Move message server to world server
- Add stubs for future systems

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -218,6 +218,7 @@ jobs:
           xi_connect
           xi_map
           xi_search
+          xi_world
 
   Linux_GCC11_64bit:
     needs: Sanity_Checks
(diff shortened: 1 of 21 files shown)


zach2good:
What other "global" systems are there that would benefit from being rewritten and moved onto world server? (logic that happens independent of a zone)

Xaver-DaRed:
If I had to guess...
- Transports (Airships, boats, etc)
- Guild patterns
- Conquest, to a certain point. Beisigued, campaign would go here aswell. But we dont have them
- Weather

zach2good:
Party and Alliance system

zach2good:
Ready for review

Of interest:
- Changes and notes in login_auth

claywar (approved):
Seems reasonable to me.  Can't remember which, but either lgtm used to, or CodeQL may complain about lack of header guards (even though they're not needed)

PR #3523, opened 2023-01-24
[Magic] Implement Nuke Wall to elemental nukes.

Xaver-DaRed:
**_I affirm:_**

Implements "Nuke Wall" mechanic, that reduces damage dealt by nukes when done consecutively in a time window.

You will need to enable/write some prints in damage_spell.lua. Recomended the effect power and remaining duration, aswell as all the diferent function returns.

You will also need a mob. Set it to be immortal for convenience. Get it's level.

Cast repeatedly spells of the same element.

Watch damage be reduced, depending on the damage dealt and the time in between casts, for a maximum of a -40% damage for the next 5 seconds after the first cast. Each consecutive spell will reset the timer and modify the reduction.

Watch Damage needed to cap potency vary depending on mob level.
Watch diferent damage amounts under cap apply different reductions.

### DIFF
--- a/scripts/globals/effects/nuke_wall.lua
+++ b/scripts/globals/effects/nuke_wall.lua
@@ -1,0 +1,15 @@
+-----------------------------------
+-- xi.effect.NUKE_WALL
+-----------------------------------
+local effectObject = {}
+
+effectObject.onEffectGain = function(target, effect)
+end
+
+effectObject.onEffectTick = function(target, effect)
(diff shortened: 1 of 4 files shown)


TeoTwawki (approved):
seems alright for now. can be revisited later as needed. Anyone with a problem with it can open an issue and comment it out on their server.

zach2good (approved):

PR #1483, opened 2022-04-01
[core] Make raycasts check for different floors when appropriate

zach2good:
**_I affirm:_**
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] that I agree to LandSandBoat's [Limited Contributor License Agreement](https://github.com/LandSandBoat/server/blob/base/.github/CONTRIBUTOR_AGREEMENT.md), as written on this date
- [x] that I have **read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)**
- [x] that I've _**tested my code and things my code changed**_ since the last commit in the PR, and will test after any later commits

Closes: https://github.com/LandSandBoat/server/issues/1470

PR #3524, opened 2023-01-24
[los] Port Eden's line-of-sight system

zach2good:
**_I affirm:_**

- Add forked submodule of Eden's LOS meshes
- Port Eden's LOS code
- Add our-style CMake stuff for Eden's code
- Make the replacements in the code for raycasting and canSee, etc.
- Tidy some header includes while I'm in there
- run_clang_format.sh
- Add some startup logging for "Hey, I notice your navmesh or los folders are empty, just letting you know"
- Also brings in some commits with misc improvements I saw on Eden while I was poking around
- Profile (its fast! but it boosts full load time by 23s):
- image
- Test and verify locally!
- Find ways to speed up startup when single-process and loading 299 zones. It's now considerably slower.
  - Yes! parallel for_each loop saves the day!
image
- Mark zones with the new misc flags, if needed
- Eden uses this:
```sql
-- Make players unable to cast through walls in cities (1), dungeon zones (3), battlefields (4), instances (6), and limbus areas (7).
UPDATE zone_settings SET misc = misc + 0x1000 WHERE misc & 0x1000 = 0 AND zonetype IN (1, 3, 4, 6, 7);
-- Also disallow casting through walls in Dynamis cities and nyzul
UPDATE zone_settings SET misc = misc + 0x1000 WHERE misc & 0x1000 = 0 AND zoneid IN (42, 77, 185, 186, 187, 188, 294, 295, 296, 297);
```

Go get chased around by mobs, get drawn-in, cast spells etc.


zach2good:
This PR is the proper implementation of https://github.com/LandSandBoat/server/pull/1474

and probably invalidates: https://github.com/LandSandBoat/server/pull/1483

zach2good:
Would probably also be good to make sure that NMs will only draw-in to points that they can see, if that was still a problem?

tools/dbtool.py
```diff
@@ -979,14 +979,23 @@ def configure_and_launch_multi_process_by_zonetype():
...
+
 def tasks_menu():
     present_menu(
         "Maintenance Tasks",
         {
-            "1": ["Set zone IP addresses", set_external_ip_dialog],
-            "2": ["Server-wide announcement", announce_menu],
-            "3": ["Show table sizes (min 2MB)", print_db_tables_by_size],
-            # "4": [
+            "1": ["Update git submodules", update_submodules],
```
claywar, line 994: Love having this baked into dbtool as well.  For future, we should probably update the wiki to instruct people to use it on first install and updates (even if we're not updating the submodules often)

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -1149,18 +1141,30 @@ bool CMobController::Cast(uint16 targid, SpellID spellid)
 bool CMobController::CanMoveForward(float currentDistance)
 {
     TracyZoneScoped;
-    if (PMob->m_Behaviour & BEHAVIOUR_STANDBACK && currentDistance < 20)
+
+    uint16 standbackRange = 20;
```
claywar, line 1145: Out of scope for this PR, but does this ever change?

claywar (approved):

src/map/ai/controllers/mob_controller.cpp
zach2good: I set it up this way to mimic what Eden is doing in this block; they have a mobmod for this and we dont, so I left it hard-coded

TeoTwawki: we probably should follow up with adding that mobmod - I am fairly sure there are a few mobs on retail that disobey the standard rules and that's also a fun customization point too. 

Note: we could also convert the one that handles if they standback at all, and  if we don't want them to, just set it zero, but that both creates a barrier to them eventually getting back in sync with us and is potentially a surprise bug situation later in existing code.

PR #3539, opened 2023-01-27
[lua] Reduce potency of Bounty Shot

zach2good:
**_I affirm:_**

👁️ 👁️


claywar (approved):

TeoTwawki:
we actually need to redo that JA entirely. its still wrong. there is no effect ID by that name in the 1st place. bad inheritance from previous project iteration.

PR #3541, opened 2023-01-27
[core] Add plumbing for TH Proc System

zach2good:
**_I affirm:_**

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1,4 +1,4 @@
-/*
+﻿/*
 ===========================================================================
 
   Copyright (c) 2010-2015 Darkstar Dev Teams
@@ -7059,7 +7059,7 @@ namespace battleutils
         return 1.0;
     }
 
-    bool TryProcTreasureHunter(CCharEntity* PChar, CMobEntity* PMob, CAttackRound* attackRound, bool feintApplied)
+    uint8 TryProcTreasureHunter(CCharEntity* PChar, CMobEntity* PMob, CAttackRound* attackRound, bool feintApplied)
     {
         // The Treasure Hunter proc system was introduced in the December 2010 update and applies to players using Thief main job, as well as the Ranger job ability Bounty Shot.
         // https://www.bg-wiki.com/ffxi/Treasure_Hunter
@@ -7099,7 +7099,13 @@ namespace battleutils
             rate *= 1 + (PChar->getMod(Mod::AUGMENTS_FEINT) * 0.25) - 0.25;
         }
 
-        return xirand::GetRandomNumber(1, 100) <= std::clamp<uint8>(rate, 1, 100);
+        // Roll!
+        if (xirand::GetRandomNumber(1, 100) <= std::clamp<uint8>(rate, 1, 100))
+        {
+            return thCap + 1;
+        }
+
+        return thCap;
     }
 
 }; // namespace battleutils
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
 ===========================================================================
 
   Copyright (c) 2010-2015 Darkstar Dev Teams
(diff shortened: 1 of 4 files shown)


TeoTwawki:
suggested pr title amendment: `TH Proc System` so people don't think this is "the one" that moves us to the retail chart for drop rates.

src/map/utils/battleutils.cpp
```diff
@@ -7087,4 +7088,43 @@ namespace battleutils
...
+        rate *= std::pow(0.7f, diff);
+
+        // Add any flat bonus
+        rate += PChar->getMod(Mod::TREASURE_HUNTER_PROC);
+
+        // If SA or TA landed, the rate is 10x
+        rate *= attackRound->GetSATAOccured() ? 10 : 1;
+
+        // If Feint landed, the rate is doubled
+        rate *= feintApplied ? 2 : 1;
```
cocosolos: Feint should be giving a 25% increase per merit level 2-5, reaching the 2x rate at 5/5 merits.

src/map/utils/battleutils.cpp
```diff
@@ -7087,4 +7088,43 @@ namespace battleutils
...
+
+        // Treasure Hunter can be increased up to a maximum value of 12
+        // (boostable by Job Points)
+        auto thCap = 12 + PChar->getMod(Mod::TREASURE_HUNTER_CAP);
+
+        // The maximum base Treasure Hunter level that can be applied from the first action against an enemy is 8.
+        if (!mobTH)
+        {
+            thCap = 8;
+        }
```
cocosolos, line 7079: Doesn't look like this is used yet, but just to note the base cap of 8 is for non-procced hits and _can_ land a 9 on the first hit with a proc.

zach2good: I still don't fully understand the flow here, so regardless of whether you proc or not, your actions will set TH8 on an enemy? and then on top of that you have a chance to proc straight to TH9?

zach2good: This'll also need a bit of a rewrite to track current TH in a way that isn't "for every entity in my enmity list, find the highest TH amount"

TeoTwawki: > I still don't fully understand the flow here, so regardless of whether you proc or not, your actions will set TH8 on an enemy? and then on top of that you have a chance to proc straight to TH9?

pretty much even your very 1st hit has the chance to proc. your base of 8 would then pop the message for 9 on that 1st proc, while raising mob to 9. If you base is 8, you never get a message for 8 because 1st proc brings it to 9.

zach2good:
I don't know enough about THF (despite having an i119 THF on retail) to do justice to the math on this one. WTB someone who knows what they're doing to help me! The code is easy, the logic is hard.

src/map/utils/battleutils.cpp
```diff
@@ -7058,4 +7059,53 @@ namespace battleutils
...
     }
 
+    uint8 TryProcTreasureHunter(CCharEntity* PChar, CMobEntity* PMob, CAttackRound* attackRound, bool feintApplied)
+    {
+        // The Treasure Hunter proc system was introduced in the December 2010 update and applies to players using Thief main job, as well as the Ranger job ability Bounty Shot.
+        // https://www.bg-wiki.com/ffxi/Treasure_Hunter
+        // https://ffxiclopedia.fandom.com/wiki/Treasure_Hunter
+
+        auto playerTH = PChar->getMod(Mod::TREASURE_HUNTER);
+        auto mobTH    = PMob->PEnmityContainer->GetHighestTH();
```
cocosolos, line 7069: I'm pretty sure the TH level of the mob is tracked independently from enmity. Wiki says player can leave the zone/party and TH will only reset when mob goes passive (it's not clear if TH resets as soon as it goes passive or until it starts to regen health). I'm not sure how the system currently works but there is m_THLevel or something on the mob entity that I assume would only need some tweaking to work with the new system.

PR #3544, opened 2023-01-28
[core] Switch from SUNSHINE weather to NONE/CLEAR at night

zach2good:
**_I affirm:_**

TODO: Check this is actually what retail does

Put `print(zone:getWeather())` in the onInit script for Bastok Markets and boot up at Vana nighttime, it'll now print 0. It printed 1 before.

### DIFF
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -637,9 +637,12 @@ void CZone::UpdateWeather()
 {
     TracyZoneScoped;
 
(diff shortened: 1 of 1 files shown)


zach2good:
Just hopped onto retail, its WEATHER_NONE at 9am vanatime, so, WEATHER_SUNSHINE is not the default weather during the day. Our weather data is all bunk ;_;

PR #3566, opened 2023-02-05
[core] Add new async helpers

zach2good:
**_I affirm:_**


Xaver-DaRed (approved):

claywar (approved):

zach2good:
TODO: This will need some docs and some anonymous namespace stuff, because since we have `sql` exposed everywhere, and you can easily try and submit a job that has `_sql` or `sql` as an argument, you may or may not shadow or typo and use the sql connection on a different thread - ruining the lives of everyone around you.

zach2good:
Thus the conversation we had the other day of wrapping all global uses of sql and lua as `global::sql` and `global::lua`.
It's a bit wordy, but it makes everything safe

PR #3567, opened 2023-02-05
[sql+core] Move various exp table data into sql, add ENABLE_EXP_PET_PENALTY setting

zach2good:
**_I affirm:_**


zach2good:
This will need testing in the following way:

- Extract the existing implementation into its own function.
- Alter the new implementation and the old one to take in just the player level, chain time, chain number, and before exp.
- It will return the new exp.
- Run the full list of levels, chain numbers, chain times etc. through it and make sure the before/afters match up exactly.

Then we know we're good to go

zach2good:
Same with the other functions/tables too I guess

Frankie-hz:
Chain 3 for level 20 has an incorrect time value. It should be:

```sql
INSERT INTO `exp_chain_values` VALUES(20, 3, 1.30, 40000);
```

Since the table only goes up to pChainNumber 6, when the chainNumber is greater than 6 it is not defaulting to the base values as in the original switch statement.

You would have to break it up into two if statements if a match is found at 6 or before and one if its later.

This should return no errors in your Exp test harness. I did NOT test this manually in-game to see if the exp returns the same value but the values from the old and new return the same...therefore it *should* work. 

```cpp
    void ApplyExpChainBonuses(uint8 mLvl, uint16& chainNumber, uint32& chainTime, float& exp, EMobDifficulty mobCheck, bool& chainactive)
    {
        if (mobCheck <= EMobDifficulty::DecentChallenge)
        {
            return;
        }

        if (chainTime > 0 || chainTime == 0)
        {
            chainactive = true;
        }
        else
        {
            chainNumber = 1;
        }

        auto mainLevel    = mLvl;
        auto applyExpMult = 1.00f;

        // This will track if a match is found inside the table. If so, break.
        bool foundMatch = false;

        // Loop through the lookup and apply multipliers and times
        for (auto& [upperLevel, pChainNumber, expMultiplier, pChainTime] : g_ExpChainValues)
        {
            if ((pChainNumber == chainNumber || pChainNumber == 6) && mainLevel <= upperLevel)
            {
                // Save this for later
                applyExpMult = expMultiplier;

                // Overwrite chainTime value as we find new valid ones
                chainTime = 0 + pChainTime;

                // Found a match!
                foundMatch = true;

                // If we've matched, we've matched at the soonest possible time.
                // No point in iterating any more.
                break;
            }
        }

        // If no match was found and chainNumber is greater than 6, use chainNumber = 6 values
        if (!foundMatch && chainNumber > 6)
        {
            for (auto& [upperLevel, pChainNumber, expMultiplier, pChainTime] : g_ExpChainValues)
            {
                if (pChainNumber == 6 && mainLevel <= upperLevel)
                {
                    // Save this for later
                    applyExpMult = expMultiplier;

                    // Overwrite chainTime value as we find new valid ones
                    chainTime = 0 + pChainTime;
                    break;
                }
            }
        }

        // Only apply exp multiplier once
        exp *= applyExpMult;
    }
```

This code returns no warnings from thee test case.
image

I am not the greatest coder so there may be an easier way to do this. But that is the issue causing the values not to match.

zach2good:
Thanks @Frankie-hz! 👍

zach2good:
WTB Module to make these values era @Frankie-hz

zach2good:
This passed the test harness, needs actual in-game testing:

- General solo XP
- XP with a party member
- XP in a signet area
- XP in a signet area with a party member
- Some XP chains
- Testing out the setting multiplier I added

Frankie-hz:
> WTB Module to make these values era @Frankie-hz

I can probably do that next week, I’m away for the weekend. 

Might be able to sneak it in this weekend.

Frankie-hz:
I have the era-module done for the chaintime/exp - I can send it to you when I get home tonight.

zach2good:
If you feel like challenging yourself, you can try and PR it to the branch I'm currently working on: `move_exp_sql`

zach2good:
Ready for code review, still needs testing though

zach2good:
Almost a year later: @Frankie-hz, did you ever get a chance to test this a bit? I want to merge this and get it out from my pile of shame

clanofartisans:
> Almost a year later: @Frankie-hz, did you ever get a chance to test this a bit? I want to merge this and get it out from my pile of shame

Hey, Zach! We're about to start testing this in earnest very soon. Do you want the SQL stuff upgraded to the `db::` stuff before we start or is it okay to go ahead with it as-is for now and we can do a quick test afterward to confirm that nothing explodes when that's done?

zach2good:
The old sql system still works, so it's fine to test as is

clanofartisans:
From our initial testing, it looks like there may be some issues with chains. I'll try to dig into the code and see where it might be going wrong. Here is the relevant bit of conversation I have:

### Majestrix
> from testing on Beta2, Chaining is very broken. tested lvl 1, 15,20, and then at 75 for merits. chain doesn't break. waited 5minutes between mobs and chain still continues. but then sometimes when you fight ep, it'll break, and then you'll go back to DC, and the chain will continue?

### Tyler
> ill try to confirm then, see if i can legit afk for 10 minutes and keep chain
> yeah it doesn't seem to have a timer at all

zach2good:
@clanofartisans thanks for following up and testing! As I look at this again, I'm wondering why this is remaining in C++. It doesn't fulfil any of my conditions for `oh my god this needs to stay in Core, what the fuck are you doing??` - so it would be faster and easier to test, and for people to module over if it was all just in Lua?

zach2good:
Going to re-do this as a Lua global

PR #3569, opened 2023-02-06
[SQL] [LUA] Add Login Campaign Rewards: Mutsunokami, "Fancy" sets

N3ckB3ard:
**_I affirm:_**

Implements Mutsunokami and Fancy items obtained through Login Campaign Rewards. 
Fixes issue with "items\per._snow_cone.lua", the name of the usable item is related to the "Sortname" from item_basic.sql and not the Full Name.

The effect "Summers Tidings" is not clear on specifics of when the +10 elemental resistance is applied.
I gave the Fancy Leg Armor a static value of (Fire + Water)Resistance +10

All Items have been tested and the food dispensed from the body armor is "working as intended".


TeoTwawki:
> The effect "Summers Tidings" is not clear on specifics of when the +10 elemental resistance is applied.
I gave the Fancy Leg Armor a static value of (Fire + Water)Resistance +10

Just to be clear we don't normally allow substitutions like that. But in this case the item appears to actually be active all the time -or- its checking if its summertime maybe? (I didn't pull mine out in winter to look again, but it was up when I got mine) 

And nobody is going to be op or have their game ruined by 10 res. So depending on what other devs say, this might go through as is.

Xaver-DaRed:
It's probably a latent effect based on date, but for now I think its good.

I would add the raw item ID to the item.lua file
And I would also use the npcUtils.giveitem function intead of addItem, unless ints not supposed to return a message when the item is obtained

N3ckB3ard:
> It's probably a latent effect based on date, but for now I think its good.
> 
> I would add the raw item ID to the item.lua file And I would also use the npcUtils.giveitem function intead of addItem, unless ints not supposed to return a message when the item is obtained

I implemented the usage of this item from preexisting code to create similarity on event items.
https://github.com/LandSandBoat/server/blob/base/scripts/globals/items/dream_hat_%2B1.lua#L20

I do understand these are different items but I followed the logic of "Dispense:" on the actual item itself and how the code was structured within Base

TeoTwawki:
> It's probably a latent effect based on date, but for now I think its good.
> 
> I would add the raw item ID to the item.lua file And I would also use the npcUtils.giveitem function intead of addItem, unless ints not supposed to return a message when the item is obtained

additem is still supposed to produce a message indicating you gained an item. always did in the past.

scripts/globals/items/fancy_gilet.lua
```diff
@@ -0,0 +1,23 @@
...
+    local result = 0
+    if target:getFreeSlotsCount() == 0 then
+        result = xi.msg.basic.ITEM_NO_USE_INVENTORY
+    end
+
+    return result
+end
+
+itemObject.onItemUse = function(target)
+    target:addItem(6565) -- Persikos Snow Cone
```
TeoTwawki: This is the lua file Xaver referred to https://github.com/LandSandBoat/server/blob/base/scripts/globals/items.lua

I'm sure we have _many_ examples that don't do this yet, but we've started placing item IDs here and then referring to items by name elsewhere in the other Lua scripts. We use the long form name shown in the item sql where possible with a few tweaks like all caps and `_M1` for `-1` and `_p1` for `+1` items so in this case: 

`PERSIKOS_SNOW_CONE                    = 6565,`

Would be what you'd add, right after DECANTER_ARCIELA_II

N3ckB3ard: Would it be alright if I went through some of the items found in "scripts\globals\items" and update the magic numbers to ID lookups? I can make a new PR updating them

TeoTwawki: not only alright, that would be very excellent of you! <3

TeoTwawki (approved):

Xaver-DaRed (approved):

zach2good (approved):

PR #3571, opened 2023-02-06
NIN "Subtle Blow" merit fix & BST "Killer Effects" merit fix.

UmeboshiXI:
**_I affirm:_**

- Allows NIN subtle blow merits to apply to auto attacks and weapon skills.

- Allows BST killer effect merits to increase intimidation rates.

- Adds !gettp command to observe a mob's tp.

NIN Merits: 
-1. Hit a mob with 0/5 merits. Use !gettp to check the mobs TP after the hit.
-2. Go to moghouse, put points into subtle blow merits.
-3. Hit a mob with merits leveled as 75 NIN. Use !gettp to check mobs TP after the hit. Mob TP should be reduced.

BST Merits:
-1. Removed Bird Killer trait from BST Job in database. (Using Bird Killer as an example to test.)
-2. Upgrade Killer Effect merits as 75 BST.
-3. Aggro a bird, observe intimidation effect.

### DIFF
--- a/scripts/commands/gettp.lua
+++ b/scripts/commands/gettp.lua
@@ -1,0 +1,48 @@
+-----------------------------------
+-- func: gettp <player>
+-- desc: Displays target's current TP
+-----------------------------------
+
+cmdprops =
+{
+    permission = 1,
+    parameters = "is"
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!gettp (player)")
+end
+
+function onTrigger(player, tp, target)
+    -- validate target
+    local targ
+    local cursorTarget = player:getCursorTarget()
+
+    if target then
+        targ = GetPlayerByName(target)
+        if not targ then
+            error(player, string.format("Player named '%s' not found!", target))
+            return
+        end
+    elseif cursorTarget and not cursorTarget:isNPC() then
+        targ = cursorTarget
+    else
+        targ = player
+    end
+
+    -- get tp
+    if targ:isAlive() then
+        targ:getTP(tp)
+        local pet = targ:getPet()
+        if pet and pet:isAlive() then
+            pet:getTP(tp)
+        end
+
+        if targ:getID() ~= player:getID() then
+            player:PrintToPlayer(string.format("%s's TP is %i.", targ:getName(), targ:getTP()))
+        end
+    else
+        player:PrintToPlayer(string.format("%s is currently dead.", targ:getName()))
+    end
+end
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2297,6 +2297,12 @@ namespace battleutils
                 float sBlow2    = std::clamp((float)PAttacker->getMod(Mod::SUBTLE_BLOW_II), -50.0f, 50.0f);
                 float sBlowMult = ((100.0f - std::clamp(sBlow1 + sBlow2, -75.0f, 75.0f)) / 100.0f);
 
+                // apply NIN subtle blow merit values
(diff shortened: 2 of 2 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -2297,6 +2297,12 @@ namespace battleutils
                 float sBlow2    = std::clamp((float)PAttacker->getMod(Mod::SUBTLE_BLOW_II), -50.0f, 50.0f);
                 float sBlowMult = ((100.0f - std::clamp(sBlow1 + sBlow2, -75.0f, 75.0f)) / 100.0f);
 
+                // Add subtle blow mods from NIN merits
+                if (PAttacker->objtype == TYPE_PC && PAttacker->GetMJob() == JOB_NIN && PAttacker->GetMLevel() > 74)
```
WinterSolstice8, line 2301: I think this might make more sense to do in `charutils::CalculateStats` or `charutils::BuildingCharTraitsTable` where you can just add a subtle blow mod. `BuildingCharTraitsTable` already adds mods.

`GetMeritValue` already accounts for level >= 75 and job being NIN, so you could do something like..

```
        if (PChar->PMeritPoints)
        {
            PChar->addModifier(Mod::SUBTLE_BLOW, PChar->PMeritPoints->GetMeritValue(MERIT_SUBTLE_BLOW_EFFECT, PChar));
        }
 ```
Adding it deep in core means Lua won't be aware of the added subtle blow mod. In addition, this current PRed logic will go above the normal subtle blow 1 cap (is there evidence that the NIN subtle blow merit does this?)

File: src/map/utils/battleutils.cpp [hunk 2]
```diff
@@ -3595,6 +3607,12 @@ namespace battleutils
                 break;
         }
 
+        // Add intimidation rate from BST killer effect merits
+        if (PDefender->objtype == TYPE_PC && PDefender->GetMLevel() > 74 && PDefender->GetMJob() == JOB_BST)
```
WinterSolstice8, line 3611: Similar thing here with a simpler check for GetMeritValue since it does the job checking for you, but I don't know about how exactly the merit actually works.

JP wiki and BG wiki don't mention if it's _all_ killer effects or only ones you have from traits. Does this let you intimidate humanoids (players)? probably not. I think this one will need a source or verification of some kind to figure out what exactly it does.

scripts/commands/gettp.lua
```diff
@@ -0,0 +1,48 @@
...
+    else
+        targ = player
+    end
+
+    -- get tp
+    if targ:isAlive() then
+        targ:getTP(tp)
+        local pet = targ:getPet()
+        if pet and pet:isAlive() then
+            pet:getTP(tp)
```
WinterSolstice8, line 39: Dead code?

WinterSolstice8 (changes requested):

src/map/utils/battleutils.cpp
TeoTwawki: mod should be set in trait table and just having the trait should already handle adding the mod
https://github.com/LandSandBoat/server/blob/base/sql/traits.sql#L25
there's a reason the column is called "modifier" and this issue with caps is exactly it.

scripts/commands/gettp.lua
TeoTwawki: there is a lot of excess code here not really needed. As long as target isn't nil and isn't an NPC, just print the name and its current TP. No reason to care what type of entity it is like pet - other than npc, because npc's don't have stats like TP. Also doesn't matter if its dead when the command executes, that won't break the world.

`getTP()`doesn't have an input parameter btw.

src/map/utils/battleutils.cpp
TeoTwawki: bug, now known: non merit trait and merit trait won't stack because code for merit'ed traits will _delete_ non merited version of the trait before applying.

 (╯°□°)╯︵ ┻━┻

PR #3578, opened 2023-02-08
Switch to checking mainly for charset at database level

TeoTwawki:
**_I affirm:_**

Techniclaly its not the collation we need, its the charset. But collation does tip us off that someone probably has a bad charset. I've modified the function to check the database rather than mariadb's global variables, because thats really what we need to know: the step in setup (use utf8) will prevent the problem for newbies tho.

This way also prevents unneeded warnings on databases other than ours.

Deliberately set bad charset and or collation, start map server

image

### DIFF
--- a/src/common/sql.cpp
+++ b/src/common/sql.cpp
@@ -192,30 +192,31 @@ void SqlConnection::SetupKeepalive()
     m_PingInterval = timeout + reserve;
 }
 
-void SqlConnection::CheckCollation()
+void SqlConnection::CheckCharset()
 {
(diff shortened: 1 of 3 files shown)


TeoTwawki:
p.s. we probably should have dbtool check for and correct these. I didn't think of it sooner or you might be looking at a pr for that instead.

Xaver-DaRed (approved):

claywar (approved):

zach2good (approved):

PR #3580, opened 2023-02-09
[core] Replace `delete` and make it clear pointers

zach2good:
**_I affirm:_**

https://www.stroustrup.com/bs_faq2.html#delete-zero

"C++ explicitly allows an implementation of delete to zero out an lvalue
operand, and I had hoped that implementations would do that, but that
idea doesn't seem to have become popular with implementers."

So, we'll ensure this behaviour ourselves.

- Tracy still works
- The overloaded new/deletes are hit with breakpoints, so everything is working as expected

### DIFF
--- a/src/common/kernel.cpp
+++ b/src/common/kernel.cpp
@@ -55,6 +55,67 @@
 #endif
 #endif
 
+#ifdef TRACY_ENABLE
+void* operator new(std::size_t count)
+{
+    void* ptr = malloc(count);
+    TracyAlloc(ptr, count);
+    return ptr;
+}
+
+void operator delete(void* ptr) noexcept
+{
+    TracyFree(ptr);
+    free(ptr);
+    ptr = nullptr;
+}
+
+void operator delete(void* ptr, std::size_t count) noexcept
+{
+    TracyFree(ptr);
+    free(ptr);
+    ptr = nullptr;
+}
+
+void operator delete[](void* ptr) noexcept
+{
+    TracyFree(ptr);
+    free(ptr);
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):

claywar (approved):

src/common/kernel.cpp
```diff
@@ -55,6 +55,67 @@
...
+    void* ptr = malloc(count);
+    TracyAlloc(ptr, count);
+    return ptr;
+}
+
+void operator delete(void* ptr) noexcept
+{
+    TracyFree(ptr);
+    free(ptr);
+    ptr = nullptr;
```
HealsCodes, line 70: Assigning to `ptr` here does not change the actual value of pointer outside this scope so `ptr = nullptr` has no effect outside (tested on clang 14.0.0).

For a little more technical detail - the variable `ptr` as parameter is a copy of the actual pointer so any modifications to the value of it (compared to the value that it is pointing _to_) are lost once the function exits.

If you want to zero out the actual pointer you'd have to pass in a `void** ptr` and write to it via `*p = nullptr` but that's not a drop-in for the default `delete` operator syntax.

Also applies to all delete implementations above and below.

zach2good: Bah humbug, you're quite right, I'll have to go with `destroy()` in place of the actual calls to delete, and set up some CI rules to keep an eye out for calls to delete. Thanks for the heads up 👍 

```cpp
template <typename T>
inline void destroy(T*& ptr)
{
    delete ptr;
    ptr = nullptr;
}
```

PR #3585, opened 2023-02-09
[core] Replace operator delete with destroy()

zach2good:
**_I affirm:_**

- Adds a CI check to block usage of delete and delete[]
- Replaced delete and delete[] with destroy() and destroy_arr
- Comments about `delete this;`, since we can't drop-in replace those

Everything should work as it did before, but now when we delete things, they are nulled out as we want.


zach2good:
Out of draft now

claywar (approved):

Xaver-DaRed (approved):

PR #3586, opened 2023-02-09
Fixed Dancer Finishing Moves Count

TonzoffunCE:
**_I affirm:_**

This fixes the calculation for Finishing Moves to include the MOD 'xi.mod.STEP_FINISH'. This will now allow the Dancer mythic weapon 'Terpsichore' to perform accurately.

This can be verified simply by doing a '!setmod 494 1 Player' and seeing that step count rewards an extra step now.

### DIFF
--- a/scripts/globals/job_utils/dancer.lua
+++ b/scripts/globals/job_utils/dancer.lua
@@ -5,6 +5,7 @@ require('scripts/globals/jobpoints')
 require('scripts/globals/magic')
 require('scripts/globals/msg')
 require('scripts/globals/status')
+require("scripts/globals/utils")
 require('scripts/globals/weaponskills')
 -----------------------------------
 xi = xi or {}
@@ -28,6 +29,9 @@ local function getStepFinishingMovesBase(player)
     elseif player:getMainJob() == xi.job.DNC then
         numAwardedMoves = 2
     end
+
+    numAwardedMoves = numAwardedMoves + player:getMod(xi.mod.STEP_FINISH)
+    utils.clamp(numAwardedMoves, 0, 5)
 
     return numAwardedMoves
 end


scripts/globals/job_utils/dancer.lua
```diff
@@ -29,6 +30,9 @@ local function getStepFinishingMovesBase(player)
         numAwardedMoves = 2
     end
 
+    numAwardedMoves = numAwardedMoves + player:getMod(xi.mod.STEP_FINISH)
```
claywar, line 33: Seems reasonable; however, since this mod is only used by Terpsichore, can anyone confirm the behavior with it regarding presto?  Is it still clamped, or will that give 6 moves?

Xaver-DaRed: imagen
Would need to also test if the Weapon is indeed needed in the main slot or not

TonzoffunCE: > Seems reasonable; however, since this mod is only used by Terpsichore, can anyone confirm the behavior with it regarding presto? Is it still clamped, or will that give 6 moves?

https://www.bg-wiki.com/ffxi/Finishing_Move says:
"Initially, up to 5 Finishing Moves can be held at once. [Dancer](https://www.bg-wiki.com/ffxi/Dancer)'s 100 and 1200 [Job Point](https://www.bg-wiki.com/ffxi/Job_Point) gifts each raise the maximum by 2, ultimately allowing up to 9 Finishing Moves to be held."

My apologies, I thought LSB was 75-cap era for some reason :(.. I will add the JP gifts into this.

TonzoffunCE:
I'm very sorry but I'm going to close this PR because I realized after the fact there's other problems with this job_utils that I want to include in this, like Wild Flourish, Building Flourish and Contradance JAs.

WinterSolstice8:
You can draft a PR to intricate the same thing, but no worries. reopen when you feel you're ready

PR #3596, opened 2023-02-12
Prevent pets from Ticking if they're marked for deletion

JRepoArchive:
**_I affirm:_**

In `CZoneEntities` we're currently updating AI on pets which are marked for deletion. This means, that doing anything that kills the PMaster will cause a server crash.  For instance, /shutdown /logout with an Avatar out will do it.  Other instances are weird zoning timing, master dying / HPing quickly, etc.

This change specifically checks if a pet is marked for deletion (Their PMaster has been nulled out), then updates AI for any pets that are left.

Perform any action that will result in the PMaster getting cleaned up before the PPet.
PPet should no longer Tick because their status will be marked to `DISAPPEAR` and they'll get removed from zone entities before ticking again.

### DIFF
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -1434,15 +1434,11 @@ void CZoneEntities::ZoneServer(time_point tick, bool check_trigger_areas)
         //     : this way, but we need to do this to keep allies working (for now).
         if (auto* PPet = static_cast<CPetEntity*>(it->second))
         {
-            PPet->PRecastContainer->Check();
-            PPet->StatusEffectContainer->CheckEffectsExpiry(tick);
-            if (tick > m_EffectCheckTime)
-            {
-                PPet->StatusEffectContainer->TickRegen(tick);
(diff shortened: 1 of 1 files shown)


zach2good (approved):
What does this end up looking like in-game? 

I don't know what we have at the moment, but I think I've seen people talking about pets existing for a moment and then fading out when their master zones. If we already don't do that, then would this cause that behaviour?

zach2good (commented):
I'd probably also put some comments here, saying why these are done in a different order to Trusts and Mobs

JRepoArchive:
> What does this end up looking like in-game?
> 
> I don't know what we have at the moment, but I think I've seen people talking about pets existing for a moment and then fading out when their master zones. If we already don't do that, then would this cause that behaviour?

This still happens properly for Avatars on zone on the client-side. I'll prepare a short gif to show.  This specifically prevents trying to tick their AI during that process.

zach2good (approved):

PR #3597, opened 2023-02-12
Optimize Daily Tally item selection

cocosolos:
**_I affirm:_**

- Reference item vectors instead of copying them every time we select an item.
- Select some goblin junk to distribute if an owned Rare item is selected instead of rerolling.

Use the box, should work as expected.

### DIFF
--- a/scripts/globals/gobbiemysterybox.lua
+++ b/scripts/globals/gobbiemysterybox.lua
@@ -71,22 +71,14 @@ local anniversaryItems =
 }
 local gobbieJunk =
 {
-    507,   -- Goblin Mail
-    508,   -- Goblin Helm
-    510,   -- Goblin Armor
-    511,   -- Goblin Mask
-    566,   -- Goblin Cup
-    568,   -- Goblin Die
-    1239,  -- Goblin Doll
-    1868,  -- Goblin Coif Cutting
-    2542,  -- Goblin Mess Tin
-    2543,  -- Goblin Weel
-    4539,  -- Goblin Pie
-    4543,  -- Goblin Mushpot
-    9777,  -- Goblin Offering
-    15203, -- Goblin Coif
-    18180, -- Goblin Grenade
-    19220, -- Goblin Cracker
+    2542, -- Goblin Mess Tin
+    2543, -- Goblin Weel
+    4324, -- Hobgoblin Chocolate
+    4325, -- Hobgoblin Pie
+    4328, -- Hobgoblin Bread
+    4458, -- Goblin Bread
+    4495, -- Goblin Chocolate
+    4539, -- Goblin Pie
 }
 xi.mystery.onTrade = function(player, npc, trade, events)
     if trade:getItemCount() == 1 then
--- a/scripts/globals/gobbiemysterybox.lua
+++ b/scripts/globals/gobbiemysterybox.lua
@@ -71,22 +71,22 @@ local anniversaryItems =
 }
 local gobbieJunk =
 {
-    507,
-    508,
-    510,
-    511,
-    566,
-    568,
-    1239,
-    1868,
-    2542,
-    2543,
-    4539,
-    4543,
-    9777,
-    15203,
-    18180,
-    19220,
+    507,   -- Goblin Mail
+    508,   -- Goblin Helm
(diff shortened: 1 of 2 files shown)


cocosolos:
I think this stuff would be a good candidate to move to the world server. Currently distributes daily points for each map process running.

src/map/daily_system.cpp
```diff
@@ -11,49 +11,55 @@ namespace daily
...
             }
             case 6:
             {
-                dialItems = specialDialItems;
+                dialItems = &specialDialItems;
                 break;
             }
         }
-        do
+        selection = std::rand() % dialItems->size();
```
zach2good: While you're in here, this could be changed to something like `xirand::GetRandomElement(dialItems)` instead of using `std::rand()`

src/map/daily_system.cpp
```diff
@@ -11,49 +11,55 @@ namespace daily
     std::vector<uint16> sundries1DialItems;
     std::vector<uint16> sundries2DialItems;
     std::vector<uint16> specialDialItems;
+    std::vector<uint16> gobbieJunk = { 507, 508, 510, 511, 566, 568, 1239, 1868, 2542, 2543, 4539, 4543, 9777, 15203, 18180, 19220 };
```
zach2good: Could you expand this table out and comment what each of these items are?

zach2good:
> I think this stuff would be a good candidate to move to the world server. Currently distributes daily points for each map process running.

Yes absolutely @cocosolos

src/map/daily_system.cpp
cocosolos: Didn't realize this existed, will look into it!

src/map/daily_system.cpp
cocosolos: Done, also commented on the Lua side.

scripts/globals/gobbiemysterybox.lua
```diff
@@ -71,22 +71,22 @@ local anniversaryItems =
...
+    1239,  -- Goblin Doll
+    1868,  -- Goblin Coif Cutting
+    2542,  -- Goblin Mess Tin
+    2543,  -- Goblin Weel
+    4539,  -- Goblin Pie
+    4543,  -- Goblin Mushpot
+    9777,  -- Goblin Offering
+    15203, -- Goblin Coif
+    18180, -- Goblin Grenade
+    19220, -- Goblin Cracker
```
TeoTwawki: After literally 4 full stacks of 99 keys (396 trades total) that coudl only produce rare flagged stuff I already had (abjuration keys) I got:

71 goblin mess tins,
106 goblin pies
102 hobgoblin pies
48 hobgoblin chocolates
40 goblin chocolates
14 goblin bread
9 hobgoblin bread
6 goblin weel
and nothing else. not a single armor helm or mask.

Which suggestes they are weighted and makes me question if some of the other goblin items can even happen at all, or if maybe abjur keys don't pull from the same list.

TeoTwawki: I actually do this for the mess tins, they desynth to darksteel :)

cocosolos: Admittedly it's been a while since this list was compiled and I'm pretty sure I just grabbed whatever "Goblin" items that had the "Obtainable from Goblin Box" flag. It's likely some items are missing and some maybe shouldn't be there, along with possible weighting.

TeoTwawki: oh. well, every time my mules build up to many keys I can get us more data xD I keep a pile of abjurs I just pass to whichever one has too many keys to make my darksteel factory go brrrr so I'll keep tracking them for later use.

cocosolos: So I went back and looked over those items and some of them were themselves Rare flagged so I'm not sure what criteria was used when originally selecting them. I went ahead and changed the list to reflect your findings, minus any possible weighting.

CatsEyeXI:
Just found this PR. Feel free to close if its duplicative 
https://github.com/LandSandBoat/server/issues/3599

zach2good (approved):

PR #3601, opened 2023-02-14
Correct Cursed Orb cutscene to prevent player pretzles

CriticalXI:
**_I affirm:_**

- Bug fix to adjust the param passed for cursed orb cutscene to animate the player depending on their race. Can see AirSkyBoat/AirSkyBoat#1020 for what currently happens as the param is hard set for mithra animation currently.

Run `!cs` within Davoi using the differing params passed in the code to see your character either pretzle or animate correctly. Can also use different addons in Ashita or Windower to change your displayed race to test.


claywar (approved):

TeoTwawki (approved):
lmao, ty Crit

Issue #3599, opened 2023-02-13
🐛 Please move UpdateDailyTallyPoints() to world server


CatsEyeXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Win 10

## Branch affected by issue 

`base`

## Steps to reproduce

Each map server calls this function, effectively increasing each players daily tally per map server process

## Expected behavior

This function should be called once.

CatsEyeXI:
```96.236.43.24454230:[02/12/23 10:00:00:836][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
96.236.43.24454230:[02/13/23 10:00:00:830][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
96.236.43.24454330:[02/12/23 10:00:01:417][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
96.236.43.24454330:[02/13/23 10:00:01:390][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
96.236.43.24454331:[02/13/23 10:00:02:321][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
96.236.43.24454332:[02/12/23 10:00:01:376][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
96.236.43.24454332:[02/13/23 10:00:01:372][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
96.236.43.24454333:[02/12/23 10:00:02:227][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
96.236.43.24454333:[02/13/23 10:00:02:209][map][debug] Distributed daily tally points (daily::UpdateDailyTallyPoints:172)
[...]

PR #3603, opened 2023-02-15
[world] Move Gobbie Box SQL update to world server

zach2good:
**_I affirm:_**

- Tidy up headers and includes as I find them
- Add setting to enable/disable HTTP server
- Move HTTP server onto own worker so it doesn't block the rest of xi_world construction
- Move default HTTP server port to something less popular
- Move vana_time to common and update all references, since its needed in xi_world now too
- Add missing copyright headers to files I wrote in 2020 (👴)
- Add small task manager instance to xi_world
- Move UpdateDailyTallyPoints to xi_world and hook up to task system with vana_time

Anyone who wants to take this fix as a patch, you can wrap the original `UpdateDailyTallyPoints` logic in a check for "which port is this process using, if its the default port, fire the logic". That will suppress the additional calls with a single line of code.

Closes https://github.com/LandSandBoat/server/issues/3599

## Testing

- Access `http://localhost:8088/api`, get served the welcome static HTML
- New setting properly enable/disables the HTTP server
- CI passes? maybe?
- Everything builds
- Make tally update happen every tick, watch the logging spam updates, it works

To the automated PR checkbox bot: I brought you into this world, and I can take you right out again.


zach2good:
> Anyone who wants to take this fix as a patch, you can wrap the original UpdateDailyTallyPoints logic in a check for "which port is this process using, if its the default port, fire the logic". That will suppress the additional calls with a single line of code.

This is also the lazy way of making conquest only fire a single time - tie it to a single zone or process thats always going to be up: the default port, or UPPER_JEUNO, whatever you like

TeoTwawki (approved):

claywar (approved):

PR #3589, opened 2023-02-10
Fixed Dancer Job Utils

TonzoffunCE:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

Adjusted the calculation of finishing moves to account for the mythic dagger 'Terpsichore'

Fixed the abilities:
- Wild Flourish
- Building Flourish

Brought the ability Contradance into the Dancer job_utils so everything is all together (this still requires some core work to convert into an AoE Waltz effect)

References:
https://www.bg-wiki.com/ffxi/Finishing_Move
https://www.bg-wiki.com/ffxi/Terpsichore
https://www.bg-wiki.com/ffxi/Wild_Flourish
https://www.bg-wiki.com/ffxi/Building_Flourish

## Steps to test these changes

Bring these changes into server and observe how the step counts are accurate and account for dancer mythic dagger now, and the Wild Flourish/Building Flourish JAs work now.

PR #3609, opened 2023-02-17
[DNC] Move logic from 3 abilities to job utils and update their functions to match new effect method

Xaver-DaRed:
**_I affirm:_**

Imports some of the changes from PR #3589 

- Reorganizes Dancer job_utils script, so local functions, check functions and use functions are together.
- Moves logic from Building Flourish, Wild Flourish and Contradance to Dancer job_utils
- Updates logic for Flourishes check functions, since we no longer use diferent status effects to count finishing moves, and makes them all use the same one.
- Properly implements Building and Wild Flourishes.

Better reviewed by commit:

1st - No existing logic changes. Moves code arround inside the same script. Only notable change is renaming a function so it's in line with how the rest are named.
2nd - No existing logic changes. Moves code from ability scripts to job_utils script.
3rd - Updates existing logic in ability check functions, so we only check for 1 status effect and it's power.
4th - Makes all flourishes use the same check function.
5th - Corrects Building and Wild Flourish logic in both job_utils and weaponskills.lua. Corrected the value of Building Flourish merit.

### DIFF
--- a/scripts/globals/job_utils/dancer.lua
+++ b/scripts/globals/job_utils/dancer.lua
@@ -152,39 +152,20 @@ xi.job_utils.dancer.checkViolentFlourishAbility = function(player, target, abili
 end
 
 xi.job_utils.dancer.checkBuildingFlourishAbility = function(player, target, ability)
-    if
-        player:hasStatusEffect(xi.effect.FINISHING_MOVE_1) or
-        player:hasStatusEffect(xi.effect.FINISHING_MOVE_2) or
-        player:hasStatusEffect(xi.effect.FINISHING_MOVE_3) or
-        player:hasStatusEffect(xi.effect.FINISHING_MOVE_4) or
-        player:hasStatusEffect(xi.effect.FINISHING_MOVE_5)
-    then
+    if player:hasStatusEffect(xi.effect.FINISHING_MOVE_1) then
         return 0, 0
+    else
+        return xi.msg.basic.NO_FINISHINGMOVES, 0
     end
-
-    return xi.msg.basic.NO_FINISHINGMOVES, 0
 end
 
 xi.job_utils.dancer.checkWildFlourishAbility = function(player, target, ability)
     if player:getAnimation() ~= 1 then
         return xi.msg.basic.REQUIRES_COMBAT, 0
     else
-        if player:hasStatusEffect(xi.effect.FINISHING_MOVE_1) then
-            return xi.msg.basic.NO_FINISHINGMOVES, 0
-        elseif player:hasStatusEffect(xi.effect.FINISHING_MOVE_2) then
-            player:delStatusEffect(xi.effect.FINISHING_MOVE_2)
-            return 0, 0
-        elseif player:hasStatusEffect(xi.effect.FINISHING_MOVE_3) then
-            player:delStatusEffectSilent(xi.effect.FINISHING_MOVE_3)
-            player:addStatusEffect(xi.effect.FINISHING_MOVE_1, 1, 0, 7200)
-            return 0, 0
-        elseif player:hasStatusEffect(xi.effect.FINISHING_MOVE_4) then
-            player:delStatusEffectSilent(xi.effect.FINISHING_MOVE_4)
-            player:addStatusEffect(xi.effect.FINISHING_MOVE_2, 1, 0, 7200)
-            return 0, 0
-        elseif player:hasStatusEffect(xi.effect.FINISHING_MOVE_5) then
-            player:delStatusEffectSilent(xi.effect.FINISHING_MOVE_5)
-            player:addStatusEffect(xi.effect.FINISHING_MOVE_3, 1, 0, 7200)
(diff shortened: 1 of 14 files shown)


scripts/globals/job_utils/dancer.lua
```diff
@@ -334,51 +411,46 @@ xi.job_utils.dancer.useViolentFlourishAbility = function(player, target, ability
...
-xi.job_utils.dancer.checkWaltzAbility = function(player, target, ability)
-    local waltzInfo = waltzAbilities[ability:getID()]
-
-    if target:getHP() == 0 then
-        return xi.msg.basic.CANNOT_ON_THAT_TARG, 0
-    elseif player:hasStatusEffect(xi.effect.SABER_DANCE) then
-        return xi.msg.basic.UNABLE_TO_USE_JA2, 0
-    elseif player:hasStatusEffect(xi.effect.TRANCE) then
-        ability:setRecast(math.min(ability:getRecast(), 6))
+xi.job_utils.dancer.useBuildingFlourishAbility = function(player, target, ability)
```
claywar, line 408: This code is not valid with the current implementation of steps/finishing moves

Xaver-DaRed: EDIT: Corrected job_utils and weaponskills code for those 2 skills

scripts/globals/job_utils/dancer.lua
```diff
@@ -55,7 +72,7 @@ local function setFinishingMoves(player, numMoves)
     numMoves              = math.min(numMoves, getMaxFinishingMoves(player))
 
     if finishingEffect then
-        if numMoves == 0 then
+        if numMoves <= 0 then
```
claywar: We should never be purposefully going below 0 moves here, and if so, that might mask a greater problem

scripts/globals/job_utils/dancer.lua
```diff
@@ -67,7 +84,62 @@ local function setFinishingMoves(player, numMoves)
     end
 end
 
-xi.job_utils.dancer.stepAbilityCheck = function(player, target, ability)
+local function getStepAnimation(weaponSkillType)
```
claywar, line 105: Not a huge fan of these long conditions for either function, but there's some overlap so I get it.  Given its only 3 conditions that do, I'd almost prefer to table this data

scripts/globals/job_utils/dancer.lua
```diff
@@ -81,6 +153,76 @@ xi.job_utils.dancer.stepAbilityCheck = function(player, target, ability)
...
+    -- Combat Check.
+    if
+        combatOnly and
+        player:getAnimation() ~= 1
+    then
+        return xi.msg.basic.REQUIRES_COMBAT, 0
+    end
+
+    -- Finishing Move check.
+    local finishingMoves = player:getStatusEffect(xi.effect.FINISHING_MOVE_1):getPower()
```
claywar: Could this be possibly changed to numFinishingMoves just for clarity?

scripts/globals/job_utils/dancer.lua
```diff
@@ -334,64 +423,58 @@ xi.job_utils.dancer.useViolentFlourishAbility = function(player, target, ability
...
+    local power = utils.clamp(availableMoves, 0, 3)
 
-    if target:getHP() == 0 then
-        return xi.msg.basic.CANNOT_ON_THAT_TARG, 0
-    elseif player:hasStatusEffect(xi.effect.SABER_DANCE) then
-        return xi.msg.basic.UNABLE_TO_USE_JA2, 0
-    elseif player:hasStatusEffect(xi.effect.TRANCE) then
-        ability:setRecast(math.min(ability:getRecast(), 6))
+    player:addStatusEffect(xi.effect.BUILDING_FLOURISH, power, 0, 60, 0, flourishMerits)
+    setFinishingMoves(player, availableMoves - 3)
```
claywar: This is attempting to always remove 3 which should not be the case.  setFinishingMoves does not clamp the lower bound, so if we want to assume and remove more, that function will need additional safety.

(This one's important)

scripts/globals/job_utils/dancer.lua
claywar: I see now, but I don't like it.  This encourages bad behavior from other functions assuming that the set will handle it.  I would prefer additional logging happen here if it is a negative value, and still handle safety in the calling function as well.

Xaver-DaRed: I reverted this change. The solution was easier than expected.

scripts/globals/job_utils/dancer.lua
Xaver-DaRed: Tabled this functions. However, this functions came as they were from weapinskills.lua. The purpose was to take stuff out of that file only.

scripts/globals/job_utils/dancer.lua
Xaver-DaRed: Fixed

claywar (approved):

PR #3623, opened 2023-02-21
[core] xi_world party system

zach2good:
**_I affirm:_**

HEAVILY IN PROGRESS

Things to consider:
- All logic for managing a party/alliance/treasure pool "object" should exist in xi_world
- the party/alliance/treasure pool implementations on xi_map should be a thin shell that just forwards inputs to xi_world, and responds to ZMQ messages to turn into packets for the client
- We're banking on xi_world being correct and bulletproof, so we don't necessarily need to back everything into the db
- Though, if we have xi_world as a single point of access, having everything backed by the db becomes safe and sane again
- If xi_world DID go down, it comes back up very quickly, so if things were backed into the db we could re-broadcast party definition packets and there'd be no disruption to players
- Similarly, if xi_map goes down, as soon as it's back up and players start logging back in, there can be logic to rebuild the party they were in - with a timeout of course

### DIFF
--- a/src/common/mmo.h
+++ b/src/common/mmo.h
@@ -127,6 +127,8 @@ enum CHATFILTERTYPE : uint64
     // Filter level is 0-3
 };
 
+// TODO: Move all messaging types to their own file
+
 enum MSGSERVTYPE : uint8
 {
     MSG_LOGIN,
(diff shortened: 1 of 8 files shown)


zach2good:
Closing this (for now).

My plan of action (for everything to do with world server, IPC+ZMQ, and the party system) revolves around adding a proper layer of abstraction to how xi_map and xi_world communicate with each other. We ABSOLUTELY SHOULD NOT be defining these structures by hand and pushing bits and bytes into them. I'm flirting with gRPC at the moment, but it will require a lot more testing, proof-of-concept-ing, and benchmarking before we can even take a step in that direciton.

PR #3624, opened 2023-02-24
[Combat] Create global functions for fSTR and fSTR2

Xaver-DaRed:
**_I affirm:_**

Creates global functions for fSTR and for fSTR2. This is done so they can be used by any skill (upcoming WS rewrite)
This new fie will contain functions for more physical stuff,like acc calculations, critical rate, etc... which could be used by more things other than Weaponskills, like certain Mobskills, job abilities (shield bash), etc...

This, however, also incorporates new caps for STR-VIT diff which I don't see in current weaponskills.lua file.

None yet. You can call them. They will return a value so long as you feed them 2 combat-prepared entities.

### DIFF
--- a/scripts/globals/damage/physical_utilities.lua
+++ b/scripts/globals/damage/physical_utilities.lua
@@ -1,0 +1,103 @@
+-----------------------------------
+-- Global, independent functions for physical calculations.
+-- Includes:
+-- fSTR, fSTR2
+-----------------------------------
+xi = xi or {}
+xi.combat = xi.combat or {}
+xi.combat.physical = xi.combat.physical or {}
+-----------------------------------
+
+-- "fSTR" in English Wikis. "SV function" in JP wiki and Studio Gobli.
+-- BG wiki: https://www.bg-wiki.com/ffxi/FSTR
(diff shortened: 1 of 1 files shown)


scripts/globals/damage/physical_utilities.lua
```diff
@@ -0,0 +1,99 @@
...
+    -- Calculate statDiff.
+    local statDiff     = actor:getStat(xi.mod.STR) - target:getStat(xi.mod.VIT)
+    local weaponRank   = actor:getWeaponDmgRank()
+    local statLowerCap = (7 + weaponRank * 2) * -2
+    local statUpperCap = (14 + weaponRank * 2) * 2
+
+    statDiff = utils.clamp(statDiff, statLowerCap, statUpperCap)
+
+    -- Calculate fSTR based on stat difference.
+    if statDiff >= 12 then
```
claywar, line 26: Every condition here is divided by 4.  This could be done afterwards for simplicity

scripts/globals/damage/physical_utilities.lua
```diff
@@ -0,0 +1,99 @@
...
```
claywar, line 72: Same as above

WinterSolstice8:
needs source links (particularly that upper/lower cap from gobli), other than that LGTM

WinterSolstice8 (approved):

PR #3626, opened 2023-02-24
Auto Translates version 30230204_0 ids + notes.

bsips:
**_I affirm:_**

Updates auto translate tables with with keys and sane strings for client version 30230204_0

Includes auto translates for Greetings, Questions, Zone Names, Job Abilities, Expansion names, etc, 
Also includes items and key items.

Added notes on deriving values for future reference.


TeoTwawki:
looks like we have a codepage problem:
`D:\a\server\server\src\map\autotranslate.cpp(2847,19): warning C4566: character represented by universal-character-name '\u2642' cannot be represented in the current code page (1252) [D:\a\server\server\build\src\map\xi_map.vcxproj]`

for whatever reason CI isn't using the utf-8 we save these files in.

bsips:
It looks like it doesn't like ♂ etc

image

TeoTwawki:
> It looks like it doesn't like ♂ etc

I don't _think_ you have to worry about it, but I don't know how to tell our CI build to stop using wrong codepage either.

WinterSolstice8:
I think the best way to do this might be to generate a sql table that can be loaded at runtime. That is a lot like what we do for version updates with the tools anyway, and this data would likely be version specific so it will be good to align to that.

Our CI won't complain about it because it can't, the auto translates are more like data than actual strings to begin with.

zach2good:
Thanks for this, we were mostly doing this by hand before 👀 

> I think the best way to do this might be to generate a sql table that can be loaded at runtime. 

Yeah, I agree with this, then at runtime we can load all of this into an `unordered_map` and do lookups from there whenever we encounter that first AT byte. I _assume_ the code I wrote to do that lookup and replace still works...

As a start, I think just generating `autotranslate.sql` and setting things up from there will be good, but I'd like to eventually roll all of the functionality in https://github.com/bsips/AutoTParser into https://github.com/LandSandBoat/UpdateExtractor, even if it means rewriting it all in Python and adding a dependency for `ResourceExtractor`, or making `MassExtractor` do what we need

zach2good:
Since this is raw data, I'm merging this into a side branch so I can get everything rebased and cleaned up, and then merged into main

PR #3628, opened 2023-02-25
Implement Master Levels 

JRepoArchive:
**_I affirm:_**

Adds all of the subsystems and logic for implementing master levels.

__Master Levels__
[BGWiki Information for Master Levels](https://www.bg-wiki.com/ffxi/Master_Levels#Increasing_your_Master_Level)
(this information has been verified against retail)

__Packet Adjustments__

S2C 0x01B - Job Update 
Bytes 0x68:0x6B - Bit-packed display of Job Mastery Status
Bytes 0x6D - Array of Master Levels 

S2C 0x061 - Char Stats
Bytes 0x65:0x6C 
- Master Level 
- Master Breaker?
- Current Exemplar Points 
- Total Exemplar Points

S2C 0x0DF - Char Health Update
(These fields are only populated when not /anon)
0x24 - Current Master Level 
0x25 - Whether character is in possession of a master breaker KI or not

### Status

- [x] Character SQL Tables 
- [x] Loading / Saving of Character ML data 
- [x] S2C Packet Adjustments (0x0DF, 0x01B, 0x061)
- [x] Enable Master Level Menu
- [x] Enable Exemplar Point Display when you have the master breaker and are on a mastered job
- [x] GM Command `setmasterlevel` 
- [ ] GM Command `addexemplarpoints`
- [ ] Mob Exemplar Point Yield 
- [ ] Exemplar Point Distribution
- [ ] Exemplar Point Loss Mechnics
- [ ] ML Stat Bonuses (Base Stats, Skills)
- [ ] Sub job cap logic
- [ ] Sub job ability locking for specific abilities
- [ ] Search result display for MLs (Accessible via + when in /sea)

image

I am opening this PR as a draft to get early feedback on the structure of the implementation.  Don't want it to feel like a bolted on system, but a strong implementation.

PENDING


Xaver-DaRed:
Regarding the 2 new sql tables used to track both lvls and exp, we could just add new fields/columns to the existing ones

Also, I think the defaults dont need to be 0, since we already have lvl 0 for not-unlocked jobs and you need to be both lvl 99 and jov mastered in order to even begin to use the system. We can save ourselves the hassle of having to change a new value to 1 once a job is unlocked.

Just my opinion though. Don't change it until everyone else gives their opinion.

JRepoArchive:
> Regarding the 2 new sql tables used to track both lvls and exp, we could just add new fields/columns to the existing ones
> 
> Also, I think the defaults dont need to be 0, since we already have lvl 0 for not-unlocked jobs and you need to be both lvl 99 and jov mastered in order to even begin to use the system. We can save ourselves the hassle of having to change a new value to 1 once a job is unlocked.
> 
> Just my opinion though. Don't change it until everyone else gives their opinion.

The contents are fundamentally different, but also identical to EXP, so we likely do need the separation. Also, no one likes a 40+ column table. In my opinion, they do need to be zero since it is possible to be ML0.  (That's what you are when you get the key item to enable master levels).  

Semantically I wanted these tables to be 1:1 representations to what you see in the game UI without any trickery on the server side.

settings/default/main.lua
```diff
@@ -90,6 +90,7 @@ xi.settings.main =
     -- CHARACTER CONFIG
     INITIAL_LEVEL_CAP              = 50, -- The initial level cap for new players.  There seems to be a hardcap of 255.
     MAX_LEVEL                      = 99, -- Level max of the server, lowers the attainable cap by disabling Limit Break quests.
+    MAX_MASTER_LEVEL               = 50, -- Maximum master level allowed. Warning: If this is enabled, SUBJOB_RATIO is overridden. -1 for disabled, 50 (RETAIL CAP)
```
TeoTwawki, line 93: It shouldn't really override per se. Master levels add to the subjob level cap (at 5:1 I think it is, haven't been on retail in a bit) which non standard ratios would have that cap at a different number is all. 

Still, I would advise server against combinations that cause subjob to go higher than 99 or higher than mainjob, something bound to break in an unsupported way which I will without hesitation call user error!

TeoTwawki: basically think of them as "bonus levels" rather than this main lv = this exact sub lv

PR #3629, opened 2023-02-26
Fix Ul'xzomit roaming and spawn behaviors

### LINKED ISSUE #2668: 🔨 Implement more accurate Ul'xzomit roaming behavior

## Describe the feature

My FFXI Capture request was completed (except for confirmation that clusters are a slotted drop). See https://www.youtube.com/watch?v=Qa66dBOMplM

Observations:

 * The respawn on both Ul'xzomit mothers and babies is 5 minutes, not 16 minutes. Was there a patch that reduced spawn times, or was the wiki just wrong (not surprising)?
 * Mother roams continuously without pause.
 * Babies follow toward the mother on a slightly meandering path. They don't specifically try to get behind her.
 * Babies follow continuously without pause except when too close.
 * If mother gets too close (crosses back over them) the baby will stop. They will not specifically turn to face the mother.
 * The babies seem to follow the same path (with slight offset) and thus stick together pretty closely while following, unless one ends up much further than the other.
 * The slight offset the babies have from eachother seems to be a static world coordinate offset, rather than relative to mother.
 * Babies and mothers have same health (~4000-4500)
 * When the mother despawns, the babies detach and roam (separately) with a similar continuous roaming pattern to the mother.
 * When the mother respawns, the babies will return to following her no matter the distance. They don't *seem* to warp.
 * When the babies respawn, they will return to following the mother, no matter the distance. They don't *seem* to force spawn next to mother.
 * The babies can respawn whether the mother is in combat or not.
 * Babies link within what looked like roughly ~5 yalms. Only really noticeable when a baby respawns.

Assumptions:

 * I didn't ask for a test on whether the babies can respawn while the mother is dead. I'm going to assume they can, and will just reattach when the mother spawns.

DiscipleOfEris:
**_I affirm:_**

Fixes #2668 

Make baby Ul'xzomits spawn normally.

Ul'xzomit babies will now roam just like mother Ul'xzomits when the mother is dead. They will re-attach when the mother respawns, and will re-attach if the mother is alive when they spawn.

Reduce Ul'xzomit respawn time to be in-line with retail.

Fix a bug where Ul'xzomit babies could be immune to physical damage (seemed to be a spawn race condition I couldn't quite figure out).

Changed some defaults of `xi.follow.follow()` params, and added option to disable warping.

1. Go to Al'taieu.
2. Kill a baby Ul'xzomit and respawn it. Observe it automatically re-attaches.
3. Kill a mother Ul'xzomit. Observe the baby Ul'xzomits start roaming by themselves.
4. Kill a baby Ul'xzomit and respawn it. Observe it starts roaming by itself.
5. Respawn the mother Ul'xzomit. Observe the babies start following it again.
6. Fight a mother Ul'xzomit. Kill a baby Ul'xzomit and respawn it. Observe the baby only links when it gets close.
7. Kill an Ul'xzomit and let it respawn normally. Observe its respawn time is now 5.5 minutes.

## Other thoughts

~~I believe baby Ul'xzomits when following the mother have the `xi.pathFlag.RUN` flag set, despite my not setting it. My observations suggest their follow speed should be the normal roam speed. On retail, if they fall behind the mother they don't catch up unless the mother changes directions.~~

~~Baby Ul'xzomits might be linking *slightly* farther than they do in retail. It's harder to test since the babies move a little too fast.~~

~~The following should probably place a limit on its max pathfinding distance it calculates at one time. If the baby Ul'xzomits are far from the mother, the mother could have moved quite far by the time the baby gets there.~~

Ul'xzomits when roaming seem to occasionally despawn on their own (they do not on retail), perhaps by getting too far from their spawn. This is a general problem with LSB's roaming behavior, and not specific to the Ul'xzomit behavior implemented in this PR.

When a baby Ul'xzomit delatches, it sometimes moves in a straight line quite some distance to return to its "area" which isn't necessarily the same as the mother. This is a general problem in LSB's roaming behavior, and not specific to the Ul'xzomit behavior implemented in this PR.

Ul'aerns, Ul'hpemdes, and Ul'phuabos also seem to have 5.5 minute respawns on retail. ~~I haven't checked Om' or other mobs in Al'taieu, but~~ I suspect most of them have 5.5 minute respawns since Al'taieu is an outdoor zone (though you cannot mount). (See #3630)


scripts/globals/follow.lua
```diff
@@ -56,30 +57,59 @@ function addFollower(leader, follower, options)
...
+    if options.roamCooldown then
+        follower:setMobMod(xi.mobMod.ROAM_COOL, options.roamCooldown)
+    end
+
     followerOptions[followerId] = options
 end
 
-function removeFollower(leader, follower)
-    local leaderId = leader:getID()
+function cleanupFollower(follower)
```
Xaver-DaRed: Since this is a global function, can it be changed to `xi.follow.cleanupFollower`?

scripts/globals/follow.lua
```diff
@@ -56,30 +57,59 @@ function addFollower(leader, follower, options)
...
+    followerOptions[follower:getID()] = nil
+    follower:removeListener("FOLLOW_ROAM_ACTION")
+    follower:removeListener("FOLLOW_DESPAWN")
+    follower:setRoamFlags(follower:getLocalVar("[follow]roamFlags"))
+    follower:setMobMod(xi.mobMod.DONT_ROAM_HOME, follower:getLocalVar("[follow]DONT_ROAM_HOME"))
+    follower:setMobMod(xi.mobMod.ROAM_COOL, follower:getLocalVar("[follow]ROAM_COOL"))
+end
+
+function xi.follow.clearFollowers(leader)
```
Xaver-DaRed: Related to the comment above, can global functions that you edit follow this structure?

xi.follow.functionName = function(params)

scripts/globals/follow.lua
```diff
@@ -56,30 +57,59 @@ function addFollower(leader, follower, options)
...
+    if not followers or #followers == 0 then
+        return
+    end
+
+    for _, follower in ipairs(followers) do
+        xi.follow.stopFollowing(follower)
+    end
+end
+
+function xi.follow.stopFollowing(follower)
```
Xaver-DaRed: Same here

scripts/zones/AlTaieu/mobs/Ulxzomit.lua
```diff
@@ -3,75 +3,83 @@
 --  Mob: Ul'xzomit
 -----------------------------------
 require("scripts/globals/follow")
+require("scripts/globals/status")
 -----------------------------------
 local entity = {}
 
-entity.onMobSpawn = function(mob)
+function entity.onMobSpawn(mob)
```
Xaver-DaRed: We follow the deleted method to name functions. For consistency across  the project. Sorry ><

scripts/zones/AlTaieu/mobs/Ulxzomit.lua
```diff
@@ -3,75 +3,83 @@
...
-entity.onMobEngaged = function(mob, target)
-    local followers = xi.follow.getFollowers(mob)
+function entity.onMobFight(mob, target)
+end
 
-    if not followers then
-        return
-    end
+function entity.onMobEngaged(mob, target)
+end
```
Xaver-DaRed: This function is empty, so it isn't really needed. Pls remove.

Same with any other empty functions, except the main 4 basic ones (spawn, death, despawn, etc)

scripts/globals/follow.lua
DiscipleOfEris: `cleanupFollower()` is a local function. I declared it local at the beginning of the file. That said, it's only called in one place (as a result of refactoring from when I added it) so I'll just roll it into `xi.follow.stopFollowing()`

TeoTwawki: Didn't catch it previously (way back in [august last year](https://github.com/LandSandBoat/server/commit/cb3458a2553f54546b3d2f52e8f53e2fbb406722#diff-840a590675fe79f6b29c82d4bb7eed93d24c0c273a5ea5a5d397aed8690670c9R18)) but I dislike the pattern of functions beginning their lives as nil variables. Which is why nobody noticed this was local and not global.

scripts/zones/AlTaieu/mobs/Ulxzomit.lua
DiscipleOfEris: No prob. It was a change I overlooked when porting my fix from my own server.

scripts/globals/follow.lua
DiscipleOfEris: For clarity, it's a way to keep functions in local scope but have the function definition after the function definition of a function that will call the local one. Since Lua is a single-pass compiler. I'm still using it for the `onMobDespawn` and `onMobRoamAction` callbacks.

If it's an approach that LSB would prefer to avoid, let me know.

scripts/zones/AlTaieu/mobs/Ulxzomit.lua
TeoTwawki: page load move where I was typing.

scripts/globals/follow.lua
TeoTwawki: Definitely want to avoid.

DiscipleOfEris: Okay, I've re-ordered the functions, as requested.

DiscipleOfEris:
I figured out why my lack of `pathFlag.RUN` was being ignored. I never actually told the call to `pathTo()` to use the configured option... And when the argument isn't provided, `pathTo()` defaults to `RUN | WALLHACK | SCRIPTED`.

DiscipleOfEris:
While the `pathFlag` is taken care, it doesn't seem to be possible to limit the length of the `pathTo`. I could see adding a new function to the pathing AI and a new Lua binding, that prunes points beyond a max range (as opposed to `pathInRange` which sets a min range).

The other option would be to add a check to `onMobRoamAction` and just clear the current path when the follower has moved more than the max distance, so it re-paths.

The final option would be to just leave it as-is.

DiscipleOfEris:
I think I'll go with a `maxSecondsBetweenPathing` option for `follow()`, that if set will record the time of each `pathTo()` and then clear the previous `pathTo()` after that many seconds, to repath.

Edit: Decided to call it `forceRepathInterval`.

DiscipleOfEris:
@zach2good @claywar @Xaver-DaRed Ready for review.

Xaver-DaRed (approved):
Looks sane enough to me.

Kind of weird that ONLY Ul'xzomit use the follow logic currently.
For instance, there is this quadav NM in Pashow Marshlands that has 2 scorts.

Out of scope of this PR, of course.

DiscipleOfEris:
Yes, I coded this with the intent that the follow logic could be used by other mobs.

DiscipleOfEris:
@zach2good @TeoTwawki @Xaver-DaRed Is there anything blocking this being merged?

PR #3641, opened 2023-02-28
[WS] Add fTP to physical utilities and table elemental Belts and Gorgets

Xaver-DaRed:
**_I affirm:_**

- Codes fTP to physical_utilities file.
- Adds elemental Belts and Gorgets (Including Fotia) to items.lua file.
- Corrects Belts and Gorgets table order, to match their actual element.

The function isnt called yet, so it does nothing ATM.
The change in weaponskills.lua fixes belts and Gorgets giving fTP bonuses to incorrect elements.. 

Do WSs with known elemental properties while wearing an apropiate belt and/or gorget.
Observe an slight damage increase. Or better yet, add debug prints to see if fTP gets a bonus or not.


scripts/globals/combat/physical_utilities.lua
```diff
@@ -19,10 +19,58 @@
...
+require("scripts/globals/items")
+-----------------------------------
 xi = xi or {}
 xi.combat = xi.combat or {}
 xi.combat.physical = xi.combat.physical or {}
 -----------------------------------
+local wsElementalProperties =
+{
+-- [Resonance] = { None, Fire, Ice, Wind, Earth, Thunder, Water, Light, Dark },
+    [       0] = {    0,    0,   0,    0,     0,       0,     0,     0,    0 }, -- Lv 0, NONE
```
claywar, line 31: Seems like there's a lot of excessive whitespace here, I'm fine with it because of the comments, but also don't want to make this a habit of folks

claywar: Also, should these values be bool instead, or is there a reason to use 0,1?

Xaver-DaRed: It can be either.

I initially wrote it as booleans when I converted the table, but visually, for some reason, it was just easier on the eye, for me, to read the table with 0s and 1s

WinterSolstice8: There are unfortunately things that aren't .1 ftp
https://www.bg-wiki.com/ffxi/Gavialis_Helm

WinterSolstice8: (maybe I misread, 0,1 as 0.1?)

Xaver-DaRed: The bonus is added after.
This table is meant to return 0/1 (or false/true, idm) depending on the weaponskill property (fusion, liquefaction, etc...) and the element we are cheking against.
It's a "match" table.

scripts/globals/weaponskills.lua
```diff
@@ -1357,9 +1358,30 @@ function handleWSGorgetBelt(attacker)
     local accBonus = 0
 
     if attacker:getObjType() == xi.objType.PC then
-        -- TODO: Get these out of itemid checks when possible.
-        local elementalGorget = { 15495, 15498, 15500, 15497, 15496, 15499, 15501, 15502 }
-        local elementalBelt =   { 11755, 11758, 11760, 11757, 11756, 11759, 11761, 11762 }
+        local elementalGorget = -- Ordered by element correctly.
```
TeoTwawki: request: change comment to `-- Ordered by element, TDODO: mods/latents instead of items`

Even if the item ID is variablized, its still using item ids and someday it'll get changed not to as long as something reminds a future editor of that

Xaver-DaRed: Done

WinterSolstice8 (approved):

PR #3651, opened 2023-03-03
[Lua] Correct an issue preventing final chest from being in Temenos-West

CatsEyeXI:
**_I affirm:_**

Correct an issue preventing final chest from being opened.

Enter Temenos West and fight to the final floor. Open chest and profit.

### DIFF
--- a/scripts/battlefields/Temenos/temenos_western_tower.lua
+++ b/scripts/battlefields/Temenos/temenos_western_tower.lua
@@ -36,13 +36,6 @@ local setupItemCrate = function(crateID, floor)
     end)
 end
 
-local setupLootCrate = function(crateID, floor)
-    local crate = GetEntityByID(crateID)
-    xi.limbus.hideCrate(crate)
-    crate:setModelId(961)
-    crate:addListener("ON_TRIGGER", "TRIGGER_CRATE", utils.bind(content.handleOpenLootCrate, content))
-end
-
 local setupTimeCrate = function(crateID, floor)
     local crate = GetEntityByID(crateID)
     xi.limbus.hideCrate(crate)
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
This doesn't seem completely right, for some reason.

Wouldnt this be fixed by just adding this npc ID `16928830`
inside `ITEM_CRATES` table in script/zones/Temenos/IDs.lua file?

jmcmorris:
So I believe this issue was caused by a regression you can see here: https://github.com/LandSandBoat/server/commit/7cdb557b6a3ec9fe2cb0f5a5d3977a39a533c385#diff-e047edfc1f16bb6b5b73212bba36d4f7c9efbb2504ce4ad551e8cfe52040e582L529-R452https://github.com/LandSandBoat/server/commit/7cdb557b6a3ec9fe2cb0f5a5d3977a39a533c385#diff-e047edfc1f16bb6b5b73212bba36d4f7c9efbb2504ce4ad551e8cfe52040e582L529-R452

Did you try the fix with only the `[ID.TEMENOS_WESTERN_TOWER.npc.LOOT_CRATE] =` change?

scripts/battlefields/Temenos/temenos_western_tower.lua
```diff
@@ -449,7 +458,7 @@ content.loot =
         },
     },
 
-    [7] =
+    [ID.TEMENOS_WESTERN_TOWER.npc.LOOT_CRATE] =
```
Xaver-DaRed, line 452: I tested jmcmorris suggestion and it worked.

This change here is the only change needed

jmcmorris: 👍

Xaver-DaRed (approved):

PR #3655, opened 2023-03-04
[core] Move CheckTriggerAreas to its own task

### LINKED ISSUE #3648: 🔨 Separate check triger areas/regions tick interval into it's own task.
**_I affirm:_**

## Describe the feature

In retail, "regions" or event triggering zones, trigger almost immediately upon entering them.
In here, it isn't that frequent, which allows to skip over this regions with enough speed and/or if the region is small enough
(Example: Salaheem sentinels stairs region)

Creating a new task for regions would allow us to mimic that behavior by making the refresh tick more frequent, without needing to affect other things that don't really need to be checked that frequently.

zach2good:
**_I affirm:_**

- Moves CheckTriggerAreas to its own job, with a higher frequency than the main tick - meaning trigger areas are checked more often.
- I've tested it with tracy, each check round comes in at about 30us, with peaks of 100us, so, not all that bad. Could be improved, but meh 🤷 

- Add some print logging to onTriggerAreaEnter for a zone, go run into that area and see the logging.

Closes https://github.com/LandSandBoat/server/issues/3648

### DIFF
--- a/src/map/map.h
+++ b/src/map/map.h
@@ -67,6 +67,9 @@ static constexpr float server_tick_rate = 2.5f;
 // Update every 400ms
 static constexpr float server_tick_interval = 1000.0f / server_tick_rate;
 
+// Check Trigger Areas every 200ms
+static constexpr float server_trigger_area_interval = server_tick_interval / 2.0f;
+
 typedef std::map<uint64, map_session_data_t*> map_session_list_t;
 extern map_session_list_t                     map_session_list;
 
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -77,23 +77,14 @@
 int32 zone_server(time_point tick, CTaskMgr::CTask* PTask)
 {
     CZone* PZone = std::any_cast<CZone*>(PTask->m_data);
-    PZone->ZoneServer(tick, false);
+    PZone->ZoneServer(tick);
     return 0;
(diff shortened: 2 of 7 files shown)


TeoTwawki:
> Stretch Goal: Refactor trigger areas to be typesafe. ie. Don't have cube and circle logic in the same place, with the same constructors.

_donates $LSB-coin$_

Would be nice tio split'em into 2 bindings as well to cut back on people using the existing one incorrectly.

zach2good:
Tested, all seems fine, timings are as follows:
```cpp

// 2.5 updates per second
static constexpr float server_tick_rate = 2.5f;

// Update every 400ms
static constexpr float server_tick_interval = 1000.0f / server_tick_rate;

// Check Trigger Areas every 200ms
static constexpr float server_trigger_area_interval = server_tick_interval / 2.0f;
```

zach2good:
I'm reluctant to move those timings into settings, because people will mess with them while having no idea the consequences of their actions - and then they'll come crying to us asking why everything is broken

TeoTwawki:
coulda swore one of those was supposed to be 2400ms (2.4s). my brain must be trying to combine zone tick and combat tick.

Xaver-DaRed (approved):

claywar (approved):

PR #3661, opened 2023-03-08
Add melee and ranged pDIF calculations to physical utilities

Xaver-DaRed:
**_I affirm:_**

Adds pDIF calculations to physical_utilities.lua
Adds support for pDIF zones.

None. The functions are currently unused.


scripts/globals/combat/physical_utilities.lua
```diff
@@ -71,6 +95,253 @@ local elementalBelt = -- Ordered by element.
...
+    xi.zone.WALK_OF_ECHOES_P1
+}
+
+-- Meant to be called from WS master function, not from pDIF function.
+-- We do not want to be checking it PER HIT after all.
+xi.combat.physical.isLevelCorrectedZone = function(actor)
+    local actorZone = actor:getZoneID()
+    local result    = false
+
+    for _, pDifZone in pairs(pDifZoneList) do
```
zach2good: Look at how we use `set` in `rhapsodies.lua` to avoid having to look over this table all the time: https://github.com/LandSandBoat/server/blob/base/scripts/globals/rhapsodies.lua

scripts/globals/combat/physical_utilities.lua
```diff
@@ -71,6 +95,315 @@ local elementalBelt = -- Ordered by element.
     xi.items.SHADOW_BELT
 }
 
+-- List of zones in which level correction still happens.
+local pDifZoneList =
```
WinterSolstice8: Level correction also applies for magic evasion/magic accuracy, accuracy/evasion

this could be moved to a different file later

scripts/globals/combat/physical_utilities.lua
```diff
@@ -294,7 +325,168 @@ xi.combat.physical.calculateFTP = function(actor, wsTP1000, wsTP2000, wsTP3000)
...
+    baseRatio = actorAttack / targetDefense
+
+    -- Step 2: cRatio (Level correction, corrected ratio) Zone based!
+    local levelDifFactor = 0
+
+    if applyLevelCorrection then
+        levelDifFactor = (target:getMainLvl() - actor:getMainLvl()) * 0.05
+    end
+
+    local cRatio = baseRatio - levelDifFactor
```
Frankie-hz: baseRatio (Attack/Defense) should be capped prior to hitting the level correction and cRatio
image

Xaver-DaRed: I added the caps to att/def ratio. TY^^

scripts/globals/combat/physical_utilities.lua
```diff
@@ -294,7 +326,178 @@ xi.combat.physical.calculateFTP = function(actor, wsTP1000, wsTP2000, wsTP3000)
...
+    local pDifBaseCap = pDifWeaponCapTable[weaponType][1] -- TODO: Does something affect this, like with final cap?
+
+    baseRatio = utils.clamp(baseRatio, 0, pDifBaseCap)
+
+    -- Step 2: cRatio (Level correction, corrected ratio) Zone based!
+    local levelDifFactor = 0
+
+    if applyLevelCorrection then
+        levelDifFactor = (target:getMainLvl() - actor:getMainLvl()) * 0.05
+    end
```
Frankie-hz, line 369: If I understand this correctly the applyLevelCorrection is only applying to the specific zones listed in the level_correction.

https://www.bluegartr.com/threads/112776-Dev-Tracker-Findings-Posts-(NO-DISCUSSION)/page8
This says that the baseRation WAS 2.0 capped and it not virtually uncapped as of 2013 so this seems to be correct.

Notes for level correction:
+ Monsters do NOT suffer from negative level correction (Attacker < Defender)
+ Mobs, Avatars and pets only get bonuses, no penalties (if Attack > Defnder then cRatio + correction)
+ Avatars and pets cap at 2.0 ratio for anything that checks tooWeak (Need to double check this if it was ever changed - I did add a lua binding for this already so if I can find evidence if wasnt changed then easy!)
+ Total PDIF cap at the end for monsters is 4.0

scripts/globals/combat/physical_utilities.lua
```diff
@@ -294,7 +326,178 @@ xi.combat.physical.calculateFTP = function(actor, wsTP1000, wsTP2000, wsTP3000)
...
+        pDifUpperCap = 1
+    elseif wRatio < 1.2 then
+        pDifUpperCap = wRatio + 0.3
+    elseif wRatio < 1.5 then
+        pDifUpperCap = wRatio + wRatio * 0.25
+    elseif wRatio < 2.625 then
+        pDifUpperCap = wRatio + 0.375
+    else
+        pDifUpperCap = 3
+    end
```
Frankie-hz: If you want to eliminate an elseif you can utils.clamp the wRatio + 0.375 to 3

scripts/globals/combat/physical_utilities.lua
Xaver-DaRed: Correct. Since the inclussion of mobs over lvl 99, any zone that has this kind of mobs has level correction turned off, even if the target mob is not over lvl 99. In most cases, it doesnt affect lvl 75 servers, since those zones will be disabled.

Unfortunately, locus and apex mobs happen to be in legacy zones aswell, because of reasons. I'm sure very good ones too. /s

However, I have added a check for lvl 75 servers, so level correction always gets applied.

> Avatars and pets cap at 2.0 ratio for anything that checks tooWeak

Which ratio do you mean? The base ratio, or the final cap?

Xaver-DaRed: Added "not player" check for negative level correction

Frankie-hz: > 

Oh sorry! cRatio should be capped for pets after level correction is applied only if its too weak. They added this in to make sure pets dont "overkill too much" - I can find the devblog this week - I should have it bookmarked somewhere I just need to dig it out

scripts/globals/combat/physical_utilities.lua
```diff
@@ -47,6 +47,31 @@ local wsElementalProperties =
     [      16] = {    0,    0,   1,    0,     1,       0,     1,     0,    1 }, -- Lv 4, Darkness
 }
 
+-- Table with pDIF caps per weapon/skill type.
+local pDifWeaponCapTable =
```
WinterSolstice8, line 51: This table looks odd. 
"1H melee weapons have an effective ratio cap of 3.625" and yet the charts show different numbers.  I wonder if this "effective" cap is including Damage Limit bonus, and the limits should be changed to match the table `Pre-Randomizer pDIF Caps by Weapon Type`

scripts/globals/combat/physical_utilities.lua
```diff
@@ -294,7 +326,204 @@ xi.combat.physical.calculateFTP = function(actor, wsTP1000, wsTP2000, wsTP3000)
...
+    if tpIgnoresDefense then
+        ignoreDefenseFactor = tpFactor
+    end
+
+    targetDefense = math.floor(targetDefense * ignoreDefenseFactor)
+
+    baseRatio = actorAttack / targetDefense
+
+    -- Apply cap to baseRatio.
+    local pDifBaseCap = pDifWeaponCapTable[weaponType][1] -- TODO: Does something affect this, like with final cap?
```
WinterSolstice8: https://www.bg-wiki.com/ffxi/Damage_Limit%2B effects this

scripts/globals/combat/physical_utilities.lua
Xaver-DaRed: From what I could gather...
Column 1: the "effective cap" is pre-level diff and wratio
Column 2: after that, but before melee random factor and ranged distance correction.

The damage limit+ is a job trait. aka, a modifier. I dont remember the exact math, but it was dead simple...
newCap = oldCap + 0.2 * trait_level <-- something like that, from memory. Bottom line, no need to table it.

WinterSolstice8 (approved):

zach2good (approved):

PR #3664, opened 2023-03-11
[QUEST] Unbridled Passion - Add alternative CS in Xarcabard

kaincenteno:
**_I affirm:_**

* assigns destinations table to xi.teleport.destination, to avoid referring to magic numbers when checking the player's position with the destination in a script.
* adds missing cs for Unbriddled Passion quest, there is a different CS if you zone into the area by means of teleport Vahzl

set charvar 'unbridledPassion' to 3
use teleport vahzl spell and watch the cs show.

### DIFF
--- a/scripts/globals/teleports.lua
+++ b/scripts/globals/teleports.lua
@@ -151,6 +151,7 @@ local destinations =
     [ids.ESCHA_RUAUN]           = {   -0.371,  -34.279, -466.980, 192, 289 },
     [ids.MISAREAUX_CONFLUENCE]  = {  -57.385,  -21.460,  568.941, 160,  25 }
 }
+xi.teleport.destination = destinations
 
 xi.teleport.type =
 {
--- a/scripts/globals/teleports.lua
+++ b/scripts/globals/teleports.lua
@@ -87,7 +87,7 @@ xi.teleport.id = ids
 -- TELEPORT TO SINGLE DESTINATION
 -----------------------------------
 
-local destinations =
+xi.teleport.destination =
 {
     [ids.DEM]                   = {  220.000,   19.104,  300.000,   0, 108 }, -- (R)
     [ids.HOLLA]                 = {  420.000,   19.104,   20.000,   0, 102 }, -- (R)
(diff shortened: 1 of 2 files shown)


scripts/globals/teleports.lua
```diff
@@ -151,6 +151,7 @@ local destinations =
     [ids.ESCHA_RUAUN]           = {   -0.371,  -34.279, -466.980, 192, 289 },
     [ids.MISAREAUX_CONFLUENCE]  = {  -57.385,  -21.460,  568.941, 160,  25 }
 }
+xi.teleport.destination = destinations
```
zach2good: If you're going to do this, make the initial local table global and update any later calls to `destinations` to use it

scripts/zones/Xarcabard/Zone.lua
```diff
@@ -5,6 +5,7 @@ local ID = require('scripts/zones/Xarcabard/IDs')
 require('scripts/quests/i_can_hear_a_rainbow')
 require('scripts/globals/conquest')
 require('scripts/globals/keyitems')
+require("scripts/globals/teleports")
```
zach2good: Single quotes in these includes, like all the others its surrounded by

scripts/zones/Xarcabard/Zone.lua
```diff
@@ -43,7 +43,15 @@ zoneObject.onZoneIn = function(player, prevZone)
     elseif quests.rainbow.onZoneIn(player) then
         cs = 9
     elseif unbridledPassionCS == 3 then
-        cs = 4
+        if
+            math.floor(pos.x) == math.floor(xi.teleport.destination[xi.teleport.id.VAHZL][1]) and
```
zach2good: I think a better way to do this would just be a distance check with the Vahzl teleport. With double/float numbers, checking against zero is one thing, but this flooring/checking pattern doesn't sit right.

kaincenteno: yeah a friend suggested that too. It's fixed now =)

scripts/zones/Xarcabard/Zone.lua
kaincenteno: fixed =)

scripts/globals/teleports.lua
kaincenteno: done, and updated the later calls to refer to the global table :)

zach2good (approved):

PR #3559, opened 2023-02-04
[lua] [sql] Removed animated_weapon.lua mixin, all references to the mixin, move fragment drops to SQL

EpicTaru:
Co-authored-by: CatsEyeXI <beneliezer@gmail.com>

**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

Removes **animated_weapon.lua** mixin and all references to it from the .lua files of the Animated Weapon mobs.  This PR also removes the code from the .lua files of the Animated Weapon mobs that dictated which fragment each Animated Weapon drops.

PR https://github.com/LandSandBoat/server/pull/3516 (authored by CatsEyeXI) (PR closed because was combined into this PR) moves the relevant 100% chance fragment drops out of the mob .lua files and into SQL where they should belong going forward.

Reference the BG-Wiki pages for the Animated Weapons (https://www.bg-wiki.com/ffxi/Category:Animated_Weapon#Event_Appearances) to see that all list their relevant fragment as a Guaranteed drop.  FFXIDB can also be referenced to display the drop rates of the fragments but the rates may be skewed as the data may contain old school dynamis drop rate data as well as drop rates reported by those that did not kill the 15 demon NMs before killing an Animated Weapon.

History:

**animated_weapon.lua** mixin was an outdated relic of the 75 cap Dynamis era where the killing of the 15 demon NMs on the 5 towers in Dynamis-Xarcabard was required before the Animated Weapons dropped their relevant fragment. The mixin, along with the previously removed function **activateAnimatedWeapon** (removed in https://github.com/LandSandBoat/server/commit/56e5f4d70424a755968e4ed7fe99e046709f586a) are what caused the animation sub-state changes of the the Animated Weapons to allow for them to drop their fragment once all 15 demon NMs were killed.

Neo Dynamis now only requires the relevant goad item to be obtained to spawn the relevant animated weapon. Once the relevant animated weapon is defeated, the relevant fragment is a **G**uaranteed 100% drop.  

## Steps to test these changes

[...]

PR #3670, opened 2023-03-13
[sql] Fix typo in groupRate value of Animated Weapon fragment drops

EpicTaru:
**_I affirm:_**

Corrects the groupRate value of the the fragment drops of the Dynamis-Xarcabard animated weapons.

No difference would be noticed before or after pulling down this PR and killing the animated weapons.  The fragment will drop with the same frequency either way.

The groupRate value was incorrectly set as ``100`` (should be ``1000``) for the fragment drops from the animated weapons in Dynamis-Xarcabard in https://github.com/LandSandBoat/server/pull/3559.  As the dropType value is set to 0 (DROP_NORMAL), the value of groupRate has no impact on the drop rate (``@ALWAYS``) of the relevant fragment off their relevant Animated Weapon.

This PR simply corrects the groupRate value for the sake of continuity.


TeoTwawki (approved):
the good news is that for non grouped drop rows, this goes unused anyway. still want it for consistency tho.

Xaver-DaRed (approved):

PR #3674, opened 2023-03-15
[Fix] Waltz recast time underflowing.

Xaver-DaRed:
**_I affirm:_**

Fixes certain cases where a combination of Waltz's recast flat reduction, combined with Fan Dance recast reduction multiplier would result in waltzes recast time being set to ludicrous hours of wait time.

This was caused, concretely, at merit value of 5 / 5.
When getting a merit value, we get the number of merits stored multiplied by a value set in database.
This resulted in fetching a value of 25 when at 5/5, instead of a value of 5, which caused the math to make the recast timer go negative.

Max fan dance merits and equip as much waltz recast reduction gear as possible.
Use Waltzes without recast timer going to 18 hours+

### DIFF
--- a/scripts/globals/job_utils/dancer.lua
+++ b/scripts/globals/job_utils/dancer.lua
@@ -181,22 +181,27 @@ xi.job_utils.dancer.checkWaltzAbility = function(player, target, ability)
     elseif player:getTP() < waltzInfo[1] then
         return xi.msg.basic.NOT_ENOUGH_TP, 0
     else
-        -- Waltz Delay (-1s per mod value)
+        local newRecast = ability:getRecast()
+
+        -- Apply Waltz Delay Modifier (-1s per mod value)
         local recastMod = player:getMod(xi.mod.WALTZ_DELAY)
+
         if recastMod ~= 0 then
-            local newRecast = ability:getRecast() + recastMod
-            ability:setRecast(utils.clamp(newRecast, 0, newRecast))
+            newRecast = newRecast + recastMod
         end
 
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (dismissed):

WinterSolstice8 (approved):

scripts/globals/job_utils/dancer.lua
```diff
@@ -181,23 +181,28 @@ xi.job_utils.dancer.checkWaltzAbility = function(player, target, ability)
...
 
         -- Apply "Fan Dance" Waltz recast reduction.  All tiers above 1 grant 5%
         -- recast reduction each.
-        local fanDanceMerits = target:getMerit(xi.merit.FAN_DANCE)
+        local fanDanceMeritValue = target:getMerit(xi.merit.FAN_DANCE)
+
         if
             player:hasStatusEffect(xi.effect.FAN_DANCE) and
-            fanDanceMerits > 1
+            fanDanceMeritValue > 5 -- 1 merit = Value of 5.
```
claywar, line 199: This is changing the logic completely for applying merit changes by themself.  This reduction should apply for every merit past the first, at a 5% reduction

claywar: I see now.  Could you please comment the fandDanceMeritValue variable as well to understand that it is receiving a value of 5 for each upgrade?

claywar (approved):

scripts/globals/job_utils/dancer.lua
Xaver-DaRed: Done

PR #3681, opened 2023-03-18
Replace mysql python module with mariadb

cocosolos:
**_I affirm:_**

Replaces call to mysql connector with mariadb.

Run the new migration without mysql module installed.

### DIFF
--- a/tools/migrations/035_escha_portals_unlock.py
+++ b/tools/migrations/035_escha_portals_unlock.py
@@ -1,5 +1,5 @@
 import array
-import mysql.connector
+import mariadb
 
 def migration_name():
     return "Adding eschan_portals column to char_unlocks table"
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
I understand that this seems to be a good idea for us given we use MariaDB, but why did this PR come about? Was there some issue you were having?

cocosolos:
> I understand that this seems to be a good idea for us given we use MariaDB, but why did this PR come about? Was there some issue you were having?

Someone in Discord was having an issue installing the server since MySql connector is no longer a requirement and dbtool fails to run when this migration tries to import it.

TeoTwawki (approved):

WinterSolstice8 (approved):
LGTM

zach2good (approved):

Xaver-DaRed (approved):
Ooops

PR #3696, opened 2023-03-26
[NPC] Ada - Abyssea Tahrongi Default Action

kaincenteno:
**_I affirm:_**

Adds the default interaction when speaking with Ada

zone to Abyssea Tahrongi canyon
find npc
talk to her


scripts/zones/Abyssea-Tahrongi/npcs/Ada.lua
```diff
@@ -0,0 +1,19 @@
+-----------------------------------
+-- Zone: Abyssea - Tahrongi
+--  NPC: Ada
```
zach2good: Does the default interaction need the empty script to work? This should also have the pos command to get to this NPC

kaincenteno: Didn't test it without it, just added it cause I looked through npcs that also had empty scripts also using default interaction so thought it was the norm. I'll go ahead and test and remove if works without it then.

kaincenteno: tested, working without script.

zach2good (approved):

PR #3714, opened 2023-03-31
Rewrites getAvailableTrickAttackChar to remove expensive trig operations

GravenImageRD:
**_I affirm:_**

Rewrites getAvailableTrickAttackChar to remove expensive trig operations by converting to Diamond Angles and comparing angles between potential partners and the TA user. Additionally makes sure to pick the viable TA partner closest to mob instead of the first viable TA partner found. Currently the value is set to about +/-2 degrees which from my testing felt like the "must be in a straight line" expectation from retail and called out on wikis.

1. Find a mob
2. Summon a Trust
3. Fight mob
4. Put the Trust between you and the mob
5. Use Trick Attack

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -4225,19 +4225,14 @@ namespace battleutils
     }
 
     /*
-     * Find if any party members are in position for trick attack.  Do this by making a narrow triangle:
-     *  one endpoint at the mob, and the other two endpoints being equidistant from the TA user, perpendicular to
-     *  the line between the mob and the TA user.  Find the slope of the line between the TA user and the mob, and
-     *  decide whether to use x or z as the dependent variable (to avoid big numbers and divide by 0 errors on
-     *  vertical slopes).  Using this slope, we can find the angle of the perpendicular line to the x or z line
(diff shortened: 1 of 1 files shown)


zach2good:
> expensive trig operations

Can you provide before/after profiles of this, please?

EDIT: This should still be up to date: https://github.com/LandSandBoat/server/wiki/Performance-Profiling-with-Tracy

zach2good:
Implementation looks good though 👍

GravenImageRD:
Glad you had me check perf more closely. While calculating a single diamond angle is faster than the atans + cos + sin being used in the current implementation, switching to angle comparisons potentially requiring multiple diamond angle calculations ended up costing more than the current slope checking method in the worst case (several possible TA targets closer to the mob than the user). I also vastly underestimated how much faster trig operations have gotten since the last time I used them in C++, which was probably ten years ago now. 😆 Not too proud to admit when I've been bested, and it was nice digging into some math heavy coding after such a long time. Will close this PR since the crux of it is unnecessary, and will put up a simpler PR that just adds the logic to make sure TA is picking the valid partner closest to the mob instead of the first one found.

PR #3716, opened 2023-04-01
Patch minor exploit opened by module.

TeoTwawki:
Also change message type to warning to match other exploit checks.

**_I affirm:_**

prevents using addons to bypass restrictions on what zone the AH menu can open in.

1. Enable the pagination module
2. on a non-gm character, load ahgo in ashita
3. `/ah` and buy something. it should hang and eventually die. before this change it would succeed.

### DIFF
--- a/modules/custom/cpp/ah_pagination.cpp
+++ b/modules/custom/cpp/ah_pagination.cpp
@@ -35,6 +35,12 @@ class AHPaginationModule : public CPPModule
         {
             TracyZoneScoped;
 
+            if (PChar->m_GMlevel == 0 && !PChar->loc.zone->CanUseMisc(MISC_AH))
+            {
+                ShowWarning("%s is trying to use the auction house in a disallowed zone [%s]", PChar->GetName(), PChar->loc.zone->GetName());
+                return;
+            }
+
             // Only intercept for action 0x05: Open List Of Sales / Wait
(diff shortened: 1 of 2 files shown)


WinterSolstice8 (approved):

modules/custom/cpp/ah_pagination.cpp
```diff
@@ -35,6 +35,12 @@ class AHPaginationModule : public CPPModule
         {
             TracyZoneScoped;
 
+            if (PChar->m_GMlevel == 0 && !PChar->loc.zone->CanUseMisc(MISC_AH))
```
ghost, line 38: Could be a good idea to do a distance check to AH counters to verify the player isn't accessing the AH from across the zone.

TeoTwawki: been thinking on how we'd approach that. in the packet system here we don't have the npc object to check against, in lua land we ***could*** set trigger areas around them in a manner similar to what is done for the "Flyers for Regine" quest but that's beyond the scope of what I intend for the pr, which is to make it the same regardless of modules.

I think I'd like to see triggerAreas improved on 1st. Right now they require preset numbered entries. I'd like to be able to dynamically create and destroy them and use non numbered index entry, then we could have the npc create them at spawn time, instead of in zone init, and without fear of index collisions

PR #3728, opened 2023-04-03
[Quest] Converts DRG flag and AF1-3 to Interaction Framework

hooksta4:
**_I affirm:_**

Converts the following quests to IF.
The Holy Crest - https://ffxiclopedia.fandom.com/wiki/The_Holy_Crest - https://www.youtube.com/watch?v=J3FI4IK3ZU4
A Craftmans Work - https://ffxiclopedia.fandom.com/wiki/A_Craftsman%27s_Work - https://www.youtube.com/watch?v=4zPw_9mV9d8
Chasing Quotas - https://ffxiclopedia.fandom.com/wiki/Chasing_Quotas - https://www.youtube.com/watch?v=wDdGUSfKW4M
Knight Stalker - https://ffxiclopedia.fandom.com/wiki/Knight_Stalker - https://www.youtube.com/watch?v=vUPuL-Bs6vc

Removes individual lines pertaining to quests from NPC's. 
Added proper textID's based on retail captures
Moved default responses to appropriate defaultActions files

Removed useless vars checks that won't apply using framework. (theHolyCrestCheck) as an example. 

The associated quests such as Trouble at the sluice are already converted on ASB - will need to cherry pick those, if desired. 

Steps are located in the comments in each of the quest lua's. 

Individuals who are currently running a live server will need to run the following to convert old variables to the interaction framework. However, use with caution I assume no liability for errors or countless hours of your GM's trying to fix thing. However, it is very easy to following along with using the framework. 

UPDATE char_vars SET varname = 'Quest[0][93]Prog' WHERE varname = "TheHolyCrest_Event";

UPDATE char_vars SET varname = 'Quest[0][94]csSeen WHERE varname = "has_seen_drgaf1_quest_already";
UPDATE char_vars SET varname = 'Quest[0][94]Prog WHERE varname = "aCraftsmanWork";
UPDATE char_vars SET varname = 'Quest[0][94]nmKilled WHERE varname = "Decurio_I_IIIKilled";

UPDATE char_vars SET varname = 'Quest[0][95]csSeen WHERE varname = "ChasingQuotas_No";
UPDATE char_vars SET varname = 'Quest[0][95]Timer WHERE varname = "ChasingQuotas_date";
UPDATE char_vars SET varname = 'Quest[0][95]nmKilled WHERE varname = "SturmtigerKilled";
UPDATE char_vars SET varname = 'Quest[0][95]Prog WHERE varname = "ChasingQuotas_Progress";

UPDATE char_vars SET varname = 'Quest[0][96]Prog WHERE varname = "KnightStalker_Progress";
UPDATE char_vars SET varname = 'Quest[0][96]Timer WHERE varname = "ChasingQuotas_Progress";
UPDATE char_vars SET varname = 'Quest[0][96]nmKilled WHERE varname = "KnightStalker_Kill";


scripts/quests/sandoria/DRG_AF1_A_Craftmans_Work.lua
```diff
@@ -0,0 +1,133 @@
...
+                    end
+                end,
+            },
+
+            onEventFinish =
+            {
+                [73] = function(player, csid, option, npc)
+                    if option == 1 then
+                        quest:begin(player)
+                        quest:setVar(player, 'Prog', 1)
```
Xaver-DaRed, line 55: Overall, this quest structure looks good.
However, I don't think you need to set the prog variable to 1 in here, since this is it's own section.
You could get away with just setting it to 1 when killing the nm

Less variable handling = less chances to screw up, if you know what I mean

scripts/quests/sandoria/DRG_AF1_A_Craftmans_Work.lua
```diff
@@ -0,0 +1,133 @@
...
+        end,
+
+        [xi.zone.NORTHERN_SAN_DORIA] =
+        {
+            ['Miaux'] =
+            {
+                onTrigger = function(player, npc)
+                    if player:getVar("Quest[0][94]csSeen") == 0 then
+                        return quest:progressEvent(73)
+                    elseif player:getVar("Quest[0][94]csSeen") == 1 then --Declined First Offering.
```
Xaver-DaRed: I think you could just use `quest:getVar(player, 'csSeen')` here since we are checking for a varable for this very quest.

Actually, I dont think `player:getVar()` would work, it would be `player:getCharVar()`

scripts/quests/sandoria/DRG_AF1_A_Craftmans_Work.lua
```diff
@@ -0,0 +1,133 @@
...
+                    end
+                end,
+            },
+
+            ['Decurio_I-III'] =
+            {
+                onMobDeath = function(mob, player, optParams)
+                    if quest:getVar(player, 'Prog') == 1 then
+                        quest:setVar(player, 'nmKilled', 1)
+                    end
```
Xaver-DaRed, line 126: So, we are using 3 variables in this quest. We could get away with just using 1 to track the whole progression.

We have to think that players leave quests unfinished. Long term, the less variables used = less variables left living in the database.

So it's more of a "good practice" thing, than a matter of it being wrong.

scripts/quests/sandoria/DRG_AF2_Chasing_Quotas.lua
```diff
@@ -0,0 +1,272 @@
...
+                        quest:setVar(player, 'csSeen', 1) -- Declined First Offering.
+                    end
+                end,
+
+                [18] = function(player, csid, option, npc)
+                    if option == 1 then -- Accept quest.
+                        quest:begin(player)
+                        quest:setVar(player, 'csSeen', 0)
+                    else
+                        quest:setVar(player, 'csSeen', 1)
```
Xaver-DaRed, line 86: Same things I said about the first quest apply in this one too.

- `quest:getVar(player, 'csSeen')` instead of `player:getVar("Quest[0][95]csSeen")`
- In this case, I see a timer variable, so we cant get away with just 1 progression variable. Probably

scripts/quests/sandoria/DRG_AF2_Chasing_Quotas.lua
```diff
@@ -0,0 +1,272 @@
...
+
+        [xi.zone.PORT_SAN_DORIA] =
+        {
+            ['Ceraulian'] =
+            {
+                onTrigger = function(player, npc)
+                    local questProgress = quest:getVar(player, 'Prog')
+
+                    if questProgress == 1 then
+                        if player:getCharVar("Timer") > os.time() then
```
Xaver-DaRed: This variable is set as a quest variable, so it WILL have a quest ID identifier in it's name.

So, basicaly, we are not checking for the correct variable. Would be fixed easily via `quest:getVar(player, 'Timer')`

scripts/quests/sandoria/DRG_AF2_Chasing_Quotas.lua
```diff
@@ -0,0 +1,272 @@
...
+                            return quest:progressEvent(7)
+                        end
+                    elseif questProgress == 2 then
+                        return quest:progressEvent(8)
+                    elseif questProgress == 3 then
+                        return quest:progressEvent(6)
+                    elseif questProgress == 4 or questProgress == 5 then
+                        return quest:progressEvent(9)
+                    elseif questProgress == 6 then
+                        return quest:progressEvent(15)
```
Xaver-DaRed, line 119: Another thing to note:

We reguarly reserve `:progressEvent` for the important events that actually progress the quest and use `:event` for the temporal dialog changes that don't actually do anything/aren't important.

The reson for this is that this npc may, or may not, be involved in other quests/missions.

`:progressEvent` will ALWAYS take priority over `:event` meaning we will never "block" any other quest or mission progression. Not really retail accurate, since retail DOES block progression in certain cases, but this falls into "be better than retail, let's not randomly make players think they did something wrong when they didnt"

scripts/quests/sandoria/DRG_AF1_A_Craftmans_Work.lua
KnowOne134: xi.quest.getVar(ares, quest, 'name')

scripts/quests/sandoria/DRG_AF1_A_Craftmans_Work.lua
KnowOne134: Like changing it to use setBit()

scripts/quests/sandoria/DRG_AF2_Chasing_Quotas.lua
KnowOne134: If it needs an override because it's important van use event():importantOnce()

scripts/quests/sandoria/DRG_AF1_A_Craftmans_Work.lua
hooksta4: the rationale for keeping the vars where they are at is because the previous writers had them setup that way. While I understand it may not be primary concern, those that have a player base and pull this will have a hell of a time comparing vars to where they are. With the above sql commands, those server owners can change the varnames and the progress is left the same.

scripts/quests/sandoria/DRG_AF1_A_Craftmans_Work.lua
hooksta4: good point - i was under the impression when I started this that it wouldnt call them but made the updates.

scripts/quests/sandoria/DRG_AF1_A_Craftmans_Work.lua
hooksta4: Understood - i went back and forth on trimming the fat but again, opted to go with what previous vars were set at on the conversion instead of a complete rewrite.

scripts/quests/sandoria/DRG_AF2_Chasing_Quotas.lua
hooksta4: Indeed, i usually mix up a couple - if it has an eventFinish then its a progressEvent otherwise, tis event.

Thanks for the feedback - made the appropriate changes.

PR #3729, opened 2023-04-04
[lua] Aern mixin fixes

WinterSolstice8:
**_I affirm:_**

Adjusts luautils.cpp GetEntityByID to have an optional silenceWarning param in case a failure can be expected.
Change Aern mix to use IDs instead of storing CLuaBaseEntity references to target player/pet after reraise. This resolves a crash.

Go to Sea, summon Ifrit, assault a aern, use Flaming Crush to kill an aern, immediately release when it's hp is 0 and do not crash.
Crash is observed on Linux, at least.

### DIFF
--- a/src/map/lua/luautils.h
+++ b/src/map/lua/luautils.h
@@ -137,7 +137,7 @@ namespace luautils
     auto GetZone(uint16 zoneId) -> std::optional<CLuaZone>;
     auto GetNPCByID(uint32 npcid, sol::object const& instanceObj) -> std::optional<CLuaBaseEntity>;
     auto GetMobByID(uint32 mobid, sol::object const& instanceObj) -> std::optional<CLuaBaseEntity>;
-    auto GetEntityByID(uint32 mobid, sol::object const& instanceObj) -> std::optional<CLuaBaseEntity>;
+    auto GetEntityByID(uint32 mobid, sol::object const& instanceObj, sol::object const& arg3) -> std::optional<CLuaBaseEntity>;
 
     void  WeekUpdateConquest(sol::variadic_args va);
     uint8 GetRegionOwner(uint8 type);
--- a/scripts/mixins/families/aern.lua
+++ b/scripts/mixins/families/aern.lua
@@ -31,17 +31,33 @@ g_mixins.families.aern = function(aernMob)
             mob:setMobMod(xi.mobMod.NO_DROPS, 1)
         end
 
-        local target = mob:getTarget()
+        local target   = mob:getTarget()
+        local targetID = 0
+        local masterID = 0
 
-        if
-            target and
-            target:isPet() and
-            not target:isAlive()
-        then
-            target = target:getMaster()
+        if target then
+            targetID = target:getID()
+        end
+
+        if target:isPet() and target:getMaster() then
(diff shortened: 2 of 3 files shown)


The-Aerec (approved):

src/map/lua/luautils.h
```diff
@@ -137,7 +137,7 @@ namespace luautils
     auto GetZone(uint16 zoneId) -> std::optional<CLuaZone>;
     auto GetNPCByID(uint32 npcid, sol::object const& instanceObj) -> std::optional<CLuaBaseEntity>;
     auto GetMobByID(uint32 mobid, sol::object const& instanceObj) -> std::optional<CLuaBaseEntity>;
-    auto GetEntityByID(uint32 mobid, sol::object const& instanceObj) -> std::optional<CLuaBaseEntity>;
+    auto GetEntityByID(uint32 mobid, sol::object const& instanceObj, bool silenceWarning) -> std::optional<CLuaBaseEntity>;
```
zach2good: is this bool optional? I don't know what sol does here if you request a concrete type instead of a sol::object (which may be nil if you don't provide it), does it default to false?

zach2good: Threw this together very fast and I may have omitted some details, but it looks like a crash?
https://godbolt.org/z/s59b1d4YT

TeoTwawki: f it gets defined with a default value instead, it'll go through safely, I think.
`auto GetEntityByID(uint32 mobid, sol::object const& instanceObj, bool silenceWarning = false)`

At least I hope so since we do that in a few places already.

zach2good: Unfortunately not (in this cut down example):
```cpp

auto GetEntityByID(bool silenceWarning = false)
{
    std::cout << "GetEntityByID" << " " << silenceWarning;
    return true;
}
```

```
[sol2] An error occurred and has been passed to an error handler: sol: runtime error: stack index 1, expected boolean, received no value: 
stack traceback:
	[C]: in function 'base.GetEntityByID'
	[string "GetEntityByID()"]:1: in main chunk
terminate called after throwing an instance of 'sol::error'
  what():  sol: runtime error: stack index 1, expected boolean, received no value: 
stack traceback:
	[C]: in function 'base.GetEntityByID'
	[string "GetEntityByID()"]:1: in main chunk
```

zach2good: It'll need to be a sol::object which is checked for being non null, and then converted into a bool

TeoTwawki: The = part needs to be at the header side, not the cpp side. based on existing examples.

zach2good: argument defaults don't care where they're declared, as long as they're not declared twice (redefinition), but it seems like sol doesnt handle accepting nil/no args as the listed types default: https://godbolt.org/z/Wf38jMjxG

TeoTwawki: Weird that VS will not even compile it for me on the cpp side then, but happily builds when done in the header only.

WinterSolstice8: Updated, optional arg seems to function as wanted now.

TeoTwawki (approved):

PR #3732, opened 2023-04-05
[Lua] Full Circle returns no MP because local var being set on wrong entity

CatsEyeXI:
**_I affirm:_**

As title states, GEO ability "Full Circle" returns no MP because the LocalVar is being set on the Luopan instead of the player.

Currently full circle returns no MP, after this change it returns the correct amount of MP.

### DIFF
--- a/scripts/globals/job_utils/geomancer.lua
+++ b/scripts/globals/job_utils/geomancer.lua
@@ -247,7 +247,7 @@ end
 
 xi.job_utils.geomancer.fullCircle = function(player, target, ability)
     local hppRemaining = target:getHPP()
-    local mpCost       = target:getLocalVar("MP_COST")
+    local mpCost       = player:getLocalVar("MP_COST")
     local fcMerit      = player:getMerit(xi.merit.FULL_CIRCLE_EFFECT)
     local crMerit      = player:getMerit(xi.merit.CURATIVE_RECANTATION)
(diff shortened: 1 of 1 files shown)


scripts/globals/job_utils/geomancer.lua
```diff
@@ -425,7 +425,7 @@ xi.job_utils.geomancer.spawnLuopan = function(player, target, spell)
     xi.job_utils.geomancer.addAura(luopan, 0, effect, finalPotency, targetType)
 
     -- Save the mp cost for use with Full Circle on the luopan
-    luopan:setLocalVar("MP_COST", spell:getMPCost())
```
TeoTwawki, line 428:
```lua
xi.job_utils.geomancer.fullCircle = function(player, target, ability)
    local hppRemaining = target:getHPP()
    local mpCost       = target:getLocalVar("MP_COST")
```
seems the author believed the _target_ of the ability would be the loupan, not the player. could you change that line to say _player_ as well? just for clarity

CatsEyeXI: all set, sir

TeoTwawki: ty

TeoTwawki (approved):

claywar (approved):

PR #3736, opened 2023-04-07
[FIX] Boiling blood

Chronos77:
When mobs use mob skill Boiling Blood, attack and defense increased excessively.

WhatsApp Image 2023-04-07 à 13 22 36

**_I affirm:_**

In this merge request, I correct the power parameter sent in the  mobBuffMove method to 25 for haste and 50 berserk
instead of 2500 and 5000

Give TP to Mesa wivre in Abyssea Konstchat until using the mob skill Boiling blood.

I check these stats with the command !getstats

Result for me: 

image

### DIFF
--- a/scripts/globals/mobskills/boiling_blood.lua
+++ b/scripts/globals/mobskills/boiling_blood.lua
@@ -17,7 +17,7 @@ end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     xi.mobskills.mobBuffMove(mob, xi.effect.HASTE, 2500, 0, 60)
-    xi.mobskills.mobBuffMove(mob, xi.effect.BERSERK, 5000, 0, 60)
+    xi.mobskills.mobBuffMove(mob, xi.effect.BERSERK, 50, 0, 60)
     skill:setMsg(xi.msg.basic.NONE)
     return 0
 end


scripts/globals/mobskills/boiling_blood.lua
```diff
@@ -16,9 +16,8 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    xi.mobskills.mobBuffMove(mob, xi.effect.HASTE, 2500, 0, 60)
-    xi.mobskills.mobBuffMove(mob, xi.effect.BERSERK, 5000, 0, 60)
-    skill:setMsg(xi.msg.basic.NONE)
+    xi.mobskills.mobBuffMove(mob, xi.effect.HASTE, 25, 0, 60)
```
TeoTwawki: this is incorrect. Hast is not scaled that way in the project at present. And you've lost the line break. You also don't clearly state what you are fixing in your commit message https://github.com/LandSandBoat/server/pull/3736/commits/0dbe6eb7bf49ec5d0a38c25d097103347a94285e or in your pr title. I presume the problem was gaining a MASSIVE amount of attack.

TeoTwawki (changes requested):

scripts/globals/mobskills/boiling_blood.lua
WinterSolstice8: looks like based on the script it would've added 5000%~ attack

TeoTwawki: yup. but the hast is 25% when using 2500 of the modifier. and if mobBuffMove does math on its value ima kick someone in the shins. Hard.

TeoTwawki: at min need the line break put back and commit message amended regardless.

Chronos77: I rewrote the merge request and commit text. line break returned!

TeoTwawki: https://github.com/LandSandBoat/server/blob/a5540475df64b3f70231ace3190c78f5130c0f4f/src/map/modifier.h#L227

TeoTwawki (approved):

PR #3738, opened 2023-04-08
Retail accurate Palborough Mines elevator

WinterSolstice8:
**_I affirm:_**

This adds some functionality and changes some scripts to accurately represent the elevators & levers in Palborough Mines.

Use a lever for the Palborough Mines elevator, get an event with a Y/N dialog, click no and it does nothing.
Click the lever and say yes, it waits about 3 seconds and then moves. During the time it is moving or waiting you should not be able to trigger the dialogue.
If another player has this dialogue open it should do nothing even during the inactivity window.

The levers should only move if the elevator is going to activate. The states are independent of eachother.


TeoTwawki (approved):

scripts/zones/Palborough_Mines/npcs/_3z7.lua
```diff
@@ -6,17 +6,48 @@ require("scripts/globals/status")
...
 local entity = {}
 
+entity.isBusy = false
+
 entity.onTrade = function(player, npc, trade)
 end
 
 entity.onTrigger = function(player, npc)
-    RunElevator(xi.elevator.PALBOROUGH_MINES_LIFT)
+    local state = player:getElevatorState(xi.elevator.PALBOROUGH_MINES_LIFT)
```
zach2good: If the binding doesn't rely on the player or the npc in this interaction, should it instead be part of lua_zone and then you call through npc:getZone():getElevetorStuff(id)?

WinterSolstice8: `RunElevator` lives in luautils so I put it there, how's this?

zach2good (approved):

claywar (approved):

scripts/zones/Palborough_Mines/npcs/_3z7.lua
TeoTwawki: if I were to (hypothetical) redo the entire system from scratch, all of it would live in baseentity. every single part of a elevator is an npc. down to the moving gears, invisible doors, and platform itself - all npc's and their animation states. the sole contribution the zone makes is to tell them when to start their repeating task for the automatic ones. Pull-lever ones like this rely on a player interaction instead.

PR #3749, opened 2023-04-10
[core] Add logic to despawn hidden NPCs if status is set to disappear

WinterSolstice8:
**_I affirm:_**

Fixes issues with QMs not disappearing, such as Scrawled Writing when trading a Goblin Drink in oldton movalpolos

Trade a Goblin Drink to the Scrawled Writing in Oldton Movalpolos, the goblin wolfman should spawn, the QM should disappear.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -2475,7 +2475,7 @@ void CLuaBaseEntity::updateToEntireZone(uint8 statusID, uint8 animation, sol::ob
         PNpc->name[8]                  = 8;
     }
 
-    PNpc->loc.zone->UpdateEntityPacket(PNpc, ENTITY_UPDATE, UPDATE_COMBAT);
+    PNpc->loc.zone->UpdateEntityPacket(PNpc, ENTITY_UPDATE, UPDATE_COMBAT, true);
 }
 
 /************************************************************************
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -617,6 +617,12 @@ void CZoneEntities::SpawnNPCs(CCharEntity* PChar)
                     PChar->updateEntityPacket(PCurrentNpc, ENTITY_DESPAWN, UPDATE_NONE);
                 }
             }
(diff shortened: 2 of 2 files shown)


zach2good:
I don't doubt that this works, my question is why did this issue appear some months ago, and why is the fix in this specific area and not modifying something that was added recently 🤔 

Spidey senses tingling

TracentEden:
> I don't doubt that this works, my question is why did this issue appear some months ago, and why is the fix in this specific area and not modifying something that was added recently 🤔
> 
> Spidey senses tingling

Yeah, the issue was introduced in this [commit](https://github.com/AirSkyBoat/AirSkyBoat/pull/2837/commits/62be9ddabe1d5289c4baed2696c548088d8f89ea) with the `if (loc.zone && updatemask && status != STATUS_TYPE::DISAPPEAR && now > m_nextUpdateTimer)` condition because  `setStatus(xi.status.DISAPPEAR)` just sets the status and mask and relies on `PostTick` to send the `UpdateEntityPacket` which is no longer does because the status is `xi.status.DISAPPEAR`.

I think this method of fixing (rather than removing the condition) better aligns with how similar logic works for chars and other types of non-NPC entities.

WinterSolstice8:
I did some more digging. This PR combined with the previous commit makes NPCs behave like all other entity types (see `PMobEntity::PostTick()`) I will still want to do more testing on far away NPCs (such as bastok bridges, etc) to see if they still function.

WinterSolstice8:
It looks like the bridges don't work unless you are close anyway -- we need another PR to fix that.

WinterSolstice8:
Ready for review, packed in with a fix for the bridges. It had nothing to do with this PR or why this PR came about. simply a missing `true` flag somewhere.

TeoTwawki:
> Fixes issues with QMs not disappearing, such as Scrawled Writing when trading a Goblin Drink in oldton movalpolos

also the FoV caskets not despawning when you fail to crack'em or time runs out.

TeoTwawki (approved):

claywar (approved):

PR #3763, opened 2023-04-14
Hide items with no buy history until listed for first time

CatsEyeXI:
**_I affirm:_**

This code was written by @OpheliaXI. I am simply porting it over to LSB.

This PR will hide any items from the Auction House that do not have a buy history.
This is a more realistic scenario for servers, as some gear is unobtainable or uncoded.
If the operator chooses, they can use a 3rd party tool like pydarkstar to populate their auction house.

Before change:
You will see every single item, whether it's been purchased before or not

After change:
Only items with a buy history are visible on the auction house.

### DIFF
--- a/src/search/data_loader.cpp
+++ b/src/search/data_loader.cpp
@@ -146,6 +146,7 @@ std::vector<ahItem*> CDataLoader::GetAHItemsToCategory(uint8 AHCategoryID, int8*
                         break;
                     }
                 }
+
                 if (!itemFound)
                 {
                     ahItem* PAHItem       = new ahItem;
--- a/src/search/data_loader.cpp
+++ b/src/search/data_loader.cpp
@@ -119,44 +119,49 @@ std::vector<ahItem*> CDataLoader::GetAHItemsToCategory(uint8 AHCategoryID, int8*
             ItemList.push_back(PAHItem);
         }
     }
-    const char* noQtyQuery = "SELECT item_basic.itemid, item_basic.stackSize "
-                             "FROM item_basic "
-                             "LEFT JOIN auction_house ON item_basic.itemId = auction_house.itemid "
-                             "LEFT JOIN item_equipment ON item_basic.itemid = item_equipment.itemid "
-                             "LEFT JOIN item_weapon ON item_basic.itemid = item_weapon.itemid "
-                             "WHERE aH = %u AND auction_house.itemid IS NOT NULL "
-                             "GROUP BY item_basic.itemid "
-                             "%s";
 
-    int32 noQtyRet = sql->Query(noQtyQuery, AHCategoryID, OrderByString);
-    if (noQtyRet != SQL_ERROR && sql->NumRows() != 0)
+    if (settings::get<bool>("search.OMIT_NO_HISTORY"))
     {
-        while (sql->NextRow() == SQL_SUCCESS)
+        const char* noQtyQuery = "SELECT item_basic.itemid, item_basic.stackSize "
+                                 "FROM item_basic "
+                                 "LEFT JOIN auction_house ON item_basic.itemId = auction_house.itemid "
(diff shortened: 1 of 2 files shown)


zach2good:
I'm not against the idea, but it isn't 'as retail', since retail doesn't have any reason to omit any results.

You could wrap the whole additional block in a setting in search.lua and use that to toggle the new behaviour, defaulting to false. OMIT_NO_HISTORY or something

CatsEyeXI:
> I'm not against the idea, but it isn't 'as retail', since retail doesn't have any reason to omit any results.
> 
> You could wrap the whole additional block in a setting in search.lua and use that to toggle the new behaviour, defaulting to false. OMIT_NO_HISTORY or something

I think I did that right, seems to work.

TeoTwawki:
What retail does is track if the item has ever been _listed_ before. even if it never sells, as long as somebody made the attempt its in the list forever after that,

TeoTwawki (approved):
if you ask me, we're better than retail. and this setting is the closest we're getting to what retail does without a more painful reworking. So I aprove.

src/search/data_loader.cpp
```diff
@@ -119,6 +119,49 @@ std::vector<ahItem*> CDataLoader::GetAHItemsToCategory(uint8 AHCategoryID, int8*
...
+                uint16 CurrItemID = sql->GetUIntData(0);
+                bool   itemFound  = false;
+                for (ahItem* PAHItem : ItemList)
+                {
+                    if (PAHItem->ItemID == CurrItemID)
+                    {
+                        itemFound = true;
+                        break;
+                    }
+                }
```
TeoTwawki, line 148: failed to notice the lack of line break here before I reflex-approved.

TeoTwawki (changes requested):

0phie:
So we would need a separate table to get this to operate the exact way Teo described retail operating as. This as although putting it in item_basic.sql would be great, it would likely lead to overwriting of that data pretty quick. It wouldn't be hard to add a new protected table that we check when an item is listed with a "Hey this has been listed before" or "This is a new item to the AH".

TeoTwawki:
> So we would need a separate table to get this to operate the exact way Teo described retail operating as. This as although putting it in item_basic.sql would be great, it would likely lead to overwriting of that data pretty quick. It wouldn't be hard to add a new protected table that we check when an item is listed with a "Hey this has been listed before" or "This is a new item to the AH".

I'm not sure anyone in or out of the team cares about that feature enough to make it 100% that way in the 1st place.

Item basic is read only atm, so I don't think we'd want to store there. We could query the AHs rows to determine if the item was ever _listed_ (a row will exist, even if it hasn't sold, if we don't discard that row on return - some adjusting would be required) but honestly a new table to store some data like "have we seen this before" seems a lot saner - but really only if we also stopped doing the whole "lets load everything from item basic into here!" that we currently do. And really in a lot of ways the AH is what I call better-than-retail, so its kind of awkward to hold to the "must be retail accurate" standard 100%

I can actually prove what retail does there tho, if anyone cares. :)

I think this is fine as is till someone just can't stand that we're imperfect.

src/search/data_loader.cpp
CatsEyeXI: Hi Teo, I'm not clear on what change I need to make here, can you please explain? :)

TeoTwawki: new line between the closing brace and the if that follows. not sure why style checks didn't complain.

TeoTwawki:
```
}
if (stuff)
{
    moreStuff();
}
```
vs 
```
}

if (stuff)
{
    moreStuff();
}
```

TeoTwawki (approved):

zach2good (approved):

PR #3767, opened 2023-04-15
[Abyssea] Set event 2180 when timer is over and move set pos in onEventFinish

Chronos77:
**_I affirm:_**

Set event 2180 when timer is over and move set pos in onEventFinish

review of the time display at the end of the timer (see it on retail)

Enter in abyssea and do not take visitant status 

Wait 5 minutes

### DIFF
--- a/scripts/globals/abyssea.lua
+++ b/scripts/globals/abyssea.lua
@@ -1098,6 +1098,13 @@ xi.abyssea.onZoneIn = function(player)
     end
 end
 
+xi.abyssea.onEventFinish = function(player, csid, option)
+    if csid == 2180 then
+        local zoneID = player:getZoneID()
+        player:setPos(unpack(xi.abyssea.exitPositions[zoneID]))
+    end
+end
+
 xi.abyssea.afterZoneIn = function(player)
     local zoneID = player:getZoneID()
     local ID = zones[zoneID]
--- a/scripts/globals/effects/visitant.lua
+++ b/scripts/globals/effects/visitant.lua
@@ -6,7 +6,7 @@ require("scripts/globals/zone")
 -----------------------------------
 local effectObject = {}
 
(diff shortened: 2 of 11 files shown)


scripts/globals/abyssea.lua
```diff
@@ -1098,6 +1098,13 @@ xi.abyssea.onZoneIn = function(player)
     end
 end
 
+xi.abyssea.onEventFinish = function(player, csid, option)
```
claywar, line 1101: onEventFinish (even if not used) should have 4 parameters (player, csid, option, npc)

scripts/globals/effects/visitant.lua
```diff
@@ -6,7 +6,7 @@ require("scripts/globals/zone")
 -----------------------------------
 local effectObject = {}
 
-local remainingTimeLimits = { 300, 120, 60, 30, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1 }
+local remainingTimeLimits = { 300, 240, 180, 120, 60, 30, 10, 5, 4, 3, 2, 1 }
```
claywar, line 9: Can you confirm that this is the behavior on retail for updates?  In original testing, the previous version was what I observed.

Chronos77: image
This is a capture i made on retail 10 minutes ago

scripts/globals/abyssea.lua
Chronos77: same for zone object? 

this is an example for AlTaieu
image

however, the name may not be correct, could I put onZoneEventFinish?

claywar: Fair enough, since there is no npc to interact with here

claywar (approved):

scripts/globals/abyssea.lua
TeoTwawki: is actually the event target, which is some edge cases is not an npc. on occasion it is the player themself. I don't think the zone object is ever it tho. On retail SE usualy uses a stand in dummy NPC for such interactions.

PR #3786, opened 2023-04-18
[core] Change ABSORB_DMG_TO_MP to work from all sources of damage

WinterSolstice8:
**_I affirm:_**

Does what it says on the tin, ABSORB_DMG_TO_MP mod will now work from all damage sources that eventually call takeDamage in C++, negative addhp and/or delhp calls will not proc this.

`!setmod ABSORB_DMG_TO_MP 100`
take damage from autoattacks and get 1:1 mp back conversion
take damage from spells and get 1:1 mp back conversion (used to be broken)
take damage from mob abilities and get 1:1 mp back conversion (used to be broken)

`!setmod REGEN_DOWN 100` and do not get any MP conversion
absorb element via `${ELEMENT}_ABSORB` mod,  do not get any MP conversion
absorb phys damage via `PHYS_ABSORB` mod, do not get any MP conversion

do not get MP conversion with stoneskin/phalanx absorbing all damage

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -587,6 +587,15 @@ int32 CBattleEntity::takeDamage(int32 amount, CBattleEntity* attacker /* = nullp
         this->SetLocalVar("weaponskillHit", 0);
     }
 
+    if (getMod(Mod::ABSORB_DMG_TO_MP) > 0)
+    {
+        int16 absorbedMP = (int16)(amount * getMod(Mod::ABSORB_DMG_TO_MP) / 100);
+        if (absorbedMP > 0)
+        {
+            addMP(absorbedMP);
+        }
+    }
+
     return addHP(-amount);
 }
 
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -5268,12 +5268,7 @@ namespace battleutils
         }
         else
         {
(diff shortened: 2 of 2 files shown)


TeoTwawki (approved):
Original author(s) possibly intended there to be 1 for magic and then later 1 for physical got added and they never bothered to rename, and never bothered to make one for both. We may want to later make one for just magic damage later to go with that physical one sitting just below the remove part at line 7050-ish of battleutils.

WinterSolstice8:
I thought that might be the case, however, the mod in modifier.h has this comment:
`ABSORB_DMG_TO_MP   = 516, // Unlike PLD gear mod, works on all damage types (Ethereal Earring)`

EDIT: there is also already `ABSORB_PHYSDMG_TO_MP   = 426,  // Absorbs a percentage of physical damage taken to MP.`

TeoTwawki:
> I thought that might be the case, however, the mod in modifier.h has this comment: `ABSORB_DMG_TO_MP = 516, // Unlike PLD gear mod, works on all damage types (Ethereal Earring)`
> 
> EDIT: there is also already `ABSORB_PHYSDMG_TO_MP = 426, // Absorbs a percentage of physical damage taken to MP.`

I'd have to check the commit log to be sure, but we made a lot of the comment long after the fact.

completing the set for both all dmg and each type separate might be useful down the line anyway, if someone feels like it another day.

PR #3795, opened 2023-04-20
Fixes skill level table in utils

Frankie-hz:
**_I affirm:_**

Added Rank G and the 70+ rank ranged for skill level calculations. These were missing for the last 8 years so why not add them now...right!?

These are currently only effecting magic hit rate.

TLDR: Mobs 70+ are missing a lot of MACC from the table being incomplete.

See: https://wiki.ffo.jp/html/2570.html for chart

Go to any mobskill that uses magic and see that you get resisted less? Not entirely sure what to put here :/

### DIFF
--- a/scripts/globals/utils.lua
+++ b/scripts/globals/utils.lua
@@ -319,8 +319,7 @@ end
 --     SKILL LEVEL CALCULATOR
 --     Returns a skill level based on level and rating.
 --
---    See the translation of aushacho's work by Themanii:
---    http://home.comcast.net/~themanii/skill.html
+--    See: https://wiki.ffo.jp/html/2570.html
 --
 --    The arguments are skill rank (numerical), and level.  1 is A+, 2 is A-, and so on.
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):

TeoTwawki (approved):
RubbadubdubSTAMP thanks for the post 75 age update. 🖖

scripts/globals/utils.lua
```diff
@@ -348,15 +350,21 @@ local function getSkillLevelIndex(level)
         rangeId = 50
     elseif level > 60 and level <= 70 then
         rangeId = 60
-    else
+    elseif level > 70 and level <= 75 and (rank > 2) then -- If this is Rank A+ or A- then skip
```
claywar: Useless parens here can be removed

Frankie-hz: Fixed

claywar (approved):

PR #3801, opened 2023-04-22
Records of Eminence: Items Global, Other, Limbus & ZNM Objectives

Era-Lusiphur:
**_I affirm:_**

Converts magic numbers of items to readable names using `items.lua` global.

Implement most of the the Content (Other) Records of Eminence objectives, excluding `Unlock Treasure Chests and Coffers`, `Mons.: Total Monsters Vanquished` and `Reaching the Crest`, which are listed in `roe_records` but commented out for now. Salvage boss mobids were set using the only ones with mob_spawn_points currently in the system.

Also implements Limbus and ZNM Records of Eminence objectives.

In a near future PR, most or all of the `mobKill` objectives for specific mobs/NMs (and not families/ecosystems) will be moved to being triggered in their own script file. This removes additional magic numbers from the RoE global but also stops processing for trial kills outside of killing that specific mob.

Set RoEs, do objectives, get rewards.


scripts/globals/roe_records.lua
```diff
@@ -5868,6 +5953,336 @@ function getRoeRecords(triggers)
...
+            reqs = { itemID = set { 1907 } },
+            flags = set { "repeat" },
+            reward = { sparks = 300, xp = 1500, accolades = 30 },
+        },
+
+        [778] =
+        { -- Spoils - Smoky Chip
+            trigger = triggers.itemLooted,
+            goal = 1,
+            reqs = { itemID = set { 1910 } },
```
Xaver-DaRed: This raw item IDs could be converted to the item.lua file.
Some of them may already be in there. I which case, we should be using the item reference.

Raw IDs = bad

Era-Lusiphur: To document: Some chatter has occurred in DMs and this will be done, along with all existing itemIDs. In addition, mob kill triggers will be moved to the script files for the mobs themselves.

Xaver-DaRed (approved):
Still missing a few item IDs, but I'm aproving anyway.
Good job

claywar (approved):

PR #3813, opened 2023-04-24
[Homepoints] General cleanup, update raw ids with enums and implement destination check.

Xaver-DaRed:
**_I affirm:_**

- Replaces raw Zone IDs with xi.zone.ZONE_NAME
- Moves homepoint.lua to `teleports` folder
- Implements a check to see if destination is actually unlocked

Teleport regularly.

### DIFF
--- a/modules/custom/lua/homepoint_heal.lua
+++ b/modules/custom/lua/homepoint_heal.lua
@@ -3,7 +3,7 @@
 -- Set if you want Home Points to heal you like in single-player Final Fantasy games.
 -----------------------------------
 require("modules/module_utils")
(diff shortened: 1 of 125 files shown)


TeoTwawki:
I'd rather we not change the path to the global file, given that it handles more than just teleporting its going to be harder to look for by people who don't know or just plain forget. And I will definitely forget and have to grep later.

----

edit:

I'm the minority in the team it seems, while I'd still rather have the parts handling not-teleport be away from the teleport folder, if 2 lsb ppl approve then its merging. Just making that clear.

PR #3822, opened 2023-04-27
Changes WEAPON_BONUS mod to flat instead of percent

Frankie-hz:
**_I affirm:_**

Changes WEAPON_BONUS mod to flat damage instead of percentage. This will make balancing NM/HNMs/etc. much more easy and accurate. Through the discovery of mob stats every mission/hnm has a very specific base damage mod. This will make it much easier to implement.

Also changed the weapon bonus to its proper ATTP according to Jimmayus retail testing shown here: https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit#gid=228593342

Dynamis damage was approximated. This needs much more retail data and should be applied per mob_pool or in each individual mobs lua.

In the light of adding these to a few NMs we know the data points for - I would like to get in the habbit of putting immunities/mods for NM/HNMs inside their Luas so when balancing fights you do not have to look in 7 different places for how or what the mob has on it.

Set prints in the code to show the damage increased? Not sure what to put here.


TeoTwawki:
side notes:
- it would probably have helped if back in the day the mob base weapon damaged wasn't made up guessed at (correct formula is known, I believe - dunno if we presently use it or if we still have made up bs)
- there is also a multiplier field in mob_pools that acts as a percentage, and maybe should not at all exist or just act as an override or bonus amount like this pr does to the mod

Frankie-hz:
> 

If you would like to remove that (please lord let me do it) I will add it here as well. Just give me the ok

Edit: Mob base damage is simply level +2 - thats it and +5 for ranged. If you want me to update that as well I will.

TeoTwawki:
> If you would like to remove that (please lord let me do it) I will add it here as well. Just give me the ok

haha I'll keep you in mind if we green light killing it! as always thanks for the PR Frankie

Frankie-hz:
> 

I got more coming! Full rework of Mob HP and mob stats

TeoTwawki:
> I got more coming! Full rework of Mob HP and mob stats

<.<
\>.\>

_Guys sorry but frankie my new bff now._


Part of why we have'em wrong is dsp treated them the same way as players and calculated stats assuming sj was always half of main. Most of the family table is copy pasta with contributors form the days of old not having any actual data about what rank of str this or that goblin had - or sometime snot even knowing what the numbers in those table columns meant.

I've also had an idea about maybe decoupling traits from mobs jobs and make them "bolt on" with a table like players jobs have a trait table but using something other than job id for the mob one (maybe assign per pool so that NMs don't need extra fam IDs) - thats a long term wish to move us away from mobs getting things from a job that they shouldn't as well as reduce usage of `addMod()`/modifier tables that I haven't had time to work on. atm such things require either the job(s) granting the trait, or adding mods manually.

in the meantime we have that mass of 3 modifier tables (id, pool, family) and mob script or mixin script.

/endbraindump

TeoTwawki (approved):

Frankie-hz:
> > I got more coming! Full rework of Mob HP and mob stats
> 
> <.< >.>
> 
> _Guys sorry but frankie my new bff now._
> 
> Part of why we have'em wrong is dsp treated them the same way as players and calculated stats assuming sj was always half of main. Most of the family table is copy pasta with contributors form the days of old not having any actual data about what rank of str this or that goblin had - or sometime snot even knowing what the numbers in those table columns meant.
> 
> I've also had an idea about maybe decoupling traits from mobs jobs and make them "bolt on" with a table like players jobs have a trait table but using something other than job id for the mob one (maybe assign per pool so that NMs don't need extra fam IDs) - thats a long term wish to move us away from mobs getting things from a job that they shouldn't as well as reduce usage of `addMod()`/modifier tables that I haven't had time to work on. atm such things require either the job(s) granting the trait, or adding mods manually.
> 
> in the meantime we have that mass of 3 modifier tables (id, pool, family) and mob script or mixin script.
> 
> /endbraindump

Well luckily for you the TLDR on mob stats is Main Job/Sub Job are always 1:1 and we have yet to find an exception. It's a one line change but without re-doing how mob stats actually work it would break everything and make mobs basically have double stats. Jimmayus just finished up what we called "Project Subjob" so all of that will be pushed upstream. It might have to come in two batches:

1) The current subjob system we have which is 95% correct
2) Add in outliers from the finalization of data that he gathered

TeoTwawki:
> Well luckily for you the TLDR on mob stats is Main Job/Sub Job are always 1:1 and we have yet to find an exception

iirc my convo with Jimmayus about it correctly we should really only be caring about the mainjobs level and what have been being added from sj would still simply be calculated off that anyway so its more like they don't have 2 levels at all.

>  without re-doing how mob stats actually work it would break everything and make mobs basically have double stats. 

yup same reason we haven't touched it yet in lsb. big game of jenga moving all the pieces.

I look forward to seeing the prs

scripts/zones/Dragons_Aery/mobs/Fafnir.lua
```diff
@@ -11,6 +11,7 @@ local entity = {}
 
 entity.onMobSpawn = function(mob)
     mob:setLocalVar("[rage]timer", 3600) -- 60 minutes
+    mob:setMobMod(xi.mobMod.WEAPON_BONUS, 60) -- Level 90 + 50 = 140 Base Weapon Damage
```
Xaver-DaRed: Not to be nitpicky, but this is adding 60. The comment suggests a 50?

scripts/zones/Dragons_Aery/mobs/Nidhogg.lua
```diff
@@ -11,6 +11,7 @@ local entity = {}
 
 entity.onMobSpawn = function(mob)
     mob:setLocalVar("[rage]timer", 3600) -- 60 minutes
+    mob:setMobMod(xi.mobMod.WEAPON_BONUS, 60) -- Level 90 + 50 = 140 Base Weapon Damage
```
Xaver-DaRed: Same here.

The rest of the PR looks good to me.

scripts/zones/Dragons_Aery/mobs/Fafnir.lua
Frankie-hz: Fixed - it was supposed to be 50

scripts/zones/Dragons_Aery/mobs/Nidhogg.lua
Frankie-hz: Fixed - it was supposed to be 50

Xaver-DaRed (approved):

PR #3827, opened 2023-04-28
[Abyssea] Update droplist for qm pop NM Abyssea Tahrongi

Chronos77:
**_I affirm:_**

Set Droplist for QM NM in abyssea Tahrongi

Kill QM NM in abyssea Tahrongi


N3ckB3ard:
I am curious about the method you used to gather the mob drop information?

RAIST5150:
> 

Same.

I get making the move from flat percentage values to the new categories and all... but seeing things like shifts between roughly 1 in 6 chance and 1 in 10.

Unless there are not entries that equate to roughly 10% and 15% that can be used to maintain roughly the same drop rate as before, why the shift to considerably more restrictive drop rate?

Chronos77:
> I am curious about the method you used to gather the mob drop information?

a mix between bgwiki and FFXI db and use this picture

https://user-images.githubusercontent.com/37684138/81483872-1a73cb00-91f6-11ea-8dc4-f3153eabedba.png

Chronos77:
> > 
> 
> Same.
> 
> I get making the move from flat percentage values to the new categories and all... but seeing things like shifts between roughly 1 in 6 chance and 1 in 10.
> 
> Unless there are not entries that equate to roughly 10% and 15% that can be used to maintain roughly the same drop rate as before, why the shift to considerably more restrictive drop rate?

Eft Egg
image
image
image

[Chunk of cockatrice tailmeat](http://ffxidb.com/items/2923)
image
image

Resilient Mane
image
image
image

When we see the data displayed, we can see that the drop rates on these mobs were not necessarily based on a TH0.

RAIST5150:
Thanks for more clarification.

On the surface it looked like there was a significant mismatch between the cited values in the source that BG was also using (ffxidb).

Screenshot_20230428-185546_Samsung Internet

TeoTwawki:
what we normally want in project is matching SE's "buckets" for TH zero using other TH tiers as datapoinst to figure out which bucket it goes in (bucket = rarity tier)

ALL sources we see out there have some skew so we use some educated guessing based on what datapoints are available. ffxiclopedia merges all TH tiers together, ffxidb merges higher tiers of TH, and displays multiples of an item merged together into 1 rate - you can't know of 1 is 100% and the 2nd drop is 50%, or if there are 2 drops of 7% rate each without going onto retail to check because it looks the same. Additionaly items that were added later in the game will be skewed on the same mob at both cites because its total samples is smaller than items that mob already had for years. bgs rarity tiers seem to be based on the wrong data column or outdated info.

some human error and data collection scrape error comes into play as well. some mobs newer items have been mistakenly "fixed" to have same total kills as their older items..Especially at ffxiclopedia. On ffxidb some mobs that never drop something have it shown in their list anyway depending what page you view. and for reasons I do not yet know, sometimes rates on an item page do not match _any_ column of the mob page on ffxidb. 

There is a lot of chaos to sift through.

RAIST5150:
Good to have all this here with these changes in case a challenge comes up down the line... you know it will happen.  :)

N3ckB3ard:
> 

Thank you for this info. I asked because I did a capture of Abyssea - La Theine and when I went to check the LSB SQL for the drop ID's, I noticed FFXIDB was copy/pasted into most of the mob drops. Some items are way over a average drop percentage and it will require a lot of research to remove the bad additions

TeoTwawki:
> Thank you for this info. I asked because I did a capture of Abyssea - La Theine and when I went to check the LSB SQL for the drop ID's, I noticed FFXIDB was copy/pasted into most of the mob drops. Some items are way over a average drop percentage and it will require a lot of research to remove the bad additions

more hands always welcome - most of what's in there is several years old and wrong from our forebears and in many cases from before anyone knew how TH works or before good data samples even existed. And when I added the variables at top of file were mass places onto things that already lined up with them -without- any error checking if that was the correct rate bucket before hand, so you know (I just took numbers already in use and variable-ized them). much work ahead of us.

eventually we'll table the buckets themselves, and do away with raw rate % except for customized drops. before that we need most of the existing table to fit into our best guestimate of the right bucket.

Xaver-DaRed (approved):

PR #3831, opened 2023-04-30
[Event] Egg Hunt Egg-stravaganza 2005-2023

loxleygames:
**_I affirm:_**

Implements Egg Hunt Egg-stravaganza 2005, 2006, 2007, 2008, 2009-2012, 2013, 2014, 2015-2017, 2018, 2019-2023
https://ffxiclopedia.fandom.com/wiki/Egg_Hunt_Egg-stravaganza
(2018 information from https://www.bg-wiki.com/index.php?title=Egg_Hunt_Guide&oldid=548386)
No additional changes up to present retail (2023).

### Overview
* Event Moogles and decorations appear during the defined period
* Additional decorations (Helpers) appear during and after 2015
* Speak to Moogles solo for a random daily egg or in a party to select an egg
* HELM awards an additional "initial egg" every day while the event is active
* Trade combinations of eggs to Moogles once per day for potential rewards

### Features
* Event can be upgraded annually with a setting for each year's rewards
* All combinations up to and including 2019 are fully defined and give correct rewards
* Implements custom words into the settings file to be configured per server
* Implements Egg Helm, where the player receives the second letter of the party member selected (If they are wearing Egg Helm)
* Implements Hatchling Shield to dispense egg related items with a conditional check for ToAU era items
* Implements Piece of Copse Candy costume consumable
* Implements Rabbit Cap to dispense 1-2 initial eggs
* Implements Egg Buffet assembly and disassembly at the MH Moogle

* Update the database, this is required to script the event Moogles
* Copy the updated default main settings into your regular settings
* Set the start/finish dates of the event for testing

```lua
------------------------------------
-- Reset once per day limits
------------------------------------

-- Daily egg(s) from speaking to Moogle
!setplayervar <player> [EGG_HUNT]DAILY_EGG 0

-- Reward for correct combination
!setplayervar <player> [EGG_HUNT]DAILY_REWARD 0

-- Trade in from Lizard Egg, Party Egg etc.
!setplayervar <player> [EGG_HUNT]DAILY_BONUS 0

-- First HELM activity per day awards random egg
!setplayervar <player> [EGG_HUNT]DAILY_HELM 0

-- Initial three must be completed before additional combos
!setplayervar <player> [EGG_HUNT]FIRST_THREE 0

------------------------------------
-- Moogle locations
------------------------------------

-- Southern San d'Oria
!gotoid 17719704

-- Northern San d'Oria
!gotoid 17723702

-- Bastok Mines
!gotoid 17735920

-- Bastok Markets
!gotoid 17740005

--Windurst Waters
!gotoid 17752413

-- Windurst Woods
!gotoid 17764682
```


scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,852 @@
...
+
+-- ----------------------------------
+-- Data
+-- ----------------------------------
+
+xi.events.egg_hunt.data =
+{
+    [xi.zone.SOUTHERN_SAN_DORIA] =
+    {
+        id = 17719704,
```
WinterSolstice8: Hardcoded NPC IDs need to be using the zone ID lua instead for ID shifts

zach2good: You can look up and use offsets like this: https://github.com/LandSandBoat/server/blob/9d1098b4d9ebd45cb66b92201c37351bcc9f5507/scripts/zones/Batallia_Downs/IDs.lua#L112

zach2good: Also the message IDs should be added to each zone, so they're included in UpdateExtractor runs

scripts/globals/moghouse.lua
```diff
@@ -169,6 +169,25 @@ xi.moghouse.moogleTrade = function(player, npc, trade)
...
+
+        local eggComponents =
+        {
+            xi.items.EGG_LOCKER,
+            xi.items.EGG_TABLE,
+            xi.items.EGG_STOOL,
+            xi.items.EGG_LANTERN,
+        }
+
+        if npcUtil.tradeHasExactly(trade, eggComponents) then
```
claywar, line 181: tradeHasExactly functions use confirmItem, while tradeComplete() consumes the entire trade inventory.  This should be changed to player:confirmTrade()

LoxleyX:
* Moved NPCs to dynamic lookups
* Moved dialog to zone IDs
* Confirmed all IDs on updated retail client
* Implemented every outstanding era from wiki
* Implemented every oustanding era reward item
* Refactored names for consistency across all files
* Corrected implementation differences from ASB base
* Fully tested all combinations and fixed numerous edge cases

scripts/globals/helm.lua
```diff
@@ -1535,6 +1535,20 @@ xi.helm.onTrade = function(player, npc, trade, helmType, csid, func)
                 movePoint(player, npc, zoneId, info)
             end
 
+            if
+                xi.events and
+                xi.events.egg_hunt and
+                xi.events.egg_hunt.enabledCheck and
+                player:getVar("[EGGHUNT]DAILY_HELM") < vanaDay()
```
TeoTwawki: Dunno if they ever changed it (was kind of annoying and I thought it was an oversight) but used to be able to cheese the event by re-zoning because every 1st helm attempt would generate another egg even tho the day had not changed. its worth making a note to confirm next year (no change requested here!)

loxleygames: Yeah this implementation definitely won't do that, hopefully that's not a thing on retail or has been patched out! If it is, switching this to a local var should be an easy fix, or maybe make it an optional setting in future.

TeoTwawki: Right, eggxactly

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,882 @@
...
+            table.insert(eggs, { egg, 1 })
+        end
+    end
+
+    return eggs
+end
+
+local hasFirstThree = function(player)
+    return (
+        player:getVar(settings.VAR.FIRST_THREE) == 1 or
```
Xaver-DaRed: Minor, but, we use getCharVar to fetch player variables.

In quest scripts we don't, but outside of them we do, for consistency.

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,882 @@
...
+        settings.ERA_2018 and
+        npcUtil.tradeHasExactly(trade, xi.items.IMPERIAL_EGG)
+    then
+        for i = 1, 8 do
+            table.insert(reward, math.random(xi.items.A_EGG, xi.items.Z_EGG))
+        end
+    end
+
+    if #reward > 0 then
+        if player:getVar(settings.VAR.DAILY_BONUS) >= vanaDay() then
```
Xaver-DaRed: There are more examples, but my previous comment applies to all cases of getVar and setVar

scripts/globals/items/rabbit_cap.lua
```diff
@@ -0,0 +1,28 @@
...
+        result = xi.msg.basic.ITEM_NO_USE_INVENTORY
+    end
+
+    return result
+end
+
+itemObject.onItemUse = function(target)
+    target:addItem(math.random(xi.items.A_EGG, xi.items.Z_EGG))
+
+    if math.random(5) > 4 then
```
Xaver-DaRed: Minor nitpick of mine, but I prefer defining a (min, max) range when using randoms.
I don't really know at glance or memory if `math.random(5)` returns 0-5 or 1-5

Xaver-DaRed:
Really good job overall

scripts/globals/items/rabbit_cap.lua
TeoTwawki: > Minor nitpick of mine, but I prefer defining a (min, max) range when using randoms. I don't really know at glance or memory if `math.random(5)` returns 0-5 or 1-5

starts at 1 when only 1 parameter - can't start at 0 unless explicitly told to, or if doing the 0 to 1 float type.

its still worth using both params just for clarity

TeoTwawki: its still worth using both params just for clarity

scripts/globals/helm.lua
Xaver-DaRed: You forgot to change one getVar() here

Xaver-DaRed (approved):

WinterSolstice8 (approved):
Thanks for your hard work

scripts/zones/Northern_San_dOria/IDs.lua
```diff
@@ -118,6 +118,18 @@ zones[xi.zone.NORTHERN_SAN_DORIA] =
         HERE_TAKE_THIS                = 13079, -- Here, take this...
         IF_YOU_WEAR_THIS              = 13080, -- If you put this on and walk around, something...unexpected might happen...
         THANK_YOU                     = 13081, -- Thank you...
+        EGG_HUNT_REWARD1              = 13116, -- Egg-cellent! Here's your prize, kupo!
```
zach2good: Given that these message IDs are all sequential, we could instead have `EGG_HUNT_OFFSET` and all the messages could use  `ID.text.EGG_HUNT_OFFSET + 2` etc. 

The comments also have to match egg-xactly what is listed by UpdateExtractor, otherwise monthly updates will break: https://github.com/LandSandBoat/UpdateExtractor/blob/main/out/scripts/zones/Northern_San_dOria/Text.lua.raw.txt#L12275

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,882 @@
...
+    end
+
+    for k, v in pairs(tbl) do
+        if xi.settings.main[name][k] then
+            tbl[k] = xi.settings.main[name][k]
+        end
+    end
+end
+
+loadSettings(settings, "EGG_HUNT")
```
zach2good, line 65: Good design here 👍

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,882 @@
...
+        end
+
+    elseif month == settings.FINISH.MONTH then
+        if day <= settings.FINISH.DAY then
+            return true
+        end
+    end
+
+    return false
+end
```
zach2good, line 83: This is good, might look at standardising this across other events

zach2good (commented):
Looking close to being ready 👍

scripts/zones/Northern_San_dOria/IDs.lua
zach2good: Note that update extractor texts havent been updated since Jan, so they're likely quite out of date

scripts/globals/events/egg_hunt_egg-stravaganza.lua
TeoTwawki: I like that its local to the event's own file

TeoTwawki: If we get more seasonal events coded, later we could have a single event settings file.

Xaver-DaRed:
This looks good to go

zach2good (approved):

settings/default/main.lua
```diff
@@ -207,6 +207,35 @@ xi.settings.main =
...
+        ERA_2019 = false, -- Allows trading Apkallu Egg
+
+        -- Consolation prizes for repeating combinations where
+        -- the player has already received the relevant reward
+        MINOR_REWARDS = true,
+
+        -- Set custom combinations, eg. WORD = 12345
+        -- Where WORD  is an arrangement of lettered eggs
+        -- Where 12345 is the itemID for the reward
+        BONUS_WORDS =
```
TeoTwawki, line 234: This and the Era settings above.. You went above and beyond 👏

TeoTwawki: I forget if we already discussed doing it, but I think we've enough here to warrant a new `settings/default/seasonal.lua` file to move everything seasonal event related to instead of storing in `main.lua`

The most recent event to be added was starlight celebration, which did its its is primary event file.

https://github.com/LandSandBoat/server/blob/base/scripts/globals/events/starlight_celebration.lua#L14

I think we can merge as-is then do that migration after.

sql/npc_list.sql
```diff
@@ -29662,7 +29662,7 @@ INSERT INTO `npc_list` VALUES (17764678,'blank','     ',132,59.501,-4.285,32.050
 INSERT INTO `npc_list` VALUES (17764679,'blank','     ',74,127.302,-4.574,97.234,32776,50,50,0,0,64,2,2075,0x0000DA0200000000000000000000000000000000,34,NULL,0);
 INSERT INTO `npc_list` VALUES (17764680,'blank','     ',199,127.790,-4.515,178.663,32769,40,40,0,0,64,2,2075,0x0000DA0200000000000000000000000000000000,34,NULL,0);
 INSERT INTO `npc_list` VALUES (17764681,'blank','     ',135,60.845,-4.635,247.504,32789,50,50,0,0,64,2,2075,0x0000DA0200000000000000000000000000000000,34,NULL,0);
-INSERT INTO `npc_list` VALUES (17764682,'Moogle','Moogle',161,104.823,-5.000,-55.745,1,40,40,0,0,0,2,4194307,0x0000520000000000000000000000000000000000,0,NULL,0);
+INSERT INTO `npc_list` VALUES (17764682,'Egg_Hunt_Moogle','Moogle',161,104.823,-5.000,-55.745,1,40,40,0,0,0,2,4194307,0x0000520000000000000000000000000000000000,0,NULL,0);
```
TeoTwawki: SE moves, deleted, and re-adds these moogles a lot. They will also reuse the same "slot" for different events,

I think we should dynamic entity the seasonal moogles so that they become shift proof.

TeoTwawki: Aside from that concern, this is eggcellent.

loxleygames: The name is to ensure the script is kept separate from Mog House or other event Moogles. They're all just called "Moogle" unfortunately. The only way I could 100% identify them was by triggering the cutscene (It turns). Perhaps in the future we will be able to automatically populate IDs like this from the cutscene data.

TeoTwawki: I know, but thats the thing - SE will move them (or delete them) then the ID is invalid. With a dynamic entity we ignore what SE is doing and ***create*** our own moogles ~~with blackjack and hookers~~ and never have to worry that it was moved/deleted

TeoTwawki: This is from Zach's example of a dynamic NPC, it goes in onZoneInitlize in his example:
```lua
    -- Insert NPC into zone
    local horro = zone:insertDynamicEntity({

        -- NPC or MOB
        objtype = xi.objType.NPC,

        -- The name visible to players
        -- NOTE: Even if you plan on making the name invisible, we're using it internally for lookups
        --     : So populate it with something unique-ish even if you aren't going to use it.
        --     : You can then hide the name with entity:hideName(true)
        -- NOTE: This name CAN include spaces and underscores.
        name = "Horro",

        -- You can use regular model ids (See documentation/model_ids.txt, or play around with !costume)
        look = 2430,

        -- You can also use the raw packet look information (as a string), as seen in npc_list and mob_pools
        -- look = "0x0100020500101120003000400050006000700000",

        -- Set the position using in-game x, y and z
        x = 5.000,
        y = 0.000,
        z = 0.000,

        -- Rotation is scaled 0-255, with 0 being East
        rotation = 128,

        -- Overriding widescan is only available to NPCs.
        widescan = 1,

        -- onTrade and onTrigger can be hooked up to your dynamic entity,
        -- just like with regular entities. You can also omit these.
            -- NOTE: We have to use getPacketName, because the regular name is modified and being used
            --     : for internal lookups
            player:PrintToPlayer("No, thanks!", 0, npc:getPacketName())
        end,

        -- The entity will not be "triggerable" unless you populate onTrigger
            -- NOTE: We have to use getPacketName, because the regular name is modified and being used
            --     : for internal lookups
            player:PrintToPlayer("Welcome to GM Home!", 0, npc:getPacketName())
        end,
    })

    -- Use the mob object however you like
    -- horro:getID() etc.
    -- We're not doing anything with it, so ignore this object
    utils.unused(horro)

    -- You could also just not capture the object
    -- zone:insertDynamicEntity({ ...end)
```
    
Horro is the NPC from the Dragon Quest crossover event.

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,898 @@
...
+    VAR =
+    {
+        DAILY_EGG     = "[EGG_HUNT]DAILY_EGG",
+        DAILY_REWARD  = "[EGG_HUNT]DAILY_REWARD",
+        DAILY_BONUS   = "[EGG_HUNT]DAILY_BONUS",
+        DAILY_HELM    = "[EGG_HUNT]DAILY_HELM",
+        FIRST_THREE   = "[EGG_HUNT]FIRST_THREE",
+    },
+
+    -- Default era is 2005
```
claywar, line 32: Not blocking this PR, but I wonder if there's a way we could consolidate this in the future with other events

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,898 @@
...
+            {   0, -105.500, -5.250,  44.500, "0x01001E0782117020173017401750006000700000", },
+            { 127,  106.500, -5.000, -33.200, "0x01001E0382117020173017401750006000700000", },
+            { 127,  106.500, -5.000, -46.800, "0x01001E0482117020173017401750006000700000", },
+            { 127, -155.000, -4.000, -27.000, "0x01001E0100107120173017401750006000700000", },
+            { 127, -155.000, -4.000, -33.000, "0x01001E0282117120173017401750006000700000", },
+        },
+    },
+}
+
+local messageOffset =
```
claywar, line 218: I like this pattern for offsets a lot.  It makes things much easier to look up when we're using them

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,898 @@
...
+    HELP    = 10, -- Hey, you look new here, kupo! If you need anything, just let me know!
+    REWARD4 = 11, -- Heh heh heh... I have something egg-specially egg-citing in store for you, kupo!
+}
+
+------------------------------------
+-- Helpers
+------------------------------------
+
+xi.events.egg_hunt.charToEgg = function(char)
+    local letter = string.byte(string.lower(char)) - 97
```
claywar, line 239: Perhaps a comment here in the future?  I get that its the ASCII math to determine which letter, but just for others in the future.

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,898 @@
...
+    return xi.items.A_EGG + letter
+end
+
+xi.events.egg_hunt.eggInList = function(eggs, egg)
```
claywar, line 243: I would prefer some more descriptive variable names for eggs, egg, and j in this case, even though j is an iterator.  The function is small enough to understand, but just thinking about future folks

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,898 @@
...
+    end
+
+    return -1
+end
+
+xi.events.egg_hunt.stringToEggs = function(text)
+    local str = string.lower(text)
+    local eggs = {}
+
+    for i = 1, #str do
```
claywar, line 257: Same as above, or comment for intent of the function

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,898 @@
...
+}
+
+local minorReward = function()
+    return minorRewards[math.random(#minorRewards)]
+end
+
+local tradeIn = function(player, npc, trade, zoneID)
+    local reward = {}
+
+    if
```
claywar, line 303: The three below conditions are exclusive, since they depend on tradeHasExactly().  There's not really a reason to have separate blocks for these outside of readability (which I definitely understand).  As a future effort, we might want to table up conditions and iterate over that.

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,898 @@
...
+    { xi.items.FLOWER_EGG, xi.items.HATCHLING_EGG, xi.items.EGG_LOCKER, },
+}
+
+local regionControl = function(player, npc, trade)
+    -- First three required
+    if not hasFirstThree(player) then
+        return
+    end
+
+    for k, v in pairs(regionNames) do
```
claywar, line 475: Same as previous comment, there's value in using better variables in loops

scripts/globals/helm.lua
```diff
@@ -1535,6 +1535,20 @@ xi.helm.onTrade = function(player, npc, trade, helmType, csid, func)
                 movePoint(player, npc, zoneId, info)
             end
 
+            if
```
claywar, line 1538: Note to future self: Move some magian functions to utils to make this nested check unnecessary

claywar (approved):
Stamping this one.  Overall it looks good, but my biggest suggestion would be commenting and more descriptive variables in the future.  Most functions are easy to discern, but always remember looking at something today is very different than looking at it a couple of days later, or even years from now.

claywar (approved):
Stamping this one.  Overall it looks good, but my biggest suggestion would be commenting and more descriptive variables in the future.  Most functions are easy to discern, but always remember looking at something today is very different than looking at it a couple of days later, or even years from now.

sql/npc_list.sql
loxleygames: While this certainly makes sense (And I've used dynamic entities for the decorations where there is no event interaction), there are still a few limitations. For example, the Moogle turns while the player is locked in the cutscene and pivots back afterwards.

We could fake this with `lookAt` but we'd need to only send the update packet to that player, then hook onto `onEventFinish` to reset it. I'm very interested in expanding these kind of functions in the core, as it helps me write convincing custom content. Maybe this is something I will look to pick up in future.

Thanks for your suggestions and kind words.

scripts/globals/events/egg_hunt_egg-stravaganza.lua
loxleygames: Thanks for your suggestions @claywar! I'm happy to follow up on these recommendations if you guys don't mind waiting on the merge.

scripts/globals/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -0,0 +1,898 @@
...
+        end
+    end
+end
+
+xi.events.egg_hunt.onEventFinish = function(player, csid, option)
+    local zoneID = player:getZoneID()
+    local ID = zones[zoneID]
+
+    -- "Forget it." or out of range
+    if option > 28 then
```
claywar, line 731: For bonus points here as well, this seems suspicious, and smells like a bitfield.  If >= 3 (0b11), I'd be willing to bet if you shift right two, its most likely a mask based on it from there.

sql/npc_list.sql
TeoTwawki: ah right, SE turns the actual NPC ID associated in their event code. amending `lookAt()` and `showText()` (both of them do turning behaviors) has been on my back burner for awhile - some NPCs we currently use it on should only appear to turn for the person interacting with them, and others should turn for _everyone_ but you are right that we'd be doing fakery even with correct event ID launching.

The method used for turning npc's by those functions also has the hilarious unintended side effect I refer to as "bobble head" - they continue to try and follow the player after the interaction, which makes their head bob up and down as you move.

TeoTwawki: I have a work in progress that should solve this facing issue. Should be done pretty soon.

TeoTwawki: `facePlayer()` is now in the upstream

PR #3838, opened 2023-05-04
[Abyssea] Update message ID for key items obtained

Chronos77:
**_I affirm:_**

Update message ID for key item obtained as retail

!messagespecial 7320 1280 
!messagespecial 7220 1280

### DIFF
--- a/scripts/globals/abyssea.lua
+++ b/scripts/globals/abyssea.lua
@@ -722,7 +722,7 @@ xi.abyssea.giveNMDrops = function(mob, player, ID)
     for k, v in pairs(normalDrops) do
         if xi.abyssea.canGiveNMKI(mob, 20) then
             playerClaimed:addKeyItem(v)
-            playerClaimed:messageSpecial(ID.text.KEYITEM_OBTAINED, v)
+            playerClaimed:messageSpecial(ID.text.PLAYER_KEYITEM_OBTAINED, v)
         end
     end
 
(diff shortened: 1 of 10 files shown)


Xaver-DaRed:
I'm not an expert on Abussea, but...

So what I'm seing here is that abyssea has a different message for key items than other zones.
I'm guessing this applies to pop key items.
Have you made sure this apply to all cases/means in which a player can obtain a Key Item though?

For example... Maybe the atma vendor?

claywar:
I would suggest keeping the previous definition, since that is used by npcUtil functions, and then create a new enum value for the ones used in Abyssea if this is indeed different.

Chronos77:
> I'm not an expert on Abussea, but...
> 
> So what I'm seing here is that abyssea has a different message for key items than other zones. I'm guessing this applies to pop key items. Have you made sure this apply to all cases/means in which a player can obtain a Key Item though?
> 
> For example... Maybe the atma vendor?

I did a lot of capturing in abyssea konstchat and tahrongi for KI and Abyssite, I only saw this message when has ki pop item and abyssite, however with my character no longer have access to the "Atma Fabricant", I would have to up another character. 

I'm actually not sure that this message is not used at all.  as Claywar suggests, I'll leave the old message.

Xaver-DaRed (approved):

zach2good (approved):

PR #3842, opened 2023-05-05
 Move individual mobkill RoE objectives to mob scripts

Era-Lusiphur:
**_I affirm:_**

Many Records of Eminence objectives are tied to specific mob IDs. In addition to presenting possible issues if the mob IDs shift, the objectives are processed on every mob killed in the game. So it was decided in a group chat that the best way forward would be to move as many of these objectives as possible into the onMobDeath function of the mob's scripts themselves. This PR will adjust as many of those as possible in this fashion unless some interesting issue appears.

Import changes, flag RoE, kill NM.

### DIFF
--- a/scripts/globals/roe_records.lua
+++ b/scripts/globals/roe_records.lua
@@ -3027,8 +3027,6 @@ function getRoeRecords(triggers)
 
         [216] =
         { -- Subjugation: Jaggedy-Eared Jack
-            trigger = triggers.mobKill,
-            reqs = { mobID = set { 17187111 } },
             reward = { sparks = 250, xp = 500 },
         },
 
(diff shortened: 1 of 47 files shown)


Kreidos:
Besides this being a waste of Lusiphur's valuable time, with no appreciable performance or readability payoff, this is gonna complicate future debugging by spreading the logic over hundreds of files. If you want to index all the mobIDs then sure, do that, but this is just lunacy. Having thousands of lines in mob files was specifically what everyone wanted to avoid when this system was built.

Did anyone even do performance testing to see how negligible this is? There's records that trigger every single time anyone damages anything, and even those run just fine; so I seriously ask how significant a much less frequent event like mob-death really is in comparison...

TeoTwawki:
it never should have been allowed to have raw IDs in the 1st place, is bad practice and leads to maintenance hell. We could do without the attitude and foss is the wrong environment to be defensive of ones code..

edit: For the record. An NM call back isn't going to be thousands of lines, its set once then never again, and its proofed against future ID shifting which _**does**_ in fact happen to mobs on occasion, like when SE suddenly adds apex or locus mobs to an older zone.  Lusi can waste his time however he likes.

PR #3844, opened 2023-05-08
[Fix] Prevent trusts from healing if master has enmity

Xaver-DaRed:
**_I affirm:_**

As title suggests.

Summon any trust.
!inmortal yourself, a mob and your trust.

Engage a mob, !hp 1 your trustand turn arround and/or disengage.

Whats the trust not heal until mob is dead or enmity is lost.

### DIFF
--- a/src/map/ai/controllers/trust_controller.cpp
+++ b/src/map/ai/controllers/trust_controller.cpp
@@ -32,6 +32,7 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "../../entities/trustentity.h"
 #include "../../items/item_weapon.h"
 #include "../../mob_spell_container.h"
+#include "../../notoriety_container.h"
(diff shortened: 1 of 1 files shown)


TeoTwawki:
Should actually check if the trust AI is active/engaged instead (not as in actively attacking, but as in it is in battle mode)

On retail I can disengage, bind a mob that is beating on me, stand back and my trusts will in fact recover because even though I have enmity the trust themselves have gone idle. This has gotten me through some fights I would have died in.

Xaver-DaRed:
I've done further testing.

The first healing tick takes about 15 seconds to happen.
During those 15 seconds, if the trust or master recieves an attack, even if it misses, the first tick timer is reset.

Hence why you, Teo, can proc the heals if you bind or sleep them. Any Status effect that prevents mob action would do.

PR #3849, opened 2023-05-08
[sql] Add missing Nashmau NPCs from capture

claywar:
**_I affirm:_**

Adds additional NPCs to Nashmau from retail captures.

NOTE: This does not implement NPC scripts

Go to Raraoon, see Raraoon


Xaver-DaRed (approved):

WinterSolstice8:
do we not have a content tag for the extra HPs? not that I particularly mind anyway

WinterSolstice8 (approved):

claywar:
> do we not have a content tag for the extra HPs? not that I particularly mind anyway

We could, but those aren't visible outside of captures in most cases

TeoTwawki:
RE: the extra homepoints

Nashmau's a pretty small town. I'm not sure where they would have used them even hypothetically, so their existence in the dat is likely a side effect of the tooling and copy pasta/templating used on SEs end.

As far as a content tag I think we should tag every homepoint that was added after the update that gave them teleport capability, regardless of if they are used or not. Then Servers can optionally turn those off if they are worried about some noob getting stranded far from home without warp because they foolishly set homepoint someplace dangerous.

PR #3850, opened 2023-05-09
[xit] Inital commit of xit

zach2good:
**_I affirm:_**

`dbtool` is the best thing since sliced bread. I'd like to expand it to be the one-stop-shop for managing a server. Once you've downloaded and built the server, everything on every platform should be handled through `dbtool`. Since this is quite the expansion of its current duties, I'm proposing we also rename it to: `xit`.

Mainly pushing it up into a draft PR so I don't accidentally lose anything as I move some things around

### DIFF
--- a/tools/xit.py
+++ b/tools/xit.py
@@ -1,0 +1,223 @@
+# xit: XI (Server Management) Tool
+# Pronounced "zit"
+# All-in-one management tool for LSB servers
+# Usage: ./xit.py <check, start, stop, help>
+# or
+# Double-click xit.py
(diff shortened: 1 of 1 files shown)


KthulupwnsAsura (approved):

zach2good:
Kind of not needed anymore - there are new maintenance tasks in dbtool for configuring and launching the server in different configurations, etc. We don't have enough want/need to spend maintainer time on writing GM tools when we're busy focusing on making the emulator itself better.

PR #3857, opened 2023-05-09
adds function to disallow mob from widescan

hooksta4:
mob:setMobMod(xi.mobMod.MOBMOD_NO_WIDESCAN, 1)

**_I affirm:_**

Allows for mobs to be removed from Widescan.

onMobSpawn
    mob:setMobMod(xi.mobMod.MOBMOD_NO_WIDESCAN, 1)

### DIFF
--- a/scripts/globals/status.lua
+++ b/scripts/globals/status.lua
@@ -2585,6 +2585,7 @@ xi.mobMod =
     MAGIC_RANGE            = 72, -- magic aggro range
     TARGET_DISTANCE_OFFSET = 73, -- Adjusts how close a mob will move to it's target. 12 = 1.2 yalm. Positive values to go closer, negative farther.
     ONE_WAY_LINKING        = 74, -- Will link with other mobs in its party (typically the same mob family) while roaming, but will not let others link with it once engaged
+    MOBMOD_NO_WIDESCAN     = 75, -- Disables widescan for a specific mod
 }
 
 -----------------------------------
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -1468,3 +1468,8 @@ bool CMobEntity::OnAttack(CAttackState& state, action_t& action)
         return CBattleEntity::OnAttack(state, action);
     }
 }
+
+bool CMobEntity::isWideScannable()
+{
+    return !(getMobMod(MOBMOD_NO_WIDESCAN));
+}
--- a/src/map/mob_modifier.h
+++ b/src/map/mob_modifier.h
@@ -103,6 +103,7 @@ enum MOBMODIFIER : int
     MOBMOD_MAGIC_RANGE            = 72, // magic aggro range
     MOBMOD_TARGET_DISTANCE_OFFSET = 73, // Adjusts how close a mob will move to it's target. 12 = 1.2 yalm. Positive values to go closer, negative farther.
     MOBMOD_ONE_WAY_LINKING        = 74, // Will link with other mobs in its party (typically the same mob family) while roaming, but will not let others link with it once engaged
+    MOBMOD_NO_WIDESCAN            = 75, // Disables widescan for a specific mod
 };
 
 #endif
--- a/src/map/entities/mobentity.h
+++ b/src/map/entities/mobentity.h
@@ -178,6 +178,7 @@ public:
 
     virtual void Spawn() override;
     virtual void FadeOut() override;
+    virtual bool isWideScannable() override;
 
     bool   m_AllowRespawn; // if true, allow respawn
     uint32 m_RespawnTime;  // respawn time


hooksta4:
Thanks @zach2good  and @Frankie-hz

src/map/entities/mobentity.cpp
```diff
@@ -1468,3 +1468,8 @@ bool CMobEntity::OnAttack(CAttackState& state, action_t& action)
         return CBattleEntity::OnAttack(state, action);
     }
 }
+
+bool CMobEntity::isWideScannable()
+{
+    return !(getMobMod(MOBMOD_NO_WIDESCAN));
```
zach2good, line 1474: Did you test this with `CBaseEntity::isWideScannable()` as part of the condition? By doing this, you're entirely bypassing that original logic for _all mobs_

TeoTwawki:
I approve of the method but you're going to have to rebase and change the ID the mobmod uses after We merge Frankie's other open pr because he uses 75 there.

src/map/mob_modifier.h
```diff
@@ -103,6 +103,7 @@ enum MOBMODIFIER : int
     MOBMOD_MAGIC_RANGE            = 72, // magic aggro range
     MOBMOD_TARGET_DISTANCE_OFFSET = 73, // Adjusts how close a mob will move to it's target. 12 = 1.2 yalm. Positive values to go closer, negative farther.
     MOBMOD_ONE_WAY_LINKING        = 74, // Will link with other mobs in its party (typically the same mob family) while roaming, but will not let others link with it once engaged
+    MOBMOD_NO_WIDESCAN            = 75, // Disables widescan for a specific mod
```
zach2good, line 106: mob*

scripts/globals/status.lua
```diff
@@ -2585,6 +2585,7 @@ xi.mobMod =
     MAGIC_RANGE            = 72, -- magic aggro range
     TARGET_DISTANCE_OFFSET = 73, -- Adjusts how close a mob will move to it's target. 12 = 1.2 yalm. Positive values to go closer, negative farther.
     ONE_WAY_LINKING        = 74, -- Will link with other mobs in its party (typically the same mob family) while roaming, but will not let others link with it once engaged
+    MOBMOD_NO_WIDESCAN     = 75, -- Disables widescan for a specific mod
```
zach2good, line 2588: mob *

src/map/entities/mobentity.cpp
hookstar: Yea so without the ! it defaults to showing no mobs on widescan -

zach2good: This was the snippet I gave you:
```cpp
bool CMobEntity::isWideScannable()
{
    return CBaseEntity::isWideScannable() && getMobMod(MOBMOD_NO_WIDESCAN);
}
```

Sure, you need to negate the `getMobMod`, but the whole point of `override` is that you're replacing and bypassing the original call. So if you don't include the original call with `CBaseEntity::isWideScannable()`, the ONLY condition being checked is the mobmod - so it'll be skipping whether they're alive, or their name is visible etc.

TeoTwawki: > Yea so without the ! it defaults to showing no mobs on widescan -

Really coz it looks like you are not asking that state and returning if your condition is right but actually just returning your condition..

Frankie-hz: > This was the snippet I gave you:
> 
> ```c++
> bool CMobEntity::isWideScannable()
> {
>     return CBaseEntity::isWideScannable() && getMobMod(MOBMOD_NO_WIDESCAN);
> }
> ```
> 
> Sure, you need to negate the `getMobMod`, but the whole point of `override` is that you're replacing and bypassing the original call. So if you don't include the original call with `CBaseEntity::isWideScannable()`, the ONLY condition being checked is the mobmod - so it'll be skipping whether they're alive, or their name is visible etc.

That snippet is gonna default to false and return nothing causing nothing to show on widescan since its an override.

zach2good: We've already talked about the negation of the `getMobMod` call, I'm talking about the omission of `CBaseEntity::isWideScannable()` , which is this:
```cpp

bool CBaseEntity::isWideScannable()
{
    return status != STATUS_TYPE::DISAPPEAR && !IsNameHidden() && !GetUntargetable();
}
``` 

Once the negation is in, what if the mob is set as untargettable, hidden, or disappear? They'll appear on widescan.

zach2good: What I'm asking for is this:
```cpp
bool CMobEntity::isWideScannable()
{
    return CBaseEntity::isWideScannable() && !getMobMod(MOBMOD_NO_WIDESCAN);
}
```

We do a very similar thing in NPCEntity:
```cpp
bool CNpcEntity::isWideScannable()
{
    return widescan == 1 && status == STATUS_NORMAL && CBaseEntity::isWideScannable();
}
```

Frankie-hz: Ah yes, I see. Thank you :D

hookstar: I see what you mean also - thanks. I'll rebase when franks gets merged and make those adjustments. I uh, not too good at reading what's going on. Was trying to do the right thing by PR it but didn't realize the second order effects. Thanks again.

Xaver-DaRed:
Frankie's PR got merged. This needs a rebase to account for it (and adressing Zach's comments) so it can be merged.

hookstar:
Got it. Will be back home tomorrow afternoon and get it sorted then. Thanks

PR #3859, opened 2023-05-10
[Proposal] Introduce tradeRelease() function

Xaver-DaRed:
**_I affirm:_**

As discussed in discord, this creates a new binding to complete trades that dont take away any items, but properly unreserves them.

Have recycle bin enabled.
Go to Sea serpent Grotto. Use a beastcoin to open a door. Throw away beastcoin used. Try to retrieve it from recycle bin.
It should work properly, becouse the item isn't reserved.

### DIFF
--- a/src/map/lua/lua_baseentity.h
+++ b/src/map/lua/lua_baseentity.h
@@ -233,6 +233,7 @@ public:
     uint8 getFreeSlotsCount(sol::object const& locID);         // Gets value of free slots in Entity inventory
     void  confirmTrade();                                      // Complete trade with an npc, only removing confirmed items
     void  tradeComplete();                                     // Complete trade with an npc
+    void  tradeRelease();                                      // Complete trade with an npc without taking any item, but properly "unreserving" them
     auto  getTrade() -> std::optional<CLuaTradeContainer>;
 
     // Equipping
--- a/scripts/globals/conquest.lua
+++ b/scripts/globals/conquest.lua
@@ -1098,11 +1098,13 @@ xi.conquest.overseerOnTrade = function(player, npc, trade, guardNation, guardTyp
                     player:setCharVar("CONQUEST_RING_RECHARGE", getConquestTally())
                     player:showText(npc, mOffset + 58, item, ring.cp, ring.charges) -- "Your ring is now fully recharged."
                 else
+                    player:tradeRelease()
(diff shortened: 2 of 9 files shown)


WinterSolstice8 (approved):
LGTM

src/map/lua/lua_baseentity.h
```diff
@@ -233,6 +233,7 @@ class CLuaBaseEntity
     uint8 getFreeSlotsCount(sol::object const& locID);         // Gets value of free slots in Entity inventory
     void  confirmTrade();                                      // Complete trade with an npc, only removing confirmed items
     void  tradeComplete();                                     // Complete trade with an npc
+    void  tradeRelease();                                      // Complete trade with an npc without taking any item, but properly "unreserving" them
```
TeoTwawki, line 236: is this a counterpart to confirm or complete? This makes it look like its an opposite to complete.

claywar:
I still don't believe this is the correct approach for things.  This is being used for NPC trades exclusively (from what I can see), and there are only two scenarios that I can think of that we need to cover:

1. Player trades to NPC, and we handle all logic in the entities onTrade function
2. Player trades to NPC, and is set into an event.

For scenario 1, the trade container should be released at the end of the function.  For scenario 2, the trade container should be released at the end of onEventFinish (as defaults.  We could confirm or complete prior)

In my opinion, the above logic could and should be handled in core, and not rely on someone in lua land manually releasing the items in that container.

src/map/lua/lua_baseentity.h
Xaver-DaRed: This IS an oposite to complete.
There is nothing to confirm as "take" after all

TeoTwawki: I had to confirm we didn't mean confirm :D 

so many trade thingies, and I thought there was some problem with items being added and then no confirm call to eat them.

Xaver-DaRed:
This PR doesn't cover all possible cases. Just some I found while looking at ASBs method and from memory.

However, I understood that, even if we agree in that the trade container should be released by itself if no event happened, we still wanted a separate binding for certain edge cases (Teo) AND so we can eventually automate checking closed/unclosed trades (Zach)

I do not have the skill to code proper behaviour for the "no-event scenarios" but since the binding is supposedly needed anyway, I think it's a good compromise to use it until some1 with more skill than I do can code the proper behavior.

It would just be a matter of then removing the one-liners.

PR #3862, opened 2023-05-11
Resolve some issues with `PrintToArea()`

TeoTwawki:
**_I affirm:_**

 - Certain kinds of messaging could not be seen by the triggering entity (because at some point the message server was adjusted to omit the sending entity)
 - The values were magic numbered and not enumerated
 - There wasn't anything to indicate user error from sending invalid values, now there is
 - Unity chat wasn't supported, now it is

Co-Authored-By: claywar

Without whom I might have taken ages to realize wtf was happening.

log in 2 characters and do :

`!exec player:PrintToArea("test!", 4, 2, "Dummy")`

While SOLO and see every characters sees **party** chat in **shout** range of the solo character, as originally intended.

Side note: I dislike this if/else tower, and considered a switch but that just made it taller if formatted properly. If you got a better idea, feel free to PR it.

### DIFF
--- a/scripts/globals/msg.lua
+++ b/scripts/globals/msg.lua
@@ -53,6 +53,7 @@ xi.msg.area =
     SHOUT       = 2, -- Will display in wide area around player
     PARTY       = 3, -- Will display to players entire party/alliance
     YELL        = 4, -- If yell is enabled in zone, will display.
+    UNITY       = 5, -- Also World Wide
 }
 
 -----------------------------------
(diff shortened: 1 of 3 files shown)


WinterSolstice8:
Unity broadcasts to everyone in any unity? Interesting. I wonder if we need to delineate out which unities that can be sent to. That's probably of limited use though, beyond the eventual Domain Invasion & random chatter dialogue.

TeoTwawki:
> Unity broadcasts to everyone in any unity? Interesting. I wonder if we need to delineate out which unities that can be sent to. That's probably of limited use though, beyond the eventual Domain Invasion & random chatter dialogue.

its mainly that we don't have unity full retail implemented yet I think.

TeoTwawki:
I last minute renamed something, forgot it, and now its telling the world how I didn't test my very finally final revision..brb

TeoTwawki:
There should be the end of my noobing it up now. Fully re-tested on my laptop.

Btw, unity chat works even if the player has joined no unities at all, let alone all the diff unities: it just broadcasts to the channel regardless. I'm ok with that given this is for fakery.

zach2good (approved):

Xaver-DaRed (approved):

PR #3865, opened 2023-05-11
Move LoadSetSpells() back before BuildingCharTraitsTable() in LoadChar()

kelppo:
**_I affirm:_**

This was moved in c652ebdec7a0fa212dc34eb2545266cdea8275e1 which meant that PChar->m_SetBlueSpells is still all 0 in blueutils::CalculateTraits(), resulting in Blue Mage players logging in with their BLU traits missing.

1. !addspell 612 \<me\>
2. !addspell 615 \<me\>
3. Equip both spells to get auto-refresh, observe how the trait is lost after logging out/zoning
4. Apply this patch
5. Observe how the trait now appears after logging in/zoning

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -860,13 +860,13 @@ namespace charutils
         charutils::LoadInventory(PChar);
 
         CalculateStats(PChar);
+        blueutils::LoadSetSpells(PChar);
         BuildingCharSkillsTable(PChar);
         BuildingCharAbilityTable(PChar);
         BuildingCharTraitsTable(PChar);
         jobpointutils::RefreshGiftMods(PChar);
 
         // Order matters as these use merits and JP gifts
-        blueutils::LoadSetSpells(PChar);
         puppetutils::LoadAutomaton(PChar);
 
         PChar->animation = (HP == 0 ? ANIMATION_DEATH : ANIMATION_NONE);
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -860,13 +860,16 @@ namespace charutils
         charutils::LoadInventory(PChar);
 
         CalculateStats(PChar);
+        jobpointutils::RefreshGiftMods(PChar);
+
+        // This must come after refreshing gift modifiers, but before loading job traits.
         blueutils::LoadSetSpells(PChar);
+
         BuildingCharSkillsTable(PChar);
         BuildingCharAbilityTable(PChar);
         BuildingCharTraitsTable(PChar);
-        jobpointutils::RefreshGiftMods(PChar);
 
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
Thank you for the PR. We will need to find out why that change happened in the first place, in order to not re-open an issue while attempting to close another. We will get back to this.

src/map/utils/charutils.cpp
```diff
@@ -860,13 +860,13 @@ namespace charutils
...
         CalculateStats(PChar);
+        blueutils::LoadSetSpells(PChar);
         BuildingCharSkillsTable(PChar);
         BuildingCharAbilityTable(PChar);
         BuildingCharTraitsTable(PChar);
         jobpointutils::RefreshGiftMods(PChar);
 
         // Order matters as these use merits and JP gifts
-        blueutils::LoadSetSpells(PChar);
         puppetutils::LoadAutomaton(PChar);
```
Xaver-DaRed: Ok, so, after speaking with Jmcmorris (and reading his PR)...

> This also fixes an issue with BLU JP gift JP_BLUE_MAGIC_POINT_BONUS is not applied properly when loading a character.

As you can see, simply reverting this change would re-introduce the issue.

After looking at other places in the code, it seems that `jobpointutils::RefreshGiftMods(PChar);` could be moved up, right below `CalculateStats(PChar);`

So the final order would be the following...
```c++
        CalculateStats(PChar);
        jobpointutils::RefreshGiftMods(PChar);
        
        // This must come after refreshing gift modifiers, but before loading job traits.
        blueutils::LoadSetSpells(PChar);

        BuildingCharSkillsTable(PChar);
        BuildingCharAbilityTable(PChar);
        BuildingCharTraitsTable(PChar);

        // Order matters as these use merits and JP gifts
        puppetutils::LoadAutomaton(PChar);
```

Of course, this would require testing.

kelppo: I did more testing and found a different issue altogether, as of commit c944d914bcdfafcf9114143e825ebf757a280db9, with my patch applied and with the order you posted here.

I found that I don't get the extra Blue Magic point immediately after spending JP, I have to zone or log out to get them.

Xaver-DaRed: If this change doesn't introduce that issue, or rather, if that issue already exists, then I would say it's "out of scope" of this PR.

Xaver-DaRed (approved):

src/map/utils/charutils.cpp
TeoTwawki: > I found that I don't get the extra Blue Magic point immediately after spending JP, I have to zone or log out to get them.

Sounds like we fail to send something to the client in some other part of the core

claywar (approved):

PR #3867, opened 2023-05-11
correct line between player ws & mob tp attacks for trust type entities

TeoTwawki:
Anything above 255 is a mobskill, anything below 256 is a player ws - this even includes fomor/shadow skill

While no trust currently uses those, if one were to ever be added it would fail from being considered the wrong type

**_I affirm:_**

Corrects the dividing point between player style ws and mob tp attacks for trust type entities

References:
 - ROM\27\80.dat showing first mob tp attack in its index:  
  `00001	Foot Kick`  
  (note: we currently use the "zeroth" ID as a dummy for another copy of self destruct, no idea why but possibly a mistake as self destruct is 255 which would be 511 with the offset, and we already have a copy at 511.. )
 - our database showing we add 256 to avoid WS ID collision:  
  `INSERT INTO mob_skills VALUES (257,1,'foot_kick',0,7.0,2000,1500,4,0,0,0,0,0,0);`

Give a trust the right model for it (player type or _maybe_ one that already has a gun) and a new skill list with Vulcan Shot in it. If the model isn't right it won't animate as expected but should say the correct skill name in the chat.

### DIFF
--- a/src/map/utils/trustutils.cpp
+++ b/src/map/utils/trustutils.cpp
@@ -734,7 +734,7 @@ namespace trustutils
         for (uint16 skill_id : skillList)
         {
             TrustSkill_t skill;
-            if (skill_id <= 240) // Player WSs
+            if (skill_id <= 255) // Player WSs
             {
                 CWeaponSkill* PWeaponSkill = battleutils::GetWeaponSkill(skill_id);
                 if (!PWeaponSkill)


TeoTwawki:
This is just something I happened to notice while meddling with trust code and the new WS that were added to the game yesterday. Work related to ws and not related to trusts incoming in separate pr.

claywar (approved):

zach2good (approved):

PR #3871, opened 2023-05-12
Roaming distance

MowFord:
**_I affirm:_**

Mob roaming was noticed to wander much further than the value of `MOBMOD_ROAM_DISTANCE`. After some digging it seemed to just be an issue with the navmesh query `findRandomPointAroundCircle`, but after some poking I noticed that the pathfind function `FindRandomPath` was adjusting `startPosition` every iteration of turns. This meant that for mobs with `MOBMOD_ROAM_TURNS` greater than 1, they could effectively chain their roaming points further and further from their spawn point. I'm not sure if this is by design or not, but it seems inaccurate.

After making each roam turn choose a point centered around the spawn position, it was still noticed that roaming was too wide. Seemed like perhaps the units of the radius are off. Traced down `PMob->GetRoamDistance()` only ever being called in the `mob_controller` roam code when calling the aforementioned random roaming circle radius. Removed the arbitrary division by 10 and stored a separate value for the `findRandomPath` call. I am completely ignorant on the units of the detour library and navmesh pointmaps... I've tried various divisors for the `maxRadius` parameter for `findRandomPosition` and nothing seems to make a difference. 

The mobs just roam less frequently with this change, as any points returned by pathfind outside of their roaming radius are just ignored. I know it's inefficient. I know it should be better, I'm just not sure how to improve it.

According to [this comment](https://github.com/recastnavigation/recastnavigation/issues/236#issuecomment-248862072), the max radius doesn't limit the distance the picked point is from the start, it limits the polygons selected for the query. It seems [very complex](https://github.com/recastnavigation/recastnavigation/discussions/606) to fully limit the area for a dtQuery. 

Curious if this is a "good enough" solution to the problem of limiting roam range? This solution results in 
 - what seems like similar roaming frequency
 - more calls to `m_navMesh->findRandomPosition` on average, since erroneous turnPoints are discarded
 - guaranteed roaming turnPoints within the `roam_distance` mobmod

I used a particular spot as a test before and after, mobid 17199307 in valk dunes.
 - !despawnmob 17199307 
 - !spawnmob 17199307 
 - !gotoid 17199307 
 - This puts you on the mob's spawn point. Select the mob and lock on with the distance addon.
 - to speed up the testing, you can `!setmod move 100` and `!setmobmod roam_cool 1`

### DIFF
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -470,13 +470,15 @@ bool CPathFind::FindRandomPath(const position_t& start, float maxRadius, uint8 m
 
     auto m_turnLength = xirand::GetRandomNumber((int)maxTurns) + 1;
 
-    position_t startPosition = start;
+    // Seemingly arbitrary value to pass for maxRadius, all values seem to give similar results, likely due to navmesh polygons being too dense?
+    float maxRadiusForPolyQuery = maxRadius / 10.f;
+    position_t startPosition    = start;
 
-    // find end points for turns
-    for (int i = 0; i < m_turnLength; i++)
+    // find end points for turns, iterate potentially twice as many times to account for erroneous turnPoints
+    for (int i = 0; i < m_turnLength * 2; i++)
     {
-        // look for new point centered around the last point
-        auto status = m_POwner->loc.zone->m_navMesh->findRandomPosition(startPosition, maxRadius);
+        // look for new turnPoint. findRandomPosition doesn't guarantee the new point is within the radius
+        auto status = m_POwner->loc.zone->m_navMesh->findRandomPosition(startPosition, maxRadiusForPolyQuery);
 
         // couldn't find one point so just break out
         if (status.first != 0)
@@ -484,12 +486,24 @@ bool CPathFind::FindRandomPath(const position_t& start, float maxRadius, uint8 m
             return false;
         }
 
-        m_turnPoints.push_back(status.second);
-        // setting this chains the roaming points, drawing circles around each subsequent roaming turn
-        // startPosition = m_turnPoints[i];
+        float distSq = distanceSquared(startPosition, status.second, true);
+        // only add the roam point if it's _actually_ within range of the spawn point...
+        if (distSq < maxRadius * maxRadius)
+        {
+            m_turnPoints.push_back(status.second);
+        }
+        // else
+        // {
+        //     ShowDebug("CPathFind::FindRandomPath (%s - %d) random point too far: sq distance (%f)", m_POwner->GetName(), m_POwner->id, distSq);
+        // }
+        if (m_turnPoints.size() >= m_turnLength)
+            break;
     }
-    m_points       = m_POwner->loc.zone->m_navMesh->findPath(start, m_turnPoints[0]);
-    m_currentPoint = 0;
+    if (m_turnPoints.size() > 0)
+    {
+        m_points       = m_POwner->loc.zone->m_navMesh->findPath(start, m_turnPoints[0]);
+        m_currentPoint = 0;
+    }
 
     return !m_points.empty();
 }
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -468,7 +468,7 @@ bool CPathFind::FindRandomPath(const position_t& start, float maxRadius, uint8 m
         return false;
     }
 
-    auto m_turnLength = xirand::GetRandomNumber((int)maxTurns) + 1;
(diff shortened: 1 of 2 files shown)


MowFord:
FWIW I was curious and loaded up the valkurm navmesh from [this asb branch](https://github.com/AirSkyBoat/xiNavmeshes/tree/2-New-and-8-Update-Navmesh) and it results in much fewer hits to the distance being out of range.

WinterSolstice8 (approved):
Seems reasonable

TeoTwawki:
> I'm not sure if this is by design or not, but it seems inaccurate.

Even its original author might not know. Suspect they changed their mind about it halfway though.

> Removed the arbitrary division by 10 and stored a separate value for the findRandomPath call. I am completely ignorant on the units of the detour library and navmesh pointmaps...

I really hope its in yalms because thats what we use everywhere else but 🤞 your guess good as mine. Division probably had to do with how much could be packed into the mobmod and no float support in the mobmod directly (they are all integers) and the original author really liking their mobmods when making new things.

> Curious if this is a "good enough" solution to the problem of limiting roam range?

I think so.

As an aside, you seem to be picking all the long standing annoyances on my back burner list ❤️ thank you.

MowFord:
After looking at detour and thinking about it, I believe the "units" are, indeed, whatever our x,y,z values are. The disconnect I think is in how detour deals in areas (polygons) vs individual points? The linked github issues and discussions are a bit over my head.

src/map/ai/helpers/pathfind.cpp
```diff
@@ -470,25 +470,40 @@ bool CPathFind::FindRandomPath(const position_t& start, float maxRadius, uint8 m
...
+        // only add the roam point if it's _actually_ within range of the spawn point...
+        if (distSq < maxRadius * maxRadius)
+        {
+            m_turnPoints.push_back(status.second);
+        }
+        // else
+        // {
+        //     ShowDebug("CPathFind::FindRandomPath (%s - %d) random point too far: sq distance (%f)", m_POwner->GetName(), m_POwner->id, distSq);
+        // }
+        if (m_turnPoints.size() >= m_turnLength)
```
WinterSolstice8, line 499: win32 CI doesnt like this line -- `auto m_turnLength` might resolve differently there, We probably need to make `m_turnLength` a uint32_t (or appropriate)

Interestingly, `maxTurns`  is defined as uint8 but casted to `int`

you can probably remove the cast there and use `xirand::GetRandomNumber<uint8_t>(maxTurns)` and then `auto` should resolve it the same everywhere

zach2good: Useful to note that STL `.size()` calls always return `std::size_t`, which tends to be equivalent to `unsigned int` (length dependant on system arch)

https://en.cppreference.com/w/c/types/size_t

MowFord: Thanks, good info. I'll resolve this compiler warning tonight.

MowFord: FWIW i couldn't static cast the rand with `uint8_t`, then used `uint32_t` and got this warning:

image

I found a similar situation in `mobentity.cpp` 

```
// The underlying function in GetRandomNumber doesn't accept uint8 as <T> so use uint32
// https://stackoverflow.com/questions/31460733/why-arent-stduniform-int-distributionuint8-t-and-stduniform-int-distri
uint8 level = static_cast<uint8>(xirand::GetRandomNumber<uint32>(m_minLevel, m_maxLevel + 1));
```

And coded it similarly

Compiled and confirmed roaming still functions as expected.

MowFord: CI failed again. I'm able to replicate this locally choosing the `x86-Release` build.

compiler debugging is not my strong suit, but i'll poke at this for a little bit

zach2good: Without looking at the code, the comparison is with the size-t from the size() method, not the incoming turn count. I'll write up what you need to change tomorrow

MowFord: seems like auto is not choosing uint8 when i do uint8 + 1, but if i wrap the whole thing in a static_cast it works <.<

zach2good: 1 is unintuitively not unsigned, you could skirt this with uint8 + 1U, since numeric literals are signed by default

zach2good: (but if everything is passing CI, don't worry about jiggling things around anymore)

WinterSolstice8:
CI looks good now 👍

zach2good (approved):

PR #3872, opened 2023-05-12
Miscelaneous fixes from ASB

Xaver-DaRed:
**_I affirm:_**

- Removes Weight effect when applying Flee and Chocobo Jig. Left TODO for possible exceptions.
- I seemingly forgot to add which slot to check for iLvl Magic Accuracy. So added it.
- Prevents players from using weaponskills while not engaged.
- Some corrections to synthesis.

- Try to perform Weaponskills while engaged and not engaged.
- Try to do synthesis regularly

### DIFF
--- a/scripts/globals/abilities/chocobo_jig.lua
+++ b/scripts/globals/abilities/chocobo_jig.lua
@@ -20,6 +20,11 @@ abilityObject.onUseAbility = function(player, target, ability)
     local durationMultiplier = 1.0 + utils.clamp(player:getMod(xi.mod.JIG_DURATION), 0, 50) / 100
     local finalDuration      = math.floor(baseDuration * durationMultiplier)
 
+    -- TODO: Chocobo Jig will not override all types of weight effect. Find out which aren't overriden.
+    if player:hasStatusEffect(xi.effect.WEIGHT) then
+        player:delStatusEffect(xi.effect.WEIGHT)
+    end
+
     player:addStatusEffect(xi.effect.QUICKENING, 20, 0, finalDuration)
 end
 
--- a/scripts/globals/abilities/chocobo_jig_ii.lua
+++ b/scripts/globals/abilities/chocobo_jig_ii.lua
@@ -20,6 +20,11 @@ abilityObject.onUseAbility = function(player, target, ability)
     local durationMultiplier = 1.0 + utils.clamp(player:getMod(xi.mod.JIG_DURATION), 0, 50) / 100
     local finalDuration      = math.floor(baseDuration * durationMultiplier)
 
+    -- TODO: Chocobo Jig will not override all types of weight effect. Find out which aren't overriden.
+    if target:hasStatusEffect(xi.effect.WEIGHT) then
+        target:delStatusEffect(xi.effect.WEIGHT)
+    end
+
     target:addStatusEffect(xi.effect.QUICKENING, 20, 0, finalDuration)
 end
 
--- a/scripts/globals/combat/magic_hit_rate.lua
+++ b/scripts/globals/combat/magic_hit_rate.lua
@@ -17,14 +17,13 @@ local elementTable =
     [xi.magic.element.DARK   ] = { xi.mod.DARKACC,    xi.mod.DARK_AFFINITY_ACC,    xi.mod.DARK_MEVA,    xi.mod.DARK_RES_RANK,    0                                 },
 }
 
--- actor Magic Accuracy
+-- Actor Magic Accuracy
 xi.combat.magicHitRate.calculateActorMagicAccuracy = function(actor, target, spell, skillType, spellElement, statUsed, bonusMacc)
-    local actorJob      = actor:getMainJob()
-    local actorWeather  = actor:getWeather()
-    local spellGroup    = spell and spell:getSpellGroup() or xi.magic.spellGroup.NONE
-    local statDiff      = actor:getStat(statUsed) - target:getStat(statUsed)
-
-    local magicAcc      = actor:getMod(xi.mod.MACC) + actor:getILvlMacc()
+    local actorJob     = actor:getMainJob()
+    local actorWeather = actor:getWeather()
+    local spellGroup   = spell and spell:getSpellGroup() or xi.magic.spellGroup.NONE
+    local statDiff     = actor:getStat(statUsed) - target:getStat(statUsed)
+    local magicAcc     = actor:getMod(xi.mod.MACC) + actor:getILvlMacc(xi.slot.MAIN)
 
     -- Get the base magicAcc (just skill + skill mod (79 + skillID = ModID))
     if skillType ~= 0 then
--- a/scripts/globals/job_utils/thief.lua
+++ b/scripts/globals/job_utils/thief.lua
@@ -38,6 +38,17 @@ local stealableSPEffects =
 -----------------------------------
 -- Local Functions
 -----------------------------------
+local function getTHFlvl(player)
+    local thfLevel = 0
+
+    if player:getMainJob() == xi.job.THF then
+        thfLevel = player:getMainLvl()
+    else
+        thfLevel = player:getSubLvl()
+    end
+
+    return thfLevel
+end
 
 local function processDebuff(player, target, ability, debuff)
     local power = 10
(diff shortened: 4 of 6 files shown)


scripts/globals/abilities/chocobo_jig_ii.lua
```diff
@@ -20,6 +20,11 @@ abilityObject.onUseAbility = function(player, target, ability)
     local durationMultiplier = 1.0 + utils.clamp(player:getMod(xi.mod.JIG_DURATION), 0, 50) / 100
     local finalDuration      = math.floor(baseDuration * durationMultiplier)
 
+    -- TODO: Chocobo Jig will not override all types of weight effect. Find out which aren't overriden.
+    if target:hasStatusEffect(xi.effect.WEIGHT) then
```
zach2good, line 24: General question, do we need to check for a status before we try and remove it, or is removal of a status that isn't there a safe no-op?

zach2good (approved):

scripts/globals/abilities/chocobo_jig_ii.lua
Xaver-DaRed: Generaly it isnt needed. It's safe.

However, to fetch the effect power, the effect has to be there first.

Since I KNOW there are certain cases where Weight effect isn't removed (probably based on effect power) I decided to keep the check in there, since I know me, or some1 else, will end up fetching the effect power inside there and removing it based on it.

scripts/globals/job_utils/thief.lua
```diff
@@ -38,6 +38,17 @@ local stealableSPEffects =
 -----------------------------------
 -- Local Functions
 -----------------------------------
+local function getTHFlvl(player)
```
claywar, line 41: I've seen this pattern used for several jobs, and would suggest in the future we create a generic utils function to return.  _also_ there is no safety for the subjob actually being thf here, but I would assume this is tied to the skill itself prior to being called.

src/map/packet_system.cpp
```diff
@@ -955,6 +955,12 @@ void SmallPacket0x01A(map_session_data_t* const PSession, CCharEntity* const PCh
         break;
         case 0x07: // weaponskill
         {
+            if (!PChar->PAI->IsEngaged()) // Prevent Weaponskill usage if player isn't engaged.
```
WinterSolstice8: This isnt retail behavior, SE allows you to do this exact thing via packet injection. Don't know if we really care though

Xaver-DaRed: I assume that if something can only be pulled off via packet injection, it's unintended behavior that would fall into "better than retail".

However, retail does block you from doing WSs while not engaged, regularly.

WinterSolstice8: I don't think it does block you at all unless they changed it recently, shortcuts + gearswap and typing out the name through `//weaponskill` injects an outgoing 0x01A, as well as you can make the argument that ranged weapon skills shouldn't need you to engage to weaponskill. I presume shorthand + luashitacast will do the same

this has been retail behavior since at least abyssea, people were able to do this with spellcast. perhaps this is a good candidate for a setting

claywar: Seconding the idea of putting this behind a setting.  Reduces the risk overall of trying to hunt down any side effect that might happen from this.

TeoTwawki: retail only "allows" it by injection or memory patching (certain add-ons will make this happen), because the devs trusted the client too much when the game was made.

setting third'd

add-on is likely why you've not seen it block you winter.

scripts/globals/abilities/chocobo_jig.lua
```diff
@@ -20,6 +20,11 @@ abilityObject.onUseAbility = function(player, target, ability)
     local durationMultiplier = 1.0 + utils.clamp(player:getMod(xi.mod.JIG_DURATION), 0, 50) / 100
     local finalDuration      = math.floor(baseDuration * durationMultiplier)
 
+    -- TODO: Chocobo Jig will not override all types of weight effect. Find out which aren't overriden.
```
TeoTwawki, line 23: probably a power/subpower/tier value needs to be involved for resolving that TODO later

scripts/globals/job_utils/thief.lua
TeoTwawki: Related problem here: the wikis are really effing wrong about how STEAL RATE is calculated. Level **_may_** be involved with a base rate, but certain items (or perhaps certain mobs that HAVE that item) can be much harder or easier to land a steal on. I can also with zero doubt say no base stat influence steal - not INT not anything.

This has been bugging me for a long time. Our steal rates are bullsh-- that do not at all match experience as a thf on retail.

I think retail steal items have a rate just like any other drop that the devs may have used level as a guideline for setting the rate or may have level corrected it comparing player to mob on each attempt.

src/map/packet_system.cpp
```diff
@@ -4926,6 +4932,21 @@ void SmallPacket0x096(map_session_data_t* const PSession, CCharEntity* const PCh
         return;
     }
 
+    // If the player is already crafting, don't allow them to craft.
+    // This prevents packet injection based multi-craft, or time-based exploits.
+    if (PChar->animation == ANIMATION_SYNTH)
```
TeoTwawki, line 4935: we might want to do this return if their animation state is anything except "normal".

src/map/packet_system.cpp
```diff
@@ -4926,6 +4932,21 @@ void SmallPacket0x096(map_session_data_t* const PSession, CCharEntity* const PCh
...
+    {
+        return;
+    }
+
+    // Force full synth duration wait no matter the synth animation length
+    // Thus players can synth on whatever fps they want
```
TeoTwawki, line 4941: Sounds like an enhancement, not against it tho. 

I wonder what retail does tehre.

src/map/packet_system.cpp
```diff
@@ -4926,6 +4932,21 @@ void SmallPacket0x096(map_session_data_t* const PSession, CCharEntity* const PCh
...
+    if (PChar->m_LastSynthTime + 15s > server_clock::now())
```
TeoTwawki, line 4942: below we had this as 10s  ... era difference? I recall SE make crafting faster at one point

TeoTwawki: This might be another thing to make configurable. We could default to whatever retail is now, and then adjust via module.

Frankie-hz: We had it at 10 and it was still allowing people to fast synth and packet injection after that 10 seconds was up. We tried to set it at 20 but if you hit lastsynth within this time frame it will add the flat time to the next wait time synth. This was the best solution we came up with. 

Edit: There is probably a better way to do this but for the short term this helped stopped almost everything we were looking it.

src/map/packet_system.cpp
Xaver-DaRed: Added a setting to this, so it can be easily toggled on/off

For retail accuracy, default is "off"

scripts/globals/combat/magic_hit_rate.lua
```diff
@@ -17,14 +17,14 @@ local elementTable =
     [xi.magic.element.DARK   ] = { xi.mod.DARKACC,    xi.mod.DARK_AFFINITY_ACC,    xi.mod.DARK_MEVA,    xi.mod.DARK_RES_RANK,    0                                 },
 }
 
--- actor Magic Accuracy
-xi.combat.magicHitRate.calculateActorMagicAccuracy = function(actor, target, spell, skillType, spellElement, statUsed, bonusMacc)
-    local actorJob      = actor:getMainJob()
-    local actorWeather  = actor:getWeather()
-    local spellGroup    = spell and spell:getSpellGroup() or xi.magic.spellGroup.NONE
-    local statDiff      = actor:getStat(statUsed) - target:getStat(statUsed)
```
claywar: CI not happy about the double newline here

Xaver-DaRed: rebased and fixed. probably

claywar (approved):

Xaver-DaRed:
@TeoTwawki Sorry for the tag. Anything you want adressed over here?

TeoTwawki:
> @TeoTwawki Sorry for the tag. Anything you want addressed over here?

oh hey, nothing required but we def want to followup about last synth time - we're increasing its time here after SE shortened it, and frank said it seemed to be basically ignored at the stock speed.

WinterSolstice8 (approved):

PR #3877, opened 2023-05-13
create `facePlayer()` for use with dynamic NPCs

TeoTwawki:
**_I affirm:_**

 When we start an event, most npc's that kick off the event face the player. In the client, **this is always tied to the correct NPC even if you fire it off from the wrong NPC**
 
This function will fake the dynamic entity doing this same facing behavior.  

If the NPC need to be restored to its original rotation via `setPosRot()` in `onEventFinish()` (we can't do that inside the `facePlayer()` because it has no idea when your event ends, but its easy enough to toss the current rot to a variable at event start)

This is different from `lookAt()` in a several ways
 - no arbitrary points
 - by default only the player interacting sees the change
 - and my reason for doing it: no changing amount of arguments to deal with

Bonus points: 
- No bobblehead effect from `m_TargID` where the NPCs head tries to follow you forever without the rest of its body.
- No regretting that the creepy stalker head can't do a full 360 rotation coz that would've been sweet for Halloween.

login 2 characters stand at same NPC. target it and then try:
`!exec target:facePlayer(player)`
`!exec target:facePlayer(player, true)`
`!exec target:facePlayer(player, false`)
And observe the expected behavior of defaulting to only seeing the rotation on 1 of your 2 characters unless explicitly state "false".

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -1562,6 +1562,43 @@ void CLuaBaseEntity::lookAt(sol::object const& arg0, sol::object const& arg1, so
 }
 
 /************************************************************************
+ *  Function: facePlayer()
+ *  Purpose : change entities rotation to the direction of a player
+ *  Example : npc:facePlayer(player, true/false)
+ *  Notes   : Arbitrary points are not supported!
+ *            Bool set "false" changes the facing for everyone.
+ ************************************************************************/
+
+void CLuaBaseEntity::facePlayer(CLuaBaseEntity* PLuaBaseEntity, sol::object const& nonGlobal)
+{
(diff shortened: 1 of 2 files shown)


Xaver-DaRed:
If I'm reading this correctly, you can feed another entity to the function and the npc/whatever will face it too, correct?
If that's the case, why not faceEntity()?

TeoTwawki:
> If I'm reading this correctly, you can feed another entity to the function and the npc/whatever will face it too, correct? If that's the case, why not faceEntity()?

2nd entity was made to be charentity, and lookAt() is entity agnostic

if 2nd entity can't be cast 2 char entity, this bails out

TeoTwawki:
if 2nd entity can't be cast 2 char entity, this bails out

WinterSolstice8 (approved):

PR #3884, opened 2023-05-15
Kill ZONE_TYPE::BATTLEFIELD

TeoTwawki:
**_I affirm:_**

I've declared war on unneeded zone typing. The mobs already know they are in a battlefield, there's no need for the zone to consider itself battlefield only when that's not even retail accurate - zones can be for more than 1 thing, look at Ghelsba Outpost - a "normal"/dungeon type zone you can exp in, that happens to also have a little fenced off BCNM area in it.
- its redundant with MOBTYPE_BATTLEFIELD anyway
- using it over a mobtype check is the cause of several edge cases in the bcnm system

Do some BCNM, see that mobs still link.

 ~~This is in draft mode until I verify that **ALL** the bcnm mobs have their type field correctly set~~ ...

... _**~~Because with it using the zone check all this time, we'd have never noticed if any were wrong.~~**_

~~Also I probably won't finish that for awhile, so if someone wants to assist that and PR just sql changes this could speed up greatly.~~

Made the SQL changes. Tested limbus as well as a few bcnm - everything seems to behave as expected after the changes. Both agro and linkage.


TeoTwawki:
~~I ran some queries to check on what BCNM mobs were missing the mobType value for BCNM mob and judging from mass of results I'd guess virtually every battlefield mob was never flagged as being a battlefield mob.~~

~~If we merged this without correcting that, we'd break basics like linking and agro.~~

fixed

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #3885, opened 2023-05-15
[codeql] Address several warnings and errors identified in code scanning

claywar:
**_I affirm:_**

Chips away at several codeql errors and warnings on the list.  See comments in the commits for problem description and resolution.

No visible changes should be observed, outside of our warning list going down.

### DIFF
--- a/src/map/grades.cpp
+++ b/src/map/grades.cpp
@@ -191,12 +191,12 @@ namespace grade
         return StatScale[rank][scale];
     }
 
-    float GetMobHPScale(uint8 rank, uint8 scale)
+    uint8 GetMobHPScale(uint8 rank, uint8 scale)
     {
         return MobHPScale[rank][scale];
     }
 
-    float GetMobRBI(uint8 rank, uint8 scale)
+    uint8 GetMobRBI(uint8 rank, uint8 scale)
     {
         return MobRBI[rank][scale];
     }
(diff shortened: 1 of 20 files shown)


src/map/grades.cpp
```diff
@@ -191,12 +191,12 @@ namespace grade
         return StatScale[rank][scale];
     }
 
-    float GetMobHPScale(uint8 rank, uint8 scale)
+    uint8 GetMobHPScale(uint8 rank, uint8 scale)
```
Frankie-hz, line 194: Ah good catch - my bad!

cocosolos:
The file logging for those migrations was because depending on the state of the database there could be a rather large amount of corrupted data. Those are the only ones that do it because AFAIK they're the only ones that work directly on blob data where each entry has a possibility of being corrupt. For instance if someone made an incorrect backup of their database then every character could potentially have corrupted data there, or if they're importing single characters doing something like a transfer from another server then those characters may have corrupt data and would need to be manually adjusted. I went with a file output for those so they could be reviewed more easily if there's any issue, which may be problematic if relying on printed output, especially if there's a lot of corrupt data possibly resulting in thousands of lines of output depending on server size. However, these migrations would only be run by someone using data from a very old database at this point, or a very screwed up database. I also think dbtool clears the screen at some points when drawing the menu, though I'm not sure if that's still the case.

Sorry for the long write up, but figured I'd explain the rationale there. I don't think I can see any of the warnings/errors you're getting for this stuff unless I'm missing it somewhere.

claywar:
Good point, restored the file logging for these migrations but changed the logic to use `with` to ensure that regardless of exception thrown, the file is closed at the end of each function.

Xaver-DaRed (approved):

TeoTwawki (approved):
lgtm

PR #3888, opened 2023-05-16
Interaction - Quest: Ayame and Kaede

claywar:
**_I affirm:_**

Converts the Bastok Quest, "Ayame and Kaede" to interaction framework

Follow guides for this quest


scripts/quests/bastok/Ayame_and_Kaede.lua
```diff
@@ -0,0 +1,213 @@
...
+
+    return false
+end
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            return status == QUEST_AVAILABLE and
+                player:getMainLvl() >= xi.settings.main.ADVANCED_JOB_LEVEL
```
Xaver-DaRed, line 57: Wont this need a settings require?

scripts/quests/bastok/Ayame_and_Kaede.lua
```diff
@@ -0,0 +1,213 @@
...
+                            quest:setVar(player, 'Prog', 3)
+
+                            if quest:getVar(player, 'isActor') == 1 then
+                                quest:setVar(player, 'isActor', 0)
+                                npc:hideNPC(xi.settings.main.FORCE_SPAWN_QM_RESET_TIME)
+                            end
+
+                            return quest:keyItem(xi.ki.STRANGELY_SHAPED_CORAL)
+                        else
+                            quest:setVar(player, 'isActor', 1)
```
Xaver-DaRed: I don't understand this variable. Its only used 3 times:
- 1 to check if its 1, to set it to 0, and thats it
- 2 to set it. 1 to 0. 1 to 1.

But it doesnt actually prevent anything from happening? Nor allow it, either.

Also, if its needed, why not set it as a local var? I'm thinking a player dying or warping out midfight.

scripts/quests/bastok/Ayame_and_Kaede.lua
claywar: Technically no, since settings is an empty file, but can add

scripts/quests/bastok/Ayame_and_Kaede.lua
claywar: isActor here is used on the followup interaction which causes the NPC to hide.  There's a lot of discussion around this quest from old times, and this may no longer be accurate, but the person triggering the NM spawn is what matters.  They spawn, everyone gets credit, but as soon as the person triggering hits the spot, everyone waits.

claywar: It _could_ be a localvar, but this is mostly a 1:1 conversion of our existing implementation outside of changing some progress states to make things easier

scripts/quests/bastok/Ayame_and_Kaede.lua
```diff
@@ -0,0 +1,214 @@
...
+                onTrigger = function(player, npc)
+                    if
+                        not isNMSpawned() and
+                        quest:getVar(player, 'Prog') == 2
+                    then
+                        if quest:getLocalVar(player, 'nmDefeated') == 1 then
+                            quest:setVar(player, 'Prog', 3)
+
+                            if quest:getLocalVar(player, 'isActor') == 1 then
+                                quest:getLocalVar(player, 'isActor', 0)
```
Xaver-DaRed: set
`quest:setLocalVar(player, 'isActor', 0)`

claywar: And this is what I get for making changes before coffee

Xaver-DaRed (approved):

PR #3894, opened 2023-05-19
[core] Mob vs. Player Hit Rate Fix

Frankie-hz:
**_I affirm:_**

Previously mobs were being penalized for dLvl. Only players should be penalized for being lower than the defender.

See: https://www.bg-wiki.com/ffxi/Hit_Rate#:~:text=99%25%20For%20one%2Dhanded%20main,95%25%20for%20ranged%20attacks.

Please note the signs are reversed on the wiki but I decided to keep our sign convention the same.

Fight any mob that is lower than you
Depending on the level of the mob you should see a slight to moderator increase in the hit rate of the mobs as they no longer suffer from dLvl penalty.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2616,7 +2616,7 @@ namespace battleutils
 
             if (shouldApplyLevelCorrection)
             {
-                int16 dLvl = PAttacker->GetMLevel() - PDefender->GetMLevel();
+                int16 dLvl = 0;
                 // Skip penalties for avatars, this should likely be all pets and mobs but I have no proof
                 // of this for ACC, ATT level correction for Pets/Avatars is the same as mobs though.
                 bool isPet    = PAttacker->objtype == TYPE_PET;
@@ -2630,17 +2630,17 @@ namespace battleutils
 
                 if (isAvatar)
                 {
-                    if (dLvl > 0)
-                    {
-                        // Avatars have a known level difference cap of 38
-                        hitrate += static_cast<int16>(std::min(dLvl, (int16)38) * 2);
-                    }
+                    dLvl = PAttacker->GetMLevel() - PDefender->GetMLevel();
+                    // Avatars have a known level difference cap of 38
+                    dLvl = std::clamp(dLvl, static_cast<int16>(0), static_cast<int16>(38));
                 }
-                else
+                // Only players are penalized for dLvl
+                else if (PAttacker->objtype == TYPE_PC && PAttacker->GetMLevel() < PDefender->GetMLevel())
                 {
-                    // Everything else has no known caps, though it's likely 38 like avatars
-                    hitrate += static_cast<int16>(dLvl * 2);
+                    dLvl = PAttacker->GetMLevel() - PDefender->GetMLevel();
                 }
+
+                hitrate += static_cast<int16>(dLvl * 2);
             }
 
             // https://www.bg-wiki.com/bg/Hit_Rate


src/map/utils/battleutils.cpp
```diff
@@ -2630,15 +2630,23 @@ namespace battleutils
...
+
+                // Only players are penalized for dLvl
+                if (PAttacker->objtype == TYPE_PC)
                 {
-                    // Everything else has no known caps, though it's likely 38 like avatars
+                    if (PAttacker->GetMLevel() < PDefender->GetMLevel())
+                    {
+                        dLvl = PAttacker->GetMLevel() - PDefender->GetMLevel();
+                    }
+
```
Xaver-DaRed: If you leave the dLvl definition like it was (directly set the level diff, first line changed) you dont need to add it to avatars.

Then, the new block of code would only need to be...
```c++
// Only players are penalized for dLvl
if (PAttacker->objtype == TYPE_PC)
{
    if (dLvl > 0)
    {
        dLvl = 0;
    }

    hitrate += static_cast<int16>(dLvl * 2);
}
```

Xaver-DaRed: Unless, of course, you have future edits in mind for this.

Frankie-hz: You are right - I originally had something like this and then changed it back to its own if statement. I could do something like this as well - whichever you prefer. 
```cpp
if (shouldApplyLevelCorrection)
{
    int16 dLvl = 0;
    bool isPet = PAttacker->objtype == TYPE_PET;
    bool isAvatar = false;

    if (isPet)
    {
        CPetEntity* petEntity = dynamic_cast<CPetEntity*>(PAttacker);
        isAvatar = petEntity && petEntity->getPetType() == PET_TYPE::AVATAR;
    }

    if (isAvatar)
    {
        dLvl = PAttacker->GetMLevel() - PDefender->GetMLevel();
        dLvl = std::min(dLvl, static_cast<int16>(38));
    }
    else if (PAttacker->objtype == TYPE_PC && PAttacker->GetMLevel() < PDefender->GetMLevel())
    {
        dLvl = PAttacker->GetMLevel() - PDefender->GetMLevel();
    }

    hitrate += static_cast<int16>(dLvl * 2);
}
```

Xaver-DaRed:
```c++
if (shouldApplyLevelCorrection)
{
    int16 dLvl    = PAttacker->GetMLevel() - PDefender->GetMLevel();
    bool isAvatar = false;

    if (PAttacker->objtype == TYPE_PET)
    {
        CPetEntity* petEntity = dynamic_cast<CPetEntity*>(PAttacker);
        isAvatar = petEntity && petEntity->getPetType() == PET_TYPE::AVATAR;
    }

    if (isAvatar)
    {
        dLvl = std::min(dLvl, static_cast<int16>(38));
    }
    else if (PAttacker->objtype == TYPE_PC && dLvl > 0)
    {
        dLvl = 0;
    }

    hitrate += static_cast<int16>(dLvl * 2);
}
```

My reasoning behind this, other than being marginaly shorter, is any case where entities arent players (avatar check is to cap it from what I see), I don't think dlvl should be 0. I may be wrong though.

Frankie-hz: If we leave that defined on top (and not set it to 0) then if the entity is a Mob lets say - regardless of its higher or lower when attacking a player there would be some sort of level correction by dLvl.

From the research jimmayus did on mob pdif/acc:
Monster hit rate is: 75 - Floor((acc-eva)/2), no level correction
-->> *there might be positive hit correction*

So in either case the dLvl should be clamping at 0 for mobs regardless. If we can find more proof that there is positive hit correction (like there is for PDIF) then we can add that in later. As for now we have no solid evidence of that until more testing has been done.

Edit: as per the BG wiki as well it only refernces player vs mob hitrate - it seems from testing and other sources that there is no correction for mobs vs players

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #3907, opened 2023-05-21
Add latch to Magian Player Cache to force reload on zoning

### LINKED ISSUE #3615: 🐛 [SQL] Trial of the Magian CharVar issue "Multi Process"
**_I affirm:_**

 Windows 10 (x64- Debug)

## Branch affected by issue 

`base`
image

## Steps to reproduce

I am currently experiencing a issue on CatsEyeXI regarding Trial of the Magian. It seems and through testing Trials are being accounted for correctly for objective total and count. Then when the trial is complete if you are to return to the moogle to turn in your progress or completed item you will be informed your trial is not completed.

Printing some of the local's within
https://github.com/LandSandBoat/server/blob/70680538d6c2a6fec0e960cd5407fbc0df142d5c/scripts/globals/magiantrials.lua#L21
I then adjusted this return to look for the charVar [trial]2

gave me a odd return...
image

```
!checkvar Neckbeard [trial]2
```
Returns the correct Variable for the trial

It seems more than this global is being effected if the getCharVar can not return the value from the DB. Thank you for anyone with information regarding this.

## Update 2/19:
I was successfully able to recreate this issue locally and it exhibits the same behaviors as the issue from CatsEyeXI.
Unfortunately the SQL log error seems to be specific to the DB of the server so I am removing most of the issues related to it.

For my test today I enabled logging as suggested from Zach and updated to the latest build. I then started my multiple processes of map.exe targeting different ports.
(Note: Ru'Lude Gardens is **NOT** on port.54230)

Checking my trial by trading it to the moogle should reflect my current progression of completed trial objectives.
image
image

After completing another objective I returned to the moogle to find my trial progress remains the same even if my CharVariable
increases.
image

I then added the following prints to my console to see if I could collect the data that is being used during this NPC interaction. 
https://github.com/LandSandBoat/server/blob/70680538d6c2a6fec0e960cd5407fbc0df142d5c/scripts/globals/magiantrials.lua#L454-L455
```lua
                local charVar      = player:getCharVar("[trial]2")
                print("_, cacheData:", _, cacheData)
                print("progress:", progress)
                print("itemId:", itemId)
                print("charVar:", charVar)
```
this was my return
image
Then after I restarted the process (map.exe that matched the port to Ru'Lude Gardens) my CharVar reflected correctly and I received the correct progress.
image

This ultimately prevents players from being able to turn in completed trial's as the CharVar I imagine is being set when the exe loads the area and somehow is not being updated when the CharVar increases.
## Expected behavior

Return of the correct CharVar ~~and solution to SQL errors~~

When you have completed the total amount of objectives related to your trial and redeem your reward without a restart of the map process

claywar:
**_I affirm:_**

Fixes #3615

Adds a localvar to returning cached magian data, and if zero (meaning the player has zoned), it is forced to reload data.

Zone between different map instances, and see that data is correct

### DIFF
--- a/scripts/globals/magiantrials.lua
+++ b/scripts/globals/magiantrials.lua
@@ -14,7 +14,10 @@ xi.magian.trialCache = xi.magian.trialCache or {}
 local function getPlayerTrials(player)
     local activeTrials = xi.magian.trialCache[player:getID()]
 
-    if activeTrials then
+    if
+        activeTrials and
+        player:getLocalVar('magianUpdated') == 1
+    then
         return activeTrials
     end
 
@@ -30,6 +33,8 @@ local function getPlayerTrials(player)
     end
 
     xi.magian.trialCache[player:getID()] = activeTrials
+
+    player:setLocalVar('magianUpdated', 1)
 
     return activeTrials
 end


TeoTwawki (approved):

TeoTwawki:
technically, we should should still fix the async issue that causes this. but that does fix the bug itself

claywar:
The issue is we're caching this data in a lua table on the map server, so changing instances could mean stale data being accessed.  The only other way that I can see fixing this would be to cache and access from world server.

TeoTwawki:
> The issue is we're caching this data in a lua table on the map server, so changing instances could mean stale data being accessed. The only other way that I can see fixing this would be to cache and access from world server.

zach was finger pointing the async sql stuff was why I comment

PR #3908, opened 2023-05-21
Change BCNM Level Restriction to only remove Dispellable effects

### LINKED ISSUE #3389: 🐛 effects stripped on battlefield entry
**_I affirm:_**

win server 2019

## Branch affected by issue 
`base`

## Steps to reproduce
Use a food enterm then enter a BCNM. Note that it is removed. On retail most effects are stripped including reraise from spells, but food effect and reraise from items remain.

## Expected behavior
Food effect should remain. This is a relatively new bug.

Reraise from items should also remain, but to my knowledge that was never seperated from spell effect reraise and is thus one of those ancient issues we forget about.

claywar:
**_I affirm:_**

Partially addresses #3389 by changing delete status effects using EFFECTFLAG_DEATH to EFFECTFLAG_DISPELABLE on BCNM entry.  This resolves food being removed, but will not fix reraise, since we currently do not differentiate between item-based and spell-based reraise.

This change mirrors how we handle Level-Sync status outside of BCNMs being applied.

Enter BCNM, still be full.

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -274,7 +274,7 @@ void CBattlefield::ApplyLevelRestrictions(CCharEntity* PChar) const
             cap = settings::get<uint8>("main.MAX_LEVEL"); // Cap to server max level to strip buffs - this is the retail diff between uncapped and capped to max lv.
         }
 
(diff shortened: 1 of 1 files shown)


TeoTwawki (approved):
omnomnomnom

PR #3910, opened 2023-05-22
[core] Fix level sync corrupting charmed pets

WinterSolstice8:
**_I affirm:_**

Changes the check for charmed pets to check if PMobEntity type and not immediately static casting to something it isn't

Charm a pet above level 18, !addeffect level_sync 18 90 on yourself and have the mob detach and not corrupt/crash the server.


TeoTwawki:
the bug works with -any- level. the pet gets its level dropped to 1 via corruption. I tried a whole buncha sync levels, any will do it even if its above what the mob was.

zach2good (approved):

TeoTwawki (approved):

PR #3915, opened 2023-05-23
[lua] [PLD] Adjust cover duration formula to match retail

WinterSolstice8:
**_I affirm:_**

verified on retail:
* This formula is dVIT (user VIT - target MND) in seconds
* https://wiki.ffo.jp/html/2612.html

Use Cover, see bonus duration of 1 second per dVIT

### DIFF
--- a/scripts/globals/job_utils/paladin.lua
+++ b/scripts/globals/job_utils/paladin.lua
@@ -75,9 +75,10 @@ xi.job_utils.paladin.useChivalry = function(player, target, ability)
     return target:addMP(amount)
 end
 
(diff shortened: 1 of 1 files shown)


claywar (dismissed):

WinterSolstice8:
Hmm. Needs more research and better control of variables. I could be wrong here.

WinterSolstice8:
I'm pretty sure the current formula matches retail. When I control for just MND or VIT it matches. ilvl gear just threw a wrench in swaps, had to use accessories.

PR #3917, opened 2023-05-23
Remove duplicate caching of content enabled settings

claywar:
**_I affirm:_**

Removes contentEnabledMap and its references in favor of using settings::get() for the appropriate tag.  This seems to resolve a infrequent startup crash where contentEnabledMap's iterator could become invalid if another thread modifies the map, whereas settings is cached earlier on startup for all values.

1. Start server (not an always occurrence for the issue), and no crashes should occur.
2. IsContentEnabled still returns valid results


Xaver-DaRed (approved):

zach2good (approved):
Ohh good catch

PR #3919, opened 2023-05-23
[Chore] synthutils minor cleaning

Xaver-DaRed:
**_I affirm:_**

Preparation work to add "no material loss" and "always material loss" recipes. Mostly cleanup while I read the code.

No real changes should have happened.If something was working, should work. If something was broken, should remain broken.

Better reviewed by commit.

### DIFF
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -243,7 +243,7 @@ namespace synthutils
 
                 if (synthDiff <= 0)
                 {
-                    if (PChar->CraftContainer->getCraftType() == 1) // if it's a desynth lower success rate
+                    if (PChar->CraftContainer->getCraftType() == CRAFT_DESYNTHESIS) // if it's a desynth lower success rate
                     {
                         success = 0.45;
                     }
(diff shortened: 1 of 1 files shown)


src/map/utils/synthutils.cpp
```diff
@@ -130,7 +130,7 @@ namespace synthutils
      *                                                                           *
      ****************************************************************************/
 
-    double getSynthDifficulty(CCharEntity* PChar, uint8 skillID)
+    int16 getSynthDifficulty(CCharEntity* PChar, uint8 skillID)
```
claywar, line 133: You're changing this return from float to int, but all other functions that call this are float.  Might want to check into usage here.

Xaver-DaRed: This function is called 2 times in the whole src folder.
I did change to int16 the 2 variables that store the result of this function and checked that the code didnt rely on it being a float.

Synth dificulty can never have non-round numbers. its just a substraction. recipe skill level - player skill level (after being clamped)

claywar (approved):

PR #3930, opened 2023-05-25
[client] May 2023 Version Update (Part 2)

claywar:
**_I affirm:_**

* Adds new zone (Gwora - Throne Room) with placeholder values
* Updates Text ID, Version, and Status Effect changes

Nothing is exposed, outside of Southern San d'Oria IDs being displayed correctly


scripts/globals/zone.lua
kaincenteno, line 1: does MAX_ZONE need to be increased to 300?

Xaver-DaRed (approved):

PR #3933, opened 2023-05-26
[Event] Vana'versary Celebration Campaign

neuromancerxi:
**_I affirm:_**

Implements the Vana'versary Celebration Campain

* Implements Event using Dynamic Entities
* Ties Event to Event system
* Creates entries in default settings to Enable/Disable Campaign, Announcement, and Announcement Text
* Implements the "Free to Take" Coffer and item acquisition
* Implements the Moogle that flanks the "Free to Take" Coffer with default greeting dialog, repeat dialog, and Moogle Shirt Gift (and re-acquisition)
* Adds placeholder lua for Chacharoon (**Note**: Mandragora Mania Madness is not implemented).
* Adds "History" Moogle dialog and 2 Item Acquisition logic. 
* Not enough information is available at this time to implement the 3rd "time based" reward.
* Shows/Hides NPCs associated with campaign when campaign becomes active/inactive.
* Adds a few missing shirts (ids and model ids).
* Adds item scripts for Coffer Rewards and makes all rewards which are usable functional.
* Adds npc_utils logic for delivering a just-used charged item (matching Retail behavior).

### Default (Campaign Active May 16 - 31)

* Test with default settings, confirm if NPCs are not visible (out of cycle)
* Pull in `VANAVERSARY_` related settings from default/main into custom settings. Adjust `VANAVERSARY_ALWAYS` to `true`
* Restart map, confirm NPCs are visible and responsive.

### Campaign Enabled

* Confirm NPCs are visible
* Confirm History Moogle, Vanaversary Moogle, and "Free to Take" Coffer have correct dialog, and provide items.
* Note Chacharoon and his flanking pets are visible.
* Note Chacharoon is selectable, pets are not.
* Note Chacharoon provides simple dialog only (Mandragora Mania Madness is currently unimplemented).


cocosolos:
Can't test this myself right now but does the coffer reset the cooldown for usable items when you take them? I know on retail when you take the behemoth suit +1 from the coffer the cooldown is reset so you can't immediately use it, drop it, then take a new one and repeat.

scripts/globals/vanaversary.lua
```diff
@@ -0,0 +1,218 @@
...
+xi.vanaversary.treasureCoffer = function(player, csid, option)
+    if player:getGender() == 1 then -- Male
+        xi.vanaversary.goods = xi.vanaversary.cofferItemsM
+    end
+
+    player:startEvent(csid)
+end
+
+xi.vanaversary.treaureCofferGoods = function(player, csid, option)
+    npcUtil.giveItem(player, xi.vanaversary.goods[option])
```
TeoTwawki: on retail the obtained items start at max cooldown, and I made a binding separate from our normal item adding to handle that in the past on the nexus cape ppl were exploiting by dropping and re-obtaining `player:addUsedItem(xi.items.NEXUS_CAPE)`

the npc utils doesn't yet have an option for this, so it would need updated or else you'd need to use the binding directly + generate the message yourself.

TeoTwawki: Note that this only applies to usable items.

neuromancerxi:
Thanks for the feedback, good call outs and suggestion. I've got an idea of how to address this.

sql/npc_list.sql
```diff
@@ -27359,14 +27359,14 @@ INSERT INTO `npc_list` VALUES (17740016,'Eggsperiment_B','Eggsperiment B',127,-1
...
 INSERT INTO `npc_list` VALUES (17740018,'Eggsperiment_T','Eggsperiment T',127,-155.000,-4.000,-33.000,0,40,40,0,1,40,6,2075,0x01001E0282117120173017401750006000700000,32,NULL,1);
 INSERT INTO `npc_list` VALUES (17740019,'Moogle','Moogle',15,-260.221,-12.021,-77.689,1,40,40,0,0,0,2,3,0x0000520000000000000000000000000000000000,0,NULL,1);
-INSERT INTO `npc_list` VALUES (17740020,'Moogle','Moogle',40,-259.440,-12.021,-79.538,1,40,40,0,0,0,2,4194307,0x0000520000000000000000000000000000000000,0,NULL,1);
--- INSERT INTO `npc_list` VALUES (17740021,'blank','_',0,0,0,0,0,50,50,0,0,112,2,2051,0x0000340000000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17740022,'Moogle','Moogle',105,-293.121,-12.021,-76.843,1,50,50,0,0,0,2,3,0x0000390900000000000000000000000000000000,0,NULL,1);
-INSERT INTO `npc_list` VALUES (17740023,'Chacharoon','Chacharoon',152,-293.500,-12.020,-59.500,1,50,50,0,0,0,2,27,0x00005F0B00000000000000000000000000000000,32,NULL,1);
+INSERT INTO `npc_list` VALUES (17740020,'History_Moogle','Moogle',40,-259.440,-12.021,-79.538,1,40,40,0,0,0,0,4194307,0x0000520000000000000000000000000000000000,0,NULL,1);
```
claywar, line 27362: Can you confirm that this Moogle is _only_ used for this event? Else we're going to have potential conflicts in the future.  This might be something to just leave in the db for a record, and create dynamically

TeoTwawki: @claywar I can already tell you with 100% certainty that it WILL be used for multiple things. SE reuses them as well as shuffling them in and out. This is why I was pushing for dynamic entity moogles on the Easter even PR.

Same thing here, using a DE and adding in a `facePlayer()` before starting the event will make them look just like retail despite the NPC ID not matching.

zach2good:
If this is historically a timed event, should it be using the event system? (eggstravaganza, etc.)

scripts/globals/vanaversary.lua
neuromancerxi: Added checks and handlers to deal with charged items.

sql/npc_list.sql
neuromancerxi: Currently, I can't confirm or deny that this particular Moogle is unused by other events. It differs from, for example, the ID(s) being used by the Eggstravaganza Moogles. I am happy to monitor as seasonal events rotate in/out to check on it, however.

If there are other events discovered where it is evident they are using this same ID, could approach it by renaming the moogle based on the sequence they appear in the Zone (e.g. `Moogle_01`), add conditionals to the script, and add additional "disappear" handling.

neuromancerxi:
> Can't test this myself right now but does the coffer reset the cooldown for usable items when you take them? I know on retail when you take the behemoth suit +1 from the coffer the cooldown is reset so you can't immediately use it, drop it, then take a new one and repeat.

Good catch. Confirmed Retail behavior and adjusted to align.

> If this is historically a timed event, should it be using the event system? (eggstravaganza, etc.)

It is tied to the Adventurer Appreciation Campaign, which happens annually: 

https://www.bg-wiki.com/ffxi/Category:Seasonal_Events#Adventurer_Appreciation_Campaign

My initial take was to build this as a flagged event to give greater control to Server Operators, allowing them enable/disable at their own discretion (or perhaps the anniversary of their own server launch). 

If the overall preference is to adapt this to the Seasonal Event format, I'm open to investigating how to accomplish that.

neuromancerxi:
May I request some guidance for the style checks?

```
Multiple parameters used without an appropriate following space or newline: scripts/globals/items/chocobo_shield_+1.lua:21
local qty = math.random(1,11)                              <-- HERE
```
https://github.com/LandSandBoat/server/pull/3933/commits/dd057571aae9a97f998afd98454db264859250d8#diff-c4e14892f9f798d8fc6db11d9d82dedcec1ee937bb6bda2369a010c46c7e5c61

The style check seems to be asking me to add an additional new line, even though there is already one present? Would that mean I should have two new lines instead of one?

claywar:
Add a space after commas in parameter lists
`local qty = math.random(1,11)` <- Bad
`local qty = math.random(1, 11) ` <- Good

sql/npc_list.sql
TeoTwawki: > Currently, I can't confirm or deny that this particular Moogle is unused by other events. It differs from, for example, the ID(s) being used by the Eggstravaganza Moogles. I am happy to monitor as seasonal events rotate in/out to check on it, however.
> 
> If there are other events discovered where it is evident they are using this same ID, could approach it by renaming the moogle based on the sequence they appear in the Zone (e.g. `Moogle_01`), add conditionals to the script, and add additional "disappear" handling.

They shift these around and even change the total number of moogles regularly, but what they use the 1st/2nd/3rd/4th moogle for changes while still occupying the same relative position. **This is known** and I've watched those moogles shift during 6+ years of npc id updating and seeing mogs captured during events repeatedly. 

So we either start using DEs or consign ppl to never being able to enable these out of season and having to deal with the ID shifting and needing to make sure its updated (client update time) immediately  ***forever*** so people can actually start the event when SE does. Starting a pattern of just using a DE eliminates all of that maintenance cost and "just works".

neuromancerxi: Firstly, I appreciate all the feedback on the draft.  Secondly, regarding using the actual NPC ID vs a DE:

Whichever of these two methods is the stated preference by the project is the one I'll work on implementing. If the preference is Dynamic Entries, should this apply to all of the actors (Vanaversary Chest, Vanaversary Moogle, History Chest, Chacharoon, etc), or is the concern about the specific Moogle that has been observed to have been reused or ID Shifted?

TeoTwawki: its even more important for non moogles. in theory if we insisted on never using DEs we could at least have logic in a re-used moogle check which role he's supposed to be playing. But Chacharoon doesn't have that option when SE rips him out of the dat in a few months time.

I've just asked the rest of the team to voice any objection they have to using DE to evade all future maintenance cost - there are pretty much zero drawbacks to it other than one I already created a workaround for. We really need to get a good standard going on these things early, because when the next person decides to add a missing seasonal event they'll be looking at what is done here to guide them.

TeoTwawki: verdict: do the DE

scripts/globals/npc_util.lua
```diff
@@ -284,6 +284,41 @@ function npcUtil.giveItem(player, items, params)
     return true
 end
 
+--[[
+    For handling of 1/1 usable items delivered in a "just used" state. Most 1/1 charged items are Rare,
+    but some exceptions do exist. Pass a value of 1 in notRare for those exceptions.
+    e.g. npc_utils.giveUsedItem(player, 22132, 1) will grant the player an Ex Artemis Bow +1.
+--]]
+function npcUtil.giveUsedItem(player, item, notRare)
```
TeoTwawki, line 289: 👍

TeoTwawki: looks like that param should just be true/false rather than a number?

neuromancerxi: Functionally the same, but will swap this to a bool instead.

sql/npc_list.sql
neuromancerxi: > verdict: do the DE

Swapped back to draft while I work out moving this from fixed to DE.

neuromancerxi: Current draft supports Dynamic Entities with some caveats:

* Cut scenes do not properly animate Dynamic Entities (DE ID is not what the CS expects)
* Player ends up looking at an incorrect xyz due to the CS Enity ID differing from the DE.

> We really need to get a good standard going on these things early, because when the next person decides to add a missing seasonal event they'll be looking at what is done here to guide them.

Next major commit will rework back to Static Entries, but leaving the DE commit in place as a potential example.

TeoTwawki: we have means to turn both entities already. its a separate function call. We can even perform other animations as well. (NPC waving and what have you) This is not an impassable barrier.

I coudl modify the recently added `facePlayer()` to also turn the player and save doing an extra call if this is typical behavior.

neuromancerxi:
Any suggestions on how to work around the sanity checks for the undefined and shadowing warnings in this case? 

I'm guessing they are due to the way `scripts/globals/events/vanaversary` is iteratively generating the dynamic NPCs, which requires defining onTrigger, onEventUpdate, etc.

scripts/globals/events/vanaversary.lua
```diff
@@ -0,0 +1,488 @@
...
+                    x = x,
+                    y = y,
+                    z = z,
+                    rotation = rot,
+                    widescan = widescan,
+                    entityFlags = entityFlags,
+                    namevis = namevis,
+                    releaseIdOnDisappear = true,
+                    onTrigger = function(player, npc)
+                        player:timer(400, function(player)
```
Xaver-DaRed: To answer your question: change `player` with `playerArg` like so...
```lua
    player:timer(400, function(playerArg)
```

Change `playerArg` with whatever you think will fit

Xaver-DaRed:
I worte that backwards originally in my comment. I edited it immediately after, but I guess github doesnt update edits automatically or something.

```lua
    player:timer(400, function(playerArg)
```

scripts/globals/events/vanaversary.lua
```diff
@@ -0,0 +1,488 @@
...
+        xi.events.vanaversary.goods = xi.events.vanaversary.cofferItemsM
+    end
+
+    player:startEvent(csid)
+end
+
+xi.events.vanaversary.treaureCofferGoods = function(player, csid, option)
+    local choice = xi.events.vanaversary.goods[option]
+
+    if player:isChargedItem(choice) then
```
TeoTwawki, line 404: 👍

scripts/globals/events/vanaversary.lua
```diff
@@ -0,0 +1,488 @@
...
+        xi.events.vanaversary.generateEntities()
+    end
+
+    for _, entityID in pairs(xi.events.vanaversary.entities) do
+        local entity = GetNPCByID(entityID)
+        if entity then
+            if enabled then
+                entity:setStatus(xi.status.NORMAL)
+            else
+                -- TODO: Why doesn't DISAPPEAR work here?
```
TeoTwawki, line 59: My guess is the answer is packets. I think we broke something somewhere. Retain comment here till we fix it because it really should be what you tried to use 1st.

neuromancerxi:
Not sure what's up with the Linux Build failing on... MultiInstance conquest? Should I do a force push to try to kick it off again?

TeoTwawki:
> Not sure what's up with the Linux Build failing on... MultiInstance conquest? Should I do a force push to try to kick it off again?

no need, we can manually klick the hamster wheel and give him a jolt

src/map/packet_system.cpp
```diff
@@ -2289,6 +2291,24 @@ void SmallPacket0x04B(map_session_data_t* const PSession, CCharEntity* const PCh
         {
             login_message = login_message + (settings::get<std::string>("main.TRUST_ALTER_EGO_EXPO_MESSAGE"));
         }
+        if (settings::get<bool>("main.VANAVERSARY_ANNOUNCE"))
```
zach2good: This logic all lives in Lua now in server.lua, this'll need to be rebased and then changed to account for that 👍 

https://github.com/LandSandBoat/server/pull/4132

neuromancerxi: Thanks! A lot has changed in the last two months. Converting back to draft while I do some reworking to address this and a few other items.

PR #3936, opened 2023-05-27
[Magic] Adds missing logic to Resist States

Frankie-hz:
**_I affirm:_**

- Corrects logic on when level correction is supposed to apply. -- Mobs vs Player: There is no dLvl for this interaction -- Player vs Mobs: Mobs do not have a disadvantage when targeted. Only a bonus.

- Adds resist rank tier change for magic burst

- Corrects logic for final resist tier results.
If the spell is a damage spell and the player does not have subtle sorcery on then you should enter the logic for resist rate / 2. In addition when the EEM tier is 150% (-3) the resist rate will always be at least 0.5.

### DIFF
--- a/scripts/globals/combat/magic_hit_rate.lua
+++ b/scripts/globals/combat/magic_hit_rate.lua
@@ -203,6 +203,11 @@ xi.combat.magicHitRate.calculateActorMagicAccuracy = function(actor, target, spe
     magicAcc = magicAcc + utils.clamp(maccFood, 0, actor:getMod(xi.mod.FOOD_MACC_CAP))
 
     -----------------------------------
+    -- magicAcc from spell base.
+    -----------------------------------
+    magicAcc = magicAcc + bonusMacc
+
+    -----------------------------------
     -- magicAcc from Magic Burst
     -----------------------------------
     local _, skillchainCount = FormMagicBurst(spellElement, target)
@@ -236,8 +241,13 @@ xi.combat.magicHitRate.calculateTargetMagicEvasion = function(actor, target, spe
         magicEva = magicEva + target:getMod(mEvaMod) + target:getMod(xi.mod.STATUS_MEVA)
     end
 
-    -- Level correction. Target gets a bonus when higher level. Never a penalty.
-    if levelDiff > 0 then
+    -- Level correction.
+    -- Mobs vs Player: There is no dLvl for this interaction
+    -- Player vs Mobs: Mobs get a bonus when higher level. Never a penalty.
+    if
+        levelDiff > 0 and
+        not target:isPC()
+    then
         magicEva = magicEva + levelDiff * 4
     end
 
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
> Corrects logic for final resist tier results.
If the spell is a damage spell and the player does not have subtle sorcery on then you should enter the logic for resist rate / 2. In addition when the EEM tier is 150% (-3) the resist rate will always be at least 0.5.

-3 resist tier means you are weak to it. resist rate should definately not be 0.5 minimum

scripts/globals/combat/magic_hit_rate.lua
```diff
@@ -271,26 +281,32 @@ xi.combat.magicHitRate.calculateResistRate = function(actor, target, skillType,
         resistRank = target:getMod(elementTable[spellElement][4])
     end
 
+    -- Handles increasing res rank if a skillchain is present
+    local _, skillchainCount = FormMagicBurst(spellElement, target)
+
+    if skillchainCount > 0 then
+        resistRank = utils.clamp(resistRank - 1, -3, 11)
+    end
```
Xaver-DaRed, line 289: I dont think this is needed. Caster already gets a magic accuracy bonus in the macc step

Frankie-hz:
> > Corrects logic for final resist tier results.
> > If the spell is a damage spell and the player does not have subtle sorcery on then you should enter the logic for resist rate / 2. In addition when the EEM tier is 150% (-3) the resist rate will always be at least 0.5.
> 
> -3 resist tier means you are weak to it. resist rate should definately not be 0.5 minimum

If a mob has the lowest tier then you hit what you call a half resist guarantee. Since it’s the weakest tier it’s impossible to get anything lower than that. Same goes for the other direction. If a mob at tier 11 (5% eem) then you’ll never be able to land an enfeeble on it because your resist will never be 0.5 or above, even if you hit the 5% dice roll. This is only for the max/min tiers. I will link everything later when I get a chance.

Xaver-DaRed:
> > > Corrects logic for final resist tier results.
> > > If the spell is a damage spell and the player does not have subtle sorcery on then you should enter the logic for resist rate / 2. In addition when the EEM tier is 150% (-3) the resist rate will always be at least 0.5.
> > 
> > 
> > -3 resist tier means you are weak to it. resist rate should definately not be 0.5 minimum
> 
> If a mob has the lowest tier then you hit what you call a half resist guarantee. Since it’s the weakest tier it’s impossible to get anything lower than that. Same goes for the other direction. If a mob at tier 11 (5% eem) then you’ll never be able to land an enfeeble on it because your resist will never be 0.5 or above, even if you hit the 5% dice roll. This is only for the max/min tiers. I will link everything later when I get a chance.

EDIT: Ah, I see what you are doing.

I'm not liking the general organization though. nothing that cant be solved, but lets remember this is meant to be as easily read as possible.

I'll wait for the finished product though. It was just a comment for the time being

Frankie-hz:
> > > > Corrects logic for final resist tier results.
> > > > If the spell is a damage spell and the player does not have subtle sorcery on then you should enter the logic for resist rate / 2. In addition when the EEM tier is 150% (-3) the resist rate will always be at least 0.5.
> > > 
> > > 
> > > -3 resist tier means you are weak to it. resist rate should definately not be 0.5 minimum
> > 
> > 
> > If a mob has the lowest tier then you hit what you call a half resist guarantee. Since it’s the weakest tier it’s impossible to get anything lower than that. Same goes for the other direction. If a mob at tier 11 (5% eem) then you’ll never be able to land an enfeeble on it because your resist will never be 0.5 or above, even if you hit the 5% dice roll. This is only for the max/min tiers. I will link everything later when I get a chance.
> 
> EDIT: Ah, I see what you are doing.
> 
> I'm not liking the general organization though. nothing that cant be solved, but lets remember this is meant to be as easily read as possible.
> 
> I'll wait for the finished product though. It was just a comment for the time being

But thats why I have you! 😬😬

Xaver-DaRed:
ty for this PR Frank. It was very helpfull

PR #3938, opened 2023-05-27
Port Eden's Spawn Slot implementation

zach2good:
**_I affirm:_**
- [not yet] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

- Eden has a series of PRs implementing shared spawn slots that I've been promising to port. I've gone through most of them (and Eden's current live files) to start getting things into LSB. They have a lot of systems that aren't compatible with upstream that need to be jiggled around to get to work.
- TODO

- TODO

## TODO

- Port all the slot data Eden has so far, it's all in yaml so I'll need to write a little script to poop it into SQL where we want it (with comments, etc.)

### DIFF
--- a/sql/mob_spawn_slots.sql
+++ b/sql/mob_spawn_slots.sql
@@ -1,0 +1,20 @@
+--
+-- Table structure for table `mob_spawn_slots`
+--
+
+/* mob spawn slots */
+DROP TABLE IF EXISTS `mob_spawn_slots`;
+CREATE TABLE `mob_spawn_slots` (
+  `spawnslotid` int(10) unsigned NOT NULL,
+  `zoneid` smallint(3) unsigned NOT NULL DEFAULT 0,
+  `chance` tinyint(4) DEFAULT '0',
+  `maxspawns` tinyint(4) DEFAULT '0',
+  PRIMARY KEY (`spawnslotid`, `zoneid`)
+) ENGINE=Aria TRANSACTIONAL=0 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
+
+LOCK TABLES `mob_spawn_slots` WRITE;
+
+-- Carpenter's Landing (Zone 2)
+INSERT INTO `mob_spawn_slots` VALUES (1,2,100,1);
+
+UNLOCK TABLES;
--- a/sql/mob_spawn_slots.sql
+++ b/sql/mob_spawn_slots.sql
@@ -8,13 +8,12 @@ CREATE TABLE `mob_spawn_slots` (
   `spawnslotid` int(10) unsigned NOT NULL,
   `zoneid` smallint(3) unsigned NOT NULL DEFAULT 0,
   `chance` tinyint(4) DEFAULT '0',
-  `maxspawns` tinyint(4) DEFAULT '0',
   PRIMARY KEY (`spawnslotid`, `zoneid`)
 ) ENGINE=Aria TRANSACTIONAL=0 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
 
 LOCK TABLES `mob_spawn_slots` WRITE;
 
 -- Carpenter's Landing (Zone 2)
-INSERT INTO `mob_spawn_slots` VALUES (1,2,100,1);
+INSERT INTO `mob_spawn_slots` VALUES (1,2,100);
 
 UNLOCK TABLES;
--- a/sql/mob_spawn_slots.sql
+++ b/sql/mob_spawn_slots.sql
@@ -15,5 +15,7 @@ LOCK TABLES `mob_spawn_slots` WRITE;
 
 -- Carpenter's Landing (Zone 2)
 INSERT INTO `mob_spawn_slots` VALUES (1,2,100);
+INSERT INTO `mob_spawn_slots` VALUES (2,2,100);
+INSERT INTO `mob_spawn_slots` VALUES (3,2,100);
 
 UNLOCK TABLES;
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,13 +19,17 @@ jobs:
   Sanity_Checks:
     runs-on: ubuntu-22.04
     steps:
-    # workaround for broken clang on ubuntu runner until fixed: https://github.com/actions/runner-images/issues/8659
-    - uses: claywar/workaround8649@b59699cec47ef4f0db8cf5564e0dc30a74cc8888
-      with:
-        os: ubuntu-22.04
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
+      # Workaround for incompatible clang and libstc++ versions, see
+      # <https://github.com/actions/runner-images/issues/8659> for more info.
+      # The basic idea here is to remove g++ 13 and its C++ standard library
+      # and replace that with a compatible version of libstdc++.
+    - name: Workaround for GHA runner images issue 8659
+      run: |
+        sudo apt-get purge -y g++-13 gcc-13 libstdc++-13-dev
+        sudo apt-get install -y --allow-downgrades libstdc++-12-dev libstdc++6=12.* libgcc-s1=12.*
     - name: Install Dependencies
       run: |
         # https://mariadb.com/docs/connect/programming-languages/c/install/#connector-c-install-repo-configure-cs
@@ -197,13 +201,17 @@ jobs:
     needs: Sanity_Checks
     runs-on: ubuntu-22.04
     steps:
-    # workaround for broken clang on ubuntu runner until fixed: https://github.com/actions/runner-images/issues/8659
-    - uses: claywar/workaround8649@b59699cec47ef4f0db8cf5564e0dc30a74cc8888
-      with:
-        os: ubuntu-22.04
     - uses: actions/checkout@v3
       with:
         fetch-depth: 1
+      # Workaround for incompatible clang and libstc++ versions, see
+      # <https://github.com/actions/runner-images/issues/8659> for more info.
+      # The basic idea here is to remove g++ 13 and its C++ standard library
+      # and replace that with a compatible version of libstdc++.
+    - name: Workaround for GHA runner images issue 8659
+      run: |
+        sudo apt-get purge -y g++-13 gcc-13 libstdc++-13-dev
+        sudo apt-get install -y --allow-downgrades libstdc++-12-dev libstdc++6=12.* libgcc-s1=12.*
     - name: Install Dependencies
       run: |
         sudo apt-get update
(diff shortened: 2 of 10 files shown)


sql/mob_spawn_slots.sql
```diff
@@ -0,0 +1,16 @@
+--
+-- Table structure for table `mob_spawn_slots`
+--
+
+/* mob spawn slots */
+DROP TABLE IF EXISTS `mob_spawn_slots`;
+CREATE TABLE `mob_spawn_slots` (
+  `id` int(10) unsigned NOT NULL,
+  `mobid` int(10) unsigned NOT NULL,
```
TeoTwawki: reminder: per prior discussions in discord, we really need to tie it all up without relying on mobid being in 2 tables when this is done with. it'll mean having the "slot" id available somewhere else to tie them together. That way when mobIDs change we don't have to change it in multiple places.

zach2good:
Current status of this: We don't want to introduce more things that are keyed onto the global IDs of specific mobs, since they can (and do, rarely) shift around!

Needs to find another mechanism that is a) easy to use, and b) is resistant to shifts

TeoTwawki:
> Current status of this: We don't want to introduce more things that are keyed onto the global IDs of specific mobs, since they can (and do, rarely) shift around!
> 
> Needs to find another mechanism that is a) easy to use, and b) is resistant to shifts

The quick fix is to call the "id" in this table "slotID" and then also have that same slotID in one of existing mob tables to match them on. Keeping the current structure of this, you'd add slotID to the spawn points table. We don't need to be perfect this on the 1st pass, we'll be tearing up mob tables to reorganize eventually anyway and at that future time we can flesh this out to being more than random mobid selection and use true shared spawn nodes.

just changes the querying a little.

TeoTwawki:
just changes the querying a little.

Frankie-hz:
We did this on Horizon and was in the process of moving it to ASB when we saw LSB was going to use a different implementation. Once this gets resolved I can port all the spawn point information we did with retail captures and how we came to each spawnset section.

We added "spawnset" to the spawn point table to identify which mobs were in which spawnset.

https://github.com/AirSkyBoat/AirSkyBoat/blob/staging/modules/custom/sql/mob_zones/02_cop/02_wilderness/002_carpenters_landing.sql

Here is an example: If you look at how the spawnsets are grouped together - thats almost the same pattern for every single zone that has spawnsets. I am not sure if that helps shed some light on things and how to proceed. Every "beastmen" group is always the same "section/group" of mobs.
image

image

image

TeoTwawki:
@Frankie-hz that looks very similar to what I was suggesting to Zach, but I called it slotID instead of spawnSet - then we'd use that same number in the other table instead of the mob ID. Basically numbering the spawns and using that as the basis for the relation between 2 tables because that won't shift around every update.

Frankie-hz:
Yeah is basically the exact same thing as you said. I was trying to give you guys an idea how how most spawn sets are actually structured in zones to maybe see if there was a different way that would pop up instead.

We also added a zone column because we didnt want to structure it where if we had to add another slotid to say "Zone 2" (Using carp landing example on top) we could just add slotID 7 instead of adding it to the end of the table.

Frankie-hz:
@zach2good Let me know if there is anything I can do to help move this along. I would love to get this into ASB and get rid of all our modules for this <3

zach2good:
We're still deciding what we want slots to represent:
- One slot represents a chance for a single mob to be spawned from a list (Eden)
- One slot represents a series of chances for multiple mobs to be spawned up to a certain limit - from a list (ASB)

Frankie-hz:
I am gonna leave this here as a "how the system works". This is what we observed during retail testing. I will try to remember everything but I will make edits if I realized I left something out. Maybe that will help make a determination on which system to use.

- If there is a group of 3 Ord IDs next to each other and only **one** is allowed to spawn -> when a mob from that group dies it can choose one of those three mobs (including itself) to spawn. At no point will there be more than 1 orc up at a time. 
- When the zone resets (verified during maintenance) not always the same mobs will be up. It seems to random what mobs are re-spawned during a reset but the number of mobs in that area (mobs per group) have a limit on how many can be up at at time. 
- Most night time mobs share spawn slots with mobs that spawn during the day. Example: When night time hits and all the mobs are alive there will be no ghouls inside the dunes. You must kill Goblins DURING night time in order to spawn ghouls.

Building on night time: Zitah for instance has hounds that spawn from courels. But not all of the hounds spawn from the same group. Some hounds are separate of that group and will always spawn when it ticks nighttime and then will continue to respawn the same ID when killed (Basically saying its not inside that group)

- Elementals in certain zones are also grouped with each other and will spawn on certain areas of the map when the other is killed. (Observed to be always 100% chance it would spawn the other elemental, NOT the one you just killed)
- Beastmen campfires are zone-wide pools. This means if you kill a goblin in a campfire on one side of the map, that goblin will NEVER spawn again until other campfire goblins are killed. I observed this in Meri Mountains when goblins were not spawning for 4 days straight and only started to re-spawn when other campfires (of the same TIER) were killed. This was also observed in ronfaure.

TeoTwawki:
> You must kill Goblins DURING night time in order to spawn ghouls.

More specifically a respawn has to happen during night so it can do the "dice roll". You can kill'em just before as long as the next spawn happens at night.

Some good example cases to prove how things work are right in Ghelsba. Including cases where ONE of the shared spawns is the PH to an NM - which can make it seem like that Nm has a much higher cooldown or much lower pop rate than it actually does if the specific mob that is the PH doesn't get chosen (the NM isn't tied to the whole group, just that one mob!)

tl;dr Teo is corroborating what Frank just said.

zach2good:
db changes added with:
```sql
ALTER TABLE xidb.mob_spawn_points
ADD COLUMN `spawnslotid` int(10) unsigned NOT NULL DEFAULT '0' AFTER `groupid`;
```

and dbtool -> tasks -> dump table -> mob_spawn_points

zach2good:
SQL query to populate testing groups (so we don't have to edit anything by hand in this PR):
```sql

UPDATE xidb.mob_spawn_points SET spawnslotid = 1 WHERE mobid = 16785440;
UPDATE xidb.mob_spawn_points SET spawnslotid = 1 WHERE mobid = 16785441;
UPDATE xidb.mob_spawn_points SET spawnslotid = 1 WHERE mobid = 16785442;

UPDATE xidb.mob_spawn_points SET spawnslotid = 2 WHERE mobid = 16785446;
UPDATE xidb.mob_spawn_points SET spawnslotid = 2 WHERE mobid = 16785447;
UPDATE xidb.mob_spawn_points SET spawnslotid = 2 WHERE mobid = 16785448;

UPDATE xidb.mob_spawn_points SET spawnslotid = 3 WHERE mobid = 16785457;
UPDATE xidb.mob_spawn_points SET spawnslotid = 3 WHERE mobid = 16785458;
UPDATE xidb.mob_spawn_points SET spawnslotid = 3 WHERE mobid = 16785459;

```

zach2good:
I guess the only point of contention is: `AFTER `groupid`;`, where should we put the new column? I've put it after the grouped, open to other ideas. The mobid, name and packetname line up nicely with whats in the dats, so I wanted to leave them laid out as-is

TeoTwawki:
> I guess the only point of contention is: `AFTER `groupid`;`, where should we put the new column? I've put it after the grouped, open to other ideas. The mobid, name and packetname line up nicely with whats in the dats, so I wanted to leave them laid out as-is

I'd put it right at the end of the pos data, after `rot`. It's also fine right before pos data which is where you look to have it right now.

Frankie-hz:
With this implementation does every mob need a group ID and all the % chances need to add up to 100? I am following how this chance thing exactly works.

zach2good:
I have no idea how it works, i just pulled it in as-is, here's the relevant code:
```cpp
bool SpawnSlot::TrySpawn()
{
    // Determine which of the mobs in the group can be spawned, and if there's one spawned already
    std::vector<std::tuple<uint32, CMobEntity*>> chanceSpawns;
    std::vector<CMobEntity*>                     remainingSpawns;

    CMobEntity* allowedSpawn = nullptr;

    uint32 totalChance = 0;

    for (auto&& entry : entries)
    {
        if (entry.mob->isAlive())
        {
            allowedSpawn = entry.mob;
            break;
        }

        if (!entry.mob->m_CanSpawn)
        {
            continue;
        }

        if (entry.spawnChance > 0)
        {
            totalChance += entry.spawnChance;
            chanceSpawns.push_back({ entry.spawnChance, entry.mob });
        }
        else
        {
            remainingSpawns.push_back(entry.mob);
        }
    }

    // Don't spawn if there's another mob in this slot already spawned.
    if (allowedSpawn)
    {
        return false;
    }

    // If there are no remaining spawns available, bail out
    if (remainingSpawns.empty())
    {
        return false;
    }

    // Check for chance spawns
    if (totalChance > 0)
    {
        uint32 roll = xirand::GetRandomNumber(100);

        // Check if roll is low enough number to make one of the chance mobs to spawn.
        if (roll < totalChance)
        {
            // Find the chance spawn which matches the roll
            uint32 accumulatedRoll = 0;

            bool spawned = false;
            for (auto&& entry : chanceSpawns)
            {
                auto mob = std::get<1>(entry);
                accumulatedRoll += std::get<0>(entry);
                if (!spawned && roll < accumulatedRoll)
                {
                    allowedSpawn = mob;
                    break;
                }
            }
        }
    }

    if (!allowedSpawn)
    {
        if (!remainingSpawns.empty())
        {
            // Pick a random mob from the non-chance spawns
            allowedSpawn = xirand::GetRandomElement(remainingSpawns);
        }
        else
        {
            return false;
        }
    }

    // Finally spawn the mob
    for (auto&& entry : entries)
    {
        if (entry.mob == allowedSpawn)
        {
            entry.mob->m_AllowRespawn = true;
            entry.mob->Spawn();
        }
        else
        {
            entry.mob->m_AllowRespawn = false;
        }
    }

    return allowedSpawn != nullptr;
}
```

Frankie-hz:
Update on this: Eden doesn't actually use the chance variable yet. In the event that there are weighted chances to spawn they added it as a precaution in case it was found this was true. As of right now we haven't confirmed that is an actual thing but other than that the implementation looks to be in-line with what we have found as well.

We added the slotid after the mobID so you can quickly tell which spawnslot things are in without having to think very much. I don't have a preference though either way.

zach2good:
Now that I've completely forgotten all context as to what was going on in this PR and what I was opposing: if you're following along with what Eden is doing - we're probably good to take it too. If we end up being wrong, then we _both_ have a lot of fixing to do :D

Frankie-hz:
> Now that I've completely forgotten all context as to what was going on in this PR and what I was opposing: if you're following along with what Eden is doing - we're probably good to take it too. If we end up being wrong, then we _both_ have a lot of fixing to do :D

I can live with that. I will get around to testing this when I get a minute to breathe. We more or less do it the same way.

MowFord:
This PR has interested me for some time and I'd like to assist in whatever way I can in getting it over the finish line. In its current form, It seems to perfectly solve the issue. I assume it's just a matter of testing? I'd be willing to add a git patch to our codebase as we continue ramping up building modules and testing things. I'd hate to spend too much time on it if the implementation plan shifts a ton, though.

Just to make sure i'm understanding it:
- `spawnslotid` is unique per zone, similar to a mob's groupid
- creating a group is simply a matter of setting non-zero `spawnslotid` as well as defining that `spawnslotid`/`zoneid` pair in the `mob_spawn_slots` table

Only concern about pulling this patch in would be
- the changes to the github actions yaml file. I assume those incompatibilities are no longer relevant to this PR?
- the merge conflict, which github isn't making obvious, but i assume is related to actions or deletion of the wiki submodule

Xaver-DaRed:
Not needed anymore.

PR #3956, opened 2023-06-01
Split status.lua into multiple files, preload enums before other globals

claywar:
**_I affirm:_**

* Creates scripts/enum directory, and splits status.lua into one file per table
* Begins to remove deprecated references to both status and settings files

Note: scripts/enum is now loaded after settings, but prior to all other globals.  There should not be any required direct require to those files.

No changes should be observed


scripts/enum/pathflag.lua
```diff
@@ -0,0 +1,11 @@
+xi = xi or {}
+
+xi.pathflag =
+{
+    NONE     = 0x00,
+    RUN      = 0x01, -- run twice the speed
+    WALLHACK = 0x02, -- run through walls if path is too long
+    REVERSE  = 0x04, -- reverse the path
+    SCRIPT   = 0x08, -- don't overwrite this path before completion (except via another script)
+    SLIDE    = 0x10,  -- Slide to end point if close enough (so no over shoot)
```
TeoTwawki: comment misaligned

TeoTwawki (approved):

PR #3960, opened 2023-06-03
[chore] Remove remaining status global requires

claywar:
**_I affirm:_**

* Removes remaining requires for scripts/globals/status
* Adds CI check to not allow status, settings, or enum requires

No changes should be observed


Xaver-DaRed (approved):
Sublime, sir

scripts/globals/items/jug_of_marys_milk.lua
```diff
@@ -12,7 +12,11 @@ itemObject.onItemCheck = function(target)
 end
 
 itemObject.onItemUse = function(target)
-    if not hasSleepEffects(target) then
+    if
+        not target:hasStatusEffect(xi.effect.SLEEP_I) and
+        not target:hasStatusEffect(xi.effect.SLEEP_II) and
```
TeoTwawki, line 17: 🤔 but you can't use item if you are asleep anyway, can you target others with this the whole time and I never knew?

incidentally, the item was always wrong anyway: its basically diablos nightmare sleep.
https://ffxiclopedia.fandom.com/wiki/Mary%27s_Milk

I'll make us an issue.

PR #3968, opened 2023-06-05
[Chore] Cleanup "interaction" battlefields

Xaver-DaRed:
**_I affirm:_**

Cleanup battlefield.lua, limbus.lua and battlefields coded under Morris's system

Only real change is a slight lowering in 1 metal chip drop rate to fall under SE's %rates.
No real changes were made.

### DIFF
--- a/scripts/globals/battlefield.lua
+++ b/scripts/globals/battlefield.lua
@@ -1507,17 +1507,3 @@ function xi.battlefield.HandleLootRolls(battlefield, lootTable, players, npc)
         battlefield:setLocalVar("lootSeen", 1)
     end
 end
-
--- TODO(jmcmorris): This is no longer needed once Limbus uses the new Battlefield system
-function xi.battlefield.HealPlayers(battlefield, players)
-    players = players or battlefield:getPlayers()
-    for _, player in pairs(players) do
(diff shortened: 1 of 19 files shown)


claywar (approved):

scripts/globals/battlefield.lua
```diff
@@ -959,10 +983,10 @@ function Battlefield:onBattlefieldEnter(player, battlefield)
         self.requiredItems.wearMessage and
         player:hasItem(self.requiredItems[1])
     then
-        local itemId = self.requiredItems[1]
-        local uses = player:incrementItemWear(itemId)
-        -- Gets the total number of item uses for the given item. Default to one since that is the majority of them.
-        local totalUses = xi.battlefield.itemUses[itemId] or 1
+        local itemId    = self.requiredItems[1]
+        local uses      = player:incrementItemWear(itemId)
+        local totalUses = xi.battlefield.itemUses[itemId] or 1 -- Gets the total number of item uses for the given item. Default to one since that is the majority of them.
```
TeoTwawki: this makes the line 163 chars wide, which was hopefully why they put the comment above instead of inline. I do the same thing when they get to wide (as defined by making me right/left scroll side-by-side diff)

TeoTwawki (approved):

PR #3975, opened 2023-06-08
[Crafting] Cleanup GP Vendors (Union Representatives)

Xaver-DaRed:
**_I affirm:_**

Cleans union representative NPCs via the following changes:
- Renames their functions
- Moves their inventories to crafting.lua
- Renames the variables used in their functions.

Additionally, this also has the following changes in the fishing guild:
- Minor correction in the required rank for an item.
- Fixes items selected not matching items given.

A note on Guild IDs: We use 2 different guild id systems. This NPCs use the correct order used by retail. However, the order we have defined doesn't match it and I don't see why this order was chosen, since it also doesn't follow the skill ID order.

Once I get to cleaning guild masters, I will clean those up, but, for the time being, that's out of the scope of this PR.
BUT it WILL get done.

Review by commit for clarity (maybe)
It should be functionally the same. Except for the fishing guild, which should work properly now.

### DIFF
--- a/scripts/globals/crafting.lua
+++ b/scripts/globals/crafting.lua
@@ -420,20 +420,20 @@ end
 -- Guild Point NPCs (Union Representatives)
 --------------------------------------------------
 xi.crafting.guildPointNPConTrade = function(player, npc, trade, csid, guildId)
-    local _, remainingPoints = player:getCurrentGPItem(guildId)
     local ID                 = zones[player:getZoneID()]
+    local _, remainingPoints = player:getCurrentGPItem(guildId)
 
     if player:getCharVar('[GUILD]currentGuild') - 1 == guildId then
         if remainingPoints == 0 then
             player:messageText(npc, ID.text.NO_MORE_GP_ELIGIBLE)
         else
             local totalPoints = 0
-            for i = 0, 8 do
-                local items, points = player:addGuildPoints(guildId, i)
+            for tradeSlot = 0, 8 do
+                local items, points = player:addGuildPoints(guildId, tradeSlot)
 
                 if items ~= 0 and points ~= 0 then
                     totalPoints = totalPoints + points
-                    trade:confirmSlot(i, items)
+                    trade:confirmSlot(tradeSlot, items)
                 end
             end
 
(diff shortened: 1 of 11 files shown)


Xaver-DaRed:
Tested it again and screenshoted the result.
imagen

scripts/globals/crafting.lua
```diff
@@ -119,7 +321,7 @@ end
 -- tradeTestItem Action
 -----------------------------------
 xi.crafting.tradeTestItem = function(player, npc, trade, craftID)
-    local guildID    = craftID - 48
+    local guildId    = craftID - 48
```
TeoTwawki, line 322: I know we normally fullLowerCamel, but I think an exception can be made for `ID` given its already short for "**Id**entification"

I really hate seeing `Id` anywhere. Just triggers the heck outa my OCD.

Xaver-DaRed: We were using both `ID` and `Id` in this file. I have no real preference.

Xaver-DaRed: I have a new, separate PR half ready touching this file (Guild Masters in this case)

If you guys want `ID` I promise to change it in that one.

Xaver-DaRed:
Rebased to resolve conflicts

scripts/globals/crafting.lua
```diff
@@ -244,74 +446,79 @@ xi.crafting.unionRepresentativeTrade = function(player, npc, trade, csid, guildI
     end
 end
 
-xi.crafting.unionRepresentativeTrigger = function(player, guildId, csid, currency, keyitems)
+xi.crafting.guildPointNPConTrigger = function(player, csid, guildId)
```
claywar: I'd prefer if we're calling this onTrigger to pass the two expected parameters (player, npc) to the function, and doing a lookup from a table in this script to determine what the guildId and other values are.

claywar: (Also, I can't help but read this as NP-Con)

claywar (approved):
Definitely an improvement overall, but I'd like us to move the passed values from the NPC scripts to a lookup in the global eventually.

scripts/globals/crafting.lua
Xaver-DaRed: I can change the parameters in the next PR, once I actually know all guild-related stuff that can be tabled.
The consensus back then was that "Union Representative" was not really known/clear.
Would you preffer `guildPointOnTrigger`?

claywar: I can get on board with this

PR #3979, opened 2023-06-09
[core] Fix a crash with a corner case in CAlliance::delParty

WinterSolstice8:
**_I affirm:_**

This PR fixes a crash that would involve the 0th member of a party having a zonewide pool and other members not being in the zone. It also improves some logic with various mixtures of being in/out of a zonewide treasure pool.
This also fixes a broken up alliance able to get the same treasure pool in 2 different zones.

Have an alliance constructed as the following:
Alliance lead in a 2+ person party
Second party with 2 members, the party leader in a zone with a zonewide pool
The alliance leader will issue `/acmd breakup` which will no longer cause a crash.

Have 2 members in two different zones with mobs able to drop stuff, break up the alliance as indicated above, and not be able to share loot between two zones.

### DIFF
--- a/src/map/alliance.cpp
+++ b/src/map/alliance.cpp
@@ -170,6 +170,12 @@ void CAlliance::removeParty(CParty* party)
 
 void CAlliance::delParty(CParty* party)
 {
+    // Don't delete parties when there's no party in the alliance
+    if (!party->m_PAlliance || party->m_PAlliance->partyList.size() == 0)
+    {
+        return;
+    }
+
     // Delete the party from the alliance list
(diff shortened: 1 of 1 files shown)


TeoTwawki (dismissed):

WinterSolstice8:
Drafting because fixing this one bug discovered another.

WinterSolstice8:
Fixed the issue, undrafted.

claywar (approved):

Xaver-DaRed (approved):

zach2good (approved):

PR #3992, opened 2023-06-13
[Quests] Add module for missable quest "Mom, the adventurer?"

Xaver-DaRed:
**_I affirm:_**

It sometimes happens that quests are missable in retail.
I don't know of any other case other than "Mom, the adventurer?", but it happens.
This PR does:
- Makes "Mom, the adventurer?" unmissable by default, as discussed.
- Adds a module to restore it's missability.

Get fame lvl 3+ in bastok, enable the setting and be able to activate this quest.
Activate module and be able to miss it.

### DIFF
--- a/modules/era/lua/missable_mom_the_adventurer.lua
+++ b/modules/era/lua/missable_mom_the_adventurer.lua
@@ -1,0 +1,20 @@
+-----------------------------------
+-- Module to make "Mom the Adventurer?" quest missable.
+-----------------------------------
+require("modules/module_utils")
+-----------------------------------
+local m = Module:new("missable_mom_the_adventurer")
+
+m:addOverride('xi.server.onServerStart', function()
+    super()
+
+    xi.module.modifyInteractionEntry('scripts/quests/bastok/Mom_the_Adventurer', function(quest)
+        quest.sections[1].check = function(player, status, vars)
+            return status ~= QUEST_ACCEPTED and
+                player:getFameLevel(xi.quest.fame_area.BASTOK) < 2 and
+                vars.Prog == 0
+        end
+    end)
+end)
+
+return m
--- a/scripts/quests/bastok/Mom_the_Adventurer.lua
+++ b/scripts/quests/bastok/Mom_the_Adventurer.lua
@@ -37,7 +37,6 @@ quest.sections =
     {
         check = function(player, status, vars)
             return status ~= QUEST_ACCEPTED and
-                player:getFameLevel(xi.quest.fame_area.BASTOK) < 2 and
                 vars.Prog == 0
         end,
 


TeoTwawki:
I thought last time this came up we decided to be "better than retail" and make this one not missable?

CatsEyeXI:
> I thought last time this came up we decided to be "better than retail" and make this one not missable?

i was literally just talking about this with Xaver, seems like we could do better than retail here, not sure if a setting is even necessary.

TeoTwawki:
> i was literally just talking about this with Xaver, seems like we could do better than retail here, not sure if a setting is even necessary.

yeah I've surprised to see the check he's fixing is in the code base presently, thought we nuked it ages ago.

..would like SE to fix it on retail I've several mules that missed it.

Xaver-DaRed:
I decided to go both ways.
- Have it retail accurate
- Allow for it to not be missed.

The setting is just in case there is another obscure quest out there that is missable aswell

TeoTwawki:
I kinda suspect SE meant for it to be repeatable after the 2nd quest in the chain and then changed their mind making it bugged (because it is repeatable until you get enough fame for 2nd quest, and normally all quest are either repeatable or not instead of changing state like this one - remember they don't have a script engine like we do)

jesserhodes:
I love this change. I'd also be in favor of making it better than retail.

TeoTwawki:
Been thinking on this, and I think it makes more sense to have this default non-missable and do the alteration allow it to be missed as a module to instead of a setting. 

Could do one for repeatability too - one could debate if SE intended it to be repeatable forever or not, depending on which way or if its considered a bug (I wonder if anyone has ever reported it on their bug forum to see if SE thinks this is intended..)

I know you thought setting because other quests could use it, but no other quests have this issue on retail at all in the 20 years the game has been running. Just my 2 cents.

Xaver-DaRed:
I don't think we can modularze a quest that uses this system. I may be wrong though

zach2good:
Now that I've added this: https://github.com/LandSandBoat/server/commit/ea34e47ffabee824439c46c9b3ba27cd569bb448

This change could become:
```lua
m:addOverride('xi.server.onServerStart', function()
    super()

    xi.module.modifyInteractionEntry('scripts/quests/bastok/Mom_the_Adventurer', function(quest)
        local DISALLOW_MISSABLE_QUESTS = Whatever value you needed?

        -- REMEMBER: Lua is 1-indexed!
        quest.sections[1].check = function(player, status, vars)
            return status ~= QUEST_ACCEPTED and
                (player:getFameLevel(xi.quest.fame_area.BASTOK) < 2 or DISALLOW_MISSABLE_QUESTS) and
                vars.Prog == 0
        end
    end)
end)
```

TeoTwawki (approved):
assuming tested current version, looks good to me

zach2good (approved):

PR #3998, opened 2023-06-16
[Proposal] Track experience lost on death, instead of level

Xaver-DaRed:
**_I affirm:_**

- Changes what is being tracked on death and raise, from `Level` to `Experience Lost`
- Fixes char variable not being deleted on certain cases.
- Fixes a few cases of incorrect exp being returned because of lvl cap

NOTE: It needs to be a character variable, because we need it to persist through "zoning", either via Tractor or via CS at the end of a battlefield (CoP 1-3 final battlefied, for example).

NOTE: This PR doesn't really introduce a new character variable.. It insteads replaces the use of an existing one, renames it and makes sure it's deleted when it should so it doesnt stay in the database forever.
Server operators should find any use of `DeathLevel` Character variable in their databases and delete it.

- Die and raise. Recover correct amount of exp.
- Die. Get tractored and then raised. Recover correct amount of exp.

For extra fun
- Die while lvl capped by battlefield. Let it kick you out and raise. Recover the correct amount of exp.

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1919,6 +1919,7 @@ void CCharEntity::HandleErrorMessage(std::unique_ptr<CBasicPacket>& msg)
 void CCharEntity::OnDeathTimer()
 {
     TracyZoneScoped;
+    charutils::SetCharVar(this, "expLost", 0);
     charutils::HomePoint(this);
 }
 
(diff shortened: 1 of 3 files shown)


zach2good (approved):
Makes sense to me 👍

TeoTwawki (approved):

PR #3999, opened 2023-06-16
Custom SQL Module - Safe Outposts

CatsEyeXI:
**_I affirm:_**

Moves notoriously dangerous mobs which have been known to gank players away from outposts.

Either include `modules/custom/sql` in your `modules/init.txt` and run DBTool or
Execute SQL script / run DB tool and restart map server

### DIFF
--- a/modules/custom/sql/safe_outposts.sql
+++ b/modules/custom/sql/safe_outposts.sql
@@ -1,16 +1,27 @@
 -- Relocates dangerous monsters away from outposts
-UPDATE mob_spawn_points SET pos_x =  210.481, pos_y =  -6.946, pos_z =  101.274 WHERE mobname = 'Goblin_Leecher' AND pos_x = 184.673 AND pos_y = -7.710 AND pos_z = 106.691;    -- Valkurm Dunes, Goblin Leecher
-UPDATE mob_spawn_points SET pos_x = -396.957, pos_y = -32.261, pos_z =   44.893 WHERE mobname = 'Goblin_Shepherd' AND pos_x = -166.429 AND pos_y = 7.910 AND pos_z = -105.478;  -- Buburimu Peninsula, Goblin Ambusher
-UPDATE mob_spawn_points SET pos_x = -411.430, pos_y = -31.429, pos_z =    3.221 WHERE mobname = 'Goblin_Bandit' AND pos_x = -134 AND pos_y = 7 AND pos_z = -96;                 -- Buburimu Peninsula, Goblin Bandit
-UPDATE mob_spawn_points SET pos_x =  -75.189, pos_y =  -1.413, pos_z = -130.057 WHERE mobname = 'Goblin_Mercenary' AND pos_x = -139.706 AND pos_y = 7.167 AND pos_z = -61.223;  -- Cape Terrigan, Goblin Shepard
-UPDATE mob_spawn_points SET pos_x =   -7.247, pos_y =   0.161, pos_z =    0.709 WHERE mobname = 'Goblin_Mercenary' AND pos_x = -144 AND pos_y = 8 AND pos_z = -21;              -- Cape Terrigan, Goblin Bandit
-UPDATE mob_spawn_points SET pos_x =  -61.728, pos_y =  -1.335, pos_z =  -99.354 WHERE mobname = 'Goblin_Bandit' AND pos_x = -154.915 AND pos_y = 7.554 AND pos_z = -38.958;     -- Cape Terrigan, Goblin Mercenary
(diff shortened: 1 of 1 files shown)


TeoTwawki:
But its not a true retail experience till the goblins in yuhtunga kill you before the screen even fades from black! /s

True story, 1st time I ever outpost warped on retail that's what happened. then it happened again and road the ship sulking the whole way.

TeoTwawki (dismissed):
`WHERE mobid = 17199305` you'll need different conditions than the ID, because they occasionally DO change (just far less often than npc ones)

I'm not sure the optimal combination of things to check but you'd likely need a join or possibly just do math to get zone id (that part of the id won't change) and then you could check for "this name in this zone at this original pos"

zach2good:
Without digging in too deeply: one way to handle this would be to do a subquery first to find mobs who are less than a certain distance from a point (`SELECT * FROM mob_spawn_points  WHERE ...`, etc.) so you can specify the center of the output, and a distance.

Given that this'll probably require some maths and navmesh validation, doing this all in SQL isn't really possible. 

If you're then not doing this in SQL, you could probably do this quite happily in a C++ module using OnInit():

https://github.com/LandSandBoat/server/blob/54db671a076ee58509e1fcf4829834027d67c3ce/src/map/utils/moduleutils.h#L52

https://github.com/LandSandBoat/server/blob/54db671a076ee58509e1fcf4829834027d67c3ce/src/map/map.cpp#L296

We have something in the navmesh for this (I think, I haven't used it yet):

https://github.com/LandSandBoat/server/blob/54db671a076ee58509e1fcf4829834027d67c3ce/src/map/navmesh.cpp#L439

https://github.com/LandSandBoat/server/blob/54db671a076ee58509e1fcf4829834027d67c3ce/src/map/lua/luautils.cpp#L5121C57-L5121C57

Which is powered by this:

https://docs.unrealengine.com/4.27/en-US/API/Runtime/Navmesh/Detour/dtNavMeshQuery/moveAlongSurface/

So in the module you'd specify:
```
outputData =
{
 { xi.zone.VALKURM_DUNES, position_t(center of output) },
 ... more entries...
}

for [zone, centerPosition] in output data:
    PZone = GetZone(zone)
    for mob in PZone->ForEachMob[]...
        if distance(mob, centerPosition) < someNumberYouSpecifyAsTooClose:
            /* something something use centerPosition, m_navmesh, mob position, and some maths to find and validate a new position for mobs */
```

Doing it this way frees you from using hard coded IDs, but you have to be careful not to move any mobs who are NMs, or are in Garrison, etc.

zach2good:
The other way to do it is to use navmesh::snapToValidPosition. Get the distance and angle between center pos and the mob pos then use those alongside your highschool trig to generate a new position at your desired distance, and snap it to the navmesh because you won't know if it's valid until you do.

zach2good:
Like Teo mentioned; if the C++/Lua module approach is difficult to pull off, then it would be acceptable to do the lookups in SQL:
```sql
UPDATE mob_spawn_points SET pos_x = 210.481, pos_y = -6.946, pos_z = 101.274 WHERE mobname = "Goblin_Leecher" AND pos_x = 184.673 AND pos_y = -7.710 AND pos_z = 106.691; -- Valkurm Dunes, Goblin Leecher
```

Just to make sure we aren't looking up by possibly shifting IDs

CatsEyeXI:
> 

Ok, this I can do.

modules/custom/sql/safe_outposts.sql
```diff
@@ -0,0 +1,16 @@
+-- Relocates dangerous monsters away from outposts
+UPDATE mob_spawn_points SET pos_x =  210.481, pos_y =  -6.946, pos_z =  101.274 WHERE mobname = 'Goblin_Leecher' AND pos_x = 184.673 AND pos_y = -7.710 AND pos_z = 106.691;    -- Valkurm Dunes, Goblin Leecher
```
zach2good: Could you split each of these chunks by Zone, and align them with themselves please, then theres no need for the trailing comments since the mob name is baked in? Otherwise looks good 👍 

```sql
-- Valkurm Dunes
UPDATE mob_spawn_points SET pos_x = 210.481, pos_y = -6.946, pos_z = 101.274 WHERE mobname = 'Goblin_Leecher' AND pos_x = 184.673 AND pos_y = -7.710 AND pos_z = 106.691;

-- Buburimu Peninsula
UPDATE mob_spawn_points SET pos_x = -396.957, pos_y = -32.261, pos_z = 44.893 WHERE mobname = 'Goblin_Shepherd' AND pos_x = -166.429 AND pos_y = 7.910 AND pos_z = -105.478;
UPDATE mob_spawn_points SET pos_x = -411.430, pos_y = -31.429, pos_z =  3.221 WHERE mobname = 'Goblin_Bandit'   AND pos_x = -134.000 AND pos_y = 7.000 AND pos_z =  -96.000;

...
```

CatsEyeXI: all set 👍

zach2good (approved):

TeoTwawki (approved):
it'll _probably_ be fine with just name and x y z, at least till we refactor something that breaks the queries here.

PR #4001, opened 2023-06-16
[Fix] Tractor and Raise being usable when they shouldn't.

Xaver-DaRed:
**_I affirm:_**

Tractor and Raise can't be casted on a player that is dead but has the raise or the tractor menu active.
Confirmed this logic in retail with the use of 2 characters.

This PR fixes this issue/replicates this behavior.
Tested and working.

Step 1: Char A -> Die
Step 2: Char B -> Cast Tractor or any raise spell on Char A. Char A: Do nothing.
Step 3: Char B -> Attempt to cast tractor or Raise on Char A. Be blocked. Char A: Negate the raise or warp.
Step 4:  Char B -> Cast Tractor or any raise spell on Char A. Now you are able to.

### DIFF
--- a/scripts/globals/spells/black/tractor.lua
+++ b/scripts/globals/spells/black/tractor.lua
@@ -7,7 +7,12 @@ require("scripts/globals/msg")
 local spellObject = {}
 
 spellObject.onMagicCastingCheck = function(caster, target, spell)
-    if target:isMob() then -- Because Prishe in CoP mission
+    if
+        target:isMob() or                             -- Because Prishe in CoP mission.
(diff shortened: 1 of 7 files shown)


zach2good (approved):

TeoTwawki (approved):

TeoTwawki:
in case any old timers pop up to say the behavior is wrong: there WAS a time when the game would happily let you actually waste your time casting, let it go all the way, and then you'd be confused as nothing actually happened. now in modern retail it'll stop you.

PR #4011, opened 2023-06-19
Add template note to not resolve your own comment threads

TeoTwawki:
**_I affirm:_**

Github is ignoring [the problem](https://github.com/orgs/community/discussions/18458), and people keep doing this and its irritating especially if they just disagreed with a requested change and then "close" the conversation. Hopefully this note reduces people closing them before we're ready.

None.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,6 +2,7 @@
 <!-- (it should look like this: - [x] I have ...) -->
 **_I affirm:_**
 - [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
+- [ ] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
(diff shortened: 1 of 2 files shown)


zach2good:
I don't think this belongs in the main template, we need to keep it as small as humanly possible. Just yeet this into the contributor guides?

TeoTwawki:
> I don't think this belongs in the main template, we need to keep it as small as humanly possible. Just yeet this into the contributor guides?

Won't get noticed. Did this big and in their face intentionally. In a perfect world I wouldn't even have to say it let along put in the main template. I'm trying to be polite about it while removing any excuse of "oh I didn't know."

TeoTwawki:
Per private discussion, converted to checkbox. also rebased to 1 commit.

We could not automate this.

Xaver-DaRed (approved):

zach2good:
This will need this chunk updated:
https://github.com/LandSandBoat/server/blob/base/.github/workflows/github.yml#L44-L62

Or slightly modified to instead normalize and then count un-filled boxes

zach2good (approved):

PR #4013, opened 2023-06-20
Move conquest system to world server

xkoredev:
**_I affirm:_**

- Weekly tally / hourly updates now happen on world server
- World server communicates of map server updates via zmq
- Adds xi_world configuration and debug symbols for Mac OS
- Maps cache all data. Only world server uses DB.

Kill stuff as level 15+ outside. Check that influence points are changed in db. Check region map.
Run !updateconquest 0 for tally.
Run !updateconquest 2 for conquest update, after killing lots of mobs.

### DIFF
--- a/src/map/CMakeLists.txt
+++ b/src/map/CMakeLists.txt
@@ -45,8 +45,11 @@ set(SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/char_recast_container.h
     ${CMAKE_CURRENT_SOURCE_DIR}/command_handler.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/command_handler.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/conquest_data.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/conquest_data.h
     ${CMAKE_CURRENT_SOURCE_DIR}/conquest_system.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/conquest_system.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/conquest_zmq.h
     ${CMAKE_CURRENT_SOURCE_DIR}/daily_system.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/daily_system.h
     ${CMAKE_CURRENT_SOURCE_DIR}/enmity_container.cpp
--- a/src/map/conquest_data.cpp
+++ b/src/map/conquest_data.cpp
@@ -1,0 +1,146 @@
+#include "conquest_data.h"
+
+#include "common/logging.h"
+#include "common/sql.h"
+#include "conquest_system.h"
+
+ConquestData::ConquestData(std::unique_ptr<SqlConnection>& sql)
+: regionControls(std::vector<region_control_t>(256))
+, influences(std::vector<influence_t>(256))
+{
+    load(sql);
+}
+
+void ConquestData::load(std::unique_ptr<SqlConnection>& sql)
+{
+    const char* Query = "SELECT region_control, region_control_prev, sandoria_influence, bastok_influence, windurst_influence, beastmen_influence \
+                             FROM conquest_system;";
+
+    int32 ret = sql->Query(Query);
+
+    if (ret != SQL_ERROR && sql->NumRows() != 0)
+    {
+        uint8 regionId = 0;
+        while (sql->NextRow() == SQL_SUCCESS)
+        {
+            region_control_t regionControl;
+            regionControl.current    = sql->GetIntData(0);
+            regionControl.prev       = sql->GetIntData(1);
+            regionControls[regionId] = regionControl;
+
+            influence_t influence;
+            influence.sandoria_influence = sql->GetIntData(2);
+            influence.bastok_influence   = sql->GetIntData(3);
+            influence.windurst_influence = sql->GetIntData(4);
+            influence.beastmen_influence = sql->GetIntData(5);
+            influences[regionId]         = influence;
+
+            regionId++;
+        }
+    }
+}
+
+int32 ConquestData::getInfluence(REGION_TYPE region, NATION_TYPE nation) const
+{
+    switch (nation)
+    {
+        case NATION_SANDORIA:
+            return influences[(uint8)region].sandoria_influence;
+        case NATION_BASTOK:
+            return influences[(uint8)region].bastok_influence;
+        case NATION_WINDURST:
+            return influences[(uint8)region].windurst_influence;
+        case NATION_BEASTMEN:
+            return influences[(uint8)region].beastmen_influence;
+        default:
+            ShowWarning("Influence not known for nation: %d", nation);
+            return 0;
+    }
+}
+
+uint8 ConquestData::getRegionOwner(REGION_TYPE region) const
+{
+    return regionControls[(uint8)region].current;
+}
+
+uint8 ConquestData::getRegionControlCount(NATION_TYPE nation) const
+{
+    uint8 count = 0;
+    for (const auto& regionControl : regionControls)
+    {
+        if (regionControl.current == nation)
+        {
+            count++;
+        }
+    }
+
+    return count;
+}
+
+uint8 ConquestData::getPrevRegionControlCount(NATION_TYPE nation) const
+{
+    uint8 count = 0;
+    for (const auto& regionControl : regionControls)
+    {
+        if (regionControl.prev == nation)
+        {
+            count++;
+        }
+    }
+
+    return count;
+}
+
+auto ConquestData::getRegionControls() -> std::vector<region_control_t> const
+{
+    return regionControls;
+}
+
+void ConquestData::addInfluencePoints(int points, NATION_TYPE nation, REGION_TYPE region)
+{
+    switch (nation)
+    {
+        case NATION_SANDORIA:
+        {
+            influences[(uint8)region].sandoria_influence += points;
+            break;
+        }
+        case NATION_BASTOK:
+        {
+            influences[(uint8)region].bastok_influence += points;
+            break;
+        }
+        case NATION_WINDURST:
+        {
+            influences[(uint8)region].windurst_influence += points;
+            break;
+        }
+        case NATION_BEASTMEN:
+        {
+            influences[(uint8)region].beastmen_influence += points;
+            break;
+        }
+        default:
+        {
+            ShowWarning("Influence not known for nation: %d", nation);
+        }
+    }
+}
+
+void ConquestData::updateInfluencePoints(std::vector<influence_t> const& influencePoints)
+{
+    influences.clear();
+    for (const auto& influence : influencePoints)
+    {
+        influences.push_back(influence);
+    }
+}
+
+void ConquestData::updateRegionControls(std::vector<region_control_t> const& updatedRegionControls)
+{
+    regionControls.clear();
+    for (const auto& regionControl : updatedRegionControls)
+    {
+        regionControls.push_back(regionControl);
+    }
+}
--- a/src/map/conquest_data.h
+++ b/src/map/conquest_data.h
@@ -1,0 +1,73 @@
+#ifndef SERVER_CONQUEST_DATA_H
+#define SERVER_CONQUEST_DATA_H
+
+#include <string>
+
+#include "common/cbasetypes.h"
+#include "conquest_zmq.h"
+#include "zone.h"
+
+/**
+ * Conquest Data that is cached by the map server. This data is used to avoid redundant DB reads.
+ * Conquest data should be update periodically, as a result of influence / regional control updates
+ * received by the world server's conquest system.
+ */
+class ConquestData
+{
+public:
+    ConquestData(std::unique_ptr<SqlConnection>& sql);
+
+    /**
+     * Gets the influence points for a given nation in a given region.
+     */
+    int32 getInfluence(REGION_TYPE region, NATION_TYPE nation) const;
+
+    /**
+     * Gets the owner of a given region. That is, the one controlling this region since last tally.
+     */
+    uint8 getRegionOwner(REGION_TYPE region) const;
+
+    /**
+     * Gets the number of regions controlled by a given nation.
+     */
+    uint8 getRegionControlCount(NATION_TYPE nation) const;
+
+    /**
+     * Gets the number of regions controlled by a given nation in the
+     * previous tally.
+     */
+    uint8 getPrevRegionControlCount(NATION_TYPE nation) const;
+
+    /**
+     * Gets the array of region controls, indexed by regionId.
+     */
+    auto getRegionControls() -> std::vector<region_control_t> const;
+
+    /**
+     * Adds the given influence points to the given nation for the given
+     * region.
+     */
+    void addInfluencePoints(int points, NATION_TYPE nation, REGION_TYPE region);
+    /**
+     * Updates the influence points to match those given.
+     */
+    void updateInfluencePoints(std::vector<influence_t> const& influencePoints);
+
+    /**
+     * Updates region controls to match those given.
+     */
+    void updateRegionControls(std::vector<region_control_t> const& regionControls);
+
+private:
+    std::vector<region_control_t> regionControls;
+    std::vector<influence_t>      influences;
+
+    /**
+     * Loads the latest conquest data from DB.
+     * Should only ever be done on map initialization.
+     * World server should be updating conquest data periodically.
+     */
+    void load(std::unique_ptr<SqlConnection>& sql);
+};
+
+#endif // SERVER_CONQUEST_DATA_H
--- a/src/map/conquest_system.cpp
+++ b/src/map/conquest_system.cpp
@@ -21,9 +21,10 @@
 
 #include "conquest_system.h"
 
+#include "common/mmo.h"
 #include "common/vana_time.h"
-
 #include "entities/charentity.h"
+#include "message.h"
 #include "utils/charutils.h"
 #include "utils/zoneutils.h"
 
@@ -32,92 +33,39 @@
 #include "latent_effect_container.h"
 #include "lua/luautils.h"
 
-/************************************************************************
- *                                                                       *
- *  Реализация namespace conquest                                       *
- *                                                                       *
- ************************************************************************/
-
 namespace conquest
 {
-    /************************************************************************
-     *                                                                       *
-     *  UpdateConquestSystem                                                *
-     *                                                                       *
-     ************************************************************************/
-
-    void UpdateConquestSystem()
+    std::shared_ptr<ConquestData> conquestData;
+    std::shared_ptr<ConquestData> GetConquestData()
     {
-        TracyZoneScoped;
-
-        uint8 ranking            = conquest::GetBalance();
-        bool  isConquestAlliance = conquest::IsAlliance();
-        // clang-format off
-        zoneutils::ForEachZone([ranking, isConquestAlliance](CZone* PZone)
+        if (conquestData == nullptr)
         {
-            // only find chars for zones that have had conquest updated
-            if (PZone->GetRegionID() <= REGION_TYPE::DYNAMIS)
-            {
-                uint8 influence = conquest::GetInfluenceGraphics(PZone->GetRegionID());
-                uint8 owner     = conquest::GetRegionOwner(PZone->GetRegionID());
+            conquestData = std::make_shared<ConquestData>(sql);
+        }
 
-                luautils::OnConquestUpdate(PZone, Conquest_Update, influence, owner, ranking, isConquestAlliance);
-                PZone->ForEachChar([](CCharEntity* PChar)
-                {
-                    PChar->PLatentEffectContainer->CheckLatentsZone();
-                });
-            }
-        });
-        // clang-format on
+        return conquestData;
     }
 
-    void UpdateInfluencePoints(int points, unsigned int nation, REGION_TYPE region)
+    void AddInfluencePoints(int points, unsigned int nation, REGION_TYPE region)
     {
-        if (region == REGION_TYPE::UNKNOWN)
-        {
-            return;
-        }
+        // Send update message to world server
+        // Note that we do not update local cache, as it would potentially become out of sync from
+        // world server due to other map updates anyway. We wait for eventual consistency.
+        const std::size_t headerLength = 2 * sizeof(uint8);
+        const std::size_t dataLength   = headerLength + sizeof(int32) + sizeof(uint32) + sizeof(uint8);
+        uint8             data[dataLength]{};
 
-        std::string Query = "SELECT sandoria_influence, bastok_influence, windurst_influence, beastmen_influence FROM conquest_system WHERE region_id = %d;";
+        // Regional event type + conquest msg type
+        ref<uint8>((uint8*)data, 0) = REGIONAL_EVT_MSG_CONQUEST;
+        ref<uint8>((uint8*)data, 1) = CONQUEST_MAP2WORLD_ADD_INFLUENCE_POINTS;
 
-        int ret = sql->Query(Query.c_str(), static_cast<uint8>(region));
+        // Payload
+        ref<int32>(data, 2)  = points;
+        ref<uint32>(data, 6) = nation;
+        ref<uint8>(data, 10) = (uint8)region;
 
-        if (ret == SQL_ERROR || sql->NextRow() != SQL_SUCCESS)
-        {
-            return;
-        }
-
-        int influences[4] = {
-            sql->GetIntData(0),
-            sql->GetIntData(1),
-            sql->GetIntData(2),
-            sql->GetIntData(3),
-        };
-
-        if (influences[nation] == 5000)
-        {
-            return;
-        }
-
-        auto lost = 0;
-        for (auto i = 0u; i < 4; ++i)
-        {
-            if (i == nation)
-            {
-                continue;
-            }
-
-            auto loss = std::min<int>(points * influences[i] / (5000 - influences[nation]), influences[i]);
-            influences[i] -= loss;
-            lost += loss;
-        }
-
-        influences[nation] += lost;
-
-        sql->Query(
-            "UPDATE conquest_system SET sandoria_influence = %d, bastok_influence = %d, "
-            "windurst_influence = %d, beastmen_influence = %d WHERE region_id = %u;",
-            influences[0], influences[1], influences[2], influences[3], static_cast<uint8>(region));
+        // Do send the message
+        message::send(MSG_MAP2WORLD_REGIONAL_EVENT, data, dataLength);
     }
 
     /************************************************************************
@@ -129,7 +77,7 @@ namespace conquest
     void GainInfluencePoints(CCharEntity* PChar, uint32 points)
     {
         points += (uint32)(PChar->getMod(Mod::CONQUEST_REGION_BONUS) / 100.0);
-        conquest::UpdateInfluencePoints(points, PChar->profile.nation, PChar->loc.zone->GetRegionID());
+        conquest::AddInfluencePoints(points, PChar->profile.nation, PChar->loc.zone->GetRegionID());
     }
 
     /************************************************************************
@@ -196,7 +144,7 @@ namespace conquest
             }
         }
 
-        conquest::UpdateInfluencePoints(points, NATION_BEASTMEN, region);
+        conquest::AddInfluencePoints(points, NATION_BEASTMEN, region);
     }
 
     /************************************************************************
@@ -285,24 +233,13 @@ namespace conquest
         }
     }
 
-    uint8 GetInfluenceGraphics(REGION_TYPE regionid)
+    uint8 GetInfluenceGraphics(REGION_TYPE region)
     {
-        int32       sandoria = 0;
-        int32       bastok   = 0;
-        int32       windurst = 0;
-        int32       beastmen = 0;
-        const char* Query    = "SELECT sandoria_influence, bastok_influence, windurst_influence, beastmen_influence \
-                             FROM conquest_system WHERE region_id = %d;";
+        int32 sandoria = GetConquestData()->getInfluence(region, NATION_SANDORIA);
+        int32 bastok   = GetConquestData()->getInfluence(region, NATION_BASTOK);
+        int32 windurst = GetConquestData()->getInfluence(region, NATION_WINDURST);
+        int32 beastmen = GetConquestData()->getInfluence(region, NATION_BEASTMEN);
 
-        int32 ret = sql->Query(Query, static_cast<uint8>(regionid));
-
-        if (ret != SQL_ERROR && sql->NumRows() != 0 && sql->NextRow() == SQL_SUCCESS)
-        {
-            sandoria = sql->GetIntData(0);
-            bastok   = sql->GetIntData(1);
-            windurst = sql->GetIntData(2);
-            beastmen = sql->GetIntData(3);
-        }
         return GetInfluenceGraphics(sandoria, bastok, windurst, beastmen);
     }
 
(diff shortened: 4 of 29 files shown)


xkoredev:
@zach2good

src/map/CMakeLists.txt
```diff
@@ -137,9 +137,9 @@ set(SOURCES
 )
 
 if(UNIX)
-    set(resource "")
+    set(resource conquest_zmq.cpp conquest_zmq.h conquest_data.cpp conquest_data.h)
 else()
-    set(resource "${CMAKE_SOURCE_DIR}/res/mapserver.rc")
+    set(resource "${CMAKE_SOURCE_DIR}/res/mapserver.rc" conquest_zmq.cpp conquest_zmq.h conquest_data.cpp conquest_data.h)
```
zach2good: These don't go here, these go in the SOURCES section above - this is just for the rc  icon

src/map/conquest_data.h
```diff
@@ -0,0 +1,68 @@
...
+     * Loads the latest conquest data from DB.
+     * Should only ever be done on map initialization.
+     * World server should be updating conquest data periodically.
+     */
+    void load(std::unique_ptr<SqlConnection>& sql);
+
+    /**
+     * Gets the influence points for a given nation in a given region.
+     */
+    int32 getInfluence(REGION_TYPE region, uint8 nation);
```
zach2good: Can we start upgrading nation to its own enum like REGION_TYPE?

src/map/conquest_data.h
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Adds the given influence points to the given nation for the given
+     * region.
+     */
+    void addInfluencePoints(int points, uint8 nation, REGION_TYPE region);
+    /**
+     * Updates the influence points to match those given.
+     */
+    void updateInfluencePoints(std::vector<influence_t> influencePoints);
```
zach2good: If we're copying in influencePoints, presumably we're also not modifying it. So we could change this to be a const& and save ourselves a copy

src/map/conquest_data.h
zach2good: Can you go through all of the methods you're adding and try to make them const, where possible (and take in larger data types as const&, where possible)

src/map/conquest_system.cpp
```diff
@@ -34,90 +35,34 @@
 
 /************************************************************************
  *                                                                       *
- *  Реализация namespace conquest                                       *
+ *  Conquest namespace implementation                                    *
```
zach2good: 🔪 🇷🇺 

(I actually quite like Russia as a country, but I take great joy when these old comments get axed)

zach2good: Does this comment bring any value though? `namespace conquest` tells us its the start of the namespace...

src/map/conquest_zmq.cpp
```diff
@@ -0,0 +1 @@
+#include "conquest_zmq.h"
```
zach2good: 👀

src/map/conquest_zmq.h
```diff
@@ -0,0 +1,20 @@
+#ifndef SERVER_CONQUEST_ZMQ_H
+#define SERVER_CONQUEST_ZMQ_H
+
+#include <zmq.hpp>
```
zach2good: Not needed?

src/map/message.cpp
```diff
@@ -656,6 +658,95 @@ namespace message
...
+                            }
+
+                            conquest::conquestData->updateRegionControls(regionControls);
+                            return;
+                        }
+
+                        break;
+                    }
+                    default:
+                        ShowWarning("Message: unhandled regional event type %d", eventType);
```
zach2good, line 746: This needs to be in a {} block

src/map/utils/battleutils.cpp
```diff
@@ -301,7 +301,7 @@ namespace battleutils
 
         if (ret != SQL_ERROR && sql->NumRows() != 0)
         {
-            for (uint32 x = 0; sql->NextRow() == SQL_SUCCESS; ++x)
+            while (sql->NextRow() == SQL_SUCCESS)
```
zach2good, line 304: 👍

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,832 @@
+﻿/*
```
zach2good: This shouldn't be part of this PR

src/world/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ list(REMOVE_ITEM SOURCES
 if(UNIX)
     set(resource "")
 else()
-    set(resource "${CMAKE_SOURCE_DIR}/res/worldserver.rc")
+    set(resource "${CMAKE_SOURCE_DIR}/res/worldserver.rc" message_handler.h)
```
zach2good: This should go in the SOURCES block

src/world/conquest_system.cpp
```diff
@@ -20,3 +20,342 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
+===========================================================================
+
+Copyright (c) 2010-2015 Darkstar Dev Teams
+
+                        This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+                                         (at your option) any later version.
+
+                                         This program is distributed in the hope that it will be useful,
```
zach2good: Formatting of this is all screwwy?

src/world/conquest_system.h
```diff
@@ -20,9 +20,37 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 */
 #pragma once
 
-class ConquestSystem
+#include "./common/sql.h"
```
zach2good: No need for the unix-style ./'s

src/world/conquest_system.cpp
```diff
@@ -20,3 +20,342 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
+}
+
+bool ConquestSystem::handleMessage(MSGSERVTYPE     type,
+                                   zmq::message_t* extra,
+                                   zmq::message_t* packet,
+                                   zmq::message_t* from,
+                                   in_addr         from_addr,
+                                   uint16          from_port)
+{
+    if (type != MSGSERVTYPE::MSG_MAP2WORLD_REGIONAL_EVENT)
```
zach2good: I like the abstract way of handling handleMessage and the message handlers, but this is quite wasteful. 

It goes through each registered handler (we'll eventually have a few: the 4x regional event things, party system, some other stuff, etc.) and then checks to see if thats the right place. During registration we could be mapping the message type to the destination and then doing a lookup to dispatch the message, instead of this loop

src/world/message_server.cpp
```diff
@@ -248,11 +297,14 @@ void message_server_listen(const bool& requestExit)
     }
 }
 
-void message_server_init(const bool& requestExit)
+void message_server_init(const std::shared_ptr<SqlConnection>&               sql,
```
zach2good: You've activated my trap card (and I think you messaged me about this).

The underlying SQL handle we use in SqlConnection is ABSOLUTELY NOT THREAD SAFE EVEN SLIGHTLY. If you access it from a different thread from where it was created it'll wipe it out silently and then the next access will crash the system. message_server lives on its own thread, so you can't pass a shared_ptr to a SqlConnection created on the main thread to it, it needs to have its own one (like I had set up before).

There's no harm in the message_server, the parent world server, and each message-handling-regional-event-thing having their own SqlConnections as part of their classes, they're very cheap and SQL Server is very performant

src/world/message_server.cpp
```diff
@@ -248,11 +297,14 @@ void message_server_listen(const bool& requestExit)
...
+                         const std::vector<std::shared_ptr<MessageHandler>>& handlers,
```
zach2good: Just as a general design thing, I would probably pass in an array of weak_ptrs here and then canonicalize them when we need to use the underlying object

src/map/conquest_system.h
```diff
@@ -49,9 +51,9 @@ class CCharEntity;
 
 namespace conquest
 {
-    void UpdateConquestSystem(); // Update conquest information in the DB
+    extern ConquestData* conquestData; // Cached data with influences / region controls
```
zach2good: Avoid using extern here, we can construct this in other ways

zach2good (commented):
Just a quick driveby before I start my work day.

In places where there's a big fat sql global: don't be afraid of using it. In places where there isn't, you can/should be creating a SqlConnection for that class.

Make sure you're sticking to the style guide: <https://github.com/LandSandBoat/server/wiki/Development-Guide-Cpp#naming--misc>

src/world/conquest_system.cpp
Xaver-DaRed: It's also uneeded I would think.
The LSB copyright is at the top of the file, no need to add back the old DSP one

src/map/conquest_system.h
xkoredev: How would you go about constructing that data? Not used to working with namespaces

src/map/conquest_system.cpp
xkoredev: Ya I was considering removing some of them. wasn't sure if we wanted to leave them for consistency, but in this case it makes sense

src/world/conquest_system.cpp
xkoredev: how would you register these? I was thinking of a map of tuple<type, subtype> but the syntax was kinda ugly

src/map/packets/conquest_map.cpp
```diff
@@ -36,77 +38,38 @@ CConquestPacket::CConquestPacket(CCharEntity* PChar)
     this->setType(0x5E);
     this->setSize(0xB4);
 
-    // TODO: Periodically cache this information inside conquest_system so that it
-    //     : isn't so on-demand
-    const char* Query = "SELECT region_id, region_control, region_control_prev, \
-                         sandoria_influence, bastok_influence, windurst_influence, \
-                         beastmen_influence FROM conquest_system;";
+    auto  conquestData     = conquest::conquestData;
```
zach2good: This is creating a copy

src/map/conquest_system.h
zach2good: You need to look into how these work: https://en.cppreference.com/w/cpp/language/storage_duration

main.cpp:
```cpp
#include "conquest.h"

int main(int argc, char** argv)
{
    conquest::init();
    
    auto* conquestData = conquest::getConquestData();

    return 0;
}
```

conquest.h
```cpp
#pragma once

namespace conquest
{
    struct ConquestData
    {

    };

    void init();
    ConquestData* getConquestData();
}
```

conquest.cpp
```cpp
#include "conquest.h"

namespace conquest
{
    // You could let the data be default constructed here
    ConquestData conquestData;

    // Or:
    // ConquestData* conquestData;

    void init()
    {
        // You could construct it with args here
        // conquestData = new ConquestData(...);

        // * Make sure you delete it later, or use shared_ptr
    }

    ConquestData* getConquestData()
    {
        // You could also return raw ptr or a shared ptr here, or a ref
        return &conquestData;
    }
}
```


^ It depends entirely on how/where you want to store the data. I would recommend NOT putting it into static storage (stored and constructed in a namespace/global scope).

zach2good: ^ Note that the storage is defined in the compilation unit, not the header *

src/world/conquest_system.cpp
zach2good: You're quite close:

In this block:
```cpp
    // Setup message handlers
    std::vector<std::shared_ptr<MessageHandler>> handlers;
    handlers.push_back(conquestSystem);
    messageServer = std::make_unique<message_server_wrapper_t>(zmqSql, handlers, std::ref(m_RequestExit));
```

You've already created the conquest system, just create a map of the defining message type to a shared_ptr to the ConquestSystem. The map will be of pointers to MessageHandlers, since it's an interface that's all fine.

zach2good: Yeah, instead of this long chain of fallthroughs, you should be mapping things to function pointers

src/map/CMakeLists.txt
xkoredev: My IDE put them there.. will move em out.

src/world/conquest_system.cpp
xkoredev: the part that i'd like to make prettier is mapping the type and subtype to a handler. Few options:
- using a tuple<type, subtype> as key for the handler map
- using a nested map<type, map<subtype, handler>>
- using a uint16 key and just use key = type | (subtype << 8)

I am leaning towards the latter

zach2good: You have 2x sets of keys, there's no need to combine them or do a single dispatch.

You do one lookup of destination key to shared ptr value of the relevant message handler in the message server, then forward the payload.

You do the next lookup of subject (message type) once you're in the correct handler, mapping the key to a function pointer of the function that'll handle the payload

zach2good: Type:  REGIONALMSGTYPE : uint8
Subtype: CONQUESTMSGTYPE : uint8

src/map/CMakeLists.txt
xkoredev: Done

src/map/conquest_system.cpp
xkoredev: Removed.

src/world/conquest_system.h
xkoredev: Removed

src/world/conquest_system.cpp
zach2good:
```cpp
#include <vector>
#include <string>
#include <memory>
#include <unordered_map>
#include <functional>
#include <iostream>

// Defined globally
enum REGIONAL_EVENT_HANDLERS
{
    CONQUEST,
};

// Defined so only conquest systems in world/map can see/care about it
enum CONQUEST_MESSAGES
{
    DO_THING,
};

class MessageHandler
{
public:
    virtual ~MessageHandler() = default;
    virtual void handleMessage(std::vector<int> payload) = 0;
};

class ConquestSystem : public MessageHandler
{
public:
    ConquestSystem()
    {
        // Set up internal handler lookup
        internalHandlerMap[CONQUEST_MESSAGES::DO_THING] = std::bind(&ConquestSystem::handleDoThing, this, std::placeholders::_1);
    }

    void handleMessage(std::vector<int> payload)
    {
        // Lookup key, call member function
        auto key = (CONQUEST_MESSAGES)payload[1];
        internalHandlerMap[key](payload);
    }

    void handleDoThing(std::vector<int> payload)
    {
        std::cout << "Called handleDoThing from ConquestSystem\n";
    }

private:
    std::unordered_map<CONQUEST_MESSAGES, std::function<void(std::vector<int>)>> internalHandlerMap;
};

int main()
{
    // Demo Setup
    std::unordered_map<REGIONAL_EVENT_HANDLERS, std::shared_ptr<MessageHandler>> handlerMap;
    handlerMap[REGIONAL_EVENT_HANDLERS::CONQUEST] = std::make_shared<ConquestSystem>();

    // Message Server gets message
    std::vector<int> message{
        REGIONAL_EVENT_HANDLERS::CONQUEST,
        CONQUEST_MESSAGES::DO_THING,
        42 // payload
    };

    // Extract lookup key
    auto regionalEventHandlerKey = (REGIONAL_EVENT_HANDLERS)message[0];

    // Use key to lookup the handler, then give the message to the handler (a copy in this case)
    handlerMap[regionalEventHandlerKey]->handleMessage(message);

    return 0;
}
```

zach2good: I'd probably make `virtual void handleMessage(std::vector<int> payload) = 0;` take `std::vector<int>&&` so things had to be `std::move`'d into them

src/map/conquest_system.h
zach2good: If each messageHandler ends up being on its own thread, then they'll have to communicate using locks or with a lockless queue, also

src/world/conquest_system.cpp
zach2good: If each messageHandler ends up being on its own thread, then they'll have to communicate using locks or with a lockless queue, also

src/world/conquest_system.cpp
```diff
@@ -20,3 +20,342 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
+    // 3- Send tally end Msg
+    sendRegionControlsMsg(CONQUEST_WORLD2MAP_WEEKLY_UPDATE_END);
+}
+
+void ConquestSystem::updateHourlyConquest()
+{
+    sendInfluencesMsg(true);
+}
+
+void ConquestSystem::updateDailyTallyPoints()
```
cocosolos: Not sure if this is the best place for this since it has no relation to the conquest system.

src/map/conquest_data.h
xkoredev: Done

src/map/conquest_data.h
xkoredev: Done

src/map/conquest_data.h
xkoredev: To be clear created the enum and used it in conquest, but wont refactor outside

xkoredev: Done

src/map/conquest_zmq.h
xkoredev: Removed

src/map/conquest_zmq.cpp
xkoredev: removed empty file

src/map/message.cpp
xkoredev: Done. Curious, is this a codestyle thing (makes sense) or is the functionality different without it?

src/world/CMakeLists.txt
xkoredev: Moved

src/map/utils/fishingcontest.cpp
xkoredev: Removed. Must have been an artifact of my local state at the time in ASB

src/world/conquest_system.cpp
xkoredev: Thanks. Idk why I copy/pasta'd it there. Removed.

src/world/conquest_system.cpp
xkoredev: Since we don't know our requirements for all handlers yet, I was going to just keep the same arguments to the message parse method (`zmq::message_t* extra, zmq::message_t* packet, from_ip, from_port`). Would there be issues with that?

xkoredev: Actually packet seems more meant for the request / reply mechanism so I think I can do with just the data vector as you suggested, and from_ip / port

zach2good: Remember (so far, it might change in the future) that the idea is to not generate actual packet data for the client in the world server.

It's a little annoying and inconvenient that they use the same structures and mechanisms, but ultimately, world server doesn't have all the same information that map does - nor does it matter.

xkoredev: so if we send around an std::vector<uint8>, is this the way we would access properties?
```
        int32 points;
        std::memcpy(&points, payload.data(), sizeof(int32));
```

Any other options?

src/world/message_server.cpp
xkoredev: I've switched to creating instances of sql connection as needed, or using the global when available.

src/world/conquest_system.cpp
xkoredev: I've changed this to use a vector<uint8> payload. Let me know if that makes sense.

src/world/message_server.cpp
xkoredev: Gotcha. I am setting up the handlers in message server now so no need to do this anymore.

src/world/conquest_system.cpp
xkoredev: Moved it back. I was confused by the word tally. Thanks!

src/world/conquest_system.cpp
zach2good: What do you mean by properties? For a given message type, you know what type of item is going in, vs what is coming out?
```
std::memcpy(&points, payload.data(), sizeof(int32));
```
Should work fine, `.data()` can be abused for storage of whatever you want - given that it's contiguous memory. If you want to store something more complex, it's defined as a vector of uint8's (bytes). So you can just construct the payload as sizeof(your_struct) and then write/read out the data from `.data()`

zach2good: If you _really really_ wanted to, you could also start using flatbuffers, but that feels overkill for this

src/map/conquest_system.h
xkoredev: Done

src/map/packets/conquest_map.cpp
xkoredev: Using shared ptr now

xkoredev:
@zach2good ammended comments. Next step will be adding the cache on the world server that is flushed every 2 min, as well as map server zone ip / port cache, tho the latter will require a world restart when maps restart if the settings change.
Should I make these changes in this PR? or follow up?

src/map/conquest_system.h
```diff
@@ -60,8 +50,9 @@ namespace conquest
     uint8 GetInfluenceRanking(int32 san_inf, int32 bas_inf, int32 win_inf, int32 bst_inf);
     uint8 GetInfluenceRanking(int32 san_inf, int32 bas_inf, int32 win_inf);
 
-    void UpdateConquestGM(ConquestUpdate type); // Update conquest system by GM (modify in the DB and use @updateconquest)
-    void UpdateWeekConquest();                  // Update conquest system every sunday
+    void HandleInfluenceUpdate(const std::vector<influence_t>& influences, bool shouldUpdateZones); // Triggered periodically when world server sends updates of the latest influence data.
```
zach2good: We're an east const (EAST IS BEAST) family around here ;)

(apart from const char*, because what is life without nonsensical conventions)

src/map/conquest_data.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "common/sql.h"
+#include "conquest_system.h"
+
+ConquestData::ConquestData()
+: regionControls(std::vector<region_control_t>(256))
+, influences(std::vector<influence_t>(256))
+{
+}
+
+void ConquestData::load(std::unique_ptr<SqlConnection>& sql)
```
zach2good, line 14: Given that this doesn't have any dependencies apart from the sql handle, I'd move all of this logic into the constructor and use the global sql handle with extern. I think its OK to lean on that design pattern for now, and at some point in the future I'll kill all the global extern usage of the sql handle etc.

zach2good: Also, given how little this data is, we could just lazy-load it just before its needed the first time?

src/world/world_server.cpp
```diff
@@ -21,12 +21,13 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
 #include "common/application.h"
-#include "common/console_service.h"
 #include "common/logging.h"
+#include "common/vana_time.h"
 
 namespace
 {
-    std::unique_ptr<SqlConnection> sql;
+    std::unique_ptr<SqlConnection>  sql;
+    std::unique_ptr<ConquestSystem> conquest;
```
zach2good: There's already a std::shared_ptr<ConquestSystem> conquestSystem; in worldServer members?

src/world/world_server.cpp
```diff
@@ -21,12 +21,13 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
 
 #include "common/application.h"
-#include "common/console_service.h"
 #include "common/logging.h"
+#include "common/vana_time.h"
 
 namespace
 {
-    std::unique_ptr<SqlConnection> sql;
+    std::unique_ptr<SqlConnection>  sql;
```
zach2good, line 29: World server itself isn't doing any SQL stuff, does it need its own sql handle?

src/world/message_server.cpp
```diff
@@ -39,18 +40,59 @@ struct chat_message_t
 
 moodycamel::ConcurrentQueue<chat_message_t> outgoing_queue;
 
-std::unique_ptr<SqlConnection> zmqSql;
+std::unique_ptr<SqlConnection>                                       sql;
+std::unordered_map<REGIONALMSGTYPE, std::shared_ptr<MessageHandler>> regionalMsgHandlers;
+
+typedef std::function<bool(zmq::message_t* data,
```
zach2good, line 46: 👍

src/world/message_server.cpp
```diff
@@ -163,6 +205,15 @@ void message_server_parse(MSGSERVTYPE type, zmq::message_t* extra, zmq::message_
             // no op
             break;
         }
+        case MSG_MAP2WORLD_REGIONAL_EVENT:
+        {
+            uint8*      data    = (uint8*)extra->data();
+            const uint8 subType = ref<uint8>(data, 0);
+
+            std::vector<uint8> bytes(data, data + extra->size());
+            regionalMsgHandlers[(REGIONALMSGTYPE)subType]->handleMessage(bytes, from_ip, from_port);
```
zach2good: 👍

zach2good: In the interest of safety, I'd look into not using operator[] and using .get() instead, and wrapping this in a try/catch block -> because world-server needs to be bulletproof. 

If an unvalidated subType comes in, operator[] access in the map will create a new shared_ptr, which we immediately call handleMessage on. I'm not sure if this would be a nullptr dereference or not, but best not to risk it

src/world/message_handler.h
```diff
@@ -0,0 +1,18 @@
+#ifndef SERVER_MESSAGE_HANDLER_H
+#define SERVER_MESSAGE_HANDLER_H
+
+#include "common/mmo.h"
+#include <zmq.hpp>
+
+class MessageHandler
```
zach2good: I'd probably call this IMessageHandler, since its a pure interface

src/world/conquest_system.cpp
```diff
@@ -20,3 +20,282 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
+        const std::size_t offset             = headerLength + sizeof(size_t) + sizeof(region_control_t) * i;
+        ref<uint8>((uint8*)data, offset)     = regionControls[i].current;
+        ref<uint8>((uint8*)data, offset + 1) = regionControls[i].prev;
+    }
+
+    // 3- Create ZMQ Message and queue it
+    zmq::message_t dataMsg = zmq::message_t(dataLen);
+    memcpy(dataMsg.data(), data, dataLen);
+
+    if (ipp == 0)
```
zach2good: I wonder if this is dangerous or not, because I think we use 0 as a designator for "Zone is turned off" in a lot of places?

src/map/message.cpp
zach2good: It functions because it just runs from the case statement downwards, but that doesn't align with what if and while do, so its cleaner and clearer to wrap things in {}'s, with the break either inside or outside - I don't know if we have a convention for that

zach2good (commented):
Getting closer. I wonder if ConquestData could/should be a private class inside the conquest system file in the conquest namespace. It doesn't do much apart from data loading and retrieval, and those can all just be functions that exist in the conquest namespace?

src/world/message_server.cpp
zach2good: [https://en.cppreference.com/w/cpp/memory/unique_ptr/operator*](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator*)

Turns out this sort of access is undefined behaviour, lets avoid that :D

src/map/conquest_system.h
xkoredev: Done

src/map/conquest_data.cpp
xkoredev: I'll lazy load it

xkoredev: actually this is in map. which global should we rely on ? the map one?

xkoredev: Done. On construction

src/world/world_server.cpp
xkoredev: We need to expose world in the namespace to then right?

xkoredev: This pointer is the one in world, assigned on construction, just like sql

src/world/world_server.cpp
xkoredev: its used in this `UpdateDailyTallyPoints` method which I moved back outside per request (since it doesnt belong to conquest system)

src/world/message_server.cpp
xkoredev: So I understand. Avoid using ->get() or avoid accessing without checking first?

src/world/message_handler.h
xkoredev: Done

src/world/conquest_system.cpp
xkoredev: Alternative would be create a different method, or maybe a boolean flag for broadcast?

xkoredev:
> Getting closer. I wonder if ConquestData could/should be a private class inside the conquest system file in the conquest namespace. It doesn't do much apart from data loading and retrieval, and those can all just be functions that exist in the conquest namespace?

Ya I think it should. I just didn't know about that being a possibility as I have not worked with namespaces in the past. Will do.

xkoredev:
> Contributor

I stacked overlown it (because im a namespace noob) and not sure that you can create a private namespace class:
https://stackoverflow.com/questions/38795042/how-make-a-class-private-inside-a-namespace-in-c#:~:text=You%20cannot.,restrict%20access%20using%20friend%20specifiers.

src/world/message_server.cpp
zach2good:
```cpp
try
{
    auto& handler = regionalMsgHandlers.get((REGIONALMSGTYPE)subType); // or is this a shared ptr? I don't remember
    handler->handleMessage(...);
}
catch (the possible exception from get() throwing)
{
...
}
```

zach2good: The idea is to use get, because get will throw if the key isn't there. If you use operator[] and it isn't there, it'll create an entry (an empty shared ptr). Accessing members of the underlying class of an empty shared ptr is undefined behaviour (it might work, it might crash, we can't know)

src/world/conquest_system.cpp
zach2good: If you're trying to use a magic number to send things to the world server, I'd probably go for 0xFFFF

zach2good:
By private I mean it's defined only in the cpp file, so it isn't advertised to consumers of the related header.

```cpp
#include <vector>

// This is main.cpp, or any cpp file:
namespace conquest
{
    class Test
    {
        public:
            Test() = default;
    };
}

// conquest::Test won't be available from the conquest header, if thats also defining and advertising the conquest namespace

int main()
{
    conquest::Test t;   
    return 0;
}
```

src/world/world_server.cpp
zach2good: An anonymous namespace is used to hide local objects/data within a given compilation unit - stopping them from being extern'd from elsewhere. You can _technically_ do either way, but both is messy and confusing. The Conquest System should be a unique ptr member of the class, it doesn't need to be hidden. The local SQL ptr makes more sense in the anonymous namespace because it is explicitly _not_ for sharing, since most systems on world server will have their own and everything is threaded

zach2good:
```cpp
WorldServer::WorldServer(int argc, char** argv)
: Application("world", argc, argv)
, conquestSystem()
, httpServer(std::make_unique<HTTPServer>())
, messageServer(std::make_unique<message_server_wrapper_t>(std::ref(m_RequestExit)))
{
    // Anonymous namespace preparation
    sql      = std::make_unique<SqlConnection>();
    conquest = std::make_unique<ConquestSystem>();
```

This means you have 2x conquest systems now. `, conquestSystem()` is constructing the WorldServer::conquestSystem member, and `conquest = std::make_unique<ConquestSystem>();` is constructing the anonymous namespace member conquest in this compilation unit.

zach2good:
See here:
image

Even though conquest::Test is defined in the cpp, the definition isn't available to the header - unless you were to use a forward declaration of it, but that's not the point here

src/world/message_server.cpp
xkoredev: gotcha thanks for explaining.

src/world/conquest_system.cpp
xkoredev: 👍

src/world/world_server.cpp
xkoredev: Oh right yes. I did that intentionally because conquestSystem is stateless and I did not know a better way to share the conquest system. However i'd prefer being able to use the world system instance from world_server.cpp, but how do we do that? (i.e for the tally tick logic)

src/world/message_server.cpp
xkoredev: Done

xkoredev:
> See here: image
> 
> Even though conquest::Test is defined in the cpp, the definition isn't available to the header - unless you were to use a forward declaration of it, but that's not the point here

Actually currently CCConquestPaquet accesses conquestData via the public method (conquest::GetConquestData()). Making conquestData private would require a lot more refactor

zach2good:
https://godbolt.org/z/hdYrn117E

xkoredev:
> 

Thanks for the handhold lol.. feel kinda bad. I misunderstood this as you wanting to avoid exposing `GetConquestData()`

xkoredev:
> https://godbolt.org/z/hdYrn117E

The only downside I see here then is that conquest data has a few methods which will all be moved to conquest system, making the file a bit harder to parse.

xkoredev:
> https://godbolt.org/z/hdYrn117E

How about accessing conquestData.x? This is what CConquestPacket() does, but I don't see a way of doing so without exposing this in the headers / API.

src/world/world_server.cpp
xkoredev: Done. The same shared_ptr can be used for both

xkoredev:
> 

As mentioned offline, we are keeping it public

PR #4015, opened 2023-06-20
[Fix] Incorrect shell effect power values in mobskill scripts

Xaver-DaRed:
**_I affirm:_**

Fixes wrong power values in mobskills that grant shell.
Missed when the modifier was converted to its new scale.

Fight a crab. Wait for it to use bubble courtain. !getmod DMGMAGIC

Same for the other 3 mobskills, but with other mobs

### DIFF
--- a/scripts/globals/mobskills/bubble_armor.lua
+++ b/scripts/globals/mobskills/bubble_armor.lua
@@ -17,7 +17,7 @@ end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local typeEffect = xi.effect.SHELL
-    local power = 50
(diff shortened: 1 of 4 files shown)


WinterSolstice8 (approved):

TeoTwawki (approved):

TeoTwawki (changes requested):
I was too fast to approve, looed an saw after.

we actually need positive power, the effect script is subtracting by power
https://github.com/LandSandBoat/server/blob/base/scripts/globals/effects/shell.lua#L7

TeoTwawki (approved):
re-aproved

PR #4023, opened 2023-06-25
[gear sets] Account for items that belong to multiple sets

kelppo:
**_I affirm:_**

Before this commit, items such as Virtue Stone only activates one of the sets it belongs to (which happens to be the Hope Staff) because its entry in the xi.gear_sets.itemToSetId table only holds one setId, fix this by turning the entry into an array containing all the setIds the item belongs to.

1. !changejob rdm 99
2. !additem 17710
3. !additem 18244
4. Equip both items, observe how !getmod AMMO_SWING returns 0
5. Apply this patch
6. !getmod AMMO_SWING should now returns 50

### DIFF
--- a/scripts/globals/gear_sets.lua
+++ b/scripts/globals/gear_sets.lua
@@ -2541,19 +2541,13 @@ xi.gear_sets.createItemToSetId = function()
 
     for setId, setData in pairs(gearSets) do
         for _, itemId in ipairs(setData.items) do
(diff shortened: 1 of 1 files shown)


kelppo:
I guess I can also turn xi.gear_sets.itemToSetId into a table of tables instead of a mix of numbers and tables. Let me know which one you guys prefer.

WinterSolstice8 (approved):

claywar (approved):

PR #4025, opened 2023-06-27
add checkbox_total variable to github workflow

TeoTwawki:
we needed to have a variable because this check happens on both issues and PRs, but they use a different number of checkboxes

**_I affirm:_**

we needed to have a variable because this check happens on both issues and PRs, but they use a different number of checkboxes

~~The variable syntax isn't clear to me what we have doesn't look like GitHub's documentation, so this is not yet tested. I've never seen variables in yaml specify a datatype before, but the existing ones do . "why are these intergers const?"~~

_its frakkin javascipt_

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -19,14 +19,16 @@ jobs:
                 return string.split(word).length - 1;
             }
 
-            const event_type = "${{ github.event_name }}"
-            var   body_text  = ""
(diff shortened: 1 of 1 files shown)


TeoTwawki:
k, now that I know that is JS in there, its _probably_ fine

zach2good (approved):

PR #4029, opened 2023-06-28
[Trust] Added distances for various Trusts

thecursedcometh2:
**_I affirm:_**

This PR adds movement types (standing back or getting in close) for Kupipi, Joachim, Ulmia, Karaha-Baruha, Ferreous Coffin, Koru-Moru, Apururu (UC), Adelheid, Ygnas, and Monberaux. All except Monberaux will stand at a far distance, with Monberaux standing slightly closer because he was out of range of his skills otherwise.

- Summon each Trust.
- Observe their idle distance from the player. All trusts except Monberaux will stand at the farthest possible distance, with Monberaux standing somewhat closer.


sql/mob_pools.sql
```diff
@@ -5954,7 +5954,7 @@ INSERT INTO `mob_pools` VALUES (5895,'Donggu','Donggu',116,0x0000780100000000000
 -- Trusts
 INSERT INTO `mob_pools` VALUES (5896,'shantotto','Shantotto',153,0x0000B80B00000000000000000000000000000000,4,0,2,240,40,1,0,0,0,0,0,32,0,3,0,0,308,0,0,1011,153);
 INSERT INTO `mob_pools` VALUES (5897,'naji','Naji',149,0x0000B90B00000000000000000000000000000000,1,0,2,240,100,0,0,0,0,0,0,32,0,3,0,0,309,0,0,1012,149);
-INSERT INTO `mob_pools` VALUES (5898,'kupipi','Kupipi',153,0x0000BA0B00000000000000000000000000000000,3,0,11,230,40,1,0,0,0,0,0,32,0,3,0,0,310,0,0,1013,153);
+INSERT INTO `mob_pools` VALUES (5898,'kupipi','Kupipi',153,0x0000BA0B00000000000000000000000000000000,3,0,11,230,40,3,0,0,0,0,0,32,0,3,0,0,310,0,0,1013,153);
```
TeoTwawki: Kupipi's behavior is somewhat dependent on where in the order she was summoned, a sign that we need to have some conditional logic we can control rather than static setting this in a field we already used for something else (I was just discussing with Zach how we should not have 1 sql field with 2 enums)

----

update:

https://github.com/LandSandBoat/server/pull/4029#discussion_r1245321721

and that's why she's at "NO_MOVE" for hers. Its less conditional than previously thought.

TeoTwawki: I'm not sure how far back "mid" range is vs long, but I also know Kupipi doesn't stand as far back as Apururu even when summoned dead last.

sql/mob_pools.sql
```diff
@@ -5967,10 +5967,10 @@ INSERT INTO `mob_pools` VALUES (5907,'lion','Lion',149,0x0000C30B000000000000000
 INSERT INTO `mob_pools` VALUES (5908,'tenzen','Tenzen',149,0x0000C40B00000000000000000000000000000000,12,0,10,440,200,0,0,0,0,0,0,32,0,3,0,0,320,0,0,1023,149);
 INSERT INTO `mob_pools` VALUES (5909,'mihli_aliapoh','MihliAliapoh',151,0x0000C50B00000000000000000000000000000000,3,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,321,0,0,1024,151);
 INSERT INTO `mob_pools` VALUES (5910,'valaineral','Valaineral',145,0x0000C60B00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,322,0,0,1025,145);
-INSERT INTO `mob_pools` VALUES (5911,'joachim','Joachim',149,0x0000C70B00000000000000000000000000000000,10,3,3,500,25,2,0,0,0,0,0,32,0,3,0,0,323,0,0,1026,149);
+INSERT INTO `mob_pools` VALUES (5911,'joachim','Joachim',149,0x0000C70B00000000000000000000000000000000,10,3,3,500,25,3,0,0,0,0,0,32,0,3,0,0,323,0,0,1026,149);
```
TeoTwawki: pretty sure he (and Ulmia) was correct at 2 for mid-range

sql/mob_pools.sql
```diff
@@ -5992,26 +5992,26 @@ INSERT INTO `mob_pools` VALUES (5932,'fablinix','Fablinix',327,0x0000DC0B0000000
...
 INSERT INTO `mob_pools` VALUES (5947,'uka_totlihn','UkaTotlihn',151,0x0000EB0B00000000000000000000000000000000,19,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,359,0,0,1062,151);
 INSERT INTO `mob_pools` VALUES (5948,'klara','Klara',149,0x0000ED0B00000000000000000000000000000000,1,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,360,0,0,1063,149);
 INSERT INTO `mob_pools` VALUES (5949,'romaa_mihgo','RomaaMihgo',151,0x0000EE0B00000000000000000000000000000000,6,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,361,0,0,1064,151);
 INSERT INTO `mob_pools` VALUES (5950,'kuyin_hathdenna','KuyinHathdenna',151,0x0000EF0B00000000000000000000000000000000,21,0,3,240,100,0,0,0,0,0,0,32,0,3,0,0,362,0,0,1065,151);
 INSERT INTO `mob_pools` VALUES (5951,'rahal','Rahal',145,0x0000F00B00000000000000000000000000000000,7,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,363,0,0,1066,145);
-INSERT INTO `mob_pools` VALUES (5952,'koru-moru','Koru-Moru',153,0x0000F10B00000000000000000000000000000000,5,0,3,240,100,1,0,0,0,0,0,32,0,3,0,0,364,0,0,1067,153);
+INSERT INTO `mob_pools` VALUES (5952,'koru-moru','Koru-Moru',153,0x0000F10B00000000000000000000000000000000,5,0,3,240,100,3,0,0,0,0,0,32,0,3,0,0,364,0,0,1067,153);
```
TeoTwawki, line 6011: Koru is def not correct as long range. he gets in my way in melee range all the time on retail. I wish he was long range.

sql/mob_pools.sql
thecursedcometh2: Should I revert the change or keep it? Not sure when that condition will be implemented.

sql/mob_pools.sql
thecursedcometh2: In my experience Joachim stood back a far way, but that could be wrong so I'll revert it.

sql/mob_pools.sql
thecursedcometh2: It might be job dependent, I've seen him stand back quite a bit on some jobs. I'll revert it, though.

sql/mob_pools.sql
TeoTwawki: I'd leave her be till we sort it out. Since it wasn't 0, someone set her to 1 on purpose.

sql/mob_pools.sql
TeoTwawki: according to wiki the deal is that he just does not move at all. same for Kupipi, which is why she'll melee if summoned early in the lineup - because she'll already be closer.

> Will remain stationary wherever he is standing upon engaging an enemy. Will not attempt to move closer to the enemy unless it moves out of his casting range.

Koru is in my way all the time because he's a non engaging statue of a caster, and you might have managed to start battles with him farther out.

sql/mob_pools.sql
TeoTwawki: My thinking is that since Apururu sits back a far way at 20 yalm. thats where I expect 3 to be. He isn't nearly that far when I use him, so it has to at least be shorter than that.

sql/mob_pools.sql
TeoTwawki: reminder

thecursedcometh2: Oh crap, I'm sorry about that. I didn't see that on my first PR for Monberaux so I missed it this time.

TeoTwawki: its fairly new

zach2good:
Don't worry about this PR, it's highlighted how unclear and hard-to-find the Trust movement behaviours are. I'm going to refactor things so that these settings live in mobmods specific for trusts.

zach2good:
Refactor is done, you can see the mobmod being applied to some trusts now (and there's lots of comments explaining the different movement types): https://github.com/LandSandBoat/server/pull/4032

thecursedcometh2:
Should this be closed? It looks like #4032 did the same thing that I was trying to, just better.

zach2good:
Probably safe to close this PR, now that there are examples and lots of notes for how to set up distance logic for each trust. Generally, unimplemented trusts should just be brainless melee machines - as a giveaway that they aren't implemented yet. Putting in movement logic but not the rest of the implementation give the wrong idea I think

PR #4030, opened 2023-06-28
Fix Chainbound duration/delay to match retail (values pulled from skillchains addon)

nfaletra:
**_I affirm:_**

Chainbound status effect from Konzen-ittai and Wild Flourish were not functioning as they should. The window to use a weaponskill was incredibly short at ~2 seconds. Add latency on top of that and it was extremely difficulty to hit the skillchain reliably. I dug into the skillchains addon and copied the timings from there since I know they are accurate for retail from testing and experience.

Total Duration increased to 10 seconds. Initial Delay reduced to 2 seconds. This gives ~7 second window to skillchain after applying Chainbound.

Use Konzen-ittai or Wild Flourish followed by a weaponskill after ~2 seconds and observe a skillchain happens. Skillchain window should fall off after ~10 seconds total. Can test while using skillchains addon to have a visual guide for timings.

### DIFF
--- a/scripts/globals/abilities/konzen-ittai.lua
+++ b/scripts/globals/abilities/konzen-ittai.lua
@@ -23,7 +23,7 @@ abilityObject.onUseAbility = function(player, target, ability, action)
         not target:hasStatusEffect(xi.effect.CHAINBOUND, 0) and
         not target:hasStatusEffect(xi.effect.SKILLCHAIN, 0)
     then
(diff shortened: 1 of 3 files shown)


Xaver-DaRed (approved):
LGTM

TeoTwawki (approved):

claywar (approved):

PR #4032, opened 2023-06-28
[trust] Add and hook up TRUST_DISTANCE mobmod

zach2good:
**_I affirm:_**

Refactors trust movement settings to be a mobMod instead of piggybacking on the mob_pools.behavior field.

Summon Shantotto, Joachim, and Semih-Lafinah. !fafnir. Start fighting. See them reposition as they're meant to - with no errors in the logs etc.


zach2good:
Ready for review

scripts/globals/trust.lua
```diff
@@ -10,6 +10,17 @@ require("scripts/globals/roe")
 xi = xi or {}
 xi.trust = xi.trust or {}
 
+xi.trust.movementType =
```
claywar, line 13: Is this more of a type or a range?

claywar (approved):

scripts/globals/trust.lua
zach2good: It's both, so, naming is hard 🤷

PR #4043, opened 2023-07-03
[Quest] Converts Something Fishy quest to Interaction Framework

hooksta4:
**_I affirm:_**

Converts Something Fishy quest to Interaction Framework. 

Speak to NPC Tokaka in Port Windurst by the Fishing guild. Turn in Bastore Sardine, zone to repeat. 
Captured on July 3 2023. https://youtu.be/c0L3K_MQPXs


hooksta4:
Relevant capture starts at around 16:00 min mark of the YT video above.

scripts/quests/windurst/Something_Fishy.lua
```diff
@@ -0,0 +1,98 @@
...
+                    then
+                        return quest:event(210, 70, xi.items.BASTORE_SARDINE)
+                    end
+                end,
+
+                onTrigger = function(player, npc)
+                    if
+                        player:getQuestStatus(quest.areaId, quest.questId) == QUEST_COMPLETED and
+                        not quest:getMustZone(player) and
+                        quest:getVar(player, 'Dialog') < 1
```
claywar: There's a lot going on here, and given two different paths, they might be better off broken out into multiple sections (accepted, completed).  There's also no reason to use less than to compare a state that can only be 0 or 1.

scripts/quests/windurst/Something_Fishy.lua
```diff
@@ -0,0 +1,98 @@
...
+                end,
+            },
+
+            onEventFinish =
+            {
+                [208] = function(player, csid, option, npc)
+                    quest:setVar(player, 'Dialog', 1)
+                end,
+
+                [210] = function(player, csid, option, npc)
```
claywar, line 71: Similar to above, this would probably be cleaner to handle in a separate section for the completed version, and then use the reward table to handle the first time completion.

scripts/quests/windurst/Something_Fishy.lua
hookstar: The issue I ran into is during the capture, after first complete and subsequent "accepting" it doesn't go back into the accepted part of the log. Additionally, you have to get the dialog before trading the fish again after zone. All the cutscenes are the same for first or subsequent turn ins. Spent a decent amount of time trying to get it to match exactly how the capture went. 

As far as the >1 for some reason it would not trigger 208 with not getvar ==1 and there needs to be a discriminator between that and the next one. Up for suggestions to make it cleaner though. Thanks for reviewing!

hookstar: Although I suppose I could do a if quest ~= available and could get rid of those 2 lines...

scripts/quests/windurst/Something_Fishy.lua
hooksta4: That squash should be a lot cleaner -  tested several times as well and mirrors my capture.

scripts/quests/windurst/Something_Fishy.lua
```diff
@@ -0,0 +1,140 @@
...
+                    end
+                end,
+            },
+        },
+    },
+
+    {
+        check = function(player, status, vars)
+            return status ~= QUEST_AVAILABLE
+        end,
```
Xaver-DaRed, line 45: Isnt this dead code? Let me explain...
Quest can only be available, acceptd and completed and we already have specific sections for all 3 in the script.

hookstar: Yea I realized that they are not different. I will combine the two in the morning and squash.

hooksta4: Got it - little rusty been a few weeks since Ive done this. Thanks for the comments to make it right.

hooksta4:
updated to take out the require for items as the pr above this suggests.

scripts/quests/windurst/Something_Fishy.lua
KnowOne134: Yes this should be accepted status as you already have a completed status as well with same npcs

scripts/quests/windurst/Something_Fishy.lua
```diff
@@ -0,0 +1,114 @@
...
+
+        [xi.zone.PORT_WINDURST] =
+        {
+            ['Tokaka'] =
+            {
+                onTrigger = function(player, npc)
+                    if quest:getVar(player, 'Dialog') == 1 then
+                        return quest:event(209, 0, xi.items.BASTORE_SARDINE)
+                    elseif quest:getMustZone(player) then
+                        return quest:event(211)
```
KnowOne134, line 55: This is after completed code?

KnowOne134 (changes requested):
Agree with previous comments. The status checks should all be split and the logic put in the appropriate places. 
Could we use the standard var of Prog or Option for tracking and it self clears on quest complete

PR #4047, opened 2023-07-05
Add res-ranks and remove elemental SDT to mobs up to entry 514

Frankie-hz:
**_I affirm:_**

Same deal as PR https://github.com/LandSandBoat/server/pull/4038 and https://github.com/LandSandBoat/server/pull/4044

- Re-structures mob_resistances.sql and changes core to work with the changes
- Add res-ranks and remove elemental SDT to mobs up to entry 514.
- Separates duplicate resistant ids and more accurately defines what they are used for
- Add resistance rank values, according to this spreadsheet:
https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit#gid=1367309866

Import the table properly. See affected mobs get more (or less) resists in both nukes and status effects (spells) and be able in some cases to be immunobroken even.


sql/mob_resistances.sql
```diff
@@ -8,6 +8,7 @@ CREATE TABLE `mob_resistances` (
   `pierce_sdt` float NOT NULL DEFAULT 1,
   `h2h_sdt` float NOT NULL DEFAULT 1,
   `impact_sdt` float NOT NULL DEFAULT 1,
+  `magical_sdt` smallint(5) NOT NULL DEFAULT 0, -- General elemetnal SDT
```
WinterSolstice8: `elemental`

Frankie-hz: Fixed and pushed.

TeoTwawki:
> Add res-ranks and remove elemental SDT to mobs **up to entry 514.**

We're tackling it in chunks and thats why you had to move the sql handles instead of delete them yeah? Just double checking, I recall discussing removing them with Xaver because they weren't going to get used nearly as often as I originally thought when they were introduced.

Xaver-DaRed:
Yeah.
We spoke about it among ourselves. This PR adds the general magic SDT column, as we spoke, which is used commonly.

In a separate PR, we will deal with elemental SDTs, after seing if there is some value in them being there.
For instance, as an example, they could be repurposed or have added functionality for elemental nullification and absorbtion with ease.

So, for the time being, i think it's easier for them to stay.

Xaver-DaRed (approved):

TeoTwawki (approved):

zach2good (approved):

PR #4049, opened 2023-07-06
Correct duration on Drain II effect

CatsEyeXI:
**_I affirm:_**

Drain II's effect duration is incorrectly set to 60 seconds and should be 180 according to BGWiki
https://www.bg-wiki.com/ffxi/Drain_II

and according to ffxiclopedia:
https://ffxiclopedia.fandom.com/wiki/Drain_II


WinterSolstice8 (approved):

Xaver-DaRed (approved):

RAIST5150:
Just in case it is needed for future challenge, here is the jpwiki:

https://wiki.ffo.jp/html/11070.html

It links to the Sept. 6, 2015 patch notes where it was noted the duration was increased from 60 to 180 seconds.

PR #4053, opened 2023-07-06
[Enum] Move "items" file to enum folder

Xaver-DaRed:
**_I affirm:_**

Moves items file, containing item definitions, to enum folder.
- Removes unneeded requires.
- Adds "scripts/globals/items" path to CI's deprecated requires.

No real changes made.


TeoTwawki:
you used git-rename to move the file right, not just moved it in windows? its acting like it was deleted and recreated at the new location instead of preserving the history. Sometimes rebasing will lose rename action's, other times git is just a finicky sod for no reason at all. I really wish git had better file-move detection.

scripts/globals/mannequins.lua
```diff
@@ -2,7 +2,6 @@
 -- Information on Mannequins
 ----------------------------
 
-require("scripts/globals/items")
```
TeoTwawki, line 5: need to kill one of the line returns here

Xaver-DaRed: I THINK I just took care of all style changes needed after the mass deletion.We'll see what the CI says XD

WinterSolstice8 (approved):

claywar (approved):

Issue #4034, opened 2023-06-29
🐛 utils.lua:379: attempt to perform arithmetic on local 'levelTableIndex' (a nil value)


kaincenteno:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base`

## Steps to reproduce

some weapon skills are erroring with the same error

```
[06/29/23 02:11:46:949][map][error] luautils::onMobWeaponSkill (mobskill) ./scripts/globals/utils.lua:379: attempt to perform arithmetic on local 'levelTableIndex' (a nil value)
stack traceback:
        ./scripts/globals/utils.lua:379: in function 'getSkillLvl'
        ./scripts/globals/magic.lua:555: in function 'getMagicHitRate'
        ./scripts/globals/mobskills.lua:392: in function 'applyPlayerResistance'
        ./scripts/globals/mobskills.lua:771: in function 'mobPhysicalStatusEffectMove'
        ./scripts/globals/mobskills/biotic_boomerang.lua:22: in function <./scripts/globals/mobskills/biotic_boomerang.lua:14> (OnMobWeaponSkill:3952)

[06/29/23 02:10:07:370][map][error] luautils::onMobWeaponSkill (mobskill) ./scripts/globals/utils.lua:379: attempt to perform arithmetic on local 'levelTableIndex' (a nil value)
stack traceback:
        ./scripts/globals/utils.lua:379: in function 'getSkillLvl'
        ./scripts/globals/magic.lua:555: in function 'getMagicHitRate'
        ./scripts/globals/mobskills.lua:392: in function 'applyPlayerResistance'
        ./scripts/globals/mobskills.lua:771: in function 'mobPhysicalStatusEffectMove'
        ./scripts/globals/mobskills/aerial_collision.lua:20: in function <./scripts/globals/mobskills/aerial_collision.lua:13> (OnMobWeaponSkill:3952)

[...]

PR #4061, opened 2023-07-08
Attempt to fix nil values in utils.getSkillLvl

Xaver-DaRed:
**_I affirm:_**

Adds fallbacks to prevent nil values in utils.getSkillLvl.
Issue #4034 

Let the mobs in the linked issue perform magic/mobabilities and not get errors in console.

### DIFF
--- a/scripts/globals/utils.lua
+++ b/scripts/globals/utils.lua
@@ -341,34 +341,38 @@ end
 local skillLevelTable =
 {
     --         A+             A-             B+             B              B-             C+             C              C-             D              E              F             G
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
Looks good but asking jimmy for the non-ranked skills for jobs in case he knows

TeoTwawki:
I'm kinda expecting every mob has max skillranks in whatever they use regardless of job, as SE is fond of having mashups bolted on that would otherwise require more than 2 jobs. or at least the ability to bolt on skill ranks separately in a system similar to pankration plates.

claywar (approved):

zach2good (approved):

zach2good:
_yeet_

PR #4064, opened 2023-07-09
Cache zone settings for world server messages

xkoredev:
**_I affirm:_**

Caches zone settings to avoid message server sql queries when not needed.

Send tells to someone, run !updateconquest.

### DIFF
--- a/src/world/message_server.cpp
+++ b/src/world/message_server.cpp
@@ -22,13 +22,11 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include <concurrentqueue.h>
 #include <memory>
 #include <queue>
+#include <set>
 
 #include "common/logging.h"
 #include "conquest_system.h"
 #include "message_server.h"
-
-zmq::context_t                 zContext;
-std::unique_ptr<zmq::socket_t> zSocket;
 
 struct chat_message_t
 {
@@ -38,17 +36,26 @@ struct chat_message_t
     zmq::message_t packet;
 };
 
-moodycamel::ConcurrentQueue<chat_message_t> outgoing_queue;
+struct zone_settings_t
+{
+    uint16 zoneid;
+    uint64 ipp;
+    uint32 misc;
+};
 
-std::unique_ptr<SqlConnection>                                        sql;
-std::unordered_map<REGIONALMSGTYPE, std::shared_ptr<IMessageHandler>> regionalMsgHandlers;
+namespace
+{
+    zmq::context_t                 zContext;
+    std::unique_ptr<zmq::socket_t> zSocket;
 
-typedef std::function<bool(zmq::message_t* data,
-                           zmq::message_t* packet,
-                           zmq::message_t* from,
-                           in_addr         from_addr,
-                           uint16          from_port)>
-    HandleMessageFunc;
+    moodycamel::ConcurrentQueue<chat_message_t> outgoing_queue;
+
+    std::unique_ptr<SqlConnection>                                        sql;
+    std::unordered_map<REGIONALMSGTYPE, std::shared_ptr<IMessageHandler>> regionalMsgHandlers;
+    std::unordered_map<uint16, zone_settings_t>                           zoneSettingsMap;
+    std::vector<uint64>                                                   mapEndpoints;
+    std::vector<uint64>                                                   yellMapEndpoints;
+} // namespace
 
 void queue_message(uint64 ipp, MSGSERVTYPE type, zmq::message_t* extra, zmq::message_t* packet)
 {
@@ -67,29 +74,9 @@ void queue_message(uint64 ipp, MSGSERVTYPE type, zmq::message_t* extra, zmq::mes
 
 void queue_message_broadcast(MSGSERVTYPE type, zmq::message_t* extra, zmq::message_t* packet)
 {
-    // TODO: Cache these on world server start
-    const char* query = "SELECT zoneip, zoneport FROM zone_settings GROUP BY zoneip, zoneport;";
-    int         ret   = sql->Query(query);
-    uint64      ip    = 0;
-
-    if (ret == SQL_ERROR)
+    for (const auto& ipp : mapEndpoints)
     {
-        return;
-    }
-
-    while (sql->NextRow() == SQL_SUCCESS)
-    {
-        inet_pton(AF_INET, (const char*)sql->GetData(0), &ip);
-        uint64  port = sql->GetUIntData(1);
-        in_addr target;
-        target.s_addr = ip;
-
-        char target_address[INET_ADDRSTRLEN];
-        inet_ntop(AF_INET, &target, target_address, INET_ADDRSTRLEN);
-        ShowDebug(fmt::format("Message: -> Sending to {}:{}", target_address, port));
-        ip |= (port << 32);
-
-        queue_message(ip, type, extra, packet);
+        queue_message(ipp, type, extra, packet);
     }
 }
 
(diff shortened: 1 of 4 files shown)


xkoredev:
@zach2good

src/world/zone_settings.h
```diff
@@ -0,0 +1,13 @@
+#ifndef SERVER_ZONE_SETTINGS_H
+#define SERVER_ZONE_SETTINGS_H
+
+#include "../common/cbasetypes.h"
+
+struct zone_settings_t
```
zach2good: No need for a whole new file here, if this struct is only being used in message_server.cpp

src/world/message_server.cpp
```diff
@@ -110,12 +88,23 @@ void message_server_send(uint64 ipp, MSGSERVTYPE type, zmq::message_t* extra, zm
     }
 }
 
+std::string ipp_to_string(uint64 ipp)
+{
+    uint32  port = (uint32)((ipp & 0xFFFFFFFF00000000LL) >> 32);
```
zach2good: No need to mask here, binary shift doesn't act like a rotate. Though, I would leave comments here saying you're taking the top/bottom 32 bytes to do different things here

src/world/message_server.cpp
```diff
@@ -110,12 +88,23 @@ void message_server_send(uint64 ipp, MSGSERVTYPE type, zmq::message_t* extra, zm
     }
 }
 
+std::string ipp_to_string(uint64 ipp)
```
zach2good, line 100: 👍

src/world/message_server.cpp
```diff
@@ -42,13 +44,9 @@ moodycamel::ConcurrentQueue<chat_message_t> outgoing_queue;
...
 std::unique_ptr<SqlConnection>                                        sql;
 std::unordered_map<REGIONALMSGTYPE, std::shared_ptr<IMessageHandler>> regionalMsgHandlers;
-
-typedef std::function<bool(zmq::message_t* data,
-                           zmq::message_t* packet,
-                           zmq::message_t* from,
-                           in_addr         from_addr,
-                           uint16          from_port)>
-    HandleMessageFunc;
+std::unordered_map<uint16, zone_settings_t>                           zoneSettingsMap;
```
zach2good: Can you slap all of these .cpp-specific objects into an anonymous namespace while you're here please?

src/world/message_server.cpp
```diff
@@ -195,9 +190,18 @@ void message_server_parse(MSGSERVTYPE type, zmq::message_t* extra, zmq::message_
...
+            uint16 zoneId = ref<uint16>((uint8*)extra->data(), 2);
+            try
+            {
+                auto zoneSettings = zoneSettingsMap.at(zoneId);
+                ShowDebug(fmt::format("Message: -> rerouting to {}", ipp_to_string(zoneSettings.ipp)));
+                message_server_send(zoneSettings.ipp, type, extra, packet);
+            }
+            catch (const std::out_of_range&)
+            {
+                ShowError("Requested ZoneId not in cache: %d", zoneId);
```
zach2good, line 212: 👍

src/world/message_server.cpp
```diff
@@ -230,39 +234,23 @@ void message_server_parse(MSGSERVTYPE type, zmq::message_t* extra, zmq::message_
         }
     }
 
+    // This is only used for cases where SQL is used to get the IPs (not cached).
+    // E.g: When we get ips for a specific account_session.
+    // Consider
```
zach2good: Consider?

src/world/message_server.cpp
```diff
@@ -308,6 +296,46 @@ void message_server_listen(const bool& requestExit)
...
+        inet_pton(AF_INET, (const char*)sql->GetData(1), &ip);
+        uint64 port = sql->GetUIntData(2);
+
+        zone_settings_t zone_settings;
+        zone_settings.zoneid = sql->GetUInt64Data(0);
+        zone_settings.ipp    = ip | (port << 32);
+        zone_settings.misc   = sql->GetUIntData(3);
+
+        zoneIppSet.insert(zone_settings.ipp);
+        if (zone_settings.misc & 1024)
```
zach2good, line 336: Presumably there's an enum somewhere to make it clearer what this 1024 is used for

src/world/message_server.cpp
```diff
@@ -308,6 +296,46 @@ void message_server_listen(const bool& requestExit)
...
+    }
+
+    // Keep track of the zones, as well as a list of unique ip / port combinations.
+    std::set<uint64> zoneIppSet;
+    std::set<uint64> yellZoneIppSet;
+    zoneSettingsMap = std::unordered_map<uint16, zone_settings_t>(sql->NumRows());
+    while (sql->NextRow() == SQL_SUCCESS)
+    {
+        uint64 ip = 0;
+        inet_pton(AF_INET, (const char*)sql->GetData(1), &ip);
```
zach2good: We have a sql->GetString() thing you can use here instead of a c-style cast

src/world/message_server.cpp
```diff
@@ -308,6 +296,46 @@ void message_server_listen(const bool& requestExit)
...
+    int         ret   = sql->Query(query);
+
+    if (ret == SQL_ERROR)
+    {
+        ShowError("Error loading zone settings from DB. World server will not send messages.");
+        return;
+    }
+
+    // Keep track of the zones, as well as a list of unique ip / port combinations.
+    std::set<uint64> zoneIppSet;
```
zach2good: I wonder if zoneIpp is the best name for this. Its what's being stored, sure, but it represents a ip/port pair, which is how we identify processes on a machine, each of which can hold multiple zones. I don't have a better name suggestion, but its worth mentioning.

src/world/message_server.cpp
```diff
@@ -308,6 +296,46 @@ void message_server_listen(const bool& requestExit)
...
 }
 
+void cache_zone_settings()
+{
+    const char* query = "SELECT zoneid, zoneip, zoneport, misc FROM zone_settings;";
+    int         ret   = sql->Query(query);
+
+    if (ret == SQL_ERROR)
+    {
+        ShowError("Error loading zone settings from DB. World server will not send messages.");
```
zach2good: If this kills the world server, I'd upgrade this to ShowFatal, or ShowException (if that exists?), log very loudly, and then actually throw a std::runtime_error with a helpful message

src/world/message_server.cpp
```diff
@@ -22,10 +22,12 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include <concurrentqueue.h>
 #include <memory>
 #include <queue>
+#include <set>
```
zach2good, line 25: Presumably this is already in the pch too yeah?

src/world/zone_settings.h
```diff
@@ -0,0 +1,13 @@
+#ifndef SERVER_ZONE_SETTINGS_H
+#define SERVER_ZONE_SETTINGS_H
+
+#include "../common/cbasetypes.h"
```
zach2good: General note: (aside from one folder in the map/ai/ subfolder) we don't need to use relative paths anymore, `common/cbasetypes.h` will work where it didn't all the way back in DSP

src/world/message_server.cpp
xkoredev: `mapEndpointSet`?

src/world/message_server.cpp
xkoredev: I was having compile time issues without the include :S

src/world/message_server.cpp
zach2good: Sure

src/world/message_server.cpp
zach2good: It's already in pch.h, nothing additional to do here

src/world/zone_settings.h
xkoredev: Removed file

src/world/message_server.cpp
xkoredev: Done

src/world/message_server.cpp
xkoredev: Done and thrown

src/world/zone_settings.h
xkoredev: Good to know. Will make sure I don't add them in future locations. Removed this file anyway.

src/world/message_server.cpp
xkoredev: Done. thanks.

src/world/message_server.cpp
xkoredev: Done

src/world/message_server.cpp
xkoredev: Renamed `zoneIpp` to `mapEndpoint` in all instances

zach2good:
```
kore — Today at 01:00
ok so both pipelines passing, except one doesnt pass due to the sql not being cached (which is in the other PR). I suggest merging both then checking pipeline on feature branch, but should work
```
_yeet (into feature branch)_!

PR #4065, opened 2023-07-09
Improve functionality of !pos and !reset

CatsEyeXI:
**_I affirm:_**

Modifies !reset to remove petrification.
Modifies !pos to include player's current rotation

Use each command

### DIFF
--- a/scripts/commands/pos.lua
+++ b/scripts/commands/pos.lua
@@ -68,7 +68,7 @@ function onTrigger(player, arg1, arg2, arg3, arg4, arg5)
 
     -- report or move position
     if x == nil or y == nil or z == nil then
-        player:PrintToPlayer(string.format("%s's position: X %.4f  Y %.4f  Z %.4f", targ:getName(), targ:getXPos(), targ:getYPos(), targ:getZPos()))
+        player:PrintToPlayer(string.format("%s's position: X %.4f  Y %.4f  Z %.4f  Rot %i", targ:getName(), targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos()))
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):

scripts/commands/pos.lua
```diff
@@ -68,7 +68,7 @@ function onTrigger(player, arg1, arg2, arg3, arg4, arg5)
 
     -- report or move position
     if x == nil or y == nil or z == nil then
-        player:PrintToPlayer(string.format("%s's position: X %.4f  Y %.4f  Z %.4f", targ:getName(), targ:getXPos(), targ:getYPos(), targ:getZPos()))
+        player:PrintToPlayer(string.format("%s's position: X %.4f  Y %.4f  Z %.4f  Rot %i", targ:getName(), targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos()))
```
claywar, line 71: Looks like you're just reporting here, was there an intent to also add an optional parameter to send as well?  Fine either way, but could easily be added as well.

CatsEyeXI: the intention was to report only, as its helpful in creating dynamic entities and other things. i can't really see a reason why i'd want to send a player to a specific rotation

TeoTwawki: I think we'd actually have problems adding rot as input, because we would normally have it before the zone param - and then all our existing uses would need updated and thats a huge pain point for NPC scripts with pos in the header (which is there because of the MASSIVE convenience it provides)

TeoTwawki (approved):

PR #4072, opened 2023-07-10
Add Influence cache and reorganize time server

xkoredev:
**_I affirm:_**

- No longer sends influence updates every time a map server updates it upstream
- Send influence updates periodically via time server
- Move time server to a separate file
- Move gobbie box logic to a separate file

Stay online while vana hour, hour, week and jp midnight updates happen.
Kill shit and check that influence is updated in the region map after 2.4 minutes (1 vana hour)

### DIFF
--- a/src/world/gobbie_box.cpp
+++ b/src/world/gobbie_box.cpp
@@ -1,0 +1,32 @@
+#include "gobbie_box.h"
+
+namespace gobbiebox
+{
+    void UpdateDailyTallyPoints(WorldServer* worldServer)
+    {
+        uint16 dailyTallyLimit  = settings::get<uint16>("main.DAILY_TALLY_LIMIT");
+        uint16 dailyTallyAmount = settings::get<uint16>("main.DAILY_TALLY_AMOUNT");
+
+        const char* fmtQuery = "UPDATE char_points \
+            SET char_points.daily_tally = LEAST(%u, char_points.daily_tally + %u) \
+            WHERE char_points.daily_tally > -1;";
+
+        int32 ret = worldServer->sql->Query(fmtQuery, dailyTallyLimit, dailyTallyAmount);
+
+        if (ret == SQL_ERROR)
+        {
+            ShowError("Failed to update daily tally points");
+        }
+        else
+        {
+            ShowDebug("Distributed daily tally points");
+        }
+
+        fmtQuery = "DELETE FROM char_vars WHERE varname = 'gobbieBoxUsed';";
+
+        if (worldServer->sql->Query(fmtQuery, dailyTallyAmount) == SQL_ERROR)
+        {
+            ShowError("Failed to delete daily tally char_vars entries");
+        }
+    }
+} // namespace gobbiebox
--- a/src/world/world_server.cpp
+++ b/src/world/world_server.cpp
@@ -22,132 +22,17 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 
 #include "common/application.h"
 #include "common/logging.h"
-#include "common/vana_time.h"
-
-namespace
-{
-    std::unique_ptr<SqlConnection>  sql;
-    std::shared_ptr<ConquestSystem> conquest;
-} // namespace
-
-void UpdateDailyTallyPoints()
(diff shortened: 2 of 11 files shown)


xkoredev:
@zach2good

src/world/gobbie_box.cpp
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        fmtQuery = "DELETE FROM char_vars WHERE varname = 'gobbieBoxUsed';";
+
+        if (worldServer->sql->Query(fmtQuery, dailyTallyAmount) == SQL_ERROR)
+        {
+            ShowError("Failed to delete daily tally char_vars entries");
+        }
+    }
+}
```
zach2good: Newline at end of file

src/world/world_server.cpp
```diff
@@ -22,132 +22,18 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
 , httpServer(std::make_unique<HTTPServer>())
 , messageServer(std::make_unique<message_server_wrapper_t>(std::ref(m_RequestExit)))
 {
-    // Anonymous namespace preparation
-    sql      = std::make_unique<SqlConnection>();
-    conquest = conquestSystem;
-
     // Tasks
-    CTaskMgr::getInstance()->AddTask("time_server", server_clock::now(), nullptr, CTaskMgr::TASK_INTERVAL, time_server, 2400ms);
+    std::shared_ptr<WorldServer> worldPtr(this);
```
zach2good: `world/main.cpp`
```cpp
int main(int argc, char** argv)
{
    auto pWorldServer = std::make_unique<WorldServer>(argc, argv);

    while (pWorldServer->IsRunning())
    {
        pWorldServer->Tick();
    }

    return 0;
}
```

WorldServer is already managed by a unique_ptr that's tied to the lifetime of the program. CTaskMgr is all static because it's set up as a singleton, so you don't need to worry about the lifetime of any of these object, so it's totally OK to just use `this` here.

If you WERE going to use shared_ptr from this here (if WorldServer was created by a shared_ptr in the first place), you'd need to use: https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

zach2good (commented):
Very close, just needs some styling and that ptr change

(no need to re-ping me, I get emails about everything that happens in the repo)

zach2good:
Even if this is going into a feature branch, the commit title (the body looks good) still needs to be meaningful: https://github.com/LandSandBoat/server/wiki/Development-Guide-Git#writing-useful-commit-messages

src/world/world_server.cpp
xkoredev: Done. Makes sense.

src/world/gobbie_box.cpp
xkoredev: Removed.

zach2good (approved):

zach2good:
Strange that CI is dinging you for these, given that this isn't your code?

```
D:\a\server\server\src\map\message.cpp(685,47): error C2220: the following warning is treated as an error [D:\a\server\server\build\src\map\xi_map.vcxproj]
D:\a\server\server\src\map\message.cpp(685,47): warning C4018: '<': signed/unsigned mismatch [D:\a\server\server\build\src\map\xi_map.vcxproj]
D:\a\server\server\src\map\message.cpp(706,47): warning C4018: '<': signed/unsigned mismatch [D:\a\server\server\build\src\map\xi_map.vcxproj]
D:\a\server\server\src\map\message.cpp(728,47): warning C4018: '<': signed/unsigned mismatch [D:\a\server\server\build\src\map\xi_map.vcxproj]
```

Could you fix them while you're here? It's normally a case of swapping in some unsigned types like std::size_t in for a loop var or something, or auto

xkoredev:
> 

Done!

zach2good:
```
kore — Today at 01:00
ok so both pipelines passing, except one doesnt pass due to the sql not being cached (which is in the other PR). I suggest merging both then checking pipeline on feature branch, but should work
```
_yeet (into feature branch)_!

PR #4080, opened 2023-07-10
Fix range and logic for `MAX_ZONE` in GM commands

claywar:
**_I affirm:_**

GM commands allowed for an invalid zone (300) to be used for !send and !zone.  Reduces the lua definition of `MAX_ZONE` to 299, and applies correct conditional logic

Try to zone or send player to zone 300 (fail)
Repeat procedure with zone 299 (success)

### DIFF
--- a/scripts/commands/zone.lua
+++ b/scripts/commands/zone.lua
@@ -333,7 +333,7 @@ function onTrigger(player, bytes)
     else
         -- destination is a zone ID.
         zone = tonumber(bytes)
-        if zone == nil or zone < 0 or zone >= xi.zone.MAX_ZONE then
+        if zone == nil or zone < 0 or zone > xi.zone.MAX_ZONE then
             error(player, "Invalid zone ID.")
             return
         end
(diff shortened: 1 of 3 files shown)


zach2good (approved):

Xaver-DaRed (approved):

scripts/commands/zone.lua
```diff
@@ -333,7 +333,7 @@ function onTrigger(player, bytes)
     else
         -- destination is a zone ID.
         zone = tonumber(bytes)
-        if zone == nil or zone < 0 or zone >= xi.zone.MAX_ZONE then
```
TeoTwawki, line 336: `>= 300` .. That should have stopped it the same as `> 299` , I'm confused here.

side note: This +1 from the actual highest ID thing is the same pattern that has been used with every MAX_WHATEVER since the old PXI days to, so I suggest checking the rest for essential updates and consistency lest contributors be confused on usage and introduce new errors in the future.

claywar: 300 is an invalid zone ID

claywar:
Bah, lack of coffee.  Carry on

PR #4083, opened 2023-07-10
Perpetuance job ability implementation

nfaletra:
**_I affirm:_**

Logic for perpetuance increasing enhancing duration was implemented but the status effect and removal logic was missing. Added the ability to the db with the correct light arts flags and removal of the status effect on the next white magic enhancing spell.

Also fixed light arts status effect flags as it was not properly being removed when logging out. I double checked that behavior on retail to make sure I wasn't crazy.

* Cast enhancing magic on yourself (e.g. Regen)
* Use Lights Arts -> Perpetuance
* Cast same enhancing magic on yourself
* Observe the duration is doubled compared to the first cast


scripts/globals/abilities/perpetuance.lua
```diff
@@ -0,0 +1,34 @@
...
+    return 0, 0
+end
+
+abilityObject.onUseAbility = function(player, target, ability)
+    player:addStatusEffect(xi.effect.PERPETUANCE, 1, 0, 60)
+
+    return xi.effect.PERPETUANCE
+end
+
+return abilityObject
```
Xaver-DaRed: Need a newline at the end of the file, for CI.

src/map/utils/charutils.cpp
```diff
@@ -5912,6 +5912,10 @@ namespace charutils
             {
                 PChar->StatusEffectContainer->DelStatusEffect(EFFECT_ACCESSION);
             }
+            if (PSpell->getSkillType() == SKILL_ENHANCING_MAGIC)
+            {
```
Xaver-DaRed, line 5916: Correct me if I'm wrong, but wouldnt this need a check for the spell group, so only white magic enhancing skills (AKA, not spikes) count?

nfaletra: it's within a spell group check.

scripts/globals/abilities/perpetuance.lua
nfaletra: Done and done.

src/map/utils/charutils.cpp
Xaver-DaRed: Very true, I saw it. I stand corrected

Xaver-DaRed (approved):

claywar (approved):

PR #4086, opened 2023-07-10
[Fix] Add safety for nil values to immunobreak

Xaver-DaRed:
**_I affirm:_**

Fix oversight when coding immunobreak, when using non immunobreakable status effects.

Cast enfeebling spells (Notably Drown, Burn and other elemental DoTs) without errors.

### DIFF
--- a/scripts/globals/spells/enfeebling_spell.lua
+++ b/scripts/globals/spells/enfeebling_spell.lua
@@ -376,8 +376,12 @@ xi.spells.enfeebling.useEnfeeblingSpell = function(caster, target, spell)
     local resistStages = pTable[spellId][8]
     local message      = pTable[spellId][9]
     local bonusMacc    = pTable[spellId][12]
-    local resistRank   = target:getMod(xi.combat.element.resistRankMod[spellElement]) or 0
-    local rankModifier = target:getMod(immunobreakTable[spellEffect][1]) or 0
+    local rankModifier = 0
+
+    -- Fetch immunobreak modifier to resistance rank if aplicable.
+    if immunobreakTable[spellEffect] then
+        rankModifier = target:getMod(immunobreakTable[spellEffect][1])
+    end
 
     -- Magic Hit Rate calculations.
     local magicAcc     = xi.combat.magicHitRate.calculateActorMagicAccuracy(caster, target, spellGroup, skillType, spellElement, statUsed, bonusMacc)
@@ -404,7 +408,13 @@ xi.spells.enfeebling.useEnfeeblingSpell = function(caster, target, spell)
 
     -- The effect will get resisted.
     if resistRate <= 1 / (2 ^ resistStages) then
-        -- Attempt immunobreak.
+        -- Attempt immunobreak. Fetch resistance rank modifier.
+        local resistRank = 0
+
+        if spellElement ~= xi.magic.ele.NONE then
+            resistRank = target:getMod(xi.combat.element.resistRankMod[spellElement])
+        end
+
         if
             caster:isPC() and
             target:isMob() and


scripts/globals/spells/enfeebling_spell.lua
```diff
@@ -376,8 +376,12 @@ xi.spells.enfeebling.useEnfeeblingSpell = function(caster, target, spell)
     local resistStages = pTable[spellId][8]
     local message      = pTable[spellId][9]
     local bonusMacc    = pTable[spellId][12]
-    local resistRank   = target:getMod(xi.combat.element.resistRankMod[spellElement]) or 0
-    local rankModifier = target:getMod(immunobreakTable[spellEffect][1]) or 0
+    local rankModifier = 0
+
+    -- Fetch immunobreak modifier to resistance rank if aplicable.
+    if immunobreakTable[spellEffect] then
+        rankModifier = target:getMod(immunobreakTable[spellEffect][1])
```
zach2good, line 383: Without me looking at the implementation, what does `getMod` return if it's asked to look up a `nil` value?

Xaver-DaRed: getMod cannot return anything in this case if `immunobreakTable[spellEffect]` is nil, because we would be trying to index a nil value (table_name[row we want which is nil][Column we want])

I first tried with `var = immunobreakTable[spellEffect][1] or 0` however, that still errored and screwed the whole thing.

So this is what we are fixing here
imagen

TeoTwawki: getMod used to implode the exe on nils. might gracefully exit now, but we'd still want to not call it on nil so its good to have that `if` first

zach2good: That error means that `immunobreakTable[spellEffect]` is returning nil, and then you're trying to index into it. So you'd probably want to do something like
```
local type = immunobreakTable[spellEffect] or { SOME BASE TYPE TABLE }
rankModifier = type[1]
```

Xaver-DaRed: The way this works is:
- Only a handful of status effects can be immunobroken. The rest, which are the majority, cant.
- The immunobreak table points the status effects (that can be immunobroken) to a set of modifiers that keep track of the ammount of times that a concrete status effect has been immunobroken.
- So the table serves 2 purposes:
    1. Knowing which status effects can be immunobroken
    2. Knowing which modifier keeps track of the ammount of times immunobreak has happened to an specific status effect.

We are looking for a numerical value at the end of the day. Most times it will be 0. And the exceptions are the immunobreakable status effcts.

The simplest solution was to check if `immunobreakTable[spellEffect]` exists. If it does, then an associated modifier exists so we fetch its value. If not, we do nothing, since the default value is already 0.

claywar: It looks like CBattleEntity::getMod can still implode if an invalid index is passed, and we don't have any safety in lua land before calling it.  I agree with Xaver's implementation, since an invalid index is nil-checked prior to the call, and only valid items should exist in that table.

Now, we should absolutely add some safety in the lua binding as well

zach2good: if the binding takes a number and a nil is passed to it, sol _SHOULD_ fail gracefully and not fire off the binding, since the argument didn't match

zach2good:
```cpp
int main()
{

	sol::state lua;
	lua.open_libraries(sol::lib::base, sol::lib::package);
    lua.set_function("func", [](int input){ return input; });

    lua.script("func(nil)");

	return 0;
}
```
```
[sol2] An error occurred and has been passed to an error handler: sol: runtime error: stack index 1, expected number, received nil: not a numeric type that fits exactly an integer (number maybe has significant decimals) (bad argument into 'int(int)')
stack traceback:
	[C]: in function 'base.func'
	[string "func(nil)"]:1: in main chunk
terminate called after throwing an instance of 'sol::error'
  what():  sol: runtime error: stack index 1, expected number, received nil: not a numeric type that fits exactly an integer (number maybe has significant decimals) (bad argument into 'int(int)')
stack traceback:
	[C]: in function 'base.func'
	[string "func(nil)"]:1: in main chunk
	```

claywar (approved):

zach2good (approved):

PR #4090, opened 2023-07-11
Merge `feature/conquest_world_server` into `base`

zach2good:
**_I affirm:_**

This is mainly for my own sanity to do another full review pass before merging into `base`. I think all points in all the other PRs are addressed and I'm happy with everything


zach2good:
Anything I find that wants changing I can quickly fixup, now that this is all in an LSB-owned branch

Xaver-DaRed:
This is also moving gobbie box daily tally to world, isnt it?
I ask because, while I do think it belongs to world, it isnt part of the conquest system

claywar:
Its a lot more than conquest if you look at the time functions, so there's a bit of a disconnect in the branch name, though looking reasonable so far.  I would suggest perhaps renaming gobbie box to daily tally or something along those lines to be more specific

zach2good:
I asked for the refactoring of time_server and gobbie box stuff as part of these PRs, to make the setup for conquest/besieged etc. easier in the future. 

gobbiebox/dailytally was already part of world server (since it only wants to be ran once per period). I asked for it to be renamed to gobbiebox, as thats the front-facing gameplay mechanic for accessing daily tally points. We can revert that back though if people want

claywar (approved):

TeoTwawki:
"gobbiebox" is the npc part players interact with. the actual points being stored is referred to as daily tally both by the in game currency menu and by the npc dialogs. So when we're dealing with that number value part, makes more sense to call it by what SE did.

zach2good:
- Added missing GPL headers to all files
- Used `#pragma` once at the top of headers
- Removed small data type file and added `influence_t` and `region_control_t` to `conquest_data.h` (there was a circular header dependency that was blocking that, so I fixed it)
- `gobbiebox::` -> `dailytally::`
- Jiggled a couple of other things I noticed around
- Messed up my interactive rebase, so I rebased again onto current main
- `./tools/run_clang_format.sh`

zach2good:
Do not merge, I have some changes I want to make in ZMQ message handling

zach2good:
OK, ready to go

Xaver-DaRed (approved):

PR #4097, opened 2023-07-12
Improves !pos command by handling commas

loxleygames:
**_I affirm:_**

* Improves `!pos` command to make it more usable for developers:

Formats now handled:
* `!pos 1.0 2.0 3.0`
* `!pos 1.0, 2.0, 3.0`
* `!pos 1.0, 2.0, 3.0,`
* `!pos 1.0,2.0,3.0`
* `!pos 1.0,2.0,3.0,`
* `!pos 1 2 3`
* `!pos 1, 2, 3`
* `!pos 1, 2, 3,`
* `!pos 1,2,3`
* `!pos 1,2,3,`

You can now easily paste positions from SQL or Lua, without needing to manually remove commas.

* Adds zone ID to the default `!pos`, when displaying current position.
* Changes message to use `xi.msg.channel.SYSTEM_3` for much more readable text.

Run `!pos` again to verify the new position is correct, or append print after setting position eg.
```lua
zoneId = targ:getZoneID()
targ:setPos(x, y, z, 0)
player:PrintToPlayer(string.format("%s's position: X %.4f  Y %.4f  Z %.4f  Rot %i  (Zone: %i)", targ:getName(), targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos(), targ:getZoneID()), xi.msg.channel.SYSTEM_3)
```


scripts/commands/pos.lua
```diff
@@ -68,7 +87,7 @@ function onTrigger(player, arg1, arg2, arg3, arg4, arg5)
 
     -- report or move position
     if x == nil or y == nil or z == nil then
-        player:PrintToPlayer(string.format("%s's position: X %.4f  Y %.4f  Z %.4f  Rot %i", targ:getName(), targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos()))
+        player:PrintToPlayer(string.format("%s's position: X %.4f  Y %.4f  Z %.4f  Rot %i  (Zone: %i)", targ:getName(), targ:getXPos(), targ:getYPos(), targ:getZPos(), targ:getRotPos(), targ:getZoneID()), xi.msg.channel.SYSTEM_3)
```
TeoTwawki, line 72: PrintToPlayer already prints to SYSTEM without it being specified.

TeoTwawki:
I'm on the fence about doing this. Because In general, i would like to globally separate parameters in some way in the future rather than for just one command using lua script methods. not necessarily always by comma, but possibly. There are many instances where breaking it up by spaces is a pain point. Say for instance you wanted a GM command that did PrintToArea with your input as the message text. Every word in your sentence winds up being a separate parameter.

I understand in this pr you are simply striping the commas off the input so that the input becomes valid again for toNumber. But I'm not sure we want to establish that pattern. I guess I'll wait for other LSB members input.

We had stripped comma's out of the lua headers ages ago for this, but I see the utility of copypasta from sql file.

scripts/commands/pos.lua
loxleygames: Yeah but that's purple on blue (And not much better on the others) and it's hard to read. When you're trying to check lots of positions, it really sucks. `SYSTEM_3` is very readable on every background.

LoxleyX:
True but this is very common use case, and would benefit many developers from now until such time when there is an alternative solution to parse the args.

Btw checks failed due to the `,` inside of strings. Not sure this is avoidable.

scripts/commands/pos.lua
TeoTwawki: oh, I missed that it was system 3, not system. Yellow by default unless one edits it in client - I use that yellow text for a lot of things, also for getting with of the header that regular system comes with.

zach2good:
I would approach this a different way: change `    parameters = "sssss"` to `    parameters = "s"` so it interprets the entire input as a single string, then do string manipulation to replace everything that isn't a number or a decimal point with a space. Then deduplicate all parts of the string that have multiple spaces, split on the remaining spaces, then use `tonumber` on the parts you have left.

zach2good:
> Btw checks failed due to the , inside of strings. Not sure this is avoidable.

Try putting your strings in single quotes, I dunno if our CI cares about that or not in this case

TeoTwawki:
> I would approach this a different way: change ` parameters = "sssss"` to ` parameters = "s"` so it interprets the entire input as a single string, then do string manipulation to replace everything that isn't a number or a decimal point with a space. Then deduplicate all parts of the string that have multiple spaces, split on the remaining spaces, then use `tonumber` on the parts you have left.

changing the number of params in the settings table in the script won't make it take the whole input as 1 string, the core will still break it up and then the additional "params" it thinks it has get discarded. that might work in cases where the input has no whitespace whatsoever but `123, 465, 789` will stop at `123,` for its input.

TeoTwawki:
to add to this, even if we're enabling `123,456,789` as input I do NOT want to break using `123 456 789` so we absolutely need multiple @zach2good

TeoTwawki:
> case 's':
   __     **if (cmdparameters.size() == 1)**

apparently undoc'd feature Zach is aware of and I am not.

> so `ss` with "`mary had a little lamb`" will receive "`mary` `had`" as 2 params, but you are telling me if its only s it'll get the whole thing.

test in game, if it works it works. if it doesn't it doesn't.

zach2good:
Some more detail:

- Replace commas with spaces
- Dedupe all the spaces that might be there
- Split on spaces

We have:
```
-- utils.splitStr("a.b.c", ".") => { "a", "b", "c" }
function utils.splitStr(s, sep)
...
```

and when you key into a lua table and nothing is there, you get back nil.

So if you key'd 4 and 5 for the optional args, you can just nil check them or use the OR pattern to add defaults etc.

While you're doing this, you could add any string manipulation functions you make/use to utils.lua

zach2good:
Also, will need to trim leading/following spaces, that'll need a util - Lua doesn't come with much string manip

LoxleyX:
Thanks both for your suggestions. On @zach2good's advice, I've combined the command into a single argument which is cleaned, trimmed by utils, then split, Following @TeoTwawki's concern about reuse in other commands, I also put the function itself into utils, so now it's just a single call to use elsewhere.

I've tested on all the above conditions listed in the PR and also any amount of whitespace between params still works. This could be useful when copying from lua tables that use extra spaces to align the numbers inside a table

ps. I updated checks to allow quoted comma, which may help in future cases also.

zach2good:
@TeoTwawki did you have anything you wanted to follow up on here?

tools/ci/lua_stylecheck.py
```diff
@@ -94,7 +94,7 @@ def check_parameter_padding(self, line):
         """
         # ,[^ \n] : Any comma that does not have space or newline following
 
-        for _ in re.finditer(",[^ \n]", line):
+        for _ in re.finditer(",[^ \n\"]", line):
```
zach2good: @claywar can you verify this doesn't do anything spicy?

TeoTwawki:
> @TeoTwawki did you have anything you wanted to follow up on here?

nope but I'll add I'm pleased with the new utils

zach2good (approved):

tools/ci/lua_stylecheck.py
zach2good: Also, the comment above this will need to be updated

claywar: This breaks CI causing the case of `(param1,"param2)` to pass
image

tools/ci/lua_stylecheck.py
```diff
@@ -92,9 +92,9 @@ def check_parameter_padding(self, line):
 
         See: TBD
         """
-        # ,[^ \n] : Any comma that does not have space or newline following
+        # ,[^ \n\"] : Any comma that does not have space or newline or close quote following
```
claywar: @LoxleyXI Can you please explain the reason for this change? Also, please see the below note

claywar (changes requested):

tools/ci/lua_stylecheck.py
claywar: Update: I see where your issue was, and corrected this in #4128

tools/ci/lua_stylecheck.py
loxleygames: It allows the comma inside of a string `","`, which was necessary for `string.gsub` to remove them. Such a string could also be used to split. If prefer me to remove it, I can, but the PR will not pass CI checks without it.

claywar: If you rebase to latest, it will pass without needing to change the checks

loxleygames: Thanks. I've just seen the change. Understandably, this didn't exist when I added this.

claywar: Looks like the comment line still needs to be changed back to reflect its original info

loxleygames: Sorry, good catch. Updated.

claywar (approved):

PR #4100, opened 2023-07-13
Move spell_data.lua to enum/magic.lua and remove requires

claywar:
**_I affirm:_**

* Moves scripts/globals/spell_data.lua to scripts/enum/magic.lua to conform with naming
* Removes all requires for scripts/globals/spell_data in lua scripts
* Adds CI rule for deprecated require

No impact should be observed


scripts/enum/magic.lua
```diff
@@ -64,22 +64,6 @@ xi.magic.element =
...
-    [xi.day.FIRESDAY]     = xi.magic.element.FIRE,
-    [xi.day.ICEDAY]       = xi.magic.element.ICE,
-    [xi.day.WINDSDAY]     = xi.magic.element.WIND,
-    [xi.day.EARTHSDAY]    = xi.magic.element.EARTH,
-    [xi.day.LIGHTNINGDAY] = xi.magic.element.THUNDER,
-    [xi.day.WATERSDAY]    = xi.magic.element.WATER,
-    [xi.day.LIGHTSDAY]    = xi.magic.element.LIGHT,
-    [xi.day.DARKSDAY]     = xi.magic.element.DARK,
-}
-
```
Xaver-DaRed, line 82: Why was this moved to magic.lua? I'm guessing it's only used there

Xaver-DaRed (approved):

scripts/enum/magic.lua
claywar: Because every table in it is attached to xi.magic

claywar: Oh, and that one depends on other enums

PR #4101, opened 2023-07-13
[fix] Poroggo Excavator should not aggro

loxleygames:
**_I affirm:_**

Poroggo Excavator in Toraimarai Canal should not aggro. The original info was likely copied from the other Poroggo (Which do aggro). According to the wiki, these particular ones don't aggro and we had a player test retail to make sure.

https://ffxiclopedia.fandom.com/wiki/Poroggo_Excavator

```
!gotoid 17469709
```


Xaver-DaRed (approved):
Confirmed this myself.

Tested sight, sound, low HP, magic casting and resting in front of them. No agro on any.

PR #4105, opened 2023-07-14
[core] Allow DEs to define unique internal lookups

loxleygames:
**_I affirm:_**

__Issue__
Currently, dynamic entities are restricted to using packet name as the internal lookup name. In the case of duplicate names, they use the same table, and cannot have unique callbacks, etc. `onTrigger`, `onTrade` etc.

__Example__
This limitation creates a barrier on using DEs for certain tasks, for example multiple event Moogles in one zone may all be called `Moogle` but need to perform different functionality. I've repeatedly hit this issue handling multiple mobs with the same name, and had to use convoluted workarounds like giving them slightly different names or checking positions, which was not ideal.

__Solution__
I've added an additional parameter `lookupName`, which when defined, will be used internally instead. The packet name (What is visible to players) remains the same. If `lookupName` is not defined, the dynamic entity will simply use `name`, which is the existing behavior.

```lua
zone:insertDynamicEntity({
    objtype    = xi.objType.NPC,
    name       = "Moogle",
    packetName = "Egg_Hunt_Moogle",
    look       = 82,
    x          = 1,
    y          = 1,
    z          = 1,
    onTrigger  = function(player, npc)
        player:PrintToPlayer("Moogle : I'm just a Moogle.", xi.msg.channel.NS_SAY)
    end,
})

zone:insertDynamicEntity({
    objtype    = xi.objType.NPC,
    name       = "Moogle",
    packetName = "Other_Moogle",
    look       = 82,
    x          = 3,
    y          = 3,
    z          = 3,
    onTrigger  = function(player, npc)
        player:PrintToPlayer("Moogle : I'm different.", xi.msg.channel.NS_SAY)
    end,
})
```

### DIFF
--- a/modules/custom/lua/test_npcs_in_gm_home.lua
+++ b/modules/custom/lua/test_npcs_in_gm_home.lua
@@ -24,6 +24,10 @@ m:addOverride("xi.zones.GM_Home.Zone.onInitialize", function(zone)
         -- NOTE: This name CAN include spaces and underscores.
         name = "Horro",
 
+        -- Optional: Define a different name that is visible to players.
+        -- "Horro" (DE_Horro) will still be used internally for lookups.
+        -- packetName = "New Horro",
+
         -- You can use regular model ids (See documentation/model_ids.txt, or play around with !costume)
         look = 2430,
 
--- a/scripts/commands/fafnir.lua
+++ b/scripts/commands/fafnir.lua
@@ -24,6 +24,10 @@ function onTrigger(player)
         -- NOTE: This name CAN include spaces and underscores.
         name = "Fafnir",
 
+        -- Optional: Define a different name that is visible to players.
+        -- "Fafnir" (DE_Fafnir) will still be used internally for lookups.
+        -- packetName = "Fake Fafnir",
+
         -- Set the position using in-game x, y and z
         x = player:getXPos(),
         y = player:getYPos(),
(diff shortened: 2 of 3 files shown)


zach2good (approved):

zach2good:
Another way of doing this, which I think I prefer, is to expose both `name` and `packetName`. That doesn't introduce a third new name idea to anyone. Then the examples would be:
```lua
zone:insertDynamicEntity({
    objtype    = xi.objType.NPC,
    name       = "Other_Moogle",
    packetName = "Moogle",
    look       = 82,
    x          = 3,
    y          = 3,
    z          = 3,
    onTrigger  = function(player, npc)
        player:PrintToPlayer("Moogle : I'm different.", xi.msg.channel.NS_SAY)
    end,
})
```

zach2good:
Luckily, I don't think "DE_" prefix is used for anything internally, it's just to differentiate entries

zach2good (changes requested):
See previous comments

zach2good:
And when adding new fields, update the example modules to reflect and document your changes: `!fafnir` and `test_npcs_in_gm_home`

TeoTwawki:
> Another way of doing this, which I think I prefer, is to expose both `name` and `packetName`. That doesn't introduce a third new name idea to anyone. Then the examples would be:

Best option in my opinion. it should also be possible to after the fact rename them, starting out as "Eastermoogle" then being renamed before packet is sent. But simpler to just expose both fields from the get go.

zach2good (approved):

PR #4113, opened 2023-07-16
[chocobo] Add raising events debug scenario & improve playout

zach2good:
**_I affirm:_**

Moves some things around in the chocobo raising state machine thing, to try and fix event condensing and playout. This is the thing that's been blocking me from releasing a test version of chocobo raising.

- Enable it the feature and debugging by changing the settings at the top of `chocobo_raising.lua`
- `!chocoboraising`
- There is an option to set up the `Debug #1` scenario
- Talk to one of the chocobo raising NPCs
- See the output I'm trying to fix:
image
I eventually want it to be:
```
1-4: 0
4-4: 33
5-10: 0
```

If you want to take a crack at fixing it, all the logic is inside `condenseEvents()`

### DIFF
--- a/scripts/globals/chocobo_raising.lua
+++ b/scripts/globals/chocobo_raising.lua
@@ -443,7 +443,7 @@ local cutscenes =
     -- 40: Player gives the chocobo x
     -- 48: Happy to see you
     -- 51: Hangs its head in shame
-    -- 53: Haven't seen you around, chocobo is sleeping (dispose of white handkerchief)
+    HAVENT_SEEN_YOU = 53, -- Haven't seen you around, chocobo is sleeping (dispose of white handkerchief)
     -- 54: Accept white handkerchief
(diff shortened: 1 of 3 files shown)


scripts/globals/chocobo_raising.lua
```diff
@@ -572,6 +572,48 @@ local compareTables = function(t1, t2)
...
+            currentEndDay = currentEndDay + 1
+        else
+            -- If there is an active span, cut it now
+            if currentEventCSTable then
+                cutEvent(condensedEvents, currentStartDay, currentEndDay, currentEventCSTable)
+            end
+
+            -- Start a new span
+            currentEventCSTable = entry[2]
+            currentStartDay = chocoState.report.day_start + idx - 1
```
claywar: Just so it doesn't get lost, changing currentStartDay and currentEndDay assignments to `entry[1]` will provide the expected output.

claywar (approved):

PR #3503, opened 2023-01-18
[Conquest] Fix zone checked , so city zones are also included

Xaver-DaRed:
Co-authored-by: zach2good <zach2good@users.noreply.github.com>

**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

Fixes an issue with conquest zone checks. City zones were not being included. Thiswould affect conquest-dependant npcs, like the traveling merchant troupe not spawning.

## Steps to test these changes

Add a print in city zones dependant on conquest and check if cities are called now.

PR #4116, opened 2023-07-16
[fix] Lower conquest region check range

zach2good:
**_I affirm:_**

Fixin' crashes. The cities and dynamis zones don't need to be iterated here.

### DIFF
--- a/src/map/conquest_data.cpp
+++ b/src/map/conquest_data.cpp
@@ -63,6 +63,12 @@ void ConquestData::load(std::unique_ptr<SqlConnection>& sql)
 
 int32 ConquestData::getInfluence(REGION_TYPE region, NATION_TYPE nation) const
 {
+    if (static_cast<std::size_t>(region) > influences.size() - 1U)
+    {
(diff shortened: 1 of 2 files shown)


WinterSolstice8 (approved):

claywar (approved):

TeoTwawki:
> Fixin' crashes. The cities and dynamis zones don't need to be iterated here.

Actually they did when that was changed last time, and it shouldn't have been built in such a way to crash from this instead of failing out gracefully.

https://github.com/LandSandBoat/server/pull/3503

TeoTwawki:
reg range technically isn't wrong, the latents are intended to work in those regions
```cpp
                    luautils::OnConquestUpdate(PZone, Conquest_Update, influence, owner, ranking, isConquestAlliance);
                    PZone->ForEachChar([](CCharEntity* PChar)
                    {
                        PChar->PLatentEffectContainer->CheckLatentsZone();
                    });
```
we'll need to solve the latents differently if we're changing the range back.

PR #4132, opened 2023-07-19
[core] Move SmallPacket0x04B server message logic into Lua

### LINKED ISSUE #3965: 🔨 move SmallPacket0x04B handling to lua
**_I affirm:_**

## Describe the feature
As more and more gets stuffed here to handle messaging that happens at player login time  
https://github.com/LandSandBoat/server/pull/3933/files#diff-5c1417f13de5d4c90301639e386e3bec2fd53753b62c4aab5563f9c2d46c28d4R2294  
it is getting more and more apparent that there is a need to store the conditional logic and text outside of this part of the core in a more convenient manner. The issue lua side is timing those messages to reach the screen (the packets can be sent before the player has fully connected). People have worked around this with use of `timer()` before but that is not ideal, as people have varying latency and what is slow enough for one person isn't for the next, or is even too slow to look like retail at all.

zach2good:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/3965

Also removed deprecated/unused SetDropRate & UpdateServerMessage

image


zach2good:
Out of draft, ready for review

scripts/globals/server.lua
```diff
@@ -16,4 +16,23 @@ end
 xi.server.onTimeServerTick = function()
 end
 
+-- Message for use with SmallPacket0x04B
+xi.server.getServerMessage = function(language)
+    local serverMessage = ""
+
+    if language == 0x02 then -- English
```
claywar: Just a thought for the future, but should we eventually create an enum for language?

claywar (approved):

scripts/globals/server.lua
zach2good: Good idea, added now

zach2good:
Interesting added aside: I think now the server messages are hot-reloadable because they're just reading from settings. Neat.

PR #4139, opened 2023-07-21
Fixes print message for !pos command

loxleygames:
**_I affirm:_**

Fixes print functionality broken by #4097

Print and return early if arg is nil

`!pos` to see your current position


zach2good (commented):
You can do all of this and bail out early if you check for nil and then return instead of checking for not nil and then doing that logic

zach2good (approved):

PR #4142, opened 2023-07-22
Fix Trial-Size Trial by Earth ID

claywar:
**_I affirm:_**

Realized that I forgot to push the last changes from my working branch.  Fixes the ID used in quest constructor

Quest should log appropriately


Xaver-DaRed (approved):
oops

PR #4144, opened 2023-07-22
Fix logging in SmallPacket 0x077

claywar:
**_I affirm:_**

String values in debug prints were receiving int data.  Uses memberName when defined (case 0), and casts to char* in case 5

Logs should not show an error


Xaver-DaRed (approved):

zach2good:
This would be a good time to use fmt::format and change to the curly brace syntax

zach2good (approved):

PR #4145, opened 2023-07-22
Interaction - Quest: A Test of True Love

claywar:
**_I affirm:_**

* Converts "A Test of True Love" to Interaction Framework
* Properly adds gil reward that was missing in previous implementation

Follow guides for this quest


Xaver-DaRed (approved):

scripts/quests/bastok/A_Test_of_True_Love.lua
```diff
@@ -0,0 +1,109 @@
...
+            onEventFinish =
+            {
+                [272] = function(player, csid, option, npc)
+                    quest:setVar(player, 'Prog', 1)
+                    quest:setMustZone(player)
+                end,
+
+                [274] = function(player, csid, option, npc)
+                    if quest:complete(player) then
+                        xi.quest.setMustZone(player, xi.quest.log_id.BASTOK, xi.quest.id.bastok.LOVERS_IN_THE_DUSK)
```
hookstar, line 101: are the KI's permanent?

claywar: Good question, and this is something I'm wondering as well.  Given the next KI references all 3, and I'm working off of the previous implementation, its most likely they should be removed here

PR #4153, opened 2023-07-23
[Magic] Add setting for immunobreak & Resist rank multiplier re-scaling

Xaver-DaRed:
**_I affirm:_**

- Adds setting to enable or dissable Immunobreaks.
- Re-scales resistance rank multiplier on magic evasion, according to Jimmayus' documentation.

- Change setting to false to dissable immunobreaks.
- The way the re-scaling goes, mobs will be less weak at negative ranks, but also weaker at low (possitive) ranks and more resistant at higher ones.

### DIFF
--- a/scripts/globals/spells/enfeebling_spell.lua
+++ b/scripts/globals/spells/enfeebling_spell.lua
@@ -414,9 +414,9 @@ xi.spells.enfeebling.useEnfeeblingSpell = function(caster, target, spell)
         end
 
         if
+            xi.settings.main.ENABLE_IMMUNOBREAK and
             caster:isPC() and
             target:isMob() and
-            xi.settings.main.ENABLE_IMMUNOBREAK and
             immunobreakTable[spellEffect] and          -- Only certain effects can be immunobroken.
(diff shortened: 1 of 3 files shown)


zach2good (approved):

scripts/globals/spells/enfeebling_spell.lua
```diff
@@ -416,6 +416,7 @@ xi.spells.enfeebling.useEnfeeblingSpell = function(caster, target, spell)
         if
             caster:isPC() and
             target:isMob() and
+            xi.settings.main.ENABLE_IMMUNOBREAK and
```
zach2good: I'd probably move this setting to be the first thing thats checked in this block

zach2good (approved):

scripts/globals/spells/enfeebling_spell.lua
Xaver-DaRed: Can do

Xaver-DaRed: Done

claywar (approved):

PR #4163, opened 2023-07-25
Improves !additem command with item name handling

loxleygames:
**_I affirm:_**

Improves `!additem` command (and `GetItemIDByName`) by adding support for `sortname` match. This often means shorter and more familiar names that don't require memorising the prefix. The item ID and augment functions of `!additem` remain unchanged.

Before:
`!additem chunk_of_rock_salt`

After:
`!additem rock_salt`
`!additem chunk_of_rock_salt`

`!additem (name)`
`!additem (sortname)`

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -5499,8 +5499,8 @@ namespace luautils
     uint16 GetItemIDByName(std::string const& name)
     {
         uint16      id    = 0;
-        const char* Query = "SELECT itemid FROM item_basic WHERE name LIKE '%s';";
-        int32       ret   = sql->Query(Query, name);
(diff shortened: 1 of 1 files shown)


TeoTwawki:
`plate_of_sole_sushi` is the bane of my existence but bear in mind certain naming conventions might make the odd item or 2 impossible to match (and nothing can be done about it - I'm just saying we should expect a few unfixables here and there in our future, the naming conventions are from necessity because we had to accommodate OS file naming bs and other limitations at the same time).

TeoTwawki (approved):

zach2good (approved):

zach2good:
Anecdotally: I think I saw that Eden uses a fuzzy matching library to do this sort of thing, but I don't want to bring in another 3rd party lib just for the sake of people wanting to go `!additem sushi`

PR #4184, opened 2023-07-28
[lua] Modify print function for full integer values

claywar:
**_I affirm:_**

Adds an exception for lua_print override to print full decimal value if it is not a float

!exec print(1.2)
!exec print(utils.MAX_UINT32)


zach2good (approved):
Sneaky, I like it

PR #4189, opened 2023-07-29
[login] Add more robust login error handling

WinterSolstice8:
**_I affirm:_**

Sometimes, a handler can throw an exception that will stop the io_context from continuing work. This commit will allow the handlers to work after an exception of that type. Usually these are socket errors such as Windows giving us an invalid handle or some other OS level issues that we can't deal with.

However, even truly fatal exceptions are ignored and we will want a list in the future of "acceptable" exceptions to ignore for socket exception reasons.

Login as usual. No known way to directly trigger this code.

### DIFF
--- a/src/login/connect_server.cpp
+++ b/src/login/connect_server.cpp
@@ -86,12 +86,30 @@ ConnectServer::ConnectServer(int argc, char** argv)
         // if io_context finishes!
         ShowInfo("starting io_context");
 
-        io_context.run();
+        // This busy loop looks nasty, however --
+        // https://think-async.com/Asio/asio-1.24.0/doc/asio/reference/io_service.html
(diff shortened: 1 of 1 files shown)


zach2good (approved):

zach2good:
My question here is could this result in the program staying alive when it should die? If there's a fundamental problem that should stop the io_context starting will it spin forever instead of dieing?

WinterSolstice8:
> My question here is could this result in the program staying alive when it should die? If there's a fundamental problem that should stop the io_context starting will it spin forever instead of dieing?

That is a distinct possibility. There really isn't a good way to do this without knowing which exceptions are "allowable"

zach2good:
ASIO docs say this is how things should be laid out. Who are were mere mortals to argue 🤷🏻‍♂️

Xaver-DaRed (approved):

PR #4196, opened 2023-07-30
[Quest] Curses, Foiled A-Golem!? Additional Title

PublicXI:
**_I affirm:_**

Add the title "Total Loser" to the quest completion for Curses, Foiled A-Golem!? quest. The title should be added to available titles at a title-changing NPC, but the title "Doctor Shantotto's Flavor of the Month" should be the one that is set upon quest completion.

!pos 122 -2 112 239 (Shantotto)
!addquest 2 63
!exec player:startEvent(342)
Check title is "Doctor Shantotto's Flavor of the Month"
!pos 0.080 -10.765 5.394 239 (Burute-Sorute)
Purchase "Total Loser" title for 400 gil
Check title is "Total Loser"

### DIFF
--- a/scripts/quests/windurst/Curses_Foiled_A_Golem.lua
+++ b/scripts/quests/windurst/Curses_Foiled_A_Golem.lua
@@ -1,7 +1,7 @@
 -----------------------------------
 -- Curses, Foiled A-Golem!?
 -----------------------------------
--- !addquest 2 34
+-- !addquest 2 63
 -- Shantotto       : !pos 122 -2 112 239
 -- Torino-Samarino : !pos 105 -20 140 111
 -- Cermet Door     : !pos -183 0 190 204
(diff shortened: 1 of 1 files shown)


scripts/quests/windurst/Curses_Foiled_A_Golem.lua
```diff
@@ -185,6 +185,9 @@ quest.sections =
             onEventFinish =
             {
                 [342] = function(player, csid, option, npc)
+                    -- Per FFXIclopedia: You can still get [Total Loser] at the
+                    -- title-changing NPC after the quest has been completed.
+                    player:addTitle(xi.title.TOTAL_LOSER)
```
zach2good, line 190: If the quest awards 2 titles on completion, should we modify the framework so that multiple titles can be given in the rewards block?

zach2good (approved):

scripts/quests/windurst/Curses_Foiled_A_Golem.lua
claywar: I can't think of many other occasions where this would be needed; however, after looking at the addTitle binding, it looks like the function was changed from its original purpose and adds _and sets_ the player's title, which means we're going to be sending two update packets back to back here unless a change is made

claywar (approved):

PR #4199, opened 2023-07-30
Fix invalid regions being passed to HandleWeeklyTallyEnd

claywar:
**_I affirm:_**

Reduces range of regions iterated over in HandleWeeklyTallyEnd to match valid regions

Weekly Tally should not crash

### DIFF
--- a/src/map/conquest_system.cpp
+++ b/src/map/conquest_system.cpp
@@ -454,7 +454,7 @@ namespace conquest
         zoneutils::ForEachZone([ranking, isConquestAlliance](CZone* PZone)
         {
             // only find chars for zones that have had conquest updated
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):
Why is this so familiar?

zach2good (approved):

TeoTwawki:
> Why is this so familiar?

because we fixed previously latent effect issues by including zones that technically don't have conquest. this is actually the second time thats been un-fixed to fix crashes. The latent handling needs some rework. (conquest latents in fact work in sea and dyna, and even towns, which have no rank points or control stuffs - basically because they are "signet zones" and toau areas latent doesn't happen because those are "sanction zones")

PR #4200, opened 2023-07-30
Fixes NPC Tucker for A Little Knowledge quest (Rolanberry to Vellum)

Aeltorian:
**_I affirm:_**

This fixes the NPC Tucker to initiate a trade for Rolanberries to Vellum.

This PR fixes trading once again with NPC Tucker. 

 What was done; 

This is a fix for the quest A_Little_Knowledge which has an optional side trade quest with NPC Tucker.

trade code npc_utils was modified to where Tucker will only accept multiples of 12 up to 4 stacks as he states. Anymore and he'll only take the 4 stacks and return the rest. You still only have 3 attempts to trade him before your variable reaches the maximum and he'll be "out" of vellum. 

tucker's npc lua was also modified to add his 2 dialogs that he didn't have previously before. So if you talk to him outside of having the quest active, he'll talk about Rolanberries. If you try to trade him, he'll give a "what am I to do with these?" dialog. If quest is active and you trade him less than a stack, he will also give the "What am I to do with these?" dialog.

Trade NPC Tucker in Crawler's Nest (S) 4 stacks of Rolanberries, 3 times. You'll get a full stack of Vellum.

### DIFF
--- a/scripts/quests/crystalWar/A_Little_Knowledge.lua
+++ b/scripts/quests/crystalWar/A_Little_Knowledge.lua
@@ -79,6 +79,7 @@ quest.sections =
                         trade:hasItemQty(xi.items.ROLANBERRY, trade:getItemCount())
                     then
                         local numAwardedVellum = math.min(4, math.floor(trade:getItemCount() / 12))
+
                         trade:confirmItem(xi.items.ROLANBERRY, numAwardedVellum * 12)
                         quest:setLocalVar(player, 'numAwarded', numAwardedVellum)
 
--- a/scripts/quests/crystalWar/A_Little_Knowledge.lua
+++ b/scripts/quests/crystalWar/A_Little_Knowledge.lua
@@ -89,6 +89,8 @@ quest.sections =
                         elseif questOption == 3 then
                             return quest:progressEvent(11)
                         end
+                    else
+                        return quest:progressEvent(14)
                     end
                 end,
 
--- a/scripts/quests/crystalWar/A_Little_Knowledge.lua
+++ b/scripts/quests/crystalWar/A_Little_Knowledge.lua
@@ -73,9 +73,10 @@ quest.sections =
                     -- In addition, need to determine if partial stacks are consumed and implement that if needed.
 
                     if
-                        npcUtil.tradeHas(trade, { { xi.items.ROLANBERRY, 12 } }) and
                         questOption >= 1 and
-                        questOption <= 3
+                        questOption <= 3 and
+                        trade:getItemQty(xi.items.ROLANBERRY) >= 12 and
+                        trade:hasItemQty(xi.items.ROLANBERRY, trade:getItemCount())
                     then
                         local numAwardedVellum = math.min(4, math.floor(trade:getItemCount() / 12))
                         trade:confirmItem(xi.items.ROLANBERRY, numAwardedVellum * 12)
--- a/scripts/quests/crystalWar/A_Little_Knowledge.lua
+++ b/scripts/quests/crystalWar/A_Little_Knowledge.lua
@@ -73,14 +73,11 @@ quest.sections =
                     -- In addition, need to determine if partial stacks are consumed and implement that if needed.
 
                     if
+                        npcUtil.tradeHas(trade, { { xi.items.ROLANBERRY, 12 } }) and
                         questOption >= 1 and
-                        questOption <= 3 and
-                        trade:hasItem(xi.items.ROLANBERRY) and
-                        trade:getItemCount() >= 12 and
-                        trade:hasItemQty(xi.items.ROLANBERRY, trade:getItemCount())
+                        questOption <= 3
                     then
                         local numAwardedVellum = math.min(4, math.floor(trade:getItemCount() / 12))
-
                         trade:confirmItem(xi.items.ROLANBERRY, numAwardedVellum * 12)
                         quest:setLocalVar(player, 'numAwarded', numAwardedVellum)
 
--- a/scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
+++ b/scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
@@ -5,7 +5,6 @@
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
-    player:startEvent(14) -- doesn't want your trade dialog
 end
 
 entity.onTrigger = function(player, npc)
--- a/scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
+++ b/scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
@@ -8,7 +8,6 @@ entity.onTrade = function(player, npc, trade)
 end
 
 entity.onTrigger = function(player, npc)
-    player:startEvent(15) -- rolanberry history dialog
 end
 
 entity.onEventUpdate = function(player, csid, option, npc)
--- a/scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
+++ b/scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
@@ -5,9 +5,11 @@
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
+    player:startEvent(14) -- doesn't want your trade dialog
 end
 
 entity.onTrigger = function(player, npc)
+    player:startEvent(15) -- rolanberry history dialog
 end
 
 entity.onEventUpdate = function(player, csid, option, npc)


scripts/quests/crystalWar/A_Little_Knowledge.lua
```diff
@@ -73,14 +73,11 @@ quest.sections =
                     -- In addition, need to determine if partial stacks are consumed and implement that if needed.
 
                     if
+                        npcUtil.tradeHas(trade, { { xi.items.ROLANBERRY, 12 } }) and
```
claywar: This is incorrect.  You can trade less and therefore get less.

Aeltorian: > This is incorrect. You can trade less and therefore get less.

This directly conflicts with my personal testing as I could not trade less than 12, you should definitely get a "take your trading somewhere else" dialog. There's no reason to trade less than 12 anyways.

claywar: Disregard, tradeHas, not tradeHasExactly

claywar: Though, looking at this, it does check for >= 12, so you can't trade less.  Are you saying that you cannot trade anything that is not a full stack?

Aeltorian: > Though, looking at this, it does check for >= 12, so you can't trade less. Are you saying that you cannot trade anything that is not a full stack?

You can trade 4 stacks and get 4 vellum. Trade 2 stacks, get 2 vellum. If it's not a full stack, he'll return those rolanberries to you. However, if you get a single piece of vellum during that trade round, you gain that 1 of 3 variables for a chance at the full vellum stack limit. You can't get 2.5 sheets of vellum. Whole numbers only.

claywar: Ok, so it looks like the only thing that needs to change then is not eating the entire container, but instead confirming multiples of stacks

Aeltorian: > Ok, so it looks like the only thing that needs to change then is not eating the entire container, but instead confirming multiples of stacks

I'm not sure what you mean.

claywar:
```lua
                    if
                        questOption >= 1 and
                        questOption <= 3 and
                        trade:hasItem(xi.items.ROLANBERRY) and
                        trade:getItemCount() >= 12 and
                        trade:hasItemQty(xi.items.ROLANBERRY, trade:getItemCount())
                    then
```
From above, you have to talk to Tucker first (option >= 1), you get 3 changes to trade (<= 3, and unavailable when the value is 4)

Trade can only have Rolanberries, and must have at least 12

```lua
local tuckerEventFinish = function(player, csid, option, npc)
    local numAwarded = quest:getLocalVar(player, 'numAwarded')

    if npcUtil.giveItem(player, { { xi.items.SHEET_OF_VELLUM, numAwarded } }) then
        player:confirmTrade()
        quest:incrementVar(player, 'Option', 1)
    end
end
```

Actually, we're doing this already.  We're only confirming `12 * numAwardedVellum`
```lua
                        local numAwardedVellum = math.min(4, math.floor(trade:getItemCount() / 12))

                        trade:confirmItem(xi.items.ROLANBERRY, numAwardedVellum * 12)
                        quest:setLocalVar(player, 'numAwarded', numAwardedVellum)
```

So the confirm only eats what we reward on.  What is the issue again?

claywar: Mind you, I get the additional dialogue, and welcome it

Aeltorian: trade:hasItem does not initiate on the server I play on. Which has the exact lua LSB has.

claywar: Yep, that's the bug.  Needs a second parameter of 12, and then can remove the condition for `trade:getItemCount() >= 12 and`

claywar: We don't want to use npcUtil with a static value of 12, because we're manually confirming later on in the code

Aeltorian: > We don't want to use npcUtil with a static value of 12, because we're manually confirming later on in the code

So what should we do here? If I don't confirm the 12 here, then a Option variable would be counted if less than 12 was traded, I would think.

claywar: I've already set confirming manually instead of relying on the npcUtil function
```lua
                        local numAwardedVellum = math.min(4, math.floor(trade:getItemCount() / 12))

                        trade:confirmItem(xi.items.ROLANBERRY, numAwardedVellum * 12)
                        quest:setLocalVar(player, 'numAwarded', numAwardedVellum)
```

claywar: You can't enter that condition if less than 12 are traded (at least once the missing parameter is fixed)

Aeltorian: > I've already set confirming manually instead of relying on the npcUtil function
> 
> ```lua
>                         local numAwardedVellum = math.min(4, math.floor(trade:getItemCount() / 12))
> 
>                         trade:confirmItem(xi.items.ROLANBERRY, numAwardedVellum * 12)
>                         quest:setLocalVar(player, 'numAwarded', numAwardedVellum)
> ```

So, remove the npc_utils line and keep the 

> if
>                         questOption >= 1 and
>                         questOption <= 3
> then

?

claywar:
```lua
                    if
                        questOption >= 1 and
                        questOption <= 3 and
                        trade:hasItem(xi.items.ROLANBERRY, 12) and
                        trade:hasItemQty(xi.items.ROLANBERRY, trade:getItemCount())
                    then
```
This verifies proper option, that there's at least 12 rolanberries, and the trade container only has rolanberries

Aeltorian: Ok, let me test.

Aeltorian: @claywar 

trade:hasItem refuses to work. Throwing a handling error with nil value.

claywar: Sheesh, where was my head space that day (and today).  Change hasItem to hasItemQty

Aeltorian: Yeah, this isn't working at all. NPC is not responding to the trade. Even removed the onTrade in npc lua too and still not responding...

claywar: Give me a moment.  I'll check locally and get back.  You did talk to Tucker first, right? That's a requirement

Aeltorian: > Give me a moment. I'll check locally and get back. You did talk to Tucker first, right? That's a requirement

yeah, It's not even giving me the correct dialogue CS neither. I've already resetted my variable. Still no change.

claywar:
```lua
                    if
                        questOption >= 1 and
                        questOption <= 3 and
                        trade:getItemQty(xi.items.ROLANBERRY) >= 12 and
                        trade:hasItemQty(xi.items.ROLANBERRY, trade:getItemCount())
                    then
 ```
Tested, and this performs as expected

Comment (Aeltorian) [reply scripts/quests/crystalWar/A_Little_Knowledge.lua]: Then I'm having an issue getting that variable back to 0.

Comment (Aeltorian) [reply scripts/quests/crystalWar/A_Little_Knowledge.lua]: recommiting. Works

File: scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua [hunk 2]
```diff
@@ -5,9 +5,11 @@
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
+    player:startEvent(14) -- doesn't want your trade dialog
```
Comment (claywar) [inline scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: Need some confirmation here.  If you're not on the quest, does this always happen?  If it doesn't, then it can move to an else block in the quest (I'll help to verify)

scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
```diff
@@ -5,9 +5,11 @@
...
 end
 
 entity.onTrigger = function(player, npc)
+    player:startEvent(15) -- rolanberry history dialog
```
Comment (claywar) [inline scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: This can probably move to default actions, but there's a run-up for other dialogues.  I want to say events 4 (one-time) and 5

scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
Comment (Aeltorian) [reply scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: > Need some confirmation here. If you're not on the quest, does this always happen? If it doesn't, then it can move to an else block in the quest (I'll help to verify)

No, it does not. The Quest stuff overrides it.

Comment (claywar) [reply scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: Right, but what happens if you trade to him when you're _not_ on the quest? This is the behavior I'm asking about here

scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
Comment (Aeltorian) [reply scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: > This can probably move to default actions, but there's a run-up for other dialogues. I want to say events 4 (one-time) and 5

I was JUST looking at that.

Comment (Aeltorian) [reply scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: As we discussed that this would require a variable since one of these require a one time only event and we don't want to leave a forever variable in the system. So we'll move past this, I've also adjusted the event 14 in as a else catchall on trade during the quest as you verified on retail as it's a trade deny after the fact. Recommited. Lets get this working for those who need the vellum.

scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua
```diff
@@ -8,6 +8,7 @@ entity.onTrade = function(player, npc, trade)
...
```
Comment (claywar) [inline scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: I'm a bit wary putting this here. Is this his new action after completing the quest? If so, it would be safer to put this in a completed section of the quest and replaceDefault

Comment (Aeltorian) [reply scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: Actually,  I agree with you. I'd like him to have some form of dialog when you talk to him before triggering this quest, else he'll say nothing when you talk to him. However. Player:startEvent may be the wrong dialog command to prompt it. I don't know all Lua just yet so the call to initiate dialog eludes me atm. Could you assist?

Comment (claywar) [reply scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: I'm working on a PR right now so that we can handle the 4->5 transition.  4 will be the default action, 5 will replace after that has passed.  We're going to need to be careful with the priority tree for things, but it'll be covered soon.  For now, can we limit the scope of the PR to fixing the trade issue?

If 14/15 happen during the quest, I'm fine with keeping them there though, but not in the NPC script.

Comment (Aeltorian) [reply scripts/zones/Crawlers_Nest_[S]/npcs/Tucker.lua]: Alright, I'll remove stuffs from npc file and the quest lua should be all ready to go.

Aeltorian:
Alright, this latest commit. I can't verify if he continues to say the exact dialog;

"Fruits have an off-season, you know. What do you expect me to do with these? Eat 'em myself?"
"Go play at trading somewhere else."

After the tradings are complete, even when the quest is eventually completed, which ID 14 would have to go into the npc.lua as well.  However, at least the quest portion is complete and works.

claywar (approved):

PR #4215, opened 2023-08-03
[core] Refactor GetFirstID() to be numeric-only

zach2good:
**_I affirm:_**

Allows you to use:

```lua
        MEDUSA                = GetFirstID("Medusa") + 1,
        LIL_APKALLU           = GetFirstID("Lil_Apkallu") + 42,
```

if you so choose.

Also hits the db a lot less than before.

Works:

- First load, without errors on invalid zone names
- Hot-reload of IDs files


TeoTwawki:
side note for naughty chilrens who look at this pr and then immediately set to using it:

Do not do this
```lua
        MEDUSA = GetFirstID("Medusa") ,
        ADD1   = GetFirstID("Medusa") + 1,
        ADD2   = GetFirstID("Medusa") + 2,
        ADD3   = GetFirstID("Medusa") + 3,
```

do this instead
```lua
        MEDUSA = GetFirstID("Medusa") ,
        ADD1   = MEDUSA + 1,
        ADD2   = MEDUSA + 2,
        ADD3   = MEDUSA + 3,
```

zach2good:
Ready for review

src/map/lua/luautils.cpp
```diff
@@ -847,81 +905,64 @@ namespace luautils
...
+        }
+        // clang-format on
+    }
+
+    void PopulateIDLookupsByZone(std::optional<uint16> maybeZoneId)
+    {
+        TracyZoneScoped;
+
+        // Make sure this has been run at least once!
+        auto result = lua.safe_script_file("scripts/globals/zone.lua");
```
claywar: If zones are a straight enum, we could probably break that table out and move over there

claywar (approved):

src/map/lua/luautils.cpp
zach2good: Quite right, these aren't needed anymore - removed

PR #4220, opened 2023-08-03
Make Lua addQuest not dependent on completed state

claywar:
**_I affirm:_**

It is possible to have a completed quest set to active again, though the current binding does not allow this by checking the completed array for quests.  This removes that check, and makes it only dependent on if the quest is already flagged.

Find a quest that you have completed, and attempt to add it again.  This should not impact most things, and most repeatable quests already handle this in a less-than-desirable manner (deleting and re-adding)

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -6749,7 +6749,6 @@ void CLuaBaseEntity::setRankPoints(uint16 rankPoints)
  *  Function: addQuest()
  *  Purpose : Adds a new quest to the character's in-progress quest log
  *  Example : player:addQuest(xi.quest.log_id.BASTOK, xi.quest.id.bastok.LURE_OF_THE_WILDCAT)
- *  Notes   :
  ************************************************************************/
 
 void CLuaBaseEntity::addQuest(uint8 questLogID, uint16 questID)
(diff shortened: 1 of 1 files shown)


zach2good (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -6765,9 +6764,8 @@ void CLuaBaseEntity::addQuest(uint8 questLogID, uint16 questID)
     if (questLogID < MAX_QUESTAREA && questID < MAX_QUESTID)
     {
         uint8 current  = PChar->m_questLog[questLogID].current[questID / 8] & (1 << (questID % 8));
-        uint8 complete = PChar->m_questLog[questLogID].complete[questID / 8] & (1 << (questID % 8));
 
-        if ((current == 0) && (complete == 0))
```
TeoTwawki, line 6770: not sure when that came in but that has to have been a mistake, it was never supposed to block based on complete state and I know somewhere in the past I was able to have active+complete like on retail

TeoTwawki (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -6765,9 +6764,8 @@ void CLuaBaseEntity::addQuest(uint8 questLogID, uint16 questID)
     if (questLogID < MAX_QUESTAREA && questID < MAX_QUESTID)
     {
         uint8 current  = PChar->m_questLog[questLogID].current[questID / 8] & (1 << (questID % 8));
```
TeoTwawki: CI complains of the alignment here since the line it was aligned to below it no longer exists

PR #4221, opened 2023-08-05
[sql] Add Endark/Enlight animation and skill type/element

WinterSolstice8:
**_I affirm:_**

Corrects missing animations for Enlight/Endark
Sets skill type so you have non-zero skill for the purposes of checking interrupt
Enables skilling up divine/dark magic for the appropriate spell
image

Use enlight/endark and get skillups on divine/dark, don't get interrupted by mobs and see correct animations.


TeoTwawki (approved):
how in the F was this not added when the scrolls and spell scripts were added

zach2good (approved):

PR #4222, opened 2023-08-05
Adds NM Olgoi Khurkoi respawn timer

Aeltorian:
Adds a spawntimer on despawn to NM Olgoi Khurkoi in North Gustaberg [s]

**_I affirm:_**

Adds a respawn timer to Olgoi-K by setting a respawn timer when he despawns. Currently, set to a random math range between 90 to 120 minutes. 

References for respawn rates:

https://wikiwiki.jp/ffxi/Verup/091110/NM/%E3%82%A2%E3%83%AB%E3%82%BF%E3%83%8A%E3%82%A8%E3%83%AA%E3%82%A2#y3efabea

https://ffxiclopedia.fandom.com/wiki/Olgoi-Khorkhoi

Find Olgoi-Khurkoi out in North Gustaberg (S). Fight it and then upon death, wait anywhere between 90 to 120 minutes in order for it to respawn.


scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
```diff
@@ -6,8 +6,17 @@ require("scripts/globals/hunts")
 -----------------------------------
 local entity = {}
 
+entity.onMobSpawn = function(mob)
+    mob:setRespawnTime(0, true)
```
Comment (claywar) [inline scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: setRespawnTime only accepts one argument

Comment (Aeltorian) [reply scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: So the true doesn't need to be there. Alright.

scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
```diff
@@ -6,8 +6,17 @@ require("scripts/globals/hunts")
 -----------------------------------
 local entity = {}
 
+entity.onMobSpawn = function(mob)
+    mob:setRespawnTime(0)
```
Comment (zach2good) [inline scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: Looking around the codebase, I don't think setting the respawn time is required - I think it's implied?

Possibly depending on what the flags in the db say for the mob

Comment (Aeltorian) [reply scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: Just for the onMobSpawn since despawn is calling it?

zach2good (approved):
Note to reviewers: remember to squash!

PR #4230, opened 2023-08-08
[lua] Partially Converting "A" Zones to Dynamic Lookups

Frankie-hz:
**_I affirm:_**

Someone has to do it...

Launch the game and see all the mob/npc interactions for the lookups work.


Frankie-hz:
Not really sure what to do with Apollyon Crates. I didn't see an easy way to do the lookups so I left them as is.

Someone with more experience of how Arrapago Remnants works should probably do that one. I can convert them all using the IDs they currently have in there but I am not familiar with the zone at all.

scripts/zones/Abyssea-Altepa/IDs.lua
```diff
@@ -71,28 +71,28 @@ zones[xi.zone.ABYSSEA_ALTEPA] =
...
-        TABLILLA                    = 17670572,
-        SHARABHA                    = 17670574,
-        WAUGYL                      = 17670576,
-        CHICKCHARNEY                = 17670577,
-        VADLEANY                    = 17670578,
-        BUGUL_NOZ                   = 17670579,
+        RANI_1              = GetFirstID("Rani"),
+        ORTHUS_1            = GetFirstID("Orthrus"),
+        DRAGUA_1            = GetFirstID("Dragua"),
+        BENNU_1             = GetFirstID("Bennu"),
```
claywar, line 77: The amount of offsets here doesn't really solve, but rather hides the problem.  There probably should be a better way to address these since there are multiple instances of the same mob.  Not going to comment on every instance, but + 201 below?  That's obscene, and there must be a better way.

Frankie-hz: Maybe a GetSecondID / GetThirdID? Unless we can make something like GetTableIDs that would return the list of same IDs and pick the number you want instead of returning a full list.

Frankie-hz: There are also instances of where placeholder of mobs are just all over the place with IDs. We would need a way to target a specific mob in a list with multiple names.

Frankie-hz: I can make a GetSecondID and GetThirdID:

Prints the correct values instead of using + 201/etc.

```lua
KUKULKAN_1              = GetFirstID("Kukulkan"),
KUKULKAN_2              = GetSecondID("Kukulkan"),
KUKULKAN_3              = GetThirdID("Kukulkan"),
```
image

Another idea would be a GetID("name", # of ID you want)
```lua
KUKULKAN_1              = GetID("Kukulkan"),
KUKULKAN_2              = GetID("Kukulkan", 1),
KUKULKAN_3              = GetID("Kukulkan", 2),
```
Which also returns the same values: 
image

I think option two would be the best one out of the two I came up with. I can add that to this PR unless someone can think of a something cleaner. Code I made:

```cpp
lua.set_function("GetID", [&](std::string const& name, std::optional<int> index = std::nullopt) -> uint32
{
    auto& elements = lookup[name];
    if (!index)
    {
        return elements.front();  // Return the first element if index is not provided
    }
    else if (index.value() >= 0 && index.value() < elements.size())
    {
        return elements[index.value()];
    }
    else
    {
        ShowWarning("GetID has no value!");
        return 0;
    }
});
```

zach2good: GetTableOfIDs(name)[1], etc?

zach2good:
Now that I'm awake and I've had a coffee:

## General Notes

`Someone has to do it...`: sadly, no they don't. Not necessarily.

What the new helpers have done is draw attention to how ridiculous our ID lookup files are, and whether or not we should be doing things this way. Cargo-culting these new helpers doesn't put us in a much better place in terms of design, clarity, or ease of use. 

You've probably noticed that things get real awkward around our current mob placeholder ID system, and you already mentioned you didn't know what to do with the Apollyn crates - so you left them. Should we jam these helpers in there? Or should we think about whether its time for a new design?

How many of these IDs _really_ need to live in these files? Are they referenced in a single place and only really needed to be organised at spawn or respawn time? Meaning they can be a runtime check and we can lookup actually by name, rather than this pseudo-step at load time.

If all of these were automatically updated with tooling (I think they partially were at one point, back when there was a Java-based update tool?) then it would make sense to lay everything out in these files. But they aren't, so we should look at what alternatives are available.

## Impl Notes

Things like this:

```lua
   PANDEMONIUM_WARDEN = GetFirstID("Pandemonium_Warden") + 1,
```
it's not clear to the casual observer _why_ it's a +1 when there isn't a +0 elsewhere in the file. This would want a comment explaining. But really, does this really need to be here?

If it's used in a single place, maybe it's `onTrigger`, can we use `zone:queryEntitiesByName(` to look it up at runtime when everything is populated? That's _even more_ shift-proof.

===

From the first round of reviews earlier:

Adding more helpers isn't the answer, nor is adding offsets of 200 or so while doing lookups.
Asking for a table of all the "Rani"'s for example, and then indexing in feels better in these cases*.

```
GetTableOfIDs(name)[1],
```

It was cheap to construct these tables before, but now it's even more so:
I've just added https://github.com/LandSandBoat/server/pull/4232

*: But like I said at the start, we don't yet know if this is a pattern we want to spend the time converting _300 large files_ to.

===

Placeholder format.

If you look at how we use placeholder ID lookups:
```lua
        PEALLAIDH_PH          =
        {
            [17100870] = 17101143, -- 333.885 -9.646 -447.557
            [17100871] = 17101143, -- 309.638 -8.548 -447.557
            [17100872] = 17101143, -- 307.320 -10.088 -451.786
            [17100873] = 17101143, -- 295.122 -12.271 -414.418
            [17100874] = 17101143, -- 287.607 -16.220 -387.671
            [17100875] = 17101143, -- 315.793 -16.336 -402.407
            [17100876] = 17101143, -- 321.809 -16.843 -373.780
        },
```

There's a lot of redundancy here right? and this table is only looked up in a single place:

```lua
    xi.mob.phOnDespawn(mob, ID.mob.PEALLAIDH_PH, 5, 3600) -- 1 hour
end
```

So, it's calling this helper. Then internally its doing a lookup on NM's ID and checking the placeholder ID etc. Setting a listener, etc.

The function in `mobs.lua` is actually quite small and simple. Could we be doing this another way?

Just because we have shitty patterns doesn't mean we have to propagate them forever. I don't have an answer yet for what a new design could look like, but maybe we should start thinking about it.

## Meta Notes

This came up recently with Garrison scoping stuff, but the idea applies here too:
> [lua] Converting "A" Zones to Dynamic Lookups

> Not really sure what to do with Apollyon Crates. I didn't see an easy way to do the lookups so I left them as is.

It's important to scope your PR, both in terms of content and in title. If you say you've converted a whole group of things - actually convert them, don't leave a caveat saying you missed out some bits. There's nothing wrong with doing something partially.

The current title doesn't reflect the work thats been done, so if it went in as-is, our patch notes (that are built automatically off of PR titles) wouldn't be accurate.

> [lua] Partially Convert "A" Zones to use Dynamic Lookups

Would be fine.

===

I don't remember the last time this happened, but it happens every few months when someone gets excited:

> [lua] Converting "A" Zones to Dynamic Lookups
> [lua] Converting "B" Zones to Dynamic Lookups

Please don't submit multiple chunks of work when the first part hasn't been approved or validated.

scripts/zones/Abyssea-Altepa/IDs.lua
TeoTwawki: > GetTableOfIDs(name)[1], etc?

yes please

Frankie-hz: Going back into my hole now.

zach2good:
Linking the discussion: https://github.com/LandSandBoat/server/discussions/4235

Does anyone have ideas about how to handle this going forwards?

We have lots of different cases we want to cover.

The main issue we have is: No matter what system we come up with, we'll never be _fully_ insulated and safe from ID shifts from version updates. Lookups by name are fine, when we aren't looking up much, or if there's a single guaranteed name or group of names.

Exceptions (I think):

- `Apollyon (Zone 38)`  https://raw.githubusercontent.com/LandSandBoat/server/base/sql/npc_list.sql

There are just _so many_ armoury crates, with no realy way to differentiate them. Is SE likely to update the zone and cause a entity ID shift? No they're not. But that doesn't free us from the not wanting to do any of the following:
- Hardcode IDs
- Do a single lookup and do large offsets from it
- Do a table lookup and do large indexes into that table

@Shozokui suggested that we key on the spawn points / positions of mobs and NPCs, which is fine, until someone recaps a zone and everything breaks as badly as if there was a huge shift.

Anyone got ideas?

zach2good:
Gonna carry this discussion on... in the discussion

Frankie-hz:
Closing as per discord announcement changes are coming.

PR #4233, opened 2023-08-08
[Interaction] Fix impossible option in "Faded Promises" quest

Xaver-DaRed:
**_I affirm:_**

- Fixes incorrect option being checked.

Event 803 "always returns option 0" in retail. (This is seemingly untrue, though. Just an issue with the capture suite)
imagen

First option return in screenshot is for second (the wrong) option.
Second option return in screenshot is for the first (the correct) option.

UPDATE: This seems to be an issue with the capture suite.

- Start quest.
- Talk to ayame.
- See the quest advance when selecting the correct option.

### DIFF
--- a/scripts/quests/bastok/Faded_Promises.lua
+++ b/scripts/quests/bastok/Faded_Promises.lua
@@ -77,7 +77,8 @@ quest.sections =
             onEventFinish =
             {
                 [803] = function(player, csid, option, npc)
-                    if option == 1 then
+                    -- This event uses a very big number for cancelation, making the capture suite wrongly return 0.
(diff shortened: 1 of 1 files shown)


claywar:
> Both are 0 as you can see. There are no event updates either.

This is actually incorrect, and a bug with the capture suite.  In the case of declining, the returned value is 1073741824 (bit 31 == 1).  This can safely use the condition `if option == 0` to fix

zach2good:
> and a bug with the capture suite.

If this is the case, this PR can still be used to leave a comment saying so for future generations

claywar (approved):

PR #4240, opened 2023-08-09
Add ability to track one-time (unique) events for a player

claywar:
**_I affirm:_**

* Adds a DB field in char_unlocks to track unique events (Things that we cannot track by any other means, that can only happen once)
* Adds set/del/hasCompletedUniqueEvent bindings
* Adds uniqueEvent enum in lua land to identify the above events
* Adds example usage for Ekokoko in Mhaura

Talk to Ekokoko, see that the response is latched to event 52 after the first interaction

### DIFF
--- a/scripts/enum/unique_event.lua
+++ b/scripts/enum/unique_event.lua
@@ -1,0 +1,16 @@
+-----------------------------------
+-- Unique Events
+-----------------------------------
+xi = xi or {}
+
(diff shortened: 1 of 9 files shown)


zach2good (approved):

zach2good:
Now that this is in, should we do an audit of things that would be useful to migrate?

PR #4241, opened 2023-08-09
Remove unused parameters from Emela-ntouka setRespawnTime

claywar:
**_I affirm:_**

Binding: `void CLuaBaseEntity::setRespawnTime(uint32 seconds)`

Removes second parameter from this script which provides no impact

No changes should be observed

### DIFF
--- a/scripts/zones/Fort_Karugo-Narugo_[S]/mobs/Emela-ntouka.lua
+++ b/scripts/zones/Fort_Karugo-Narugo_[S]/mobs/Emela-ntouka.lua
@@ -27,7 +27,7 @@ entity.onMobSpawn = function(mob)
         end
     end)
 
-    mob:setRespawnTime(0, true)
+    mob:setRespawnTime(0)
 end
 
 entity.onMobDeath = function(mob, player, optParams)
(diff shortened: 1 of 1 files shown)


TeoTwawki (commented):
looks like it used to exist, since this is still in the comment block:
```
 *  Notes   : Haven't seen the second argument option being used
 *
 * Note: Removed optional (and unused) parameter 2:
 *      if (!lua_isnil(L, 2) && lua_isboolean(L, 2) && lua_toboolean(L, 2))
        { // set optional parameter to true to only modify the timer
            return 0;
        }
 ```

claywar:
Given ` Removed optional (and unused) parameter 2` Is this something we want/need to support?

TeoTwawki:
> Given ` Removed optional (and unused) parameter 2` Is this something we want/need to support?

I'm not sure why we ever did tbh

TeoTwawki:
tell you what, if we need it later, lets give it its own function not a param

TeoTwawki (approved):

zach2good:
This is also present in [Xolotl.lua](https://github.com/LandSandBoat/server/blob/4b87f4e35dc0f7dd43196eb90c61ea00599e31fa/scripts/zones/Attohwa_Chasm/mobs/Xolotl.lua#L10) if anyone wants a free PR

https://github.com/search?q=repo%3ALandSandBoat%2Fserver+mob%3AsetRespawnTime%280&type=code

PR #4244, opened 2023-08-10
Fix Scroll of Burst II being unusable

kelppo:
**_I affirm:_**

Remove the extra "xi.magic.spell"
This was broken in ea9df8c67896efbb37ce35929f01bb0d717447d0.

1. !changejob blm 75
2. !additem burst_ii
3. Try to use the scroll and observe how it doesn't work
4. Apply this patch
5. Observe how it now works

### DIFF
--- a/scripts/globals/items/scroll_of_burst_ii.lua
+++ b/scripts/globals/items/scroll_of_burst_ii.lua
@@ -6,11 +6,11 @@
 local itemObject = {}
 
 itemObject.onItemCheck = function(target)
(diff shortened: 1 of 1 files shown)


zach2good (approved):

zach2good:
A quick search confirms that this is the only place this has happened:
https://github.com/search?q=repo%3ALandSandBoat%2Fserver+.xi&type=code

Thanks for the PR 👍

PR #4247, opened 2023-08-10
[SQL] Add missing drop lists to Dynamis bosses

CatsEyeXI:
**_I affirm:_**

Add drop lists to missing Dynamis bosses

Execute the SQL statements


claywar:
Do you have a source for confirming that these are valid drops?

CatsEyeXI:
> Do you have a source for confirming that these are valid drops?

All of these were cross referenced with FFXIDB. 
Is there something beyond that you are looking for?

zach2good:
Be very careful when you use FFXIDB as a source. The first column they show is the AVERAGE drop rate - which combines ALL LEVELS OF TREASURE HUNTER.

You want to be using the TH0 values, unless you want to vastly inflate the droprates of these items.

CatsEyeXI:
> Be very careful when you use FFXIDB as a source. The first column they show is the AVERAGE drop rate - which combines ALL LEVELS OF TREASURE HUNTER.
> 
> You want to be using the TH0 values, unless you want to vastly inflate the droprates of these items.

just as a quick gloss-over, it looks like many of the TH0 values a values are 0, would the desired change then be to remove these items completely?

CatsEyeXI:
image

zach2good:
Without looking into it, presumably, some of those are items that have to be procc'd to drop? Which FFXIDB won't have any additional filtering for.

Otherwise, there's no data.

Then the options are to go in line with other similar items on similar mobs, and cross reference other sources - I think there are pretty good dynamis docs on JP sites?

Otherwise-otherwise, its not hard to jump on retail with a TH0 character and farm for a couple of hours to get a vague idea of rarities rather than completely guessing.

TeoTwawki:
Not only that, but you want to use the th data as a guide to figuring which rate bucket it fits in - if the th0 says its rare but both th1 and th2 say its common, its probably a low sample size skewing some results

Frankie-hz:
That screenshot you have only has 13 data points - it would be more of a guess than anything without drop rate data from retail. Drop rates are very tricky when there is limited data on some of these...unfortunate truth :(

PR #4248, opened 2023-08-12
Fix for SOB6 Wild Card quest incorrect vars

Aeltorian:
Fixes the unobtainable var 2 portion of the quest.

**_I affirm:_**

This fix fixes the unobtainable var 2 that you get from talking to the 2nd NPC. Vars 2, 3, 4, 5, have been adjusted to complete the quest which you couldn't complete prior.
 

You can now continue onwards to complete the quest once you obtain it.

### DIFF
--- a/scripts/quests/windurst/SOB6_Wild_Card.lua
+++ b/scripts/quests/windurst/SOB6_Wild_Card.lua
@@ -78,7 +78,7 @@ quest.sections =
             ['_6n2'] =
             {
                 onTrigger = function(player, npc)
-                    if quest:getVar(player, 'Prog') == 2 then -- First meeting at house of hero.
+                    if quest:getVar(player, 'Prog') == 1 then -- First meeting at house of hero.
                         if player:getRank(xi.nation.WINDURST) < 9 then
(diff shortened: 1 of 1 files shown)


TeoTwawki:
> hotfix

Please stop using that word - both in PR titles and commit messages. 

It is not appropriate for the context. If I were quick fixing my personal server, I'd tell my players I made a hotfix. Fixing an issue in the upstream source code, I would not call a hotfix. **Microsoft** has "hotfixes" that are short quick fixes for an issue that are not the final form of whatever correct measures eventually come and often break other things as a trade off. We don't do that here.

Aeltorian:
> > hotfix
> 
> Please stop using that word - both in PR titles and commit messages.
> 
> It is not appropriate for the context. If I were quick fixing my personal server, I'd tell my players I made a hotfix. Fixing an issue in the upstream source code, I would not call a hotfix. **Microsoft** has "hotfixes" that are short quick fixes for an issue that are not the final form of whatever correct measures eventually come and often break other things as a trade off. We don't do that here.

o7

Xaver-DaRed (approved):

PR #4250, opened 2023-08-12
Convert zone types field to a bitmask

TeoTwawki:
**_I affirm:_**

The intention here is to pave the way towards asking the zone itself if it is a "signet zone" and the like. I thought this made more sense here than in the zone misc flags, which are more about what you can do while in a zone. here we want to know if the zone "counts" as these types, so a signet type made sense to me. Note that this is not the same as being a conquest zone. Several things related to signet will function in zones where conquest doesn't happen.
- Changed from enum class to just enum to ease bitwise maths.
- Added zonetypes for signet/sanction/sigil/ionis in the mask.
- Updated all core references to use bitwise checks. values can now be combined.
- Did update existing enum values in zone settings, but did not add new enums to the values in zone_settings yet

Log in and verify using prints that dungeons still count as dungeons, dynamis still counts as dynamis. There should be no observable changes before and after merging this PR.


zach2good:
Suggestion:

Change GetType() to GetTypeMask(), really try to hammer home that its a mask!

TeoTwawki:
> Suggestion:
> 
> Change GetType() to GetTypeMask(), really try to hammer home that its a mask!

done, good idea

src/map/zone.h
```diff
@@ -415,15 +415,18 @@ enum class CONTINENT_TYPE : uint8
     OTHER_AREAS            = 4
 };
 
-enum class ZONE_TYPE : uint8
+enum ZONETYPE
```
zach2good: This can still be `enum ZONETYPE : uint16`

src/map/zone.h
```diff
@@ -530,7 +533,7 @@ class CZone
 {
 public:
     ZONEID             GetID();
-    ZONE_TYPE          GetType();
+    uint16             GetTypeMask();
```
zach2good: This can still be `ZONETYPE`, it'll be convertible back into a uint16

src/map/lua/lua_zone.cpp
```diff
@@ -453,7 +453,7 @@ void CLuaZone::Register()
     SOL_REGISTER("getID", CLuaZone::getID);
     SOL_REGISTER("getName", CLuaZone::getName);
     SOL_REGISTER("getRegionID", CLuaZone::getRegionID);
-    SOL_REGISTER("getType", CLuaZone::getType);
+    SOL_REGISTER("getType", CLuaZone::getTypeMask);
```
zach2good: This will want to be updated in all the Lua too:  SOL_REGISTER("getType", CLuaZone::getTypeMask);

TeoTwawki: fidn and replace is failing me, brb with multiple corrections after this rebuild

TeoTwawki:
retested and rebased

Xaver-DaRed (approved):

src/map/utils/zoneutils.cpp
```diff
@@ -430,9 +430,9 @@ namespace zoneutils
                     {
                         while (sql->NextRow() == SQL_SUCCESS)
                         {
-                            ZONE_TYPE zoneType = PZone->GetType();
+                            ZONE_TYPE zoneType = PZone->GetTypeMask();
 
-                            if (zoneType != ZONE_TYPE::DUNGEON_INSTANCED)
+                            if (!(zoneType & ZONE_TYPE::INSTANCED))
```
claywar, line 435: In the future, I think we might want to avoid bit compares in the condition, and perhaps abstract it out to a function.  Out of scope though

claywar (approved):

src/map/utils/zoneutils.cpp
TeoTwawki: eventually when I have more time I think we want whats incoming for status effect flag functions: `GetZoneType(Enumhere)` to check single bits of the mask. Just not tonight unless someone else wants to dive in. :)

PR #4252, opened 2023-08-13
[core] Attempt to prevent "undead" actions, see commit text

WinterSolstice8:
**_I affirm:_**

This commit is attempting to prevent actions from occuring after a mob/player has died

If a counter kills you, you should not be able to continue your attack beyond that attack round If you are dead, you should not be able to auto attack, or finish a skill or ranged attack of any kind

TODO: verify if any interrupt packets on some of the harder cases to occur (such as weaponskill state) upon death

Kill mob/player in various states of readying TP moves, auto attacks, etc, don't see any "beyond the grave" attacks

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -1744,8 +1744,10 @@ bool CBattleEntity::OnAttack(CAttackState& state, action_t& action)
 
     /////////////////////////////////////////////////////////////////////////
     //  Start of the attack loop.
+    //  Make sure our target is alive on each iteration to not overkill;
+    //  And make sure we aren't dead in case we died to a counter.
     /////////////////////////////////////////////////////////////////////////
-    while (attackRound.GetAttackSwingCount() && !(PTarget->isDead()))
+    while (attackRound.GetAttackSwingCount() && PTarget->isAlive() && this->isAlive())
     {
         actionTarget_t& actionTarget = list.getNewActionTarget();
         // Reference to the current swing.
--- a/src/map/ai/states/attack_state.cpp
+++ b/src/map/ai/states/attack_state.cpp
@@ -170,5 +170,5 @@ bool CAttackState::CanAttack(CBattleEntity* PTarget)
 
 bool CAttackState::AttackReady()
 {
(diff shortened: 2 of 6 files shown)


TeoTwawki:
for server ops and layers noticing this: those times you see your dead player or mob still continue swinging is because latency and packet arrival order. He dead, he jus don't know it yet 😉

Xaver-DaRed (approved):
Extra-checking aint a bad thing. I do wonder if this may or may not conflict with mobs reraise (SEA mobs)

src/map/entities/battleentity.cpp
```diff
@@ -1744,8 +1744,10 @@ bool CBattleEntity::OnAttack(CAttackState& state, action_t& action)
 
     /////////////////////////////////////////////////////////////////////////
     //  Start of the attack loop.
+    //  Make sure our target is alive on each iteration to not overkill;
+    //  And make sure we aren't dead in case we died to a counter.
     /////////////////////////////////////////////////////////////////////////
-    while (attackRound.GetAttackSwingCount() && !(PTarget->isDead()))
+    while (attackRound.GetAttackSwingCount() && !PTarget->isDead() && this->isAlive())
```
TeoTwawki: kinda weird for the previous author to use "not isDead" when thats exactly the same as checking "isAlive". Both return the same true/false-ness on of the entity is alive.

TeoTwawki (approved):

src/map/entities/battleentity.cpp
TeoTwawki: up to you if you'd like to turn both checks into 
` while (attackRound.GetAttackSwingCount() && PTarget->isAlive() && this->isAlive())`

TeoTwawki:
> Extra-checking aint a bad thing. I do wonder if this may or may not conflict with mobs reraise (SEA mobs)

shouldn't, technically its not the entity performing an action

src/map/entities/battleentity.cpp
WinterSolstice8: True, no real reason to use both. I'll change it.

PR #4255, opened 2023-08-13
Retail accurate AH listings

### LINKED ISSUE #3856: 🐛 AH sorting broken
**_I affirm:_**

Windows Server 2019

## Branch affected by issue 
`base`

## Steps to reproduce
Browse rings at AH, notice items aren't sorted. Choose to sort by level. notice rings STILL aren't sorted, be confused try to reset then re-sort, exit and re-enter AH, give up after a few minutes. Note that the order DOES change, just not correctly.

## Expected behavior
See highest level rings at the top, lower level rings below.

cocosolos:
**_I affirm:_**

Track which items have been listed on the AH using a new protected table
and trigger. Only show these items in the AH by default. This is controlled
by the `search.OMIT_NO_HISTORY` setting and is now on by default.

Run the following query to populate the new table using your existing
auction house history:
```
INSERT IGNORE INTO auction_house_items(itemid)
SELECT itemid FROM auction_house
GROUP BY itemid;
```

Fixes #3856

With empty auction_house and auction_house_items tables:
Make sure OMIT_NO_HISTORY is true in your search.lua settings file.
Search any category on AH, should all be empty.
List something on AH, it now appears in searches.
Remove listing, item still shows in searches.
Change OMIT_NO_HISTORY to false.
Search any category on AH, should all be fully populated.

### DIFF
--- a/settings/default/search.lua
+++ b/settings/default/search.lua
@@ -11,7 +11,7 @@ xi.settings = xi.settings or {}
 
 xi.settings.search =
 {
-    -- Omit items with no buy history from auction house results
+    -- Omit items with no listing history from auction house results
     OMIT_NO_HISTORY = false,
 
     -- After EXPIRE_DAYS, will listed auctions expire?
--- a/sql/auction_house_items.sql
+++ b/sql/auction_house_items.sql
@@ -1,0 +1,9 @@
+--
+-- Table structure for table `auction_house_items`
+--
+
+DROP TABLE IF EXISTS `auction_house_items`;
+CREATE TABLE `auction_house_items` (
+  `itemid` smallint(5) unsigned NOT NULL,
+  PRIMARY KEY `itemid` (`itemid`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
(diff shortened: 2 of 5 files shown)


settings/default/search.lua
```diff
@@ -11,8 +11,8 @@ xi.settings = xi.settings or {}
 
 xi.settings.search =
 {
-    -- Omit items with no buy history from auction house results
-    OMIT_NO_HISTORY = false,
+    -- Omit items with no listing history from auction house results
+    OMIT_NO_HISTORY = true,
```
TeoTwawki: despite being technically retail, we probably want to default it off anyway if only because ppl who didn't read/notice it will then think its bugged

cocosolos: I agree people probably expect the false behavior but I'm conflicted because I think settings should default to retail-like. Existing servers won't notice the difference if they have this set already in their copy of the settings file and it's one of only a few settings for the search server so would be hard to miss.

TeoTwawki: Normally we want all default retail. I think this and thinking show all was better than retail was the main reasoning behind defaulting off in the PR it came in on. I'm ok with whatever other devs prefer here, but thought it worth mentioning since we had it off in here already.

TeoTwawki: > one of only a few settings for the search server so would be hard to miss.

You would think, but people often don't notice when new settings land in the default files to add them to their edited files and wind up with whatever the default setting behavior is (or an error). Even catches me now and then after I pull on my testbed not realizing how long its been. One of the few drawbacks of making it so nobody is accidently committing their settings anymore :)

cocosolos: I wanted to squash that style commit so I changed the default setting back to false since I was already rebasing. If you guys decide you want it the other way just let me know and I'll change it back np.

Xaver-DaRed (approved):

PR #4257, opened 2023-08-14
[Core] Ranged attack adjustment to max distance checks

sarca571ca:
**_I affirm:_**

- https://github.com/AirSkyBoat/AirSkyBoat/pull/3376

Use a ranged weapon(preferably with a long delay), target a mob near max range start a ranged attack as they are moving out of range. Observe that when the mob goes beyond the 25 yalm limit the ranged attack will still go off but always miss.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1547,6 +1547,12 @@ namespace battleutils
         hitrate = hitrate + (acc - eva) / 2 + (PAttacker->GetMLevel() - PDefender->GetMLevel()) * 2;
 
         uint8 finalhitrate = std::clamp(hitrate, 20, 95);
+
+        // Check to see if distance is greater than 25 and force hitrate to be 0
+        if (distance(PAttacker->loc.p, PDefender->loc.p) > 25)
+        {
+            finalhitrate = 0;
+        }
         return finalhitrate;
     }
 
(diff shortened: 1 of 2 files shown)


sarca571ca:
First time doing a cherry-pick so please do correct me if I have made errors.

src/map/utils/battleutils.cpp
```diff
@@ -1547,6 +1547,12 @@ namespace battleutils
         hitrate = hitrate + (acc - eva) / 2 + (PAttacker->GetMLevel() - PDefender->GetMLevel()) * 2;
 
         uint8 finalhitrate = std::clamp(hitrate, 20, 95);
+
+        // Check to see if distance is greater than 25 and force hitrate to be 0
+        if (distance(PAttacker->loc.p, PDefender->loc.p) > 25)
+        {
+            finalhitrate = 0;
+        }
```
Xaver-DaRed, line 1555: This is an exception, which means that it could be placed at the very top of the function and avoid having to do any calculations or additional checks for an attack thats gonna miss always anyways.
```c++
        if (distance(PAttacker->loc.p, PDefender->loc.p) > 25)
        {
            return 0;
        }
```

sarca571ca: Is this preferred? I'm not well versed in C. It appears I need to do some work anyways since the checks are failing so I might just do a separate PR for this.

PR #4276, opened 2023-08-16
[luacheck] Fix error-prone negation in conditionals

claywar:
**_I affirm:_**

After upgrading luacheck to 1.1.1, there were additional issues called out by it regarding error-prone negation (`if not x == y`).  This PR addresses those issues.

No impact should be observed, outside of fixing incorrect logic in LB9 (Part 2)

### DIFF
--- a/scripts/globals/npc_util.lua
+++ b/scripts/globals/npc_util.lua
@@ -364,7 +364,7 @@ end
 function npcUtil.giveCurrency(player, currency, amount)
     local ID = zones[player:getZoneID()]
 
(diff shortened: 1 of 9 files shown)


zach2good:
Is it possible to get/force this version of luacheck in CI?

zach2good (approved):

claywar:
> Is it possible to get/force this version of luacheck in CI?

We're pulling that right now.  My local environment didn't have it though, due to luarocks handling the update as opposed to apt

Xaver-DaRed (approved):

PR #4280, opened 2023-08-17
[core] Additions to Packet Guard

zach2good:
**_I affirm:_**

- Changes the internal timing inside packet guard to use second doubles instead of milliseconds uint32
- Adds some more rate limited packets
- Track m_LastPacketType, and add PacketsArrivingInCorrectOrder

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -124,12 +124,6 @@ namespace
     uint32 TotalPacketsDelayedPerTick = 0U;
 } // namespace
 
-/************************************************************************
- *                                                                       *
- *  mapsession_getbyipp                                                  *
- *                                                                       *
- ************************************************************************/
-
 map_session_data_t* mapsession_getbyipp(uint64 ipp)
 {
     TracyZoneScoped;
@@ -145,11 +139,20 @@ map_session_data_t* mapsession_getbyipp(uint64 ipp)
     return nullptr;
 }
 
-/************************************************************************
- *                                                                       *
- *  mapsession_createsession                                             *
- *                                                                       *
- ************************************************************************/
+map_session_data_t* mapsession_getbychar(CCharEntity* PChar)
+{
+    TracyZoneScoped;
+
+    for (const auto& [_, session] : map_session_list)
+    {
+        if (PChar && session->PChar->id == PChar->id)
+        {
+            return session;
+        }
+    }
+
+    return nullptr;
+}
 
 map_session_data_t* mapsession_createsession(uint32 ip, uint16 port)
 {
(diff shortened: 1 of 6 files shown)


zach2good:
This is cleaned up and up-to-date now, needs testing to make sure it doesn't force logout people for legitimate usage

zach2good:
As a general point, what is more inconvenient for a bad actor? Having your bad packets dropped and you not knowing about it, or being force-logged out of the game immediately?

I guess the _cruel_ thing would be to queue them up for automatic logout in a random 30s-60s range, then they won't know what packet of theirs triggered it. But then, that's stepping into "this is the job of operators" territory, since we should be already logging when someone is being abusive.

WinterSolstice8:
> As a general point, what is more inconvenient for a bad actor? Having your bad packets dropped and you not knowing about it, or being force-logged out of the game immediately?
> 
> I guess the _cruel_ thing would be to queue them up for automatic logout in a random 30s-60s range, then they won't know what packet of theirs triggered it. But then, that's stepping into "this is the job of operators" territory, since we should be already logging when someone is being abusive.

Retail behavior for bad injects is to drop your connection and let your client timeout, that's harder to notice the first time

zach2good:
> Retail behavior for bad injects is to drop your connection and let your client timeout, that's harder to notice the first time

I like it, I'll leave the helper for force logout, but I'll also add drop connection

zach2good:
Added some plumbing to allow flicking a switch to just ignore all packet traffic for a given session. This leads to eventual client timeout and looks like it looks when the server goes down. Will be very annoying.

zach2good:
Ready for review

src/map/map.cpp
```diff
@@ -785,7 +792,13 @@ int32 parse(int8* buff, size_t* buffsize, sockaddr_in* from, map_session_data_t*
             {
                 ShowWarning("[PacketGuard] Caught mismatch between player substate and recieved packet: Player: %s - Packet: %03hX",
                             PChar->getName(), SmallPD_Type);
-                // TODO: Plug in optional jailutils usage
+
+                if (settings::get<bool>("map.PACKETGUARD_DROP_CONNECTION"))
+                {
+                    ShowWarning("[PacketGuard] Removing %s's session from the server", PChar->getName());
```
WinterSolstice8: super nit: this isn't actually removing their session but making their packets ignored, if we wanted to actually remove it we could drop their session entry in sql so the key no longer works (but that would also spam the server with pointless "ignoring connection from...")

That might allow them to relog immediately, though...

zach2good:
I've scaled this back and removed the connection dropping logic. This is now just some tidying of packet guard, and enforcing `PacketsArrivingInCorrectOrder`

Xaver-DaRed (approved):

PR #4282, opened 2023-08-17
[core] Lay out casts and null checks more around gardening

zach2good:
**_I affirm:_**

### DIFF
--- a/src/map/utils/gardenutils.cpp
+++ b/src/map/utils/gardenutils.cpp
@@ -95,7 +95,7 @@ namespace gardenutils
                 CItem* PItem = PContainer->GetItem(slotID);
                 if (PItem != nullptr && PItem->isType(ITEM_FURNISHING))
                 {
-                    CItemFlowerpot* PPotItem = static_cast<CItemFlowerpot*>(PItem);
+                    CItemFlowerpot* PPotItem = dynamic_cast<CItemFlowerpot*>(PItem);
                     if (PPotItem != nullptr && PPotItem->canGrow() && vanatime >= PPotItem->getStageTimestamp())
                     {
                         uint8  stageDuration        = GetStageDuration(PPotItem);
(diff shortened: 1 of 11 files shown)


src/map/utils/gardenutils.cpp
```diff
@@ -95,15 +95,18 @@ namespace gardenutils
...
                         uint32 daysSinceStageChange = (vanatime - PPotItem->getStageTimestamp()) / VANADAY_SECONDS;
                         uint8  wiltTime             = VANADAYS_TO_WILT + PChar->getMod(Mod::GARDENING_WILT_BONUS);
                         bool   wasExamined          = PPotItem->wasExamined();
-                        if ((!wasExamined && (stageDuration > wiltTime || (stageDuration + daysSinceStageChange > wiltTime))) ||
-                            daysSinceStageChange > VANADAYS_TO_GUARANTEE_WILT + wiltTime)
+                        bool   ignoredDuringStage   = !wasExamined && stageDuration > wiltTime;
+                        bool   ignoredSinceChange   = !wasExamined && stageDuration + daysSinceStageChange > wiltTime;
+                        bool   shouldForceWilt      = daysSinceStageChange > VANADAYS_TO_GUARANTEE_WILT + wiltTime;
+
+                        if (ignoredDuringStage || ignoredSinceChange || shouldForceWilt)
```
zach2good, line 109: I gave this logical split to downstream and they are currently testing it to see if one of these is redundant and possibly causing spurious wilting

zach2good:
I'm going to be pretty busy this week, if someone could quickly smoke test this against a day or two of gardening it would be much appreciated

Xaver-DaRed:
I made a fresh build of your branch.

Maybe I'm doing something wrong, however, after placing the flowerpots, when attempting to feed it a seed to grow it, it will just ignore the imput. Won't consume the seed and the flowerpot will remain empty.

This happens with all 10 placed flowerpots.

zach2good:
I was hoping Horizon would have made some similar changes to this and had some feedback, but they haven't had time/bandwidth, and I don't know gardening well enough to weigh in on anything apart from how C++ works.

Closing, pending more reports/info

Issue #4182, opened 2023-07-28
🐛 Party chat is not getting passed through map server instances


CatsEyeXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base`

## Steps to reproduce

Form a party across different map server instances (at least 3 people)
Party leader tends to sometimes see the party chat (not sure if this is a red herring)
But most times players in different instances cannot see the party chat sent from another instance.
This is a newish bug. (as of a few weeks)

## Expected behavior

Party chat should be visible across all map server instances

zach2good:
Presumably this is fixed, since people aren't screaming from the rooftops. If this is still valid, please feel free to re-open the issue.

PR #4289, opened 2023-08-19
[core] Handle party chat zmq messaging by party ID and not character ID

WinterSolstice8:
**_I affirm:_**

This changes ZMQ messaging for party chat to be handled with party ID and not character ID --
The rationale is that in a different instance that character doesn't exist so the lookup will always fail.

Load 2 instances of xi_map with different zone lists, create a party between two different instances and be able to see party chat between the two.

### DIFF
--- a/src/map/message.cpp
+++ b/src/map/message.cpp
@@ -140,26 +140,45 @@ namespace message
             }
             case MSG_CHAT_PARTY:
             {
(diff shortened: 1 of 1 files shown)


TeoTwawki (approved):

claywar (approved):

Xaver-DaRed (approved):

zach2good:
This is possibly a fix for:https://github.com/LandSandBoat/server/issues/4182

zach2good (approved):

PR #4290, opened 2023-08-19
Fix yell messages not being instantly delivered

xkoredev:
**_I affirm:_**

Fixes yell messages delivered in batches

Yell a lot


zach2good (approved):
I would be suspicious about this, since it's bypassing the lockless queues, but the other chat message types do it and the send function looks thread safe (providing access to the socket is thread safe)

zach2good:
In this case, messages are parsed and responses are all sent from the ZMQ thread, so it's all safe 👍 

If you want to submit a message to be send from _a different thread_, you should be using `queue_message`

zach2good:
Oh good, the thread-unsafe things are private functions, nothing to worry about

PR #4302, opened 2023-08-20
Move mobskills to scripts/mobskills, remove global requires

claywar:
**_I affirm:_**

Moves mobskills out of scripts/globals so that we can take advantage of the mobskills global prior to loading these files.  Removes the mobskills global require out of all mobskill scripts

Go to a mob (tested on Bibiki Bay lv72 Goblin, Bomb) and feed it tp until it uses a mobskill


Xaver-DaRed (approved):

Xaver-DaRed:
We spoke about this, but, I think we should have all "actions" inside a folder called actions.
Aproved anyway

claywar:
I can get on board with this.  Will move these items out individually first, then followup with one change to move and sort.  I first want to see what's feasible to move

PR #4314, opened 2023-08-21
Fix conquest tally issues

xkoredev:
**_I affirm:_**

- Fixes conquest data issues due to ZQM message length mismatch
- Fixes client crash after conquest tally
- Fixes certain areas incorrectly getting tally update messages

Check region map.
Run !updateconquest 0.

### DIFF
--- a/src/map/conquest_system.cpp
+++ b/src/map/conquest_system.cpp
@@ -59,12 +59,13 @@ namespace conquest
             }
             case CONQUEST_WORLD2MAP_WEEKLY_UPDATE_END:
             {
-                const std::size_t             headerLength   = 2 * sizeof(uint8);
-                uint32                        size           = ref<uint32>(data, 2);
+                const std::size_t headerLength = 2 * sizeof(uint8) + sizeof(std::size_t);
+                const std::size_t size         = ref<std::size_t>(data, 2);
+
                 std::vector<region_control_t> regionControls = std::vector<region_control_t>(size);
                 for (std::size_t i = 0; i < size; i++)
                 {
-                    const std::size_t start = headerLength + sizeof(size_t) + i * sizeof(region_control_t);
+                    const std::size_t start = headerLength + i * sizeof(region_control_t);
 
                     region_control_t regionControl{};
                     regionControl.current = ref<uint8>(data, start);
@@ -78,13 +79,13 @@ namespace conquest
             }
             case CONQUEST_WORLD2MAP_INFLUENCE_POINTS:
             {
-                const std::size_t        headerLength      = 2 * sizeof(uint8);
+                const std::size_t        headerLength      = 2 * sizeof(uint8) + sizeof(bool) + sizeof(std::size_t);
                 bool                     shouldUpdateZones = ref<bool>(data, 2);
-                uint32                   size              = ref<uint32>(data, 3);
+                const std::size_t        size              = ref<std::size_t>(data, 3);
                 std::vector<influence_t> influencePoints   = std::vector<influence_t>(size);
                 for (std::size_t i = 0; i < size; i++)
                 {
-                    const std::size_t start = headerLength + sizeof(bool) + sizeof(size_t) + i * sizeof(influence_t);
+                    const std::size_t start = headerLength + i * sizeof(influence_t);
 
                     influence_t influence{};
                     influence.sandoria_influence = ref<uint16>(data, start);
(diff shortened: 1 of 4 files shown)


xkoredev:
I have tested all citiy zones but I am getting a weird artifact only in south / north / port sandy where I see double tally messages. No clue why.

Edit: Aha! Turns out some zone ids were incorrect, which was causing odd player messages. This was not part of my change, so as a bonus, we get to fix them. Tally messages in sandy must have been broken all along.

src/map/conquest_system.cpp
```diff
@@ -416,13 +418,17 @@ namespace conquest
...
+            if (regionId <= REGION_TYPE::TAVNAZIA)
             {
                 uint8 influence = conquest::GetInfluenceGraphics(PZone->GetRegionID());
                 uint8 owner     = conquest::GetRegionOwner(PZone->GetRegionID());
-
                 luautils::OnConquestUpdate(PZone, Conquest_Tally_Start, influence, owner, ranking, isConquestAlliance);
             }
+            else if (regionId > REGION_TYPE::TAVNAZIA && regionId < REGION_TYPE::DYNAMIS)
+            {
+                luautils::OnCityConquestUpdate(PZone, Conquest_Tally_Start, ranking, isConquestAlliance);
```
xkoredev: We want to call certain city logic for conquest updates, but cities don't have owner / influence data as they are not conquest zones per se. I think this is the cleanest way to handle this. Even though its more boilerplate, it does no longer call the same method for zones that have different assumptions of owner / influence values, which could potentially cause issues.

zach2good: Why change all of these files, when you can keep everything as it is and add branching logic inside xil.conquest.OnConquestUpdate to do something different when `regionId > REGION_TYPE::TAVNAZIA && regionId < REGION_TYPE::DYNAMIS`?

xkoredev: ya.. this is why I shoulda stopped at 4 am. Will do that.

xkoredev: oh ya so what I was trying to accomplish os not having the consumer of this logic depend on some undetermined owner / influence values. This could prevent future bugs (at the cost of more boilerplate). I.e: anyone could write

onConquestUpdate(zone, updatetype, influence, owner, ranking, isConquestAlliance)
    .. . code here assuming owner and influence is valid

in any city zone.lua

xkoredev: Additionally, the core code becomes a bit harder to follow imo. we have to do something like:

```
   zoneutils::ForEachZone([ranking, isConquestAlliance](CZone* PZone)
        {
            REGION_TYPE regionId = PZone->GetRegionID();
            if (regionId < REGION_TYPE::DYNAMIS)
            {
                // Cities do not have owner or influence
                uint8 influence = 0;
                uint8 owner = 0;
                if (regionId <= REGION_TYPE::TAVNAZIA) {
                    uint8 influence = conquest::GetInfluenceGraphics(PZone->GetRegionID());
                    uint8 owner     = conquest::GetRegionOwner(PZone->GetRegionID());
                }

                luautils::OnConquestUpdate(PZone, Conquest_Tally_End, influence, owner, ranking, isConquestAlliance);

                if (regionId <= REGION_TYPE::TAVNAZIA) {
                    PZone->ForEachChar([](CCharEntity* PChar)
                    {
                        PChar->pushPacket(new CConquestPacket(PChar));
                        PChar->PLatentEffectContainer->CheckLatentsZone();
                    });
                }
            }
        });
```

scripts/zones/Northern_San_dOria/IDs.lua
```diff
@@ -31,7 +31,7 @@ zones[xi.zone.NORTHERN_SAN_DORIA] =
         GP_OBTAINED                   = 7017,  -- Obtained: <number> guild points.
         NOT_HAVE_ENOUGH_GP            = 7018,  -- You do not have enough guild points.
         RENOUNCE_CRAFTSMAN            = 7034,  -- You have successfully renounced your status as a [craftsman/artisan/adept] of the [Carpenters'/Blacksmiths'/Goldsmiths'/Weavers'/Tanners'/Boneworkers'/Alchemists'/Culinarians'] Guild.
-        CONQUEST_BASE                 = 7288,  -- Tallying conquest results...
+        CONQUEST_BASE                 = 7274,  -- Tallying conquest results...
```
Xaver-DaRed: The Ids should not be changed. That new ID stands for
`It's all the fault of those Royal Knights! Does Prince Trion think to undo all that my son has wrought? Does he?`

Same applies for other IDs in other files, with different messages.

src/map/conquest_system.cpp
claywar: I'd prefer this as opposed to creating a new binding that people need to juggle on the lua side, and that we have to call separately (calling out to lua isn't expensive, but its still cheaper to handle it in core)

xkoredev: Done in the current PR now.

scripts/zones/Northern_San_dOria/IDs.lua
xkoredev: Ah.. this made me waste so much time. I think it was an issue with my client? Otherwise I can't explain. My offsets were different. I recently switched to retail for caps so that may be it :S

xkoredev: Changed back

zach2good (approved):

PR #4321, opened 2023-08-21
Remove remaining non-global and non-mixin global requires

claywar:
**_I affirm:_**

Removes remaining global requires which have been ensured to be preloaded by caching on startup

No game changes should be observed

### DIFF
--- a/scripts/effects/gambit.lua
+++ b/scripts/effects/gambit.lua
@@ -4,14 +4,14 @@
 local effectObject = {}
 
 effectObject.onEffectGain = function(target, effect)
--- intentionall blank, handled in scripts/globals/job_utils/rune_fencer
+    -- intentionally blank, handled in scripts/globals/job_utils/rune_fencer
 end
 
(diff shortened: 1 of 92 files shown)


Xaver-DaRed (approved):

scripts/effects/gambit.lua
```diff
@@ -4,14 +4,14 @@
...
--- intentionall blank, handled in scripts/globals/job_utils/rune_fencer
+    -- intentionally blank, handled in scripts/globals/job_utils/rune_fencer
 end
 
 effectObject.onEffectTick = function(target, effect)
 end
 
 effectObject.onEffectLose = function(target, effect)
---intentionally blank, effects removed in CPP
+    --intentionally blank, effects removed in CPP
```
Xaver-DaRed: no separation between --text here?

claywar: Not my doing, but can fix

PR #4323, opened 2023-08-22
Remove unused global exceptions, add magicburst globals to xi table and enum

claywar:
**_I affirm:_**

* Adds enum for skillchainType and updates references
* Creates xi.magicburst and attaches functions in that global to it
* Removes duplicate xi.magicburst functions
* Removes non-class mission and quest variables from global CI exceptions

No changes should be observed


Xaver-DaRed (approved):
Note to myself:
Go to combat/physical_utilities and use this enums in there

claywar:
Looks like that table duplicates as well.  I'd love to just use yours in the future

PR #4327, opened 2023-08-22
Merge `feature/garrison_enhancements` into `base`

zach2good:
**_I affirm:_**


zach2good:
Currently testing:

- Enable debug logging
- Goto Dunes, `!garrison start Testo`:

image

- Doesn't work initially - will investigate closer next week

```lua
---------------------------------------------------------------------------------------------------
-- func: garrison <command> (player)
-- commands:
-- !garrison start (player) starts the garrison for the given player (or targetted one). This bypasses requirements like lockout
-- !garrison stop  (player) stops the garrison (if any) currently running in the player's zone
-- !garrison win (player) win the garrison (if any) currently running in the player's zone
---------------------------------------------------------------------------------------------------
```

Garrison GM command should also support default-self target if there is no name provided or cursor target - like all the other GM commands

zach2good:
(Testing in Dunes, the Garrison I've played the most)

Somewhat working now, but traders aren't being allocated stats correctly:
image

So when mobs spawn, they one-hit the traders

EDIT: Fixed

zach2good:
When NPCs are created, there's no associated pool for them?

EDIT: Fixed

zach2good:
TODO: Change the module to use `super(...)` so it doesn't have to be such a big copy/paste of the existing logic

DONE:
image

zach2good:
```lua
    -- If info is missing, a debug message will be logged and Garrison will not begin
    if
        xi.garrison.getAllyInfo(zoneData, nationID) == nil or
        xi.garrison.getMobInfo(zone) == nil
    then
        player:PrintToPlayer("Garrison is currently unavailable for this region.", xi.msg.channel.SYSTEM_3)
        debugLog("Garrison was cancelled due to missing data.")

        return false
    end

```

`xi.garrison.getMobInfo(zone)` isn't defined, so everything is unplayable without GM commands

LoxleyX:
I think the functionality was moved to other functions and this line can now be removed

zach2good:
image
image

zach2good:
From caps: multiple races can spawn as NPCs:
image

So, we can either ignore this, or we can use existing pools for the base different races, and combine them with groups per-nation in each zone.

zach2good:
Even with that missing function call removed, the code _only_ works when the module is applied.

EDIT:
image

Indexing into a null table, could/should be done in steps

LoxleyX:
NPC race doesn't necessarily match the nation. As seen here, Bastok owns outpost but NPCs are Mithra with Windurst attire.
cape_terrigan_bastok

Xaver-DaRed:
That's not the point, I think. The point is each race has different stats, meaning, the look they use should match a race and with it, a set of stats

claywar (approved):

PR #4330, opened 2023-08-22
[Interaction] Fix "Acting in good faith" quest not starting

Xaver-DaRed:
**_I affirm:_**

As pointed by Claywar, the KI npcUtil was missing the player argument, which made the script error and fail to give both the KI and the quest-

Have fame 4+ with windurst and lvl 10+. Speak with NPC and get the quest started, the quest recorded in your quest log and the KI in your temporal KI list.

### DIFF
--- a/scripts/quests/windurst/Acting_in_Good_Faith.lua
+++ b/scripts/quests/windurst/Acting_in_Good_Faith.lua
@@ -36,7 +36,7 @@ quest.sections =
             {
                 [10019] = function(player, csid, option, npc)
                     if option == 0 then
-                        npcUtil.giveKeyItem(xi.ki.SPIRIT_INCENSE)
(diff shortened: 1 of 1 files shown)


scripts/quests/windurst/Acting_in_Good_Faith.lua
```diff
@@ -35,10 +35,8 @@ quest.sections =
             onEventFinish =
             {
                 [10019] = function(player, csid, option, npc)
-                    if option == 0 then
-                        npcUtil.giveKeyItem(xi.ki.SPIRIT_INCENSE)
-                        quest:begin(player)
-                    end
+                    quest:begin(player)
```
claywar: This option check is still required, since the CS can be cancelled with a non-zero option
image

The real issue is with npcUtil missing the player argument for the first parameter

claywar (approved):

PR #4342, opened 2023-08-24
[dbtool] Add function to dump DB to SQL files

cocosolos:
**_I affirm:_**

Adds a function to the dbtool maintenance tasks section to dump database data in-place into the sql files.

Queries database then scans corresponding SQL file, replacing values in INSERT statements with data from the database.

Is not smart about which INSERT statement gets which data; line-end comments can get shuffled around if SQL file INSERTs are out of order or missing. **If you're doing INSERT or DELETE operations, make sure to add a blank ``INSERT INTO `table` VALUES ();`` or manually remove the INSERT statement respectively before dumping tables.**

Almost everything except dbtool and item_latents is just formatting changes from the tool. item_latents has had some manual reordering and comment unshuffling done.

There's some issues with Python correctly padding floats according to the database settings, so trailing 0s are trimmed all throughout. None of the values should have actually changed, and the padding is readded when imported into the database. I'm not sure if there's an easy way around this, but it's just a visual thing.

1. Import a fresh database.
2. Perform a query on your DB to update some set of values.
3. Run dbool, select Maintenance Tasks (t), dump either a single table or the whole db (d or a), check git diffs and verify only your changed values have changed.
    - Alternatively use `py -3 dbtool.py dump` or `py -3 dbtool.py dump TABLE_NAME`.


cocosolos:
`'Selh\teus'`
Why? D:
Is mob_resistances.name even used anywhere? Is it safe to change this to `'Selh\'teus'`?

TeoTwawki:
> `'Selh\teus'` Why? D: Is mob_resistances.name even used anywhere? Is it safe to change this to `'Selh\'teus'`?

That was likely accidental, only reason to have the slash is to escape the single quotemark. Which we only really _needed_ for the pool name but it was copied to other tables for consistencies sake.

p.s. resistances name is purely for humans to easy see wtf mob it belongs to without having to execute join statements or control+f multiple files

Xaver-DaRed (approved):

zach2good (approved):

zach2good:
Is it possible to make the export for float-type numbers always be 3 decimal places?

TeoTwawki:
> Is it possible to make the export for float-type numbers always be 3 decimal places?

This is dictated by the column definition `float(7,3)`

to change it, just change the line defining the float

TeoTwawki:
to change it, just change the line defining the float

cocosolos:
The database knows how much to pad, but that gets lost when it gets pulled into Python. I could probably make _all_ floats 3 decimal places, which would make some columns longer. I tried looking into it and it didn't seem possible to have variable padding based on what's set in the DB. The DB does handle it though so when viewing in the DB everything is padded as it should be.

TeoTwawki:
personally I only care about the format of the dump (and mysqldump.exe is going to obey the column define for how many trailing zero to retain - if using something that doesn't, reconsider) and the data inside the live db (same deal there) so whatever from it passes through on the way to and from those shouldn't matter, the end result is going to look like `1.000` and `1.250` at both ends.

cocosolos:
Okay after some further research it looks like I may be able to use the precision and decimal settings from the DB, will see what I can do.

cocosolos:
Sorry about all the revisions, should be set now. It should match the decimals used by the database, which is 3 in most places and a few 1, 2, and undefined that are handled as well.

TeoTwawki (approved):

zach2good (approved):
Yeet, yeet I say!

PR #4344, opened 2023-08-24
[lua] Add conditional logger factory

zach2good:
**_I affirm:_**

Uses some Lua black magic to allow us to automatically look up `player` in the context in which your logger was called, and print out the same message to that client.

Converted chocobo raising logging as an example

EDIT: Also refactors our cpp binding for Lua `print` and refactors things to make Lua's `tostring` use that internal logic.


TeoTwawki:
image

Xaver-DaRed (approved):

PR #4346, opened 2023-08-24
[core] Changed any Russian comments to English within mixed files

EpicTaru:
**_I affirm:_**

- Translated instances of Russian language to English language in comments within following files:
  - src/map/ability.cpp
  - src/map/ability.h
  - src/map/char_recast_container.cpp
  - src/map/conquest_system.cpp
  - src/map/enmity_container.cpp
- Also corrected any incorrect grammar in comments found while performing above changes.


src/map/conquest_system.cpp
```diff
@@ -713,7 +713,7 @@ namespace conquest
             charutils::AddPoints(PChar, charutils::GetConquestPointsName(PChar).c_str(), points);
             GainInfluencePoints(PChar, points / 2);
         }
-        return 0; // added conquest points (пока не вижу в этом определенного смысла)
+        return 0; // added conquest points (I don't see any point in it yet)
```
TeoTwawki: some of these can be deleted instead of translated. pretty much when it is non helpful like this one is, axe it.

past me is guilty of this in English BUT: comments should be guiding the reader one understanding the code, not simply talking to the reader. if its not doing that, 99% chance it can be removed.

EpicTaru:
Also attempting to fix that error with initial commit message length.

EpicTaru:
Thank you, Teo, for the assist!

src/map/ability.cpp
```diff
@@ -315,7 +315,7 @@ uint16 CAbility::getAoEMsg() const
 
 /************************************************************************
  *                                                                       *
- *  Реализация namespase для работы со способностями                     *
+ *  Namespace implementation for working with abilities                  *
```
TeoTwawki, line 318: reminder to future self to wage war against header bricks stating the obvious.

TeoTwawki (approved):

src/map/ability.cpp
EpicTaru: I'll also join in this war in further translation commits I make!

PR #4357, opened 2023-08-25
[Fix] Elemental mobs physical resistances and detection in Central Temenos 2F and Eastern tower

kelppo:
**_I affirm:_**

Before this commit, the Elementals in Central Temenos - 2nd Floor take 0 physical damage because of how they were set up, fix this by flipping the sign of xi.mod.UDMGPHYS and xi.mod.UDMGMAGIC to match scripts/battlefields/Temenos/temenos_eastern_tower.lua and update the comment in there.

1. !pos 580.000 -2.375 104.000 37
2. !addkeyitem white_card
3. !addkeyitem cosmo_cleanse
4. Enter Central Temenos - 2nd Floor, observe how the elementals take no physical damage
5. Apply this patch
6. Observe how they now take physical damage

### DIFF
--- a/scripts/battlefields/Temenos/temenos_eastern_tower.lua
+++ b/scripts/battlefields/Temenos/temenos_eastern_tower.lua
@@ -279,12 +279,16 @@ content.groups =
             "Dark_Elemental_E",
         },
 
-        -- NOTE: Elementals take double physical damage because their family resistance is 25% so it totals to 50% resistance
+        -- NOTE: Elementals in here take 50% physical damage instead of the usual 25%
+        -- TODO: Verify if the Elementals here should detect sound
         mods =
         {
-            [xi.mod.UDMGPHYS ] = 10000,
-            [xi.mod.UDMGMAGIC] = -5000,
-            [xi.mod.FASTCAST ] = 20,
+            [xi.mod.SLASH_SDT   ] = 500,
+            [xi.mod.PIERCE_SDT  ] = 500,
+            [xi.mod.IMPACT_SDT  ] = 500,
+            [xi.mod.HTH_SDT     ] = 500,
+            [xi.mobMod.DETECTION] = xi.detects.HEARING,
+            [xi.mod.FASTCAST    ] = 20,
         },
 
         mobMods =
--- a/scripts/battlefields/Temenos/central_temenos_2nd_floor.lua
+++ b/scripts/battlefields/Temenos/central_temenos_2nd_floor.lua
@@ -76,11 +76,14 @@ content.groups =
             "Light_Elemental",
         },
 
-        -- NOTE: Elementals take double physical damage because their family resistance is 25% so it totals to 50% resistance
+        -- NOTE: Elementals in here take 50% physical damage instead of the usual 25%
+        -- TODO: Verify if the Elementals here should detect sound
         mods =
         {
-            [xi.mod.UDMGPHYS    ] = -10000,
-            [xi.mod.UDMGMAGIC   ] = 5000,
+            [xi.mod.SLASH_SDT   ] = 500,
+            [xi.mod.PIERCE_SDT  ] = 500,
+            [xi.mod.IMPACT_SDT  ] = 500,
+            [xi.mod.HTH_SDT     ] = 500,
             [xi.mobMod.DETECTION] = xi.detects.HEARING,
         },
     },


Xaver-DaRed:
Ok,so...
```c++
        float resist = 1.f + PDefender->getMod(Mod::UDMGPHYS) / 10000.f;
        resist       = std::max(resist, 0.f);
        damage       = (int32)(damage * resist);

        resist = 1.f + PDefender->getMod(Mod::DMGPHYS) / 10000.f + PDefender->getMod(Mod::DMG) / 10000.f;
        resist = std::max(resist, 0.5f);                        // PDT caps at -50%
```

Translation: regular DMGPHYS + DMG are capped at 50% and multiplicative to UPHYSDMG.
This is a mess already, because we also have this...
```c++
    SLASH_SDT  = 49, // Slash Damage Taken
    PIERCE_SDT = 50, // Piercing Damage Taken
    IMPACT_SDT = 51, // Impact Damage Taken
    HTH_SDT    = 52, // Hand-To-Hand Damage Taken
```

This are the values in database for its physical resistance by dmg type
```sql
INSERT INTO `mob_resistances` VALUES (99,'Elemental-Air',0.25,0.25,0.25,0.25,0,0,0,0,0,0,0,0,0,0,-3,11,11,0,0,0,0);
```

Abilities that affect physical damage are supposed to work on this mobs though, so I propose that, since the elementals at both Temenos are supposed to take 1/2 damage from physical, instead of  their regular 1/4, instead of fucking arround with UDMGPHYS, lets just set the dmgtype modifiers to a 1/2 to override the originals.

So
```lua
[xi.mod.SLASH_SDT ] = 500,
[xi.mod.PIERCE_SDT] = 500,
[xi.mod.IMPACT_SDT] = 500,
[xi.mod.HTH_SDT   ] = 500,
```
Yes, those are the actual values, those modifier use a different scale.

This, regarding physical damage. More to come on magical damage once I actually research it.
This would need an update both on the mobs of central Temenos 2F and eastern tower

Xaver-DaRed:
Ok, as for magical SDT, that modifier entry should be just removed entirely. Both in Central Temenos 2nd floor and in Temenos Eastern Tower.

In short, final result should look like this for both:
```lua
mods =
{
    [xi.mod.SLASH_SDT   ] = 500,
    [xi.mod.PIERCE_SDT  ] = 500,
    [xi.mod.IMPACT_SDT  ] = 500,
    [xi.mod.HTH_SDT     ] = 500,
    [xi.mobMod.DETECTION] = xi.detects.HEARING,
},
```

Xaver-DaRed (changes requested):
Read the comments I left and apply them to the PR.
Or propose a better method. That works aswell.

Also, the comments in the code should be updated to reflect the new way of handling (removing mention to "doubling damage") since this modifiers are now being set to their proper values instead of applying a multiplicative modifier on top.

jmcmorris:
@Xaver-DaRed Much better solution than I had. I didn't consider changing the SDT directly while doing this. Nice work! 👍

kelppo:
PR updated, I re-tested both East and 2nd Floor and they take approximately the same damage with Xaver's changes.

Xaver-DaRed (approved):

scripts/battlefields/Temenos/temenos_eastern_tower.lua
```diff
@@ -279,12 +279,15 @@ content.groups =
...
         mods =
         {
-            [xi.mod.UDMGPHYS ] = 10000,
-            [xi.mod.UDMGMAGIC] = -5000,
-            [xi.mod.FASTCAST ] = 20,
+            [xi.mod.SLASH_SDT   ] = 500,
+            [xi.mod.PIERCE_SDT  ] = 500,
+            [xi.mod.IMPACT_SDT  ] = 500,
+            [xi.mod.HTH_SDT     ] = 500,
+            [xi.mobMod.DETECTION] = xi.detects.HEARING,
```
TeoTwawki, line 290: Detection change ~~and fast cast change~~ weren't mentioned in the commit or the pr description. If accurate to retail, they should have been mentioned.

_partially fooled by the highlighting_

kelppo: I can't test on retail but ffxiclopedia says that they only detect sound in East.
https://ffxiclopedia.fandom.com/wiki/Fire_Elemental
Should we leave it like before this PR? Only the Elementals in East detect sound.

kelppo: Never mind, before this PR only 2nd Floor Elementals detect sound, just let me know how to go about this.

TeoTwawki: Next time just add it to your commit message. We can add it in if we squash your commits, but I don't like editing people messages.

TeoTwawki: > Next time just add it to your commit message. We can add it in if we squash your commits, but I don't like editing people messages.

I thought this was pretty clear. I was **not** looking for another commit removing it after a commit adding it. I just want commit messages to reflect everything happening in a commit.

Xaver-DaRed:
This one is good to go

PR #4359, opened 2023-08-25
[core] Battleentity.cpp/.h: Translated Russian to English

EpicTaru:
**_I affirm:_**

Within **Battleentity.cpp**/**.h**:
- Translated Russian comments to English
- Also removed any un-necessary/non-helpful translations
  - If I need to remove more that are deemed un-necessary/non-helpful, please let me know.
- Corrected grammar/formatting of any pre-existing comments that I came across during the above processing

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -151,7 +151,7 @@ bool CBattleEntity::isSitting()
 
 /************************************************************************
  *                                                                       *
- *  Пересчитываем максимальные значения hp и mp с учетом модификаторов   *
+ *  Recalculate hp and mp maximum values, taking into account modifiers  *
  *                                                                       *
  ************************************************************************/
 
@@ -185,7 +185,7 @@ void CBattleEntity::UpdateHealth()
 
 /************************************************************************
  *                                                                       *
- *  Получаем текущее количество очков жизней                             *
+ *  Get current number of hit points                                     *
  *                                                                       *
  ************************************************************************/
 
@@ -201,7 +201,7 @@ int32 CBattleEntity::GetMaxHP() const
 
 /************************************************************************
  *                                                                       *
- *  Получаем текущее количество очков маны                               *
+ *  Get current number of mana points                                    *
  *                                                                       *
  ************************************************************************/
 
@@ -463,7 +463,7 @@ uint16 CBattleEntity::GetRangedWeaponRank()
 
 /************************************************************************
  *                                                                       *
- *  Изменяем количество TP сущности                                      *
+ *  Change the amount of TP of an entity                                 *
  *                                                                       *
  ************************************************************************/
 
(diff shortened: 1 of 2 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -527,7 +527,7 @@ int32 CBattleEntity::addHP(int32 hp)
     hp        = health.hp - cap;
     health.hp = cap;
 
-    // если количество жизней достигает нуля, то сущность умирает
+    // if the number of lives reaches zero, then the entity dies
```
TeoTwawki: this can be removed. "lives" is incorrect and we all know what HP reaching zero means

TeoTwawki: actually the majority of comments in the header bricks are uselessly stating the obvious

src/map/entities/battleentity.cpp
```diff
@@ -512,7 +512,7 @@ int16 CBattleEntity::addTP(int16 tp)
 
 /************************************************************************
  *                                                                      *
- *  Изменяем количество жизней сущности                                 *
+ *  Change the number of lives of the entity                            *
```
TeoTwawki: `Change current HP of entity` if kept at all

src/map/entities/battleentity.cpp
```diff
@@ -151,7 +151,7 @@ bool CBattleEntity::isSitting()
 
 /************************************************************************
  *                                                                       *
- *  Пересчитываем максимальные значения hp и mp с учетом модификаторов   *
+ *  Recalculate hp and mp maximum values, taking into account modifiers  *
```
TeoTwawki, line 154: as an example: this one is actually mildly useful, because it lets the reader know this function INCLUDES modifiers before they go into the actual code to find that out

src/map/entities/battleentity.cpp
EpicTaru: Removing, thanks.

src/map/entities/battleentity.cpp
EpicTaru: Noted, I'll do a sweep and remove any other Capt. Obvious comments too

src/map/entities/battleentity.h
```diff
@@ -239,9 +239,9 @@ enum SLOTTYPE
 
 #define MAX_SLOTTYPE 18
 
-// CROSSBOW и GUN - это Piercing, разделение сделано из-за одинакового skilltype
-// для возможности различить эти орудия при экипировке и избавиться от ошибки
-// использования пуль с арбалетом и арбалетных стрел с огнестрельным оружием (только персонажи)
+// CROSSBOW and GUN are Piercing, the division is made due to the same skilltype
```
TeoTwawki: This can go because skill type does not universally correlate to damage type - there exceptions in multiple weapon categories where the skill type remains the same but the damage type does not. The most obvious ones being some old items like Heavy Halberd and Vampiric Claws. The lack of ranged weapons with unusual types at time of comment doesn't preclude the possibility of them existing later, but the comment leads readers to think this will always be true.

Thus it only serves to potentially misinform the reader.

TeoTwawki (approved):

PR #4363, opened 2023-08-26
Correct Automaton maneuver tracking with optic fiber and negative mods

claywar:
**_I affirm:_**
- [x] I have _**tested the hell out of my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

* Adds two Lua bindings for getting a single or a table of automaton attachments
* Corrects issue with tracking negative-value mods for attachments
* Moves status effect removal calls to automatons to a point after the effect is removed
* Calculates optic fiber bonuses outside of mod when juggling effects

Equip ARK3, Tension Spring 4, Optic Fiber, and Armor Plate II attachments and observe values transitioning to proper values for any combination of maneuvers

### DIFF
--- a/scripts/globals/automaton.lua
+++ b/scripts/globals/automaton.lua
@@ -187,8 +187,6 @@ local function getRefreshModValue(pet, attachmentName, numManeuvers)
     return regenRefreshFormulas[attachmentName][1][numManeuvers + 1] + petMaxMP * (regenRefreshFormulas[attachmentName][2][numManeuvers + 1] / 100)
 end
 
--- TODO: This may still be necessary with future updates
---[[
 local function isOpticFiber(attachmentName)
     if string.find(attachmentName, 'optic_fiber') ~= nil then
         return true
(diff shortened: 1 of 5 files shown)


Xaver-DaRed (approved):

scripts/globals/automaton.lua
```diff
@@ -187,16 +187,30 @@ local function getRefreshModValue(pet, attachmentName, numManeuvers)
...
-]]--
+
+local function calculatePerformanceBoost(pet, numManeuvers)
+    local master = pet:getMaster()
+    local performanceBoost = 0
+
+    local numLightManeuvers = master and master:countEffect(xi.effect.LIGHT_MANEUVER) or 0
+    for _, attachmentObj in ipairs(pet:getAttachments()) do
+        local attachmentName = attachmentObj:getName()
+        print(attachmentName)
```
WinterSolstice8: `print`

claywar: Doh! Thought I nuked them all out.  Good catch

WinterSolstice8 (approved):

PR #4367, opened 2023-08-27
[ci] Add rule to not allow code after `then` in conditional

claywar:
**_I affirm:_**

Adds CI rule to check for any non-comment text after the word `then` in lua scripts and corrects all issues raised from it.

No changes should be observed

### DIFF
--- a/scripts/actions/mobskills/astral_flow_pet.lua
+++ b/scripts/actions/mobskills/astral_flow_pet.lua
@@ -30,6 +30,19 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
     return 0
 end
 
+-- [mobskillId] = { petFamily1, petFamily2, ... }
+local petAstralFlowAbility =
+{
+    [839] = { 36, 381 }, -- Fenrir (Howling Moon)
+    [913] = { 38, 383 }, -- Ifrit (Inferno)
+    [914] = { 45, 388 }, -- Titan (Earthen Fury)
+    [915] = { 40, 384 }, -- Leviathan (Tidal Wave)
+    [916] = { 37, 382 }, -- Garuda (Aerial Blast)
+    [917] = { 44, 387 }, -- Shiva (Diamond Dust)
+    [918] = { 43, 386 }, -- Ramuh (Judgment Bolt)
+    [919] = { 34, 379 }, -- Carbuncle (Searing Light)
+}
+
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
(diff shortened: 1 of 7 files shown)


scripts/actions/mobskills/astral_flow_pet.lua
```diff
@@ -42,18 +55,12 @@ mobskillObject.onMobWeaponSkill = function(target, mob, skill)
...
-    elseif petFamily == 38 or petFamily == 383 then skillId = 913 -- ifrit     inferno
-    elseif petFamily == 40 or petFamily == 384 then skillId = 915 -- leviathan tidal wave
-    elseif petFamily == 43 or petFamily == 386 then skillId = 918 -- ramuh     judgment bolt
-    elseif petFamily == 44 or petFamily == 387 then skillId = 917 -- shiva     diamond dust
-    elseif petFamily == 45 or petFamily == 388 then skillId = 914 -- titan     earthen fury
-    else
-        printf("[astral_flow_pet] received unexpected pet family %i. Defaulted skill to Searing Light.", petFamily)
+    for mobSkillId, petFamilyList in ipairs(petAstralFlowAbility) do
+        if utils.contains(petFamily, petFamilyList) then
+            skillId = mobSkillId
```
Xaver-DaRed, line 62: no break for the loop when a match happens?

Xaver-DaRed: Also, `skillId` default value is already 919. Seems reduntant to set it to 919 again if a match happens.

Xaver-DaRed (approved):
Nitpicks apart, lgtm

scripts/actions/spells/black/bio_iii.lua
```diff
@@ -45,20 +45,31 @@ spellObject.onSpellCast = function(caster, target, spell)
     -- http://wiki.ffo.jp/html/1954.html
     -- this is a tiered calculation that has at least three tiers,
     -- so I'll use breakpoints for human readability
-    local dotdmg = 0
-    if     skillLvl > 400 then dotdmg = 17
```
TeoTwawki, line 49: ya know, I even called out this exact styling in the style guide years ago. _Reeeeeeee no padding between the if and the condition_

Ty for fixing.

TeoTwawki (approved):

scripts/actions/mobskills/astral_flow_pet.lua
claywar: You're right, this is a carryover from a way it was previously handled

PR #4374, opened 2023-08-29
Mob being incorrectly classified as NMs

cmastro24:
**_I affirm:_**

These values were erroneously changed in #3884 causing various mobs to check Impossible to Gauge.  This change reverts some of that work to make the mob types regular.

Find an elemental and check to see if not impossible to gauge.


personaone:
Nice I wanted to report some random Ice Elementals flying around as NMs 😊

TeoTwawki:
this likely means some battlefield mobs are sharing data with non battlefield mobs
```cpp
enum MOBTYPE
{
    MOBTYPE_NORMAL      = 0x00,
    MOBTYPE_0X01        = 0x01, // available for use
    MOBTYPE_NOTORIOUS   = 0x02,
    MOBTYPE_FISHED      = 0x04,
    MOBTYPE_CALLED      = 0x08,
    MOBTYPE_BATTLEFIELD = 0x10,
    MOBTYPE_EVENT       = 0x20
};
```
looking back in my original pr many were flagged as `2` and I moved them to `16` (0x10) and some were indeed `0` and moved `16` by the same query, which looked at every bcnm row and hunted down the mobid in it and pool of said mob ID to do the edit.

so when we find BCNM mob that are not checking notorious and or misbehaving, those bcnm mob will need new pools so that their flagging can be diff --or- perhaps we can on-the-fly change mobtype at bcnm start? I wasn't expecting this scenario to happen or I'd have just done that to start with. testing rando mobs unrelated to what I was changing didn't cross my mind.

TeoTwawki (approved):

PR #4376, opened 2023-08-29
[core] Add UUID util

zach2good:
**_I affirm:_**

To be used later for item disambiguation

```cpp

#include "common/uuid.h"

...

    ShowInfo(uuid::GenerateUUID());
```

```
[08/29/23 15:40:23:299][connect][info] 40dba40e-8784-4650-8eef-6f51957c58d4 (ConnectServer::ConnectServer:74)
[08/29/23 15:40:23:299][connect][info] 91e64cc5-477d-410d-97b1-d48810ecc93d (ConnectServer::ConnectServer:75)
```

### DIFF
--- a/ext/CMakeLists.txt
+++ b/ext/CMakeLists.txt
@@ -131,7 +131,7 @@ CPMAddPackage(
 CPMAddPackage(
     NAME json
     GITHUB_REPOSITORY nlohmann/json
-    GIT_TAG v3.11.2
+    GIT_TAG 254395e5f7731b5595e7c11ac4b929c923582114
 ) # defines: nlohmann_json::nlohmann_json
 
 CPMAddPackage(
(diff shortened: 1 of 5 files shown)


TeoTwawki (dismissed):
I'm assuming it'll still be ok after you're done force pushing formatting over it

claywar (approved):

zach2good:
I brought in the content of `crashoz/uuid_v4` into a single header, since that repo builds the examples by default on import and it was starting to be a pain in the ass. I might submit some changes back to them to fix it, but for now, Win64+Linux gets the SIMD code, and Win32+OSX gets the manual code

zach2good:
I'll probably need to add in more compiler flags to force this behaviour, since people out there use strange architectures that don't have AVX2 and friends

zach2good:
I've opted to rip out the fancy SIMD code for now, with a TODO to maybe add it back in later. The majority of our UUID generation will be happening in MariaDB with `UUID()` anyway

PR #4377, opened 2023-08-29
Add check for CDespawnState to getCurrentAction()

TeoTwawki:
**_I affirm:_**

This used to be here but was removed because in most places, we consider despawned mobs to also be "dead" and the next step is then considered waiting for respawn.

But if this is a mob that was not killed, but rather despawned by script in a zone suspended because its empty then it is not "dead" and it is not waiting to respawn either.

It won't transition until the 1st tick after the zone wakes (this is expected behavior).

The action type is enumerated so I've given it the enumerated value rather than lumping it in with death, and placed its check after the existing death check - this keeps all existing behavior intact but in the event a mob doesn't "count" as dead or waiting to respawn, we'll at least know its in this state.

I am not 100% on this fixing what made me notice it, but it for sure does no harm and everything works as it did before.

`!getcurrentaction` some mobs, hopefully not get any messages about unhandled states.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -1471,7 +1471,7 @@ void CLuaBaseEntity::setStatus(uint8 status)
  *  Purpose : Returns the current state of a non-NPC entity
  *  Example : if target:getCurrentAction() ~= xi.act.MOBABILITY_USING then
  *  Notes   : Function name ambiguous, but getCurrentState() in use already
- *          : See globals/status.lua for action definitions
+ *          : See scripts/enum/action.lua for action definitions
  ************************************************************************/
 
 uint8 CLuaBaseEntity::getCurrentAction()
@@ -1486,7 +1486,7 @@ uint8 CLuaBaseEntity::getCurrentAction()
 
     if (m_PBaseEntity->PAI->IsStateStackEmpty())
     {
-        action = 16;
+        action = 16; // For mobs, this means roaming
     }
     else if (m_PBaseEntity->PAI->IsCurrentState<CRespawnState>())
     {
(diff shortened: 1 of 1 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -1471,7 +1471,7 @@ void CLuaBaseEntity::setStatus(uint8 status)
  *  Purpose : Returns the current state of a non-NPC entity
  *  Example : if target:getCurrentAction() ~= xi.act.MOBABILITY_USING then
  *  Notes   : Function name ambiguous, but getCurrentState() in use already
- *          : See globals/status.lua for action definitions
+ *          : See \scripts\enum\action.lua for action definitions
```
claywar: Can you change these to forward slashes to mimic what is used in lua?

claywar (approved):

src/map/lua/lua_baseentity.cpp
TeoTwawki: done

zach2good:
It took me a second to realise what those awful compiler warnings were about:

You need to add `despawn_state.h` to this block at the top of that file, it doesn't know about it:
```cpp
#include "ai/states/ability_state.h"
#include "ai/states/attack_state.h"
#include "ai/states/death_state.h"
#include "ai/states/inactive_state.h"
#include "ai/states/item_state.h"
#include "ai/states/magic_state.h"
#include "ai/states/mobskill_state.h"
#include "ai/states/petskill_state.h"
#include "ai/states/raise_state.h"
#include "ai/states/range_state.h"
#include "ai/states/respawn_state.h"
#include "ai/states/weaponskill_state.h"
```

zach2good (approved):

PR #4380, opened 2023-08-30
Custom Module - Chigoes

CatsEyeXI:
**_I affirm:_**

Implements a custom module that will spawn chigoes from the mobs defined in the table.

Source: https://www.bg-wiki.com/ffxi/Category:Chigoe

**Features**
- A chigoe will spawn the each time the defined mob uses a mobskill (up to 5 times)
- Chigoe will get one-shot with a critical hit
- Chigoe will get one-shot when hit with a pre-defined list of job abilities (bashes, etc)
- Chigoe will get one-shot when hit with a weaponskill

Load module, visit one of the defined mobs in the table, engage it and !tp 3000 it.
It will spawn a chigoe up to 5 times when using a mob skill.


modules/custom/lua/mixin_marid.lua
```diff
@@ -0,0 +1,151 @@
...
+    { 'Caedarva_Mire',      'Mosshorn'         },
+}
+
+-- validate that the file exists before trying to override it or weird things happen...
+local pathWithFilename = io.popen('cd'):read'*all'
+
+for _, entry in pairs(mobs) do
+    local zoneName = entry[1]
+    local mobName  = entry[2]
+    local filename = pathWithFilename:sub(1, -2) .. string.format('\\scripts\\zones\\%s\\mobs\\%s', zoneName, mobName) .. '.lua'
```
claywar: There's absolutely no reason a module should have to hit the disk.  All of these scripts should be cached beforehand.

CatsEyeXI: some mobs don't have existing scripts. using xi.module.ensureTable on files that do exist has undesired results, i.e. some will stop working all together. I think fixing that is out of scope for this module (and I have no clue how to fix it), what do you propose?

claywar: Everything is cached in the xi table.  It should be as simple as looking up there if something exists.

TeoTwawki:
If I am not mistaken those chigoe exist in the dats right before and after the Marid IDs and don't have to be dynamic if offsets are used, tho they can be.

Retail accurate content doesn't belong in a module tho. Base implimentation would go into a mixin before any customizing in a module.

p.s. there is also a limit on active/live chigs out at once, and if its hit new ones don't pop from any marid till a corpse fades out

CatsEyeXI:
> If I am not mistaken those chigoe exist in the dats right before and after the Marid IDs and don't have to be dynamic if offsets are used, tho they can be.
> 
> Retail accurate content doesn't belong in a module tho. Base implimentation would go into a mixin before any customizing in a module.

I wasn't sure this would be accepted as "retail accurate" based on the fact that it's using dynamic entities, but if you are OK with that, then I can massage this into a mixin and try to figure out the disk stuff...

CatsEyeXI:
will resubmit

PR #4350, opened 2023-08-24
Adds more capture data for Garrison (For feature branch)

loxleygames:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?
Captured and updated starting positions for:
* West Sarutabaruta
* Jugner Forest
* Meriphataud Mountains
* Qufim Island
* Beaucedine Glacier
* The Sanctuary of Zi'tah
* Yhoator Jungle

Captured and added NPC looks for:
* San d'Oria 50
* Bastok 30
* Windurst 20/30/40

Removed deprecated `xi.garrison.getMobInfo`
Updated placeholders module to better reflect the estimated missing NPCs

## Steps to test these changes

Go to a Garrison area and use `!garrison start`

PR #4382, opened 2023-08-30
Adds capture data for Garrison (Windurst 99)

loxleygames:
**_I affirm:_**

Adds capture data for the last missing combination: Windurst 99 (75)

N/A

### DIFF
--- a/scripts/globals/garrison_data.lua
+++ b/scripts/globals/garrison_data.lua
@@ -21,7 +21,7 @@ xi.garrison.allyNames =
 {
     [20] = { [xi.nation.SANDORIA] = "Patrician",       [xi.nation.BASTOK] = "Recruit",         [xi.nation.WINDURST] = "Candidate"       },
     [30] = { [xi.nation.SANDORIA] = "Trader",          [xi.nation.BASTOK] = "Mariner",         [xi.nation.WINDURST] = "Scholar"         },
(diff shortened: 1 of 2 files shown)


zach2good (approved):
LGTM

Same question as last time, can you link the capture these are from?

https://github.com/LandSandBoat/server/pull/4350#issuecomment-1692407025

scripts/globals/garrison_data.lua
```diff
@@ -151,7 +151,8 @@ xi.garrison.allyLooks =
 
         [xi.nation.WINDURST] =
         {
-            -- (Missing capture)
+            "0x010009043E106F203E303E403E501C611B700000",
+            "0x0100020349106F204930494049501F611B700000",
```
Frankie-hz, line 155: Not sure if I should comment on this here or in the other PR but can we add comments to know what model these actually are? If someone wants to check all of these in the future it would very helpful.

LoxleyXI: It says the level and the nation, which is specific enough (They are always the same) and more descriptive than the source version. Unless you want me to describe what they're wearing, I'm not sure what you're asking for.

LoxleyX:
Captures have been submitted

scripts/globals/garrison_data.lua
zach2good: @Frankie-hz I'll add descriptions of these looks onto the feature branch, it's done easily enough with `!costume lookstring` or `!exec player:setModelId(lookstring)`

zach2good: Tracking here: https://github.com/LandSandBoat/server/issues/4386

PR #4428, opened 2023-09-03
[DNC] Add Terpsichore Finishing Move bonus to job_util

Xaver-DaRed:
**_I affirm:_**

Adds (all versions of) Terpsichore bonus to Finishing Moves.
As per BG wiki, this only happens with weapon equiped on your main hand, so added a TODO to confirm.

Equip any Terpsichore (except lvl 75 unfinished version) in main hand.
Do whatever dancers do to get finishing moves.
See a bonus.


scripts/globals/job_utils/dancer.lua
```diff
@@ -62,6 +76,14 @@ local function getStepFinishingMovesBase(player)
         numAwardedMoves = 2
     end
 
+    -- Terpsichore main-hand FM bonus.
+    -- TODO: Confirm in retail the bonus only applies when the weapon is equiped in main hand.
+    local mainHandWeapon = player:getEquipID(xi.slot.MAIN)
+
+    if terpsichoreTable[mainHandWeapon] then
```
claywar, line 82: Steps != Finishing moves.  Steps are the number of applications of a specific type of step (Cap is 10, regardless)

claywar (approved):

PR #4430, opened 2023-09-03
[core] Various item_.cpp/.h files: Translated Russian to English

EpicTaru:
**_I affirm:_**

Within /src/map/items/**item_equipement.cpp**/**.h**, **item_shop.h**,  and **item_usable.cpp**/**.h**
- Translated Russian comments to English but ultimately just removed the comments because they were deemed non-helpful/not needed as the code itself preceding the comments already gave sufficient explanation of what the code was doing.
  - If more are deemed un-necessary/non-helpful, please let me know, I'm doing my best to weed out these.


src/map/items/item_equipment.cpp
```diff
@@ -43,7 +43,6 @@ CItemEquipment::CItemEquipment(uint16 id)
 
 CItemEquipment::~CItemEquipment()
 {
-    // ни в коем случае не освобождать здесь указатели на модификатоты и спецеффекты. они глобальны.
```
Xaver-DaRed, line 46: This function is empty.
Google claims that comment means "in no case free here pointers to modifiers and special effects. they are global."
Which I would guess that means something for someone.

Xaver-DaRed (approved):

src/map/items/item_equipment.cpp
TeoTwawki: slightly more accurate: `in any case, do not release pointers to modifiers and special effects here. they are global.`

TeoTwawki: its just a warning to not destroy those objects here

TeoTwawki (approved):

PR #4432, opened 2023-09-04
Adds Phantom Roll recast modifier and corrects Phantom Roll recast reductions

N3ckB3ard:
**_I affirm:_**

 This PR aims to add the "Phantom Roll" delay modifier and Augment modifier. During the creation of this modifier I noticed the 11 recast timer was not matching retail. On retail when a 11 is rolled, you will receive half of the base cost of the ability with the addition of the merit and modifier. In my testing, I was able to receive a half of the recast timer without the effect of "Bust", when the "Bust" effect was active I only received the merit and modifier bonus. 
---
Example Recast:
Base - 1min 
Merits - 50 sec
Augmented Pants - 45 sec
11 Roll - 30 sec
11 Roll + Merits - 20 sec
11 Roll + Merits + Pants - 15 sec

Bust + 11 Roll + Merits + Pants - 45 sec

---
You can see these recast changes in the video's shared below
https://drive.google.com/file/d/1wfOajWcn6R60oMVeaSqe5SEbgy71x4Fd/view?usp=sharing
https://drive.google.com/file/d/1iEnv6WH4HrKzPG5VGptoNV4wLEHS5Yun/view?usp=sharing

Gear and Merits:
image
image

I rebuilt my local server and used the !setmod command for the modifier 1076 set to 5 and received a 5 sec recast reduction.
This set my recast to 55 sec
Under the effects of a 11 with the modifier set, I received a 25 sec recast timer.
Fully merited I received a 15 sec recast timer


claywar (approved):
Barring CI, seems reasonable to me.  My only request outside of that would to be to rename the `bust` variable to something more descriptive, such as `numBustEffects`

Xaver-DaRed (approved):

claywar (approved):

PR #4437, opened 2023-09-04
Add support for charVars and serverVars that expire based on timestamp

### LINKED ISSUE #4218: 🔨 Create interaction variable type that automatically clears on JST Midnight (or other intervals)
**_I affirm:_**

## Describe the feature
To help to prevent forever vars, create a naming convention for a variable that is searched and cleared on JST Midnight.  This might need some additional work in case the server is not alive when this is hit to ensure clear

claywar:
**_I affirm:_**

Adds an additional column to char_vars and server_variables which accepts a unix timestamp for when the variable will expire.  On next access, if expired the variable will be deleted from the db.  No changes are required, and this addition is backwards compatible with existing charvar usage.

**NOTE**: All times referenced are based on the CVanaTime instance of `getSysTime()`.  The helper functions created will return system time offsets, and _NOT_ vanadiel time.

### Why?
Between this and unique events, this should be one of the final nails in the coffin for "Forever" vars.  There are some things that we track in the database that can never go away (such as reset times for Avatar fights).  This allows for an alternative, where we can set a variable to expire, instead of persisting when the next time is available.

An expired variable as described above, is one that is either deleted on access, or on server startup if its timestamp exists in the past.

### Usage
An optional parameter has been added to setCharVar (and its variants, including Interaction Framework) for an optional expiry field.  In addition, the following helper functions have been added/updated:

**Lua**
```
xi.mission.setVar           = function(player, areaId, missionId, name, value, expiry)
xi.mission.setVarExpiration = function(player, areaId, missionId, name, expiry)
xi.quest.setVar             = function(player, areaId, missionId, name, value, expiry)
xi.quest.setVarExpiration   = function(player, areaId, missionId, name, expiry)
```

**Interaction**
```
function Container:setVar(player, name, value, expiry)
function Container:setVarExpiration(player, name, expiry)

-- Wrapper for setVar where the expiration matters more than any other
-- stored value.  There is no need to get specifically, since it will
-- be removed.  (Returned value is either 1 or 0)
function Container:setTimedVar(player, name, expiry)
```

**Core**
```
CLuaBaseEntity::setCharVar(std::string const& varName, int32 value, sol::object const& expiry)
CLuaBaseEntity::setCharVarExpiration(std::string const& varName, uint32 expiry)
```

This is system-time based, and there have been the following Lua bindings created to assist:

* `NextGameTime(uint32 intervalSeconds)`

This function accepts any value, but should be exclusively used with the xi.vanaTime enum (see below):
```lua
local secondsPerTick = 2.4

xi.vanaTime =
{
    YEAR  = 518400 * secondsPerTick,
    MONTH = 43200 * secondsPerTick,
    WEEK  = 11520 * secondsPerTick,
    DAY   = 1440 * secondsPerTick,
    HOUR  = 60 * secondsPerTick,
}
```

By passing one of these enum values in, you will receive a system timestamp for when the next event will occur.

* `NextJstDay()` -> This is just another name for `JstMidnight`, since there will most likely be a CI rule to ensure that a vanaTime value is not passed into these functions

* `NextJstWeek()` -> This will return the next time a weekly reset can occur, such as Conquest Tally

### What if what I need isn't covered?
There's a lot of potential things that we could possibly need to time, so this solution gives you a starting point, and then math can be applied to get the desired value.  For example, let's say you need 3 Vana'diel days from now:

`NextGameTime(xi.vanaTime.DAY)` will get you through one, but we now need two more, so:

`NextGameTime(xi.vanaTime.DAY) + xi.vanaTime.DAY * 2` will get you the rest of the way.

### But I need real time!
Simple!  Just take your current system timestamp, and apply the offset, so if I needed Exactly 3 Vana'diel days from right now:

`GetSystemTime() + xi.vanaTime.DAY * 3`

(Foreshadowing, do not rely on os.time(), since things will be changing soon)

#### Char Vars
* [x] Set a charVar in lua with no expiration time set
* [x]  Set a charVar in lua with an expiration in the future
* [x]  Set a charVar with expiration in the past
* [x]  Get a charVar with no expiration
* [x]  Get an unexpired charVar
* [x]  Get an expired charVar
* [x]  Set a charVar with the player on a separate process that expires in the future
* [x]  Set a charVar with the player on a separate process that does not expire
* [x] Set a charVar with the player on a separate process that exists in the past
* [x]  Get a charVar with the player on a separate process that does not expire
* [x]  Get a charVar with the player on a separate process with expiration in the past
* [x]  Get a charVar with the player on a separate process with expiration in the future
* [x]  Set a volatile charVar with no expiration
* [x]  Set a volatile charVar with expiration in the past
* [x]  Set a volatile charVar with expiration in the future
* [x]  Test DB migration script
* [x]  Test Fresh DB without migration
* [x]  Set a charVar in Lua without specifying an expiration time
* [x] Test expired charVar removal on server start
* [x] Verify unexpired charVars are retained after server start
* [x] Verify non-expiring charVars are retained after server start
* [x] Increment an Expired variable
* [x] Set Expiration of an expired variable
* [x] Set Expiration of an unexpired variable
* [x] Set Expiration of a variable that does not expire

#### Server Vars
NOTE: Server Vars don't have the same logic to notify separate processes at this time.

* [x] Set a Server Var with no expiration
* [x] Set a Server Var with an empty expiration field
* [x] Set a Server Var with an expiration value in the past
* [x] Set a Server Var with an expiration value in the future
* [x] Set a Volatile Server Var with no expiration
* [x] Set a Volatile Server Var with an empty expiration field
* [x] Set a Volatile Server Var with an expiration value in the past
* [x] Set a Volatile Server Var with an expiration value in the future
* [x] Get a non-expiring Volatile Server Var
* [x] Get an Unexpired Volatile Server Var
* [x] Get an Expired Volatile Server Var
* [x] Get a Server Var with no expiration
* [x] Get an Unexpired Server Var
* [x] Get an Expired Server Var
* [x] Verify Expired Server Var removal on server start
* [x] Verify Non-Expiring Server Var on server start is retained
* [x] Verify Server Var with expiration in the future is retained on server start
* [x] Test DB Migration Script
* [x] Test Fresh DB without Migration

#### Helper Functions
* [x] Verify returned value for NextGameTime (Hour, Day, Week, Month, Year intervals)
* [x] Verify returned value for NextJstWeek
* [x] Set an expiring charvar with Interaction Framework (`quest:setVar, mission:setVar`)
* [x] Check Interaction Container: setTimedVar
* [x] Check xi.quest helper: setVar
* [x] Check xi.quest helper: setVarExpiration
* [x] Check xi.mission helper: setVar
* [x] Check xi.mission helper: setVarExpiration

### DIFF
--- a/scripts/enum/vana_time.lua
+++ b/scripts/enum/vana_time.lua
@@ -1,0 +1,11 @@
+xi = xi or {}
+
+-- Number of earth seconds for each Vana'diel time interval
+xi.vanaTime =
+{
+    YEAR  = 518400,
+    MONTH = 43200,
+    WEEK  = 11520,
+    DAY   = 1440,
+    HOUR  = 60,
+}
--- a/scripts/enum/vana_time.lua
+++ b/scripts/enum/vana_time.lua
@@ -1,11 +1,14 @@
 xi = xi or {}
 
--- Number of earth seconds for each Vana'diel time interval
+-- Number of earth seconds for each Vana'diel time interval.
+-- See: https://www.bg-wiki.com/ffxi/Vana%27diel_Time
+local secondsPerTick = 2.4
+
 xi.vanaTime =
 {
-    YEAR  = 518400,
-    MONTH = 43200,
-    WEEK  = 11520,
-    DAY   = 1440,
-    HOUR  = 60,
+    YEAR  = 518400 * secondsPerTick,
+    MONTH = 43200 * secondsPerTick,
+    WEEK  = 11520 * secondsPerTick,
+    DAY   = 1440 * secondsPerTick,
+    HOUR  = 60 * secondsPerTick,
 }
--- a/scripts/globals/interaction/container.lua
+++ b/scripts/globals/interaction/container.lua
@@ -105,8 +105,19 @@ function Container:getVar(player, name)
     return player:getVar(self.varPrefix .. name)
 end
 
-function Container:setVar(player, name, value)
-    return player:setVar(self.varPrefix .. name, value)
+function Container:setVar(player, name, value, expiry)
+    return player:setVar(self.varPrefix .. name, value, expiry)
+end
+
+function Container:setVarExpiration(player, name, expiry)
+    return player:setCharVarExpiration(self.varPrefix .. name, expiry)
+end
+
+-- Wrapper for setVar where the expiration matters more than any other
+-- stored value.  There is no need to get specifically, since it will
+-- be removed.  (Returned value is either 1 or 0)
+function Container:setTimedVar(player, name, expiry)
+    return player:setVar(self.varPrefix .. name, 1, expiry)
 end
 
 function Container:isVarBitsSet(player, name, ...)
--- a/scripts/globals/missions.lua
+++ b/scripts/globals/missions.lua
@@ -800,34 +800,38 @@ xi.mission.getMissionMask = function(player)
 end
 
 -- Interaction Framework Helper Functions
-local function getVarPrefix(areaId, questId)
-    return string.format('Mission[%d][%d]', areaId, questId)
+local function getVarPrefix(areaId, missionId)
+    return string.format('Mission[%d][%d]', areaId, missionId)
 end
 
-xi.mission.incrementVar = function(player, areaId, questId, name, value)
-    return player:incrementCharVar(getVarPrefix(areaId, questId) .. name, value)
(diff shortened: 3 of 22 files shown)


claywar:
Testing completed, ready for review

EpicTaru (approved):

Xaver-DaRed (approved):

PR #4441, opened 2023-09-05
[tools] Enforce TLS 1.3 on hxiclient.py

WinterSolstice8:
**_I affirm:_**

Probably fixes CodeQL warning about insecure TLS?

CI should run without error


zach2good (approved):

WinterSolstice8:
Undrafted, but not 100% sure it will fix the CodeQL warning. Passes CI at least. I am enforcing the highest current version of TLS which should satisfy it provided CodeQL is smart enough to know.

PR #4446, opened 2023-09-05
[SQL] Add "Gold Ring" Desynth

Xaver-DaRed:
**_I affirm:_**

Allows Gold Rings to be desynthed.
- https://www.bg-wiki.com/ffxi/Gold_Ring
- https://ffxiclopedia.fandom.com/wiki/Gold_Ring

Get a gold ring
Get a lightning crystal
Get lvl 60 in goldsmithing
Profit


hookstar:
My repo went completely sideways for some reason. Ended up having to blow it up. Thanks for the save.

WinterSolstice8 (approved):

Frankie-hz:
Should this go at the end of 29000? I believe the IDs are sorted/sectioned by craft

TeoTwawki:
> Should this go at the end of 29000? I believe the IDs are sorted/sectioned by craft

when the full list was redone by tooling ages ago, it sort of happened to work out that way but as a word of warning the numbers are arbitrary and must never be re-sequenced just to keep them together with a category. If there is a hole from a mistake being removed, the hole can be filled in. Otherwise final/first free ID +1'd

Xaver-DaRed:
No harm in having all goldsmithing desynths together when theres room for it.

Moved it there

TeoTwawki:
right. we just don't gaf if in the future they don't stay together.

Frankie-hz:
Oh I completely understand they are arbitrary numbers - might as well *try* to keep it going lol

TeoTwawki:
yeah as long as there is gap to use (at this particular section...whole LOTTA gap o.O) but when we run outa gap, its not important

PR #4451, opened 2023-09-06
[core] Spell damage gives TP

sarca571ca:
**_I affirm:_**

When an entity takes damage from a spell, they gain a flat 50 tp.

Find a monster that cast spells. Attack them and wait for them to cast on you and note that you gain 50 tp.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2521,7 +2521,7 @@ namespace battleutils
             PAttacker->addTP(tp);
 
             // Targets of damaging spells gain 50 tp
-            PDefender->addTP((uint16)   50);
+            PDefender->addTP((uint16)50);
         }
 
         return damage;
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2513,10 +2513,11 @@ namespace battleutils
         if (PSpell->canTargetEnemy() && damage > 0 && PSpell->dealsDamage())
         {
             PDefender->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DAMAGE);
+
             // Check for bind breaking
             BindBreakCheck(PAttacker, PDefender);
 
-            // Do we get TP for damaging spells?
+            // Add TP for damaging spells (Only player chars who have the Occult Accumen trait)
             int16 tp = battleutils::CalculateSpellTP(PAttacker, PSpell);
             PAttacker->addTP(tp);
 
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

src/map/utils/battleutils.cpp
```diff
@@ -2519,6 +2519,9 @@ namespace battleutils
             // Do we get TP for damaging spells?
             int16 tp = battleutils::CalculateSpellTP(PAttacker, PSpell);
             PAttacker->addTP(tp);
+
+            // Targets of damaging spells gain 50 tp
+            PDefender->addTP((uint16)50);
```
claywar: This cast is misleading, and most likely unnecessary.  `addTP()`'s parameter is a signed int16, not unsigned

sarca571ca: that  makes sense. i had borrowed code from [battleutils.cpp#L2327](https://github.com/LandSandBoat/server/blob/219a712da21f650eddf07ac3d7b0eaa990d9ef96/src/map/utils/battleutils.cpp#L2327) so i just went with what was already used. Even in the start of that ```if statement``` they used int16.

zach2good:
We just thought of some things?
- Has this amount existed forever? Should it be hooked up to a setting for people that want to crank it or turn it off for era/custom purposes?
- Is the 50 amount affected by Store TP etc.?

WTB: Someone go on retail with SAM or some other high-store-TP job and see if you get more TP from getting nuked

zach2good:
Ok, I went to Castle.O on my i119 NIN (lots of Store TP) and let a Yagudo Theologist cast Thunder on me: I got 60TP - so this is affected by Store TP in some way

zach2good:
For reference, I think I have +22 Store TP on at the moment

zach2good:
In other places, I see this being used for giving TP:
```
                PAttacker->addTP(
                    (int16)(tpMultiplier * (baseTp * (1.0f + 0.01f * (float)((PAttacker->getMod(Mod::STORETP) + getStoreTPbonusFromMerit(PAttacker)))))));
```

Xaver-DaRed:
That would suggest that it follows regular store TP equation
`TP/hit = floor( Base TP/hit × (100 + Store TP Total) ÷ 100 )`

base is 50 in this case. your Store TP would give a 1.2 bonus, so 50 * 1.2 = 60

zach2good:
Maths:
```
float storeTPMultiplier = 1.0f + 0.01f * static_cast<float>(PDefender->getMod(Mod::STORETP) + getStoreTPbonusFromMerit(PDefender));
            PDefender->addTP(static_cast<int16>(50 * storeTPMultiplier);
```
```
multiplier = 1.0 + 0.01 * (22 + 0) = 1.22
50 * 1.22 = 61
```
^ I don't actually remember if I had 20 or 22 store TP, but the maths here is correct

claywar (approved):

Xaver-DaRed (approved):

PR #4454, opened 2023-09-06
[Nyzul] Centralize mob selection, spawn logic and information/data

Xaver-DaRed:
**_I affirm:_**

Part of Nyzul's rewrite.
This one deals (mostly) with the logic regarding generating a new floor. Focused on mobs. Lamps will come in a later PR, which will be much smaller and also delete the points.lua file in the nyzul zone folder.

Spawn point Table:
- Removed rotation entry, because its always the same.
- Defined entries as x, y and z.
- Renamed table.
- Made table local.

Other tables:
- Renamed tables.
- Made tables local.

Logic.
- Simplify logic using the the defined entries in the spawn point table and adapt to the removal of rotation entry.
- Renamed pickMobs function to prepareMobs. Made function global and moved to floor_generation, along a lamp logic function only used in said function. No changes were made to said function.
- Corrected function call to reflect change to global.
- Corrected table calls to reflect name change and global to local change.

Other:
- Added a respectable amount of comments.

Review by commit.


scripts/globals/nyzul/floor_generation.lua
```diff
@@ -0,0 +1,1245 @@
...
+        [46] = { x = 406.5, y = 0, z = -593.4 },
+        [47] = { x = 415.5, y = 0, z = -584.3 },
+        [48] = { x = 419.9, y = 0, z = -575.5 },
+        [49] = { x = 424.7, y = 0, z = -584.4 },
+        [50] = { x = 433.2, y = 0, z = -593.2 },
+        [51] = { x = 405.5, y = 0, z = -499.8 },
+    },
+
+    [2] =
+    {
```
TeoTwawki, line 70: is there anyway we can make this less of a sea of numbers? I know it was one before you got here, but I can't brain.

Xaver-DaRed: It's a table of positions based on several layouts.

I COULD enum the layouts, like SOUTH_WEST, NORTH, etc...
But the actual positions cant be helped until we have a system that allows to define a region and select valid positions from said region.

We do need this many entries (and in one of this layouts, we need even more than we have) to account for all the possible combinations of both lamps and mobs that can appear.

I did the count and if I remember correctly, just for mobs, the minimum number is 23

Xaver-DaRed: I did manage to remove one full column (rotation) from the positions, couse it was always `math.random(1, 255)`

TeoTwawki: maybe comment what the keys are rather than enuming directions. anything to break it up visually and break it down for the reader. like imagine the intem_mods table with zero commenting. that's what this feels like.

Xaver-DaRed: Sure, I can do that

claywar (approved):

PR #4455, opened 2023-09-06
Removes Dynamis items from Star Ruby Quadav

loxleygames:
**_I affirm:_**

This mob is located in Castle Zvahl Baileys [S] and I don't see any evidence that it should drop Dynamis items.
https://www.bg-wiki.com/ffxi/Star_Ruby_Quadav
https://ffxiclopedia.fandom.com/wiki/Star_Ruby_Quadav

N/A


TeoTwawki:
Likely matches the version in Beauduex [S] but till someone actually records a drop from it we should leave it dropping nothing.

Castle Zvahl Baileys [S] one has its own group ID, so that row should be drop ID zero as no source online has any listed drops for it at all. The one in Beauduex [S] has only Obsidian shard as its sole drop (at a rate that looks "common").

Mark drop ID 1949 "free to use" so someone fills in the gap next time a new drop ID needs created.

TeoTwawki:
brain farted for a sec, group != pool

TeoTwawki (approved):

PR #4461, opened 2023-09-07
[ci] Require lua comment headers consisting of only '-' to be 35 characters

claywar:
**_I affirm:_**

* Fixes some occurrences where the header line length was not 35 characters.
* Adds CI rules to test for the above

For Reference:  In scripts/ these lines are used 20002 times throughout the code.  484 were not 35 characters and 19518 were.

And LSB spake, saying, _"First shalt thou check for '-'.  Then shalt thou count to thirty-five, no more, no less.  Thirty-five shall be the number thou shalt count, and the number of the counting shall be thirty-five.  Thirty-six shalt thou not count, neither count thou thirty-four, excepting that thou then proceed to thirty-five.  Thirty-seven is right out.  If the number thirty-five, being the thirty-fifth number, is not reached, then lobbest thou thy error message towards thy contributor, who, being naughty in CI's sight, shall fix it."_


Xaver-DaRed (approved):

TeoTwawki (approved):
You had to fix more of them than I thought. I believed nearly all to already be 35.

Xaver-DaRed:
Guilty of this. I don't like it when my text goes past the header and bottom line

PR #4466, opened 2023-09-07
Remove errant logic in Check Kill Credit binding

### LINKED ISSUE #3621: 🐛 Eschan Il'aern are not counting for multiple kill RoE Objectives
**_I affirm:_**

Linux

## Branch affected by issue 

`base`, noticed on xiweb

## Steps to reproduce

Register the multiple kill Records of Eminence Objectives (ie: daily 20 or 30 kills, the repeatable 200 kill objective)
Go to Escha Ru'aun and start taking out mobs there.  Notice that the Zdei, Gargouilles, and Phuabo all register their kills towards the objectives, but the Il'aern do not.

## Expected behavior

All experience point weilding mobs in the zone register kills towards the mulitple kill objectives.

claywar:
**_I affirm:_**

Fixes #3621

Removes errant check in CLuaBaseEntity::checkKillCredit() which would fail if the player was more than 15 levels below the mob's level.

* Add RoE objectives for vanquish multiple enemies
* Set level to 99
* Kill Escha Il'Aern (level 116)
* Kill bunny in West Ronfaure

Aern will count for credit, bunny will not

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -10420,7 +10420,7 @@ uint8 CLuaBaseEntity::checkSoloPartyAlliance()
  *  Notes   : and the mob is able to give exp to the members
  ************************************************************************/
 
-bool CLuaBaseEntity::checkKillCredit(CLuaBaseEntity* PLuaBaseEntity, sol::object const& arg1, sol::object const& arg2)
+bool CLuaBaseEntity::checkKillCredit(CLuaBaseEntity* PLuaBaseEntity, sol::object const& minRange)
 {
     if (m_PBaseEntity->objtype != TYPE_PC || PLuaBaseEntity->GetBaseEntity()->objtype != TYPE_MOB)
     {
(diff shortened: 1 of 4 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -10420,23 +10420,20 @@ uint8 CLuaBaseEntity::checkSoloPartyAlliance()
...
         ShowWarning("CLuaBaseEntity::checkKillCredit() - Non-PC type calling function, or PLuaBaseEntity is not a MOB.");
         return false;
     }
 
-    CCharEntity* PChar = static_cast<CCharEntity*>(m_PBaseEntity);
-    CMobEntity*  PMob  = static_cast<CMobEntity*>(PLuaBaseEntity->GetBaseEntity());
-
-    bool  credit  = false;
-    int   lvlDiff = (int)(PMob->GetMLevel()) - (int)(PChar->GetMLevel());
-    int   maxDiff = (arg1 != sol::lua_nil) ? arg1.as<int>() : 15;
```
TeoTwawki, line 10436: if we're no longer baking in a default value, FoV/GoV global scripts will need a small update to the calling line to have same 15 level check they use now.

claywar: Is that confirmed that there is a 15 level hard cap of being over the mob?

TeoTwawki: I honestly can't say if thats "the line". its been there since forever and retained all the way to present day. I'd need to go dig in previous project iterations issues and pr's t see the exact argument for why it was added, or go load up retail and test. There def is some limit beyond simply "did you get exp" though, and removing that results in exploitable behaviors. I'd rather err on getting it imperfect than getting it exploitable.

TeoTwawki (approved):

PR #4472, opened 2023-09-09
[core] Party/Alliance zmq message audits

WinterSolstice8:
**_I affirm:_**

Fixes some crashes; lots of little changes to use party/alliance IDs more sanely with less crashing
Party/alliance message fixes

this code is "battle tested" to work on horizon, there's some subtle issues occasionally that I can't trace down because my brain is melted from the marathon debugging yesterday, but no more crashes = better

cross-process and not:
change party leader, see it update
change alliance leader, see it update
kick a player, they will be kicked and both the party and the player will see that correctly
kick a party, they will be kicked and both the alliance and party will see that correctly

### DIFF
--- a/src/common/mmo.h
+++ b/src/common/mmo.h
@@ -132,6 +132,7 @@ enum MSGSERVTYPE : uint8
     MSG_LOGIN,
     MSG_CHAT_TELL,
     MSG_CHAT_PARTY,
+    MSG_CHAT_ALLIANCE,
     MSG_CHAT_LINKSHELL,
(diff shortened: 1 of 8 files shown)


Frankie-hz:
Can confirm testing on horizon beta + live stopped all party related crashes + everything updates in every condition we have tested so far. Went from almost 40 crashes to 0 and no bug reports of any jank party/alliance issues so far. 

Note this is cross-clusters.

claywar (approved):

PR #4480, opened 2023-09-10
Fixes incorrect logic in FoV EXP threshold

loxleygames:
**_I affirm:_**

Original PR #4466 causing players to receive no EXP for level appropriate pages. It is the wrong way around. As an aside, I can't find any evidence for this feature and it doesn't seem retail accurate (For example, Cape Terrigan page 1 would award nothing at 75). The retail level limit on pages was related to the relevant mobs giving EXP to progress the page, I've never seen a cut off for the page reward itself, and this isn't documented on any of the usual resources. It would be helpful if this could be verified, to avoid future bug reports.

Current version:
```lua
player:getMainLvl() < math.max(1, page[5] - xi.settings.main.REGIME_REWARD_THRESHOLD)
```

Step through:
```lua
{ 7, 1, 0, 0, 15, 19, 475,  7 }, -- Valkurm Dunes Page 1

page[5] --15
xi.settings.main.REGIME_REWARD_THRESHOLD -- 15 (Default)
math.max(1, 15 - 15) -- 1

-- Award EXP if
player:getMainLvl() < 1
```

Complete level appropriate FoV and EXP should be awarded

### DIFF
--- a/scripts/globals/regimes.lua
+++ b/scripts/globals/regimes.lua
@@ -1494,9 +1494,9 @@ xi.regime.checkRegime = function(player, mob, regimeId, index, regimeType)
         player:setCharVar('[regime]lastReward', vanadielEpoch)
     end
 
-    -- award XP every page completion.  If the player is more than REGIME_REWARD_THRESHOLD below
-    -- the minimum suggested level, then do not award.
-    if player:getMainLvl() < math.max(1, page[5] - xi.settings.main.REGIME_REWARD_THRESHOLD) then
+    -- Award EXP for page completion
+    -- Player must be equal or greater than REGIME_REWARD_THRESHOLD levels below the minimum suggested level
+    if player:getMainLvl() >= math.max(1, page[5] - xi.settings.main.REGIME_REWARD_THRESHOLD) then
         player:addExp(reward * xi.settings.main.BOOK_EXP_RATE)
     end
 


claywar:
Can you explain the instance where they are not getting experience? This seems to be as intended.  If the player is more than 15 levels below the minimum level for the page, then do not award XP.

This was originally in checkKillCredit, but moved to the regime global iteself, and is intended to not allow a high level player to farm XP for someone much lower (for example, taking a lv96 page, and letting a lv1 leech).

I agree though, that retail behavior needs to be confirmed, and if so, removed all together in the future.

Edit: I see now, backwards logic

claywar (approved):

TeoTwawki:
> As an aside, I can't find any evidence for this feature and it doesn't seem retail accurate

I believe there was sort of cutoff but it wasn't properly documented - I think 15 was somebodies guess or only accurate under an older algo for exp and check (and even if we all were to accept this was a thing there would still be a question of if it is STILL a thing, but I def recall some situations where a just-barely EP gave exp but didn't count).
﻿
But its not being wholesale deleted for the forseeable future. As it is its configurable, so its "grandfathered in" with the hundreds of other imperfections from the old days..Been there since the original dsp implementation and carried forward for literally 5 iterations/refactors without anyone saying anything, lol.

PR #4492, opened 2023-09-13
[SQL][Items] Add item mods for items 23087-23091

helixhamin:
Finish some other TODOs and correct some wrong mods.

**_I affirm:_**

Adds mods to items 23087-23091.
It also adds some missing mods and corrects one wrong latent effect value.
Note: PFLUG mod ID exists, but the function does not appear to be implemented yet. TRUE_SHOT_EFFECT mod ID exists but the function does not appear to be implemented yet.

As GM !additem 10730|23087|23088|23089|23090|23091|23129|23330|23464|23665|25962|25963|27473|27474|28207
Equip item and make sure effects take effect (please see note above as not all mods actually work at the moment)

### DIFF
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -35,7 +35,11 @@ INSERT INTO `item_latents` VALUES (10679,11,10,10,0);
 INSERT INTO `item_latents` VALUES (10719,12,10,11,0);
 INSERT INTO `item_latents` VALUES (10719,13,10,11,0);
 INSERT INTO `item_latents` VALUES (10719,14,10,11,0);
-INSERT INTO `item_latents` VALUES (10730,63,10,13,56);  -- WAR AF2 +2 Feet Reduces defense penalty by 10% if Berserk Active
+
+-- -------------------------------------------------------
+-- Warrior's Calligae +2
+-- -------------------------------------------------------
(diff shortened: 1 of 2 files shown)


sql/item_latents.sql
```diff
@@ -35,7 +35,11 @@ INSERT INTO `item_latents` VALUES (10679,11,10,10,0);
 INSERT INTO `item_latents` VALUES (10719,12,10,11,0);
 INSERT INTO `item_latents` VALUES (10719,13,10,11,0);
 INSERT INTO `item_latents` VALUES (10719,14,10,11,0);
-INSERT INTO `item_latents` VALUES (10730,63,10,13,56);  -- WAR AF2 +2 Feet Reduces defense penalty by 10% if Berserk Active
+
+-- -------------------------------------------------------
```
claywar, line 39: Just looking at the rest, do we really need these two long lines of comments to create a header here? I'd say let's keep it simple to a single-line comment (and if the rest of the things in latents do that, nuke it from orbit.

claywar (approved):

sql/item_latents.sql
TeoTwawki: They do. let this pass, and later tomorrow I'll mass kill the dashed lines above and below the item name.

This is a case of past conversations everyone forgot about. In the dark days before tooling.

TeoTwawki: we'll start using same format `item_mods` now does after that

TeoTwawki: as promised: #4505

Issue #4493, opened 2023-09-13
🔨 New API (and performance improvements) for mixins


zach2good:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature

Before I went on holiday, we were talking about how mixins are one of the last holdouts for "things that are read off this disk and are repeatedly dumped into the global scope".

Even though I commented on these functions saying "it's OK to do these additional file reads at load-time", it'll be artificially inflating the startup time significantly:

```cpp
// Called during server startup, file reads are OK!
    int32 ApplyMixins(CBaseEntity* PMob)
    {
        TracyZoneScoped;

        if (PMob == nullptr || PMob->objtype != TYPE_MOB)
        {
            return -1;
        }

        // Clear out globals
        lua.set("mixins", sol::lua_nil);
        lua.set("mixinOptions", sol::lua_nil);

        auto zone_name = PMob->loc.zone->GetName();
        auto name      = PMob->GetName();

        auto filename = fmt::format("./scripts/zones/{}/mobs/{}.lua", zone_name, name);

        auto script_result = lua.safe_script_file(filename);
        if (!script_result.valid())
        {
            return -1;
        }

        // get the global function "applyMixins"
        sol::function applyMixins = lua["applyMixins"];
        if (!applyMixins.valid())
        {
            return -1;
        }

        // get the parameter "mixins"
        auto mixins = lua["mixins"];
        if (!mixins.valid())
        {
            return -1;
        }

        // get the parameter "mixinOptions" (optional)
        auto mixinOptions = lua["mixinOptions"];

        auto result = applyMixins(CLuaBaseEntity(PMob), mixins, mixinOptions);
        if (!result.valid())
        {
            sol::error err = result;
            ShowError("luautils::applyMixins: %s", err.what());
        }

        return 0;
    }

    // Called during server startup, file reads are OK!
    int32 ApplyZoneMixins(CBaseEntity* PMob)
    {
        TracyZoneScoped;

[...]

PR #4496, opened 2023-09-13
[core] Don't re-read mixins when applying them

zach2good:
**_I affirm:_**

Despite my rambling in https://github.com/LandSandBoat/server/issues/4493, with some careful planning its possible to keep the current API and script layout of mixins, and still get rid of the file reads. It does rely on us holding onto global state _between function calls_, which is bad and confusing juju. 

This is a good first step for performance, while leaving the door open to refactoring later on if we want it.

I'm not sure why the magian stuff has come in with this, I'll clean it up another time.

### DIFF
--- a/scripts/mixins/spawn_casket.lua
+++ b/scripts/mixins/spawn_casket.lua
@@ -2,12 +2,10 @@
 require('scripts/globals/caskets')
 require('scripts/globals/mixins')
 
-g_mixins = g_mixins or {}
-
 -----------------------------------
 -- Casket zone check
 -----------------------------------
-g_mixins.spawn_casket = function(casketMob)
+local spawn_casket = function(casketMob)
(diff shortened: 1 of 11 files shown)


zach2good:
Also, has ScopeGuard been removed? I can't seem to find it anymore

zach2good:
While the logic isn't quite there yet, it still shaves 3s off the startup time of the server: 
image

claywar:
Was removed in the command caching PR since that was the only place that was using it

zach2good:
WTB: Someone to smoke test this and make sure mixins (zone, family and mob) all still work as intended.

Xaver-DaRed:
I'll test this

zach2good:
Can't get boxes to spawn, I've goofed somewhere, probably in cleaning out the state between Mixins and ZoneMixins

zach2good:
This will be eventually replaced by https://github.com/LandSandBoat/server/pull/5378

PR #4418, opened 2023-09-02
Cache GM commands into xi.commands, remove global CI exceptions for command functions

claywar:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?
* Caches GM commands into xi.commands to no longer need to reload from disk every time one is executed
* Removes unused ScopeGuard class after changing command handler
* Removes global exceptions for cmdprops, onTrigger, fileExists, and error from CI

## Steps to test these changes
* Use a GM command in commands
* Use a GM command in modules
* Reload a GM command in modules
* Reload a GM command in commands
* Add a new GM command in commands
* Add a new GM command in modules

PR #4500, opened 2023-09-13
Trickattack rework

MowFord:
**_I affirm:_**

Trick Attack is, by many accounts, unreliable. In trying to dismantle how it works most people give up due to the complex nature of the code. Also Cover utilizes the same code. I've got to assume the many floats and trig functions carry over approximations that make it not always behave as expected.

This PR:
 - Simplifies the logic for determining if 3 entities are "in a line with a little wiggle room" for use in TA and Cover functions
 - Makes TA work by first collecting all potential TA targets (based on distance), then returning the closest to the mob that is positioned properly.

Note that there are 2 "magic numbers" in this new code (I'm unsure of how many the old code has, but there's at least one to to determine minimum and maximum slope):
 - Minimum distance from mob: .5 yalms
   - This is being careful not to get close to the `worldAngle()` defaulting to mob's rotation
 - Angle Diff threshold: 4 world angle rotation units in either direction (5 degress in either direction, rounded to the nearest rotation which really gives about 11.3 degress, cenetered on the angle between mob and TA user)

I originally brought this up https://github.com/LandSandBoat/server/discussions/3784 to see if there would be interest in making the change, but after utilizing it ourselves over at WingsXI for a bit over a week, i'm glad to say it's very reliable. 

The paradigm shift for this code is to approach the problem a different way. Instead of extrapolating various lines to build triangles around the mob, we just consider what it means to draw a line from mob to thf. 

The algorithm is best understood thinking of looking from the sky with a compass overlay on top. The line drawn could be simply represented as a compass direction. You can then also represent the potential ta target as another compass direction. The resulting angles can be subtracted to know how close to linear the three points are.

Included is a new function `!getta`, which reports the TA target that would be selected if the player running the command attacked the cursor target (Note that this is utilizing `getTrickAttackChar()` directly, so the player running the command does need Trick Attack buff active).

Simply join parties with another player and try various positions in front of and behind while using the `!getta` function. Gifs included of this testing:

Before changes: https://imgur.com/QEO1kJb

After changes: https://imgur.com/Nxvg6Hi

### DIFF
--- a/scripts/commands/getta.lua
+++ b/scripts/commands/getta.lua
@@ -2,18 +2,16 @@
 -- func: getta
 -- desc: returns the name of the entity that would be chosen for trick attack given the current (mob) target
 -----------------------------------
-cmdprops =
+local commandObj = {}
+
+commandObj.cmdprops =
 {
-    permission = 3,
+    permission = 1,
     parameters = ''
 }
 
-function error(player, msg)
-    player:PrintToPlayer(msg)
-    player:PrintToPlayer('!getta with a target')
-end
-
-function onTrigger(player)
+-- function onTrigger(player, extendedMode)
+commandObj.onTrigger = function(player)
     local targ = player:getCursorTarget()
     if targ ~= nil then
         local tatarget = player:getTrickAttackChar(targ)
@@ -26,3 +24,5 @@ function onTrigger(player)
         player:PrintToPlayer('Must select a target using in game cursor first.')
     end
 end
+
+return commandObj
--- a/scripts/commands/getta.lua
+++ b/scripts/commands/getta.lua
@@ -1,0 +1,28 @@
+-----------------------------------
+-- func: getta
+-- desc: returns the name of the entity that would be chosen for trick attack given the current (mob) target
+-----------------------------------
+cmdprops =
+{
+    permission = 3,
+    parameters = ""
+}
+
+function error(player, msg)
+    player:PrintToPlayer(msg)
+    player:PrintToPlayer("!getta with a target")
+end
+
+function onTrigger(player)
+    local targ = player:getCursorTarget()
+    if targ ~= nil then
+        local tatarget = player:getTrickAttackChar(targ)
+        if tatarget ~= nil then
+            player:PrintToPlayer(string.format("Trick attack would select: %s", tatarget:getName()))
+        else
+            player:PrintToPlayer("No valid TA target found.")
+        end
+    else
+        player:PrintToPlayer("Must select a target using in game cursor first.")
+    end
+end
--- a/scripts/commands/getta.lua
+++ b/scripts/commands/getta.lua
@@ -5,12 +5,12 @@
 cmdprops =
 {
     permission = 3,
-    parameters = ""
+    parameters = ''
 }
 
 function error(player, msg)
     player:PrintToPlayer(msg)
-    player:PrintToPlayer("!getta with a target")
+    player:PrintToPlayer('!getta with a target')
 end
 
 function onTrigger(player)
@@ -18,11 +18,11 @@ function onTrigger(player)
     if targ ~= nil then
         local tatarget = player:getTrickAttackChar(targ)
         if tatarget ~= nil then
-            player:PrintToPlayer(string.format("Trick attack would select: %s", tatarget:getName()))
+            player:PrintToPlayer(string.format('Trick attack would select: %s', tatarget:getName()))
         else
-            player:PrintToPlayer("No valid TA target found.")
+            player:PrintToPlayer('No valid TA target found.')
         end
     else
-        player:PrintToPlayer("Must select a target using in game cursor first.")
+        player:PrintToPlayer('Must select a target using in game cursor first.')
     end
 end
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -103,9 +103,8 @@ namespace battleutils
      *                                                                       *
      ************************************************************************/
 
-    const float worldAngleMinDistance                = 0.5f;
-    const float worldAngleMinDistanceSquared         = worldAngleMinDistance * worldAngleMinDistance;
-    const uint8 worldAngleMaxDeviance                = 4;
+    const float worldAngleMinDistance = 0.5f;
+    const uint8 worldAngleMaxDeviance = 4;
 
     void LoadSkillTable()
     {
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -4255,8 +4255,9 @@ namespace battleutils
     }
 
     /*
-        pass worldAngle(anchorMob, firstPlayer) instead of calculating everytime to allow TA to be more efficient
-        Calculates world angle between otherPlayer and anchor mob,
+        pass result of worldAngle(anchorEntity, firstEntity) instead of calculating everytime to allow TA to be more efficient
+        Note the order of worldAngle calls, the anchor must be first in all comparisons
+        Calculates world angle between other entity and anchor entity,
         then determines if the difference of those angles is within acceptable range for moves that require the three to be "in a straight line"
         Used for Trick Attack and Cover: separate checks for distance/party membership must be done to confirm eligability
     */
@@ -4305,6 +4306,7 @@ namespace battleutils
                 taPartyList.emplace_back(taUser->PParty);
             }
 
+            // Collect all potential TA targets who are closer to the mob than the TA user
             for (auto&& party : taPartyList)
             {
                 for (auto&& member : party->members)
@@ -6960,6 +6962,7 @@ namespace battleutils
                 uint8 angleTAmob = worldAngle(PMob->loc.p, PCoverAbilityUser->loc.p);
                 float distTAmob  = distanceSquared(PCoverAbilityUser->loc.p, PMob->loc.p);
 
+                // check if cover user is within melee range and that cover target is in-line behind
                 if (distTAmob <= static_cast<float>(PMob->GetMeleeRange() * PMob->GetMeleeRange()) && // make sure cover user is within melee range
                     distTAmob >= worldAngleMinDistanceSquared &&                                      // require closer target not be closer than .5 yalms (.5*.5=.25 distsquared) to mob
                     distTAmob < distanceSquared(PCoverAbilityTarget->loc.p, PMob->loc.p) &&           // make sure cover user is closer to the mob than cover target
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -4251,114 +4251,45 @@ namespace battleutils
     }
 
     /*
-     * Find if any party members are in position for trick attack.  Do this by making a narrow triangle:
-     *  one endpoint at the mob, and the other two endpoints being equidistant from the TA user, perpendicular to
-     *  the line between the mob and the TA user.  Find the slope of the line between the TA user and the mob, and
-     *  decide whether to use x or z as the dependent variable (to avoid big numbers and divide by 0 errors on
-     *  vertical slopes).  Using this slope, we can find the angle of the perpendicular line to the x or z line
-     *  (depending on what the dependent var is), and using that angle, the disassembled x and z components to that
-     *  line.  Divide those by 2 for a half yalm length line for each side of the base of the triangle, and we get
-     *  the min and max values for x/z around the TA user.  Now it's simply a matter of first: making sure the
-     *  TA target is closer than the TA user, and that the TA targets x and z coordinates fall within the triangle
-     *  we made.  Using the min/max points and the mobs coordinate, we can construct min and max slopes, check that
-     *  the x or z coordinates are between the mob and player, and finally calculate where the z coordinate should
-     *  be based on the users x coordinate (or the other way around in a z dependent scenario) and check if the
-     *  actual z coordinate is between those two values.
+        pass worldAngle(anchorMob, firstPlayer) instead of calculating everytime to allow TA to be more efficient
+        Calculates world angle between otherPlayer and anchor mob,
+        then determines if the difference of those angles is within acceptable range for moves that require the three to be "in a straight line"
+        Used for Trick Attack and Cover: separate checks for distance/party membership must be done to confirm eligability
+    */
+    inline bool areInLine(uint8 firstPlayerWA, CBattleEntity* anchorMob, CBattleEntity* otherPlayer)
+    {
+        // X-degree angle threshold, centered on the firstPlayer's world angle
+        // X = 10 => 10/2 * 255 / 360 ~ 3.5 rotation diff, rounded to 4 which really gives X~11.3
+        int16 angleDiff = angleDifference(firstPlayerWA, worldAngle(anchorMob->loc.p, otherPlayer->loc.p));
+        ShowDebug("InLine check angleDiff: %d\n",angleDiff);
+        if (abs(angleDiff) <= 4)
+        {
+            return true;
+        }
+        return false;
+    }
+
+
+    /*
+     *  Find if any party members are in position for trick attack.  Do this by comparing the world angle between:
+     *  1. the TA user and the TA target
+     *  2. the TA user and the Mob
+     *  First, build a list of players that are closer to the mob than the TA user,
+     *   then sort by distance and choose the first that succeeds in meeting the criteria of areInline function
      */
 
     CBattleEntity* getAvailableTrickAttackChar(CBattleEntity* taUser, CBattleEntity* PMob)
     {
+        TracyZoneScoped;
         if (!taUser->StatusEffectContainer->HasStatusEffect(EFFECT_TRICK_ATTACK))
         {
             return nullptr;
         }
 
-        float taUserX = taUser->loc.p.x;
-        float taUserZ = taUser->loc.p.z;
-        float mobX    = PMob->loc.p.x;
-        float mobZ    = PMob->loc.p.z;
-
-        float xdif       = taUserX - mobX;
-        float zdif       = taUserZ - mobZ;
-        float slope      = 0;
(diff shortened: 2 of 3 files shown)


MowFord:
Oh the irony of failing CI checks when i copied the `!getmod` function. commit cincoming

scripts/commands/getta.lua
```diff
@@ -0,0 +1,28 @@
+-----------------------------------
```
claywar, line 1: This GM command will not function.  Needs to be updated to used the caching format

MowFord: it definitely functions on my local, which command can i use as an example of the caching format?

claywar: All of them

claywar: See #4418

MowFord: Wow. Apologies. No idea why my local has the old format. I'm sure it's due to not doing a clean rebase pull when switching from ASB and LSB

Xaver-DaRed (approved):
Not too convinced about the gm command name. But thats just a nitpick.

This has been working on wings for some time, I take it?

src/map/utils/battleutils.cpp
```diff
@@ -4385,18 +4325,63 @@ namespace battleutils
...
                 }
             }
         }
+        */
+
+        if (taTargetList.size() > 0)
+        {
+            // sorts by distance then by pointer id (only if floats are equal)
+            sort(taTargetList.begin(), taTargetList.end());
+            for (std::pair<float, CBattleEntity*> const& potentialTAtarget : taTargetList)
```
zach2good: I forget what this is called, but you can decompose this to be a bit cleaner:
```cpp
int main()
{
    auto pairVec = { std::make_pair(10, 42) };
    for (auto const& [id, ptrWhatever] : pairVec)
    {
        // 
    }
    return 0;
}
```

src/map/utils/battleutils.cpp
```diff
@@ -4385,18 +4325,63 @@ namespace battleutils
...
                     }
                 }
             }
         }
+        */
+
+        if (taTargetList.size() > 0)
+        {
+            // sorts by distance then by pointer id (only if floats are equal)
+            sort(taTargetList.begin(), taTargetList.end());
```
zach2good: Please use the std namespace here: `std::sort`

src/map/utils/battleutils.cpp
```diff
@@ -4385,18 +4325,63 @@ namespace battleutils
...
+            for (std::pair<float, CBattleEntity*> const& potentialTAtarget : taTargetList)
+            {
+                if (taUser->id == potentialTAtarget.second->id || // can't TA self
+                    potentialTAtarget.second->isDead())           // Dead entity should not be TA-able
+                {
+                    continue;
+                }
+                if (areInLine(angleTAmob, PMob, potentialTAtarget.second))
+                    return potentialTAtarget.second;
```
zach2good: This needs to be wrapped in {}'s

src/map/utils/battleutils.cpp
```diff
@@ -4385,18 +4325,63 @@ namespace battleutils
...
+        {
+            // sorts by distance then by pointer id (only if floats are equal)
+            sort(taTargetList.begin(), taTargetList.end());
+            for (std::pair<float, CBattleEntity*> const& potentialTAtarget : taTargetList)
+            {
+                if (taUser->id == potentialTAtarget.second->id || // can't TA self
+                    potentialTAtarget.second->isDead())           // Dead entity should not be TA-able
+                {
+                    continue;
+                }
```
zach2good, line 4365: Need a linebreak here

src/map/utils/battleutils.cpp
```diff
@@ -4385,18 +4325,63 @@ namespace battleutils
...
                     {
-                        return potentialTrust;
+                        taTargetList.emplace_back(distTAtarget, fellow);
                     }
                 }
             }
         }
+        */
+
+        if (taTargetList.size() > 0)
```
zach2good: `!taTargetList.empty()`

src/map/utils/battleutils.cpp
```diff
@@ -4251,114 +4251,44 @@ namespace battleutils
...
+        Calculates world angle between otherPlayer and anchor mob,
+        then determines if the difference of those angles is within acceptable range for moves that require the three to be "in a straight line"
+        Used for Trick Attack and Cover: separate checks for distance/party membership must be done to confirm eligability
+    */
+    inline bool areInLine(uint8 firstPlayerWA, CBattleEntity* anchorMob, CBattleEntity* otherPlayer)
+    {
+        // X-degree angle threshold, centered on the firstPlayer's world angle
+        // X = 10 => 10/2 * 255 / 360 ~ 3.5 rotation diff, rounded to 4 which really gives X~11.3
+        int16 angleDiff = angleDifference(firstPlayerWA, worldAngle(anchorMob->loc.p, otherPlayer->loc.p));
+        ShowDebug("InLine check angleDiff: %d\n", angleDiff);
```
zach2good: I don't know how often this is called, but if it's called _a lot_ this will spam? 

Newline is a part of logging, no need to add your own 👍

src/map/utils/battleutils.cpp
```diff
@@ -6984,73 +6969,16 @@ namespace battleutils
...
-                    zDependent = false;
-                }
-                else
-                {
-                    slope = zdif / xdif;
-
-                    float angle = (float)atan((double)1) * 2 - atan(slope);
+                // using same variable names as trick attack function, for consistent understanding
+                uint8 angleTAmob = worldAngle(PMob->loc.p, PCoverAbilityUser->loc.p);
+                float distTAmob  = distanceSquared(PCoverAbilityUser->loc.p, PMob->loc.p);
```
zach2good: I see a bunch of `distanceSquared` and `sqrt`, whats the reason for this, converting things to absolute values? In which case I'd say use `std::abs`

src/map/utils/battleutils.cpp
```diff
@@ -4251,114 +4251,44 @@ namespace battleutils
...
-     *  we made.  Using the min/max points and the mobs coordinate, we can construct min and max slopes, check that
-     *  the x or z coordinates are between the mob and player, and finally calculate where the z coordinate should
-     *  be based on the users x coordinate (or the other way around in a z dependent scenario) and check if the
-     *  actual z coordinate is between those two values.
+        pass worldAngle(anchorMob, firstPlayer) instead of calculating everytime to allow TA to be more efficient
+        Calculates world angle between otherPlayer and anchor mob,
+        then determines if the difference of those angles is within acceptable range for moves that require the three to be "in a straight line"
+        Used for Trick Attack and Cover: separate checks for distance/party membership must be done to confirm eligability
+    */
+    inline bool areInLine(uint8 firstPlayerWA, CBattleEntity* anchorMob, CBattleEntity* otherPlayer)
```
zach2good: I'd rename `firstPlayerWA` to `firstEntityWorldAngle` - since we're specifically dealing with `CBattleEntity`'s and not mobs/chars

src/map/utils/battleutils.cpp
```diff
@@ -4251,114 +4251,44 @@ namespace battleutils
...
+    {
+        // X-degree angle threshold, centered on the firstPlayer's world angle
+        // X = 10 => 10/2 * 255 / 360 ~ 3.5 rotation diff, rounded to 4 which really gives X~11.3
+        int16 angleDiff = angleDifference(firstPlayerWA, worldAngle(anchorMob->loc.p, otherPlayer->loc.p));
+        ShowDebug("InLine check angleDiff: %d\n", angleDiff);
+        if (abs(angleDiff) <= 4)
```
zach2good: Please use `std:abs`

src/map/utils/battleutils.cpp
```diff
@@ -4251,114 +4251,44 @@ namespace battleutils
...
+        {
+            return true;
+        }
+        return false;
```
zach2good: `return abs(angleDiff) <= 4;`, etc.

src/map/utils/battleutils.cpp
```diff
@@ -4368,15 +4298,25 @@ namespace battleutils
                 {
                     for (std::size_t i = 0; i < a->members.size(); ++i)
```
zach2good: Please use range-based-for here, unless you have a very solid reason for needing an index into a->members

zach2good: I can't comment on the line that has it, but while you're in here and understand what's going on, can you rename the `a` in `for (auto& a : taUser->PParty->m_PAlliance->partyList)` to something a bit more descriptive? Thanks!

src/map/utils/battleutils.cpp
```diff
@@ -6984,73 +6969,16 @@ namespace battleutils
...
-                    float minZpoint = mobZ - zoffset;
-
-                    maxSlope = (maxZpoint - coverAbilityTargetZ) / (maxXpoint - coverAbilityTargetX);
-                    minSlope = (minZpoint - coverAbilityTargetZ) / (minXpoint - coverAbilityTargetX);
-                }
-
-                if (distance(PCoverAbilityUser->loc.p, PMob->loc.p) <= (float)PMob->GetMeleeRange() && // make sure cover user is within melee range
-                    distance(PCoverAbilityUser->loc.p, PMob->loc.p) <=
-                        distance(PCoverAbilityTarget->loc.p, PMob->loc.p)) // make sure cover user is closer to the mob than cover target
+                if (sqrt(distTAmob) <= (float)PMob->GetMeleeRange() &&                      // make sure cover user is within melee range
```
zach2good: If you need to be using sqrt, please use `std::sqrt`

zach2good: Please avoid using C-style casts, use a `static_cast` here

zach2good (commented):
Looks a lot less fiddly than we have at the moment, thanks!

src/map/utils/battleutils.cpp
MowFord: Suppose I could also use melee range squared

src/map/utils/battleutils.cpp
MowFord: I'll take a look, thanks

src/map/utils/battleutils.cpp
MowFord: Haha sure. I'll handle the rest of these comments and test it again. Thanks for the feedback!

MowFord:
> Not too convinced about the gm command name. But thats just a nitpick.
> 
> This has been working on wings for some time, I take it?

Yea couple weeks at this point. 

Suggestion for a better command name?

src/map/utils/battleutils.cpp
MowFord: Haven't tested it but while I'm in here how do you feel about compressing this into one loop:

```
            std::vector<CParty*> taPartyList;
            if (taUser->PParty->m_PAlliance != nullptr)
            {
                taPartyList = taUser->PParty->m_PAlliance->partyList;
            }
            else
            {
                taPartyList.emplace_back(taUser->PParty);
            }
            for (auto&& party : taPartyList)
            {
                for (auto&& member : party->members)
                {
                        ...
```

src/map/utils/battleutils.cpp
MowFord: I tend to use `distanceSquared` if i just need to compare two distances. I only used sqrt once to match the style that cover originally had (but i've since changed that to `distTAmob <= static_cast<float>(PMob->GetMeleeRange() * PMob->GetMeleeRange())`)

src/map/utils/battleutils.cpp
MowFord: this is called for every partymember/trust closer to the mob when a thf attacks with trick attack buff. It was mostly for confirmation of the system. It's not helpful at all now that the logic is confirmed to be reliable.

MowFord:
I believe I've addressed every improvement listed, we'll see what CI thinks.

src/map/utils/battleutils.cpp
zach2good: Putting performance considerations to the side for a moment (`distance` is just `sqrt(distanceSquared(...)`, and `sqrt` hasn't been an expensive operation since the 90's):

The issue I have with `distanceSquared` is that it abstracts measurements in yalms into yalms^2, and it introduces the need to continually multiply out the number you're comparing to. We have a `constexpr square(...)` function, but if the numbers being brought in aren't known at compiletime you lose the benefit of the `constexpr`.

At the cost of a couple more `sqrt` calls you can use `distance` and you wouldn't need to multiply `PMob->GetMeleeRange()` with itself and the code would be more intuitive. Since this is a complex and mathsy area of code, I'd lean towards readability over pure performance (especially when there's a sea of stupid things we do elsewhere that make optimisation of float comparisons irrelevant).

src/map/utils/battleutils.cpp
```diff
@@ -4251,149 +4251,116 @@ namespace battleutils
...
-     *  we made.  Using the min/max points and the mobs coordinate, we can construct min and max slopes, check that
-     *  the x or z coordinates are between the mob and player, and finally calculate where the z coordinate should
-     *  be based on the users x coordinate (or the other way around in a z dependent scenario) and check if the
-     *  actual z coordinate is between those two values.
+        pass worldAngle(anchorMob, firstPlayer) instead of calculating everytime to allow TA to be more efficient
+        Calculates world angle between otherPlayer and anchor mob,
+        then determines if the difference of those angles is within acceptable range for moves that require the three to be "in a straight line"
+        Used for Trick Attack and Cover: separate checks for distance/party membership must be done to confirm eligability
+    */
+    inline bool areInLine(uint8 firstEntityWorldAngle, CBattleEntity* anchorMob, CBattleEntity* otherEntity)
```
zach2good: anchorMob -> anchorEntity

src/map/utils/battleutils.cpp
```diff
@@ -4251,149 +4251,116 @@ namespace battleutils
...
+     *  Find if any party members are in position for trick attack.  Do this by comparing the world angle between:
+     *  1. the TA user and the TA target
+     *  2. the TA user and the Mob
+     *  First, build a list of players that are closer to the mob than the TA user,
+     *   then sort by distance and choose the first that succeeds in meeting the criteria of areInline function
      */
 
     CBattleEntity* getAvailableTrickAttackChar(CBattleEntity* taUser, CBattleEntity* PMob)
     {
+        TracyZoneScoped;
```
zach2good, line 4285: If you added this tag, did you happen to do any profiling to see what this costs as before/after?

src/map/utils/battleutils.cpp
```diff
@@ -4251,149 +4251,116 @@ namespace battleutils
...
-                float memberXdif = PEntity->loc.p.x - mobX;
-                float memberZdif = PEntity->loc.p.z - mobZ;
-                if (zDependent)
+                taPartyList = taUser->PParty->m_PAlliance->partyList;
+            }
+            else
+            {
+                taPartyList.emplace_back(taUser->PParty);
+            }
+            for (auto&& party : taPartyList)
```
zach2good, line 4309: Needs linebreak here

src/map/utils/battleutils.cpp
zach2good: Sure, as long as it's tested and it works, and doesn't introduce any fuckery

src/map/utils/battleutils.cpp
```diff
@@ -6984,73 +6951,16 @@ namespace battleutils
...
+                float distTAmob  = distanceSquared(PCoverAbilityUser->loc.p, PMob->loc.p);
 
-                    maxSlope = ((maxXpoint - coverAbilityTargetX) / (maxZpoint - coverAbilityTargetZ));
-                    minSlope = ((minXpoint - coverAbilityTargetX) / (minZpoint - coverAbilityTargetZ));
-
-                    zDependent = false;
-                }
-                else
+                if (distTAmob <= static_cast<float>(PMob->GetMeleeRange() * PMob->GetMeleeRange()) && // make sure cover user is within melee range
+                    distTAmob >= 0.25f &&                                                             // require closer target not be closer than .5 yalms (.5*.5=.25 distsquared) to mob
```
zach2good: There's a couple of magic numbers floating (hah) around here, can you extract them into an anonymous namespace at the top of the file please? Here's an example from the top of zone_entities.cpp:
```cpp
namespace
{
    const float CHARACTER_SYNC_DISTANCE                = 45.0f;
    const float CHARACTER_DESPAWN_DISTANCE             = 50.0f;
    const int   CHARACTER_SWAP_MAX                     = 5;
    const int   CHARACTER_SYNC_LIMIT_MAX               = 32;
    const int   CHARACTER_SYNC_DISTANCE_SWAP_THRESHOLD = 30;
    const int   CHARACTER_SYNC_PARTY_SIGNIFICANCE      = 100000;
    const int   CHARACTER_SYNC_ALLI_SIGNIFICANCE       = 10000;
    const int   PERSIST_CHECK_CHARACTERS               = 20;
} // namespace
```

src/map/utils/battleutils.cpp
MowFord: I didn't, actually. I added that a while back when I first took over wings just to get more granular stats when i was playing whack-a-mole with performance issues

src/map/utils/battleutils.cpp
MowFord: Cool idea. I'll take a look

src/map/utils/battleutils.cpp
MowFord: Understood, I'll update to use distance everywhere

MowFord:
Appreciate all the feedback. Sorry to go one step forward and two steps back each commit.

I think we're in a solid place now with the remarks (I'm actually struggling to view them in a sane way, but I think i've addressed them all).

I've compiled and re-tested on my local after this last commit

zach2good (approved):
Looks good 👍 

(remember to squash, whoever is merging)

PR #4502, opened 2023-09-14
Update terpsichore acc value

cmastro24:
**_I affirm:_**

This fixes a small problem with Terpischore where the step accuracy was off by 10.  I updated the value to 20 which is what BG/JP Wiki show as the correct value
https://www.bg-wiki.com/ffxi/Terpsichore_(Level_75)

Obtain a Terpischore and look to see if the step accuracy correctly reflects what is in BG Wiki and JP Wiki.  Validate sql file has correct value of 20 for level 75


cmastro24:
Investigating some more

cmastro24:
Validated change only affects 75 version of the weapon.

claywar (approved):

PR #4504, opened 2023-09-14
[core] Add equality operator to Lua classes

zach2good:
**_I affirm:_**

Adding the equality operator to the Lua classes makes it possible to compare them directly while in Lua, and it will be true if they are the same underlying C++ entity, even though they are separate CLuaXYZ objects.

```lua
local mob1 = GetMobByID(mobId)
local mob2 = GetMobByID(mobId)
if mob1 == mob2 then
  -- This will succeed after this PR, even though they are separate Lua objects
end
```

### DIFF
--- a/src/map/lua/lua_ability.h
+++ b/src/map/lua/lua_ability.h
@@ -59,6 +59,11 @@ public:
     void   setVE(uint16 ve);
     void   setRange(float range);
 
+    bool operator==(const CLuaAbility& other) const
+    {
+        return this->m_PLuaAbility == other.m_PLuaAbility;
+    }
+
     static void Register();
 };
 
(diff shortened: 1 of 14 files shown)


claywar (approved):

TeoTwawki:
I almost missed that they were the same mob ID in those local variables and thought "wait what?" before it hit me.

Just to clarify for future readings:
This will **only** pass when its exactly the same, right? Not any 2 of same kind but _exactly_ the same one.

TeoTwawki (approved):

TeoTwawki:
p.s.
in general, this is just one of several reasons doing entity object compares has been historically a bad idea as opposed to ID compares, but less so than it used to be

zach2good:
This is comparing the underlying pointer, so only if it's the same area of memory ie. the same object

zach2good:
CI is currently broken, but everything else is passing, so gonna merge (without the commits where I was tinkering with CI)

PR #4511, opened 2023-09-16
Add missing augmentable item logic to Tenshodo Treasure Coffer

MowFord:
**_I affirm:_**

First things first, there's a non-retail setting included as it's a simple change to support that is likely relevant for many people: `NEXUS_CAPE_FIN`. Since the expansions aren't purchased, simply enabled by the server, this setting ties nexus cape to completion of all 3 mini expansions.

Tenshodo Treasure Coffer had basic structure but didn't support scenario rewards as well as had many missing augmented items with incorrect/missing augments. The augments were meticulously audited to reflect ffxiclopedia. Also the structure was updated to be easier to read/confirm things later. Example of structure

image

Scenario rewards weren't coded at all, so additional option handling in event update/event complete. 

I've spot checked that each key item rewards an item as well as spot checked augment mods apply as expected

image

image

etc

Finally, note that `scenarioArmor` is passed a bool whether or not to give the item. This allows the menu to preview the augments

Test the `NEXUS_CAPE_FIN` variable by
 - setting to 0 and confirming the box gives nexus cape menu only if all 3 scenarios are enabled
 - setting to 1 and confirming the box gives nexus cape menu only if all 3 "fin" missions are added for player:
   - `!addmission 9 11`, `!addmission 10 14`, `!addmission 11 14`

Test the key item trades by simply giving the key item and interacting with the chest (mission progress isn't needed to receive a key-item reward)


MowFord:
Would love some guidance on the table structure. Also not sure what "Incorrectly defined table" means. I'll hold off on fixing the quote issues from sanity checks until i get feedback on the lua structure. It's a massive dataset so i'd hate to do it wrong and spin my wheels.

claywar:
So the two things that I'm seeing common are (from the style guide: https://github.com/LandSandBoat/server/wiki/Development-Guide#lua):

```
-- Correct ✔️ 
local table =
{
    content = 1,
}

-- Wrong ❌
local table = {
    content = 1,
}
```

And one that's not in there which I will fix, but a requirement for padding for same line tables (example below)

```
-- Correct ✔️ 
local table = { content = 1 }

-- Wrong ❌
local table = {content = 1}
```

claywar:
While you're in there, CI isn't going to fail this, but would prefer that this is broken out to multiple lines:
```
expansion = xi.mission.log_id.AMK, mission = xi.mission.id.amk.WELCOME_TO_MY_DECREPIT_DOMICILE, repeatable = true,
```

MowFord:
Did a quick pass to make sure what i'm doing is right. I'll break out each key item metadata, too

MowFord:
confirmed code still works after styling fixes

TeoTwawki:
> First things first, there's a non-retail setting included 

Module would be preferred if possible. If not keep the setting but an issue would be appreciated citing this as an example of something that the system can't handle.

> Also the structure was updated to be easier to read/confirm things later. Example of structure

Placing the IDs into the item enum and using the enum here would be preferred over using the raw item ID. We had no such enum back in the day, but since you are reformatting this for the 2nd (3rd?) time in its life, now is a good chance to right that wrong.

---

Just some fyi/infos


> The augments were meticulously audited to reflect ffxiclopedia.

be aware that:
1. those wiki pages are living documents updated by people as they attain ones or sometimes as they remember and feel like editing, without much in the way of being able to prove or disprove a given combination, and the script was made way long ago with assuredly some editing happening after. (build it and they will come!...Ok built it! hey where is everyone?) 
2. I did most of the original values that were present for the crystalline prophecy ones. I may not have added my own augs to wiki but I had screenshots of my own in game items as well as whatever wiki had the day it went into dsp. some people after me may not have realized augment power values are the number in code+1 (really wish I had changed that core side before we used it. would be panful to shift now)
3. at least a few of the things wiki says are wrong. _Especially_ 'clopedia. But something is better than nothing. Especially if we can get multiple sources and be reasonably sure they weren't just copying each other.

> Scenario rewards weren't coded at all, so additional option handling in event update/event complete. 

I didn't have a capture of them to work with :( and the missions weren't finished anyway, so I left'em out. Glad to see somebody fill it in finally.

----

_p.s. boy later we need some bit logic love in here to kill all those value ranges and adding of params. I still feel bad about wren (I think it was wren) having cleaned up my crappy if/else and about not using bit logic back then. Even now those params feel intimidating tho so... not it._

MowFord:
> Module would be preferred if possible

Are you saying a module to change the behavior of that one logical check? Seems like overkill. Alternative are you saying is just to make an issue so it's easily searchable?

> Placing the IDs into the item enum

I had not considered using the item enums, good call! I'll see about getting that updated.

> be aware that...

For what it's worth a lot of the augments are unchanged, the audit simply broke them out and matched them explicitly to what was listed in the wiki. I'm not sure how better we can do this other than start here and have people contribute changes with evidence in the issue/PR

> ...some bit logic love...

Explain what you mean and I don't mind taking a crack at it. I'm very well acquainted with this code as I've refactored it and squintrox very recently on wings (squintrox is my next PR)

TeoTwawki:
> Are you saying a module to change the behavior of that one logical check? Seems like overkill. 

Honestly so is the setting.

The issue comment was "when we run into things the module system is incapable of, document it or it never will be"

TeoTwawki:
> For what it's worth a lot of the augments are unchanged, the audit simply broke them out and matched them explicitly to what was listed in the wiki. I'm not sure how better we can do this other than start here and have people contribute changes with evidence in the issue/PR

sometimes I comment purely to document why things were what they were.

claywar:
I agree as well that a setting here is probably not the right path, especially nested between enabling expansions.  This is a very small change, and is most likely better enabled or disabled within a module.

TeoTwawki:
> Explain what you mean and I don't mind taking a crack at it. I'm very well acquainted with this code as I've refactored it and squintrox very recently on wings (squintrox is my next PR)

I intend for it to be its own thing at a later date, but a lot of the values of the cs option code are actually bits in masks and we really should not be doing `if option < 345678 and option > 4567654 then`

zach2good:
+1 for the module to add non-retail behaviour, please. We have a bunch of settings to enable/disable this sort of thing, but they were implemented long before we had modules. These days we have the tools to allow people to run wild with obvious QOL, enhancements, timesavers, or custom shenanigans without muddying the waters of "this is a retail FFXI emulator"

MowFord:
> +1 for the module to add non-retail behaviour, please. We have a bunch of settings to enable/disable this sort of thing, but they were implemented long before we had modules. These days we have the tools to allow people to run wild with obvious QOL, enhancements, timesavers, or custom shenanigans without muddying the waters of "this is a retail FFXI emulator"

I suppose the suggestion related to this PR is to revert the logic and anyone who wants to do this different logic should just make it a module (which i assume is trival to just change the logic of setting `eligableNexusCape` within the `onTrigger`)? Or are you saying to add the module to this PR. Just asking for clarification

MowFord:
> > Explain what you mean and I don't mind taking a crack at it. I'm very well acquainted with this code as I've refactored it and squintrox very recently on wings (squintrox is my next PR)
> 
> I intend for it to be its own thing at a later date, but a lot of the values of the cs option code are actually bits in masks and we really should not be doing `if option < 345678 and option > 4567654 then`

ah fair enough... I'll leave it for a later date then :)

MowFord:
utilizing a script to pull item ids from the db, create a friendly name, then sorting them into item.lua, I've updated item.lua and the magic numbers related to `itemId`. little powershell to pull the item ids here for posterity:

```
Get-Content .\scripts\zones\Lower_Jeuno\npcs\Treasure_Coffer.lua  | %{if($_ -match '(.+itemId[ ]+=[ ]+)(\d+)( )'){$Matches.2}}
```

Working on cleaning it all up to commit again, but wanted to just address that i'm going to revert the setting of `eligableNexusCape` to what it was before this PR and i'll open an issue to discuss the best way to create a framework for modifying slight logic within a function via module without replacing the whole function.

TeoTwawki:
> Or are you saying to add the module to this PR

I wouldn't turn down a module included. Its ok if its small. 

The main purpose of module system is to keep non-retail customization out of the main code to allow downstream to do more without touching the main files directly. Some of our settings have been slowly migrating that way, more of them will over time. The more complicated things that were once retail but changed will likely stay as settings while simpler ones or totally non retail things will move to modules.

MowFord:
Even though it may uh increase the work here, I wanted to point out items like this can be optimized with my new code:

https://ffxiclopedia.fandom.com/wiki/Varius_Torque

It would be very surprising if it could give two different combat skill augments (research needed ofc), and if that's true then instead of listing all the combat skill augments in the same tier, would could make them a sub-list. I've accounted for that format to account for stats that can range positive or negative (not to pollute the augment pool with both entries since str +1 and str -1 are two different augments, for example)

TL;DR - if someone has reliable info on items like that I could adjust during this PR

MowFord:
well that was tedious, but it's done and confirmed the coffer still functions by spamming a few key item retrievals

image

MowFord:
Ugh, submodule got added to my commit, Will figure out how to revert...

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
```diff
@@ -532,6 +1619,7 @@ local argumentKeyItems =
 }
 
 entity.onTrigger = function(player, npc)
+    local eligableNexusCape = (xi.settings.main.ENABLE_ACP * xi.settings.main.ENABLE_AMK * xi.settings.main.ENABLE_ASA) ~= 0
```
zach2good: Even though the values for each of these settings is 1 or 0, they're mentally treated like boolean values. I think it would be more appropriate to treat these as bool values rather than integer values that we're trying to multiply out in case one of them is zero.
```lua
local eligableNexusCape = setting1 == 1 and 
                          ... and 
                          ...

File: scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua [hunk 2]
```diff
@@ -30,426 +30,1427 @@ local optionToKI =
...
+        { { augment =  336, power =  4 } }, -- "Sic" & "Ready" ability delay -5
+        { { augment =  337, power =  2 } }, -- Song Recast Delay -3
+        { { augment =  338, power =  0 } }, -- "Barrage" +1
+        { { augment =  339, power = 19 } }, -- "Elemental Siphon" +20
+        { { augment =  340, power =  4 } }, -- "Phantom Roll" ability delay -5
+        { { augment =  341, power =  9 } }, -- "Repair" potency +10%
+        { { augment =  342, power =  4 } }, -- "Waltz" TP cost -50
+        { { augment =   96, power =  6 } }, -- Pet: Accuracy +7 Ranged Accuracy +7
+        { { augment =   97, power =  6 } }, -- Pet: Attack +7 Ranged Attack +7
+        { { augment =  115, power =  7 }  , { augment = 116, power =  7 } }, -- Pet: "Store TP" +8 "Subtle Blow" +8
```
zach2good: For these double augments:
```
{ { augment = 115, power = 7 }, { augment = 116, power = 7 } },
```

^ note the spacing in the middle

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
```diff
@@ -30,426 +30,1427 @@ local optionToKI =
...
+                {
+                    { 516, 0, 5 }, -- INT+1-6
+                    { 517, 0, 4 }, -- MND+1-5
+                    { 518, 0, 4 }, -- CHR+1-5
+                    {  35, 0, 3 }, -- Magic Accuracy+1-4
+                    {  39, 0, 2 }, -- Enmity+0-3
+                    {  53, 0, 2 }, -- Spell Interruption Rate-1-3%
+                }
+            },
+            { cutoff =  600, itemId = xi.item.FETTLE_BELT, augments =
```
zach2good: Can you format all of these uniformly please?
```lua
{
    cutoff =  600,
    itemId = xi.item.FETTLE_BELT,
    augments =
    {
        {  49, 0, 4 }, -- Haste+1-5
```

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
```diff
@@ -492,6 +1513,72 @@ local function givePrize(player, ki)
...
+
+    while x > 1 do
+        bin = tostring(x % 2) .. bin
+        x = math.floor(x / 2)
+    end
+
+    bin = tostring(x) .. bin
+
+    return bin
+end
```
zach2good: If we don't have a util for this already, can you put this in `utils.lua` please?

zach2good (commented):
Apart from some formatting stuff looks pretty solid 👍

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
```diff
@@ -532,6 +1619,7 @@ local argumentKeyItems =
 }
 
 entity.onTrigger = function(player, npc)
+    local eligableNexusCape = (xi.settings.main.ENABLE_ACP * xi.settings.main.ENABLE_AMK * xi.settings.main.ENABLE_ASA) ~= 0
     local receivedNexusCape = player:getCharVar('receivedNexusCape')
```
zach2good: The logic in here, is this a "you've recieved this item once and then thrown it away, so you're boned" scenario? or could this be changed to just look in a player's inventory?

Or is this one of those scenarios I'm a bit afraid of where it's an item _per-account_ and we'd have to look up whether or not any mules have it, or it's in dbox, etc.?

EDIT: I'm informed that it isnt sendable

TeoTwawki: its also given used and on cooldown after the 1st time.

I created addusedItem (which is used in the script) specifically because of this.

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
MowFord: hmm okie dokie

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
MowFord: yea the box has 2 different menu options, one for getting it the first time and one for getting it an additional time

MowFord: also `eligableNexusCape` is determining if nexus cape should be obtainable at all. retail is "purchased all 3 mini expansions", i'll be making a module to adjust this bool to be "completed all 3 mini expansion"

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
MowFord: are you asking for just the augmented items or all items?

this seems like formatting overkill, but idm it's whatever your preference is

image

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
MowFord: added comments to make it more explicit what is going on

scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
zach2good: There are lots of examples of one-line tables, and lots of examples of pure Allman style tables, but (hopefully) not both. Keep the existing oneliners as oneliners and format the larger tables fully:

```lua
    [78] = -- Marduk's Attire Set +1: Fast Cast +3-9%
    {
        items =
        {
            xi.item.MARDUKS_TIARA_P1,
            xi.item.MARDUKS_JUBBAH_P1,
            xi.item.MARDUKS_DASTANAS_P1,
            xi.item.MARDUKS_SHALWAR_P1,
            xi.item.MARDUKS_CRACKOWS_P1,
        },
        minEquipped = 2,
        mods =
        {
            { xi.mod.FASTCAST, 3, 5, 7, 9 },
        },
    },
```

We don't have a rule for it (yet), but I'd say it's never OK to combine the two styles

MowFord:
I don't think i've missed any comments and i've run through random key items to spot check things still work

zach2good (approved):
LGTM! 🚀 

Reviewers: remember to squash!

OP: You could look into `git rebase -i` and `git commit --amend` for your workflow in the future, so you can have more agency over what commits might get taken when a PR is merged. We don't want to squish all your hard work into a single commit, but conversely, we also don't want to take these into our git history:
image

zach2good:
My assumption here is that the "you get the cape already on cooldown thing" already exists and that this PR doesn't change it. It'll need to eventually be refactored (with a migration too, because: database) to use the `UniqueEvent` system and enums.

You can read more about that [here](https://github.com/LandSandBoat/server/pull/4240). TLDR: Its a way for us to get rid of/move "forever-vars" into a bitfield in `char_unlocks`

PR #4513, opened 2023-09-16
Updates Abyssea Atma and Normal Keyitems list and by zone

Mortalelite:
**_I affirm:_**

Updates the Keyitem list by zone name and ID
Added missing Keyitems
Set monsters names in order

[BG-Wiki Abyssea](https://www.bg-wiki.com/ffxi/Category:Abyssea)

[ffxiclopedia](https://ffxiclopedia.fandom.com/wiki/Category:Abyssea?so=search#Vision_of_Abyssea_Areas)

Make sure each monster gives the Keyitem


scripts/globals/abyssea.lua
```diff
@@ -236,114 +236,120 @@ local demiluneKeyItems =
...
-    ['Yaanei']        = { ['Atma'] = { xi.ki.ATMA_OF_THE_IMPREGNABLE_TOWER }, ['Normal'] = {                                  } },
-
-    -- Konschtat Highlands
-    ['Gangly_Gean']        = { ['Atma'] = {                                    }, ['Normal'] = { xi.ki.FRAGRANT_TREANT_PETAL    } },
-    ['Raskovnik']          = { ['Atma'] = { xi.ki.ATMA_OF_THRASHING_TENDRILS   }, ['Normal'] = { xi.ki.FETID_RAFFLESIA_STALK    } },
-    ['Clingy_Clare']       = { ['Atma'] = {                                    }, ['Normal'] = { xi.ki.DECAYING_MORBOL_TOOTH    } },
-    ['Fistule']            = { ['Atma'] = { xi.ki.ATMA_OF_VICISSITUDE          }, ['Normal'] = { xi.ki.TURBID_SLIME_OIL         } },
-    ['Kukulkan']           = { ['Atma'] = { xi.ki.ATMA_OF_THE_NOXIOUS_FANG     }, ['Normal'] = { xi.ki.VENOMOUS_PEISTE_CLAW     } },
-    ['Tonberry_Lieje']     = { ['Atma'] = {                                    }, ['Normal'] = { xi.ki.TWISTED_TONBERRY_CROWN   } },
+    -- Abyssea - Konschtat (zone 15)
```
TeoTwawki, line 241: for future reference:

1. When you are going to re-sequence the order of a thing, do that as its own separate commit in the same pr either before or after other changes. having them combined leads to additional difficulties for reviewers and makes it take longer to merge in most cases.
2. It is a good idea to check with the LSB team before re-sequencing anything just in case, because there will be some times where we don't want that at all.

Mortalelite: Ok I will keep that in mind for future PR. I just seen a -- TODO: Separate by zone and was just adjust by zone. I did confirm each list to the drops according to Bg-Wiki. Hopefully it looks good beyond that.

Mortalelite:
This should be ready for review.

zach2good (approved):

Xaver-DaRed (approved):

PR #4515, opened 2023-09-17
Add missing mini-expansion key item trading logic to Squintrox Dryeyes

MowFord:
**_I affirm:_**

This is the partner to my Tenshodo Coffer PR. Squintrox logic is not fun to look at in this first commit, but I hope to utilize the setbit logic from tenshodo coffer and build a couple tables to shorten the core logic. 

It is tested and technically works, but I am not satisfied with the layout, just putting it here to get feedback on a path to the end result.

Edit: Testing is complete and I'm satisfied with the behavior

Notes about the logic for this NPC:
 - The order of page 1 `onTrigger` does NOT match the order of event update options and event finish options, so this is manually defined per option value. 
 - The order of page 2 `onTrigger` DOES match the order of event finish options, so this was simple to map `option - 299` to the `menuMetadata` index

This one is tricky to test as there's multiple layers of logic here:
- Trades:
  - Trading the required items for: `CRIMSON_KEY`, `WHITE_CORAL_KEY`, and `MOOGLE_KEY` are the simplest. 
  - Trade should only succeed if the `charVar` for that particular key item maps to a timestamp on the player's charvar less than now (this value is set to next jst midnight upon a successful trade)
- Page 1 key items:
  - Page 1 lists the key items available based on
    - The respective mini expansion being complete (mission log shows the `FIN` mission as active)
    - the corresponding timestamp in the player's `charVar` player variable.
    - player not currently having the key item
  - if the player chooses that key item, the requirements are checked:
    - if the player has all required key items (most require full list, `BIRD_KEY` only requires 3 of the 6 key items) they receive the key item and all req (and prereq) key items are removed
    - if the player doesn't meet the requirements, info is given:
      - for all but `BLACK_CORAL_KEY` and `MOOGLE_KEY`, a simple text is given stating the requirements for obtaining the repeat key item
        - `MOOGLE_KEY`: the particular enfeeblement kit required is chosen/listed
        - `BLACK_CORAL_KEY`: the particular zone for digging is chosen/listed
- Page 2 key items:
  - These are all prerequisite key items for completing the mini-expansion battles again to obtain the key item rewards
  - They are not displayed if any is true
    - The respective mini expansion is not complete (mission log shows the `FIN` mission as active)
    - you currently have _ALL_ the key items 
    - you're on cooldown due to the `charVar` timestamp on your player
  - If chosen, they cost a varied amount of gil/beastmen seals, with custom messages for each type of result

Various screenshots of testing

image

image

image


scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
+            now >= player:getCharVar('LastMoogleKey'))             and
+                64 or 0) +
+        ((finishedASA                                              or
+            player:hasKeyItem(xi.ki.BIRD_KEY)                      and
+            now >= player:getCharVar('LastBirdKey'))               and
+                128 or 0) +
+        ((finishedASA                                              or
+            player:hasKeyItem(xi.ki.BOMB_KEY)                      and
+            now >= player:getCharVar('LastBombKey'))               and
+                256 or 0)
```
zach2good: This looks suspiciously like it could be broken out into a table and then iterated over

```lua
-- { expansion, keyItem, maskValue }
local lookup =
{
    { <mark ACP condition somehow>, xi.ki.CRIMSON_KEY, bit.lshift(1, 1) },
    ...
}

local arg1 = 0
for _, entry in pairs(lookup) do
    local hasCompletedExpansion = player:checkCompleteEtc...(entry[1])
    local hasRelevantKeyItem        = player:hasKI...(entry[2])
    if hasCompletedExpansion or hasRelevantKeyItem then
        local maskValue = entry[3]
        arg1 = arg1 + maskValue
    end
end
```

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
 end
 
 entity.onTrigger = function(player, npc)
-    if
-        xi.settings.main.ENABLE_ACP == 0 and
-        xi.settings.main.ENABLE_AMK == 0 and
-        xi.settings.main.ENABLE_ASA == 0
-    then
+    local now          = os.time()
+    local nextMidnight = getMidnight()
```
zach2good: These two aren't used in here

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
-            os.time() > lastCrimson
-        then -- and timer stuff here) then
+            not player:hasKeyItem(ki) and
+            now >= player:getCharVar('LastCrimsonKey')
+        then
             player:tradeComplete()
-            player:addKeyItem(xi.ki.CRIMSON_KEY)
-            player:setCharVar('LastCrimsonKey', getMidnight())
+            player:addKeyItem(ki)
+            player:setCharVar('LastCrimsonKey', nextMidnight)
```
zach2good: With the different key trades, you can probably write some helper funcs to deal with the repeated logic

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
+    local hasMoogleItems        = kit ~= 0 and
+                                    trade:hasItemQty(kit, 1) and
+                                    count == 1
+    local hasCrimsonItems       = trade:hasItemQty(xi.item.SEEDSPALL_LUX, 1)        and
+                                    trade:hasItemQty(xi.item.SEEDSPALL_LUNA, 1)     and
+                                    trade:hasItemQty(xi.item.SEEDSPALL_ASTRUM, 1)   and
+                                    count == 3
+    local hasWhiteCoralItems    = trade:hasItemQty(xi.item.ORCISH_PLATE_ARMOR, 1)   and
+                                    trade:hasItemQty(xi.item.QUADAV_BACKSCALE, 1)   and
+                                    trade:hasItemQty(xi.item.YAGUDO_CAULK, 1)       and
```
zach2good: This giant block of things defined at the top is quite unsightly, but this is quite an overloaded NPC - so the option are to use these blocks inline if they're only used one (like here), or to break them out into local functions that takes the trade and/or the player as args and returns a bool

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
+    then
+        local ki = xi.ki.MOOGLE_KEY
+        if
+            not player:hasKeyItem(ki) and
+            now >= player:getCharVar('LastMoogleKey')
+        then
+            player:tradeComplete()
+            player:addKeyItem(ki)
+            player:setCharVar('LastMoogleKey', nextMidnight)
+            player:setCharVar('ASA_kit', 0)
```
zach2good: I don't know the current state of ASA in LSB, but this will eventually be replaced by the numbered vars of the interaction framework, so I'd look into what those will look like for their eventual migration

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
+            player:messageSpecial(ID.text.KEYITEM_OBTAINED, ki)
+        else
+            player:messageSpecial(ID.text.DRYEYES_3, ki)
         end
+    -- Moogle Key: trade proper enfeebling kit
+    elseif
+        hasEnfeebKit and
+        finishedASA
+    then
+        local ki = xi.ki.MOOGLE_KEY
```
zach2good: This extraction doesn't save much, just keep the usage inline

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
     else
-        player:startEvent(323)
+        player:startEvent(323, arg1)
     end
 end
 
+-- validation of prereqs done here, event option updated based on checks to be used in onEventFinish
 entity.onEventUpdate = function(player, csid, option, npc)
+    if csid == 323 then
+        if option == 100 then -- Viridian Key
```
zach2good, line 405: Each of the blocks in here are quite unique, and take up quite varying amounts of space - I'd probably break all of them out into their own functions

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
+            -- Reminder that "True" means the option should be excluded from the player's menu
+            local arg1 = (not finishedACP                                                or
+                             (player:hasKeyItem(xi.ki.SEEDSPALL_ROSEUM)                 and
+                             player:hasKeyItem(xi.ki.SEEDSPALL_CAERULUM)                and
+                             player:hasKeyItem(xi.ki.SEEDSPALL_VIRIDIS))                and
+                                 2 or 0) +
+                         (not finishedACP                                                or
+                             (player:hasKeyItem(xi.ki.MARK_OF_SEED))                    and
+                                 4 or 0) +
+                         (not finishedACP                                                or
```
zach2good: I'd look into the pattern I mentioned before about this being a data table that's then iterated over. Remember you can nest tables, so you can have a table of KI requirements as an entry in the main table

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
+            local ki = xi.ki.BLACK_BOOK
+            if not player:hasKeyItem(ki) then
+                player:addKeyItem(ki)
+                player:messageSpecial(ID.text.KEYITEM_OBTAINED, ki)
+            end
+        elseif option == 307 then -- Elemental Saps
+            player:delSeals(15, 0)
+            player:delGil(1500)
+
+            for i, ki in pairs({
```
zach2good: For things like this, I'd define the table first and then use it, rather than inline in the loop definition

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,117 +3,534 @@
...
+                if not player:hasKeyItem(ki) then
+                    player:addKeyItem(ki)
+                    player:messageSpecial(ID.text.KEYITEM_OBTAINED, ki)
+                end
+            end
+        elseif option == 308 then -- Tablet of Hexes KIs
+            player:delSeals(20, 0)
+            player:delGil(2000)
+
+            for i, ki in pairs({
```
zach2good: For things like this, I'd define the table first and then use it, rather than inline in the loop definition.

Think about how these sorts of things are specific to the expansion they come from. Can they be added to the helpers files for those expansions and then referenced here like `xi.asaHelpers.keyItemsWhatever`?

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: Fair point

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: Good call I copy pasted definitions for consistent naming

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: The real lpt is always in the comments

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: Figured it was better to define all the complex requirements in one place at the top, but I could put them directly in the if blocks if that's preferred

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: I figured this would just be a placeholder until that happens (it'll likely be me and flibe doing them)

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: I think that is the route I will take. The same data is also used in the next event function in the same order so it should extend nicely

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: I like it

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: Oh interesting idea. I think I'd rather keep this file self contained tbh though. Since except for a could char var checks it's mostly self contained

Would it be ok to define the single table in this file with all of it and then update later when we audit the expansions?

In doing the expansions on wings it was very frustrating not having this file fleshed out first

MowFord:
I have learned to run CI checks locally (go me), I also plan on investigating the commit ammending/force push to not require squashing every PR of mine. I don't mind if you squash this one or I can go back and combine the two commits myself.

I hope i didn't get too in my head with the `menuMetadata` table of key item info. I think I struck a balance between not repeating code and readability.

Note that I _have not_ checked the "i'm fully tested this code" box as I'm waiting on feedback on the formatting before I thoroughly check every iteration.

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,57 +3,348 @@
...
+        player:getCurrentMission(xi.mission.log_id.AMK) == xi.mission.id.amk.A_MOOGLE_KUPO_DETAT_FIN
+    then
+        ki = xi.ki.WHITE_CORAL_KEY
+        tradeForKeyItem(player, trade, ki)
+    elseif -- Moogle Key: trade proper enfeebling kit
+        asaKit ~= 0                                                    and
+        trade:hasItemQty(asaKit, 1)                                    and
+        count == 1                                                     and
+        player:getCurrentMission(xi.mission.log_id.ASA) == xi.mission.id.asa.A_SHANTOTTO_ASCENSION_FIN
+    then
```
zach2good: The arbitrary gaps between the statement and the and/or keywords aren't needed here. Only put in extraneous whitespace if you're aligning things to something: like variable names or the = operators

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,57 +3,348 @@
...
+        player:messageSpecial(ID.text.DRYEYES_3, ki)
+        return false
+    end
+end
 
-    -- Crimson Key: Trade Seedspall's Lux, Luna, Astrum
+local function verifyReqKeyItems(player, ki)
+    local entry = menuMetadata[1][ki]
+    local count = 0
+    for i, reqKeyItem in pairs(entry.reqKeyItems) do
```
zach2good: If you're not using i, you should use _ to signal that the key isn't being used

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,57 +3,348 @@
...
+            {
+                xi.ki.MOLDY_WORMEATEN_CHEST,
+            },
+        },
+        [xi.ki.MOOGLE_KEY] =
+        {
+            expansion      = 3,
+            charVar        = 'LastMoogleKey',
+            reqItemCharVar = 'ASA_kit',
+            reqItems =
```
zach2good: This table is nice, but the expansion, charVar and reqItemCharVar keys aren't correctly aligned like the rest of the table - choose one style and stick to it

zach2good (commented):
Liking the new layout 👍

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: Ok so just

```
Statement and
Longer statement and
Short and
Blah
```
?

zach2good: Yessir

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -65,53 +367,162 @@ entity.onTrigger = function(player, npc)
...
+                player:messageSpecial(ID.text.DRYEYES_4)
+            elseif player:getGil() < entry.costGil then
+                player:messageSpecial(ID.text.DRYEYES_5)
             else
-                -- player:showText(player, ?)
-                -- Doesn't seem to be a message for trying when you don't have the key items?
+                player:delSeals(entry.costSeals, 0)
+                player:delGil(entry.costGil)
+
+                for i, ki in pairs(entry.relevantKeyItems) do
```
zach2good: i -> _

MowFord: :( sorry

zach2good (approved):

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -3,58 +3,360 @@
...
+    local entry = menuMetadata[1][ki]
+    local reqItemCount = 0
+    if entry.reqItemCount ~= nil then
+        reqItemCount = entry.reqItemCount
+    end
+
+    local count = 0
+    for _, reqKeyItem in pairs(entry.reqKeyItems) do
+        if reqItemCount > 0 then
+            print(reqKeyItem)
```
claywar: Debug print can be removed

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -65,53 +367,162 @@ entity.onTrigger = function(player, npc)
...
+                player:messageSpecial(ID.text.DRYEYES_5)
             else
-                -- player:showText(player, ?)
-                -- Doesn't seem to be a message for trying when you don't have the key items?
+                player:delSeals(entry.costSeals, 0)
+                player:delGil(entry.costGil)
+
+                for _, ki in pairs(entry.relevantKeyItems) do
+                    if not player:hasKeyItem(ki) then
+                        player:addKeyItem(ki)
```
claywar: These two lines can be replaced by the npcUtil function (which can handle either single input or tables of KIs)

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
```diff
@@ -65,53 +367,162 @@ entity.onTrigger = function(player, npc)
...
+            end
+        elseif option == 103 then -- Bird Key prereqs to run the bcnms
+            npcUtil.giveKeyItem(player, menuMetadata[1][xi.ki.BIRD_KEY].prereqKeyItems)
+        elseif option == 104 then -- Bomb Key
+            takeReqKeyItems(player, xi.ki.BOMB_KEY)
+        elseif option == 105 then -- Bird Key
+            takeReqKeyItems(player, xi.ki.BIRD_KEY)
+        elseif option >= 300 and option <= 308 then
+            local entry = menuMetadata[2][option - 299]
+            if player:getSeals(0) < entry.costSeals then
```
claywar, line 510: Not as familiar with this NPC, but does he really pull from stored seals and don't require a trade?

MowFord: Sure does!

https://ffxiclopedia.fandom.com/wiki/Talk:Squintrox_Dryeyes

> She doesn't take the beastmen seals from you, she takes them from [Shami](https://ffxiclopedia.fandom.com/wiki/Shami). That's why she mentions that she and Shami have a deal. Trade the seals to him and you'll be able to buy the key items from Squintrox. --[Kyrie](https://ffxiclopedia.fandom.com/wiki/User:Kyrial) 02:28, June 30, 2010 (UTC)

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
MowFord: Good catch.

MowFord:
I've updated the things clay found, but decided not to force push since zach already approved it

zach2good:
Don't worry about force pushing things etc. it doesn't dismiss the approval (I think) and GitHub is pretty good at showing `New changes since you last viewed` and even keeps copies of earlier commits.

We're also pretty good at not yeeting merges in if other people have ongoing conversations

scripts/zones/Port_Jeuno/npcs/Squintrox_Dryeyes.lua
zach2good: And still does:
image

MowFord:
Sorry to hijack a different PR but I just wanted to pose a quick question about the AMK work we'll be starting up.

Once this PR is merged I'd like to update the feature branch with `base` (to capture this and tenshodo coffer updates). Would you like me to `git merge base` for my first PR or would you like to update the feature branch before I submit the first PR? I'm not well versed in the best ways to handle merge commits.

zach2good:
Once this merges I'll rebase the feature branch, so you can start targetting that without having to do any merges 👍

PR #4516, opened 2023-09-17
Instrument swaps wont remove Aftermath effect

overefined:
**_I affirm:_**

Aftermath effect will persist when swapping instruments.

1. equip Mandau
2. !tp 1000
3. use Mercy Stroke
4. swap instrument

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -1805,8 +1805,8 @@ namespace charutils
                     if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
                     {
                         PChar->health.tp = 0;
+                        PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH);
                     }
-                    PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH);
                     BuildingCharWeaponSkills(PChar);
                     UpdateWeaponStyle(PChar, equipSlotID, nullptr);
                 }
@@ -2651,6 +2651,7 @@ namespace charutils
             {
                 // If the weapon ISN'T a wind based instrument or a string based instrument
                 PChar->health.tp = 0;
+                PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH);
             }
 
             if (!PChar->getEquip(SLOT_MAIN) || !PChar->getEquip(SLOT_MAIN)->isType(ITEM_EQUIPMENT) ||
@@ -2659,7 +2660,6 @@ namespace charutils
                 CheckUnarmedWeapon(PChar);
             }
 
-            PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH);
             BuildingCharWeaponSkills(PChar);
             PChar->pushPacket(new CCharAbilitiesPacket(PChar));
         }


cocosolos:
Is this accurate?
> The Aftermath effect immediately wears off if any weaponry is equipped, switched, or unequipped, whether this is done by the player or caused by an enemy or other gameplay mechanic. (In short, if an equipment change would cause your TP to reset to zero, it will also remove Aftermath.)

https://ffxiclopedia.fandom.com/wiki/Aftermath

This PR also looks like it only applies to the level 75 relic weapons.

claywar:
Easy to test, but also what happens in the case of SAM where there's a potential of two separate aftermath effects? I think there needs to be a check to remove the appropriate one.

I think (with a bit of rework to magian), we could just use the onItemEquip/Unequip functions to remove the effect if this is proven to be what actually happens.  I don't like the idea of having to create giant enums in core to handle this.

overefined:
> Is this accurate?
> 
> > The Aftermath effect immediately wears off if any weaponry is equipped, switched, or unequipped, whether this is done by the player or caused by an enemy or other gameplay mechanic. (In short, if an equipment change would cause your TP to reset to zero, it will also remove Aftermath.)
> 
> https://ffxiclopedia.fandom.com/wiki/Aftermath
> 
> This PR also looks like it only applies to the level 75 relic weapons.

Yeah that is inaccurate according to a couple sources. Aftermath should stay up if a bard switches an instrument for example. It may be only instruments that this applies to. I could extend it to work on all weapons that give the Aftermath effect.

> Easy to test, but also what happens in the case of SAM where there's a potential of two separate aftermath effects? I think there needs to be a check to remove the appropriate one.
> 
> I think (with a bit of rework to magian), we could just use the onItemEquip/Unequip functions to remove the effect if this is proven to be what actually happens. I don't like the idea of having to create giant enums in core to handle this.

The case where there are two separate Aftermath effects came to mind when I wrote this. I wanted to post it here to see if there was a better way to implement this. I like the idea of doing it in magian. I will see if I can get that to work and update the PR.

Edit: If it is only instruments this applies to it will be an easy fix and not require any additional enums.

if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
{
    PChar->health.tp = 0;
}
PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH);

will be changed to:

if (((CItemWeapon*)PItem)->getSkillType() != SKILL_STRING_INSTRUMENT && ((CItemWeapon*)PItem)->getSkillType() != SKILL_WIND_INSTRUMENT)
{
    PChar->health.tp = 0;
    PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH);
}

Aeltorian:
There is also the bard instruments that is not suppose to remove it's aftermath upon changing to another instrument or mainhand weapon. 

I do have a retail video to verify this.

cocosolos:
> There is also the bard instruments that is not suppose to remove it's aftermath upon changing to another instrument or mainhand weapon.
> 
> I do have a retail video to verify this.

You have video of aftermath staying after changing weapon, not instrument? It's my understanding that changing instrument doesn't wipe TP and shouldn't remove aftermath, but changing weapon should.

Aeltorian:
> > There is also the bard instruments that is not suppose to remove it's aftermath upon changing to another instrument or mainhand weapon.
> > I do have a retail video to verify this.
> 
> You have video of aftermath staying after changing weapon, not instrument? It's my understanding that changing instrument doesn't wipe TP and shouldn't remove aftermath, but changing weapon should.

Big oof on my part, let me correct myself. Changing instruments should not wipe the mainhand weapon's aftermath. Here's the video of going through different instruments and ammo slot changes after applying Carnwenhan's aftermath. Notice aftermath doesn't get removed and TP remains. 

[117MB video](https://drive.google.com/file/d/1DEpd8G_F6ViIPQb6BHIxzY-ItBgWqO3q/view?usp=sharing)

cocosolos:
> > > There is also the bard instruments that is not suppose to remove it's aftermath upon changing to another instrument or mainhand weapon.
> > > I do have a retail video to verify this.
> > 
> > 
> > You have video of aftermath staying after changing weapon, not instrument? It's my understanding that changing instrument doesn't wipe TP and shouldn't remove aftermath, but changing weapon should.
> 
> Big oof on my part, let me correct myself. Changing instruments should not wipe the mainhand weapon's aftermath. Here's the video of going through different instruments and ammo slot changes after applying Carnwenhan's aftermath. Notice aftermath doesn't get removed and TP remains.
> 
> [117MB video](https://drive.google.com/file/d/1DEpd8G_F6ViIPQb6BHIxzY-ItBgWqO3q/view?usp=sharing)

Okay this makes more sense and I can see where there is an issue with LSB. You can see in that vid that TP stays at 55 the whole time. Looks like the only change really needed is to make sure all calls to `PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH)` are paired with the calls to `PChar->health.tp = 0`. There's 2 instances where it isn't: [here](https://github.com/LandSandBoat/server/blob/791f5efec80df7a6a467c07c47b31b8015de792c/src/map/utils/charutils.cpp#L1805-L1809) and [here](https://github.com/LandSandBoat/server/blob/791f5efec80df7a6a467c07c47b31b8015de792c/src/map/utils/charutils.cpp#L2651-L2662).

src/map/utils/charutils.cpp
```diff
@@ -1780,7 +1780,6 @@ namespace charutils
                     PChar->look.sub            = 0;
                     PChar->m_Weapons[SLOT_SUB] = itemutils::GetUnarmedItem(); // << equips "nothing" in the sub slot to prevent multi attack exploit
                     PChar->health.tp           = 0;
-                    PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH);
```
cocosolos: This line should stay. When TP gets reset aftermath gets cleared unless you have video showing otherwise.

src/map/utils/charutils.cpp
```diff
@@ -1837,7 +1836,6 @@ namespace charutils
                     }
 
                     PChar->health.tp = 0;
-                    PChar->StatusEffectContainer->DelStatusEffect(EFFECT_AFTERMATH);
```
cocosolos: This one too.

cocosolos (approved):
Lgtm

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #4522, opened 2023-09-19
Adjust gardening quantity calculation and elemental ore yield

lapislosh:
**_I affirm:_**

This does three things related to gardening results:

1. **Increases the range of the RNG strength assigned to each plant from 0-31 to 0-32**. The RNG strength is later divided by 32 and therefore the maximum previous value was 96.8% making it impossible to ever hit 100 strength, making the maximum yield of certain gardening combinations impossible.

2. Elemental ores were previously impossible to hit their max yield of 2. With the first fix, this is now possible. However, there doesn't seem to be any evidence that 2 elemental ores is possible without the use of arcane pots. **Therefore, I'm reducing the maximum yield of all ores to 1**.

3. **Normalizes the distribution of quantity so that all potential quantities have equal ranges of when they're chosen.**
The old algorithm for quantity is:
min+(max-min)*percent+0.1
The new algorithm is:
min+(1+max-min)*percent
I'm not really sure how the original algorithm is supposed to work, and I suspect the +0.1 on the end may have been a hack to try and account for the fact that the RNG was previously capped at 96.8% instead of 100%. It's easier to compare the old and new results here, using Persikos from Tree Saplings + double Dark Crystal feeds:

```
strength:60 (0.00) old: 4 new: 4
strength:61 (0.04) old: 4 new: 4
strength:62 (0.07) old: 4 new: 4
strength:63 (0.11) old: 4 new: 4
strength:64 (0.14) old: 4 new: 4
strength:65 (0.18) old: 4 new: 4
strength:66 (0.21) old: 4 new: 5
strength:67 (0.25) old: 5 new: 5
strength:68 (0.29) old: 5 new: 5
strength:69 (0.32) old: 5 new: 5
strength:70 (0.36) old: 5 new: 5
strength:71 (0.39) old: 5 new: 5
strength:72 (0.43) old: 5 new: 6
strength:73 (0.46) old: 5 new: 6
strength:74 (0.50) old: 6 new: 6
strength:75 (0.54) old: 6 new: 6
strength:76 (0.57) old: 6 new: 6
strength:77 (0.61) old: 6 new: 7
strength:78 (0.64) old: 6 new: 7
strength:79 (0.68) old: 6 new: 7
strength:80 (0.71) old: 6 new: 7
strength:81 (0.75) old: 7 new: 7
strength:82 (0.79) old: 7 new: 7
strength:83 (0.82) old: 7 new: 8
strength:84 (0.86) old: 7 new: 8
strength:85 (0.89) old: 7 new: 8
strength:86 (0.93) old: 7 new: 8
strength:87 (0.96) old: 7 new: 8
strength:88 (1.00) old: 8 new: 8
```

The old algorithm slightly over-represented the minimum and greatly under-represented the maximum, while the new version is an even split across all quantities.

Note that while I did fix elemental ore yields, this will significantly impact tree sapling yields as well. Previously there would have been roughly a 5% chance to 2x saplings from a harvest and now it will be 50%. I don't have any hard data to back this up as accurate but I did see several random forum posts suggesting that it wasn't uncommon to get more saplings harvest than cuttings planted, so it seems closer to reality in my opinion.

!additem 217 10
Place flowerpots in mog house

Open the char_inventory db table. Find the entries for your 10 pots and set the extra field to:
```
0a40080807001000110009008691d9288619dc2800000000
0a40080807021100110008008691d9288619dc2800000000
0a40080807081100100007008891d9288819dc2800000000
0a400808070a1000100006008991d9288919dc2800000000
0a400808071810000f0005008a91d9288a19dc2800000000
0a400808072011000f0004008b91d9288b19dc2800000000
0a400808072c11000e0003008c91d9288c19dc2800000000
0a400808072e10000e0002008d91d9288d19dc2800000000
0a400808073c10000d0001008f91d9288f19dc2800000000
0a400808074111000d0000009191d9289119dc2800000000
```
Assign the extra field of each pot to 1 row above, in hex. These are tree saplings with double dark feeds set to harvest state, with a different strength for each pot. Respectively, the strengths and expected yields (compared to the data posted above) when harvesting:
```
60 - 4 persikos
61 - 4 persikos
65 - 4 persikos
66 - 5 persikos
75 - 6 persikos
80 - 7 persikos
87 - 8 persikos
88 - 1 dark ore
97 - 1 dark ore
100 - 1 dark ore
```

### DIFF
--- a/sql/gardening_results.sql
+++ b/sql/gardening_results.sql
@@ -1944,421 +1944,421 @@ INSERT INTO `gardening_results` VALUES (1930,6,8,8,1228,4,12,20); -- Darksteel N
 INSERT INTO `gardening_results` VALUES (1931,7,0,0,936,12,24,30); -- Rock Salt
 INSERT INTO `gardening_results` VALUES (1932,7,0,0,4449,1,2,30); -- Reishi Mushroom
 INSERT INTO `gardening_results` VALUES (1933,7,0,0,1309,4,12,28); -- Gold Leaf
-INSERT INTO `gardening_results` VALUES (1934,7,0,0,1261,1,2,12); -- Light Ore
+INSERT INTO `gardening_results` VALUES (1934,7,0,0,1261,1,1,12); -- Light Ore
 INSERT INTO `gardening_results` VALUES (1935,7,0,1,936,12,24,15); -- Rock Salt
 INSERT INTO `gardening_results` VALUES (1936,7,0,1,4105,1,2,24); -- Ice Cluster
 INSERT INTO `gardening_results` VALUES (1937,7,0,1,4449,1,2,15); -- Reishi Mushroom
 INSERT INTO `gardening_results` VALUES (1938,7,0,1,628,1,2,20); -- Cinnamon
 INSERT INTO `gardening_results` VALUES (1939,7,0,1,1309,4,12,14); -- Gold Leaf
-INSERT INTO `gardening_results` VALUES (1940,7,0,1,1255,1,2,6); -- Fire Ore
-INSERT INTO `gardening_results` VALUES (1941,7,0,1,1261,1,2,6); -- Light Ore
+INSERT INTO `gardening_results` VALUES (1940,7,0,1,1255,1,1,6); -- Fire Ore
+INSERT INTO `gardening_results` VALUES (1941,7,0,1,1261,1,1,6); -- Light Ore
 INSERT INTO `gardening_results` VALUES (1942,7,0,2,936,12,24,15); -- Rock Salt
 INSERT INTO `gardening_results` VALUES (1943,7,0,2,4106,1,2,24); -- Wind Cluster
 INSERT INTO `gardening_results` VALUES (1944,7,0,2,4449,1,2,15); -- Reishi Mushroom
 INSERT INTO `gardening_results` VALUES (1945,7,0,2,1310,6,12,20); -- Platinum Leaf
 INSERT INTO `gardening_results` VALUES (1946,7,0,2,1309,4,12,14); -- Gold Leaf
-INSERT INTO `gardening_results` VALUES (1947,7,0,2,1256,1,2,6); -- Ice Ore
-INSERT INTO `gardening_results` VALUES (1948,7,0,2,1261,1,2,6); -- Light Ore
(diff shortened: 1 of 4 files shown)


WinterSolstice8 (approved):
LGTM for code changes, I have no input on ele ore quantity.

WinterSolstice8:
Looks like your git commit title is too long,

We enforce a 72 character limit, you can use `git commit --amend` to change it then `git push --force` to update your branch

TeoTwawki (approved):
> However, there doesn't seem to be any evidence that 2 elemental ores is possible without the use of arcane pots.

anecdotal and I can't prove: but its possible just rare as F. Approving as is, because we don't even have full understanding of every detail of retail and might never unless SE discloses more like they did with TH mechanics.. We should try and remember this next time an SE rep does a Q&A with the community.

lapislosh:
> > However, there doesn't seem to be any evidence that 2 elemental ores is possible without the use of arcane pots.
> 
> anecdotal and I can't prove: but its possible just rare as F. Approving as is, because we don't even have full understanding of every detail of retail and might never unless SE discloses more like they did with TH mechanics.. We should try and remember this next time an SE rep does a Q&A with the community.

Yeah I couldn't find any conclusive evidence either way. My reasoning for this was mainly that with the old algorithm, it wasn't possible to get 2 (unless the gardening "myths" were turned on), so the safest thing is to make sure you can't get them with the new algorithm either.
If we did want to keep the maximum as 2 but make it super rare, we would need to adjust the gardening_results.sql to have multiple entries for each ore, and only the last entry would have a yield of 2.

TeoTwawki:
as for myself I've only got personal memories for what I say, and that's not a good enough to base code on. I used to do a lot of gardening (mainly for the platinum leaves resulting from failed ores).

TeoTwawki:
> If we did want to keep the maximum as 2 but make it super rare, we would need to adjust the gardening_results.sql to have multiple entries for each ore, and only the last entry would have a yield of 2.

I'd prefer to keep what you have for now, and hope that someday we squeeze some more info out of SE or someone really passionate about this goes on a long data collecting spree.

zach2good (approved):
Good maths and comments, thanks! 🚀

PR #4523, opened 2023-09-20
DRG AF Quest: "Knight Stalker" NM Spawn Fix

UmeboshiXI:
**_I affirm:_**

Allows the 2 shadow NMs to spawn when:

- Player is DRG.
- Player has pet wyvern out and alive.

Previously the NMs would not spawn as intended, stopping quest progression.

- `!changejob drg 75' and flag quest (or comment out the charvar quest requirements).
- `!gotoid 17428862`
- Use Call Wyvern.
- Click on ??? (ID:17428862).
- See that 2 shadow NMs spawn.

Note: In future, this quest should be converted to the new framework but this will allow the quest to function in the mean time.

### DIFF
--- a/scripts/zones/Temple_of_Uggalepih/npcs/qm15.lua
+++ b/scripts/zones/Temple_of_Uggalepih/npcs/qm15.lua
@@ -17,11 +17,11 @@ entity.onTrigger = function(player, npc)
         player:getCharVar('KnightStalker_Progress') == 4 and
         player:getCharVar('KnightStalker_Kill') == 0 and
         player:getMainJob() == xi.job.DRG and
-        pet and
-        pet:getPetID() == xi.petId.WYVERN and
-        npcUtil.popFromQM(player, npc, { ID.mob.CLEUVARION_M_RESOAIX, ID.mob.ROMPAULION_S_CITALLE }, { hide = 0, claim = false })
+        pet ~= nil and
+        pet:getPetID() == xi.petId.WYVERN
     then
         player:messageSpecial(ID.text.SOME_SORT_OF_CEREMONY + 1) -- Your wyvern reacts violently to this spot!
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (dismissed):

zach2good:
Logically, does this allow 2x different players to spawn the NM at the same time? I believe thats why `npcUtil.popFromQM` is in the if block, because it checks for such a scenario

scripts/zones/Temple_of_Uggalepih/npcs/qm15.lua
```diff
@@ -17,7 +17,7 @@ entity.onTrigger = function(player, npc)
         player:getCharVar('KnightStalker_Progress') == 4 and
         player:getCharVar('KnightStalker_Kill') == 0 and
         player:getMainJob() == xi.job.DRG and
-        pet and
+        pet ~= nil and
```
zach2good, line 20: `player:getPet()` can only return a valid CLuaBaseEntity or nil, checking `pet` and `pet ~= nil` are equivalent if the pet has come out correctly, what did you observe here?

zach2good:
```cpp

std::optional<CLuaBaseEntity> CLuaBaseEntity::getPet()
{
    auto* PBattle = static_cast<CBattleEntity*>(m_PBaseEntity);

    if (PBattle->PPet != nullptr)
    {
        return std::optional<CLuaBaseEntity>(PBattle->PPet);
    }

    return std::nullopt;
}
```

UmeboshiXI: Might end up closing this. Could be something weird on my end? I wasn't getting any lua errors. When I restarted my server/client, it happened to work fine afterwards. Before it was just not popping or displaying a message.

PR #4530, opened 2023-09-21
Cache char_equip data prior to processing in charutils::LoadEquip()

claywar:
**_I affirm:_**

`charutils::LoadEquip()` was querying char_equip table, and then other functions could have sub-queries prior to exiting, which could clobber data.  This PR caches the results of the initial query prior to any other processing to prevent that.

1. Equip an item in mainhand with a magian trial inscribed, and have other gear equipped
2. Log out
3. Log in

All equipment should be loaded successfully

image

Also confirmed that trials are unaffected, and proc as necessary.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -1052,38 +1052,45 @@ namespace charutils
 
         if (ret != SQL_ERROR)
         {
+            // equipSlotData[equipSlotId] = { slotId, containerId }
+            std::unordered_map<uint8, std::pair<uint8, uint8>> equipSlotData;
+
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):
This would fix weird stuff happening like characters getting fully naked on zone when zoning with a weapon with a trial?

claywar:
> This would fix weird stuff happening like characters getting fully naked on zone when zoning with a weapon with a trial?

Yes, if I can figure out why the win32 build hates me

WinterSolstice8 (approved):
LGTM but I'd add a comment on why we're caching because it's not something we'd normally do

PR #4540, opened 2023-09-23
Make inject command use a subdir instead of repository root directory.

TeoTwawki:
**_I affirm:_**

Prior to this change, binary files used with the `!inject` command had to sit in the root of the repo, cluttering it up. I'm made a subdirectory named `injection` and adjusted the command to use it.

Drop a packet bin into the `injection` folder, and `!inject` it.

### DIFF
--- a/injection/README.md
+++ b/injection/README.md
@@ -1,0 +1,5 @@
+# Packet injection
+
+This directory is for placing binary files for use with the GM command `!inject` 
+
+You can construct these binary files using packet capture and a hex editor.
--- a/scripts/commands/inject.lua
+++ b/scripts/commands/inject.lua
@@ -23,7 +23,7 @@ commandObj.onTrigger = function(player, packet)
     end
 
     -- inject packet
-    player:injectPacket(packet)
+    player:injectPacket('./injection/' .. packet)
 end
 
 return commandObj


zach2good:
People actually use !inject? I wouldn't think it's in common enough use to warrant it's own folder in the repo root?

zach2good:
It also loads from an arbitrary path:
```
    uint8 size = 0;
    FILE* File = fopen(filename.c_str(), "rb");
```

So you could specify: `./injection/packet_name` or `C:/ffxi/server/injection/packet_name` etc.

Adding the `/injection/` folder to the path breaks being able to specify an arbitrary path

zach2good:
Internally, we could update this to use `std::filesystem::path` to make sure Windows/Unix paths are handled correctly all the time

TeoTwawki:
1. I am possibly the last person left that even remembers this is a thing, but it should continue to exist because we should at least be _able_ to do it without reliance on ashita or windower. While its been years since I used one, I used to have a folder with like 50 prebuilt packets in it for testing purposes and now that I think about it, this could be useful for automated testing of certain exploits to be sure future contributors do not UN-fix one. 

2. Typing out a path every time as part of your input is irritating and this was just never a good feature: everyone who ever used this command threw their packet files in the repo root or if savvy enough like me just changed the command as I did in my own fork years ago (except I named the dir "packets" which modern me feels is bad naming due to `src/packets` existing).

3. Before this change. the path you can specify is relative to the repo root `../uponeLevel2mySideDir/packet`, and after its from that new directory `./injection + /../newpath/packet`

4. This is either a feature nobody but me cares about or is a feature I am breaking, can't be both. I'd actually rather remove the specify path feature entirely than anything else related to that pathing (just being honest).

5.  To give you a laugh: Darkstar used to have it as the recommended way to check item appearance on their wiki. IIRC its a carry-over from PXI.

Let me know what you wanna for-sure do with it and I'll adjust. I don't really want to murder it from repo because of point1, but if we did I'd just keep it on my fork and it wouldn't matter.

PR #4551, opened 2023-09-26
[LUA] Fixes self target in "!givemagianitem" command

N3ckB3ard:
**_I affirm:_**

Corrects issue with "target" in the !givemagianitem command

image
image

Use command and give the item to a player, rather than yourself

### DIFF
--- a/scripts/commands/givemagianitem.lua
+++ b/scripts/commands/givemagianitem.lua
@@ -32,9 +32,9 @@ commandObj.onTrigger = function(player, target, trialId, isRewardItem)
         player:PrintToPlayer(string.format('Player \'%s\' does not have free space for that item!', target))
     else
         if giveRewardItem then
-            xi.magian.giveRewardItem(player, trialId)
+            xi.magian.giveRewardItem(target, trialId)
         else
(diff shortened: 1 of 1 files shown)


scripts/commands/givemagianitem.lua
```diff
@@ -32,7 +32,7 @@ commandObj.onTrigger = function(player, target, trialId, isRewardItem)
         player:PrintToPlayer(string.format('Player \'%s\' does not have free space for that item!', target))
     else
         if giveRewardItem then
-            xi.magian.giveRewardItem(player, trialId)
+            xi.magian.giveRewardItem(target, trialId)
```
claywar, line 35: This isn't an entirely accurate fix, and only half-implements your desired intent.  I'm fine with requiring a target, but you didn't update the else condition, which in this case, giveRewardItem bool determines if its the requiredItem or rewardItem in the magian table.

N3ckB3ard: I updated the additional player to target

Xaver-DaRed (approved):

Issue #3632, opened 2023-02-27
🐛 [CoP] Timers of spawn ???s for Jailer of Faith and Ix'aern (DRK) are wrong


DiscipleOfEris:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Windows10

## Branch affected by issue 

`base`

## Steps to reproduce

1. Find the ??? for JoF or Ix'DRK (or go to it with a command).
2. Mark time down when ??? disappears.
3. Find it again.
4. Mark down time when it disappears.
5. For JoF ???, the time varies dramatically, from 2:34 (a single in-game hour) to pretty much any other multiple of in-game hours.
6. For Ix'DRK, the *duration* is *almost* correct, but it always an in-game hour that's a multiple of 12.

The code really raises some eyebrows: https://github.com/LandSandBoat/server/blob/base/scripts/zones/The_Garden_of_RuHmet/Zone.lua#L88

```Lua
local vanadielHour = VanadielHour()
local qmDrk = GetNPCByID(ID.npc.QM_IXAERN_DRK) -- Ix'aern drk
local s = math.random(6, 12) -- wait time till change to next spawn pos, random 15~30 mins.

-- Jailer of Faith spawn randomiser
if vanadielHour % s == 0 then
    local qmFaith = GetNPCByID(ID.npc.QM_JAILER_OF_FAITH) -- Jailer of Faith
    qmFaith:hideNPC(60) -- Hide it for 60 seconds
    qmFaith:setPos(unpack(ID.npc.QM_JAILER_OF_FAITH_POS[math.random(1, 5)])) -- Set the new position
end
```

So for every in-game hour divisible by the numbers 6-12 (which includes 0, 14, 16, 18, 20, 21), it has a chance to move the JoF ???.

## Expected behavior

[...]

PR #4552, opened 2023-09-26
Move Jailer of Faith QM every 30 minutes

lapislosh:
**_I affirm:_**

Moves the Jailer of Faith QM exactly every 30 minutes from when it lasts appears. The original implementation is completely broken, as mentioned in #3632. There is some slight contention on whether it moves every 30 minutes or 15 minutes, but as best as I can tell it was either updated in a patch to 30 minutes, or people are confusing the generic 15m (at the time) respawn on QMs vs the 30m move timer.

The code mimics Phantom Worm's movement code but requires some extra logic to track the case where JoF may pop and die before the timer triggers, requiring the timer to be reset (which is not particularly easy and instead I opted to just start a second timer and ignore the initial timer when it triggers). Phantom Worm conveniently ignores this problem because the move timer is so short it's not really noticeable.

I did not update Ix'Drk but if this PR goes through I can give it similar treatment.

5 locations for pop:
-- North / Hume tower  !pos -420 0 -157 35
-- NE / Elvaan tower   !pos -157 0 -340 35
-- SE / Galka tower    !pos -260 0 -643 35
-- SW / Tarutaru tower !pos -580 0 -644 35
-- NW / Mithra tower   !pos -683 0 -340 35

!additem 1899 for HQ euvhi organ

1. Find ??? and wait for it to hide
2. Wait 60s then find ??? in new location
3. Wait for it to hide again exactly 30m after it re-appeared (test default move logic)
4. Spawn Jailer of Faith and kill before the 30m timer triggers (test timer overlap code)
5. Make sure the QM moves exactly 30m from when the QM reappears after killing him
6. Spawn Jailer of Faith, and hold it for 30 minutes, then kill it (test timer cancel code)
7. Make sure the QM moves exactly 30m from when the QM reappears after killing him
It's much easier to test by changing the timer to 30 seconds!

### DIFF
--- a/scripts/zones/The_Garden_of_RuHmet/Zone.lua
+++ b/scripts/zones/The_Garden_of_RuHmet/Zone.lua
@@ -56,10 +56,6 @@ zoneObject.onInitialize = function(zone)
     qmDrk:setPos(unpack(ID.npc.QM_IXAERN_DRK_POS[qmDrkPos]))
     qmDrk:setLocalVar('hatedPlayer', 0)
 
-    -- Give the Faith ??? a random spawn
-    local qmFaith = GetNPCByID(ID.npc.QM_JAILER_OF_FAITH)
-    qmFaith:setPos(unpack(ID.npc.QM_JAILER_OF_FAITH_POS[math.random(1, 5)]))
-
     -- Give Ix'DRG a random placeholder by picking one of the four groups at random, then adding a random number of 0-2 for the specific mob.
     local groups = ID.mob.AWAERN_DRG_GROUPS
     SetServerVariable('[SEA]IxAernDRG_PH', groups[math.random(1, #groups)] + math.random(0, 2))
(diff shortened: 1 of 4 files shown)


zach2good (approved):
LGTM 🚀

Is it known what happens during server restart on retail? Whether or not the previously chosen point persists, or if its rerolled?

claywar (approved):

lapislosh:
> LGTM 🚀
> 
> Is it known what happens during server restart on retail? Whether or not the previously chosen point persists, or if its rerolled?

Unknown afaik, however we don't currently persist any of the randomization in Sea. Faith, Temperance, Ix'Drg, Ix'Drk, and Fortitude are all randomized on zone initialization

lapislosh:
Sorry, still learning but just found out there's a clearTimerQueue function which I can use instead of tracking the number of timers running, so I'd like to switch to that and simplify the code a bit before submitting.

PR #4557, opened 2023-09-27
[Fix] Souleater effect modifier stacking

Xaver-DaRed:
**_I affirm:_**

As per this BG article https://www.bg-wiki.com/ffxi/Souleater
Equipment that enhances Souleater (Souleater +X) does not stack. An exception happens with augments.

This PR (attempts to):
- Correct this by using a very useful function found in lua_baseentity: `getMaxGearMod()` which returns the highest value of a modifier found in gear, insstead of the total value.
- Implements that function in battleentity, so it can be used in core.

Use Souleater with different pieces of gear that enhance Souleater and see if it only takes the higest value.

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -1188,6 +1188,41 @@ int16 CBattleEntity::getMod(Mod modID)
     return m_modStat[modID];
 }
 
+/************************************************************************
+ *                                                                      *
+ *  Get the highest value of the specified modifier across all gear     *
+ *                                                                      *
+ ************************************************************************/
+int16 CBattleEntity::getMaxGearMod(Mod modID)
+{
+    TracyZoneScoped;
+    CCharEntity* PChar       = dynamic_cast<CCharEntity*>(this);
+    uint16       maxModValue = 0;
+
+    if (!PChar)
+    {
+        ShowWarning("CBattleEntity::getMaxGearMod() - Entity is not a player.");
+
+        return 0;
+    }
+
+    for (uint8 i = 0; i < SLOT_BACK; ++i)
+    {
+        auto* PItem = PChar->getEquip((SLOTTYPE)i);
+        if (PItem && (PItem->isType(ITEM_EQUIPMENT) || PItem->isType(ITEM_WEAPON)))
+        {
+            uint16 modValue = PItem->getModifier(modID);
+
+            if (modValue > maxModValue)
+            {
+                maxModValue = modValue;
+            }
+        }
+    }
+
+    return maxModValue;
+}
+
 void CBattleEntity::addPetModifier(Mod type, PetModType petmod, int16 amount)
 {
     TracyZoneScoped;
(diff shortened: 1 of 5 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -1188,6 +1188,41 @@ int16 CBattleEntity::getMod(Mod modID)
     return m_modStat[modID];
 }
 
+/************************************************************************
+ *                                                                      *
+ *  Get the highest value of the specified modifier across all gear     *
+ *                                                                      *
+ ************************************************************************/
+int16 CBattleEntity::getMaxGearMod(Mod modID)
```
zach2good, line 1196: Now that this exists here, you can update the lua binding to use this implementation instead of having the same code exist twice

CatsEyeXI:
tested this patch today, still having similar results as before. i will build a clean LSB server later on and have the player test again, though im positive we're not doing anything custom with this ability.
**note**: this is a normal hit, not weaponskill.

image

src/map/entities/battleentity.cpp
Xaver-DaRed: Been a busy week, but I havent forgotten about this.

Xaver-DaRed: Done, I think. Do tell if any other hing needs a change.

src/map/lua/lua_baseentity.cpp
```diff
@@ -12666,28 +12666,14 @@ bool CLuaBaseEntity::delLatent(uint16 condID, uint16 conditionValue, uint16 mID,
...
-        {
-            uint16 modValue = PItem->getModifier(modId);
-            if (modValue > maxModValue)
-            {
-                maxModValue = modValue;
-            }
-        }
-    }
-    return maxModValue;
+    return static_cast<CBattleEntity*>(m_PBaseEntity)->getMaxGearMod(static_cast<Mod>(modId));
```
zach2good: Inside the function that this calls, you're checking for whether or not this is a Char entity, so why not cast to a char here?

src/map/lua/lua_baseentity.cpp
```diff
@@ -12666,28 +12666,14 @@ bool CLuaBaseEntity::delLatent(uint16 condID, uint16 conditionValue, uint16 mID,
...
 int16 CLuaBaseEntity::getMaxGearMod(Mod modId)
 {
-    CCharEntity* PChar       = dynamic_cast<CCharEntity*>(m_PBaseEntity);
-    uint16       maxModValue = 0;
-
-    if (!PChar)
+    if (m_PBaseEntity->objtype != TYPE_PC)
     {
-        ShowWarning("CLuaBaseEntity::getMaxGearMod() - m_PBaseEntity is not a player.");
+        ShowWarning("Invalid Entity (NPC: %s) calling function.", m_PBaseEntity->GetName());
```
zach2good: This could get called on a Pet or a Mob, not necessarily just an NPC

zach2good (approved):
Seems fine apart from those two points

src/map/lua/lua_baseentity.cpp
Xaver-DaRed: Changed it to Non-PC, since it is meant for player characters only.

src/map/lua/lua_baseentity.cpp
Xaver-DaRed: Changed it to CCharEntity*

TeoTwawki (approved):

PR #4567, opened 2023-09-30
Faster server variable handling

TeoTwawki:
**_I affirm:_**

This definitely DOES speed it up:
- Before: 277ms - 550ms
- After:  7ms - 329ms

Buuuuut...

The thing is that range on these bothers me. those are quite large gaps between the smallest and largest times I saw. Different sources online make conflicting claims about which operations are faster. One person claiming REPLACE statements are 32x slower than  ON DUPLICATE KEY UPDATE yet in my testing REPLACE was 200% faster, other claiming ON DUPLICATE KEY UPDATE was faster than a regular UPDATE statement, which I don't see how that can be physically possible - on duplicate attempt an insert and then when it fails, does an update. its literally the same update. REPLACE does a delete followed by an insert, and has minefields for situations involving foreign keys and auto increments.

So I did "real world testing" my dang self. You clearly can't trust what you read. And then I reached out to some experts I'm waiting to hear back from. 

I'm not sure I want to move forward wit this yet, so its in draft mode pending response from persons more experienced than myself in SQL communities.

Here's some data in the meantime:

image

Print those sql times
https://github.com/LandSandBoat/server/blob/8f2779f9bbacb7f5e7ee8cffb3112de6508dde7c/src/common/sql.cpp#L353
```cpp
        else
        {
            ShowInfo(fmt::format("SQL query took {}ms: {}", dTime.count(), self->buf));
        }
```
and run some queries that were updated in this pr.

### DIFF
--- a/scripts/zones/Mhaura/Zone.lua
+++ b/scripts/zones/Mhaura/Zone.lua
@@ -16,7 +16,7 @@ zoneObject.onGameHour = function(zone)
         GetNPCByID(ID.npc.LAUGHING_BISON):setAnimationSub(0)
     end
 
-    SetServerVariable('Mhaura_Deastination', math.random(1, 100))
+    SetServerVariable('Mhaura_Destination', math.random(1, 100))
 end
 
 zoneObject.onInitialize = function(zone)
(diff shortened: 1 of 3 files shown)


TeoTwawki:
Decided this is definitely NOT the direction I want to try and patch this up with. I still don't understand why the servervar queries are being this slow on having to update existing rows. There only one key here and neither the table nor the row should have a bad locking problem going on (we do 3 tries at reading it to confirm it was written, but that is going to take < 30ms in total).

PR #4578, opened 2023-10-03
AMK mission 7 progress: Choco digging spots

Flibe-XI:
**_I affirm:_**

Progress on AMK missions
- Update helm logic to simpler format
- Added all chocobo digging locations

- !addmission 10 6
- Click on inconspicuous door in upper jeuno for cutscene, get a zone assigned
- Go dig that zone for moldy worm-eaten chest


Flibe-XI:
I'm new to this codebase, I think I'm doing this right but apologies if there is something I missed.

PR #4579, opened 2023-10-03
[AMK] helm logic updated, chocobo digging spots added

Flibe-XI:
**_I affirm:_**

Progress on AMK missions
- Update helm logic to simpler format
- Added all chocobo digging locations

- !addmission 10 6
- Click on inconspicuous door in upper jeuno for cutscene, get a zone assigned
- Go dig that zone for moldy worm-eaten chest

### DIFF
--- a/scripts/missions/amk/helpers.lua
+++ b/scripts/missions/amk/helpers.lua
@@ -32,38 +32,30 @@ local validRegions = set{
 xi.amk.helpers.helmTrade = function(player, helmType, broke)
     local amkChance = 5
     local regionId = player:getCurrentRegion()
+    local helmMapping =
+    {
+        [xi.helm.type.MINING] = xi.ki.STURDY_METAL_STRIP,
+        [xi.helm.type.LOGGING] = xi.ki.PIECE_OF_RUGGED_TREE_BARK,
+        [xi.helm.type.HARVESTING] = xi.ki.SAVORY_LAMB_ROAST,
+    }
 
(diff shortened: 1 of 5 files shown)


scripts/missions/amk/helpers.lua
```diff
@@ -78,128 +70,273 @@ local digZoneIds =
...
 
-xi.amk.helpers.tryRandomlyPlaceDiggingLocation = function(player)
-    local diggingZoneOffset = xi.amk.helpers.getDiggingZone(player)
-    local diggingSiteTable  = xi.amk.helpers.digSites[diggingZoneOffset]
+xi.amk.helpers.getDiggingZone = function(player)
+    -- Returns zone id of 1 of the 13 possible zones, and sets variable
+    local diggingZone = player:getCharVar('Mission[10][6]diggingZone')
 
-    player:setLocalVar('AMK_DIG_SITE_INDEX', math.random(1, #diggingSiteTable))
+    if diggingZone == 0 then
```
MowFord: Should probably just change to check if `xi.amk.helpers.digSites[diggingZone] == nil` This will catch zero but also bunk values.

Storing the actual zone instead of an arbitray index is A++ though from a GM perspective.

WinterSolstice8 (approved):
Don't really see anything wrong with this, deferring to Clay for nitpicks

Xaver-DaRed (approved):

PR #4594, opened 2023-10-09
Dark Ixion: Spawn/Spook mechanics

MowFord:
DI will wander/change zones but is unable to be claimed by a PC

**_I affirm:_**

This is step 1/2 to PR Dark Ixion. I figured his spawning mechanics were novel enough to be its own PR before I push the rest of the code from WingsXI.

DI's zone and last spawn time are stored as server variables. Each zone runs `onInit` and `onGameHour` to confirm DI's spawning.

He can spawn in any of these zones:
```
xi.zone.EAST_RONFAURE_S
xi.zone.FORT_KARUGO_NARUGO_S
xi.zone.BATALLIA_DOWNS_S
xi.zone.GRAUBERG_S
xi.zone.ROLANBERRY_FIELDS_S
xi.zone.WEST_SARUTABARUTA_S
xi.zone.JUGNER_FOREST_S
```

Once spawned, he follows a pre-set roaming path, sometimes reversing the path. The Paths were built by hand using ffxidb maps as source data.

If he is aggroed by any means, he pauses for 10 seconds then runs away at much higher run speed than even his normal fast speed (normal is +75%, run-away speed is +350%)

The next PR will include claim mechanics using Stygian Ash to set a localvar that allows him to claim. This PR will just engage then immediately disengage/run away

One issue I have is that the pathing doesn't seem to pick the closest point to continue when pathing is interrupted. i.e. 
 - executing `mob:pathThrough(xi.darkixion.zoneinfo[mob:getZoneID()].pathList, xi.path.flag.RUN)` after he's run through his path for a while seems to make him run straight shot to the first point in his path list for that zone. Minor issue tbh, but still an issue I'd like to resolve.

This PR includes only the spawn/spook mechanics for DI, so proper testing involves confirming only a few things:
 - on a fresh server startup he gets initialized to a zone. 
   - This is done via `xi.darkixion.zoneOnInit`. It will only do the reset logic if the server doesn't map to a key in `xi.darkixion.zoneinfo`
 - on game hour properly cycles zones every 24 hours
   - This can easily be confirmed by setting the server var `DarkIxion_PopTime` to zero (or if you want to be more exact, set it to `os.time() - 24 * 60 * 60`)
 - on game hour properly spawns him in the next zone after doing previous check
 - final (tedious) step: all zones function properly
   - do the step to reset poptime to zero, then you have until next game hour to reset the zoneid to whichever zone you want DI to pop in

### DIFF
--- a/scripts/globals/dark_ixion.lua
+++ b/scripts/globals/dark_ixion.lua
@@ -1,0 +1,464 @@
+-----------------------------------
+-- Area: One of many zones in Shadowreign zones
+--   NM: Dark Ixion
+-- https://ffxiclopedia.fandom.com/wiki/Dark_Ixion
+--[[
+    Find DI by:
+    - checking the server's zone var: "!checkvar server DarkIxion_ZoneID"
+    - go to that zone: !zone <zoneID>
+    - go to DI: !gotoname Dark_Ixion
+
+    Outline of the mob's behavior:
+
+    Will run away, most of the time to another zone, if:
+    - Campaign Battle begins
+    - a player scares it by getting too close
+    - attacks or attempts to claim with a Stygian Ash but misses
+    - or if whoever has successfully claimed it wipes.
+    - His HP is preserved after wiping and him running away to another zone
+    - If Dark Ixion is due to spawn or is already spawned during maintenance, he will spawn shortly after server comes back online.
+    - If he was not due to spawn during this time frame, his spawn window will reset to 21 hours after servers come online.
+
+    Roaming: runs very fast and can be spooked via normal aggro/claim actions. If spooked, it does the same thing as attempting a claim but missing with a Stygian Ash
+    Claim: can only be claimed by landing a hit with Stygian Ash
+    Targetting: Can damage and kill players riding Chocobo with Area of Effect attacks even if player is not in the Alliance fighting him.
+
+    Mobskills:
+    - has a normal set of TP moves, but telegraphs most/all of them
+    - if he has an aura then his TP moves are doubled up back to back
+    - has a special move that isn't listed in the combat log, that many have called 'Trample'
+        - Trample: Charges forward, dealing high damage to,(400-1000) and lowering the MP (10-30%) of, anyone in his path. No message is displayed in the chat log.
+        - When Dark Ixion's HP is low, he can do up to 3 Tramples in succession.
+        - Can be avoided easily by moving out of its path.
+        - May charge in the opposite, or an entirely random, direction from the one he is currently facing.
+        - Will load a set number of targets in his path before ramming forward.
+        - Occasionally, a person in his path will not be hit, as well as those wandering in its path after it has begun its charge.
+--]]
+-----------------------------------
+require('scripts/globals/npc_util')
+
+xi = xi or {}
+xi.darkixion = xi.darkixion or {}
+
+-- TODOs, notes, and reminders
+
+-- mob:AnimationSub() -- 0 is normal || Charging is animation sub 1  || 2 is broken horn || 3 is glowing and causes horn to repair
+-- TODO: dmg taken from front/rear (if we can)
+
+xi.darkixion.zoneinfo =
+{
+    [xi.zone.JUGNER_FOREST_S] =
+    {
+        pathList =
+        {
+            { x = -191.5, y = 000.0, z =  092.0 },
+            { x = -139.6, y = 005.0, z =  097.0 },
+            { x = -064.0, y = 002.5, z =  093.0 },
+            { x = -025.5, y = 004.5, z =  057.0 },
+            { x = 0025.0, y = 006.0, z =  040.0 },
+            { x = 0087.0, y = 002.0, z =  029.0 },
+            { x = 0074.0, y = 000.0, z = -011.0 },
+            { x = 0087.0, y = 000.0, z = -083.0 },
+            { x = 0118.0, y = 000.0, z = -109.0 },
+            { x = 0088.0, y = 000.0, z = -170.0 },
+            { x = 0061.0, y = 002.0, z = -156.0 },
+            { x = 0030.8, y = 002.0, z = -179.5 },
+            { x = -031.8, y = 000.0, z = -196.5 },
+            { x = -098.5, y = 000.0, z = -169.0 },
+            { x = -177.5, y = 000.0, z = -152.0 },
+            { x = -218.0, y = 000.0, z = -234.0 },
+            { x = -323.0, y = 002.0, z = -260.7 },
+            { x = -225.0, y = 007.0, z = -110.0 },
+            { x = -220.7, y = 006.0, z = -107.0 },
+            { x = -260.6, y = 006.0, z = -091.0 },
+            { x = -259.0, y = 001.0, z = -030.5 },
+            { x = -304.0, y = 001.4, z =  004.0 },
+            { x = -299.0, y = 002.7, z =  072.5 },
+            { x = -291.0, y = 009.0, z =  140.5 },
+            { x = -202.7, y = 002.0, z =  139.0 },
+        }
+    },
+    [xi.zone.WEST_SARUTABARUTA_S] =
+    {
+        pathList =
+        {
+            { x = 0231.8, y = -025.0, z = 258.5 },
+            { x = 0227.5, y = -017.5, z = 209.0 },
+            { x = 0134.6, y = -020.5, z = 195.0 },
+            { x = 0068.0, y = -012.4, z = 194.8 },
+            { x = 0068.0, y = -020.0, z = 254.0 },
+            { x = -027.0, y = -012.0, z = 296.4 },
+            { x = -019.0, y = -011.7, z = 349.0 },
+            { x = 0016.0, y = -018.0, z = 405.6 },
+            { x = 0092.0, y = -022.0, z = 391.5 },
+            { x = 0143.7, y = -037.5, z = 344.6 },
+            { x = 0098.8, y = -027.0, z = 269.0 },
+        }
+    },
+    [xi.zone.ROLANBERRY_FIELDS_S] =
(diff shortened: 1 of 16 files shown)


scripts/zones/Batallia_Downs_[S]/mobs/Dark_Ixion.lua
```diff
@@ -1,11 +1,69 @@
...
+entity.onMobEngaged = function(mob, target)
+    xi.darkixion.onMobEngaged(mob, target)
+end
+
+entity.onMobDisengage = function(mob)
+    xi.darkixion.onMobDisengage(mob)
+end
+
+entity.onMobFight = function(mob, target)
+    xi.darkixion.onMobFight(mob, target)
```
Comment (zach2good) [inline scripts/zones/Batallia_Downs_[S]/mobs/Dark_Ixion.lua]: You can replace these with a setup helper function in the global:

```
local entity = {}

xi.darkixion.setupEntity(entity)

return entity
```

```
-- All the functions you're going to use in the setup helper should be defined before, so they're available

xi.darkixion.setupEntity = function(enitity)
    entity.onMobEngaged = function(mob, target)
        xi.darkixion.onMobEngaged(mob, target)
    end

    entity.onMobDisengage = function(mob)
        xi.darkixion.onMobDisengage(mob)
    end

    -- etc.
end

File: scripts/globals/dark_ixion.lua [hunk 2]
```diff
@@ -0,0 +1,471 @@
...
+-- TODOs, notes, and reminders
+
+-- mob:AnimationSub() -- 0 is normal || Charging is animation sub 1  || 2 is broken horn || 3 is glowing and causes horn to repair
+-- TODO: dmg taken from front/rear (if we can)
+
+xi.darkixion.zoneinfo =
+{
+    [xi.zone.JUGNER_FOREST_S] =
+    {
+        mobID = 17113468,
```
zach2good: No need to hard-code the ID here, you can do a dynamic lookup using the zone object whenever you need to mobID (everything is cached and fast):

```
    local darkIxion = zone:queryEntitiesByName('Dark_Ixion')[1]
    local mobID     = darkIxion:getID()
```

scripts/globals/dark_ixion.lua
```diff
@@ -0,0 +1,471 @@
...
+end
+
+xi.darkixion.onCriticalHit = function(mob, attacker)
+end
+
+xi.darkixion.onWeaponskillHit = function(mob, attacker, weaponskill)
+end
+
+xi.darkixion.onMobInitialize = function(mob)
+end
```
zach2good: Will these be needed later? If not ,they don't need to exist

scripts/globals/dark_ixion.lua
```diff
@@ -0,0 +1,471 @@
...
+
+xi.darkixion.onMobSkillFinished = function(mob, target, skill)
+end
+
+xi.darkixion.onAdditionalEffect = function(mob, target, damage)
+end
+
+xi.darkixion.onMobRoamAction = function(mob)
+end
```
zach2good: Same with these: Will these be needed later? If not ,they don't need to exist

scripts/globals/dark_ixion.lua
```diff
@@ -0,0 +1,471 @@
...
+    -- wiki says 'It can pop there in less than 10 seconds or take the whole 15 minutes'
+    SetServerVariable('DarkIxion_PopTime', os.time() + math.random(1, 15 * 60)) -- based on onGameHour function timing
+end
+
+xi.darkixion.roamingMods = function(mob)
+    -- don't take damage until the fight officially starts
+    mob:setMod(xi.mod.UDMGPHYS   , -100)
+    mob:setMod(xi.mod.UDMGRANGE  , -100)
+    mob:setMod(xi.mod.UDMGBREATH , -100)
+    mob:setMod(xi.mod.UDMGMAGIC  , -100)
```
zach2good: Extraneous space and alignment after UDMGBREATH

scripts/globals/dark_ixion.lua
```diff
@@ -0,0 +1,471 @@
...
+        table.insert(keys, k)
+    end
+
+    local randZoneID = keys[math.random(#keys)]
+    SetServerVariable('DarkIxion_ZoneID', randZoneID)
+    -- wiki says 'It can pop there in less than 10 seconds or take the whole 15 minutes'
+    SetServerVariable('DarkIxion_PopTime', os.time() + math.random(1, 15 * 60)) -- based on onGameHour function timing
+end
+
+xi.darkixion.roamingMods = function(mob)
```
zach2good, line 286: This could be more descriptive, it looks like its doing a lot more than just adding mods. Maybe break out a bunch of the stuff at the end into a reset/reinit function?

zach2good (commented):
Looks pretty good, I'm not sure if it's worth doing RPC to handle Ixion moving between zones, so ServerVars are a good shout 👍

scripts/globals/dark_ixion.lua
MowFord: Yea. I basically wanted to build the framework for the combat portion so next PR is just filling in those functions

scripts/globals/dark_ixion.lua
MowFord: I'll add a comment, I made it a function so I didn't have to replicate the logic in two different mob functions. Basically contains all the things that are needed when DI is roaming.

scripts/globals/dark_ixion.lua
MowFord: oh, wow. OK cool I'll update things. Makes my table really just a skeleton to contain the patrol paths and valid zones.

scripts/zones/Batallia_Downs_[S]/mobs/Dark_Ixion.lua
Comment (MowFord) [reply scripts/zones/Batallia_Downs_[S]/mobs/Dark_Ixion.lua]: Fancy. I'll play around with it thanks

MowFord:
> Looks pretty good, I'm not sure if it's worth doing RPC to handle Ixion moving between zones, so ServerVars are a good shout 👍

Yea the way DI moves in retail appears to be the same type of logic: DI leaves a zone, trigger happens between 10s and 15m (so likely on some kind of game time trigger). There's no real need for real-time updates

scripts/globals/dark_ixion.lua
MowFord: Would it be ok to leave the mobIDs in there as reference. Debugging is much simpler when you can `!gotoid <mobid>` unless there's a similar gm command?

zach2good: I've never used it, but it looks like this exists:

`!gotoname Dark_Ixion`

scripts/globals/dark_ixion.lua
MowFord: I'll remove them for this PR since I don't have to touch all the files again using your helper function above

MowFord:
I found path flags that weren't defined in lua land. Used that to define a set of patrol pathpoints (very slick) and those appear to resume nicely after `mob:stun()`. I believe this is ready for review: round 2

scripts/zones/Batallia_Downs_[S]/mobs/Dark_Ixion.lua
Comment (MowFord) [reply scripts/zones/Batallia_Downs_[S]/mobs/Dark_Ixion.lua]: Well I thought it worked, but seems like the server crashes using this trickery. I tried creating local entity in `setupEntity` like so
```
xi.darkixion.setupEntity = function()
    local entity = {}

        xi.darkixion.onMobDeath(mob, player, optParams)
    end

        xi.darkixion.onMobDespawn(mob)
    end

        xi.darkixion.onMobSpawn(mob)
    end

    entity.onMobRoam = function(mob)
        xi.darkixion.onMobRoam(mob)
    end

    entity.onMobEngaged = function(mob, target)
        xi.darkixion.onMobEngaged(mob, target)
    end

    entity.onMobDisengage = function(mob)
        xi.darkixion.onMobDisengage(mob)
    end

    return entity
end
```

and did this in the `Dark_Ixion.lua`

```
-- Area: West Sarutabaruta [S]
--   NM: Dark Ixion
require('scripts/globals/dark_ixion')

return xi.darkixion.setupEntity()
```

> The thread 0x39d0 has exited with code 0 (0x0).
Exception thrown at 0x00007FFC11A9CF19 (KernelBase.dll) in xi_map.exe: 0xE24C4A02.
The thread 0x564 has exited with code 0 (0x0).
The thread 0xda70 has exited with code 0 (0x0).
The thread 0x8864 has exited with code 0 (0x0).
The thread 0x8910 has exited with code 0 (0x0).
The thread 0x17d8 has exited with code 0 (0x0).
The thread 0x5cd4 has exited with code 0 (0x0).
The thread 0xa7dc has exited with code 0 (0x0).
The thread 0xe920 has exited with code 0 (0x0).
The thread 0xa460 has exited with code 0 (0x0).
The thread 0xc33c has exited with code 0 (0x0).
The thread 0x3f8 has exited with code 0 (0x0).
The thread 0x771c has exited with code 0 (0x0).
The thread 0x25f4 has exited with code 0 (0x0).
The thread 0xf3cc has exited with code 0 (0x0).
The thread 0xb598 has exited with code 0 (0x0).
The thread 0x7f04 has exited with code 0 (0x0).
The thread 0xd414 has exited with code 0 (0x0).
The thread 0x8954 has exited with code 0 (0x0).
Exception thrown at 0x00007FFC11A9CF19 (KernelBase.dll) in xi_map.exe: 0xE24C4A02.

Comment (MowFord) [reply scripts/zones/Batallia_Downs_[S]/mobs/Dark_Ixion.lua]: To be clear I also tried your method first (the code is currently in this PR using that original method)

Comment (MowFord) [reply scripts/zones/Batallia_Downs_[S]/mobs/Dark_Ixion.lua]: Disregard ... it was an issue with `mob:atPoint()`. Not sure why I couldn't capture with the Visual Studio debugger, but all working now.

zach2good (approved):

PR #4600, opened 2023-10-12
[chore] CCharEntity::ValidTarget cleanup and use TARGET_PLAYER_ALLIANCE

zach2good:
**_I affirm:_**

I got told that we don't handle `TARGET_PLAYER_ALLIANCE` anywhere, so I had a look at the relevant code and saw it was a clusterfuck - so I cleaned it up and made it readable, and added the new condition. Nothing SHOULD change here, but the it was so messy I may have slipped up doing it

Lots of testing to do to make sure this does what I think it does

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -873,10 +873,12 @@ void CCharEntity::delTrait(CTrait* PTrait)
 bool CCharEntity::ValidTarget(CBattleEntity* PInitiator, uint16 targetFlags)
 {
     TracyZoneScoped;
+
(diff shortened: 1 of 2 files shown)


zach2good:
Extracting out all the conditions into their own bools, and then replacing them into the original code gives this exactly:
```cpp
        if ((targetsParty || hasPianissimo) && 
            (isSameParty || isPartyPetMaster) &&
            isDifferentChar)
```
So the addition of the alliance stuff seems to make sense

claywar (approved):

zach2good:
Made an alliance, tested a bunch of different spells on self/party/alliance, everything seems to work as before. No actions are currently set with TARGET_PLAYER_ALLIANCE, but they should be supported now

PR #4601, opened 2023-10-12
[core, SQL] Refactor to use mariadb-connector-cpp and introduce new db namespace

zach2good:
**_I affirm:_**

https://github.com/LandSandBoat/server/wiki/Database-Library-Upgrade

Test as normal (I changed some things and invalidated all my old testing, so we're starting from the beginning!):
- [x] All xi_world functionality (HTTP server, conquest system, daily tally)
- [x] All xi_search functionality (/sea all, AH, and anything else that goes through it)
- [x] Regular xi_map functionality continues to work, including staying up for 12+ hours (to make sure TryPing hasn't broken)
- [x] Send a /tell to yourself
- [x] /yell
- [x] Do some quests
- [x] Fight some monsters
- [x] Log in, log out
- [x] Create a character
- [x] Delete a character
- [x] /sea all
- [x] Teleport a full party somewhere
- [x] Look up some stuff on AH
- [x] Look at someone's seacom message
- [x] Hang around and wait for conquest stuff to happen
- [x] Set up and use the HTTP API
- [x] Form a party, chat, disband
- [x] Form a party with an added trust, chat, disband
- [x] Form an alliance, chat, disband
- [x] Talk in a Linkshell
- [x] Make sure RPC still works (SendLuaFuncStringToZone in mog_tablets: `!exec xi.mogTablet.hideAllTablets()`)
- [x] Form a party, chat, etc. across multiple processes
- [x] Console system: gm Testo 5
- [x] Before/after Tracy profiling of regular queries and some query-heavy routines
- [x] Before/after Tracy profiling of a couple of test spots that will use prepared statements
- [x] Confirm everything works with task_system:
```cpp
        // clang-format off
        {
            ts::task_system ts;
            for (auto zoneId : zonesOnThisProcess)
            {
                ts.schedule([zoneId]()
                {
                    TracyZoneScoped;
                    auto* PZone = g_PZoneList[zoneId];

                    auto Query = "SELECT \
```

image


zach2good:
Evening discoveries: both C++-style mariadb connectors are very much 1-indexed. 😷

zach2good:
Added some basic locking to make this use case safe (it would crash out otherwise):
```cpp

    auto thVec = std::vector<std::jthread>{};
    for (int i = 0; i < 32; ++i)
    {
        thVec.emplace_back(std::jthread(
        [&]()
        {
            auto rset = db::query("SELECT * FROM char_inventory;");
            if (rset && rset->rowsCount())
            while (rset->next())
            {
                ShowInfo("%u", rset->getUInt("charid"));
            }
            std::this_thread::sleep_for(std::chrono::milliseconds(xirand::GetRandomNumber(100, 500)));
        }));
    }
```

zach2good:
Another round of tests, this time the new system has been applied to _most_ of the db calls in LoadChar, and it's _ever so slightly faster_.:

Old System:
image

New System:
image

I've been having a lot of issues trying to get consistent results in my performance testing. This time, I made sure to close all processes and restart my MariaDB service - this seems to have given me a more stable baseline to test against.

I'll be happy to go ahead with these changes if we get _the same_ performance, since the library is so much better to use and so much safer than what we use now. If its faster in some cases: great. I just need to hunt around and poke it to see if I can find any places where it's worse: actually worse, not worse because my testing methodology is flawed.

EDIT:
Another test of the new system, where I've given it a couple of warmup calls to prime it, like the old system gets, before the section that I'm profiling. This makes it a fairer fight:
image

zach2good:
In these synthetic benchmarks, it's hard to factor out MariaDB doing caching magic. In a more realistic workload we'd be sending raw queries to the db and the db would have to compile them (or whatever it does) and execute them. The preparedStmt system skips the compiling step for all but the first time the query is called - so in a varied workload we'd see consistent performance improvement. 

At least in these tests, things aren't getting worse, and the code is cleaner and safer.

zach2good:
Disabling the maria query cache: https://mariadb.com/kb/en/server-system-variables/#query_cache_size

Old system, query cache disabled, 7000+ queries:
image

New system, query cache disabled, 7000 queries:
image

Pretty much exactly the same. Which is enough for me to go ahead.

zach2good:
Resolved conflicts. TODO: Need to check some stuff that went down in petutils

zach2good:
petutils is fine

zach2good:
Test rig:
```cpp

    struct TestBlob
    {
        int a;
    };

    TestBlob tb{ 70 };
    auto     blobString = db::toBlobString(tb);
    ShowInfo("do_init: TestBlob: %s", db::toBlobString(tb).c_str());
    ShowInfo("do_init: initializing TEST TABLE");
    {
        auto rset = db::preparedStmt("DROP TABLE IF EXISTS `test_table`");
    }
    {
        auto rset = db::preparedStmt(R"""(
            CREATE TABLE `test_table` (
            `id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT,
            `name` VARCHAR(15) NOT NULL,
            `flag` TINYINT(1) UNSIGNED NOT NULL,
            `smallint3u` SMALLINT(3) UNSIGNED NOT NULL,
            `smallint3` SMALLINT(3) NOT NULL,
            `float7_3` FLOAT(7,3) NOT NULL,
            `blob` BLOB NOT NULL,
            `dt` DATETIME NOT NULL,
            `ts` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
            `txt` TEXT NOT NULL,
            PRIMARY KEY (`id`)
            ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
        )""");
    }

    // Insert into the test table using ? syntax
    ShowInfo("do_init: inserting into TEST TABLE");
    {
        auto rset = db::preparedStmt(R"""(
            INSERT INTO `test_table` (`name`, `flag`, `smallint3u`, `smallint3`, `float7_3`, `blob`, `dt`, `txt`)
            VALUES (?, ?, ?, ?, ?, ?, ?, ?)
        )""",
                                     "Test1", 1, 65535, 32767, 9999.999, blobString, "2021-12-31 23:59:59", "Test1");
    }

    // update and select using the maximum values
    ShowInfo("do_init: selecting from TEST TABLE");
    {
        auto rset = db::preparedStmt("SELECT * FROM `test_table` WHERE `id` = (SELECT MAX(`id`) FROM `test_table`)");
        if (rset->next())
        {
            ShowInfo("do_init: id: %u, name: %s, flag: %u, smallint3u: %u, smallint3: %d, float7_3: %f, blob: %s, dt: %s, txt: %s",
                     rset->getUInt("id"),
                     rset->getString("name").c_str(),
                     rset->getUInt("flag"),
                     rset->getUInt("smallint3u"),
                     rset->getInt("smallint3"),
                     rset->getDouble("float7_3"),
                     rset->getString("blob").c_str(),
                     rset->getString("dt").c_str(),
                     rset->getString("txt").c_str());
        }
    }

    // update and select using the minimum values
    ShowInfo("do_init: updating and selecting from TEST TABLE");
    {
        auto rset = db::preparedStmt("UPDATE `test_table` SET `name` = ?, `flag` = ?, `smallint3u` = ?, `smallint3` = ?, `float7_3` = ?, `blob` = ?, `dt` = ?, `txt` = ? WHERE `id` = (SELECT MIN(`id`) FROM `test_table`)",
                                     "Test2", 0, 0, -32768, -69.0, blobString, "1970-01-01 00:00:00", "Test2");
    }
    {
        auto rset = db::preparedStmt("SELECT * FROM `test_table` WHERE `id` = (SELECT MIN(`id`) FROM `test_table`)");
        if (rset->next())
        {
            // Use db::extractFromBlob to extract data into a vector from 'blob'
            TestBlob outBlob{};
            db::extractFromBlob(rset, "blob", outBlob);

            ShowInfo("Trying to extract blob value: %s", rset->getString("blob").c_str());
            ShowInfo("Got: %d", outBlob.a);

            ShowInfo("do_init: id: %u, name: %s, flag: %u, smallint3u: %u, smallint3: %d, float7_3: %f, blob: %s, dt: %s, txt: %s",
                     rset->getUInt("id"),
                     rset->getString("name").c_str(),
                     rset->getUInt("flag"),
                     rset->getUInt("smallint3u"),
                     rset->getInt("smallint3"),
                     rset->getDouble("float7_3"),
                     std::to_string(outBlob.a),
                     rset->getString("dt").c_str(),
                     rset->getString("txt").c_str());
        }
    }

    return 0;
```

zach2good:
TODO: Rebuild mariadbcpp dlls now that I don't need to maintain my own fork

zach2good:
Test rig values all as expected:
```
[03/30/24 09:06:42:144][map][info] do_init: initializing TEST TABLE (do_init:225)
[03/30/24 09:06:42:161][map][info] do_init: inserting into TEST TABLE (do_init:248)
[03/30/24 09:06:42:162][map][info] do_init: selecting from TEST TABLE (do_init:258)
[03/30/24 09:06:42:163][map][info] do_init: id: 1, name: Test1, flag: 1, smallint3u: 65535, smallint3: 32767, float7_3: 9999.999000, blob: F, dt: 2021-12-31 23:59:59, txt: Test1 (do_init:272)
[03/30/24 09:06:42:163][map][info] do_init: updating and selecting from TEST TABLE (do_init:277)
[03/30/24 09:06:42:165][map][info] Trying to extract blob value: F (do_init:290)
[03/30/24 09:06:42:165][map][info] Got: 70 (do_init:291)
[03/30/24 09:06:42:165][map][info] do_init: id: 1, name: Test2, flag: 0, smallint3u: 0, smallint3: -32768, float7_3: -69.000000, blob: 70, dt: 1970-01-01 00:00:00, txt: Test2 (do_init:302)
```

zach2good:
TODO: A startup sql query that confirms that the db is there, a couple of tables we expect are there, and that there are rows in them. If these fail we should bail out early - instead of getting to the routines that load mob data and spamming errors

zach2good:
CI passes, holy shit 👀

zach2good:
TODO: Add proper doxygen comments to all of the new API, with examples, etc.

zach2good:
TODO: Either wrap the db mutex with `TracyLockable` so it can be observed, or refactor to use `mutex_guarded<T>` on the connection, or on a struct of the connection and prepped stetements

claywar (approved):

zach2good:
TODO: Triple check that if a simple query fails, it'll return nullptr as the rset

zach2good:
TODO: Run through BLU logic to make sure this query works:
```
    void LoadSetSpells(CCharEntity* PChar)
    {
        TracyZoneScoped;

        if (PChar->GetMJob() == JOB_BLU || PChar->GetSJob() == JOB_BLU)
        {
            const char* Query = "SELECT set_blue_spells FROM "
                                "chars WHERE charid = (?);";

            auto rset = db::preparedStmt(Query, PChar->id);
            if (rset && rset->rowsCount() && rset->next())
            {
                db::extractFromBlob(rset, "set_blue_spells", PChar->m_SetBlueSpells);
            }
```

zach2good:
TODO: Ensure that blobs stored in the old system can be loaded by the new system, and that everything works from a before/after standpoint (logging in as BLU, messing around with spells, and zoning is sufficient to check this)

CMakeSettings.json
```diff
@@ -47,30 +47,6 @@
       "ctestCommandArgs": "",
       "inheritEnvironments": [ "msvc_x64_x64" ],
       "variables": []
-    },
-    {
-      "name": "x86-Release",
```
WinterSolstice8, line 52: RIP in pieces, 32bit windows

cmake/FindMariaDB.cmake
```diff
@@ -1,13 +1,16 @@
 if(WIN32)
     # How to build MariaDB Libs:
     #
-    # <Clear your cache>
+    # It's easier to do this in two fresh build folders, so create build_mariadb_32 and build_mariadb_64 folders
+    # in your project root
+   #
```
WinterSolstice8: nit: missing space here to align with the rest

src/common/database.cpp
```diff
@@ -0,0 +1,117 @@
...
+            while ((start_pos = result.find(substr, start_pos)) != std::string::npos)
+            {
+                result.replace(start_pos, substr.length(), to);
+                start_pos += to.length();
+            }
+        }
+        return result;
+    };
+
+    return replaceAll(query, { "--", "/*", "*/", ";" }, "");
```
WinterSolstice8: Isn't this supposed to replace "bad" characters with escaped ones? `'` -> `\'` and so on so that things like chat, search comments, bazaar comments, etc can still include them, but they also wreck injection?

Speaking of, `'` is missing from this list.

WinterSolstice8: see:
https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

TeoTwawki: also the autostranslate junk people can put in bazaar and seacom, thats not even human visible characters.

zach2good: > Isn't this supposed to replace "bad" characters with escaped ones? ' -> \' and so on so that things like chat, search comments, bazaar comments, etc can still include them, but they also wreck injection?
> 
> Speaking of, ' is missing from this list.

My blob encoding handles escaping characters and things, this sanitise routine is specifically to make super ultra mega sure we're removing any and all possible vectors for SQL injection, even if the c connector, cpp connector and other mechanisms all promise they stop it on our behalf

zach2good: autotranslate, bazaar, seacon etc. are all handled as blobs, which I now have acting equivalently to the c-connector, and it's now obvious what it's doing.

This routine is for the cases where plain text is pulled out of incoming packets and used as strings, rather than blobs for the db or the client

WinterSolstice8: are gm/chat audits blobs?

zach2good: Should be, easy test thanks!

zach2good: Will do this in later PRs, I already have evidence of a blob being read from the old system, rewritten and reread correctly

zach2good:
All my TODOs are addressed, all my testing is done, all my comments and thoughts are addressed. Will finish the article, test `TryPing` and then ready for merge

PR #4610, opened 2023-10-15
[monstrosity] Add Terynon Instinct, Mon purchases, Odyssean Passage teleports

claywar:
**_I affirm:_**

* Adds Terynon functions for purchasing Instincts and MONs
* Adds support for Odyssean Passage teleports

NOTE: Odyssean Passage needs additional data populated.  If a location for a zone does not exist, it will default to 0,0,0 (where the Zone will set default onZoneIn), and log

Buy instincts, buy MONs, teleport all around the world.


zach2good (approved):
LGTM once purchasing is fixed

PR #4614, opened 2023-10-17
Pass wsID to xi.weaponskills.takeWeaponskillDamage()

kelppo:
**_I affirm:_**

Before this patch, PUP's Automaton's ranged attack and weaponskill shows 0 damage done because of an error in xi.weaponskills.takeWeaponskillDamage():

[map][error] luautils::onUseWeaponSkill: ./scripts/globals/weaponskills.lua:1012: bad argument #1 to 'lshift' (number expected, got nil)

Strangely enough, Atonement still correctly shows its damage even though it causes the same error, fix this by passing wsID to xi.weaponskills.takeWeaponskillDamage() as expected.

1. Unlock PUP (I had to complete the PUP unlock quest to get the Automaton to be spawned correctly)
2. !addallattachments
3. Equip Sharpshot head and body
4. Deploy Automaton to an enemy
5. Observe how its ranged attack and weaponskill shows 0 damage (target HP is still correctly diminished)
6. Apply this patch
7. Observe how its ranged attack and weaponskill now shows the damage done

### DIFF
--- a/scripts/actions/weaponskills/atonement.lua
+++ b/scripts/actions/weaponskills/atonement.lua
@@ -45,6 +45,7 @@ weaponskillObject.onUseWeaponSkill = function(player, target, wsID, tp, primary,
     }
     local calcParams =
     {
+        wsID = wsID,
         criticalHit = false,
         tpHitsLanded = 0,
         extraHitsLanded = 0,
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):

TeoTwawki (approved):
same issue we saw with _spirits within_ because these don't pass through the "do" functions.

kelppo:
> same issue we saw with _spirits within_ because these don't pass through the "do" functions.

I looked through all the call sites and these seem to be the only ones not passing wsID.

zach2good (approved):

PR #4617, opened 2023-10-17
Dark Ixion: Fight Mechanics

MowFord:
**_I affirm:_**

Makes the pony fight

Claim mechanic confirmed to work via new AE type

image

Once the fight starts, mobskills are all used via logic in onMobFight and TP is preserved via localvar each time his "auto attack" swings happen (since we use `mob:setMobSkillAttack(39)`)

DI's animation subs define mobskill behavior:
- 0 is normal
- 1 is Charging (trample)
- 2 is broken horn, cannot trample
- 3 is glowing and uses double offensive mobskills
  - also goes into this briefly when repairing horn, goes back to 0 afterwards

everytime DI autoattacks there's a chance (which grows as his HP lowers) to stack a localvar "run" counter. If this is nonzero and trample is available, then he starts the trample sequence.
- finds a random player within 30 yalms to trample. if found, he looks in that direction and starts trampling
- immediately places previous target (current battle target) in the `hitlist` and does damage if that player is in front of DI
- every onMobFight tick nearby players are placed in the hitlist and trampled if in front
- each trample path can only hit a player once (though if "run" is > 1 he may trample then immediately trample again and a player will get hit twice)
- Note that he can choose and hit _any_ player (doesn't have to be on the hate table or even in the alliance with claim)

All of his damage mobskills have flag 8 meaning they can hit any player/pet in range

Core fixes needed during testing:
- targetfind wasn't respecting findflags when an ability was conal
- rear conals weren't properly flagged as conal, so they were defaulting to single target. The `findWithinCone` function also didn't support a rear conal angle shift.
  - The only other mobskill with aoe type > 4 is rear lasers so i assume this also fixes it to work like acheron kick :)
- new lua util function `entity:getPlayersInRange()` to get all players within X range of entity. pass zero or nothing to get all players in zone.
- knockback was still applied even for miss/shadow absorb (This fix doesn't work 100% when there's multiple targets, but it _is_ better than it was)
  - I believe the issue is related to checking `PSkill->hasMissMsg()` in the loop of each target. Might be fundamentally flawed for aoe abilities, but i'll leave that for the reviewers to decide a course of action

Finally, there's a few things not perfectly accurate:
- his mobskills should not be able to be ranged to cancel the animation. 
  - I.E. if two players are standing in front of him and he uses acheron kick twice, the second one should go off even though the player got pushed out of range
  - I fixed this in WingsXI with a very hacky change to `targetfind.cpp` (allowed aoe mobskills to target the mob itself) that I'd rather not repeat unless 
    - it's generalized to not specifically only work for DI 
    - and fixes the animations to not show on the mob itself (really silly to see that)
  - It's fixed in this PR by making the db range for his aoe skills 35 and setting the message to NONE if the target is out of range
    - this is a bit of trickery I've done to make the fight work _mechanically_ closer to retail, by not letting a player cause the mob to cancel his ability if the primary target gets too far
- His mobskills shouldn't show an animation on the target if they aren't in range 
  - This may be fixable by setting the animation id in the same section that we adjust knockback, but I imagine that is not the best way to accomplish any of this. I am striving for fight difficulty/accuracy first and foremost, the animation without a combat message is minor imo.
- He should take reduced damage from the front (or increased from the back would work too). I attempted a few methods of using takedmg listeners but couldn't come up with something clean

Overall the fight feels very similar to WingsXI, which has had DI roaming around for close to a year now.

Previous PR on DI spawn/spook mechanics outlines how to get him to spawn, but the short version is
```
!zone 81
!exec SetServerVariable('DarkIxion_ZoneID', 81)
!spawnmob 17109367
!gotoid 17109367
<target DI and quickly run !stun>
<repeat these 3 until name turns red (you must land the hit)>
   !additem 19210
   <equip the ash>
   /ra <t>
<test fight>
```

The fight is fairly straightforward I think, there are multiple failsafes to avoid DI getting stuck in any particular phase/form

### DIFF
--- a/scripts/actions/mobskills/acheron_kick.lua
+++ b/scripts/actions/mobskills/acheron_kick.lua
@@ -1,0 +1,48 @@
+-----------------------------------
+--  Acheron Kick
+--
+--  Description: Physical Cone Attack damage behind user.
+--  Type: Physical
+--  Utsusemi/Blink absorb: 2-3 shadows
+--  Range: Back
+--  Dark Ixion CAN turn around to use this move on anyone with hate, regardless of their original position or even distance.
+-----------------------------------
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    -- to allow primary player to run out of conal and to respect the real range of the mobskill
+    if
+        not target:isBehind(mob, 64) or
+        target:checkDistance(mob) > 10
+    then
+        skill:setMsg(xi.msg.basic.SKILL_MISS)
+        return
+    end
+
+    local numhits = math.random(2, 3)
+    local accmod = 1
+    local dmgmod = 1
+
+    local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod)
+
+    local dmg = 0
+    if info.hitslanded > 0 then
+        dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.PHYSICAL, xi.damageType.SLASHING, info.hitslanded)
+    else
+        skill:setMsg(xi.msg.basic.SKILL_MISS)
+        return
+    end
+
+    if skill:getMsg() ~= xi.msg.basic.SHADOW_ABSORB then
+        target:takeDamage(dmg, mob, xi.attackType.PHYSICAL, xi.damageType.SLASHING)
+        skill:setMsg(xi.msg.basic.DAMAGE)
+    end
+
+    return dmg
+end
+
+return mobskillObject
--- a/scripts/globals/dark_ixion.lua
+++ b/scripts/globals/dark_ixion.lua
@@ -4,7 +4,7 @@
 -- https://ffxiclopedia.fandom.com/wiki/Dark_Ixion
 --[[
     Find DI by:
-    - checking the server's zone var: "!checkvar server DarkIxion_ZoneID"
+    - checking the server's zone var: '!checkvar server DarkIxion_ZoneID'
     - go to that zone: !zone <zoneID>
     - go to DI: !gotoname Dark_Ixion
 
@@ -267,6 +267,26 @@ xi.darkixion.setupEntity = function(entity)
     entity.onMobDisengage = function(mob)
         xi.darkixion.onMobDisengage(mob)
     end
+
+    entity.onCriticalHit = function(mob, attacker)
+        xi.darkixion.onCriticalHit(mob, attacker)
+    end
+
+    entity.onWeaponskillHit = function(mob, attacker, weaponskill)
+        xi.darkixion.onWeaponskillHit(mob, attacker, weaponskill)
+    end
+
+    entity.onMobWeaponSkillPrepare = function(mob, target)
+        xi.darkixion.onMobWeaponSkillPrepare(mob, target)
+    end
+
+    entity.onMobWeaponSkill = function(mob, target, skill)
+        xi.darkixion.onMobWeaponSkill(mob, target, skill)
+    end
+
+    entity.onMobFight = function(mob, target)
+        xi.darkixion.onMobFight(mob, target)
+    end
 end
 
 xi.darkixion.repop = function(mob)
@@ -309,8 +329,8 @@ xi.darkixion.roamingMods = function(mob)
     end
 
     mob:setMobSkillAttack(39)
-    mob:setLocalVar('charging', 0)
-    mob:setLocalVar('double', 0)
+    mob:setLocalVar('trampling', 0)
+    mob:setLocalVar('doubleUpWS', 0)
     mob:setLocalVar('lastHit', 0)
     mob:setBehaviour(0)
     mob:setAutoAttackEnabled(true)
@@ -370,11 +390,16 @@ xi.darkixion.zoneOnGameHour = function(zone)
     end
 end
 
-xi.darkixion.onMobDeath = function(mob, player, isKiller)
-    player:addTitle(xi.title.IXION_HORNBREAKER)
-    -- only reset hp after being killed
-    SetServerVariable('DarkIxion_HP', 0)
-    SetServerVariable('DarkIxion_HornStatus', 0)
+xi.darkixion.onMobDeath = function(mob, player, optParams)
+    if optParams.isKiller or optParams.noKiller then
+        -- only reset hp after being killed
+        SetServerVariable('DarkIxion_HP', 0)
+        SetServerVariable('DarkIxion_HornStatus', 0)
+    end
+
+    if player then
+        player:addTitle(xi.title.IXION_HORNBREAKER)
+    end
 end
 
 xi.darkixion.onMobDespawn = function(mob)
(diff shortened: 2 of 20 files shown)


scripts/actions/mobskills/acheron_kick.lua
```diff
@@ -0,0 +1,55 @@
...
+--  Range: Back
+--  Dark Ixion CAN turn around to use this move on anyone with hate, regardless of their original position or even distance.
+-----------------------------------
+local mobskillObject = {}
+
+local conalArc = 64
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    if not target:isBehind(mob, conalArc) then
+        print(target:getName())
```
zach2good: Leftover debug print

scripts/actions/mobskills/acheron_kick.lua
```diff
@@ -0,0 +1,55 @@
...
+--  Utsusemi/Blink absorb: 2-3 shadows
+--  Range: Back
+--  Dark Ixion CAN turn around to use this move on anyone with hate, regardless of their original position or even distance.
+-----------------------------------
+local mobskillObject = {}
+
+local conalArc = 64
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    if not target:isBehind(mob, conalArc) then
```
zach2good: `return target:isBehind(mob, conalArc) and 1 or 0`

scripts/globals/additional_effects.lua
```diff
@@ -285,6 +286,14 @@ xi.additionalEffect.attack = function(attacker, defender, baseAttackDamage, item
             msgParam = xi.effect.KO
             defender:setHP(0)
         end
+
+    elseif addType == procType.LOCAL_VAR then
+        if
+            item:getID() == 19210 -- Pinch Of Stygian Ash
```
zach2good: Can you add this to items.lua please?

scripts/globals/additional_effects.lua
```diff
@@ -130,6 +130,7 @@ xi.additionalEffect.attack = function(attacker, defender, baseAttackDamage, item
         ABSORB_STATUS = 11,
         SELF_BUFF     = 12,
         DEATH         = 13,
+        LOCAL_VAR     = 14, -- arbitrary logic to apply a local var to the defender
```
zach2good: Summoning @TeoTwawki for AE changes, I don't know how they work or if this is OK etc.

scripts/globals/additional_effects.lua
```diff
@@ -285,6 +286,14 @@ xi.additionalEffect.attack = function(attacker, defender, baseAttackDamage, item
             msgParam = xi.effect.KO
             defender:setHP(0)
         end
+
+    elseif addType == procType.LOCAL_VAR then
```
zach2good: This looks like it's adding a new proc type, but it's entirely hard-coded for Pinch Of Stygian Ash, this is probably not a good pattern

scripts/globals/dark_ixion.lua
```diff
@@ -396,6 +416,101 @@ xi.darkixion.onMobSpawn = function(mob)
...
+    -- preserve tp when doing melee swings, due to SetMobSkillAttack
+    local tpPerHit = 64 -- give tp for the auto attack (amount confirmed by not replacing autos with mobskill list)
+    mob:setLocalVar('skill_tp', mob:getTP() + tpPerHit)
+end
+
+xi.darkixion.onMobWeaponSkill = function(target, mob, skill)
+    -- skill chosen, tp already wiped
+    local skillID = skill:getID()
+    if
+        skillID == 2341 or
```
zach2good, line 465: Can you give these comments or extract them into local vars that say the name please?

scripts/globals/dark_ixion.lua
```diff
@@ -462,3 +577,301 @@ xi.darkixion.onMobDisengage = function(mob)
...
+        ((mob:getTP() >= 2900 and mob:getHPP() > 66) or
+        (mob:getTP() >= 1900 and mob:getHPP() > 33) or
+        (mob:getTP() >= 900 and mob:getHPP() > 0)) and
+        mob:getLocalVar('timeSinceWS') < os.time() - 3 and
+        mob:checkDistance(target) < 15 and
+        mob:getLocalVar('charging') == 0 and
+        mob:actionQueueEmpty()
+    then
+        mob:setLocalVar('timeToWS', 1)
+    end
```
zach2good: At a glance, this condition is awful, can you find a way to break it up?

sql/mob_droplist.sql
```diff
@@ -26775,6 +26775,13 @@ INSERT INTO `mob_droplist` VALUES (3284,0,0,1000,19120,@UNCOMMON); -- Avis (Unco
 -- ZoneID:  88 - Olgoi-Khorkhoi
 INSERT INTO `mob_droplist` VALUES (3285,0,0,1000,18768,@COMMON); -- Poppet Katars (Uncommon, 15%)
 
+-- ZoneID: 81, 82, 84, 89, 91, 95, 96 - Dark Ixion
```
zach2good: I don't know if this is acceptable, or if this messes with the auto-generated formatting of this file? @claywar ?

src/map/ai/helpers/targetfind.cpp
```diff
@@ -197,7 +197,7 @@ void CTargetFind::findWithinCone(CBattleEntity* PTarget, float distance, float a
 
     // Confirmation on the center of cones is still needed for mob skills; player skills seem to be facing angle
     // uint8 angleToTarget = worldAngle(m_PBattleEntity->loc.p, PTarget->loc.p);
-    uint8 angleToTarget = m_APoint->rotation;
+    uint8 angleToTarget = relativeAngle(m_APoint->rotation, 128 * (aoeType == 8)); // adds 180 degree rotation if rear type
```
zach2good, line 200: 👍

src/map/lua/lua_baseentity.cpp
```diff
@@ -9894,6 +9894,38 @@ void CLuaBaseEntity::recalculateAbilitiesTable()
...
+ *  Notes   : if the passed argument is nil or 0, just returns all players in the zone
+ ************************************************************************/
+
+sol::table CLuaBaseEntity::getPlayersInRange(uint32 dist = 0)
+{
+    auto players = lua.create_table();
+
+    uint32 size = 0;
+    // clang-format off
+    zoneutils::GetZone(m_PBaseEntity->getZone())->ForEachChar([&](CCharEntity* PChar) {
```
zach2good: Format this as allman braces please

src/map/lua/lua_baseentity.cpp
```diff
@@ -9894,6 +9894,38 @@ void CLuaBaseEntity::recalculateAbilitiesTable()
...
+        {
+            players.add(CLuaBaseEntity(PChar));
+            size++;
+        }
+    });
+    // clang-format on
+
+    if (!size)
+    {
+        return sol::lua_nil;
```
zach2good: I would still return an empty table here instead of nil, then you can have loops that work without having to add an additional case instead of having to explicitly nil check

scripts/globals/dark_ixion.lua
```diff
@@ -462,3 +577,301 @@ xi.darkixion.onMobDisengage = function(mob)
...
+            mob:setTP(0)
+            mob:setBehaviour(xi.behavior.NO_TURN + xi.behavior.STANDBACK)
+
+            -- re-random lightning spear
+            local xM = mob:getXPos()
+            local zM = mob:getZPos()
+            local yM = mob:getYPos()
+            local xD = math.random(-4, 4)
+            local zD = math.random(-4, 4)
+            local away = { x = xM - xD, y = yM, z = zM - zD }
```
zach2good, line 691: You have this block as a similar looking pattern in a few places, break it out to a support function at the top of the file

zach2good (commented):
Looking good so far

Lots of very short an undescriptive local var names and vars in general, can you try and make things a bit clearer please?

Things should be styled the same: `skill_tp` vs `timeSinceWS`, etc.

Good: `timeSinceWS` etc.
Bad: `run`? `double`? `zap`?

scripts/globals/additional_effects.lua
TeoTwawki: This was a pattern we decided was bad previously on Brigandish Blade and on the 4 winds for Gods. Better to do via a listener instead, even if that new listener itself connects with additional effect and not simply on hit I think that's the way to go for these things.

MowFord: can you point to an example of using a listener for this? Would it be something like adding a ranged attack listener if the player equips a stygian ash?

scripts/globals/dark_ixion.lua
MowFord: they are distinct, but I should be able to make a helper function that takes a parameter I suppose.

scripts/globals/additional_effects.lua
TeoTwawki: @MowFord its going to need a whole new one, not an existing one. Not sure if we'd want it on the mob or on the item, either way one will need to check for the other. since there are technically multiple ixions (diff ID per zone) but only one item, probably on the mob checking that this was the item that hit it rather than the item having the item check if it was hitting "an ixion".

scripts/globals/additional_effects.lua
MowFord: it's hardcoded to apply this particular LocalVar for stygian ash but easily extensible if we had the need for some other AE that applied a localvar on the target.

scripts/globals/additional_effects.lua
MowFord: for clarity, we only need to set the localvar to claim the mob, so seems like it would make more sense to do it for the person wielding the (consumable) stygian ash.

Looks like chaosbringer has an example of adding a listener on equip, maybe that's a fine method?

TeoTwawki: oh, I was thinking "hit by ash" = set claim. multi tasking at my dayjob atm

> Looks like chaosbringer has an example of adding a listener on equip, maybe that's a fine method?

yes, thats fine

scripts/globals/dark_ixion.lua
MowFord: I'm essentially reinventing the core logic that decides if a mob should use a mobskill. I could definitely collapse the top three into some continuous function, but imo the rest are needed.

src/map/lua/lua_baseentity.cpp
MowFord: looking for an example of this type of function call and they all have `claing-format off` <.<

I'll do allman and see how CI likes it

zach2good: They're all surrounded with the clang-format exception because clang-format straight up can't handle making lambda functions not look like ass 😢

MowFord:
What a wild ride. I've got:
- RANGED_STATE_EXIT listener along with a new endpoint for actions to get the action message for a target
- adjusted a bunch of logic on the mobskills to clean that up
- renamed many DI localvars as well as added comments

sql/mob_skill_lists.sql
```diff
@@ -125,13 +125,8 @@ INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ifrit',38,844);
...
 INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ifrit',38,846);
 INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ifrit',38,847);
--- INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ixion',39,1542);
-INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ixion',39,2334);
-INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ixion',39,2335);
-INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ixion',39,2336);
-INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ixion',39,2337);
-INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ixion',39,2338);
-INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ixion',39,2369);
+INSERT INTO `mob_skill_lists` VALUES ('Avatar-Ixion',39,2341); -- di hoof attack
```
TeoTwawki: looks like these 2 IDs are special melee, which you are correct to separate from regular tp-skills but should be named in a way that identifies the skill list is expressly for that purpose (see other special melee skill lists such as trust Shantotto II or August, or Wyrms in Flight mode).

So you'd wind up with 2 lists named like (just example) : 
1. "Dark_Ixion"
2. "Dark_Ixion_Melee"

MowFord: Fair enough. I actually don't use a separate skill list. The logic for which skill to use is done in the mob lua

scripts/globals/additional_effects.lua
MowFord: Was not as simple as I hoped, but I got it done. The standard order of listeners is to run the state end after at other logic, but doing that meant it was happening after the ash was consumed/unequipped.

All working now

MowFord:
Well decided to check on the PR and noticed it had a merge conflict :(

probably didn't do it the best way but i branched this `ixion_fight` branch on my local, undid all 4 commits, did a `git rebase base`, then cherry-picked the commits, fixing the conflict along the way. Any advice on a better approach is appreciated

MowFord:
Perhaps I should break out `.sql` changes into another PR for this and PW?

MowFord:
went ahead and broke up this PR into the required sql changes here: https://github.com/LandSandBoat/server/pull/4706

sql/mob_droplist.sql
MowFord: Xaver informed me that I should just leave no comment, but i've since moved the sql [to a smaller PR](https://github.com/LandSandBoat/server/pull/4706)

PR #4619, opened 2023-10-18
Add mod for "Enhances effect of Spikes spells"

overefined:
**_I affirm:_**

Adds the mod SPIKES_BONUS which implements "Enhances effect of Spikes spells" found on Duelist's Tights. I used the JP wiki for the source.
https://wiki.ffo.jp/html/4278.html claims +20% bonus 
https://wiki.ffo.jp/html/26042.html claims 30% bonus

1. !additem 15121
2. let a mob attack you with the different spikes effects on

### DIFF
--- a/src/map/modifier.h
+++ b/src/map/modifier.h
@@ -669,12 +669,13 @@ enum class Mod
     LIFE_CYCLE_EFFECT    = 1029, // Adds bonus HP% returned to the luopan when using Life Cycle
     AURA_SIZE            = 1030, // Used to extend aura size, the formula is 6.25 + (PEntity->getMod(Mod::AURA_SIZE) / 100) so adding 100 will make this 7.25
 
-    ENSPELL           = 341, // stores the type of enspell active (0 if nothing)
-    ENSPELL_DMG       = 343, // stores the base damage of the enspell before reductions
-    ENSPELL_DMG_BONUS = 432, //
-    ENSPELL_CHANCE    = 856, // Chance of enspell activating (0 = 100%, 10 = 10%, 30 = 30%, ...)
-    SPIKES            = 342, // store the type of spike spell active (0 if nothing)
-    SPIKES_DMG        = 344, // stores the base damage of the spikes before reductions
+    ENSPELL           = 341,  // stores the type of enspell active (0 if nothing)
+    ENSPELL_DMG       = 343,  // stores the base damage of the enspell before reductions
+    ENSPELL_DMG_BONUS = 432,  //
+    ENSPELL_CHANCE    = 856,  // Chance of enspell activating (0 = 100%, 10 = 10%, 30 = 30%, ...)
+    SPIKES            = 342,  // store the type of spike spell active (0 if nothing)
+    SPIKES_DMG        = 344,  // stores the base damage of the spikes before reductions
+    SPIKES_DMG_BONUS  = 1079, // Increases Blaze/Ice/Shock spikes damage by percentage (e.g. mod value 50 = +50% spikes damage)
 
     TP_BONUS    = 345, //
     SAVETP      = 880, // SAVETP Effect for Miser's Roll / ATMA / Hagakure.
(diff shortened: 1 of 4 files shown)


WinterSolstice8:
There's no source on the bg wiki page, and JP wiki claims +20% bonus https://wiki.ffo.jp/html/4278.html
https://wiki.ffo.jp/html/26042.html claims 30% bonus
image
shows +30
JP wiki seems like it has more basis in reality

overefined:
> There's no source on the bg wiki page, and JP wiki claims +20% bonus https://wiki.ffo.jp/html/4278.html https://wiki.ffo.jp/html/26042.html claims 30% bonus image shows +30 JP wiki seems like it has more basis in reality

Updated, thanks for the links!

src/map/modifier.h
```diff
@@ -669,12 +669,13 @@ enum class Mod
...
-    ENSPELL_CHANCE    = 856, // Chance of enspell activating (0 = 100%, 10 = 10%, 30 = 30%, ...)
-    SPIKES            = 342, // store the type of spike spell active (0 if nothing)
-    SPIKES_DMG        = 344, // stores the base damage of the spikes before reductions
+    ENSPELL           = 341,  // stores the type of enspell active (0 if nothing)
+    ENSPELL_DMG       = 343,  // stores the base damage of the enspell before reductions
+    ENSPELL_DMG_BONUS = 432,  //
+    ENSPELL_CHANCE    = 856,  // Chance of enspell activating (0 = 100%, 10 = 10%, 30 = 30%, ...)
+    SPIKES            = 342,  // store the type of spike spell active (0 if nothing)
+    SPIKES_DMG        = 344,  // stores the base damage of the spikes before reductions
+    SPIKES_BONUS      = 1079, // Enhances effect of "Spikes" spells
```
TeoTwawki: Lets follow naming convention for the enspell type above: `SPIKES_DM_BONUS` if its purely a damage increase. The comment should also explain how its used, a percent or additive, how many digits of precision ie is 20 making a 2% or a 20% increase, and so on. Basically I am asking for you to be more verbose here because you'd be shocked how many ppl in the future will not find the line of code will you did this in let alone understand C++ code over in battleutils.

src/map/utils/battleutils.cpp
```diff
@@ -748,6 +748,16 @@ namespace battleutils
         ELEMENT spikeElement = (ELEMENT)((uint8)GetSpikesDamageType(Action->spikesEffect) - (uint8)DAMAGE_TYPE::ELEMENTAL);
         int32   damage       = Action->spikesParam;
 
+        if (PDefender->getMod(Mod::SPIKES_BONUS) > 0)
```
WinterSolstice8: `Mod::SPIKES_DMG_BONUS` then it should compile

src/map/utils/battleutils.cpp
```diff
@@ -748,6 +748,16 @@ namespace battleutils
...
+        {
+            if (static_cast<SPIKES>(Action->spikesEffect) == SPIKES::SPIKE_BLAZE ||
+                static_cast<SPIKES>(Action->spikesEffect) == SPIKES::SPIKE_ICE ||
+                static_cast<SPIKES>(Action->spikesEffect) == SPIKES::SPIKE_SHOCK)
```
WinterSolstice8: Did we implement Deluge Spikes yet? It's castable from some Porrogo cosmetic +1.

overefined: should it just check if not dread spikes?

WinterSolstice8: RDM can't cast dread spikes and the only way to really know would be to use Contagion Transfer from that one shield... maybe we don't need a check at all?

TeoTwawki (approved):
lgtm

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #4623, opened 2023-10-20
Use Task System to schedule Conquest tasks

xkoredev:
**_I affirm:_**

- Conquest system no longer running two separate instances
- Use task system to schedule conquest system tasks

Run conquest updates. Make sure conquest map is consistent. Check all zones and make sure there are no crashes.


src/world/conquest_system.cpp
```diff
@@ -23,36 +23,50 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
+ * - All public methods that may modify the database are enqueued in the task system.
+ * - The task system processes tasks in its own thread
+ * - Private methods are not guarded via the task system, but are only called from
+ *   public methods, so we can assume that they are always called from the task system.
+ */
 ConquestSystem::ConquestSystem()
-: sql(std::make_unique<SqlConnection>())
 {
+    submit([this]()
+           { sql = std::make_unique<SqlConnection>(); });
```
zach2good, line 39: I don't think task_system is going to work for your use case here if you have to submit a task to configure it - I think it's too simple of a building block to achieve what you want.

zach2good: What _might_ work here, is if you make IMessageHandler also inherit from task_system, giving you access to the guts and the construction of everything. This would allow you to lazily set up the sql handle upon submitting the first task, or on construction, or whatever.

src/world/conquest_system.cpp
```diff
@@ -62,8 +76,11 @@ bool ConquestSystem::handleMessage(std::vector<uint8> payload,
         uint64 ipp = from_addr.s_addr;
         ipp |= (((uint64)from_port) << 32);
 
-        // Send influence data to the requesting map server
-        sendInfluencesMsg(true, ipp);
+        submit([this, ipp]()
+               {
```
zach2good, line 80: I appreciate that clang-format probably did this for you, and that this is a more common layout in the Mac dev space, but we tend to wrap all lambda's in `clang-format off` to avoid this broken indentation: https://github.com/LandSandBoat/server/wiki/Development-Guide#Lambdas

src/world/conquest_system.h
```diff
@@ -38,12 +38,10 @@ class ConquestSystem : public IMessageHandler
...
     /**
      * IMessageHandler implementation. Used to handle messages from message_server.
-     * NOTE: The copy of payload here is intentional, since these systems will eventually
-     *     : be moved to their own threads.
      */
-    bool handleMessage(std::vector<uint8> payload,
-                       in_addr            from_addr,
-                       uint16             from_port) override;
+    bool handleMessage(const std::vector<uint8>& payload,
+                       in_addr                   from_addr,
```
zach2good, line 43: Most of the rest of the code base (aside from `const char*` - bane of my life) is `type const&`

src/world/message_handler.h
```diff
@@ -23,18 +23,56 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
+ * the submit() method.
+ *
+ * This is done to ensure thread-safety outside of the message_server thread.
+ */
 class IMessageHandler
 {
 public:
+    IMessageHandler()
+    {
+        ts = new ts::task_system(1);
```
zach2good, line 47: `std::make_unique`

src/world/message_handler.h
```diff
@@ -23,18 +23,56 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
+    }
+
     virtual ~IMessageHandler()
     {
+        delete ts;
```
zach2good, line 52: If you use smart pointers, there's no need for this

src/world/message_handler.h
```diff
@@ -23,18 +23,56 @@ along with this program.  If not, see http://www.gnu.org/licenses/
...
-                               uint16             from_port) = 0;
+    virtual bool handleMessage(const std::vector<uint8>& payload,
+                               in_addr                   from_addr,
+                               uint16                    from_port) = 0;
+
+    /**
+     * Submits work to be executed asynchronously by the task system.
+     * This work will happen serialy, on a single thread.
+     */
+    void submit(std::function<void()> func)
```
zach2good, line 66: If you don't remember to use `std::move` when calling this, this will create copies of the func on the way in. `void submit(std::function<void()>&& func)` will force the callee to use `std::move`

src/world/message_server.cpp
```diff
@@ -43,18 +44,20 @@ struct zone_settings_t
     uint32 misc   = 0;
 };
 
+using InternalHandler = std::function<void(std::vector<uint8>&&, in_addr, uint16)>;
```
zach2good, line 47: This is a different signature to whats used inside message handler, should they be the same?

src/world/message_server.cpp
```diff
@@ -359,15 +362,18 @@ void cache_zone_settings()
     std::copy(yellMapEndpointSet.begin(), yellMapEndpointSet.end(), std::back_inserter(yellMapEndpoints));
 }
 
-void message_server_init(const bool& requestExit)
+void message_server_init(WorldServer* worldServer, const bool& requestExit)
```
zach2good, line 365: Could this be a const& to World Server?

src/world/message_handler.h
zach2good: But as I mentioned above, I'd look into making this class: `class IMessageHandler : public ts::task_system`

zach2good: You could also make it inherit from `SqlConnection` 👀

src/world/conquest_system.cpp
zach2good: and just because this was the review point that made me remember the term I've been trying to remember for _WEEKS_: https://en.wikipedia.org/wiki/Convention_over_configuration

zach2good:
You can also edit Async to look like this:
```cpp

#pragma once

#include "sql.h"
#include <task_system.hpp>

#include <functional>
#include <string>

class Async : public ts::task_system
{
public:
    ~Async() = default;

    static Async* getInstance();

    void query(std::string const& query);
    void query(std::function<void(SqlConnection*)> const& func);
    void submit(std::function<void()> const& func);

private:
    Async();
};
```
```cpp

#include "async.h"

#include "sql.h"
#include <task_system.hpp>

namespace
{
    // Allocated on and owned the main thread
    std::unique_ptr<Async> _instance = nullptr;

    // Allocated on and owned by the task_system worker thread
    std::unique_ptr<SqlConnection> _sql = nullptr;
} // namespace

Async::Async()
: ts::task_system(1)
{
}

Async* Async::getInstance()
{
    if (_instance == nullptr)
    {
        _instance = std::make_unique<Async>();
    }
    return _instance.get();
}

void Async::query(std::string const& query)
{
    // clang-format off
    this->schedule([query]()
    {
        TracySetThreadName("Async Worker Thread");
        if (_sql == nullptr)
        {
            _sql = std::make_unique<SqlConnection>();
        }
        _sql->Query(query.c_str());
    });
    // clang-format on
}

// NOTE: Be _very_ careful when defining your sql argument in the passed-in function.
//     : If you define your arg as _sql, but then call sql, it will use the global
//     : SQLConnection, which is on the main thread.
//     : Remember that SQLConnection is NOT THREAD-SAFE!
void Async::query(std::function<void(SqlConnection*)> const& func)
{
    // clang-format off
    this->schedule([func]()
    {
        TracySetThreadName("Async Worker Thread");
        if (_sql == nullptr)
        {
            _sql = std::make_unique<SqlConnection>();
        }
        func(_sql.get());
    });
    // clang-format on
}

void Async::submit(std::function<void()> const& func)
{
    // clang-format off
    this->schedule([func]()
    {
        TracySetThreadName("Async Worker Thread");
        func();
    });
    // clang-format on
}
```

And then use `Async::getInstance()->query()/submit()`, then you won't have to deal with any of this crap on your own. Only downside would be that all systems that call it share the same work queue, but I think that'll be OK?

zach2good:
But you'd want to make ts::task_system inheritance private, so you don't leak it to consumers

src/world/conquest_system.h
xkoredev: I knew it was something like that! And I can never remember the right one... thanks

src/world/message_handler.h
xkoredev: will use a smart pointer. I was following the `Async` example. Thought there was a reason for not using a smart pointer

src/world/message_handler.h
xkoredev: good point thanks

src/world/message_server.cpp
xkoredev: good catch. It's not. I'm not sure how it compiles / works.

src/world/message_server.cpp
xkoredev: 👍

xkoredev:
> But you'd want to make ts::task_system inheritance private, so you don't leak it to consumers

Think ill use the modified async version. The only downside is that single thread for all systems, but in theory that is okay with our current usage. If it does become a problem, couldn't we just create non singleton versions of `Async` per system?

xkoredev:
> > But you'd want to make ts::task_system inheritance private, so you don't leak it to consumers
> 
> Think ill use the modified async version. The only downside is that single thread for all systems, but in theory that is okay with our current usage. If it does become a problem, couldn't we just create non singleton versions of `Async` per system?

hmm I do see that currently chat audits go through this thread. It may be a better idea to have separate Async instances per system

zach2good:
Chat audits are on the map server and this targets world server no?

xkoredev:
> Chat audits are on the map server and this targets world server no?

oooh good point. We can do this then.

zach2good:
I don't think it's worth worrying too much about performance until it becomes an actual problem, everything is lock less and on a different thread, this will hold the workload easily

src/world/conquest_system.cpp
xkoredev: oh ya I had applied linter but did not even see the style.. I actually had it as you describe before. Looks so odd otherwise.

xkoredev:
> You can also edit Async to look like this:
> 
> ```c++
> #pragma once
> 
> #include "sql.h"
> #include <task_system.hpp>
> 
> #include <functional>
> #include <string>
> 
> class Async : public ts::task_system
> {
> public:
>     ~Async() = default;
> 
>     static Async* getInstance();
> 
>     void query(std::string const& query);
>     void query(std::function<void(SqlConnection*)> const& func);
>     void submit(std::function<void()> const& func);
> 
> private:
>     Async();
> };
> ```
> 
> ```c++
> #include "async.h"
> 
> #include "sql.h"
> #include <task_system.hpp>
> 
> namespace
> {
>     // Allocated on and owned the main thread
>     std::unique_ptr<Async> _instance = nullptr;
> 
>     // Allocated on and owned by the task_system worker thread
>     std::unique_ptr<SqlConnection> _sql = nullptr;
> } // namespace
> 
> Async::Async()
> : ts::task_system(1)
> {
> }
> 
> Async* Async::getInstance()
> {
>     if (_instance == nullptr)
>     {
>         _instance = std::make_unique<Async>();
>     }
>     return _instance.get();
> }
> 
> void Async::query(std::string const& query)
> {
>     // clang-format off
>     this->schedule([query]()
>     {
>         TracySetThreadName("Async Worker Thread");
>         if (_sql == nullptr)
>         {
>             _sql = std::make_unique<SqlConnection>();
>         }
>         _sql->Query(query.c_str());
>     });
>     // clang-format on
> }
> 
> // NOTE: Be _very_ careful when defining your sql argument in the passed-in function.
> //     : If you define your arg as _sql, but then call sql, it will use the global
> //     : SQLConnection, which is on the main thread.
> //     : Remember that SQLConnection is NOT THREAD-SAFE!
> void Async::query(std::function<void(SqlConnection*)> const& func)
> {
>     // clang-format off
>     this->schedule([func]()
>     {
>         TracySetThreadName("Async Worker Thread");
>         if (_sql == nullptr)
>         {
>             _sql = std::make_unique<SqlConnection>();
>         }
>         func(_sql.get());
>     });
>     // clang-format on
> }
> 
> void Async::submit(std::function<void()> const& func)
> {
>     // clang-format off
>     this->schedule([func]()
>     {
>         TracySetThreadName("Async Worker Thread");
>         func();
>     });
>     // clang-format on
> }
> ```
> 
> And then use `Async::getInstance()->query()/submit()`, then you won't have to deal with any of this crap on your own. Only downside would be that all systems that call it share the same work queue, but I think that'll be OK?

Compiler is not liking the private Async ctor when calling: std::make_unique<Async>(). Make this is why it wasn't a unique ptr before?

zach2good:
Use the regular boilerplate for a singleton for Async, now that we're going to use it as a singleton:
```cpp
#include <iostream>
#include <string>
#include <memory>

// async.h
class Async
{
public:
    static Async* getInstance();

    // other methods here

private:
    Async() = default;
    ~Async() = default;

    Async(const Async&) = delete;
    Async& operator=(const Async&) = delete;
};

// async.cpp
Async* Async::getInstance()
{
    static Async instance;
    return &instance;
}

int main()
{
    return 0;
}

```

xkoredev:
> Use the regular boilerplate for a singleton for Async, now that we're going to use it as a singleton:
> 
> ```c++
> #include <iostream>
> #include <string>
> #include <memory>
> 
> // async.h
> class Async
> {
> public:
>     static Async* getInstance();
> 
>     // other methods here
> 
> private:
>     Async() = default;
>     ~Async() = default;
> 
>     Async(const Async&) = delete;
>     Async& operator=(const Async&) = delete;
> };
> 
> // async.cpp
> Async* Async::getInstance()
> {
>     static Async instance;
>     return &instance;
> }
> 
> int main()
> {
>     return 0;
> }
> ```

oh... I don't know why I was using std pointer ... mb

xkoredev:
> You can also edit Async to look like this:
> 
> ```c++
> #pragma once
> 
> #include "sql.h"
> #include <task_system.hpp>
> 
> #include <functional>
> #include <string>
> 
> class Async : public ts::task_system
> {
> public:
>     ~Async() = default;
> 
>     static Async* getInstance();
> 
>     void query(std::string const& query);
>     void query(std::function<void(SqlConnection*)> const& func);
>     void submit(std::function<void()> const& func);
> 
> private:
>     Async();
> };
> ```
> 
> ```c++
> #include "async.h"
> 
> #include "sql.h"
> #include <task_system.hpp>
> 
> namespace
> {
>     // Allocated on and owned the main thread
>     std::unique_ptr<Async> _instance = nullptr;
> 
>     // Allocated on and owned by the task_system worker thread
>     std::unique_ptr<SqlConnection> _sql = nullptr;
> } // namespace
> 
> Async::Async()
> : ts::task_system(1)
> {
> }
> 
> Async* Async::getInstance()
> {
>     if (_instance == nullptr)
>     {
>         _instance = std::make_unique<Async>();
>     }
>     return _instance.get();
> }
> 
> void Async::query(std::string const& query)
> {
>     // clang-format off
>     this->schedule([query]()
>     {
>         TracySetThreadName("Async Worker Thread");
>         if (_sql == nullptr)
>         {
>             _sql = std::make_unique<SqlConnection>();
>         }
>         _sql->Query(query.c_str());
>     });
>     // clang-format on
> }
> 
> // NOTE: Be _very_ careful when defining your sql argument in the passed-in function.
> //     : If you define your arg as _sql, but then call sql, it will use the global
> //     : SQLConnection, which is on the main thread.
> //     : Remember that SQLConnection is NOT THREAD-SAFE!
> void Async::query(std::function<void(SqlConnection*)> const& func)
> {
>     // clang-format off
>     this->schedule([func]()
>     {
>         TracySetThreadName("Async Worker Thread");
>         if (_sql == nullptr)
>         {
>             _sql = std::make_unique<SqlConnection>();
>         }
>         func(_sql.get());
>     });
>     // clang-format on
> }
> 
> void Async::submit(std::function<void()> const& func)
> {
>     // clang-format off
>     this->schedule([func]()
>     {
>         TracySetThreadName("Async Worker Thread");
>         func();
>     });
>     // clang-format on
> }
> ```
> 
> And then use `Async::getInstance()->query()/submit()`, then you won't have to deal with any of this crap on your own. Only downside would be that all systems that call it share the same work queue, but I think that'll be OK?

I'm not sure we can just rely on ->query() to run queries. How would you then do something like
`sendInfluencesMsg()` where we obtain data from DB (we would need to fetch data, so query doesn't work as it only updates data) in the same queue as we would perform other operations like `updateVanaWeekly`.

src/world/conquest_system.cpp
xkoredev: So been thinking about this. I wouldn't need to do this if it wasn't for the warning that pops up when checking if sql query is executed in the same thread it's instantiated. 

Because all queries are going to happen in the same thread (the task system thread), it's okay that we instantiate sql in a different thread. Ideally, the check would only happen if two queries are called from different threads. Would you be okay with that change?

xkoredev: Ah I do see that sqlconnection fires a query on instantiation, which makes this change harder

zach2good: Absolutely not. I don't play up my 'trap card' to be cute or funny, I do it because it's a legitimate problem that was very annoying to diagnose and fix. https://github.com/LandSandBoat/server/blob/b61c1498f7c639a06bcb1822eaf97d8c4368b1d1/src/common/sql.h#L30

zach2good: The warning triggers in the exact scenario that results in UB, most often seen as the sql handle getting wiped out - so don't ignore it

zach2good:
I've done the refactoring to make Async a singleton, and it'll be scalable to multiple threads if the workload ever demands it:
https://github.com/LandSandBoat/server/pull/4628

zach2good:
```
I'm not sure we can just rely on ->query() to run queries. How would you then do something like
sendInfluencesMsg() where we obtain data from DB (we would need to fetch data, so query doesn't work as it only updates data) in the same queue as we would perform other operations like updateVanaWeekly.
```

At this point, just abandon multi-threading and revert everything to lockstep ticks on the world server - being fed messages from the ZMQ queue, and pushing messages back onto the ZMQ queue. Once besieged is "done" we can come back and look at whether or not it's worth breaking things out onto individual threads.

zach2good:
Use this PR to fix this: `Conquest system no longer running two separate instances`

If that requires mutexes, so be it. I'll vet it as it happens

zach2good:
I've done a different version of this refactoring here: https://github.com/LandSandBoat/server/pull/4632

This should be sufficient for allowing you to fix those other conquest bugs (if this was needed for that), and for your work on besieged. If things need to be re-threaded later, we'll deal with it later - but I think it's unlikely.

xkoredev:
> I've done a different version of this refactoring here: #4632
> 
> This should be sufficient for allowing you to fix those other conquest bugs (if this was needed for that), and for your work on besieged. If things need to be re-threaded later, we'll deal with it later - but I think it's unlikely.

works for me. will integrate it to ASB once merged

PR #4625, opened 2023-10-20
[Avatar] Implement Cait Sith

MowFord:
**_I affirm:_**

Implements the Cait Sith smn avatar!

~~I'm starting this off as a draft to get feedback on the core changes I've had to make for Level ? Holy, as well as the path I've taken for the pet skills. Specifically, the buff spells don't report in the chat log properly~~

~~I'm running into a weird situation with using `xi.msg.basic.SKILL_GAIN_EFFECT` when these are mobskills. It makes sense, as it's likely reporting the player ability ID as the mob skill ID. If i set the `pet_skill_finish_id` as player JA finish, then the chat log reports properly but the animation is wrong <.<~~

~~I've got the buff abilities to work via `xi.msg.basic.IS_STATUS`, but it isn't the cleanest.~~

~~Note that I've not adjusted the meat of the offensive abilities to actually function, I'm merely making sure I'm on the right track here. It will technically work this way, but I want to make sure it's the way LSB would prefer.~~

Level ? Holy working via clay's adjustments:
This was a level 5, animation was proper and it properly had no effect on the mob that wasn't divisible by 5
image

Onto a structured list of the changes:
- Cait sith has proper name prefix and named properly to be "Cait Sith" instead of "The CaitSith"
  - Done via proper name in `pet_list` and `name_prefix` in `mob_pools` set to 32
- abilities enabled in `abilities.sql` and set to have proper MP cost (and other metadata)
  -  Also fixed a bug with one of fenrir's abilities being grouped in when Cait Sith was the player's pet
- lua files and entries in the new `pet_skills.sql` table created for all blood pacts:
  - Regal Slash (BP:Rage): 3-hit physical
  - Level ? Holy (BP:Rage): aoe magical
    - Rolls a die and does dmg proportional to roll
    - Only does damage if the target's level is divisible by the roll
  - Mewing Lullaby (BP:Ward): AoE lullaby that resets TP on non-NM mobs
  - Eerie Eye (BP:Ward): conal silence/amnesia with appropriate elemental resist check for amnesia, but retail does light check for silence 
  - Reraise II (BP:Ward): single-target 60-minute reraise II buff for any party member
  - Raise II (BP:Ward): single-target raise II for any party member
  - Altana's Favor (BP:Ward): 2-hour ability gives arise to all party members in range (raise 5 and reraise III with infinite duration)
    - couldn't get targetfind to work properly between dead and alive players, so I loop over all party members and raise them manually

This PR does not include the changes from [clay's PR](https://github.com/LandSandBoat/server/pull/4627/files), but was tested using the changes

Testing of Level ? Holy is best done on mobs that range around lvl 60 since that level is divisible by all numbers 1-6.

The others are pretty straightforward I think. Giving your pet `!immortal` makes things easier to test the BP:Rages as well.


MowFord:
The crux of my issue with level ? holy is that there seems to be a fundamental flaw in how pet abilities process. The onabilitycheck is executed on the owner's ability, so you can't edit the animation ID of the petskill. (i.e. like I [do in this mobskill](https://gitlab.com/ffxiwings/wings/-/blob/master/scripts/globals/mobskills/self-destruct_cluster.lua#L46))

WinterSolstice8:
I'm busy with other stuff, I wrote the petskill system and there _should_ be a way to iron out the kinks, i'll try to look into this at some point on the weekend. Thanks for your efforts

MowFord:
> I'm busy with other stuff, I wrote the petskill system and there _should_ be a way to iron out the kinks, i'll try to look into this at some point on the weekend. Thanks for your efforts

Sounds good. I'm going to work on cleaning up the lua I just wanted to get something in front of you so I'm not completely in the dark.

Also just to be clear: this code works on a technical level but
- I'm not a fan of hard coding the animation changes in cpp
- the messages don't work perfectly as it seems the ability id is referencing mob skills instead of pet skills (i.e. if we set reraise_ii to use message id 317, it says cait sith used berserk, because berserk is mob skill id 526

WinterSolstice8:
Battlemod doesn't seem to like Cait Sith (LSB image)
image
testing retail now...

WinterSolstice8:
Cait Sith's name is wrong (missing the space) and the pool/group has "The" flagged, retail msg:

image

Level ? Holy needs to have the primary message set to `skill:setMsg(xi.msg.basic.JA_NO_EFFECT_2)` if it misses per a retail cap

image

WinterSolstice8:
Reraise II message is `JA_GAIN_EFFECT_2       = 316, -- <user> uses <ability>. <target> gains the effect of <effect>.`
(previously undocumented in msg.lua)

WinterSolstice8:
I see how to fix the problem with level ? holy, because the animation (presumably) stores the "roll" for what the level ? holy will do damage to, and rerolling it is not retail behavior.

I think I just need to add an accessor to get the action's first target and from that, you'll be able to get the action's first animation and/or set it appropriately. I'll hopefully get a PR tomorrow for the change.

WinterSolstice8:
Level ? holy script works with some backend changes like this:

```
abilityObject.onPetAbility = function(target, pet, skill, master, action)
    local holyRollOneAnimID = 164
    local primaryTargetID = action:getPrimaryTargetID()

    -- If primary target, roll for power by setting random animation.
    if primaryTargetID == target:getID() then
        action:setAnimation(primaryTargetID, holyRollOneAnimID + math.random(0, 5))
    else
        action:setAnimation(target:getID(), action:getAnimation(primaryTargetID))
    end

    local power = action:getAnimation(target:getID()) - holyRollOneAnimID + 1
    local dMND = math.floor(pet:getStat(xi.mod.MND) - target:getStat(xi.mod.MND))
    local ele = xi.damageType.LIGHT
    local dmg = 0

    local dmgmod = 1
    local basedmg = pet:getMainLvl() * power + (dMND * 1.5)
```

MowFord:
Apologies if it fails sanity checks, have to step away before doing that step, but bps are tested to work

MowFord:
Unfortunately the issue i found with `altana's favor` is also happening for `raise ii`. I thought i had confirmed it works but I cannot get the ability to consider a dead target valid for the ability :(

I diagnosed a good bit and best I can tell the issue is related to `CCharEntity::ValidTarget`. I set the flag 32 for `target_player_dead` but it just doesn't even call the onAbilityCheck function

MowFord:
If it's simpler to get things moving we could disable Raise II in `abilities.sql` to get Cait Sith implemented the rest of the way for now. I know you'd rather not have half-implemented thing, but Cait Sith being unimplemented at all might be a special case?

WinterSolstice8:
might be able to look at it later tonight, stay tuned

WinterSolstice8:
image
soon

WinterSolstice8:
image

WinterSolstice8:
Is there anything else you're missing? Work has been a bit busy.

MowFord:
> Is there anything else you're missing? Work has been a bit busy.

That was everything needed for this PR. I appreciate you. I'll clean up my PR to utilize your changes later this week

Edit: Did a rebase and tested with the changes. All working

src/map/utils/charutils.cpp
```diff
@@ -2915,7 +2915,7 @@ namespace charutils
                     }
                     else if (PetID == PETID_CAIT_SITH)
                     {
-                        if (PAbility->getID() > ABILITY_SOOTHING_RUBY && PAbility->getID() <= ABILITY_MOONLIT_CHARGE)
+                        if (PAbility->getID() > ABILITY_SOOTHING_RUBY && PAbility->getID() < ABILITY_MOONLIT_CHARGE)
```
WinterSolstice8, line 2918: Why was this changed? Was it necessary?

scripts/actions/abilities/pets/mewing_lullaby.lua
```diff
@@ -0,0 +1,43 @@
...
+-- aoe light based sleep and lowers mob TP to zero
+-----------------------------------
+local abilityObject = {}
+
+abilityObject.onAbilityCheck = function(player, target, ability)
+    return 0, 0
+end
+
+abilityObject.onPetAbility = function(target, pet, skill, master, action)
+    local typeEffect = xi.effect.LULLABY
```
Xaver-DaRed: I know it's called lullaby. However, It was proven that lullaby BRD spells actually aply sleep effect, not lullaby effect, as weird as it sounds.

My point? Either we follow suit OR we confirm in retail the actual effect

scripts/actions/spells/summoning/cait_sith.lua
```diff
@@ -0,0 +1,32 @@
+-----------------------------------
+-- Spell: Diabolos
+-- Summons Diabolos to fight by your side
```
Xaver-DaRed: Nitpick: Wrong spell comment and description

src/map/utils/charutils.cpp
MowFord: yes, when cait sith was out, moonlit charge was in the list of usable blood pacts

scripts/actions/spells/summoning/cait_sith.lua
MowFord: Oops, good catch.

Guess you can see what I copied from :)

scripts/actions/abilities/pets/mewing_lullaby.lua
MowFord: I'm not 100% knowledgeable on the specifics, and we could apply sleep effect instead of lullaby, but doesn't the chat log explicitly say `lullaby` instead of `sleep`?

TeoTwawki: We can easily use one in the addStatus and the other in the message return, that's not a hurdle. But if you think it actually uses the other status, that's easy to check with a packet cap.

Xaver-DaRed (approved):

scripts/actions/abilities/pets/mewing_lullaby.lua
MowFord: Xaver DM'd me about it and that is exactly what we updated it to.

scripts/actions/abilities/pets/regal_scratch.lua
```diff
@@ -0,0 +1,30 @@
+-----------------------------------
+-- Regal Scratch
+-- M=3,1
```
WinterSolstice8: what's this `M=3,1`? doesn't really bother me that it's in there i guess

WinterSolstice8 (approved):

scripts/actions/abilities/pets/regal_scratch.lua
MowFord: ha good question, was a direct copy from what we had on wings... i honestly have no idea

PR #4632, opened 2023-10-22
[world] Refactor regionalMsgHandlers handling

zach2good:
**_I affirm:_**

ConquestSystem was being instantiated twice, and to make handling of messages for regional event systems easier, I hooked everything up to a task system task. If we need to put everything on their own threads later down the line, we'll handle it then - not now.

- Set breakpoints
- `!updateconquest 2`
- See the new task get hit, see the handler get looked up, see the message payload get forwarded into the conquest system handler


claywar (approved):

xkoredev:
🎉 thanks. I will cherry pick this to ASB and PR asap

PR #4635, opened 2023-10-24
Deaths while charmed will not count twice

overefined:
**_I affirm:_**

Deaths while player is charmed no longer take twice the experience points. Removes the call to Die() in uncharm()

1. !addeffect 14
2. die to something


WinterSolstice8:
Huh. Legacy kludge?

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #4636, opened 2023-10-24
Fix monstrosity rank display, add additional passage teleport location

claywar:
**_I affirm:_**

* Adds formula for monipulator1 packet to limit between 0..2, and apply correct rank at the appropriate interval
* Adds teleport location for Odyssean Passage to South Gustaberg

* Boundary test infamy: 0, 10000, 10001, 20000, 20001, 50000
* Use Odyssean Passage in Feretory to travel to S. Gustaberg


zach2good (approved):
Eyyy we love maths!

PR #4639, opened 2023-10-24
Add piece of shiromochi

cmastro24:
**_I affirm:_**

Adds piece of shiromochi and the +1 variant to the usable item list and adds the correct stats for the food.

Eat the food, verify that the stats for it are correct with /checkparam <me> as well as your main stats.  Use antacid or let it wear off to make sure the buffs are removed.


TeoTwawki:
don't need to close to fix anything that goes wrong. force pushing over the same branch name would auto-update the pr for you

cmastro24:
Yea sorry, I thought it would be best to start over because in trying to squash it I really messed up the branch

PR #4642, opened 2023-10-26
Weaponskills: fix edge case allowing 9 hits

MowFord:
**_I affirm:_**

The changes in https://github.com/LandSandBoat/server/pull/1355 were useful and numerous, but swept H2H numHits under the rug. H2H weaponskills get the exact number defined in `numHits`, but have a minimum of 2 hits (things like backhand blow, etc). Conversely, dual-wield gives an extra hit to a weaponskill's `numHits`. 

Both scenarios have a hard cap of 8 total swings.

This PR adjusts H2H weaponskills to have sane `numHits` values and updates the `weaponskills.lua` functions to track actual `hitsDone` and handles the extra DW hit in `getMultiAttacks` before doing the clamp on 8 total hits.

The end result solves two issues:
- Clears up confusion of H2H weaponskill `numHits` settings
- Resolves edge case where weaponskills could get 9 hits:
  - H2H or DW weaponskills with specific procs of DA or TA. Screenshot is from before and after code changes printing out the initial hits and the extra hits:
  
image

some clarification on the above: 
the issue fixed here lies in a bit of a kludge/abuse of variables. This PR attempts to fix it by standardizing what is being tracked:
- `hitsDone` is updated to track the _actual_ number of hits attempted
- `numHits` is updated to track the _actual_ total number of hits for the ws (this is now properly clamped at 8 regardless of extra offhand hit or not)

Simplest way to test is to 
- add `printf("%u - %u", calcParams.tpHitsLanded, calcParams.hitsLanded)` before the return in `doPhysicalWeaponSkill` ([right here](https://github.com/LandSandBoat/server/blob/2b45287b702fcf2118d59588216a73267a3118c5/scripts/globals/weaponskills.lua#L766))
- give 50% DA and 50% TA to player
- run many raging fists, stringing pummel, and dancing edge (with dual wield) before and after these changes

Due to the limit being on _swings_ not _landed hits_ you won't always get 8 hits printed, but the point is that this new code tracks explicitly total `hitsDone` (really swing count) and `numHits` including extra swings from dual wield, multi attacks, and hard cap of 8.

### DIFF
--- a/scripts/actions/weaponskills/final_paradise.lua
+++ b/scripts/actions/weaponskills/final_paradise.lua
@@ -13,7 +13,7 @@ local weaponskillObject = {}
 
 weaponskillObject.onUseWeaponSkill = function(player, target, wsID, tp, primary, action, taChar)
     local params = {}
-    params.numHits = 2
+    params.numHits = 1
     params.ftp100 = 1 params.ftp200 = 1.5 params.ftp300 = 2
     params.str_wsc = 0.0 params.dex_wsc = 0.0 params.vit_wsc = 0.6 params.agi_wsc = 0.0 params.int_wsc = 0.0 params.mnd_wsc = 0.0 params.chr_wsc = 0.0
     params.crit100 = 0.0 params.crit200 = 0.0 params.crit300 = 0.0
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -248,9 +248,19 @@ local function getMultiAttacks(attacker, target, wsParams)
         end
     end
 
-    numHits = utils.clamp(numHits + bonusHits, 1, 8)
+    -- Only get an additional offHand hit with H2H if the wsParams.numHits == 1
+    -- "Note that Hand-to-Hand weaponskills' listed hits do include the offhand hit. However, if a weaponskill does not list multiple hits, it still has an additional offhand hit."
+    -- see "hits" column: https://wiki-ffo-jp.translate.goog/html/19049.html?_x_tr_sch=http&_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=it&_x_tr_pto=wapp
+    -- basically h2h always gets 2 "initial" hits, but `numHits` is hard respected for the extra hits
+    -- whereas dual wiedling gets 2 "initial" hits, but `numHits` for e.g. dancing edge is respected as though only one "initial" was performed
+    -- tl;dr: DW gives an extra WS hit everytime, H2H simply ensures every WS gets at least 2 hits (yes backhand blow and dragon kick are both 2-hit ws...)
+    numHits = numHits + bonusHits
+    if attacker:getOffhandDmg() > 0 then
+        numHits = numHits + 1
+    end
 
-    return numHits
+    -- cap total hits at 8
+    return utils.clamp(numHits, 1, 8)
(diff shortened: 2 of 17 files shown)


scripts/globals/weaponskills.lua
```diff
@@ -248,9 +248,19 @@ local function getMultiAttacks(attacker, target, wsParams)
         end
     end
 
-    numHits = utils.clamp(numHits + bonusHits, 1, 8)
```
TeoTwawki, line 251: its not hitting 9 here before your change. show me this mythical 9 hit ws bruh.

MowFord: I should have been more clear in the original post, but I have a screenshot showing a possible 9 hit. Dancing edge while dual welding and proc multi hit in each initial

TeoTwawki: > [10:49 AM]MowFord: it's not clamping the total hits
> [10:49 AM]MowFord: it's clamping hitsDone
> [10:49 AM]MowFord: which isn't accounting for the offhand hit

(from our discord DMs)

RAIST5150: Just to play devil's advocate I guess...

Are we ABSOLUTELY sure extra hit(s)/swing(s) perceived past the WS limit were not a seperate trigger outside of the WS (ie, retaliation, unfortunate timing of a nornal attack round proc, etc.)?

MowFord: Yes because I'm not confirming via tp gains or anything. In my OP you can see I'm confirming via a print in the `weaponskills.lua` function, just before the return

scripts/actions/weaponskills/final_paradise.lua
```diff
@@ -13,7 +13,7 @@ local weaponskillObject = {}
 
 weaponskillObject.onUseWeaponSkill = function(player, target, wsID, tp, primary, action, taChar)
     local params = {}
-    params.numHits = 2
+    params.numHits = 1
```
WinterSolstice8, line 16: https://www.bg-wiki.com/ffxi/Final_Paradise
The description of this WS says its two hits though I can't verify...

WinterSolstice8:
This looks functionally fine but the numhits on the WS is a bit dubious without actually going to check on retail. JP wiki claims things like Shoulder Tackle are 2 hit but that's because the offhand is being taken into account. The same thing with Final Heaven, Final Paradise, etc. They are "one hit" but the offhand gives them the second one. Not all WS are given a stated number of hits (thanks relic WS)

MowFord:
> This looks functionally fine but the numhits on the WS is a bit dubious without actually going to check on retail. JP wiki claims things like Shoulder Tackle are 2 hit but that's because the offhand is being taken into account. The same thing with Final Heaven, Final Paradise, etc. They are "one hit" but the offhand gives them the second one. Not all WS are given a stated number of hits (thanks relic WS)

For sure. The point really is that for h2h, single hit and 2-hit weaponskills are confirmed to have 2 hits total, so it really doesn't matter what we set numhits to. Setting to 1 or 2 to make me intuitive sense is what I went with.

The whole goal of this PR is to make the variables sanely represent things (and patch that mythical 9-hit ws)

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #4645, opened 2023-10-27
[core] Remove usage of "*std::find_if" pattern

zach2good:
**_I affirm:_**

Removes a pattern that immediately tries to dereference a "maybe" value - replaces them with intermediate checks and logging.

### DIFF
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -740,11 +740,19 @@ namespace petutils
         uint32 petID = PPet->m_PetID;
 
         // clang-format off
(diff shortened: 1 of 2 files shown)


zach2good:
Holding until I can do more than basic smoke testing

zach2good:
This code is pretty much equivalent to what was there before. Before, it ran the risk of dereferencing an invalid iterator, now it'll just log. Nothing else has changed. 

Tested with summoning Trusts, summoning an avatar, summoning a luopan.

Aside from summoning a Luopan while I had a summon out, all clear:
```
[11/21/23 11:18:10:370][map][warn] Pet was not null for Testo. (petutils::SpawnPet:1106)
```

claywar (approved):

PR #4646, opened 2023-10-27
add racc modifier for shigeto bow

lapislosh:
**_I affirm:_**

Adds RACC modifier for Shigeto bow for SAM only, which required a new latent effect for main job (previously only subjob-based latents existed)

https://ffxiclopedia.fandom.com/wiki/Shigeto_Bow

!additem 18142
!additem 18143

Change job to rng and then same and compare racc with/without bow. RNG RACC should remained unchanged while SAM RACC should increase.
```
Job	Weapon		RACC / RATK
SAM	Shortbow	50 / 71
SAM	Shigeto		57 / 86
SAM	Shigeto+1	58 / 87
RNG	Shortbow	103 / 67
RNG	Shigeto		103 / 82
RNG	Shigeto+1	103 / 83
```

### DIFF
--- a/src/map/latent_effect.h
+++ b/src/map/latent_effect.h
@@ -91,9 +91,10 @@ enum class LATENT : uint16
     VS_ECOSYSTEM          = 59, // Vs. Specific Ecosystem ID (e.g. Vs. Plantoid: Accuracy+3)
     VS_FAMILY             = 60, // Vs. Specific Family ID (e.g. Vs. Korrigan: Accuracy+3)
     VS_SUPERFAMILY        = 61, // Vs. Specific SuperFamily ID (e.g. Vs. Mandragora: Accuracy+3)
+    MAINJOB               = 62, // mainjob - PARAM: JOBTYPE
 };
 
-#define MAX_LATENTEFFECTID 61
+#define MAX_LATENTEFFECTID 63
 
 /************************************************************************
  *                                                                       *
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -1628,6 +1628,12 @@ INSERT INTO `item_latents` VALUES (17212,26,20,37,3);
 INSERT INTO `item_latents` VALUES (17212,26,20,37,5);
 INSERT INTO `item_latents` VALUES (17212,26,25,37,4);
 
+-- Shigeto Bow
+INSERT INTO `item_latents` VALUES (18142,26,7,62,12);     -- RACC +7 for Samurai main job
+
+-- Shigeto Bow +1
+INSERT INTO `item_latents` VALUES (18143,26,8,62,12);     -- RACC +8 for Samurai main job
+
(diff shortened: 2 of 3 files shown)


TeoTwawki (approved):

EpicTaru (approved):

src/map/latent_effect.h
```diff
@@ -91,6 +91,7 @@ enum class LATENT : uint16
     VS_ECOSYSTEM          = 59, // Vs. Specific Ecosystem ID (e.g. Vs. Plantoid: Accuracy+3)
     VS_FAMILY             = 60, // Vs. Specific Family ID (e.g. Vs. Korrigan: Accuracy+3)
     VS_SUPERFAMILY        = 61, // Vs. Specific SuperFamily ID (e.g. Vs. Mandragora: Accuracy+3)
+    MAINJOB               = 62, // mainjob - PARAM: JOBTYPE
 };
 
 #define MAX_LATENTEFFECTID 61
```
EpicTaru: Doesn't this value also need to be updated?

TeoTwawki: good catch, it does. And looks like it was missed previously or overwritten in prior pr also\*, it should be 63 as of this pr.

\* probably 2 pr's came in at same time and conflicted on this spot

lapislosh: yes and no, I actually started it before I read this comment!
the value doesn't seem to be used anything but should be updated nonetheless

EpicTaru: For clarity's sake, I'm referring to Line 97, doesn't the MAX_LATENTEFFECTID value need updated?

lapislosh: ah you are correct, it starts at 0 and was missed... I'll fix it to 63 then

lapislosh: orrrrrr it technically says "max latent ID" not "count" maybe it is supposed to be 62? maybe I should just delete it

TeoTwawki: the way we use max IDs requires them to be 1+highest used ID.

(this is one reason I hate what we name that and also hate that it exists, and am happy anytime one gets refactored out.. sometimes the alternative to it has other negative impacts tho)

TeoTwawki: If you removed that max id line, the server would implode in odd places. its necessary.

lapislosh: I guess we'll leave it at 63 and hope for the best

TeoTwawki: 63 is correct, so that is the best outcome.

EpicTaru (approved):

lapislosh:
pushed one last time to reduce down to 1 commit

TeoTwawki:
> pushed one last time to reduce down to 1 commit

bless you kind sir

TeoTwawki (approved):

PR #4652, opened 2023-10-30
NM Olgoi-Khorkhoi's first spawn fix 

Aeltorian:
Retains spawntype to 128. First spawn in zone.lua and respawn in it's mob lua. Uses name to get ID instead of calling ID directly. 

**_I affirm:_**

Olgoi is set to 128 for scripted spawn so working with that, I've coded within zone.lua to have a 60-90 minute respawn on initialization of the zone. Instead of calling for a mobID, I have set it to spawn entity by name to save ourselves any future trouble of any ID shifts. The correct respawn is located within it's lua when it despawns. This originally was not spawning so this fixes this. 

Go to North Gustaberg S, up in the northeast corner in I-4, finally see the NM pop, fight it, kill it, watch it repop for up to a couple hours later.

### DIFF
--- a/scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
+++ b/scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
@@ -13,7 +13,7 @@ end
 
 entity.onMobDespawn = function(mob)
     -- Sets to respawn between 90 to 120 minutes
-    UpdateNMSpawnPoint(mob:getID())
+    -- UpdateNMSpawnPoint(mob:getID()) TODO: need rows in nm_spawn_points.sql
     mob:setRespawnTime(math.random(5400, 7200))
 end
 
--- a/scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
+++ b/scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
@@ -13,6 +13,7 @@ end
 
 entity.onMobDespawn = function(mob)
     -- Sets to respawn between 90 to 120 minutes
+    UpdateNMSpawnPoint(mob:getID())
     mob:setRespawnTime(math.random(5400, 7200))
 end
 
--- a/sql/mob_groups.sql
+++ b/sql/mob_groups.sql
@@ -5298,7 +5298,7 @@ INSERT INTO `mob_groups` VALUES (23,6546,88,'Enchanted_Bones_blm',330,1,677,0,0,
 INSERT INTO `mob_groups` VALUES (24,6419,88,'Rock_Eater',330,0,1725,0,0,32,35,0);
 INSERT INTO `mob_groups` VALUES (25,1623,88,'Gloomanita',0,32,1006,0,0,42,53,0);
 INSERT INTO `mob_groups` VALUES (26,6464,88,'Fledermaus',330,2,82,0,0,45,47,0);
-INSERT INTO `mob_groups` VALUES (27,5870,88,'Olgoi-Khorkhoi',0,128,3285,4500,0,53,54,0);
+INSERT INTO `mob_groups` VALUES (27,5870,88,'Olgoi-Khorkhoi',0,0,3285,4500,0,53,54,0);
 INSERT INTO `mob_groups` VALUES (28,6239,88,'Veteran_Quadav',330,0,2578,0,0,59,63,0);
 INSERT INTO `mob_groups` VALUES (29,6237,88,'Greater_Quadav',330,0,1233,0,0,59,64,0);
 INSERT INTO `mob_groups` VALUES (30,6238,88,'Onyx_Quadav',330,0,1861,0,0,59,63,0);
--- a/sql/mob_groups.sql
+++ b/sql/mob_groups.sql
@@ -5298,7 +5298,7 @@ INSERT INTO `mob_groups` VALUES (23,6546,88,'Enchanted_Bones_blm',330,1,677,0,0,
 INSERT INTO `mob_groups` VALUES (24,6419,88,'Rock_Eater',330,0,1725,0,0,32,35,0);
 INSERT INTO `mob_groups` VALUES (25,1623,88,'Gloomanita',0,32,1006,0,0,42,53,0);
 INSERT INTO `mob_groups` VALUES (26,6464,88,'Fledermaus',330,2,82,0,0,45,47,0);
-INSERT INTO `mob_groups` VALUES (27,5870,88,'Olgoi-Khorkhoi',0,0,3285,4500,0,53,54,0);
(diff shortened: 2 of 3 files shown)


sql/mob_groups.sql
```diff
@@ -5298,7 +5298,7 @@ INSERT INTO `mob_groups` VALUES (23,6546,88,'Enchanted_Bones_blm',330,1,677,0,0,
 INSERT INTO `mob_groups` VALUES (24,6419,88,'Rock_Eater',330,0,1725,0,0,32,35,0);
 INSERT INTO `mob_groups` VALUES (25,1623,88,'Gloomanita',0,32,1006,0,0,42,53,0);
 INSERT INTO `mob_groups` VALUES (26,6464,88,'Fledermaus',330,2,82,0,0,45,47,0);
-INSERT INTO `mob_groups` VALUES (27,5870,88,'Olgoi-Khorkhoi',0,128,3285,4500,0,53,54,0);
```
TeoTwawki: it was likely intended to use the zone lua's onZoneItitialize to set up the 1st spawn period, as the spawn time in the group is also zero and it shouldn't already be spawned at server start.

This would be why someone set it to scripted.

scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
```diff
@@ -13,6 +13,7 @@ end
 
 entity.onMobDespawn = function(mob)
     -- Sets to respawn between 90 to 120 minutes
+    UpdateNMSpawnPoint(mob:getID())
```
Comment (TeoTwawki) [inline scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: Does this NM have rows in nm_spawn_points.sql ? if he does nobody ever commented the mobs name. if there are no rows this won't work as expected.

sql/mob_groups.sql
Aeltorian: That'll call for a rewrite of many other NMs that have this coded this way. Can't say all but some random timed spawns that I looked at are coded as spawntype 0 with respawn timer onMobDespawn.

TeoTwawki: uh, no. we've ton set up exactly as I just described. some others have a static initial spawn period that then gets overwritten. Very few if any are zero time and zero spawn type and then also set a random time on death.

scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
Comment (Aeltorian) [reply scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: lmao, he does in mob_spawn_points.sql. Is Olgoi even a NM? hmmm.

Comment (TeoTwawki) [reply scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: those are 2 very different things. all mobs need to be in mob_spawn_points while only mobs that will be using the UpdateSpawnPoint function need to exist in nm_spawn_points - if an Nm doesn't use it the NM doesn't need to go there.

sql/mob_groups.sql
Aeltorian: Does the slot to the left of the spawntype slot, start running on server initialize? It's supposedly the respawn slot. How would I do this onInitialize? Is there a particular mob/nm that has code like what you're asking?

scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
Comment (TeoTwawki) [reply scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: incidentally needing a random spawn point is another sign he should be set up in the zone init - else every 1st spawn is the exact same spot

Comment (Aeltorian) [reply scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: I'll remove this on next commit. It's a worm, it's likely it won't go far anyways

sql/mob_groups.sql
TeoTwawki: there are a **lot** of existing examples:
- https://github.com/LandSandBoat/server/blob/base/scripts/zones/Beaucedine_Glacier/Zone.lua#L10
- https://github.com/LandSandBoat/server/blob/base/scripts/zones/Bhaflau_Thickets/Zone.lua#L13
- https://github.com/LandSandBoat/server/blob/base/scripts/zones/Caedarva_Mire/Zone.lua#L9

scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua
Comment (TeoTwawki) [reply scripts/zones/North_Gustaberg_[S]/mobs/Olgoi-Khorkhoi.lua]: > I'll remove this on next commit. It's a worm, it's likely it won't go far anyways

that's not how we roll here. retail accurate or bust.

TeoTwawki (dismissed):
set spawn point and spawn time at zone init time for 1st spawn, on mob death will handle subsequent repops

spawn point gets `TODO`'d and command commented out if no spawn points were placed in `nm_spawn_ponts.sql`

sql/mob_groups.sql
Aeltorian: Ok, so call it in the zone.lua itself for first spawn. Does that `GetMobByID(ID.mob.AYNU_KAYSEY):setRespawnTime(math.random(900, 10800))` spawn it when the zone is initializes? To me, that just looks to reset the respawn timer. Olgoi doesn't have an entry in nm_spawn_points.sql so having it updatenmspawnpoints won't work. I'm just trying to understand this. Sorry if I'm being stupid

TeoTwawki: it sets the time period. if it didn't have one. it will after that line executes. if it did have a timer that wasn't hit yet, it'll replace it.

TeoTwawki: this also means you can set spawn type 0 and give it a timer in the group, and this will replace it. I just don't want it to be 0 AND ALSO have a zero spawn time. make sense?

Aeltorian: The respawn time is already coded within the onMobDespawn in it's lua. That's why I opted to use 0.

TeoTwawki: > The respawn time is already coded within the onMobDespawn in it's lua. That's why I opted to use 0.

you aren't understanding there - the mob's script doesn't set the initial spawn timing, the sql does. these are not interchangeable.

TeoTwawki: at the time the mobs onMobDeath is read, it was ALREADY initialized. if that weren't they case we'd never need to use the zone to set them up.

Aeltorian:
CI doesn't like capitalized words. Should I change this?

Aeltorian (commented):
this has been done.

TeoTwawki:
> this has been done.

tested in game, is he spawning and respawning for you now?

Aeltorian:
> > this has been done.
> 
> tested in game, is he spawning and respawning for you now?

He is up, wiggling menacingly at me.  It's working.

TeoTwawki (approved):

PR #4653, opened 2023-10-31
Smite job trait should be additive with other attp

MowFord:
**_I affirm:_**

Smite is applying a percentage increase to base attack, then applying the rest of the `ATTP` modifiers. I can’t find evidence that the current code is how it should work. Other `ATTP` adjustments work by stacking onto `ATTP`, and [this thread](https://www.ffxiah.com/forum/topic/52498/smite-atk-bonus-not-calculating-correctly-w-buffs/) seems to corroborate that line of thinking

Before changes

image

Base before attp or smite trait

image

Smite tier iv gives `64/256 = 25%` attack, shown above: `483/387 ~ 1.248` (rounding errors)

image

And after giving `25% ATTP`, we see double-dipping with the attack bonus: `603/387 ~ 1.558 => 55.8%` (since `1.248 * 1.25 = 1.56`) instead of `50%`. 

After compiling changes in this PR:

image

Base is same as before ^^

image

With smite value 64, same as before ^^

image

With additional attp as before, we get only `580`, meaning we have `580 / 387 ~ 1.49` instead of `~ 1.56`

The discrepancy of course would get more extreme the more base attack you have and higher tiers of smite.

Testing outlined above, but confirm ATTP is additive with value of `SMITE` mod (the value is converted to ATTP via dividing by 256) when wielding 2h weapon or h2h.

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -636,6 +636,7 @@ uint16 CBattleEntity::ATT()
     TracyZoneScoped;
     // TODO: consider which weapon!
     int32 ATT    = 8 + m_modStat[Mod::ATT];
+    auto  ATTP   = m_modStat[Mod::ATTP];
     auto* weapon = dynamic_cast<CItemWeapon*>(m_Weapons[SLOT_MAIN]);
     if (weapon && weapon->isTwoHanded())
     {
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):
LGTM

WinterSolstice8 (approved):

WinterSolstice8:
confirmed on retail

PR #4658, opened 2023-11-02
enable/disabling expansions in main.lua enables/disables appropriate zones

Grahf0085:
**_I affirm:_**
- [x ] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x ] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x ] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

This PR makes it so that when you enable an expansion via main.lua then that expansions zones will be available. If you disable an expansion in main.lua than that expansions zones will be disables.  People can leave their IP address alone in the zone_settings.sql (unless they want to disable individual zones).

Disable an expansion and try to zone to one of its zones. You will get a message from the map program stating the expansion is disabled.

### DIFF
--- a/settings/default/main.lua
+++ b/settings/default/main.lua
@@ -25,6 +25,8 @@ xi.settings.main =
     RESTRICT_CONTENT = 0,
 
     -- Enable Expansion (1 = Enabled, 0 = Disabled)
+    ENABLE_VANILLA   = 1;
+    ENABLE_ROZ       = 1;
     ENABLE_COP       = 1,
     ENABLE_TOAU      = 1,
     ENABLE_WOTG      = 1,
--- a/.luarc.json
+++ b/.luarc.json
@@ -1,0 +1,3 @@
+{
+    "workspace.maxPreload": 10000
+}
--- a/sql/zone_settings.sql
+++ b/sql/zone_settings.sql
@@ -1,7 +1,14 @@
+
+-- MariaDB dump 10.19  Distrib 10.11.4-MariaDB, for debian-linux-gnu (x86_64)
+--
+-- Host: localhost    Database: xidb
+-- ------------------------------------------------------
+-- Server version	10.11.4-MariaDB-1~deb12u1
+
 /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
 /*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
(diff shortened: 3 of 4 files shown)


TeoTwawki:
I think this has come up before and not made it into LSB. 

While I'm not _completely_ opposed to this, I don't want to have that decided by the content tag match alone. Make sure we're also obeying the restrict content setting (I figure that was good enough for NPCs so why not here too) and that players properly bounce back into the zone they were coming from.

settings/default/main.lua
```diff
@@ -25,6 +25,8 @@ xi.settings.main =
     RESTRICT_CONTENT = 0,
 
     -- Enable Expansion (1 = Enabled, 0 = Disabled)
+    ENABLE_VANILLA   = 1;
```
zach2good, line 28: I don't think that exposing this is a good idea, why would anyone want to disable the base zones.

Also, seeing as the English client launched with ROZ, it's not super useful to have this additional tag

zach2good (commented):
This doesn't feel like minor and simple customisation, which we're normally happy to have in settings. This would be better placed as a module to disable whatever arbitrary lists of zones you want. If a zone is properly disabled you'll be turned away from it - so the effect is the same without introducing 300 lines of arbitrary changes to the main codebase

TeoTwawki:
I think port being zero/nonezero is a better way to control which zones are enabled than tags anyway

settings/default/main.lua
Grahf0085: It's to stop the IsContentEnabled function from making an error. It has a line where it's going through the settings file and checking it for the content tag passed to it. In this case the content tag is vanilla for many of the zones since they are vanilla zones and then the function produces an error because it can't find vanilla in main.lua

I could just have it return true if it receives a value of vanilla or Roz

Grahf0085:
> I think this has come up before and not made it into LSB.
> 
> While I'm not _completely_ opposed to this, I don't want to have that decided by the content tag match alone. Make sure we're also obeying the restrict content setting (I figure that was good enough for NPCs so why not here too) and that players properly bounce back into the zone they were coming from.

Thanks for the response. The IsContentEnabled function is already checking the restrict content setting. Unless I read it wrong which is possible

Grahf0085:
> I think port being zero/nonezero is a better way to control which zones are enabled than tags anyway

I was just trying to make the settings and main.lua to be accurate

PR #4659, opened 2023-11-02
Replace nil function call for spell:addEffectFlag()

CatsEyeXI:
**_I affirm:_**

Addresses an issue that was raised, where Maiden's Virelai could not be cast. A similar call was also found in death.lua, so I replaced that one as well.
image

image

### DIFF
--- a/scripts/actions/spells/black/death.lua
+++ b/scripts/actions/spells/black/death.lua
@@ -6,7 +6,7 @@
 local spellObject = {}
 
 spellObject.onMagicCastingCheck = function(caster, target, spell)
-    spell:addEffectFlag(xi.magic.spellFlag.IGNORE_SHADOWS)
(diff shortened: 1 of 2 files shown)


claywar (approved):

TeoTwawki:
thats not nil (exactly)
https://github.com/search?q=repo%3ALandSandBoat%2Fserver%20addEffectFlag&type=code

its just not hooked up correctly for use outside of effect scripts

(it was only meant for effects, and these must have gotten caught in find+replace)

TeoTwawki:
its just not hooked up correctly for use outside of effect scripts

TeoTwawki (approved):

TeoTwawki:
at some future point I'm moving spell flags out of this spot (it was a band-aid at the time)

PR #4663, opened 2023-11-02
Fix Fire Resist mod on Flame Shield

CatsEyeXI:
**_I affirm:_**

Corrects the fire resist mod on flame shield.

Equip Item and notice you will gain the correct amount of fire resist.


zach2good (approved):
https://ffxiclopedia.fandom.com/wiki/Flame_Shield

claywar (approved):

PR #4665, opened 2023-11-02
Implement Kupo Rod (Removes Weakness)

CatsEyeXI:
**_I affirm:_**

Makes Kupo Rod usable.
Removes Weakness on use

!addeffect <me> WEAKNESS 0 3600
Use Kupo rod
lose weakness status

### DIFF
--- a/scripts/items/kupo_rod.lua
+++ b/scripts/items/kupo_rod.lua
@@ -1,0 +1,22 @@
+-----------------------------------
+-- ID: 21074
+-- Item: Kupo Rod
+-- Enchantment: Moogle Blessing (removes Weakness)
+-----------------------------------
+local itemObject = {}
+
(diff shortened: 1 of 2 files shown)


scripts/items/kupo_rod.lua
```diff
@@ -0,0 +1,18 @@
+-----------------------------------
+-- ID: 21074
+-- Item: Kupo Rod
+-- Enchantment: Moogle Blessing (removes Weakness)
+-----------------------------------
+local itemObject = {}
+
+itemObject.onItemCheck = function(target)
+    return 0
```
claywar: Can this item be used if the target is not weakened?

CatsEyeXI: ive added a check for this

Xaver-DaRed (approved):

PR #4667, opened 2023-11-02
[lua] Change PromathiaStatus variable value to 0 instead of 1

EpicTaru:
**_I affirm:_**
- [ ] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [ ] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).


EpicTaru:
Stupid rushed brain, did not mean to do THIS PR yet!

PR #4674, opened 2023-11-03
Monk mob weapon damage adjustments to match PC

MowFord:
**_I affirm:_**

Monk mobs hit like a truck because their weapon damage is calculated the same as all other mob types: `mlvl + bonus`

This adjusts them to follow the h2h curve for weapon damage based on skill for that level. I've also added a bit of bonus I had added on WingsXI to simulate a h2h weapon's additional bonus to weapon damage. I assume we'd remove it but better to include at the start for more information.

Before:
image

After:
image

Simple enough, fight various monk mobs and notice they don't Dual Wield 2h weapons with h2h delay anymore.

### DIFF
--- a/src/map/utils/mobutils.cpp
+++ b/src/map/utils/mobutils.cpp
@@ -56,6 +56,7 @@ namespace mobutils
     uint16 GetWeaponDamage(CMobEntity* PMob, uint16 slot)
     {
         uint16 lvl    = PMob->GetMLevel();
+        auto*  weapon = dynamic_cast<CItemWeapon*>(PMob->m_Weapons[SLOT_MAIN]);
         int8   bonus  = 2;
         uint16 damage = 0;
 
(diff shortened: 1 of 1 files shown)


Frankie-hz:
Where did you derive this formula from?

MowFord:
> Where did you derive this formula from?

https://ffxiclopedia.fandom.com/wiki/Category:Hand-to-Hand

it's already in multiple places in the codebase: `m_naturalH2hDamage = (int32)(m_attacker->GetSkill(SKILL_HAND_TO_HAND) * 0.11f) + 3;`

Frankie-hz:
That is for players only.

I have to dig through my DMs for screenshots of damage matching from retail but the general formula should be:

```
there are 2 calcs for h2h base damage for monsters
pre-toau = normal * .45 rounded to integer
toau+ = normal * .625 rounded to integer
```

`mobdamage = (floor((level + 2 + fstr) * .9) / 2) * pdif` (pre-toau)

Edit: It will take me a bit to find the screenshots - ill have to ask Jimmayus if he has any

MowFord:
So even lower than what i'm lowering it by?

Frankie-hz:
Not sure the math on both - would have to run the numbers. That is the total damage at the end of the attack. So this probably belongs in attack.cpp instead of the base damage section. I did this for ASB back in November of last year so excuse me if the coding is a little...not up to par 

https://github.com/AirSkyBoat/AirSkyBoat/pull/598

MowFord:
It absolutely belongs here, as it's defining the damage of the "weapon". Monk mobs hit with both fists, so their weapon damage should reflect that of a h2h weapon. Whatever numbers we figure out will reflect in `attack.cpp` at the end of it

Frankie-hz:
Sounds good.

MowFord:
Thought about it more and I'm not convinced that monk mobs don't follow the same DMG formula as players. If you have end-result numbers that show monk mobs dealing 45% of non-monk DMG pre-toau then 65% after, that could be due to the fstr changes for h2h and the weapon damage was simply using the formula above the whole time.

Unless there's good evidence I'm not sure why we wouldn't go with well established natural h2h DMG calculations

Sidenote: I'm going to push without my extra DMG formula, but it's in the PR history for posterity

MowFord:
Found a small oversight in this logic: there are monk mobs that don't use h2h combat skill.

I've adjusted to match based on the combat skill instead of mob main job

MowFord:
I'll get it right one of these days

PR #4676, opened 2023-11-03
Fixes "Augment Convert" effect on Murgleis. 

Aeltorian:
**_I affirm:_**

Murgleis:
Augments "[Convert](https://www.bg-wiki.com/ffxi/Convert)" Effect gives the following:

- Instead of converting a player's current MP and HP when they use [Convert](https://www.bg-wiki.com/ffxi/Convert), this weapon restores HP by 50% of a player's maximum amount. MP restoration is not changed.
- Does not affect the amount of MP converted from HP.

This corrects and fixes the effect of "[Healing 50% of player's Max HP](https://www.bg-wiki.com/ffxi/Convert)" upon use of Convert with any Murgleis from it's complete [75 version](https://www.bg-wiki.com/ffxi/Murgleis_(Level_75)) up to [i119 III](https://www.bg-wiki.com/ffxi/Murgleis_(Level_119_III)) version.

1. Change to RDM
2. Equip Murgleis lvl 75 or better version.
3. Set your hp to 500 and mp to 1.
4. Use convert.
5. Notice that you heal for half your max HP on top of the 1 MP that was converted over.


Xaver-DaRed (approved):

scripts/enum/item.lua
```diff
@@ -5840,12 +5840,15 @@ xi.item =
     GLEAMING_ZAGHNAL                    = 18963,
     DIRE_SCYTHE                         = 18964,
     TERPSICHORE_75                      = 18989,
+    MURGLEIS_75                         = 18995,
```
claywar, line 5843: Could you also please change the entries in item_basic (just the name, not sortname) to match these enums?  This will be vital to add magian support for upgrades

CatsEyeXI: all set

claywar (approved):

PR #4677, opened 2023-11-04
Improve Messaging on Monk's Mantra Ability

CatsEyeXI:
**_I affirm:_**

Implement Monk's Mantra ability so that it provides an HP boost to party members based on the Monk's merits.

https://www.bg-wiki.com/ffxi/Mantra

Join a party
Use Mantra
Enjoy HP Boost with percentage consistent with Monk's merit points rank


CatsEyeXI:
After realizing the code didn't really change much, I went back and double checked and realized this is, well, an "us" problem. So I won't be offended if you want to close/reject this, Sorry.

Xaver-DaRed:
Won't be closing this for now though. This brings up a few questions of my own. Like, how it interacts with other HP boosts like giants potion, for example.

PR #4678, opened 2023-11-04
Implement NM Flockbock in Lufaise Meadows

CatsEyeXI:
**_I affirm:_**

Implemented logic for NM Flockbock based on captures.

https://www.bg-wiki.com/ffxi/Flockbock
https://www.youtube.com/watch?v=j8fK86BnQbE&ab_channel=RainbowChaser

Mob should be up on server reset.
Fight, notice fight closely resembles that of video.


TeoTwawki:
> Mob should be up on server reset.

Typically NMs are not pre-spawned when server boots. They are usually waiting to spawn, sometimes on a shorter than usual cooldown, sometimes not.

Trivia:
SE leaves the server up for a semi random amount of time before we're able to log in, so mobs that spawn 1hr or less can occasionally be up if they didn't need a PH to die simply because the time already passed. Since this is not consistent, the Nm will not be up and waiting with every server restart. Lotto pops can sometimes be up by pure luck of 1st spawn winning the lotto, if they have no cooldown!

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
```diff
@@ -1,11 +1,45 @@
...
 --   NM: Flockbock
+-- https://www.bg-wiki.com/ffxi/Flockbock
+-- TODO: NM has several possible spawn locations
+-----------------------------------
+mixins = { require('scripts/mixins/rage') }
 -----------------------------------
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:addMod(xi.mod.BINDRES,      100)
```
TeoTwawki: are you trying to make immune, or just highly resistant?

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
```diff
@@ -1,11 +1,45 @@
...
+    mob:addMod(xi.mod.SLEEPRES,     100)
+    mob:addMod(xi.mod.GRAVITYRES,   100)
+    mob:addMod(xi.mod.SILENCERES,   100)
+
+    -- Has enhanced double attack, movement speed, and uses WS more often than normal mobs.
+    mob:setMod(xi.mod.DOUBLE_ATTACK, 50)
+    mob:setMod(xi.mod.MOVE,          25)
```
TeoTwawki: how did you confirm this monsters move speed, and how did you confirm +25 mod move would bring it to that speed?

sql/mob_groups.sql
```diff
@@ -893,7 +893,7 @@ INSERT INTO `mob_groups` VALUES (28,1583,24,'Gigas_Slinger',330,0,992,0,0,35,38,
 INSERT INTO `mob_groups` VALUES (29,4127,24,'Vampire_Bat',330,0,234,0,0,42,44,0);
 INSERT INTO `mob_groups` VALUES (30,4353,24,'Wingrats',330,0,80,0,0,41,44,0);
 INSERT INTO `mob_groups` VALUES (31,20,24,'Abraxas',330,0,2,0,0,85,86,0);
-INSERT INTO `mob_groups` VALUES (32,4689,24,'Flockbock',0,128,0,17000,0,82,82,0);
```
TeoTwawki, line 896: its extremely likely he should be getting a 1 to 2 hr timer set in the zone lua, or a hr timer set here - and then also a 1 to 2 hr in his depsawn script for every subsequent repop regardless of which of those two happens.

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
Xaver-DaRed: Those mods set to 100 will make them effectively immune to their associated status effect.
However, those mods are used for the resist job traits, and shouldnt be used for this. (They trigger the "Resist!" message)

I kind of remember we got functionality to set immunities via script. Key word: __kind__

TeoTwawki: theres a field that takes them as a bitmask in mob_pools table, and the script ability was added so that we coudl both add and remove them because some monsters _change_ their immunities mid fight. I wouldn't mind converting it all to script and eliminating the filed but that'd be a lot of work and might cause some non-nm to have scripts that otherwise wouldn't need them, so for now I still recommend using pools to static immunities.

If he just meant to add really potent resistance tho, then its a matter of is-this-a-trait and is-this-the-right-amount  
(do we clamp before 100% ? coz we prolly should)

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
TeoTwawki: gonna add to this that using `mod_mov` in a mob script is extremely prone to being problematic. Its better to cap the real speed in retail (gonna go look if mob is up right now) and set it as that speed in either its family or if that is not possible, the `setSpeed()` function.

Some mobs also have a "chase speed" which LSB does not yet have a method for. Not all mobs do that.

edit: confirmed base speed 40, jumps to 100 only when agro with target out of range. once target is in melee distance drops back to 40. This can be todo'd till we have a standard method to handle this in a retail accurate manner or you can handle it in a close-enough way using setSpeed with notes at your option.

CatsEyeXI: I had Spicy take a video for me, didn't bother posting it in the captures discord because Silknoz already had one, but I'm happy to provide it if you like ... also happy to use setSpeed instead of the mod.

CatsEyeXI:
not able to squash this further on my end due to merge conflict

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
CatsEyeXI: I've removed these modifiers in favor of adding the immunities directly to mob_pools.

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
```diff
@@ -1,11 +1,50 @@
...
 --   NM: Flockbock
+-- https://www.bg-wiki.com/ffxi/Flockbock
+-- TODO: NM has several possible spawn locations
+-----------------------------------
+mixins = { require('scripts/mixins/rage') }
 -----------------------------------
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    -- Has enhanced double attack, movement speed, and uses WS more often than normal mobs.
```
TeoTwawki: comment about move speed here need removed

ws'ing more often might not mean storetp. was his tp measured during the fight? sorry I can't scroll through video's to see everything that happens

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
```diff
@@ -1,11 +1,50 @@
...
+    mob:addListener('WEAPONSKILL_USE', 'PETRIBREATH_HATE_RESET', function(mobArg, target, wsid, tp, action)
+        if wsid == petribreath then
+            mob:resetEnmity(target)
+        end
+    end)
+
+    mob:setLocalVar('[rage]timer', 3600) -- 60 minutes
+end
+
+-- TODO: Mob's movement speed is increased while chasing target.
```
TeoTwawki, line 31: perfect 👍

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
TeoTwawki: other ways it could go more often: regain, different tp use thresholds, reaction mechanics (unlikely but just listing things here) [savetp](https://www.bg-wiki.com/ffxi/Save_TP), and [conservetp](https://www.bg-wiki.com/ffxi/Conserve_TP)

its is likely storetp or else regain. 

storetp is confirmable by counting up its tp, regain by just holding it without feeding any.

TeoTwawki (approved):
There's a possibility that this tp gain this thing happens isn't stp. I ruled out regain for sure by holding it for a long while while preventing my feeding any tp (shadows stoneskin and dealt zero dmg to it). It took it 11 hits to get tp enough to use a skill, and each time it hit me I gained 87/1000 tp. 

It doesn't have high delay but its tp gains act like it does.. high base delay plus high haste to counter it maybe? needs further research. Because of that, I am calling the current script "good enough" barring any objections from others.

p.s. still will need research on the overall DA rate this should have as well, remembering that mainjob war already adds some. In any case, this mobs modifiers will definitely be getting a future second look.

TeoTwawki:
> not able to squash this further on my end due to merge conflict

should be able to rebase onto latest LSB `base` which will solve that

scripts/zones/Lufaise_Meadows/mobs/Flockbock.lua
```diff
@@ -1,11 +1,51 @@
...
+entity.onMobDisengage = function(mob, target)
+    mob:setSpeed(40)
+end
+
 entity.onMobDeath = function(mob, player, optParams)
     xi.hunts.checkHunt(mob, player, 442)
 end
 
+entity.onMobDespawn = function(mob)
+    UpdateNMSpawnPoint(mob:getID())
```
TeoTwawki: so, this function requires that the mob already have rows in `nm_spawn_points`, It needs multiple rows to do its job correctly. Somewhere in the future that table will get nuked from orbit but thats a long ways off I think.

if we don't have retail captured positions plural, just comment it out with a TODO

TeoTwawki: Also same in the zone lua.

Xaver-DaRed (approved):

TeoTwawki (approved):

PR #4685, opened 2023-11-05
[lua] Add Wall of Banishing open/close animation

WinterSolstice8:
**_I affirm:_**

Add wall of banishing open/close animation

animation 8:
image
animation 9:
image
approximately 20~ seconds

image

open the wall of banishing, see the portal open a bit.

### DIFF
--- a/scripts/zones/Davoi/npcs/_45d.lua
+++ b/scripts/zones/Davoi/npcs/_45d.lua
@@ -25,7 +25,13 @@ end
 entity.onEventFinish = function(player, csid, option, npc)
     if csid == 42 and option == 0 then
         player:messageSpecial(ID.text.POWER_OF_THE_ORB_ALLOW_PASS)
-        npc:openDoor(16)
+        npc:openDoor(20)
+
+        -- Animation for the "door" being open
+        npc:setAnimation(8)
+        npc:timer(1000 * 20, function(npcArg)
+            npcArg:setAnimation(9)
+        end)
     end
 end
 


Xaver-DaRed (approved):

scripts/zones/Davoi/npcs/_45d.lua
```diff
@@ -25,7 +25,13 @@ end
 entity.onEventFinish = function(player, csid, option, npc)
     if csid == 42 and option == 0 then
         player:messageSpecial(ID.text.POWER_OF_THE_ORB_ALLOW_PASS)
-        npc:openDoor(16)
+        npc:openDoor(20)
+
+        -- Animation for the "door" being open
+        npc:setAnimation(8)
+        npc:timer(1000 * 20, function(npcArg)
```
TeoTwawki, line 32: `openDoor()` itself is just doing animation changes, since its normal changes are incorrect here and you are setting the correct ones via `setAnimation()`, there is no more need for the command at all.

```cpp
        m_PBaseEntity->animation = ANIMATION_OPEN_DOOR;
```
There is also an update packet sent, but the same thing happens when we `setAnimation()`

TeoTwawki: oh actually no, scratch that because:
https://github.com/LandSandBoat/server/blob/base/scripts/enum/animation.lua#L16

TeoTwawki: was going to say use the enum, imagine my surprise

PR #4698, opened 2023-11-07
Blu magic core updates

MowFord:
**_I affirm:_**

This ports the SATA code from CEXI for blu magic.

It also enables TA to function for blu magic skillchain damage

End result is (using cannonball as an example)
- a SA cannonball will auto crit
- if taChar is in position
  - a TA cannonball will transfer enmity
  - a TA cannonball that skillchains will transfer enmity of both spell and skillchain

`PHYSICAL_POTENCY` now gives acc and bonus to cratio

Finally, some adjustments were made to support `params.bonusacc` and `params.critchance`. If these params aren't passed both will be zero by default

_Also_ (sorry for the late discovery, but didn't want to make a new PR since this one is still open): Found that the base multiplier is being set to 1 unless Chain Affinity is in effect. This is wrong: https://www.bg-wiki.com/ffxi/Calculating_Blue_Magic_Damage

The base multiplier is in the table as ftp0 and is only applied to the first hit.

`!stun` and `!immortal` a mob
position for SA and compare damage before and after cannonball
position for TA and remove `!stun` after to see the mob target the player
TA for skillchain damage is a bit trickier to test, but you can use `!getenmity` to confirm your enmity doesn't spike after using TA (open the skillchain with the second player)

### DIFF
--- a/scripts/globals/bluemagic.lua
+++ b/scripts/globals/bluemagic.lua
@@ -224,6 +224,10 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
                               spell:isAoE() == 0 and
                               params.attackType ~= xi.attackType.RANGED and
                               (caster:isBehind(target) or caster:hasStatusEffect(xi.effect.HIDE))
+    local trickAttackTarget = (caster:hasStatusEffect(xi.effect.TRICK_ATTACK) and
+                              spell:isAoE() == 0 and
+                              params.attackType ~= xi.attackType.RANGED) and
+                              caster:getTrickAttackChar(target) or nil
 
     while hitsdone < params.numhits do
         local chance = math.random()
@@ -256,12 +260,13 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
 
         if params.attackType ~= xi.attackType.RANGED then
             caster:delStatusEffect(xi.effect.SNEAK_ATTACK)
+            caster:delStatusEffect(xi.effect.TRICK_ATTACK)
         end
 
         hitsdone = hitsdone + 1
     end
 
-    return xi.spells.blue.applySpellDamage(caster, target, spell, finaldmg, params)
+    return xi.spells.blue.applySpellDamage(caster, target, spell, finaldmg, params, trickAttackTarget)
 end
 
 -- Get the damage for a magical Blue Magic spell
@@ -306,7 +311,7 @@ xi.spells.blue.useMagicalSpell = function(caster, target, spell, params)
     -- MAB/MDB/weather/day/affinity/burst effect on damage
     finaldmg = math.floor(addBonuses(caster, spell, target, finaldmg))
 
-    return xi.spells.blue.applySpellDamage(caster, target, spell, finaldmg, params)
+    return xi.spells.blue.applySpellDamage(caster, target, spell, finaldmg, params, nil)
 end
 
 -- Perform a draining magical Blue Magic spell
@@ -333,7 +338,7 @@ xi.spells.blue.useDrainSpell = function(caster, target, spell, params, softCap, 
             caster:addMP(dmg)
         else
             dmg = utils.clamp(dmg, 0, target:getHP())
-            dmg = xi.spells.blue.applySpellDamage(caster, target, spell, dmg, params)
+            dmg = xi.spells.blue.applySpellDamage(caster, target, spell, dmg, params, nil)
             caster:addHP(dmg)
         end
     end
@@ -382,13 +387,13 @@ xi.spells.blue.useBreathSpell = function(caster, target, spell, params, isConal)
     -- Final damage
     dmg = target:breathDmgTaken(dmg)
 
-    results[1] = xi.spells.blue.applySpellDamage(caster, target, spell, dmg, params)
+    results[1] = xi.spells.blue.applySpellDamage(caster, target, spell, dmg, params, nil)
     results[2] = resistance
     return results
 end
 
 -- Apply spell damage
-xi.spells.blue.applySpellDamage = function(caster, target, spell, dmg, params)
+xi.spells.blue.applySpellDamage = function(caster, target, spell, dmg, params, trickAttackTarget)
     if dmg < 0 then
         dmg = 0
     end
@@ -419,7 +424,13 @@ xi.spells.blue.applySpellDamage = function(caster, target, spell, dmg, params)
     dmg = utils.stoneskin(target, dmg)
 
     target:takeSpellDamage(caster, spell, dmg, attackType, damageType)
-    target:updateEnmityFromDamage(caster, dmg)
+    if not target:isPC() then
+        if trickAttackTarget then
+            target:updateEnmityFromDamage(trickAttackTarget, dmg)
+        else
+            target:updateEnmityFromDamage(caster, dmg)
+        end
+    end
     target:handleAfflatusMiseryDamage(dmg)
 
     return dmg
--- a/scripts/globals/bluemagic.lua
+++ b/scripts/globals/bluemagic.lua
@@ -220,10 +220,14 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
     local hitsdone = 0
     local hitslanded = 0
     local finaldmg = 0
+    local sneakIsApplicable = caster:hasStatusEffect(xi.effect.SNEAK_ATTACK) and
+                              spell:isAoE() == 0 and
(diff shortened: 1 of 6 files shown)


MowFord:
This is an initial submission, I'm going to confirm I can do a removal of the `, nil)` addition to a few spells as well as fully test it

scripts/globals/bluemagic.lua
```diff
@@ -220,23 +220,36 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
...
             -- Generate a random pDIF between min and max
             local pdif = math.random(cratio[1] * 1000, cratio[2] * 1000)
             pdif = pdif / 1000
 
             -- Add it to our final damage
-            if hitsdone == 0 then
-                finaldmg = finaldmg + (finalD * (multiplier + correlationMultiplier) * pdif) -- first hit gets full multiplier
+            if hitsdone == 0 then -- first hit gets full multiplier
+                if sneakIsApplicable then
+                    finaldmg = finaldmg + (finalD * (multiplier + correlationMultiplier) * (pdif + 0.7))
```
Xaver-DaRed: finaldmg = finaldmg + finalD * (multiplier + correlationMultiplier) * (pdif + 0.7)

(Unneeded parenthesis)

scripts/globals/bluemagic.lua
```diff
@@ -220,23 +220,36 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
...
+                else
+                    finaldmg = finaldmg + (finalD * (multiplier + correlationMultiplier) * pdif)
```
Xaver-DaRed: finaldmg = finaldmg + finalD * (multiplier + correlationMultiplier) * pdif

(Unneeded parenthesis)

scripts/globals/bluemagic.lua
```diff
@@ -46,6 +46,29 @@ local function calculateWSC(attacker, params)
     return wsc
 end
 
+local function calculateDEXvsAGICritRate(attacker, target)
```
WinterSolstice8: Should move this into https://github.com/LandSandBoat/server/blob/base/scripts/globals/combat/physical_utilities.lua

Xaver-DaRed: IF thats the case, I have a request:
```lua
xi.combat.physical.calculateMeleeCriticalRate = function(actor, target)
    -- See reference at https://www.bg-wiki.com/ffxi/Critical_Hit_Rate
    local criticalRate = 0
    local dDex         = actor:getStat(xi.mod.DEX) - target:getStat(xi.mod.AGI)

    if dDex > 50 then
        criticalRate = 0.15
    elseif dDex >= 40 then
        criticalRate = (dDex - 35) / 100
    elseif dDex >= 30 then
        criticalRate = 0.04
    elseif dDex >= 20 then
        criticalRate = 0.03
    elseif dDex >= 14 then
        criticalRate = 0.02
    elseif dDex >= 7 then
        criticalRate = 0.01
    end

    -- We should seriously consider adding merits, job points and other modifiers here. 
    
    return criticalRate
end
```

Reasoning:
- This files are the base of weaponskills + other actions being rewriten to be... less chaotic.
- Variable naming: I personally prefer to name variables after what wikis name them. Same terms = more clarity. I also dislike abreviations.
- Order: Critical hit rate is already defined as 0, so theres no need for a step setting it to 0.
- We should also really consider puting mods that affect critical rate in there aswell. Because theres no need to add them in a different function, nor independently in blue magic, weaponskills, TP moves, JAs, etc

Xaver-DaRed: Those are my dominions. Feel spooked. UHHHhHhHHhH.... :ghost:

MowFord: Done and done.

Added a third param that if passed as true, bypasses the other modifiers so i could change weaponskills.lua as little as possible (and for flexibility)

scripts/globals/combat/physical_utilities.lua
```diff
@@ -428,6 +428,34 @@ xi.combat.physical.calculateMeleePDIF = function(actor, target, weaponType, wsAt
     return pDif
 end
 
+-- calculates melee critical rate, if true is passed, then only returns crit from dex vs agi
+xi.combat.physical.calculateMeleeCriticalRate = function(actor, target, nativeOnly)
+    -- See reference at https://www.bg-wiki.com/ffxi/Critical_Hit_Rate
+    local nativecrit = 0
```
WinterSolstice8: Shouldn't this be 0.05? That's the base crit rate -- the rest should be adding to it, no?

Xaver-DaRed: The elseif chain below could also use some simplification.

```lua
xi.combat.physical.calculateMeleeCriticalRate = function(actor, target)
    -- See reference at https://www.bg-wiki.com/ffxi/Critical_Hit_Rate
    local finalCriticalRate = 0
    local baseCriticalRate  = 0.05
    local dDex              = actor:getStat(xi.mod.DEX) - target:getStat(xi.mod.AGI)
    local statBonus         = 0

    if dDex > 50 then
        statBonus = 0.15
    elseif dDex >= 40 then
        statBonus = (dDex - 35) / 100
    elseif dDex >= 30 then
        statBonus = 0.04
    elseif dDex >= 20 then
        criticalRate = 0.03
    elseif dDex >= 14 then
        statBonus = 0.02
    elseif dDex >= 7 then
        statBonus = 0.01
    end
    
    finalCriticalRate = baseCriticalRate + statBonus + (actor:getMod(xi.mod.CRITHITRATE) + (actor:getMerit(xi.merit.CRIT_HIT_RATE) - target:getMerit(xi.merit.ENEMY_CRIT_RATE)) / 100
    
    return finalCriticalRate
end
```

I also dont like that `nativeOnly` argument in there.
In which cases would we want the critical rate from dStat only?

MowFord: Yea the goal of this function was to be a drop-in replacement for what was used in `weaponskills.lua` as well as for use in `bluemagic.lua`. 

You might be right that we should include the base .05, i just copied the function over. Ditto for the chained if/else.

WinterSolstice8 (dismissed):

scripts/globals/combat/physical_utilities.lua
WinterSolstice8: On second thought, go with Xaver's function here

MowFord: Understood. I've tweaked as I think youl would want while keeping the functions separate for base and extra mods

scripts/globals/combat/physical_utilities.lua
```diff
@@ -428,6 +428,46 @@ xi.combat.physical.calculateMeleePDIF = function(actor, target, weaponType, wsAt
...
+    local statBonus         = 0
+
+    if dDex > 50 then
+        statBonus = 0.15
+    elseif dDex >= 40 then
+        statBonus = (dDex - 35) / 100
+    elseif dDex >= 30 then
+        statBonus = 0.04
+    elseif dDex >= 20 then
+        criticalRate = 0.03
```
Xaver-DaRed: error

MowFord: image

I'm never copying from your test again

Xaver-DaRed: I never said I tested it

MowFord: Btw if you squash we lose the co-author commits i cherry-picked

TeoTwawki: WELL THEN! LETS LEARN ADVANCED REBASING! (by "lets" I mean "you people who are not me")

TeoTwawki: ...I ***guess*** you could just put all the people as "co authored by" in the new single commit but that's a lot less ~~fun for me to watch~~ _learning a valuable new skill._

Xaver-DaRed: I actually meant to just `hold` it. Couse now I need to see what we have in weaponskills.lua

Couse I'll be honest, I dont see why we cant just have all logic regarding critical hit rate in 1 function instead of several, even if that means purging whats already in weaponskills.lua.

The only thing that could potentially conflict is the fTP step. Maybe. A mere matter of just passing nill and doing a nill check, maybe. I dunnow, hence why I have to read.

MowFord: Well this PR does remove the crit rate function in ws.lua 

I made two functions because you didn't like the parameter in a single one

And by "you" I mean "you people who are not me"

PR #4701, opened 2023-11-07
[Trust] Check for valid item before attempting to Despoil

CatsEyeXI:
**_I affirm:_**

Nanaa Mihgo will try to add a nil item to the player when Despoil fails.
This adds a check to prevent trusts from using Despoil when no item is able to be despoiled.

Hang out with Nanaa Mihgo out and fight until she tries to Despoil a mob and fails. You should not get a 'blank' item message.

### DIFF
--- a/scripts/globals/job_utils/thief.lua
+++ b/scripts/globals/job_utils/thief.lua
@@ -91,13 +91,16 @@ xi.job_utils.thief.checkCollaborator = function(player, target, ability)
 end
 
 xi.job_utils.thief.checkDespoil = function(player, target, ability)
-    if player:getFreeSlotsCount() == 0 then
-        return xi.msg.basic.FULL_INVENTORY, 0
-    end
-
-    if player:getObjType() == xi.objType.TRUST then
-        if player:getMaster():getFreeSlotsCount() == 0 then
+    if player:getObjType() == xi.objType.TRUST then -- Trust
+        if
+            player:getMaster():getFreeSlotsCount() == 0 or
+            not target:getDespoilItem()
+        then
             return 1, 0
+        end
+    else -- Player
+        if player:getFreeSlotsCount() == 0 then
+            return xi.msg.basic.FULL_INVENTORY, 0
         end
     end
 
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (dismissed):

scripts/globals/job_utils/thief.lua
```diff
@@ -212,9 +212,17 @@ xi.job_utils.thief.useDespoil = function(player, target, ability, action)
...
+        target:isMob() and
+        math.random(1, 100) <= despoilChance and
+        stolen
+    then
         if player:getObjType() == xi.objType.TRUST then
-            player:getMaster():addItem(stolen)
+            if stolen then
+                player:getMaster():addItem(stolen)
+            else
+                return xi.msg.basic.SKILL_NO_EFFECT
```
TeoTwawki: on retail the trust won't even attempt this if its not going to succeed. it will NEVER give you a no effect message. Aside form that this is only checking the bool `stolen` twice, it being in the original check already should be `false` when the item is nil and thus never reach this point.

CatsEyeXI: true... could just remove the else statement then, if i understand correctly? that seems to make sense.

Xaver-DaRed: I think a better solution would be to add a check for nil despoil item in `onAbilityCheck`
That will match what Teo is saying

TeoTwawki: > I think a better solution would be to add a check for nil despoil item in `onAbilityCheck`

check if trust and nil, so we don't affect players. the original scritp in lsb that this r originally altered can stay liek we had it and if a nil is every being given to player, rechecking the variable a 2nd time wouldn't have fixed it but would still be an issue to report to us. hopefully the issue was limited to trust despoils anyway.

Xaver-DaRed:
```lua
xi.job_utils.thief.checkDespoil = function(player, target, ability)
    if player:getObjType() == xi.objType.TRUST then -- Trust
        if
            player:getMaster():getFreeSlotsCount() == 0 or
            not target:getDespoilItem()
        then
            return 1, 0
        end
    else -- Player
        if player:getFreeSlotsCount() == 0 then
            return xi.msg.basic.FULL_INVENTORY, 0
        end
    end

    return 0, 0
end
```

Xaver-DaRed (approved):

TeoTwawki (approved):

PR #4712, opened 2023-11-09
[client] November 2023 Version Update

claywar:
**_I affirm:_**

* Updates Text IDs, Version, and NPCs for current update
* Adds existing Black Sanctus effect to core enum
* Adds new Effect to enums "Animated"

No changes should be observed when using latest client


zach2good (approved):

src/map/status_effect.h
```diff
@@ -698,6 +698,8 @@ enum EFFECT
     EFFECT_MOOGLE_AMPLIFIER        = 629,
     EFFECT_TAINT                   = 630,
     EFFECT_HAUNT                   = 631,
+    EFFECT_BLACK_SANCTUS           = 632,
+    EFFECT_ANIMATED                = 633,
```
Xaver-DaRed, line 702: of course the lua and the helper didnt match

Xaver-DaRed (approved):

PR #4717, opened 2023-11-10
[blue spells] "crit/acc varies with tp" adjustments

MowFord:
**_I affirm:_**

Goes in tandem with #4698 . Essentially handles the "crit/acc varies with TP" outside of `bluemagic.lua` for simplicity.

Ideally one day the tpmod param would define what happens inside `bluemagic.lua`. I would love to do it one day, but I don't have the time to fully audit all of them (also looks like we're missing a tpmod_chance to increase chance to apply debuffs instead of duration).

Note that the changes in the other PR are required for this to affect any change

Have 3k tp, use blu skills that vary acc or crit with TP, see that they do not get a bonus. Use CA and see that they do. Azure Lore gives a flat (large) bonus as well

### DIFF
--- a/scripts/actions/spells/blue/asuran_claws.lua
+++ b/scripts/actions/spells/blue/asuran_claws.lua
@@ -22,6 +22,13 @@ spellObject.onSpellCast = function(caster, target, spell)
     local params = {}
     params.ecosystem = xi.ecosystem.BEAST
     params.tpmod = TPMOD_ACC
+    params.bonusacc = 0
+    if caster:hasStatusEffect(xi.effect.AZURE_LORE) then
+        params.bonusacc = 70
+    elseif caster:hasStatusEffect(xi.effect.CHAIN_AFFINITY) then
+        params.bonusacc = math.floor(caster:getTP() / 50)
+    end
+
     params.attackType = xi.attackType.PHYSICAL
     params.damageType = xi.damageType.HTH
     params.scattr = xi.skillchainType.LIQUEFACTION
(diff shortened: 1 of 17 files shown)


scripts/actions/spells/blue/asuran_claws.lua
```diff
@@ -22,6 +22,7 @@ spellObject.onSpellCast = function(caster, target, spell)
     local params = {}
     params.ecosystem = xi.ecosystem.BEAST
     params.tpmod = TPMOD_ACC
+    params.bonusacc = caster:hasStatusEffect(xi.effect.AZURE_LORE) and 70 or (0 + (caster:hasStatusEffect(xi.effect.CHAIN_AFFINITY) and math.floor(caster:getTP() / 50 or 0)))
```
Xaver-DaRed: Dont take this as a must, just my opinion AND personal preference, however, I would rather bonusacc be defined as 0 and then use if/else to set its value, instead of this. For clarity.

MowFord: Was simpler to copy paste a one liner but I do see yours as easier to read. Ideally this logic gets put into the main lua, I just don't see a half solution getting approved and I can't do a full implementation at this time

Let me know if you want it updated. Otherwise I think it best to leave as a style choice when it's moved to core

TeoTwawki: Def update it. The line length alone bothers me. 

(Using multiple variables could solve that as well. I'm just talking about length.)

claywar: Yeah, nested ternaries in lua might be a bit too much.  If you still wanted to keep using the style, I'd suggest splitting into two variables

MowFord: Understood

Xaver-DaRed (approved):

PR #4719, opened 2023-11-12
[core] Add PEntity->GetUntargetable() in CAIContainer::IsUntargetable

WinterSolstice8:
**_I affirm:_**

This will prevent bad packets from being sent out if a mob is changed to be untargetable mid-action

start casting a long cast time spell like freeze on a mob
!exec target:setUntargetable(true) and see you get an interrupt animation instead of nothing/jank packet


Xaver-DaRed (approved):
Great

TeoTwawki (approved):

PR #4721, opened 2023-11-12
[Lua] Add Campaign Arbiter teleport/retrace behavior

MowFord:
**_I affirm:_**

Created new `teleports` lua file to support two new sets of functions for teleporting from nations to the campaign front and retracing from the campaign front to the aligned nation.

The retrace NPCs now don't show teleport option if you are in the zone you'd teleport to. I've include the additional Campaign Arbiters' lua files in each city zone, but their status is still set to 2 in the db. 

The teleport NPCs give a menu that lists all zones you've visited and cost based on wiki sources.

Showing the teleport to battlefront in action:
image

The menu didn't show CN_S until i zoned to CN_S and went back to npc

Interact with the CA NPCs in the 3 nations in the past and get the default "Civilian what are you doing" message. Adjust your allegiance and get the menu based on zones visited (with a catchall if somehow you end up in the past nation, with an allegiance, and never having visited the zones)

Interact with the Retrace NPCs in the 3 nations in the past and see that they give you the option to Retrace only if you're aligned with a nation and you are not currently in that nation.

### DIFF
--- a/scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua
+++ b/scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua
@@ -53,8 +53,6 @@ return {
     ['Riautte']                  = { event = 41 },
     ['Rongelouts_N_Distaud']     = { event = 604 },
     ['Sabiliont']                = { event = 608 },
-    ['Saphiriance_TK']           = { event = 450 },
-    ['Scarlette_CA']             = { event = 459 },
     ['Suspicious_Object']        = { text = ID.text.NOTHING_OUT_OF_ORDINARY },
     ['Thierride']                = { event = 333 },
     ['Touttaures']               = { event = 603 },
--- a/scripts/effects/teleport.lua
+++ b/scripts/effects/teleport.lua
@@ -27,6 +27,9 @@ effectObject.onEffectLose = function(target, effect)
         xi.teleport.toHomeNation(target)
     elseif destination == xi.teleport.id.RETRACE then
         xi.teleport.toAlliedNation(target)
+    elseif destination == xi.teleport.id.CAMPAIGN then
+        local campaignDestination = effect:getSubPower()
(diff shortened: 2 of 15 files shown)


Xaver-DaRed (approved):
Please add the positions to the npcs.

Xaver-DaRed:
This needs a rebase to resolve conflicts.

MowFord:
> This needs a rebase to resolve conflicts.

Thanks for lookin out. All clear now

scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua
```diff
@@ -53,8 +53,6 @@ return {
     ['Riautte']                  = { event = 41 },
     ['Rongelouts_N_Distaud']     = { event = 604 },
     ['Sabiliont']                = { event = 608 },
-    ['Saphiriance_TK']           = { event = 450 },
```
Comment (claywar) [inline scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua:56]: Why are these being removed? These are absolutely valid from caps when unavailable, which would be prior to access from quests.  These _are_ the default action.

claywar (changes requested):

scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua
Comment (MowFord) [reply scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua:56]: The updated npc lua handles this event when the player doesn't have campaign allegiance

Comment (claywar) [reply scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua:56]: I may be misunderstanding, but you seem to be using base event of 458 - 1 for no allegiance.  This is a global, and I would suggest testing this with other zones.

Comment (MowFord) [reply scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua:56]: Saphiriance was flagged in this request but the question about 458 - 1 was about a different npc.

I see now that there was a mistake and it should be +1 for no allegiance:

image

I've rebased, then made that adjustment and tested the NPCs with and without allegiance

PR #4722, opened 2023-11-12
[Lua] Scarlette_CA: incorrect default event id

MowFord:
**_I affirm:_**

The PR fixes a typo in the event id for this NPC. 

Partner PR with https://github.com/LandSandBoat/server/pull/4721

I'm splitting this out for my own information because I'd love to know what exactly the reasoning is for `DefaultActions.lua` to be round robined with an actual `onTrigger` function for an NPC.

Tracked the history for this entry to just when this zone's `DefaultActions.lua` file was created, the event id was copied wrong...but I'd like to know why exactly it's used when an NPC has a proper `onTrigger` function.

Story:
I updated `Scarlette_CA.lua` to perform an actual startevent, but hitting Scarlette the first time would lock the client. You could hit esc to unlock and interact again to get the menu to function properly (sounds trivial, but it actually wasted a solid hour of my time trying to figure out wth was going on there). So basically an NPC defined in `DefaultActions.lua` uses round robin between the npc's lua and the default action lua event id

### DIFF
--- a/scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua
+++ b/scripts/zones/Southern_San_dOria_[S]/DefaultActions.lua
@@ -54,7 +54,7 @@ return {
     ['Rongelouts_N_Distaud']     = { event = 604 },
     ['Sabiliont']                = { event = 608 },
     ['Saphiriance_TK']           = { event = 450 },
-    ['Scarlette_CA']             = { event = 489 },
+    ['Scarlette_CA']             = { event = 459 },
     ['Suspicious_Object']        = { text = ID.text.NOTHING_OUT_OF_ORDINARY },
     ['Thierride']                = { event = 333 },
     ['Touttaures']               = { event = 603 },


claywar (approved):

MowFord:
For posterity, I found out what was going on. The interaction framework by design alternate between the npc lua file and the interaction framework logic (including default actions). So the above was working as intended, but default actions aren't needed of an npc has a functional lua file.

claywar:
There's two different things at play.  When moving the event to defaults, it was fat-fingered from 459 to 489.  That was the root cause of this issue.  Default actions in Interaction are the _lowest_ priority item.  Any event call should override this, be it in an npc script or quest/mission script.  In this case, there is no difference, as the npc only had one action.

Also, the preferred method for any NPC that has _only_ quest or mission functions is to not use the NPC script in favor of interaction.

MowFord:
> There's two different things at play. When moving the event to defaults, it was fat-fingered from 459 to 489. That was the root cause of this issue. Default actions in Interaction are the _lowest_ priority item. Any event call should override this, be it in an npc script or quest/mission script. In this case, there is no difference, as the npc only had one action.
> 
> Also, the preferred method for any NPC that has _only_ quest or mission functions is to not use the NPC script in favor of interaction.

But it was still doing round robin on the incorrect default event id, other than that yep everything else makes sense, thanks!

PR #4725, opened 2023-11-13
[core] Add 0x051 equipset to retrigger latents, retrigger on action

WinterSolstice8:
**_I affirm:_**

This commit refactors the retrigger latent variable in CCharEntity to be more generic in case latents need to be retriggered elsewhere
This commit also will allow retriggering of latents sent from equipsets, not just normal equip packets.
Actions (0x01A) will now reprocess latents if an equipset/equip was sent in the same chunk.

Try to trigger various latents that would exist only in precast (ala gearswap/luashitacast) and see they work correctly


Xaver-DaRed (approved):
LGTM

TeoTwawki (approved):

PR #4729, opened 2023-11-15
[Core] Rework getSpeed()

### LINKED ISSUE #2482: 🐛 Movement speed equipment bonuses stack unexpectedly

## Branch affected by issue 

`base`

## Steps to reproduce

Equip multiple movement speed bonus items. For this example I used
Danzo Sun-ate +12%
Kinddom Aketon +12% in San d'Oria
Shneddick Ring +18%
This returns an equipment movement speed bonus of +42%. Verified with speedchecker and targetinfo. 
## Expected behavior
Movement speed equipment bonuses should not stack. Only the highest bonus should take effect.
https://www.bg-wiki.com/ffxi/Movement_Speed

speed
ffxi_2022 08 08_18 44 55

Xaver-DaRed:
**_I affirm:_**

Tries to get us closer to matching what retail does with speed in the following ways:
- Creates a set of modifiers to separate movement speed sources.
- Makes possitive movement speed bonuses from gear NOT stack with each other and take the highest only instead.
- Makes Mazurka and quickening not stack and take the highest of the two.
- Makes weight and curse, among others, stack with each other and be multiplicative to other sources.
- Fixes an issue with !speed GM command when using movement gear.

Resolves #2482, resolves #2499, resolves #2501 

## Retail observations:
- Positive movement speed from gear DOES NOT stack with each other. Only highest applies.
- Negative movement speed from gear DOES stack with each other.
- Flee puts you at 80 speed. 2 pieces of dusk gear did NOT affect movement speed. The third one did. We concluded flee gives a bonus higher than the cap.
- Max speed is 80. There is no possible way that we could come up a different combination which allowed us to surpass this limit.
  - Combination -> +18% speed ring + Chocobo mazurka + sprinter shoes useable item effect + PERFECT Bolters roll + KI resulted in flee speed. 
- Positive movement speed from gear resulted in diminishing returns when combined with negative gear.
  - Bloody pants + dusk mask should result in ABOUT the same speed as regular speed. However, it was notizably slower, just not THAT much.
- Weight effects are concluded to be multiplicative. Full dusk gear + mollusca mantle + weight + curse resulted in a very slow movement speed, but not much different than just gear. If it was additive, we would have just stoped.
  - This was not the case of a lower limit cap. Beadeaux curse + weight is slower than just curse. Similarly to positive movement speed from gear, we saw diminishing returns (in this case, a diminishing penalty, since just curse with no negative gear is still very notizable)

Go to retail and check the actions described above.
Perform those same actions here and see if its at the very least SOMEWHAT close.


TeoTwawki:
> Max speed is 80. There is no possible way that we could come up a different combination which allowed us to surpass this limit.

status effects can exceed this. stronger versions of quickening can be stacked up (kupower maxed from resting) and **feast of swords in modern puts you at ~~200~~ 155**)

so its not a hard cap on overall speed. its more like the way we handle haste where each source has a cap and dual wield and delay reductions and then they all get shmushed together to create your attack speed.

key item from mog garden monster rearing (cactaur) functions like dusk but in positive instead of negative (it stacks with yer stuff)

src/map/entities/battleentity.cpp
```diff
@@ -226,17 +226,44 @@ int32 CBattleEntity::GetMaxMP() const
...
-    int16 startingSpeed = isMounted() ? 40 + settings::get<int8>("map.MOUNT_SPEED_MOD") : speed;
-    // Mod::MOVE (169)
-    // Mod::MOUNT_MOVE (972)
-    Mod mod = isMounted() ? Mod::MOUNT_MOVE : Mod::MOVE;
+    if (isMounted())
+    {
+        baseSpeed   = 40 + settings::get<int8>("map.MOUNT_SPEED_MOD");
+        outputSpeed = baseSpeed * (100 + getMod(Mod::MOUNT_MOVE)) / 100;
+
+        return std::clamp<uint8>(outputSpeed, 0, 80);
```
TeoTwawki: please don't clamp final chocobo/mount to 80. We want the setting to allow ridiculous chocobo speeds. It won't get that high without someone intentionally putting the setting way up there.

scripts/enum/mod.lua
```diff
@@ -912,6 +912,12 @@ xi.mod =
 
     STEP_TP_CONSUMED        = 1077, -- Modifies the amount of TP consumed by dancer steps
 
+    MOVE_SPEED_STACKABLE      = 75, -- Gear movement speed penalties, flee bonus, etc.
```
TeoTwawki: since these are all together they can be placed earlier in the list instead of at the end

src/map/entities/battleentity.cpp
```diff
@@ -226,17 +226,44 @@ int32 CBattleEntity::GetMaxMP() const
...
+    // We have all the modifiers needed. Calculate final speed.
+    float modifiedSpeed = static_cast<float>(baseSpeed) * (additiveMods + effectBonus) * gearBonus * weightPenalties
 
-    float modAmount     = (100.0f + static_cast<float>(getMod(mod))) / 100.0f;
-    float modifiedSpeed = static_cast<float>(startingSpeed) * modAmount;
-    uint8 outputSpeed   = static_cast<uint8>(modifiedSpeed);
+    outputSpeed = static_cast<uint8>(modifiedSpeed);
 
-    return std::clamp<uint8>(outputSpeed, std::numeric_limits<uint8>::min(), std::numeric_limits<uint8>::max());
+    return std::clamp<uint8>(outputSpeed, 0, 80);
```
TeoTwawki: same here about clamping to 80. Due to settings, you want to take the setting into account not just use the 80 retail has

src/map/modifier.h
```diff
@@ -980,13 +980,19 @@ enum class Mod
     DIG_BYPASS_FATIGUE = 1074, // Chocobo digging modifier found in "Blue Race Silks". Modifier works as a direct percent.
     BREATH_DMG_DEALT   = 1075, // Breath damage dealt
 
+    // New movement speed modifiers.
```
TeoTwawki: maybe just throw them right under where `move` is waiting to get deleted and before `mount_move` sits

src/map/entities/battleentity.cpp
TeoTwawki: side thought: we really at some point need to map the flag bits that are packed with the speed value. it was being capped at uint8 max before (255) but it actually stops before that because of the non-speed bits stealing some space. you can witness the speed "rolling over" in the client by using the GM command to go "too high" and finding yourself in slow motion past 240

src/map/entities/battleentity.cpp
```diff
@@ -226,17 +226,44 @@ int32 CBattleEntity::GetMaxMP() const
...
-    Mod mod = isMounted() ? Mod::MOUNT_MOVE : Mod::MOVE;
+    if (isMounted())
+    {
+        baseSpeed   = 40 + settings::get<int8>("map.MOUNT_SPEED_MOD");
+        outputSpeed = baseSpeed * (100 + getMod(Mod::MOUNT_MOVE)) / 100;
+
+        return std::clamp<uint8>(outputSpeed, 0, 80);
+    }
+
+    // Flee, KIs, Gear penalties, Bolters Roll.
```
MowFord, line 242: so bolter's roll fits interestingly into this:

> Stacks multiplicatively with [Movement Speed](https://www.bg-wiki.com/ffxi/Movement_Speed) equipment and additively with either [Mazurka](https://www.bg-wiki.com/ffxi/Mazurka) or [Quickening](https://www.bg-wiki.com/ffxi/Quickening) (but not both).

As usual I've done no retail testing, just wanted to add my 2cents that according to this wording it appears that bolter's roll does need to just be part of the additiveMods here. I.E. the logic LGTM

Xaver-DaRed: Yeah, I was happy to see that it "fitted nicely" so to say.

settings/default/map.lua
```diff
@@ -104,6 +104,10 @@ xi.settings.map =
     -- Modifier to apply to agro'd monster speed. 0 is the retail accurate default. Negative numbers will reduce it.
     MOB_SPEED_MOD = 0,
 
+    -- Modifier that limits how fast players can move. Bypassed by feast of swords. This does not prevent cheats, btw.
+    SPEED_CAP       = 80,
```
TeoTwawki: I don't believe this needs a setting, nor that the final speed value needs clamped to 80. By definition when we adjust the existing speed settings value we're already going non-retail, we just need to include the settings value for the final results max value (and also stay within uint8 bounds). We shouldn't be trying to decide the speed limit there, only that we're not rolling into an overflow/underflow.

N3ckB3ard:
Linked to other topics:
#2501
#2499 
#2482

cocosolos:
Just gonna leave this here. Maybe relevant for future research.
https://www.youtube.com/watch?v=qBBSZnut42Q

Xaver-DaRed:
I saw that video a few days ago.
I think this was fixed in retail, but if this wasnt fixed, I refuse to emulate it.

Thats why the clamps are there, so it can never underflow

TeoTwawki:
p.s. technically, #2501 is an infodump in discussions rather than an issue to close. But this does take care of 90% of what it lays out as being wrong so ima have to update it after this merges :)

TeoTwawki (approved):

PR #4740, opened 2023-11-18
[LUA] Update aoe teleport items

MowFord:
add a small delay to whole party is warped

**_I affirm:_**

AoE teleport items: Tidal, Moogle/Nomad cap

They run onItemUse in whichever order cpp decides from the target list, but the `onItemUse` was immediately using `setPos`, so more often than not people would be left behind. This PR changes their behavior to update `xi.effect.teleport` with a `TIDAL_TALISMAN` type, and updates nomad/moogle cap to utilize the already-built `HOME_NATION` type

Confirmed behavior is similar for all three items:

image

Get in party in single cluster, use tidal, if both players are teleported, give tidal to other member and use. Other member should be left behind

Update with changes from this PR and see that the party is reliably teleported.

### DIFF
--- a/scripts/items/moogle_cap.lua
+++ b/scripts/items/moogle_cap.lua
@@ -3,6 +3,8 @@
 --  Moogle Cap
 --  Transports the user to their Home Nation
 -----------------------------------
+require('scripts/globals/teleports')
+-----------------------------------
 local itemObject = {}
 
 itemObject.onItemCheck = function(target)
@@ -10,15 +12,7 @@ itemObject.onItemCheck = function(target)
 end
 
 itemObject.onItemUse = function(target)
-    local nation = target:getNation(target)
-
-    if nation == 0 then -- San d'Oria
-        target:setPos(126, 0, -1, 122, 231)
-    elseif nation == 1 then -- Bastok
-        target:setPos(106, 0, -71, 130, 234)
-    elseif nation == 2 then -- Windurst
-        target:setPos(197, -12, 224, 65, 240)
-    end
+    target:addStatusEffectEx(xi.effect.TELEPORT, 0, xi.teleport.id.HOME_NATION, 0, 4)
 end
 
 return itemObject
--- a/scripts/items/tidal_talisman.lua
+++ b/scripts/items/tidal_talisman.lua
@@ -2,39 +2,17 @@
 -- ID: 11290
 -- Item: tidal talisman
 -----------------------------------
+require('scripts/globals/teleports')
+-----------------------------------
 local itemObject = {}
-
-local teleportData =
-{
-    [xi.zone.CHATEAU_DORAGUILLE]   = {   0,   3,   2,  64, xi.zone.RULUDE_GARDENS },
-    [xi.zone.NORTHERN_SAN_DORIA]   = {   0,   3,   2,  64, xi.zone.RULUDE_GARDENS },
-    [xi.zone.SOUTHERN_SAN_DORIA]   = {   0,   3,   2,  64, xi.zone.RULUDE_GARDENS },
-    [xi.zone.PORT_SAN_DORIA]       = {   0,   3,   2,  64, xi.zone.RULUDE_GARDENS },
-    [xi.zone.BASTOK_MARKETS]       = {   0,   3,   2,  64, xi.zone.RULUDE_GARDENS },
-    [xi.zone.BASTOK_MINES]         = {   0,   3,   2,  64, xi.zone.RULUDE_GARDENS },
-    [xi.zone.METALWORKS]           = {   0,   3,   2,  64, xi.zone.RULUDE_GARDENS },
-    [xi.zone.PORT_BASTOK]          = {   0,   3,   2,  64, xi.zone.RULUDE_GARDENS },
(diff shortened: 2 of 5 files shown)


scripts/items/moogle_cap.lua
```diff
@@ -5,20 +5,23 @@
...
-    local nation = target:getNation(target)
-
-    if nation == 0 then -- San d'Oria
-        target:setPos(126, 0, -1, 122, 231)
-    elseif nation == 1 then -- Bastok
-        target:setPos(106, 0, -71, 130, 234)
-    elseif nation == 2 then -- Windurst
-        target:setPos(197, -12, 224, 65, 240)
-    end
+    target:timer(1000, function(playerArg)
```
TeoTwawki: I don't think timer() is the best way to do this. There's a delay baked into a status effect we're supposed to use, and the set pos need to be moved to it. For all items like this. They need their destinations added to the teleports global and enumerated.

zach2good:
Rather than bake a delay into each of these, I'd rather know why using setPos in onItemUse isn't working as expected - if its to allow playout of an animation thats one thing, but it should _work_

TeoTwawki:
> Rather than bake a delay into each of these, I'd rather know why using setPos in onItemUse isn't working as expected - if its to allow playout of an animation thats one thing, but it should _work_

its a problem of not using a feature that was already made, these were incorrectly scripted before the pr author got here and not a problem with onItemUse

example:
https://github.com/LandSandBoat/server/blob/480de76a43687c19467baf31c6c5362c3d4599f1/scripts/items/kingdom_stables_collar.lua#L18

```lua
    target:addStatusEffectEx(xi.effect.TELEPORT, 0, xi.teleport.id.CHOCO_SANDORIA, 0, 4)
end
```
that teleporting item functions correctly with a 4 second delay. the item animates properly, and then the teleport does after.

(being aoe just means mutliple persons are target and will get the status hidden effect, which will perform the transport upon its end)

scripts/items/moogle_cap.lua
MowFord: Agreed, was just trying to take the path of least resistance. Suppose we can make it a proper teleport using xi.effect.teleport but that would need to happen after my campaign PR gets merged.

Also it is literally 3 items that have this paradigm

MowFord:
I guess the consensus is that you'd rather update xi.teleports with a new one for each of these? This would allow us to use `target:addStatusEffectEx(xi.effect.TELEPORT`

TeoTwawki:
> I guess the consensus is that you'd rather update xi.teleports with a new one for each of these? This would allow us to use `target:addStatusEffectEx(xi.effect.TELEPORT`

that is what the Teo said yes

MowFord:
My CA teleport PR is pending changes to that file, I'll update this but i suspect i'm going to merge conflict myself :(

scripts/items/moogle_cap.lua
MowFord: also 2 of the 3 were able to use the pre-existing teleport type 👍

scripts/items/moogle_cap.lua
```diff
@@ -3,22 +3,16 @@
...
-    local nation = target:getNation(target)
-
-    if nation == 0 then -- San d'Oria
-        target:setPos(126, 0, -1, 122, 231)
-    elseif nation == 1 then -- Bastok
-        target:setPos(106, 0, -71, 130, 234)
-    elseif nation == 2 then -- Windurst
-        target:setPos(197, -12, 224, 65, 240)
-    end
+    target:addStatusEffectEx(xi.effect.TELEPORT, 0, xi.teleport.id.HOME_NATION, 0, 2)
```
TeoTwawki: if these use same animation as a standard teleport, use 4 instead of 2 at the end. if they use a warp style animation, use 3.

MowFord: all 3 have simple item animations, it seems:



https://github.com/LandSandBoat/server/assets/131182600/0e9cb792-bb65-4c5d-ae44-771923064b97

MowFord: This is on local btw, I don't have a retail account

TeoTwawki: > This is on local btw, I don't have a retail account

well that doesn't help us at all then and I have to assume the animations are likely to be wrong in lsb. use 4 for now.

scripts/items/tidal_talisman.lua
```diff
@@ -43,7 +21,7 @@ itemObject.onItemCheck = function(target)
 end
 
 itemObject.onItemUse = function(target)
-    target:setPos(unpack(teleportData[target:getZoneID()]))
+    target:addStatusEffectEx(xi.effect.TELEPORT, 0, xi.teleport.id.TIDAL_TALISMAN, 0, 2)
```
TeoTwawki: missed a 2. soon as thats updated this pr is ready to merge.

MowFord: -10 to mowfordor

TeoTwawki (approved):

Xaver-DaRed (approved):

PR #4744, opened 2023-11-20
[Fix] nil value in healing.lua effect script

Xaver-DaRed:
**_I affirm:_**

- Fixes an issue when doing GEO unlock quest.

As spoken about in discord...
imagen

- Also, minor OCD cleanup.

Unlock GEO via quest with no errors in log.
Yes, unlock GEO, I know.


TeoTwawki (approved):

zach2good:
I was going to question whether the IDs for the zone in question could be nil
But that's where the player is, so it has to be loaded, all good

PR #4747, opened 2023-11-20
[Fix] Magian Trials +DMG augment audit

MowFord:
**_I affirm:_**

Was found that Trial 1826 rewarded an incorrect +DMG augment. Wrote a quick script to parse the file for any other instances... and found none:

image

```
with open("magian_data.lua") as file:
   data = file.read()

lines = data.split("\n")

for line in [line for line in lines if '-- DMG' in line]:
   dmg = int(line.split('DMG:')[1].strip().split(' ')[0].strip()[1:])
   aug = int(line.split('{')[1].split(',')[1].strip().split(' ')[0].strip()) + 1
   if dmg - aug != 0:
     print(line)
     print('{} - {}'.format(dmg,aug))
```

Trade spharai with +6 dmg augment for trial 1826, see that you receive sparai with +8 dmg

### DIFF
--- a/scripts/globals/magian_data.lua
+++ b/scripts/globals/magian_data.lua
@@ -4036,7 +4036,7 @@ xi.magian.trials =
             itemId       = xi.item.SPHARAI,
             itemAugments =
             {
-                [1] = { 740, 5 }, -- DMG:+8
+                [1] = { 740, 7 }, -- DMG:+8
             },
         },
     },


Xaver-DaRed (approved):
We are going to need to audit the item enums aswell. Couse that SPHARAI should be SPHARAI_75

TeoTwawki (approved):

PR #4751, opened 2023-11-21
[CPP] Reduce expensive calls to SaveCharEquip()

MowFord:
**_I affirm:_**

Intro via quote from my original PR at WingsXI gitlab:
> Problem: when latency between map server and mysql is >1ms, the map server lags hard due to the nature of all sql queries being run in series and the map process waiting for the completion to continue. For most operations this isn't a problem ,but one thing in particular causes real issues: gear swaps.
Example of a macro swapping to do a nuke spell:
image

>This MR simply rate-limits the UPDATE to sql for saving equipment to the database. The only thing that reads the equipment from the db is when a character zones in (or logs in), so I've also added a save before logging out and zoning out.

The situation is similar on LSB it seems. When I originally found this it was from Tracy logging seeing excessive time allocated to processing packet 0x050 and 0x051. Equipsets are somewhat more efficient as they send the whole list of gear that need to be swapped. 0x050 is sent by Ashita when gearswap is called. It sends separate packets for every piece of gear, which (before this PR) calls `SaveCharEquip()` for each packet. This function could possibly be made more efficient with a single transaction for all 18 slots, but it's also mostly unnecessary since the equipment is only read on login/zone. One thing I've found that LSB has but doesn't seem to use is `CCharEntity::PersistData()`. Curious if this change could come with some schedule of persisting equipment. I don't see any references in the code to flagging `dataToPersist & CHAR_PERSIST::EQUIP`. 

At any rate, this change greatly improves input latency when players use Ashitacast to swap gear. CatsEyeXI has a group event that had 50+ people in it over the weekend and input latency was multiple seconds. We implemented these changes with great success.

Testing can be mostly done by confirming that equipment is saved on every way to leave a zone (zoneline, logout, teleport, etc). AFAICT every teleport-type mechanism calls SendToZone, and WingsXI has used this code for a long while with no complaints of gear being unequipped.

Specifically, run

`SELECT * FROM char_look WHERE charid = 1;`

then change one piece and run it again, see that the slot you equipped updated, then swap and run a third time to see it doesn't save.

Furthermore, check equip saving on logout/zone via this query:

`SELECT * FROM char_equip WHERE charid = 1; `

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -173,6 +173,11 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "packets/zone_in.h"
 #include "packets/zone_visited.h"
 
+namespace
+{
+    const std::string swapRateLimitVar = "core-LastGearSwap";
+} // namespace
+
 uint8 PacketSize[512];
 
 std::function<void(map_session_data_t* const, CCharEntity* const, CBasicPacket&)> PacketParser[512];
@@ -3487,8 +3492,15 @@ void SmallPacket0x050(map_session_data_t* const PSession, CCharEntity* const PCh
     }
 
     charutils::EquipItem(PChar, slotID, equipSlotID, containerID); // current
-    charutils::SaveCharEquip(PChar);
-    charutils::SaveCharLook(PChar);
+    // Only save char look to db every X seconds (or more) to avoid db spam when changing individual pieces
+    // Do not save equipment at all since SaveCharEquip is a very expensive operation
+    auto lastSwapTime   = PChar->GetLocalVar(swapRateLimitVar.c_str());
+    auto timeNowSeconds = std::chrono::time_point_cast<std::chrono::seconds>(server_clock::now());
+    if (lastSwapTime == 0 || (timeNowSeconds.time_since_epoch().count() - lastSwapTime) > 30)
+    {
+        charutils::SaveCharLook(PChar);
+    }
+    PChar->SetLocalVar(swapRateLimitVar.c_str(), static_cast<uint32>(timeNowSeconds.time_since_epoch().count()));
     luautils::CheckForGearSet(PChar); // check for gear set on gear change
     PChar->UpdateHealth();
     PChar->retriggerLatents = true; // retrigger all latents later because our gear has changed
@@ -3520,8 +3532,15 @@ void SmallPacket0x051(map_session_data_t* const PSession, CCharEntity* const PCh
             charutils::EquipItem(PChar, slotID, equipSlotID, containerID);
         }
     }
-    charutils::SaveCharEquip(PChar);
-    charutils::SaveCharLook(PChar);
+    // Only save char look to db every X seconds (or more) to avoid db spam when changing individual pieces
+    // Do not save equipment at all since SaveCharEquip is a very expensive operation
+    auto lastSwapTime   = PChar->GetLocalVar(swapRateLimitVar.c_str());
+    auto timeNowSeconds = std::chrono::time_point_cast<std::chrono::seconds>(server_clock::now());
+    if (lastSwapTime == 0 || (timeNowSeconds.time_since_epoch().count() - lastSwapTime) > 30)
+    {
+        charutils::SaveCharLook(PChar);
+    }
+    PChar->SetLocalVar(swapRateLimitVar.c_str(), static_cast<uint32>(timeNowSeconds.time_since_epoch().count()));
     luautils::CheckForGearSet(PChar); // check for gear set on gear change
     PChar->UpdateHealth();
     PChar->retriggerLatents = true; // retrigger all latents later because our gear has changed
@@ -6653,6 +6672,9 @@ void SmallPacket0x0E7(map_session_data_t* const PSession, CCharEntity* const PCh
     {
         return;
     }
+
+    charutils::SaveCharEquip(PChar);
+    charutils::SaveCharLook(PChar);
 
     if (PChar->m_moghouseID || PChar->nameflags.flags & FLAG_GM || PChar->m_GMlevel > 0)
     {
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -6402,6 +6402,8 @@ namespace charutils
             PChar->resetPetZoningInfo();
         }
 
+        SaveCharEquip(PChar);
+        SaveCharLook(PChar);
         PChar->pushPacket(new CServerIPPacket(PChar, type, ipp));
     }
 


src/map/packet_system.cpp
```diff
@@ -3487,8 +3487,15 @@ void SmallPacket0x050(map_session_data_t* const PSession, CCharEntity* const PCh
     }
 
     charutils::EquipItem(PChar, slotID, equipSlotID, containerID); // current
-    charutils::SaveCharEquip(PChar);
-    charutils::SaveCharLook(PChar);
+    // Only save char look to db every X seconds (or more) to avoid db spam when changing individual pieces
+    // Do not save equipment at all since SaveCharEquip is a very expensive operation
+    auto lastSwapTime   = PChar->GetLocalVar("core-LastGearSwap");
```
zach2good: Please extract the key `std::string` "core-LastGearSwap" to the top of the file into an anonymouse namespace, like in zone_entities.cpp:
```cpp

namespace
{
    const float CHARACTER_SYNC_DISTANCE                = 45.0f;
    const float CHARACTER_DESPAWN_DISTANCE             = 50.0f;
    const int   CHARACTER_SWAP_MAX                     = 5;
    const int   CHARACTER_SYNC_LIMIT_MAX               = 32;
    const int   CHARACTER_SYNC_DISTANCE_SWAP_THRESHOLD = 30;
    const int   CHARACTER_SYNC_PARTY_SIGNIFICANCE      = 100000;
    const int   CHARACTER_SYNC_ALLI_SIGNIFICANCE       = 10000;
    const int   PERSIST_CHECK_CHARACTERS               = 20;
} // namespace

```

src/map/packet_system.cpp
```diff
@@ -3487,8 +3487,15 @@ void SmallPacket0x050(map_session_data_t* const PSession, CCharEntity* const PCh
...
+    auto timeNowSeconds = std::chrono::time_point_cast<std::chrono::seconds>(server_clock::now());
+    if (lastSwapTime == 0 || (timeNowSeconds.time_since_epoch().count() - lastSwapTime) > 30)
+    {
+        charutils::SaveCharLook(PChar);
+    }
+    PChar->SetLocalVar("core-LastGearSwap", (uint32)timeNowSeconds.time_since_epoch().count());
```
zach2good: Please replace the c-style cast with static_cast

src/map/packet_system.cpp
```diff
@@ -3520,8 +3527,15 @@ void SmallPacket0x051(map_session_data_t* const PSession, CCharEntity* const PCh
             charutils::EquipItem(PChar, slotID, equipSlotID, containerID);
         }
     }
-    charutils::SaveCharEquip(PChar);
-    charutils::SaveCharLook(PChar);
+    // Only save char look to db every X seconds (or more) to avoid db spam when changing individual pieces
+    // Do not save equipment at all since SaveCharEquip is a very expensive operation
+    auto lastSwapTime   = PChar->GetLocalVar("core-LastGearSwap");
```
zach2good: See the previous comment

src/map/packet_system.cpp
```diff
@@ -3520,8 +3527,15 @@ void SmallPacket0x051(map_session_data_t* const PSession, CCharEntity* const PCh
...
+    auto timeNowSeconds = std::chrono::time_point_cast<std::chrono::seconds>(server_clock::now());
+    if (lastSwapTime == 0 || (timeNowSeconds.time_since_epoch().count() - lastSwapTime) > 30)
+    {
+        charutils::SaveCharLook(PChar);
+    }
+    PChar->SetLocalVar("core-LastGearSwap", (uint32)timeNowSeconds.time_since_epoch().count());
```
zach2good: Please replace the c-style cast with static_cast

src/map/packet_system.cpp
```diff
@@ -6654,6 +6668,8 @@ void SmallPacket0x0E7(map_session_data_t* const PSession, CCharEntity* const PCh
         return;
     }
 
+    charutils::SaveCharEquip(PChar);
```
zach2good, line 6676: If 2x swaps happen in quick succession and then a player logs out, the most recent gear won't be saved by this?

You could add a forceSave bool to the call to make sure it saves

src/map/utils/charutils.cpp
```diff
@@ -6402,6 +6402,8 @@ namespace charutils
             PChar->resetPetZoningInfo();
         }
 
+        SaveCharEquip(PChar);
+        SaveCharLook(PChar);
```
zach2good, line 6406: Same here

zach2good:
Also, have you profiled this to confirm that it's actually a problem? I'm sure it is, but I'd like to see the numbers

MowFord:
> Also, have you profiled this to confirm that it's actually a problem? I'm sure it is, but I'd like to see the numbers

I knew you'd ask :)

Unfortunately I haven't dipped my feet in capturing Tracy in LSB. But dozens of players can't be wrong... right? In all seriousness though the scale at which ashitacast causes this to balloon there's no way it's not still a thing. I think due to the multi-threaded nature of the packet system on lsb maybe it's not quite as noticable, but it was definitely noticable with 50+ people in the same area.

src/map/packet_system.cpp
MowFord: I'm not sure why it would be ignored here? The rate-limit is only being imposed by incoming 050 and 051, `SaveCharEquip()` isn't rate limited at all inside of the actual function

zach2good:
> the multi-threaded nature of the packet system on lsb

The main runtime tasks of Darkstar, Topaz, and LSB are all entirely single-threaded. An explanation of why can be found here: <https://github.com/LandSandBoat/server/wiki/Miscellaneous-Server#what-is-multi-process> and a walkthrough:  <https://github.com/LandSandBoat/server/wiki/Development-Server-Startup-Tutorial>

> I haven't dipped my feet in capturing Tracy in LSB.

No time like the present to learn: <https://github.com/LandSandBoat/server/wiki/Performance-Profiling-with-Tracy>

> But dozens of players can't be wrong... right?

I will bless you with an internal LSB meme:
image

src/map/packet_system.cpp
zach2good: Ah I see what you mean, please disregard

src/map/packet_system.cpp
```diff
@@ -6654,6 +6668,8 @@ void SmallPacket0x0E7(map_session_data_t* const PSession, CCharEntity* const PCh
...
+    charutils::SaveCharLook(PChar);
```
zach2good, line 6677: Could do with a newline here

src/map/packet_system.cpp
```diff
@@ -3487,8 +3487,15 @@ void SmallPacket0x050(map_session_data_t* const PSession, CCharEntity* const PCh
     }
 
     charutils::EquipItem(PChar, slotID, equipSlotID, containerID); // current
-    charutils::SaveCharEquip(PChar);
-    charutils::SaveCharLook(PChar);
+    // Only save char look to db every X seconds (or more) to avoid db spam when changing individual pieces
+    // Do not save equipment at all since SaveCharEquip is a very expensive operation
+    auto lastSwapTime   = PChar->GetLocalVar("core-LastGearSwap");
+    auto timeNowSeconds = std::chrono::time_point_cast<std::chrono::seconds>(server_clock::now());
+    if (lastSwapTime == 0 || (timeNowSeconds.time_since_epoch().count() - lastSwapTime) > 30)
```
WinterSolstice8, line 3499: Free optimization: lockstyled players will never need look updates unless their main/sub ranged/ammo change to different skill types

src/map/packet_system.cpp
WinterSolstice8: Why not use a variable to store this in CCharEntity over using localvars?

src/map/packet_system.cpp
MowFord: Yes looks like savecharlook doesn't return if lockstyled, so we could add a check here. Though i'm not sure how weapons fall into the equation. savecharlook is very cheap, and the way I have this is that it only runs savecharlook if it's been 30s since your last gearswap (if you gearswap every 10s doing random things the call won't happen)

MowFord:
My mistake about packet_system, i could have sworn i read about it being multi-threaded at some point, thanks for the reading

src/map/packet_system.cpp
MowFord: It's contained within one `.cpp` file seems less complex to use a localvar.

zach2good:
I remember writing some code that looked very similar to this for limiting auction actions or something a few years ago, might be worth extracting the whole thing out into a helper so it can be re-used more easily:
```cpp
// Definition
void localVarTimeGateMS(CCharEntity* PChar, std::string key, uint32 ms, std::function<void()> func)
{
    auto lastCheckTime = PChar->GetLocalVar(key.c_str());
    auto timeNowMs     = std::chrono::time_point_cast<std::chrono::milliseconds>(server_clock::now());

    if (lastCheckTime == 0 || (timeNowMs.time_since_epoch().count() - lastCheckTime) > ms)
    {
        func();
    }
    
    PChar->SetLocalVar(key.c_str(), static_cast<uint32>(timeNowMs.time_since_epoch().count()));
}

// Usage
localVarTimeGateMS(PChar, "lastCheckTime", 1000, [&]()
{
    charutils::SaveCharLook(PChar);
});
```

Names are a bit crap, but the idea is there

MowFord:
> I remember writing some code that looked very similar to this for limiting auction actions or something a few years ago, might be worth extracting the whole thing out into a helper so it can be re-used more easily:
> 
> ```c++
> // Definition
> void localVarTimeGateMS(CCharEntity* PChar, std::string key, uint32 ms, std::function<void()> func)
> {
>     auto lastCheckTime = PChar->GetLocalVar(key.c_str());
>     auto timeNowMs     = std::chrono::time_point_cast<std::chrono::milliseconds>(server_clock::now());
> 
>     if (lastCheckTime == 0 || (timeNowMs.time_since_epoch().count() - lastCheckTime) > ms)
>     {
>         func();
>     }
>     
>     PChar->SetLocalVar(key.c_str(), static_cast<uint32>(timeNowMs.time_since_epoch().count()));
> }
> 
> // Usage
> localVarTimeGateMS(PChar, "lastCheckTime", 1000, [&]()
> {
>     charutils::SaveCharLook(PChar);
> });
> ```
> 
> Names are a bit crap, but the idea is there

pretty slick, do you want to PR that into some global util (or would this be strictly inside charutils i guess?) then i can rebase and update my code?

MowFord:
Updated per the requests and compiled/tested again

src/map/packet_system.cpp
WinterSolstice8: The goal of this PR is optimization, no? using localvars is objectively slower than a direct access for no real reason unless for some reason we need this in lua, at which point we'd write a function hook to that variable

MowFord: I'll defer to your judgement on that one. I might need some hand-holding to do what you're suggesting, though. Are you saying add a private member in `charentity.h` and save the timestamp there instead?

WinterSolstice8: Yeah, that's the ticket

WinterSolstice8:
Ino talked to us about this PR, it seems the aforementioned `CHAR_PERSIST` is already available to use in LSB but just hasn't been implemented. Would you like to switch over to that here? I could make another PR that does the job as well -- it's well tested on Eden and has even fewer DB calls than what we have here.

WinterSolstice8:
See #4773, i tested it out and seems to work great.

Xaver-DaRed:
With the merge of PR #4773 this changes are no longer necesary. Still, thanks for the effort.

WinterSolstice8:
> With the merge of PR #4773 this changes are no longer necesary. Still, thanks for the effort.

Moreover, it brought out an issue I didn't know we had. So for that it was actually really important to begin with. Thanks.

MowFord:
Thanks all. I don't mind who gets attribution to what, just happy to be part of making the project better.

PR #4756, opened 2023-11-21
Followup to getSpeed() updates

MowFord:
Updates to augments, item latents, and mob spawn mods

**_I affirm:_**

Was auditing our modules to prepare for the updates from https://github.com/LandSandBoat/server/pull/4729 and found some missing things from the rework to use new mods:
- movement speed augment should use gear movement mod
- item latents updated to use gear movement for positive values and movement stackable for negative
- mob family mods updated to use movement stacking mod

Merge in the new getSpeed rework and see that items such as nation aketon properly gear speed stack with other items

### DIFF
--- a/sql/mob_spawn_mods.sql
+++ b/sql/mob_spawn_mods.sql
@@ -605,16 +605,13 @@ INSERT INTO `mob_spawn_mods` VALUES (17236201,55,180,1); -- IDLE_DESPAWN: 180
 -- Boreal Hound
 INSERT INTO `mob_spawn_mods` VALUES (17236202,12,15,1);  -- DRAW_IN: 15
 INSERT INTO `mob_spawn_mods` VALUES (17236202,160,50,0); -- DMG: 50
-INSERT INTO `mob_spawn_mods` VALUES (17236202,169,10,0); -- MOVE: 10
 
 -- Boreal Coeurl
 INSERT INTO `mob_spawn_mods` VALUES (17236203,12,15,1);  -- DRAW_IN: 15
 INSERT INTO `mob_spawn_mods` VALUES (17236203,23,8,1);   -- IMMUNITY: 8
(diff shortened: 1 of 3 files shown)


WinterSolstice8 (approved):

Xaver-DaRed (approved):
Ooops. Oh well

sql/mob_spawn_mods.sql
```diff
@@ -605,16 +605,16 @@ INSERT INTO `mob_spawn_mods` VALUES (17236201,55,180,1); -- IDLE_DESPAWN: 180
 -- Boreal Hound
 INSERT INTO `mob_spawn_mods` VALUES (17236202,12,15,1);  -- DRAW_IN: 15
 INSERT INTO `mob_spawn_mods` VALUES (17236202,160,50,0); -- DMG: 50
-INSERT INTO `mob_spawn_mods` VALUES (17236202,169,10,0); -- MOVE: 10
+INSERT INTO `mob_spawn_mods` VALUES (17236202,75,10,0); -- MOVE_SPEED_STACKABLE: 10
```
TeoTwawki: actually these should not _exist_

mobs were never supposed to use what was there before anyway. please delete.

the ones in scripts will need nuked later as well - mobs should be having their speeds correctly set in their sql, or calling `setSpeed()` when they need to change on the fly or when sql isn't an option.

TeoTwawki (changes requested):
can just straight delete the spawn mods- they were never correct before anyway

sql/mob_spawn_mods.sql
Xaver-DaRed: We should really just move all the mods here to their mob scrits and be done with this table.
That's out of the scope of this PR though, of course.

TeoTwawki: yeah but the speed ones don't belong there either. the rest tho, agreed death to this table.

TeoTwawki (approved):

PR #4763, opened 2023-11-23
[SQL] Remove flag on Quickening effect

MowFord:
**_I affirm:_**

Quickening, for a long time it seems, has had the flag `EFFECTFLAG_MAGIC_END` (0x80 = 128). This PR removes that. Hard to prove as absence of evidence is not evidence of absence, but hopefully someone with retail access can confirm. should be simple to test

Before changes, addeffect quickening and cast any spell. You will lose quickening.

After changes, see that it stays

Notice the buff still in effects bar after casting and zoning
image

### DIFF
--- a/sql/status_effects.sql
+++ b/sql/status_effects.sql
@@ -210,7 +210,7 @@ INSERT INTO `status_effects` VALUES (172,'intension',40,0,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (173,'dread_spikes',41,34,0,0,0,0,0,0,800);
 INSERT INTO `status_effects` VALUES (174,'magic_acc_down',16418,0,0,0,0,0,1,0,0);
 INSERT INTO `status_effects` VALUES (175,'magic_atk_down',16418,0,0,0,0,0,4,0,0);
-INSERT INTO `status_effects` VALUES (176,'quickening',4194476,0,0,0,0,0,0,0,0);
+INSERT INTO `status_effects` VALUES (176,'quickening',4194348,0,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (177,'encumbrance',8388608,0,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (178,'firestorm',4194336,0,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (179,'hailstorm',4194336,0,0,0,0,0,0,0,0);


TeoTwawki:
Likely just someone's math mistake (some folks are hex2dec challanged). Or a misunderstanding about casing _on a mob_ would would trigger a mob to be agro and constitute an attack - and you do lose it on attacking. Buffing/healing yourself etc. doesn't matter.

Anyway, can confirm via use of my Talaria, which function identically to sprinters boots but never run out of charge:

Xaver-DaRed (dismissed):

TeoTwawki:
follow-up: yeah somebody messed up their math big time. its also flagged as synth support.

sql/status_effects.sql
```diff
@@ -210,7 +210,7 @@ INSERT INTO `status_effects` VALUES (172,'intension',40,0,0,0,0,0,0,0,0);
 INSERT INTO `status_effects` VALUES (173,'dread_spikes',41,34,0,0,0,0,0,0,800);
 INSERT INTO `status_effects` VALUES (174,'magic_acc_down',16418,0,0,0,0,0,1,0,0);
 INSERT INTO `status_effects` VALUES (175,'magic_atk_down',16418,0,0,0,0,0,4,0,0);
-INSERT INTO `status_effects` VALUES (176,'quickening',4194476,0,0,0,0,0,0,0,0);
+INSERT INTO `status_effects` VALUES (176,'quickening',4194348,0,0,0,0,0,0,0,0);
```
TeoTwawki, line 213: somebody meant to have job change in here, but they missed a zero before converting to decimal and wound up with synth support instead

changing this from
`40002C` (`4194348`)
to
`400002C` (`67108908`)

TeoTwawki: also if its not supposed to be copy-able by drg's empathy 2c at the end becomes 25.

..and if generic damage when you weren't the target can take it then `EFFECTFLAG_DAMAGE = 0x00000010` needs added

I can test neither of these right now. but I'll accept the missing zero as "good enough" for this pr.

MowFord: I'm slow at hex math. But yes it should wear on damage, which I thought it did... I will revisit later tonight thanks!

Xaver-DaRed: `EFFECTFLAG_ON_JOBCHANGE = 0x00400000`
This would mean theres no need to change that concrete one.

https://github.com/LandSandBoat/server/blob/base/src/map/status_effect.h#L68C1-L68C84

As for the other 2 cases, we can leave for now until we go test them

Xaver-DaRed (approved):

sql/status_effects.sql
TeoTwawki: /shrug

windows calculator, hex mode, add flags, then switch decimal. copy paste.

PR #4764, opened 2023-11-23
[Magic] Update aspir/drain spell floors to be last

MowFord:
cap amount after all adjustments are made
return 0 when target is undead

**_I affirm:_**

When using drain spells, you cannot do more damage than the mob has HP. Due to the order of things before this PR, often times the resultant damage for a mob at low hp would be less than the mob's total hp due to the call of `target:magicDmgTaken`, which considers `UDMGMAGIC`, etc

This PR does two things:
- reorders the magic adjustments to do the cap of dmg after all other things are considered
  - This is done for drain and aspir for consistency
- returns 0 when target is undead for aspir/drain

Find a mob somewhat resistant to drain, drain it and see how much damage you do, set its hp to 10 and drain it, see that you _do_ kill it

### DIFF
--- a/scripts/actions/spells/black/aspir.lua
+++ b/scripts/actions/spells/black/aspir.lua
@@ -27,15 +27,15 @@ spellObject.onSpellCast = function(caster, target, spell)
     dmg = adjustForTarget(target, dmg, spell:getElement())
     --add in final adjustments
 
+    dmg = dmg * xi.settings.main.DARK_POWER
+
     if dmg < 0 then
         dmg = 0
     end
 
-    dmg = dmg * xi.settings.main.DARK_POWER
-
     if target:isUndead() then
         spell:setMsg(xi.msg.basic.MAGIC_NO_EFFECT) -- No effect
-        return dmg
+        return 0
     end
 
     if target:getMP() > dmg then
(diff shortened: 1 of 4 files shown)


TeoTwawki (approved):
lgtm

Xaver-DaRed (approved):

EpicTaru (approved):

RAIST5150:
"...set its hp to 10 and drain it, see that you do not kill it"

A bit confused by this statement... is this something new in retail?  Seem to recall draining things to death on BLM back when I actively played in retail.

To clarify, situationally with resists, yes it could fail to kill.  Just wanting to make sure whether this was meant to say that you flat out can not drain something to death as a rule now.

WinterSolstice8:
> "...set its hp to 10 and drain it, see that you do not kill it"
> 
> A bit confused by this statement... is this something new in retail? Seem to recall draining things to death on BLM back when I actively played in retail.
> 
> To clarify, situationally with resists, yes it could fail to kill. Just wanting to make sure whether this was meant to say that you flat out can not drain something to death as a rule now.

It's just bad/copypasted wording. This PR fixes that problem. I'll update the PR text.

WinterSolstice8 (approved):

zach2good (approved):

PR #4768, opened 2023-11-24
[Fix] Remove incorrect usage of SLEEPRES and LULLABYRES modifier

Xaver-DaRed:
**_I affirm:_**

This are remnants of the old distinction done between Light sleep and dark sleep as if they were 2 completely different things.

With the introduction of resistance ranks, mobs having the correct elemental resistance rank will naturally have resistance against one or the other.

Also, this modifiers are supposed to be used for the job traits `Resist status` which trigger the special `Resist! <target> resists the effects of the spell.` message.

Cast Repose or lullaby on skelletons and said kind of mobs. Watch it work.


TeoTwawki (approved):
really need a traits table for mobs anyway, instead of adding raw modifiers.

`#someday`

EpicTaru (approved):

PR #4770, opened 2023-11-25
Adjust warp timings.

### LINKED ISSUE #4769: 🐛 Warp ring animation doesn't play completely
**_I affirm:_**

Ubuntu

## Branch affected by issue 

`base`

## Steps to reproduce

with 1 character:
use warp ring
see animation, animation wont finish playing (will play for like a split of a second), screen will go black to loading screen

Its more noticeable if you have 2 characters:
Have 2 characters  logged in.
have 1 of them use the warp ring.
The warp ring animation is so short that the warping effect doesn't play completely, char just almost disappears

## Expected behavior

Use warp ring.
see animation
allow a couple seconds before screen goes black (no fade to black)
https://youtu.be/zTLLugHJ6zo?si=LhXRvgLDctQxLxlr&t=1205

TeoTwawki:
**_I affirm:_**

Retail takes just short of 3 seconds to animate this before warping you. It's closer to 2800ms but the code expects the answer to be in full seconds, so I am using 3 (3000ms).

That's not far off enough for a human to really notice considering there is also the somewhat random lag factor of packets going to japan and back on retail making _exactness_ tough to pin down. My estimate could be off by up to 300ms (so it could actually be 2500ms at the **lowest** - any faster is going to be abnormal).

Resolves #4769

`!additem warp_cudgel`
`additem scroll_of_instant_warp`
`additem warp_ring`

Try them out. see that the animation plays in full before you are zoned.

Go to any grounds of Valor tome and try "homing instinct" from the support menu. see the same there.

### DIFF
--- a/scripts/globals/additional_effects.lua
+++ b/scripts/globals/additional_effects.lua
@@ -248,7 +248,7 @@ xi.additionalEffect.attack = function(attacker, defender, baseAttackDamage, item
 
     elseif addType == procType.SELF_BUFF then
         if addStatus == xi.effect.TELEPORT then -- WARP
-            attacker:addStatusEffectEx(xi.effect.TELEPORT, 0, xi.teleport.id.WARP, 0, 1)
+            attacker:addStatusEffectEx(xi.effect.TELEPORT, 0, xi.teleport.id.WARP, 0, 3)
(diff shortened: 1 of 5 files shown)


TeoTwawki:
I should prolly audit teleport speeds next

claywar (approved):

PR #4783, opened 2023-11-30
Multihit WS & OaT jump audit

WinterSolstice8:
**_I affirm:_**

Audits usage of multihit WS such as follows:

Mainhand, offhand, mainhand extra WS hits (>1 hit WS), mainhand DA/TA/QA/etc procs, offhand DA/TA/QA/etc procs
All WS can proc DA/TA/QA etc twice, but the second proc requires >= 2 base hits on WS or an offhand
Offhand hits now use offhand hit damage on DA/TA/QA etc
WSD no longer applies to Jump.
Skillups on offhand hand now use the appropriate skill if different.

Use weaponskills, see stuff look pretty normal.
Use Jump and see DA/TA/QA and OaX work properly with normal TP returns.


scripts/globals/weaponskills.lua
```diff
@@ -216,51 +215,35 @@ local function getMultiAttacks(attacker, target, wsParams)
...
-    -- whereas dual wiedling gets 2 "initial" hits, but `numHits` for e.g. dancing edge is respected as though only one "initial" was performed
-    -- tl;dr: DW gives an extra WS hit everytime, H2H simply ensures every WS gets at least 2 hits (yes backhand blow and dragon kick are both 2-hit ws...)
     numHits = numHits + bonusHits
-    if attacker:getOffhandDmg() > 0 then
-        numHits = numHits + 1
-    end
 
-    -- cap total hits at 8
-    return utils.clamp(numHits, 1, 8)
+    return numHits
```
Xaver-DaRed, line 246: No clamp needed?

WinterSolstice8: Clamp is moved to the hit count loops, it's not possible for that function to return > 4 anyway

Xaver-DaRed (dismissed):

WinterSolstice8:
I just realized that there's a potential bug with TP gain here, drafted

WinterSolstice8:
Please re-review, should be good to go.

Xaver-DaRed (approved):

PR #4786, opened 2023-12-01
[SQL][Items] Item mods 23209-23214

helixhamin:
**_I affirm:_**

1. Adds item mods to items 23209-23214.
2. Fixes item 11672. The mod was incorrect, since the mod used was Magic Burst Bonus, but the item actually gives Magic Burst Bonus 2. However, since MAG_BURST_BONUS_II is not yet implemented, it was changed to a TODO item. The difference is that MBB2 is uncapped while MBB is capped.
3. Fixed item 23217. Many mods were incorrect. These  have been corrected.
4. Fixed TODO label (Change to "TODO:") for items 26109 and 26200.

As GM, !additem 23209|23210|23211|23212|23213|23214|23217
Equip item and check to make sure proper stats were added.

### DIFF
--- a/sql/item_mods.sql
+++ b/sql/item_mods.sql
@@ -8131,7 +8131,7 @@ INSERT INTO `item_mods` VALUES (11671,25,3); -- ACC: 3
 
 -- Mujin Band
 INSERT INTO `item_mods` VALUES (11672,175,5); -- SKILLCHAINDMG: 5
-INSERT INTO `item_mods` VALUES (11672,487,5); -- MAG_BURST_BONUS: 5
+-- TODO: MAG_BURST_BONUS_II: 5
 
 -- Demonry Ring
 INSERT INTO `item_mods` VALUES (11673,23,7);   -- ATT: 7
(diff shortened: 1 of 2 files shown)


sql/item_mods.sql
```diff
@@ -8131,7 +8131,7 @@ INSERT INTO `item_mods` VALUES (11671,25,3); -- ACC: 3
 
 -- Mujin Band
 INSERT INTO `item_mods` VALUES (11672,175,5); -- SKILLCHAINDMG: 5
-INSERT INTO `item_mods` VALUES (11672,487,5); -- MAG_BURST_BONUS: 5
+-- TODO: MAG_BURST_BONUS_II: 5
```
Xaver-DaRed, line 8134: Could you expand on this?

I ask because I could add support for it with enough info.

sql/item_mods.sql
```diff
@@ -58092,7 +58211,7 @@ INSERT INTO `item_mods` VALUES (26108,165,5); -- CRITHITRATE: 5
 -- Snotra Earring
 INSERT INTO `item_mods` VALUES (26109,13,8);  -- MND: 8
 INSERT INTO `item_mods` VALUES (26109,30,10); -- MACC: 10
--- TODO Enfeeble Magic Duration: +10%
+-- TODO: Enfeeble Magic Duration: +10%
```
Xaver-DaRed, line 58214: I thought we had a modifier for this already. However, easy to implement aswell.

This is a note for "future me", btw, feel free to ignore this particular comment.

sql/item_mods.sql
helixhamin: I think the only difference between the Magic Burst Damage and Magic Burst Damage II is:
Exceeds the 40% [Magic Burst](https://www.bg-wiki.com/ffxi/Magic_Burst) cap
Adding in MBBII to mods would allow 5 of the current 85 TODO items in item_mods.sql to be completed.

sql/item_mods.sql
helixhamin: Unfortunately, at this point, these are the only enfeeble mods I see:
```
	Line  165:     ENFEEBLE  = 114, // Enfeebling Magic Skill
	Line  301:     // Resistances to enfeebles - Job Traits/Job Abilities/Atmas/Items/Gear
	Line  320:     // MEVA bonus to enfeebles (Bar-Effect, for example. And modifiers in mobs)
```

Xaver-DaRed (approved):

PR #4787, opened 2023-12-01
[AMK][Lua] Added check and reminder text to AMK 6, enabled repeatability

Flibe-XI:
**_I affirm:_**

- added check to make sure player has at least one orb (cardian flavor orbs) before initiating cardian fight in amk mission 6
- added flavor text when on the mission but player has no orbs
- enabled ability to repeat the mission after completing it
- found an off-by-one error in the zone text ids

`!addmission 10 5`
`!gotoid 17572302`
Touch QM, it should tell player that an orb could cause something to happen
`!addkeyitem orb_of_swords`
Touch QM again, it should start the fight prompt
`!addmission 10 6`
Touch QM again, it should still start the fight prompt if player has an orb and doesn't have the ripe_starfruit KI


zach2good:
Just how repeatable is this? What are the mechanics for these key items dropping from local Cardians? Are the cardians ALWAYS meant to drop these items and you can come back later if you've bought them again from Squintox? Or are the drop mechanics for these KIs not implemented yet?

zach2good (approved):

zach2good:
These two should land at the same time: https://github.com/LandSandBoat/server/pull/4800

PR #4788, opened 2023-12-01
[LUA] Drain/Aspir undead fixes

MowFord:
- Moved undead check to the top so we don't quietly deal dmg to undead
- save target's HP before finalMagicAdjustments to properly display the drain message and heal caster

**_I affirm:_**

Unfortunate oversight in my previous PR for drain/aspir: undead should exit earlier.

In its current form, Drain actually does damage to undead, then displays the message about having no effect (doesn't heal the player at least)

For consistency i've put the aspir check at the top as well.

Also, when killing a mob with drain, the spell wasn't healing the player due to the fact that `finalMagicAdjustments` actually deals damage. Fixed to save the target's HP before calling that and tested all the permutations in kuftal on crabs and ghosts.

Also `finalMagicAdjustments` has checks to heal the target if `dmg < 0`, so moved this check above the call

drain a ghost and see that its hp stays full (would go down before this PR)
`!hp 50` a mob that's engaged with you and drain it to see you gain 50 hp and the mob dies

### DIFF
--- a/scripts/actions/spells/black/drain.lua
+++ b/scripts/actions/spells/black/drain.lua
@@ -9,9 +9,15 @@ spellObject.onMagicCastingCheck = function(caster, target, spell)
 end
 
 spellObject.onSpellCast = function(caster, target, spell)
+    if target:isUndead() then
+        spell:setMsg(xi.msg.basic.MAGIC_NO_EFFECT) -- No effect
+        return 0
+    end
+
     --calculate raw damage (unknown function  -> only dark skill though) - using http://www.bluegartr.com/threads/44518-Drain-Calculations
     -- also have small constant to account for 0 dark skill
     local dmg = 10 + (1.035 * caster:getSkillLevel(xi.skill.DARK_MAGIC))
+    local targetHP = target:getHP()
 
     if dmg > (caster:getSkillLevel(xi.skill.DARK_MAGIC) + 20) then
         dmg = (caster:getSkillLevel(xi.skill.DARK_MAGIC) + 20)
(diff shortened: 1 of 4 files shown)


Xaver-DaRed (approved):

TeoTwawki (approved):

scripts/actions/spells/black/drain.lua
```diff
@@ -30,20 +36,16 @@ spellObject.onSpellCast = function(caster, target, spell)
...
 
     if dmg < 0 then
         dmg = 0
     end
 
-    if target:getHP() < dmg then
-        dmg = target:getHP()
-    end
+    --add in final adjustments and deal damage
+    dmg = finalMagicAdjustments(caster, target, spell, dmg)
```
TeoTwawki, line 45: er..crap. I just hit merge and didn't realize you have this after the check for negative dmg. Adjustments to the damage happen inside this function, so that's potentially problematic. Can you followup?

MowFord: but drain should never be negative, no? that's the point of that check in this function

TeoTwawki: thats why we normally have the check for it being neagive coming after everything else. maybe finalMagicAdjustments has no chance of turning it negative - but I don't know that at a glance unless I have that function memorized.

TeoTwawki: is there any instance of a mob being healed by drain? would that..unheal the player? At a minimum, I'd want to explain in comment in file.

MowFord: I believe vanilla custard in nyzul absorb dark. Without that check in LSB i imagine you'd drain it and damage yourself. Kinda funny, would be interested to know what happens in retail

PR #4791, opened 2023-12-02
DNC job ability: Contradance

MowFord:
- Doubles curing power of next waltz
- Converts Healing Waltz to AoE

**_I affirm:_**

Codes the DNC ability: [Contradance](https://www.bg-wiki.com/ffxi/Contradance)

Simple enough. Though one small annoyance is getting the bonus for all targets of divine waltz (aoe cure)

I did it via a timer to delete the effect 100ms after the aoe resolves all the targets. This works but unsure if there's a different way LSB would prefer it done.

image

image

image

Use the contradance ability, test waltzes for bonus and that the effect wears on use

### DIFF
--- a/scripts/actions/abilities/healing_waltz.lua
+++ b/scripts/actions/abilities/healing_waltz.lua
@@ -8,6 +8,12 @@
 local abilityObject = {}
 
 abilityObject.onAbilityCheck = function(player, target, ability)
+    if player:getStatusEffect(xi.effect.CONTRADANCE) then
+        ability:setAOE(1)
+        ability:setRange(10)
+        player:delStatusEffect(xi.effect.CONTRADANCE)
+    end
+
     local waltzCost = 200 - player:getMod(xi.mod.WALTZ_COST) * 10
     if target:getHP() == 0 then
         return xi.msg.basic.CANNOT_ON_THAT_TARG, 0
@@ -39,10 +45,13 @@ abilityObject.onAbilityCheck = function(player, target, ability)
     end
 end
 
-abilityObject.onUseAbility = function(player, target, ability)
+abilityObject.onUseAbility = function(player, target, ability, action)
     local waltzCost = 200 - player:getMod(xi.mod.WALTZ_COST) * 10
     -- Only remove TP if the player doesn't have Trance.
-    if not player:hasStatusEffect(xi.effect.TRANCE) then
+    if
+        not player:hasStatusEffect(xi.effect.TRANCE) and
+        action:getPrimaryTargetID() == target:getID()
+    then
         player:delTP(waltzCost)
     end
 
--- a/scripts/globals/job_utils/dancer.lua
+++ b/scripts/globals/job_utils/dancer.lua
@@ -460,9 +460,8 @@ xi.job_utils.dancer.useWildFlourishAbility = function(player, target, ability, a
     return 0
 end
 
--- TODO: Implement Contradance status effect.
 xi.job_utils.dancer.useContradanceAbility = function(player, target, ability)
-    -- player:addStatusEffect(xi.effect.CONTRADANCE, 19, 1, 60)
+    player:addStatusEffect(xi.effect.CONTRADANCE, 19, 1, 60)
 end
 
 xi.job_utils.dancer.useWaltzAbility = function(player, target, ability, action)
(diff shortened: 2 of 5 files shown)


MowFord:
animation is Divine Seal:
https://www.youtube.com/watch?v=7GEaEOCDuKY&t=833s

WinterSolstice8:
image
Packet from retail

scripts/actions/abilities/healing_waltz.lua
```diff
@@ -42,7 +49,10 @@ end
 abilityObject.onUseAbility = function(player, target, ability)
     local waltzCost = 200 - player:getMod(xi.mod.WALTZ_COST) * 10
     -- Only remove TP if the player doesn't have Trance.
-    if not player:hasStatusEffect(xi.effect.TRANCE) then
+    if
+        not player:hasStatusEffect(xi.effect.TRANCE) and
+        player:getLocalVar('waltzTarget') == target:getID()
```
WinterSolstice8: I don't like the use of localvar -- we can remove it

Change the onUseAbility signature to `abilityObject.onUseAbility = function(player, target, ability, action)`
and then check like this:

```
    if
        not player:hasStatusEffect(xi.effect.TRANCE) and
        action:getPrimaryTargetID() == target:getID()
    then
    ...
    end
```

scripts/globals/job_utils/dancer.lua
```diff
@@ -493,6 +492,13 @@ xi.job_utils.dancer.useWaltzAbility = function(player, target, ability, action)
     amtCured = (target:getStat(xi.mod.VIT) + player:getStat(xi.mod.CHR)) * statMultiplier + waltzInfo[3]
     amtCured = math.floor(amtCured * (1.0 + (math.min(50, player:getMod(xi.mod.WALTZ_POTENCY)) / 100)))
     -- TODO: Account for Waltz Potency Received
+    if player:getStatusEffect(xi.effect.CONTRADANCE) then
+        amtCured = amtCured * 2
+        -- slight delay to allow the effect to apply to all targets of divine waltz
+        player:timer(100, function(playerArg)
+            playerArg:delStatusEffect(xi.effect.CONTRADANCE)
```
WinterSolstice8, line 500: Don't like this, but there's not much we can do about it. Maybe we need some kind of self-removing listener? I think it's okay to leave it in like this for now with a TODO to improve it

scripts/actions/abilities/healing_waltz.lua
WinterSolstice8: onUseAbility is already fed a CLuaAction so no extra modification is needed
```
        sol::function onUseAbility = GetCacheEntryFromFilename(filename)["onUseAbility"];
        if (!onUseAbility.valid())
        {
            ShowWarning("luautils::onUseAbility - Ability %s not found.", PAbility->getName());
            return 0;
        }

        auto result = onUseAbility(CLuaBaseEntity(PUser), CLuaBaseEntity(PTarget), CLuaAbility(PAbility), CLuaAction(action));
        ```

File: scripts/globals/job_utils/dancer.lua (hunk 2 above)
Comment (TeoTwawki) [reply scripts/globals/job_utils/dancer.lua:500]: https://github.com/LandSandBoat/server/blob/ece956579adf1d2f9aa7f97552e2637a070a132c/src/map/utils/battleutils.cpp#L5675

Need something like that, but for abilities instead of spells. There is no correct way to do this in lua because it should not be happening in lua.

File: scripts/actions/abilities/healing_waltz.lua (hunk 1 above)
Comment (MowFord) [reply scripts/actions/abilities/healing_waltz.lua]: Sounds good. I can fix a couple other AOE job abilities like this

Comment (MowFord) [reply scripts/actions/abilities/healing_waltz.lua]:
```
if
            abilityId == xi.jobAbility.DIVINE_WALTZ or
            abilityId == xi.jobAbility.DIVINE_WALTZ_II
        then
            if player:getID() == target:getID() then
                player:delTP(waltzCost)
            end
```

My mistake divine waltz is a self-only so this logic works just fine, but yep your solution works for healing waltz

MowFord: image

image

all still workie

scripts/globals/job_utils/dancer.lua
MowFord: BTW a TODO was added for this. Unsure what the hold label was added for

TeoTwawki: > Unsure what the hold label was added for

_point back at own previous comment_

was so nobody merged ahead of conferring on if this was acceptable temporarily, exactly how temporary that would be, and who if anybody would be coming back to do it right. Worth mentioning that temporary measures are technical debt, and past iterations of the project added more technical debt than the GDP of some real world nations.

@claywar  is our resident DNC expert I think. If He's ok with this I'll remove the hold I placed.

MowFord: I realize that this runs for every target, would it be more appropriate to set the contradance effect duration to zero instead of a timer (still will run for every target) so it falls off immediately after the targetfind loop?

PR #4793, opened 2023-12-02
[core] Improve Auras

WinterSolstice8:
**_I affirm:_**

* Reduce `new` CStatusEffects on every update of Auras through means of tracking auras through a new effect flag
* Improve retail accuracy by constantly renewing the expiry time of the buff

Note: this may be followed up with future improvements, there may be some extreme corner case oddities moving from the same effect aura into another where power doesn't change. It's 3:30 AM give me a break :)

Use auras on mobs/players and see them work.

example: `!exec target:addStatusEffectEx(612, 612, 6, 3, 0, xi.effect.POISON, 4, 1, xi.effectFlag.AURA)` and get poisoned, see that the poison aura does _not_ flicker but stays constantly in a "going to wear off" state

### DIFF
--- a/scripts/enum/effect_flag.lua
+++ b/scripts/enum/effect_flag.lua
@@ -35,4 +35,5 @@ xi.effectFlag =
     AURA            = 0x04000000,
     HIDE_TIMER      = 0x08000000,
     ON_ZONE_PATHOS  = 0x10000000,
+    ALWAYS_EXPIRING = 0x20000000,
 }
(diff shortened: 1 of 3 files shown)


WinterSolstice8:
Silence Aura from Scylla shows the timer is _increasing_ meaning the server must be updating it's expiry
image
image

zach2good (approved):
Loving that proof, can you also test this with GEO and the aura-giving trusts? Then this is good to go

Xaver-DaRed (approved):

PR #4800, opened 2023-12-03
[AMK][Lua] Cardian orb drop mechanic

Flibe-XI:
**_I affirm:_**

Adds a function to cardians in outer horutoto ruins that allows them to drop their respective key items "orb of sword/baton/coin/cup" for AMK mission 6. 

`!addmission 10 5`
Kill cardians in outer horutoto ruins until they cause the player to obtain their key items.  If the player isn't on the right mission, or already has the orb, the orb won't drop and no error is caused.

The drop rate has very little information other than the level of the mob is involved, and the higher the level, the higher the droprate. Some sources say drop rate on lower level mobs is very poor, so the range is set it at is between 4 - 25.8%, which increases with size of the party as well.

image

### DIFF
--- a/scripts/missions/amk/helpers.lua
+++ b/scripts/missions/amk/helpers.lua
@@ -58,9 +58,8 @@ end
 -- The rest of the players in alliance get the same outcome as the first
 xi.amk.helpers.cardianOrbDrop = function(mob, player, orb)
     if
-        player:getCurrentMission(xi.mission.log_id.AMK) < xi.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE or
         player == nil or
-        orb == nil
+        player:getCurrentMission(xi.mission.log_id.AMK) < xi.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE
     then
         return
     end
--- a/scripts/missions/amk/helpers.lua
+++ b/scripts/missions/amk/helpers.lua
@@ -52,6 +52,45 @@ xi.amk.helpers.helmTrade = function(player, helmType, broke)
     end
 end
 
+-- AMK 6 (index 5) - An Errand! The Professor's Price
+-- Cardian orb KI logic: All or nothing drop of the orb is handled by
+-- a local var that is set once by the first player is called by onMobDeath
+-- The rest of the players in alliance get the same outcome as the first
+xi.amk.helpers.cardianOrbDrop = function(mob, player, orb)
+    if
+        player:getCurrentMission(xi.mission.log_id.AMK) < xi.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE or
+        player == nil or
+        orb == nil
+    then
+        return
+    end
+
+    -- Chance of drop increases both based on size of party and level of mob killed
+    if mob:getLocalVar('Mission[10][5]cardianOrbDrop') == 0 then
+        local partySize = 0
+        for _, member in pairs(player:getAlliance()) do
+            if member:getZoneID() == xi.zone.OUTER_HORUTOTO_RUINS then
+                partySize = partySize + 1
+            end
+        end
+
+        -- Chance ranges from 4% to 25.8% (based on max mob lvl of 44)
+        local dropChance = (3 * mob:getMainLvl()) + (30 * utils.clamp(partySize, 1, 6)) - 10
+        local roll = math.random(1000)
+
+        if roll < dropChance then
+            mob:setLocalVar('Mission[10][5]cardianOrbDrop', 1)
+        else
+            mob:setLocalVar('Mission[10][5]cardianOrbDrop', 2)
+        end
+    end
+
+    -- Give orb
+    if mob:getLocalVar('Mission[10][5]cardianOrbDrop') == 1 then
+        npcUtil.giveKeyItem(player, orb)
+    end
+end
+
 -- AMK 7 (index 6) - Select/lookup the digging zone
 local digZoneIds =
 {
--- a/scripts/zones/Outer_Horutoto_Ruins/mobs/Two_of_Swords.lua
+++ b/scripts/zones/Outer_Horutoto_Ruins/mobs/Two_of_Swords.lua
@@ -1,12 +1,11 @@
 -----------------------------------
 -- Area: Outer Horutoto Ruins
---  Mob: Five of Coins
+--  Mob: Two of Swords
 -----------------------------------
 local entity = {}
 
 entity.onMobDeath = function(mob, player, optParams)
-    xi.amk.helpers.cardianOrbDrop(mob, player, xi.ki.ORB_OF_COINS)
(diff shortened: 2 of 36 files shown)


zach2good:
Well well well, the questions from my other PR comment are answered :D

scripts/missions/amk/helpers.lua
```diff
@@ -52,6 +52,44 @@ xi.amk.helpers.helmTrade = function(player, helmType, broke)
     end
 end
 
+-- AMK 6 (index 5) - An Errand! The Professor's Price
+-- Cardian orb KI logic: All or nothing drop of the orb is handled by
+-- a local var that is set once by the first player is called by onMobDeath
+-- The rest of the players in alliance get the same outcome as the first
+xi.amk.helpers.cardianOrbDrop = function(mob, player, orb)
+    if player:getCurrentMission(xi.mission.log_id.AMK) < xi.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE then
```
zach2good: If the cardian dies from DoT, player will be nil - should be checking for that

scripts/missions/amk/helpers.lua
```diff
@@ -52,6 +52,44 @@ xi.amk.helpers.helmTrade = function(player, helmType, broke)
...
+    -- Chance of drop increases both based on size of party and level of mob killed
+    if mob:getLocalVar('Mission[10][5]cardianOrbDrop') == 0 then
+        local partySize = 0
+        for _, member in pairs(player:getAlliance()) do
+            if member:getZoneID() == xi.zone.OUTER_HORUTOTO_RUINS then
+                partySize = partySize + 1
+            end
+        end
+
+        -- Chance ranges from 4% to 25.8% (based on max mob lvl of 44)
```
zach2good, line 76: 👍

zach2good (dismissed):
As a note to passers-by, another way of doing this would have been to create and apply the `cardian` family mixin to all these files - resulting in the same one-line-per-file change. These mobs aren't particularly special, so there's little difference in either approach. The meat is in the new helper.

scripts/zones/Outer_Horutoto_Ruins/mobs/Two_of_Swords.lua
```diff
@@ -0,0 +1,12 @@
+-----------------------------------
+-- Area: Outer Horutoto Ruins
+--  Mob: Two of Swords
+-----------------------------------
+local entity = {}
+
+entity.onMobDeath = function(mob, player, optParams)
+    xi.amk.helpers.cardianOrbDrop(mob, player, xi.ki.ORB_OF_SWORDS)
+    xi.regime.checkRegime(player, mob, 665, 3, xi.regime.type.GROUNDS)
```
TeoTwawki: Not seeing any regime pages that use Two of Swords
https://www.bg-wiki.com/ffxi/Grounds_of_Valor#Outer_Horutoto_Ruins

Watch what you copy and paste. Make sure the things you add actually belong in new mob scripts.

TeoTwawki (dismissed):

Flibe-XI:
> As a note to passers-by, another way of doing this would have been to create and apply the `cardian` family mixin to all these files - resulting in the same one-line-per-file change. These mobs aren't particularly special, so there's little difference in either approach. The meat is in the new helper.

Good to know thanks.

Also, I spoke with Xaver about possibly creating different mob_family_system ids for each type of cardian, as they are supposed to drop different crystals, but they are all the same element currently.  When they are all different families, the orb can be removed as an argument and the KI's can just be mapped to the family and determined by mob:getFamily() with the helpers function. Probably not worth the effort and is fine the way it is.

scripts/missions/amk/helpers.lua
```diff
@@ -52,6 +52,45 @@ xi.amk.helpers.helmTrade = function(player, helmType, broke)
...
 
+-- AMK 6 (index 5) - An Errand! The Professor's Price
+-- Cardian orb KI logic: All or nothing drop of the orb is handled by
+-- a local var that is set once by the first player is called by onMobDeath
+-- The rest of the players in alliance get the same outcome as the first
+xi.amk.helpers.cardianOrbDrop = function(mob, player, orb)
+    if
+        player:getCurrentMission(xi.mission.log_id.AMK) < xi.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE or
+        player == nil or
+        orb == nil
```
zach2good: This new change won't work, because you're checking the player's mission before you're checking if player is really there.
Lua properly has short-circuit logic, so you can just move the check for player or orb being nil before the mission check and you'll be fine.

https://www.tutorialspoint.com/execute_lua_online.php :
```lua
local player = nil

if player == nil or player:thisFunctionDoesntExist() == true then
    print("Short circuit success")
end
```

Prints correctly.

```lua
    if
        player == nil or
        player:getCurrentMission(xi.mission.log_id.AMK) < xi.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE or
```

You don't need to check whether or not orb is nil, because you're the one making the call and you know you've populated it

zach2good:
Don't worry about the family/crystal stuff, that's out of the scope of this PR

scripts/missions/amk/helpers.lua
Flibe-XI: D'oh!  Understood, thanks.  Will fix soon.

scripts/missions/amk/helpers.lua
```diff
@@ -57,11 +57,7 @@ end
 -- a local var that is set once by the first player is called by onMobDeath
 -- The rest of the players in alliance get the same outcome as the first
 xi.amk.helpers.cardianOrbDrop = function(mob, player, orb)
```
zach2good, line 59: What are the conditions for cardians to drop this key item? Can the item drop forever if you've completed AMK? Or do they only drop during this mission and if you want repeat KIs then you have to buy them from squintox?

Either way, your most recent commit removed the mission check all together

Flibe-XI: This is what happens when I try to rush something at work.  I'll focus and fix it when I get home.

Flibe-XI: image

I can't, however, find any information about whether or not the key items can drop before flagging mission 6 from shantotto.

zach2good: Easy to go get caps on retail

zach2good: For now, this condition seems fine: `player:getCurrentMission(xi.mission.log_id.AMK) < xi.mission.id.amk.AN_ERRAND_THE_PROFESSORS_PRICE`

zach2good (approved):

scripts/missions/amk/helpers.lua
Flibe-XI: I made a request in FFXI Captures discord. I don't have a retail account personally.

Xaver-DaRed (approved):

PR #4804, opened 2023-12-04
Implements Mythic WS quest climb scaling

MowFord:
**_I affirm:_**

Emulates the partial relic logic to apply a WS to nyzul base weapons only when the climb progress dictates the WS points are sufficient

Also updates Zalsuhn to scale the requirements based on climb progress.

The item mods were built via a SQL query to (hopefully) ensure accuracy.

Finally, the base weapons are set to the highest possible ws points. This shouldn't affect any other logic, as we don't need to track when the weapon is "broken" for anything other than the transient access to the WS. When the ws shows up in the menu, it's available to trade to Zalsuhn and fully unlock the WS for use with any weapon (regardless of if the module is in use or not).

Created a lua binding to reload the player's ws list for use in the weapon listeners

To summarize:
- trading a completed weapon to Zalsuhn:
  - This was already coded, but I've adjusted to use a helper function in `xi.equipment` to allow module overrides for the required ws points
  - I've also adjusted the structure of the related tables to encode more data and be more explicit
- Getting a WS to show up as a latent on the weapon while in the wild
  - This is done via a `WEAPONSKILL_STATE_EXIT` listener. It checks the current ws points against the helper function and adds a mod to the weapon to signify the ws is available (similar to relic weapons)
  - The listener function also verifies that the weapon is not in the offhand slot

Testing is 2 parts (as we know ws points work well):
 - `!addweaponskillpoints <slotid> 250` will give the ws points to be able to trade the item
   - https://imgur.com/Yvdz9n1
 - adding slightly less will allow you to test the listener functionality: 
   - https://imgur.com/j4IF5Kp


MowFord:
Was informed this is not how retail works anymore with the scaled ws points requirements... Bleh

I'm not sure a module would be able to do these things in cpp, so perhaps I can rework it to check _something_ to determine which method to use, and update the lua and SQL side to be a module?

MowFord:
I've pivoted to make this function the way it does in retail and also included a module to adjust the required weaponskill points for ~2014-era behavior

Note: the ws doesn't become available until the next weaponskill due to `addWeaponskillPoints` being after the listener call in the stack. I think this is a fine compromise.

MowFord:
rebased and confirmed it still works:

image

Edit: didn't think to re-compile. Thanks CI! Fixes incoming

Confirmed learning in the field by gaining points still works:

image

MowFord:
btw the source branch was deleted but afaik this is ready to merge. Was in use as-is on cexi for many months

PR #4808, opened 2023-12-04
[CPP] Set 60s cooldown on all blu spells

MowFord:
when adding or removing spell set

**_I affirm:_**

I was told at one point there was a cooldown for blu spells when setting (or removing) spells, but I couldn't find any remnants.

This code should be negligable performance-wise, and solves the problem 100%

Set blue spells, see 60s cooldown on using any blu spells
Remove blu spells, see same thing

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -7610,6 +7610,20 @@ void SmallPacket0x102(map_session_data_t* const PSession, CCharEntity* const PCh
                 ShowDebug("No match found. ");
             }
         }
+
+        // Regardless what the set spell action is, force recast on all currently-set blu spells
+        for (uint8 i = 0; i < 20; i++)
+        {
+            if (PChar->m_SetBlueSpells[i] != 0)
+            {
+                auto  spellId = static_cast<SpellID>(PChar->m_SetBlueSpells[i] + 0x200);
+                auto* PSpell  = spell::GetSpell(spellId);
+                if (CBlueSpell* PBlueSpell = dynamic_cast<CBlueSpell*>(PSpell))
+                {
+                    PChar->PRecastContainer->Add(RECAST_MAGIC, static_cast<uint16>(PBlueSpell->getID()), 60);
+                }
+            }
+        }
     }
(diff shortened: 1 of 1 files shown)


N3ckB3ard (approved):
I am officially starting a conspiracy theory that this logic was removed on LSB... I have no factual proof other than tears that this existed 
giphy

src/map/packet_system.cpp
```diff
@@ -7609,6 +7609,19 @@ void SmallPacket0x102(map_session_data_t* const PSession, CCharEntity* const PCh
             {
                 ShowDebug("No match found. ");
             }
+
+            // Regardless what the set spell action is, force recast on all currently-set blu spells
+            for (uint8 i = 0; i < 20; i++)
+            {
+                if (PChar->m_SetBlueSpells[i] != 0)
+                {
+                    CBlueSpell* BlueSpell = (CBlueSpell*)spell::GetSpell(static_cast<SpellID>(PChar->m_SetBlueSpells[i] + 0x200));
```
zach2good: The check for BlueSpell being nullptr is only likely to work if GetSpell returns nullptr, if it returns an incompatible spell then it'll succeed and strange things may happen. This is a use case for `dynamic_cast`:

```cpp
auto  spellId = static_cast<SpellID>(PChar->m_SetBlueSpells[i] + 0x200);
auto* PSpell  = spell::GetSpell(spellId);
if (CBlueSpell* PBlueSpell = dynamic_cast<CBlueSpell*>(PSpell))
{
    PChar->PRecastContainer->Add(RECAST_MAGIC, static_cast<uint16>(PBlueSpell->getID()), 60);
}
````

Also, prefer static cast to c-style casting

zach2good (changes requested):

src/map/packet_system.cpp
MowFord: Changes made, also noticed it was accidentally in the `then` section only applying it when you _set_ a spell, not when you _remove_ a spell.

Tested and confirmed both cases lock out blu spells for 60s

Thanks!

zach2good (approved):
Code is fine, I never played BLU so someone else will need to confirm if the logic is right

Xaver-DaRed (approved):

TeoTwawki:
YEET

Issue #4785, opened 2023-11-30
🐛 Penguin Ring's Enchantment unable to use


Aeltorian:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Windows10

## Branch affected by issue 

`base`

## Steps to reproduce

Obtain Penguin Ring. Attempt to use it. Find yourself unable to use. 

## Expected behavior

The [Penguin Ring](https://www.bg-wiki.com/ffxi/Penguin_Ring)'s enchantment is suppose greatly increases the rate of stamina drain of a fish while you try to reel it in. There is a modifier for it already. Ring is also labeled on LSB as R/EX. With the captures, it's only EX. 

![image](https://github.com/LandSandBoat/server/assets/115962399/427a59ba-9788-43c5-8178-d9bf80e4f5cf)

The ring will lose it's effect after 20 minutes, if you zone, or if you remove the ring. It will stack with [Albatross Ring](https://www.bg-wiki.com/ffxi/Albatross_Ring) and itself.

Sources: 
[BG-Wiki](https://www.bg-wiki.com/ffxi/Penguin_Ring)
[JP Wiki](https://wiki.ffo.jp/html/1344.html)

Captures: 

Penguin Ring + Albatross Ring
https://www.youtube.com/watch?v=6axIzVfu4eg
https://drive.google.com/open?id=1iTjp0T0-C5o72n6oBd5xbegM3h4ngRVu

Penguin Ring x2
https://youtu.be/HB8HbIblaUw
https://drive.google.com/open?id=1uiw8J9to-INLIy0Wj96LmilTwwrtSS8W

TeoTwawki:
[That modifier only does exactly what its comment says - and not at all what the wiki's say](https://github.com/search?q=repo%3ALandSandBoat%2Fserver%20penguin_ring_effect&type=code)

> Ring is also labeled on LSB as R/EX. With the captures, it's only EX.

[...]

PR #4816, opened 2023-12-05
[FISHING] Implement rings: Penguin, Pelican, Albatross, Noddy

MowFord:
**_I affirm:_**

Penguin, Pelican, and Albatross correspond to mods, respectively. These mods are on/off style and coded in `fishingutils.cpp`. All that's needed is for the rings to properly apply the mod via a buff that lasts for 20 minutes

Noddy ring is a bit more in depth but luckily there was already example code for fisherman's apron. Emulated that logic. Also I see no evidence that "decreasing the chance of items/mobs" should simply shift that chance to getting nothing. As these pools are weights and a random number is generated based on their total, I've removed the addition of the weight to `NoCatchWeight`

I've adjusted the `overwrite` value for `ENCHANTMENT` to be 4. This means that it doesn't overwrite other effects with the same "power". This is, as far as I can tell, the only way to have the same status effect multiple times on an entity.

Finally, a video showing it wears on unequip and on zone:

https://imgur.com/76zvqVm

Equip the 3 effect rings, use them to see that you can have multiple buffs. Unequip to see that only that particular buff wears off.

Noddy is hard to test effectively, but the code is very simple. I did test and confirm fishing still works.

### DIFF
--- a/src/map/utils/fishingutils.cpp
+++ b/src/map/utils/fishingutils.cpp
@@ -1061,8 +1061,8 @@ namespace fishingutils
         gear.waist           = (waist == FISHERS_ROPE) ? waist : 0;
         gear.legs            = (legs == FISHERMANS_HOSE || legs == ANGLERS_HOSE) ? legs : 0;
         gear.feet            = (feet == FISHERMANS_BOOTS || feet == ANGLERS_BOOTS) ? feet : 0;
-        gear.ring1           = 0;
-        gear.ring2           = 0;
+        gear.ring1           = (PChar->getEquip(SLOT_RING1) != nullptr) ? PChar->getEquip(SLOT_RING1)->getID() : 0;
+        gear.ring2           = (PChar->getEquip(SLOT_RING2) != nullptr) ? PChar->getEquip(SLOT_RING2)->getID() : 0;
         gear.ranged          = 0;
         gear.ammo            = 0;
 
@@ -2346,7 +2346,21 @@ namespace fishingutils
             if (sub > 0)
             {
                 ItemPoolWeight -= sub;
-                NoCatchWeight += sub;
+                // These are truly weights, so we don't need any particular totals
+                // NoCatchWeight += sub;
+            }
+        }
+
+        // Fishing skill (artisan and above): Reduces chances of fishing up monsters.
+        if (fishingSkill >= 68 && (gear.ring1 == NODDY_RING || gear.ring2 == NODDY_RING) && MobPoolWeight > 0)
+        {
+            uint16 sub = (uint16)std::floor(MobPoolWeight * 0.25f);
+
+            if (sub > 0)
+            {
+                MobPoolWeight -= sub;
+                // These are truly weights, so we don't need any particular totals
+                // NoCatchWeight += sub;
             }
         }
 
--- a/scripts/items/albatross_ring.lua
+++ b/scripts/items/albatross_ring.lua
@@ -1,0 +1,49 @@
+-----------------------------------
+-- ID: 15552 and 15555
+-- Albatross Ring
+--  Enchantment: Increases stamina while fishing
+-----------------------------------
+local itemObject = {}
+
+local itemMod = xi.mod.ALBATROSS_RING_EFFECT
+
+itemObject.onItemCheck = function(target)
+    if target:getMod(itemMod) > 0 then
+        return xi.msg.basic.ITEM_UNABLE_TO_USE_2
+    end
+
+    return 0
+end
+
+itemObject.onItemUse = function(target, player, item)
+    local effectType = xi.effect.ENCHANTMENT
+    local itemId = item:getID()
+    -- will not overwrite other enchantments due to effect and power being 0
+    target:addStatusEffectEx(effectType, effectType, 0, 0, 1200, 0, itemId)
+    for _, effect in pairs(target:getStatusEffects()) do
+        if
+            effect:getEffectType() == effectType and
+            effect:getSubPower() == itemId
+        then
+            effect:addEffectFlag(xi.effectFlag.ON_ZONE)
+            -- gives the player the mod, and stacks the mod to the effect so it is removed when the effect wears off
+            effect:addMod(itemMod, 1)
+            target:addMod(itemMod, 1)
+        end
+    end
+end
+
+itemObject.onItemUnequip = function(target, item)
+    local effectType = xi.effect.ENCHANTMENT
+    local itemId = item:getID()
+    for _, effect in pairs(target:getStatusEffects()) do
+        if
+            effect:getEffectType() == effectType and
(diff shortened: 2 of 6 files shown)


cocosolos:
Not sure if the functionality exists yet, but the ring enchantments should wear off if the ring is removed (also on logout/zone, but that should just be an effect flag that might already be set.)

MowFord:
> Not sure if the functionality exists yet, but the ring enchantments should wear off if the ring is removed (also on logout/zone, but that should just be an effect flag that might already be set.)

Oh interesting. Ok that makes things even more complicated... wear on zone is easily set as an effect flag, but I don't know how to make effects wear on removing gear

Xaver-DaRed:
itemObject.onItemEquip = function(target, item)
itemObject.onItemUnequip = function(target, item)

in this case:
```lua
itemObject.onItemUnequip = function(target, item)
    if target:hasStatusEffect(X)
        target:delStatusEffect(X)
    end
end
```

TeoTwawki:
for enchantment effect last I knew attaching to the item was done via a subid and stats added in the item script, like with food.

unless someone broke that (its been years since I did one, but breath mantle/high breath mantle does same thing - attach to item)

TeoTwawki:
https://github.com/LandSandBoat/server/blob/2377b2a3cee236fdad3357366762c9dd86a402ca/scripts/items/high_breath_mantle.lua#L22

you can have multiple unrelated enchantment effects from different items. this is why on retail you can stack the 2 different pelican rings if you have the old version from before SE replaced it.

cocosolos:
> itemObject.onItemEquip = function(target, item)
> itemObject.onItemUnequip = function(target, item)
> 
> in this case:
> ```lua
> itemObject.onItemUnequip = function(target, item)
>     if target:hasStatusEffect(X)
>         target:delStatusEffect(X)
>     end
> end
> ```

I actually wonder what happens if you're wearing 2 Albatross, Duck, or Pelican rings on retail. Does the effect wear only when you remove both rings, or specifically the ring that was used? Would be funny and convenient if it removed the effect when you remove the first ring.

TeoTwawki:
> I actually wonder what happens if you're wearing 2 Albatross, Duck, or Pelican rings on retail. Does the effect wear only when you remove both rings, or specifically the ring that was used?

I wanna say on retail its specific to the actual item you used, but I'm not positive. And I think our implementation will act on item ID regardless of if you have 2 of the same ID :(

edit: actually I forgot - you can't activate 2 of the same item unless they are different IDs. thats why the double I mentioend can work, SE replaced the ring with a new one of diff ID

Xaver-DaRed:
We could just use effect subpower to save the slot of the item used and THEN add a check for the ring slot placement before removing the effect

MowFord:
i thought i tried using enchantment and it didn't remove one removing the gear. Also `enchantment` status effect overwrites other enchantment status effects due to `overwrite == 0` on it

TeoTwawki:
> i thought i tried using enchantment and it didn't remove one removing the gear. Also `enchantment` status effect overwrites other enchantment status effects due to `overwrite == 0` on it

oh boy. we have things to fix.. the only enchants that shouldn't be tied to item are if the item ID is not passed it (its optional).

At least the effects flags can be actually changed in script! https://github.com/LandSandBoat/server/blob/base/src/map/lua/lua_statuseffect.h#L68

cocosolos:
> edit: actually I forgot - you can't activate 2 of the same item unless they are different IDs. thats why the double I mentioend can work, SE replaced the ring with a new one of diff ID

You might not be able to activate them but you can still equip 2 since those ones are only Ex.

> We could just use effect subpower to save the slot of the item used and THEN add a check for the ring slot placement before removing the effect

This sounds like it would work.

MowFord:
> At least the effects flags can be actually changed in script! https://github.com/LandSandBoat/server/blob/base/src/map/lua/lua_statuseffect.h#L68

afaik you can't set flags on an effect before applying it, and when you apply the overwrite check is done

TeoTwawki:
> afaik you can't set flags on an effect before applying it, and when you apply the overwrite check is done

That's true, but the second the effect object exists you can chain a `owningObject:getStatusEffect(ENUM):setStuffAndThings()` together to act on it - even immediately after its born. we do that in a few places. inside an effect script or an item script you don't even have to go that far. Remember all those `effect:getPower()` all over the repo? Same concept. And changes there will only apply to that specific instance of the effect.

MowFord:
Got an example somewhere of that? I thought the flag checking was done within the addedfect binding

TeoTwawki:
> Got an example somewhere of that? I thought the flag checking was done within the addedfect binding

we actually stripped flags+add death flag immediately after adding spike effects (so right after effect was born) to a mob to prevent it being removed, so you can for sure change flags and they will correctly be checked outside of the addEffect binding.

MowFord:
Also the issue being discussed is the `overwrite` value of an effect, not the flag specifically.

If `overwrite` is 4, then it lets two buffs with the same power coexist. I suspect we just need to make the `enchantment` buff have `overwrite == 4`, but that's a big change for a lowly dev like me

MowFord:
Also linking to a related issue

https://github.com/LandSandBoat/server/issues/4785

MowFord:
Allright I did the thing. Note that `ENCHANTMENT` being able to live alongside other of the same effect means this logic is now flawed:

image

I might have opened a can of worms on this one, but I think that in general `ENCHANTMENT` effects should be able to live alongside other `ENCHANTMENT` effects

src/map/utils/fishingutils.cpp
```diff
@@ -2346,7 +2346,21 @@ namespace fishingutils
             if (sub > 0)
             {
                 ItemPoolWeight -= sub;
-                NoCatchWeight += sub;
+                // These are truly weights, so we don't need any particular totals
+                // NoCatchWeight += sub;
```
Xaver-DaRed, line 2350: As you pointed out, you've found `no evidence that "decreasing the chance of items/mobs" should simply shift that chance to getting nothing` which means you also have not found evidence of the contrary either, so change this back, please.

src/map/utils/fishingutils.cpp
```diff
@@ -2346,7 +2346,21 @@ namespace fishingutils
...
+        // Fishing skill (artisan and above): Reduces chances of fishing up monsters.
+        if (fishingSkill >= 68 && (gear.ring1 == NODDY_RING || gear.ring2 == NODDY_RING) && MobPoolWeight > 0)
+        {
+            uint16 sub = (uint16)std::floor(MobPoolWeight * 0.25f);
+
+            if (sub > 0)
+            {
+                MobPoolWeight -= sub;
+                // These are truly weights, so we don't need any particular totals
+                // NoCatchWeight += sub;
```
Xaver-DaRed, line 2363: Do add the wight to `NoCatchWeight` please

PR #4828, opened 2023-12-08
[LUA] Quick Draw: Set recast on the action, not the ability

MowFord:
**_I affirm:_**

The person that coded `QUICK_DRAW_RECAST` (me) erroneously set the recast on the ability. Since this is a charge-based grouped ability, the recast is actually the charge cost. So `ability:getRecast()` was 1, subtract the 5s mod from ASA and you get 0 charge cost... this meant Quick Draw could be spammed nonstop.

This confusion was due to a few of the quick draw abilities not having an `action` param, and attempting to `ability:setRecast()` in `onUseAbility` does nothing... my mistake

I've tested and this is fixed now. Apologies I thought I had tested but I don't know COR very well and clearly I didn't test it properly

Give a weapon and bullets: `!additem martial_gun`, `!additem bronze_bullet 99`. Use a shot: `/ja "fire shot" <t>`, see the recast is 60s. give yourself the cooldown reduction: `!setmod QUICK_DRAW_RECAST 5`, shoot again and see lower cooldown

### DIFF
--- a/scripts/actions/abilities/dark_shot.lua
+++ b/scripts/actions/abilities/dark_shot.lua
@@ -19,14 +19,14 @@ abilityObject.onAbilityCheck = function(player, target, ability)
         player:hasItem(xi.item.DARK_CARD, 0) or
         player:hasItem(xi.item.TRUMP_CARD, 0)
     then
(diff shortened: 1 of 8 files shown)


TeoTwawki (approved):
_"oops"_ 🤣

Xaver-DaRed (approved):

PR #4830, opened 2023-12-08
Fixes mob evasion using wrong formula

Frankie-hz:
**_I affirm:_**

When implementing the mob stat re-work there a mix-up in how we calculate player evasion vs mob evasion.

See the explanation below by Jimmayus.

Player evasion: https://www.bg-wiki.com/ffxi/Evasion
Evasion = floor( AGI ÷ 2 ) + ( Evasion from Skill ) + ( Evasion from Traits/JAs/gear )

with a breakpoint at 200 skill resulting in a bifurcated formula:

Accuracy from Skill =
For Skill ≤ 200: Evasion = Skill
For Skill ≥ 201: Evasion = floor( (Evasion Skill-200) × 0.9 ) + 200

Mob Evasion: https://w.atwiki.jp/studiogobli/pages/25.html
敵防御 ＝ [f(Lv,種族の防御ランク) + 8 + [VIT/2] + ジョブ特性]×種族特性
敵回避 ＝ f(Lv,メインジョブの回避スキルランク) + [AGI/2] + ジョブ特性

Monster Defense: [f(Family Defense rank) + 8 + VIT/2] + job traits] * bonuses Monster Evasion f(main job skill rank) + AGI/2 + job traits + bonuses where "f" is defined by the following table:
Rank    Lv-1～50    Lv51～
Ａ    6+[(Lv-1)*3.0]    153+[(Lv-50)*5.0]?
Ｂ    5+[(Lv-1)*2.9]    147+[(Lv-50)*4.9]?
Ｃ    5+[(Lv-1)*2.8]    142+[(Lv-50)*4.8]
Ｄ    4+[(Lv-1)*2.7]    136+[(Lv-50)*4.7]?
Ｅ    4+[(Lv-1)*2.5]    126+[(Lv-50)*4.5]

the "?" is an ongoing problem because sometimes monsters express half ranks, but that's a can of worms and there's only so many researchers Anyway, in our current analysis the best way to explain why this is a big deal is to use Greater Colibri a level 82 Greater Colibri has 339 evasion and 67 agility using the player evasion skill analysis:

Colibri are RDM|RDM, RDMs have rank D evasion skill:
level 82 rank D skill is 240, 200 + (240-200)*.9 = 236 base vs
level 82 rank D skill monster is 136+[(Lv-50) * 4.7], 136 + (82 - 50) * 4.7, 136 + 150 (values are generally floored in XI) = 286
236 vs 286, or a base evasion drop of 50

1) 236 vs 286 evasion base, vs the known 339 evasion value
2) the only other 3 sources of evasion are job traits (none for rdm), agi & bonuses
3) 67 known agility is only 33 evasion additional, resulting in 269 and 319 evasion pre-bonuses

I'll spare use some additional math and link only to the excellent work already done on this subject: https://ffxiclopedia.fandom.com/wiki/Category_talk:Colibris

Before: 
image

After:
image

I had to reproduce a level 82 Colibri so for this test with a known evasion value we had to do the following:
1) Change `Locus_Colibri` minLevel and maxLevel to be 82 to mimic the known values of these birds. `UPDATE mob_groups SET minLevel = 82, maxLevel = 82 WHERE name = 'Locus_Colibri'; `
2) !gotoid 16990444
3) !getstats

### DIFF
--- a/src/map/utils/mobutils.cpp
+++ b/src/map/utils/mobutils.cpp
@@ -82,7 +82,7 @@ namespace mobutils
     }
 
     // Gest base skill rankings for ACC/ATT/EVA/MEVA
-    uint16 GetBase(CMobEntity* PMob, uint8 rank)
+    uint16 GetBaseSkill(CMobEntity* PMob, uint8 rank)
     {
         int8 mlvl = PMob->GetMLevel();
 
@@ -100,25 +100,28 @@ namespace mobutils
                 return battleutils::GetMaxSkill(SKILL_THROWING, JOB_MNK, mlvl); // E Skill (5)
         }
 
-        ShowError("mobutils::GetBase rank (%d) is out of bounds for mob (%u) ", rank, PMob->id);
+        ShowError("mobutils::GetBaseSkill rank (%d) is out of bounds for mob (%u) ", rank, PMob->id);
         return 0;
     }
 
     uint16 GetMagicEvasion(CMobEntity* PMob)
     {
         uint8 mEvaRank = PMob->evaRank;
-        return GetBase(PMob, mEvaRank);
+        return GetBaseSkill(PMob, mEvaRank);
     }
 
     /************************************************************************
      *                                                                       *
-     *  Base value for defense                       *
+     *  Base value for defense and evasion                                   *
      *                                                                       *
      ************************************************************************/
 
-    uint16 GetDefense(CMobEntity* PMob, uint8 rank)
+    uint16 GetBaseDefEva(CMobEntity* PMob, uint8 rank)
     {
-        // family defense = [floor(defRank) + 8 + vit / 2 + job traits] * family multiplier
+        // See: https://w.atwiki.jp/studiogobli/pages/25.html
+        // Enemy defense = [f(Lv, racial defense rank) + 8 + [VIT/2] + job characteristics] x racial characteristics
+        // Enemy evasion = f(Lv, main job evasion skill rank) + [AGI/2] + job characteristics
+        // The funcion f is below
         uint8 lvl = PMob->GetMLevel();
 
         if (lvl > 50)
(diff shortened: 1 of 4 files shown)


WinterSolstice8 (approved):

src/map/utils/mobutils.cpp
```diff
@@ -100,25 +100,28 @@ namespace mobutils
...
 
     /************************************************************************
      *                                                                       *
-     *  Base value for defense                       *
+     *  Base value for defense and evasion                                   *
      *                                                                       *
      ************************************************************************/
 
-    uint16 GetDefense(CMobEntity* PMob, uint8 rank)
+    uint16 GetBase(CMobEntity* PMob, uint8 rank)
```
Xaver-DaRed: This is gonna get confusing.

I think we could name them `GetBaseSkill` and `GetBaseDefEva` or something along those lines, so we know what we are doing.

I do like the changes and the detailed explanation though. But future ourselves will come back to this eventually and I dont want "future me" hating "current me".

Frankie-hz: I was struggling to find a good naming convention...thank you

src/map/utils/mobutils.h
```diff
@@ -66,8 +66,8 @@ namespace mobutils
 
     uint16 GetWeaponDamage(CMobEntity* PMob, uint16 slot);
     uint16 GetMagicEvasion(CMobEntity* PMob);
-    uint16 GetEvasion(CMobEntity* PMob);
```
Xaver-DaRed, line 69: How is this possible? named in header as `GetEvasion` and in .cpp as `GetDefense`?

Oh well

Xaver-DaRed (approved):

src/map/utils/mobutils.h
Frankie-hz: I was just talking to teo about this...I have no idea how that happened. GetEvasion was deleted so long ago.

Xaver-DaRed: This proves all this really works via black magic.

Not gonna be me the one to complain, though XD

Frankie-hz: I don't want to talk about magic at 3am...

MowFord:
This is great, looks good to me (just compiled/tested locally)

Pretty funny seeing this being used all over in the code, though:

image

PR #4835, opened 2023-12-09
Adjust logic of lantern trades in den of rancor

TeoTwawki:
**_I affirm:_**

This was likely to go unnoticed given the error is not easily observable from the client side.

`npcUtil.tradeHas()` does not handle full inventory or messages about the item you cannot get (it can't - how would it know?) and can't be used here.

Fortunately, `npcUtil.giveItem()` does handle that properly.

Also adjusted some login in the zone's global to avoid locking lanterns on failed trades to already lit NPCs

- !zone den**tabkey** to den of rancor.
- `!speed 200` and zoom yourself to the upper level lamp lighting area
- `!additem unlit_lantern`
- trade and see that it functions as expected to get rancor flame
- zoom down to the gates and trade a rancor flame to see it lights
- `!additem rancor_flame` to get another, and trade to the already lit one, see it tell you its already lit, then light the other 3 while continuing to add more as needed with the gm command till you light all 4 and the gate opens.

If you want to see what the hidden error is, you'll need to debug on the server side without these changes to check the items state before and after each trade attempt while also filling your inventory and pre-lighting the lamps at the gate. As stated previously, you'd normally never know anything was wrong at the client side.

### DIFF
--- a/scripts/zones/Den_of_Rancor/globals.lua
+++ b/scripts/zones/Den_of_Rancor/globals.lua
@@ -10,10 +10,13 @@ local denOfRancorGlobal =
         trade to lanterns next to Sacrificial Chamber (Rancor Flame)
         ..............................................................................................]]
     onTradeLanternChamber = function(player, npc, trade)
-        if npcUtil.tradeHas(trade, xi.item.RANCOR_FLAME) then -- Rancor Flame
-            if npc:getAnimation() == xi.anim.OPEN_DOOR then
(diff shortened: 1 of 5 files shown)


TeoTwawki:
This is mainly so people don't copy paste from these scripts into new ones and find places this hidden error would actually matter at.

Xaver-DaRed (approved):

PR #4836, opened 2023-12-09
[docs] Add "Thanks" section to README

zach2good:
**_I affirm:_**

We're not required to thank GH or JB for all the freebies they give us, but it's the nice thing to do, given that these packs have a value of 200GBP/year each. I'd hate to think how much the hosting and CI minutes would cost us if we were paying for them. 💀 

N/A


claywar (approved):

zach2good:
image

PR #4842, opened 2023-12-09
[core, Lua] Rename rename PrintToPlayer and PrintToArea to lowercase

zach2good:
**_I affirm:_**

Been meaning to do this since the Sol refactor. This brings these two bindings in line with every other binding we have. They're only used for prints in GM commands and a couple of left over debug prints for broken content.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -170,8 +170,9 @@ log/leaks.txt
 .project
 .vs
 /win32/.vs/xi/v15
+.luarc.json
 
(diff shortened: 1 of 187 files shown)


claywar (approved):

TeoTwawki (approved):
muscle memory forgive me... _stamps_

Xaver-DaRed (approved):

PR #4850, opened 2023-12-11
[AMK, Lua, SQL] Mission 9, stone hunt in quicksands caves complete

Flibe-XI:
**_I affirm:_**

Implementation of Rescue! A Moogles Labor of Love, mission 9 of AMK.  The player needs to collect 9 'stone' key items by clicking various question marks in quicksands caves.  There are 20 possible QMs, and 10 possible sets of combinations required to check.  The set which is assigned randomly in this commit is then passed as an arg to the geologist event which marks the map.

`!addmission 10 8`
`!pos -737 -6 -550 208`
Talk to goblin geologist, he will assign a marker set to you.  Go around quicksands and click the markers shown on the map for your stone KIs.  (alternatively, `!gotoid ###`, ids are in the mission lua file)
Return to goblin to get KI Navaratna Talisman. 
Get CS at shimmering circle in chamber of oracles.

### DIFF
--- a/scripts/missions/amk/09_Rescue_A_Moogles_Labor_of_Love.lua
+++ b/scripts/missions/amk/09_Rescue_A_Moogles_Labor_of_Love.lua
@@ -2,9 +2,22 @@
 -- Rescue! A Moogle's Labor of Love
 -- A Moogle Kupo d'Etat M9
 -- !addmission 10 8
+-- Geologist cutscene args : csid, progress, has QC map: 1 or 0, markerset: 1-10
+-- Goblin Geologist        : !pos -737 -6 -550 208
+-- STONE_OF_SURYA          : !addkeyitem 1145
+-- STONE_OF_CHANDRA        : !addkeyitem 1146
+-- STONE_OF_MANGALA        : !addkeyitem 1147
+-- STONE_OF_BUDHA          : !addkeyitem 1148
+-- STONE_OF_BRIHASPATI     : !addkeyitem 1149
+-- STONE_OF_SHUKRA         : !addkeyitem 1150
+-- STONE_OF_SHANI          : !addkeyitem 1151
+-- STONE_OF_RAHU           : !addkeyitem 1152
+-- STONE_OF_KETU           : !addkeyitem 1153
+-- NAVARATNA_TALISMAN      : !addkeyitem 1158
 -----------------------------------
 require('scripts/globals/missions')
 require('scripts/globals/interaction/mission')
+local ID = zones[xi.zone.QUICKSAND_CAVES]
 -----------------------------------
 
 local mission = Mission:new(xi.mission.log_id.AMK, xi.mission.id.amk.RESCUE_A_MOOGLES_LABOR_OF_LOVE)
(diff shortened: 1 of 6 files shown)


scripts/missions/amk/09_Rescue_A_Moogles_Labor_of_Love.lua
```diff
@@ -14,83 +27,202 @@ mission.reward =
...
+    [18] = 17629762,
+    [19] = 17629763,
+    [20] = 17629764,
+    [21] = 17629765,
+    [22] = 17629766,
+    [23] = 17629767,
+    [24] = 17629768,
+    [26] = 17629770,
+    [27] = 17629771,
+}
```
Xaver-DaRed: Using raw IDs for NPCs or mobs in not "zone" files is not a desirable method.
The main reason is because this IDs can and will shift and theres no way for us to remember this IDs have to be updated.

I would recommend defining this npcs in their zone ID file and call their enums here

scripts/zones/Quicksand_Caves/IDs.lua
```diff
@@ -117,6 +120,26 @@ zones[xi.zone.QUICKSAND_CAVES] =
         CHAINS_THAT_BIND_US_QM = 17629746,
         TREASURE_COFFER        = 17629747,
         ANTICAN_TAG_QM         = 17629769,
+        QM_AMK_1               = 17629668,
```
zach2good: These are still raw IDs, you need to be using `GetFirstID` or `GetTableOfIDs` to pull these from the db at load time - which makes them more shift-proof

scripts/zones/Quicksand_Caves/IDs.lua
```diff
@@ -40,6 +41,7 @@ zones[xi.zone.QUICKSAND_CAVES] =
         HOMEPOINT_SET                 = 11435, -- Home point set!
         UNITY_WANTED_BATTLE_INTERACT  = 11493, -- Those who have accepted % must pay # Unity accolades to participate. The content for this Wanted battle is #. [Ready to begin?/You do not have the appropriate object set, so your rewards will be limited.]
     },
+
```
zach2good: No need for these additional linebreaks, none of the other ID files have them

scripts/zones/Quicksand_Caves/DefaultActions.lua
```diff
@@ -1,6 +1,7 @@
 local ID = zones[xi.zone.QUICKSAND_CAVES]
 
 return {
+    ['qmAMK']             = { messageSpecial = ID.text.NOTHING_OUT_OF_ORDINARY },
```
zach2good: Looking at all the other qm's we have in the codebase, they're all in `snake_case`, can you make this match? `qm_amk`, etc.

scripts/missions/amk/09_Rescue_A_Moogles_Labor_of_Love.lua
```diff
@@ -14,83 +27,202 @@ mission.reward =
...
+    [20] = ID.npc.QM_AMK_20,
+    [21] = ID.npc.QM_AMK_21,
+    [22] = ID.npc.QM_AMK_22,
+    [23] = ID.npc.QM_AMK_23,
+    [24] = ID.npc.QM_AMK_24,
+    [26] = ID.npc.QM_AMK_26,
+    [27] = ID.npc.QM_AMK_27,
+}
+
+local markerSets =
```
zach2good, line 30: 👍

sql/npc_list.sql
```diff
@@ -23470,26 +23470,26 @@ INSERT INTO `npc_list` VALUES (17629748,'qm7','???',115,-506.764,17.375,-420.372
...
-INSERT INTO `npc_list` VALUES (17629763,'qm19','???',0,-142.000,1.252,-271.000,1,40,40,0,0,32,0,3,0x0000340000000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17629764,'qm20','???',0,-65.000,1.413,-270.000,1,40,40,0,0,32,0,3,0x0000340000000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17629765,'qm21','???',0,631.000,0.187,-419.000,1,40,40,0,0,32,0,3,0x0000340000000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17629766,'qm22','???',0,862.000,-8.500,-415.000,1,40,40,0,0,32,0,3,0x0000340000000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17629767,'qm23','???',0,613.000,8.981,-537.000,1,40,40,0,0,32,0,3,0x0000340000000000000000000000000000000000,0,NULL,0);
-INSERT INTO `npc_list` VALUES (17629768,'qm24','???',0,770.000,0.460,-668.000,1,40,40,0,0,32,0,3,0x0000340000000000000000000000000000000000,0,NULL,0);
+INSERT INTO `npc_list` VALUES (17629752,'qmAMK','qm8',0,-345.000,0.500,826.000,1,40,40,0,0,32,0,3,0x0000340000000000000000000000000000000000,0,NULL,0);
```
zach2good, line 23473: I forget what if the polutils name is used for anything, but typically this isn't a thing we change. The first `name` field is the one we use to do internal lookups and fire off scripts etc. The second one is pulled from POLUtils (I assume, from the name?) and may or may not be what's send to the client. On other entity types it is anyway.

scripts/zones/Quicksand_Caves/npcs/amk_qm.lua
```diff
@@ -0,0 +1,12 @@
+-----------------------------------
+-- Area: Quicksand Caves
+-- NPC: qm (???)
```
zach2good: Name doesn't match the filename

scripts/zones/Quicksand_Caves/npcs/amk_qm.lua
```diff
@@ -0,0 +1,12 @@
+-----------------------------------
```
zach2good: This is the correct naming for file here, can you update npc_list and the mission script to key on this?

scripts/zones/Quicksand_Caves/npcs/amk_qm.lua
```diff
@@ -0,0 +1,12 @@
...
+-- Area: Quicksand Caves
+-- NPC: qm (???)
+-- Mission: AMK - Rescue! A Moogle's Labor of Love
```
zach2good: Probably worth adding a comment saying this script is shared between all the AMK qm's for this mission, to avoid confusion or someone bloating out the zone with multiple empty files - one for each qm

zach2good (changes requested):
Getting closer, thanks for all the hard work 🖖 (there's no regular salute emoji?)

zach2good:
I skipped over implementing this mission on my branch way-back-when, because it looked like a pain in the ass, thanks for getting to it 👍

scripts/zones/Quicksand_Caves/IDs.lua
Flibe-XI: Oh, I had found one that did, so I thought that was the default.

sql/npc_list.sql
Flibe-XI: Ya, I was trying to preserve the identity of the qm in the db somehow, as the first name field is all going to be identical.  Not sure of another way to do it.

Flibe-XI:
> I skipped over implementing this mission on my branch way-back-when, because it looked like a pain in the ass, thanks for getting to it 👍

That seems to be the theme SE was going for with AMK missions.  This one was just tedious but the puzzles were even worse.

scripts/zones/Quicksand_Caves/IDs.lua
Flibe-XI: See Qufim Island, the only one near it like that haha. Bad luck.

scripts/zones/Quicksand_Caves/IDs.lua
Flibe-XI: That is really cool.  Had to rejig the markerSets table in the mission lua.  Did it via find/replace and spot checked it. Works great!

scripts/zones/Quicksand_Caves/IDs.lua
```diff
@@ -117,6 +118,7 @@ zones[xi.zone.QUICKSAND_CAVES] =
         CHAINS_THAT_BIND_US_QM = 17629746,
         TREASURE_COFFER        = 17629747,
         ANTICAN_TAG_QM         = 17629769,
+        QM_AMK                 = GetTableOfIDs('qm_amk')
```
zach2good, line 121: 👍

zach2good (approved):

PR #4860, opened 2023-12-14
Misc fixes

WinterSolstice8:
**_I affirm:_**

Fixes crash with transports if they fail to load and try to resize a string on a nullptr npc
Fix corner case with EquipItem

Break Transports with bad NPC ID, boot up the server and don't see it crash
Equip sub items and see them work properly.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -2613,6 +2613,13 @@ namespace charutils
                 PChar->pushPacket(new CMessageBasicPacket(PChar, PChar, 0, 0, 0x200));
                 return;
             }
+
+            // Disallow everything but shields if you're using H2H
+            // Equipping a shield will unequip the H2H weapon and you will go barefisted with a shield
+            if (PMainItem && PMainItem->getSkillType() == SKILL_HAND_TO_HAND)
+            {
+                return;
+            }
         }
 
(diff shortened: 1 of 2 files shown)


src/map/utils/charutils.cpp
```diff
@@ -2613,6 +2613,12 @@ namespace charutils
                 PChar->pushPacket(new CMessageBasicPacket(PChar, PChar, 0, 0, 0x200));
                 return;
             }
+
+            // Can't equip anything in sub slot if you're using H2H other than shields
```
TeoTwawki: Eh? Comment confuses, because actual h2h can't use shield either. Only the no-skilled 1 hand punch can.

WinterSolstice8: when you equip the shield it unequips your h2h so you're bare-fisted with one punch

TeoTwawki: the way it was worded is a lil wonky if that's what you meant to say, makes it sound like saying "if you are h2h, only a shield can be here". Maybe `// equipping a shield will remove h2h item, else sub slot must remain empty`

WinterSolstice8: updated comment

TeoTwawki (approved):

PR #4867, opened 2023-12-15
[core] Remove exceptions from AI states

### LINKED ISSUE #4024: 🔨 Exceptions should be exceptional
**_I affirm:_**

## Describe the feature

Mainly in the AI state system we use throwing exceptions as a way to bail out of the current state and bubble up an error message or some state to the original AI controller. This works, but exception handing is slow and should only be used _in exceptional circumstances_. In our case, being out of range if you enter range_state is not an exception worthy event, we should be setting some sort of error condition and then popping it off the stack. 

Reference:

https://github.com/LandSandBoat/server/blob/base/src/map/ai/states/range_state.cpp#L39-L53

https://pspdfkit.com/blog/2020/performance-overhead-of-exceptions-in-cpp/

zach2good:
**_I affirm:_**

Prepare for a lesson in Core + C++.

- Inside ChangeState, the template argument tells the function what state it needs to construct and then put onto the state stack.
- While it's doing this, it's listening (via a try/catch block) for exceptions to be bubbled up.
- This custom exception we have carries a packet that will be emitted to the entity - if they can receive it (players).
- We have done it this way because the only way to bail out of a class constructor early is to set flags or throw an exception.
- We're also unable to ferry values out of a constructor, unless we throw a custom exception and attach the data to it, and then catch it in a try/catch block.

===

- What I'm going to do instead is make the construction of a state _very simple_, it will take the arguments passed in (maybe I should capture these as a tuple?), and store them. Nothing else.
- Next up we will call state->Initialize(), which will return `StateResult`.
- This is a typedef of `nonstd::expected<void, std::unique_ptr<CBasicPacket>>`. In the happy case it holds nothing (void), and in the error case it holds our error packet.
- All the other logic is replacing `throw` with `return nonstd::make_unexpected(...)`

Everything works as normal.

Closes https://github.com/LandSandBoat/server/issues/4024

### DIFF
--- a/src/map/ai/CMakeLists.txt
+++ b/src/map/ai/CMakeLists.txt
@@ -8,5 +8,7 @@ set(AI_SOURCES
     ${AI_STATE_SOURCES}
     ${CMAKE_CURRENT_SOURCE_DIR}/ai_container.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/ai_container.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/state.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/state.h
     PARENT_SCOPE
 )
--- a/src/map/ai/ai_container.h
+++ b/src/map/ai/ai_container.h
@@ -32,7 +32,7 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "helpers/pathfind.h"
 #include "helpers/targetfind.h"
 #include "packets/message_basic.h"
-#include "states/state.h"
+#include "state.h"
 
 class CBaseEntity;
 class CCharEntity;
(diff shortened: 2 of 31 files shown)


zach2good:
Meh, I don't even need `nonstd::expected`, I can just set the error message and return early, and then check to see if the error message has been set - then extract it.

zach2good:
Maybe I do need expected, because I can either return success, failure with no packet, or failure with a packet. Which would be:
expected<void, std::optional<CBasicPacket>>

- void for success
- nullopt for silent fail
- the packet for sending a failure packet

zach2good:
Probably want to not do the Init() calls, and just set the error on the way out of the constructor, and check for the error immediately afterwards

zach2good:
Then state can have CState::StateResult as a member, and that's what we check

zach2good:
This will end up being implemented in https://github.com/LandSandBoat/server/pull/5185

PR #4868, opened 2023-12-15
Dark Ixion: Fight Mechanics

MowFord:
**_I affirm:_**

Recreated my PR as I was doing cleanup and didn't realize my older LSB PRs were linked to the Wings-XI repo

Please see previous PR for any lingering info: https://github.com/LandSandBoat/server/pull/4617

Makes the pony fight

Claim mechanic confirmed to work via new range exit listener

Once the fight starts, mobskills are all used via logic in onMobFight and TP is preserved via localvar each time his "auto attack" swings happen (since we use `mob:setMobSkillAttack(39)`)

DI's animation subs define mobskill behavior:
- 0 is normal
- 1 is Charging (trample)
- 2 is broken horn, cannot trample
- 3 is glowing and uses double offensive mobskills
  - also goes into this briefly when repairing horn, goes back to 0 afterwards

everytime DI autoattacks there's a chance (which grows as his HP lowers) to stack a localvar "run" counter. If this is nonzero and trample is available, then he starts the trample sequence.
- finds a random player within 30 yalms to trample. if found, he looks in that direction and starts trampling
- immediately places previous target (current battle target) in the `hitlist` and does damage if that player is in front of DI
- every onMobFight tick nearby players are placed in the hitlist and trampled if in front
- each trample path can only hit a player once (though if "run" is > 1 he may trample then immediately trample again and a player will get hit twice)
- Note that he can choose and hit _any_ player (doesn't have to be on the hate table or even in the alliance with claim)

All of his damage mobskills have flag 8 meaning they can hit any player/pet in range

Core fixes needed during testing:
- targetfind wasn't respecting findflags when an ability was conal
- rear conals weren't properly flagged as conal, so they were defaulting to single target. The `findWithinCone` function also didn't support a rear conal angle shift.
  - The only other mobskill with aoe type > 4 is rear lasers so i assume this also fixes it to work like acheron kick :)
- new lua util function `entity:getPlayersInRange()` to get all players within X range of entity. pass zero or nothing to get all players in zone.
- knockback was still applied even for miss/shadow absorb (This fix doesn't work 100% when there's multiple targets, but it _is_ better than it was)
  - I believe the issue is related to checking `PSkill->hasMissMsg()` in the loop of each target. Might be fundamentally flawed for aoe abilities, but i'll leave that for the reviewers to decide a course of action

Finally, there's a few things not perfectly accurate:
- his mobskills should not be able to be ranged to cancel the animation. 
  - I.E. if two players are standing in front of him and he uses acheron kick twice, the second one should go off even though the player got pushed out of range
  - I fixed this in WingsXI with a very hacky change to `targetfind.cpp` (allowed aoe mobskills to target the mob itself) that I'd rather not repeat unless 
    - it's generalized to not specifically only work for DI 
    - and fixes the animations to not show on the mob itself (really silly to see that)
  - It's fixed in this PR by making the db range for his aoe skills 35 and setting the message to NONE if the target is out of range
    - this is a bit of trickery I've done to make the fight work _mechanically_ closer to retail, by not letting a player cause the mob to cancel his ability if the primary target gets too far
- His mobskills shouldn't show an animation on the target if they aren't in range 
  - This may be fixable by setting the animation id in the same section that we adjust knockback, but I imagine that is not the best way to accomplish any of this. I am striving for fight difficulty/accuracy first and foremost, the animation without a combat message is minor imo.
- He should take reduced damage from the front (or increased from the back would work too). I attempted a few methods of using takedmg listeners but couldn't come up with something clean

Overall the fight feels very similar to WingsXI, which has had DI roaming around for close to a year now.

Previous PR on DI spawn/spook mechanics outlines how to get him to spawn, but the short version is
```
!zone 81
!exec SetServerVariable('DarkIxion_ZoneID', 81)
!spawnmob 17109367
!gotoid 17109367
<target DI and quickly run !stun>
<repeat these 3 until name turns red (you must land the hit)>
   !additem 19210
   <equip the ash>
   /ra <t>
<test fight>
```

The fight is fairly straightforward I think, there are multiple failsafes to avoid DI getting stuck in any particular phase/form

### DIFF
--- a/scripts/globals/dark_ixion.lua
+++ b/scripts/globals/dark_ixion.lua
@@ -4,7 +4,7 @@
 -- https://ffxiclopedia.fandom.com/wiki/Dark_Ixion
 --[[
     Find DI by:
-    - checking the server's zone var: "!checkvar server DarkIxion_ZoneID"
+    - checking the server's zone var: '!checkvar server DarkIxion_ZoneID'
     - go to that zone: !zone <zoneID>
     - go to DI: !gotoname Dark_Ixion
 
@@ -267,6 +267,26 @@ xi.darkixion.setupEntity = function(entity)
     entity.onMobDisengage = function(mob)
         xi.darkixion.onMobDisengage(mob)
     end
+
+    entity.onCriticalHit = function(mob, attacker)
+        xi.darkixion.onCriticalHit(mob, attacker)
+    end
+
+    entity.onWeaponskillHit = function(mob, attacker, weaponskill)
+        xi.darkixion.onWeaponskillHit(mob, attacker, weaponskill)
+    end
+
+    entity.onMobWeaponSkillPrepare = function(mob, target)
+        xi.darkixion.onMobWeaponSkillPrepare(mob, target)
+    end
+
+    entity.onMobWeaponSkill = function(mob, target, skill)
+        xi.darkixion.onMobWeaponSkill(mob, target, skill)
+    end
+
+    entity.onMobFight = function(mob, target)
+        xi.darkixion.onMobFight(mob, target)
+    end
 end
 
 xi.darkixion.repop = function(mob)
@@ -309,8 +329,8 @@ xi.darkixion.roamingMods = function(mob)
     end
 
     mob:setMobSkillAttack(39)
-    mob:setLocalVar('charging', 0)
-    mob:setLocalVar('double', 0)
+    mob:setLocalVar('trampling', 0)
+    mob:setLocalVar('doubleUpWS', 0)
     mob:setLocalVar('lastHit', 0)
     mob:setBehaviour(0)
     mob:setAutoAttackEnabled(true)
@@ -370,11 +390,16 @@ xi.darkixion.zoneOnGameHour = function(zone)
     end
 end
 
-xi.darkixion.onMobDeath = function(mob, player, isKiller)
-    player:addTitle(xi.title.IXION_HORNBREAKER)
-    -- only reset hp after being killed
-    SetServerVariable('DarkIxion_HP', 0)
-    SetServerVariable('DarkIxion_HornStatus', 0)
+xi.darkixion.onMobDeath = function(mob, player, optParams)
+    if optParams.isKiller or optParams.noKiller then
+        -- only reset hp after being killed
+        SetServerVariable('DarkIxion_HP', 0)
+        SetServerVariable('DarkIxion_HornStatus', 0)
+    end
+
+    if player then
+        player:addTitle(xi.title.IXION_HORNBREAKER)
+    end
 end
 
 xi.darkixion.onMobDespawn = function(mob)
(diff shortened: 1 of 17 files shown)


src/map/ai/states/range_state.cpp
```diff
@@ -135,6 +135,8 @@ bool CRangeState::Update(time_point tick)
             m_PEntity->OnRangedAttack(*this, action);
             m_PEntity->loc.zone->PushPacket(m_PEntity, CHAR_INRANGE_SELF, new CActionPacket(action));
         }
+
+        m_PEntity->PAI->EventHandler.triggerListener("RANGE_STATE_EXIT", CLuaBaseEntity(m_PEntity), CLuaBaseEntity(PTarget), CLuaAction(&action));
```
cocosolos: Can you document this in documentation/AI_Events.txt?

MowFord: Sure thing, done!

scripts/items/pinch_of_stygian_ash.lua
```diff
@@ -0,0 +1,26 @@
+-----------------------------------
+-- ID: 19210
+-- Stygian Ash
+-----------------------------------
+local itemObject = {}
+
+itemObject.onItemEquip = function(target, item)
+    target:addListener('RANGE_STATE_EXIT', 'STYGIAN_ASH_THROWN', function(playerArg, targetArg, action)
```
zach2good, line 8: This being in the item looks a bit stinky, wouldn't it make more sense to put range_attack_give/ranged_attack_take triggers in `CCharEntity::OnRangedAttack`; like we have with weaponskills?

```cpp
            this->PAI->EventHandler.triggerListener("WEAPONSKILL_USE", CLuaBaseEntity(this), CLuaBaseEntity(PTargetFound), PSkill->getID(), state.GetSpentTP(), CLuaAction(&action), damage);
            PTarget->PAI->EventHandler.triggerListener("WEAPONSKILL_TAKE", CLuaBaseEntity(PTargetFound), CLuaBaseEntity(this), PSkill->getID(), state.GetSpentTP(), CLuaAction(&action));
```

zach2good: I wouldn't remove the new trigger point though, more triggers isn't a bad thing 👍

zach2good: (if there are no attached listeners, the call is next to free)

zach2good:
```cpp
    // calls event from core
    template <class... Args>
    void triggerListener(std::string const& eventname, Args&&... args)
    {
        TracyZoneScoped;
        TracyZoneString(eventname);
        if (eventListeners.count(eventname))
        {
            for (auto&& event : eventListeners.at(eventname))
            {
                auto result = event.lua_func(std::forward<Args&&>(args)...);
                if (!result.valid())
                {
                    sol::error err = result;
                    ShowError("Error in listener event %s: %s", eventname, err.what());
                }
            }
        }
    }

```

MowFord: Interesting so you're saying we should create a ranged_take listener to go alon with the range_state_end listener?

Then, add the ranged take listener to DI? I suppose we could, though we probably wouldn't want it on DI all the time since it's only relevant when he's not engaged? Not impossible to add it on spawn, remove it on engage, and add it on disengage.

Definitely seems simpler in my mind to have it on the item. Also we would need to pass the itemid if the listener on DI would be deciding if he gets claimed or not. (Though I do agree having the listener being removed via a timer because the ammo gets unequipped before the range state ends is a bit wonky)

zach2good: The exit of an entity state is not always perfectly in line with the give/take of what is happening with the state (ranged attack, weaponskill, item use, etc.). I don't think it _hurts_ to keep range_state_end trigger, even if we don't use it. I think it makes more sense for the flow of:
- I (Ixion) install a listener for when I'm hit with a ranged attack. If that ranged attack is from a player with Stygian Ash equipped, do X.

Instead of:
- When a player equips Stygian Ash, install a listener on them so that if they hit something with it it'll have a localvar set. The target can check for this localvar at their convenience. They will do X on their next tick if they have the localvar. 

That seems a lot more steps and a lot less straightforward

src/map/lua/lua_baseentity.cpp
```diff
@@ -10036,6 +10036,30 @@ void CLuaBaseEntity::recalculateAbilitiesTable()
...
+
+    // clang-format off
+    zoneutils::GetZone(m_PBaseEntity->getZone())->ForEachChar([&](CCharEntity* PChar)
+    {
+        if (!dist || distance(PChar->loc.p, m_PBaseEntity->loc.p) < dist)
+        {
+            players.add(CLuaBaseEntity(PChar));
+        }
+    });
+    // clang-format on
```
zach2good, line 10091: This will return ALL players within range, including dead ones. You do check for this in the instances I can see, but we have a whole `TargetFind` system that you should be using for this.

If/when we improve TargetFind, this block will likely be missed from that by accident. All target-finding should exist in TargetFind. It's much better to use TargetFind, even if you have to refactor it in another PR before its safe/easy/possible to use here.

scripts/globals/dark_ixion.lua
```diff
@@ -462,3 +584,292 @@ xi.darkixion.onMobDisengage = function(mob)
...
+        end
+    end)
+
+    mob:setMobSkillAttack(39)
+
+    xi.darkixion.pos     = nil
+    xi.darkixion.hitList = nil
+end
+
+xi.darkixion.itsStompinTime = function(mob, target)
```
zach2good, line 804: Can we get a better name for this please?

scripts/globals/dark_ixion.lua
```diff
@@ -462,3 +584,292 @@ xi.darkixion.onMobDisengage = function(mob)
...
+    -- remove bind
+    mob:delStatusEffect(xi.effect.BIND)
+    mob:delStatusEffect(xi.effect.WEIGHT)
+    mob:setSpeed(95)
+    mob:setAutoAttackEnabled(false)
+    mob:setMobAbilityEnabled(false)
+    mob:setMobSkillAttack(0)
+    mob:setBehaviour(xi.behavior.NO_TURN + xi.behavior.STANDBACK)
+
+    local nearbyPlayers = mob:getPlayersInRange(30)
```
zach2good, line 820: So does this include ALL players, even those not in the fighting alliance or on the enmity list?

scripts/globals/dark_ixion.lua
```diff
@@ -462,3 +584,292 @@ xi.darkixion.onMobDisengage = function(mob)
     mob:setAggressive(false)
     mob:setLocalVar('StygianLanded', 0)
 end
+
+-- Disable cyclomatic complexity check for this function:
+-- luacheck: ignore 561
```
zach2good, line 589: Please fix this

scripts/globals/dark_ixion.lua
```diff
@@ -462,3 +584,292 @@ xi.darkixion.onMobDisengage = function(mob)
...
+        if
+            mob:isFollowingPath() and
+            os.time() < mob:getLocalVar('runPathTime')
+        then
+            local nearbyPlayers = mob:getPlayersInRange(7)
+            if nearbyPlayers ~= nil then
+                for i = 1, #nearbyPlayers do -- look for players that are too close to ixion while he tramples, hit the ones in front
+                    local stomp = false
+
+                    local dork = nearbyPlayers[i]
```
zach2good, line 739: Please rename this to something more relevant

scripts/globals/dark_ixion.lua
```diff
@@ -462,3 +584,292 @@ xi.darkixion.onMobDisengage = function(mob)
...
+    end
+
+    if mob:getLocalVar('trampling') == 1 then
+        -- cleanly exit trample when reaching the point (TODO check explicitly for a scripted path?)
+        -- timestamp hard exit in case of navmesh abuse
+        if
+            mob:isFollowingPath() and
+            os.time() < mob:getLocalVar('runPathTime')
+        then
+            local nearbyPlayers = mob:getPlayersInRange(7)
```
zach2good, line 734: Like before, will this hit non-engaged players? Spectators? etc.

src/map/lua/lua_baseentity.cpp
zach2good: I'll probably refactor targetfind on Sunday, Eden has some tasty treats I've been meaning to take since 2020 and now is probably a good time to do that

MowFord: Interesting yea it didn't occur to me to use target find for this. Doesn't targetfind technically also loop over everyone in the zone? 

I agree it would be cleaner to use that system though. I've learned since making the PW and DI PRs that I should be breaking off these core changes to separate PRs to get the features I need for the main PR

scripts/items/pinch_of_stygian_ash.lua
MowFord: True. also this code i believe spits out a warning in the luabaseentity constructor for PTarget when a ranged attack is interupted

scripts/globals/dark_ixion.lua
MowFord: yes. can even hit people on a chocobo

scripts/items/pinch_of_stygian_ash.lua
MowFord: btw I haven't touched this in a while, but 
> I (Ixion) install a listener for when I'm hit with a ranged attack. If that ranged attack is from a player with Stygian Ash equipped, do X.

Note that when the ranged attack finishes, the item is no longer equipped. My original implementation (in my previous PR that got accidentally deleted due to branch deletion) was to create a new AE type that would be re-usable for other arbitrary AE types. Could make it something like `setLocalVar(itemID.."hit")`... but I digress, as it was requested to use a listener.

The listener mechanism in this PR works well and is simple to bypass the issue of single-use ranged items being gone once the hit is registered. I've thought a few times and I just don't know a clean way to do it your suggested way. It would also be a listener that gets called throughout the whole fight. Also, imho it's more obvious where to see how the stygian ash works by looking at the ash lua file

scripts/globals/dark_ixion.lua
TeoTwawki: kinda late for me to mention this, bit Nidhog's spike flail is the same way - he'll whip your rear for being in the vicinity. I forget whether that was handled with a target flag or a skill flag. Had this been a skill script instead of a mob script, it'd get handled for you.

MowFord: In this case we aren't actually triggering an ability, we're choosing a target to run to then every tick looking for nearby targets to see if they get trampled

TeoTwawki: I know, I wasn't confused. its unfortunate the only place we have a pre-made handler of it is in mobskills.

MowFord: I believe Zach said he'd implement something for lua to arbitrarily use target find

MowFord:
rebased the branch

PR #4873, opened 2023-12-17
[Core] Add isTrust() lua function

Xaver-DaRed:
**_I affirm:_**

Adds functionality to check if an entity is a trust in lua.

None atm, unless you create a custom module to use it to check if an entity is a trust. Which you can now do.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -1450,6 +1450,18 @@ bool CLuaBaseEntity::isPet()
 }
 
 /************************************************************************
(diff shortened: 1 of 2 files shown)


WinterSolstice8 (approved):

CatsEyeXI (approved):

zach2good:
No need for this, we use getObjType everywhere to differentiate base entities into whatever type

zach2good:
I suppose this is harmless, given that we have bindings for all the others 🤷🏻‍♂️

zach2good (approved):

PR #4876, opened 2023-12-17
Setting: EXP_PARTY_GAP_NO_EXP

CatsEyeXI:
**_I affirm:_**

Provides an optional setting for server operators to use that will nullify a party member's experience points if the level difference with the highest-level party member exceeds a given value.

1. Join an exp party with two level 75's, both will get experience points,
2. Adjust the value of the setting as desired, and !changejob to a level past the allowed threshold and the offending party member will receive no exp.

### DIFF
--- a/settings/default/map.lua
+++ b/settings/default/map.lua
@@ -55,6 +55,8 @@ xi.settings.map =
     EXP_RATE                = 1.0,
     EXP_LOSS_RATE           = 1.0,
     EXP_PARTY_GAP_PENALTIES = true,
+    -- A party member's experience points are nullified if the level difference with the highest-level party member exceeds this value.
+    EXP_PARTY_GAP_NO_EXP    = 0,
 
     -- Capacity Point Settings
     CAPACITY_RATE = 1.0,
--- a/settings/default/map.lua
+++ b/settings/default/map.lua
@@ -55,8 +55,11 @@ xi.settings.map =
     EXP_RATE                = 1.0,
     EXP_LOSS_RATE           = 1.0,
     EXP_PARTY_GAP_PENALTIES = true,
+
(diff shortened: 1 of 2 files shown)


src/map/utils/charutils.cpp
```diff
@@ -4138,7 +4138,11 @@ namespace charutils
                 {
                     if (settings::get<bool>("map.EXP_PARTY_GAP_PENALTIES"))
                     {
-                        if (maxlevel > 50 || maxlevel > (memberlevel + 7))
+                        if (settings::get<uint8>("map.EXP_PARTY_GAP_NO_EXP") > 0 && maxlevel >= (memberlevel + settings::get<uint8>("map.EXP_PARTY_GAP_NO_EXP")))
```
zach2good: Instead of calling this twice, extract this out into:
```
uint8 partyGapNoExp = settings::get<uint8>("map.EXP_PARTY_GAP_NO_EXP");
```

Then use it, it'll be less wide and easier to read

settings/default/map.lua
```diff
@@ -55,6 +55,8 @@ xi.settings.map =
     EXP_RATE                = 1.0,
     EXP_LOSS_RATE           = 1.0,
     EXP_PARTY_GAP_PENALTIES = true,
+    -- A party member's experience points are nullified if the level difference with the highest-level party member exceeds this value.
+    EXP_PARTY_GAP_NO_EXP    = 0,
```
zach2good: I'd add another linebreak before the comment and make the comment explain in more detail what this does, what the accepted number range is, and a worked example of what it does. If its away from the other exp settings then the = won't need to be aligned with the rest

zach2good (approved):

PR #4878, opened 2023-12-19
[Synthesis] Add new crafting type for "no material loss" recipes

Xaver-DaRed:
**_I affirm:_**

AKA Lu-Shang fishing rod, which is notorious for never loosing the base material (broken rod) when failing the synthesis.
- Adds new type of crafting type for said recipes. (Currently only 2 known by me)
- Fixes an issue with HQ rate modifier from food/gear operating at a different scale, returning laughable values.

Additional note:
- Renamed zone enum for zone 0. It was unused but labeled as "Residential area", when it isnt. Its the monorail unfinished zone.

Get Woodworking to apropiate level, get broken lushang or broken lushang +1 and craft away. Never loose materials.

### DIFF
--- a/src/map/trade_container.h
+++ b/src/map/trade_container.h
@@ -36,8 +36,9 @@
 
 enum CRAFT_TYPE
 {
-    CRAFT_SYNTHESIS   = 0,
-    CRAFT_DESYNTHESIS = 1,
+    CRAFT_SYNTHESIS         = 0,
+    CRAFT_DESYNTHESIS       = 1,
+    CRAFT_SYNTHESIS_NO_LOSS = 2,
 };
 
 class CItem;
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -658,80 +658,94 @@ namespace synthutils
      *                                                                         *
      **************************************************************************/
 
+    int32 handleMaterialLoss(CCharEntity* PChar)
+    {
+        uint8 currentCraft     = PChar->CraftContainer->getInvSlotID(0);
+        int16 synthDifficulty  = getSynthDifficulty(PChar, currentCraft);
+        int16 modSynthFailRate = PChar->getMod(Mod::SYNTH_FAIL_RATE);
+
+        // We are able to get the correct elemental mod here by adding the element to it since they are in the same order
+        double reduction = PChar->getMod((Mod)((int32)Mod::SYNTH_FAIL_RATE_FIRE + PChar->CraftContainer->getType()));
+
+        // Similarly we get the correct craft mod here by adding the current craft to it since they are in the same order
+        reduction += PChar->getMod((Mod)((int32)Mod::SYNTH_FAIL_RATE_WOOD + (currentCraft - SKILL_WOODWORKING)));
+        reduction /= 100.0;
+
+        uint8 invSlotID  = 0;
+        uint8 nextSlotID = 0;
+        uint8 lostCount  = 0;
+        uint8 totalCount = 0;
+
+        double random   = 0;
+        double lostItem = std::clamp(0.15 - reduction + (synthDifficulty > 0 ? synthDifficulty / 20 : 0), 0.0, 1.0);
+
+        // Translation of JP wiki for the "Synthesis failure rate" modifier is "Synthetic material loss rate"
+        // see: http://wiki.ffo.jp/html/18416.html
+        lostItem += (double)modSynthFailRate * 0.01;
+
+        invSlotID = PChar->CraftContainer->getInvSlotID(1);
+
+        for (uint8 slotID = 1; slotID <= 8; ++slotID)
+        {
+            if (slotID != 8)
+            {
+                nextSlotID = PChar->CraftContainer->getInvSlotID(slotID + 1);
+            }
+
+            random = xirand::GetRandomNumber(1.);
+
+            if (random < lostItem)
+            {
+                PChar->CraftContainer->setQuantity(slotID, 0);
+                lostCount++;
+            }
+            totalCount++;
+
+            if (invSlotID != nextSlotID)
+            {
+                CItem* PItem = PChar->getStorage(LOC_INVENTORY)->GetItem(invSlotID);
+
+                if (PItem != nullptr)
+                {
+                    PItem->setSubType(ITEM_UNLOCKED);
+                    PItem->setReserve(PItem->getReserve() - totalCount);
+                    totalCount = 0;
+
+                    if (lostCount > 0)
+                    {
+                        charutils::UpdateItem(PChar, LOC_INVENTORY, invSlotID, -(int32)lostCount);
+                        lostCount = 0;
+                    }
+                    else
+                    {
+                        PChar->pushPacket(new CInventoryAssignPacket(PItem, INV_NORMAL));
+                    }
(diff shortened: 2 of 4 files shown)


src/map/trade_container.h
```diff
@@ -36,8 +36,9 @@
 
 enum CRAFT_TYPE
 {
-    CRAFT_SYNTHESIS   = 0,
-    CRAFT_DESYNTHESIS = 1,
+    CRAFT_SYNTHESIS        = 0,
+    CRAFT_DESYNTHESIS      = 1,
+    CRAFT_NO_MATERIAL_LOSS = 2,
```
zach2good: I wouldn't class this as an entirely new type of craft, so I'd probably call this `CRAFT_SYNTHESIS_NO_MATERIAL_LOSS` - even if it is super long

src/map/utils/synthutils.cpp
```diff
@@ -96,7 +96,7 @@ namespace synthutils
                 PChar->CraftContainer->setItem(10 + 2, (uint16)sql->GetUIntData(11), (uint8)sql->GetUIntData(15), 0); // RESULT_HQ
                 PChar->CraftContainer->setItem(10 + 3, (uint16)sql->GetUIntData(12), (uint8)sql->GetUIntData(16), 0); // RESULT_HQ2
                 PChar->CraftContainer->setItem(10 + 4, (uint16)sql->GetUIntData(13), (uint8)sql->GetUIntData(17), 0); // RESULT_HQ3
-                PChar->CraftContainer->setCraftType((uint8)sql->GetUIntData(18));                                     // Store if it's a desynth
+                PChar->CraftContainer->setCraftType((uint8)sql->GetUIntData(18));                                     // Store synh type (regular, desynth or "no material loss")
```
zach2good: * synth

src/map/utils/synthutils.cpp
```diff
@@ -659,77 +659,80 @@ namespace synthutils
 
     int32 doSynthFail(CCharEntity* PChar)
     {
-        uint8 currentCraft     = PChar->CraftContainer->getInvSlotID(0);
-        int16 synthDifficulty  = getSynthDifficulty(PChar, currentCraft);
-        int16 modSynthFailRate = PChar->getMod(Mod::SYNTH_FAIL_RATE);
+        if (PChar->CraftContainer->getCraftType() != CRAFT_NO_MATERIAL_LOSS) // If it's an unbreakable synth (Ex: Lu Shan), skip break calculations.
```
zach2good: Instead of wrapping everything in this new condition, you can see that the function only returns 0 (ie, the return is useless).

So instead you can avoid touching the whole rest of the function by returning early if its the new type:
```cpp
if (PChar->CraftContainer->getCraftType() == CRAFT_NO_MATERIAL_LOSS)
{
    return 0
}
```

src/map/utils/synthutils.cpp
```diff
@@ -659,77 +659,80 @@ namespace synthutils
...
-            }
-            nextSlotID = 0;
-            if (invSlotID == 0xFF)
-            {
-                break;
             }
         }
 
-        if (PChar->loc.zone->GetID() != 255 && PChar->loc.zone->GetID() != 0)
+        if (PChar->loc.zone->GetID() != 255 && PChar->loc.zone->GetID() != 0) // Why?
```
zach2good: While you're here, can you swap these out for the enums from zone.h so this is more readable?
This is zone 0: unknown and zone 255: Abyssea-Empyreal_Paradox

and maybe we should think about zones where crafting isn't allowed and expand this list?

src/map/utils/synthutils.cpp
Xaver-DaRed: I thought about doing that at first, however, we still need the packet pushes at the end of the function, for the messages to player and nearby victims, so if I were to return early, I would need to duplicate that logic.

Not oposing, just wondering if thats the best course of action.

src/map/trade_container.h
Xaver-DaRed: How about `CRAFT_SYNTHESIS_NO_LOSS` ?

src/map/utils/synthutils.cpp
Xaver-DaRed: This check I find it very weird. What it does is just prevent sending the SYNTH_FAIL message to players in range. Thats it.

I will update the ids with their enums, I was just wondering the need for this check for those 2 concrete zones.

src/map/zone.h
```diff
@@ -42,8 +42,7 @@ enum ZONEID : uint16
 {
     // Note: "residential zones" aren't really zones of their own.
     // It's more of a sub zone - the dats for messages and entities will all be from the zone you entered from.
-    ZONE_RESIDENTIAL_AREA = 0, // Old Tech Demonstration zone from pre-release (aka "the monorail place")
-                               // The Above should NOT be labeled "RESIDENTIAL_AREA"
+    ZONE_MONORAIL_PRE_RELEASE           = 0, // Old Tech Demonstration zone from pre-release (aka "the monorail place")
```
TeoTwawki, line 45: I used to have this labeled as `ZONE_TECH_DEMO` but someone reverted it.\*

trivia: client dat actually labels it as `unknown` and it is the only zone name in full lowercase.

\*people kept landing there, they thought the label was why

Xaver-DaRed: Everyone knows the legendary monorail XD

Kind of always gave me FF VIII vives tbh

src/map/utils/synthutils.cpp
```diff
@@ -659,77 +660,83 @@ namespace synthutils
...
-            if (invSlotID == 0xFF)
-            {
-                break;
             }
         }
 
-        if (PChar->loc.zone->GetID() != 255 && PChar->loc.zone->GetID() != 0)
+        // Push "Synthesis failed" messages.
+        if (PChar->loc.zone->GetID() != ZONE_MONORAIL_PRE_RELEASE &&
+            PChar->loc.zone->GetID() != ZONE_ABYSSEA_EMPYREAL_PARADOX) // Why?
```
TeoTwawki: `MAX_ZONEID                          = 300,`

max used to be 255, prolly how that got there

src/map/utils/synthutils.cpp
zach2good: If its just a matter of sending one packet, that's fine to go inside the new chunk I'm suggesting, then the whole rest of the function doesn't have to be indented

Xaver-DaRed: Its a bit more complex than a one liner.

I didnt feel comfortable with repeating all this code:
```c++
        // Push "Synthesis failed" messages.
        uint16 currentZone = PChar->loc.zone->GetID();

        if (currentZone &&
            currentZone != ZONE_MONORAIL_PRE_RELEASE &&
            currentZone != ZONE_49 &&
            currentZone < MAX_ZONEID)
        {
            PChar->loc.zone->PushPacket(PChar, CHAR_INRANGE, new CSynthResultMessagePacket(PChar, SYNTH_FAIL));
        }

        PChar->pushPacket(new CSynthMessagePacket(PChar, SYNTH_FAIL, 29695));
```

So instead I opted with separating the material loss logic into a new function called from the original function. I think that is a good compromise

zach2good (approved):

PR #4880, opened 2023-12-19
Add GM command to test renaming non-player entities

TeoTwawki:
**_I affirm:_**

- Sometimes it is helpful to see in game before finalizing a change
- Or maybe you want to prank your players

Does not work on players. Does not alter database. Limited to 16 characters in length (player sized names).

1. `!rename xyz` while cursor target is an NPC. see name became `xyz`
2. `!rename ""` while targeting the same NPC, see name change back to stock.
3. Try both again but adding the NPC's ID and not targeting it, or targeting anything else (won't matter). You can get the ID by using `!getid` if need be,
4. Now repeat the previous steps with a MOB.
5. Now try it on a player, and see the error message.
6. Now try with invalid IDs, and the cursor not on anything, also see same error message.

Enjoy ~~pranking your pla~~- I mean, _testing_, your rename powers.

### DIFF
--- a/scripts/commands/rename.lua
+++ b/scripts/commands/rename.lua
@@ -1,0 +1,41 @@
+-----------------------------------
+-- func: !rename
+-- desc: Renames target NPC or MOB. Limited to 16 character spaces.
+--       Does not work on players. Does not alter database.
+-----------------------------------
+local commandObj = {}
+
+commandObj.cmdprops =
+{
+    permission = 1,
+    parameters = 'si'
+}
+
+local function error(player, msg)
+    player:printToPlayer(msg)
+    player:printToPlayer('!rename (string) (optional npcID/mobID)')
+end
+
+commandObj.onTrigger = function(player, inputString, entityID)
+    local entity = player:getCursorTarget()
+
+    if entityID then
+        entity = GetMobByID(entityID)
+        if not entity then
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

scripts/commands/rename.lua
```diff
@@ -0,0 +1,41 @@
...
+        end
+    end
+
+    if not entity then
+        error(player, "Entity not found")
+        return
+    end
+
+    if inputString == '""' then
+        entity:renameEntity("", true)
```
Xaver-DaRed: Gotta replace the `"` with an `'`

TeoTwawki: _frownyface_

TeoTwawki: I'm leaving the undo input as double quotes though, because I am NOT escaping single quotes inside single quotes for this convention.

scripts/commands/rename.lua
```diff
@@ -0,0 +1,41 @@
...
+
+    if entityID then
+        entity = GetMobByID(entityID)
+        if not entity then
+            entity = GetNPCByID(entityID)
+        end
+    end
+
+    if not entity then
+        error(player, "Entity not found")
```
TeoTwawki: dang it, I missed these ones too

PR #4882, opened 2023-12-19
[CPP] scale +hq mod by 100 to match adjustments to chanceHQ

MowFord:
**_I affirm:_**

After the change in this commit: https://github.com/LandSandBoat/server/commit/bdfcae0bd22e208d4dce55e15e729cc487f6efcf

the modifier for synth HQ rate no longer modifies appropriately. It was originally adding X/512 to the float between 0 and 1, which represented the chance of an HQ result. Now that `chanceHQ` is between 0 and 100, we need to adjust this accordingly.

I haven't done extensive testing on the statistics of the results here, but simply going with logical conclusions based on the changes from the relevant commit.


Xaver-DaRed:
I already have a PR open for this #4878

PR #4888, opened 2023-12-21
[core] Don't mutate ELEVATOR or SHIP NPCs stored name

### LINKED ISSUE #4879: 🐛 Strange getName behavior for some NPCs
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

simple reproducing of the cosmetic issue (notice the missing remainder of the print output for npcid 17363346):

image

but here's the similar thing happening in ShowDebug (added code to the lua getLocalVar function):
```
uint32 CLuaBaseEntity::getLocalVar(std::string const& var)
{
    ShowDebug(fmt::format("LocalVar: {} ({})", m_PBaseEntity->getName(), m_PBaseEntity->id).c_str());
    return m_PBaseEntity->GetLocalVar(var.c_str());
}
```

image

I'm guessing the issue is related to an `@` symbol in the npc name and calling `c_str()`?

This issue report isn't just to report that weird printing issue, though. I'm unable to reproduce the original issue that spawned this rabbit hole, but I'll past the crash text:

image

image

Not sure if `ShowTrace` calls different underlying functions than `ShowDebug` and maybe that's why it crashes vs just doesn't display the rest of the text. But would appreciate some guidance here

zach2good:
**_I affirm:_**

When we call string::resize(), the increased buffer is uninitialised. If this buffer ends up containing a % character, if that gets passed to one of our old-style sprintf-based loggers it'll explode. While I want to migrate to the fmt-based loggers over time, that's not the answer here. We shouldn't be mutating the stored name of entities for the benefit of the client, that should be done on a temporary copy of the name just before the packet is sent.

Tested:
- Got retail caps to compare
- Metalworks elevators still work
- Port Bastok Bridge + Airship still work (got booted off the bridge and it was still up for a little bit, then closed slowly)
- Mhaura ship came and went

No truncation:
image

Closes https://github.com/LandSandBoat/server/issues/4879

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -2088,7 +2088,6 @@ void CLuaBaseEntity::setElevator(uint8 id, uint32 lowerDoor, uint32 upperDoor, u
 
     elevator.zoneID = m_PBaseEntity->loc.zone->GetID();
 
-    elevator.Elevator->name.resize(10);
     CTransportHandler::getInstance()->insertElevator(elevator);
 }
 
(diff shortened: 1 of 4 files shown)


zach2good:
TODO: Go get retail captures to see exactly what's coming in for entity_udpate packet for these model types

zach2good:
This has broken the elevators in Metalworks >.>

zach2good:
Tested and ready to go

WinterSolstice8 (approved):

PR #4891, opened 2023-12-21
[CPP] Slight plumbing changes for spirit pacts to cast outside combat

MowFord:
**_I affirm:_**

This adjusts the behavior of smn spirit pacts to attempt casting outside combat. This plumbing will allow lua-based logic handling for their casting behavior via the use of 

- `xi.pets.avatar.onMobSpawn` to set their initial magic cooldown and adding a roam/combat tick listener to adjust their magic cool afterwards
- `xi.pets.avatar.onMobMagicPrepare` to do the heavy lifting of deciding which spells to cast when

To be clear though, this PR doesn't make the spirits any smarter, it just allows them to randomly buff/heal outside of combat. I will be working on a separate PR to create `scripts/globals/pets/avatar.lua`

Summon light spirit and see that it will randomly cast cures/buffs even while idle.

See that there are no errors when calling `onMobMagicPrepare`.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -425,7 +425,7 @@ bool CMobController::TrySpecialSkill()
 bool CMobController::TryCastSpell()
 {
     TracyZoneScoped;
-    if (!PTarget || !CanCastSpells())
+    if (!CanCastSpells())
     {
         return false;
     }
@@ -446,7 +446,8 @@ bool CMobController::TryCastSpell()
     }
 
     // Try to get an override spell from the script (if available)
-    auto possibleOverriddenSpell = luautils::OnMobMagicPrepare(PMob, PTarget, chosenSpellId);
+    auto PSpellTarget            = PTarget ? PTarget : PMob;
+    auto possibleOverriddenSpell = luautils::OnMobMagicPrepare(PMob, PSpellTarget, chosenSpellId);
     if (possibleOverriddenSpell.has_value())
     {
         chosenSpellId = possibleOverriddenSpell;
--- a/src/map/ai/controllers/pet_controller.cpp
+++ b/src/map/ai/controllers/pet_controller.cpp
@@ -68,6 +68,14 @@ void CPetController::DoRoamTick(time_point tick)
     {
         return;
     }
+    else if (PPet->m_PetID <= PETID_DARKSPIRIT)
+    {
+        // this will respect the pet's mob casting cooldown properties via MOBMOD_MAGIC_COOL
+        if (CMobController::IsSpellReady(0) && CMobController::TryCastSpell())
+        {
+            return;
+        }
+    }
 
     float currentDistance = distance(PPet->loc.p, PPet->PMaster->loc.p);
 


src/map/ai/controllers/pet_controller.cpp
```diff
@@ -68,6 +68,14 @@ void CPetController::DoRoamTick(time_point tick)
     {
         return;
     }
+    else if (PPet->m_PetID <= PETID_DARKSPIRIT)
+    {
+        // this will respect the pet's mob casting cooldown properties via MOBMOD_MAGIC_COOL
+        if (CMobController::IsSpellReady(0) && CMobController::TryCastSpell())
```
zach2good, line 74: Wow, `IsSpellReady` is a terrible function 😆

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -446,7 +446,7 @@ bool CMobController::TryCastSpell()
     }
 
     // Try to get an override spell from the script (if available)
-    auto possibleOverriddenSpell = luautils::OnMobMagicPrepare(PMob, PTarget, chosenSpellId);
+    auto possibleOverriddenSpell = luautils::OnMobMagicPrepare(PMob, PTarget ? PTarget : PMob, chosenSpellId);
```
zach2good: Instead of doing this inline, can you break this out into a section just before that sets PTarget to be PMob if PTarget is nullptr, leaving the lua call to OnMobMagicPrepare clean

MowFord: I was wary to change ptarget in the controller. I think it needs to be null when out of combat

src/map/ai/controllers/pet_controller.cpp
MowFord: Yea... It ain't dumb if it works... Right?

MowFord:
So I ran into something I'm not sure why I didn't anticipate, but `curaga` cannot be cast across alliance members. I think a `lua_spell` binding for `setTargetFlags` would solve our problem. Would that be ok to add to this PR?

Edit: this would be an extension of `spell.cpp`'s `setValidTarget()`

zach2good:
There's already enough going on in this PR, do it another time

src/map/ai/controllers/mob_controller.cpp
zach2good: Then don't change it in the controller, add:

```
auto* PSpellTarget = PTarget ? PTarget : PMob;
```

and fill in the lua binding appropriately

zach2good (approved):

zach2good:
I wonder, will this effect other mobs that cast things on themselves in combat, but not while idle? Haste etc.? Could do with some confirmation here (still merged it)

MowFord:
> I wonder, will this effect other mobs that cast things on themselves in combat, but not while idle? Haste etc.? Could do with some confirmation here (still merged it)

It shouldn't... they use a whole separate section of code instead of `TryCastSpell`:
```
                else if (PMob->GetMJob() == JOB_SMN && CanCastSpells() && PMob->SpellContainer->HasBuffSpells() &&
                         m_Tick >= m_LastMagicTime + std::chrono::milliseconds(PMob->getBigMobMod(MOBMOD_MAGIC_COOL)))
                {
                    // summon pet
                    auto spellID = PMob->SpellContainer->GetBuffSpell();
                    if (spellID)
                    {
                        CastSpell(spellID.value());
                    }
                }
                else if (CanCastSpells() && xirand::GetRandomNumber(10) < 3 && PMob->SpellContainer->HasBuffSpells())
                {
                    // cast buff
                    auto spellID = PMob->SpellContainer->GetBuffSpell();
                    if (spellID)
                    {
                        CastSpell(spellID.value());
                    }
                }
```

PR #4894, opened 2023-12-22
[LUA] Provide more robust casting behavior for light spirit

MowFord:
**_I affirm:_**

I've not found a more thorough outline of summoner spirit behavior than [this guide](https://ffxiclopedia.fandom.com/wiki/Spirit_Guide_by_Spira#Elemental_Spirit_Casting_Timer), so this is what i'm using to base the code on

This PR piggybacks on https://github.com/LandSandBoat/server/pull/4891 (which may need some additional tweaks to cleanly support this code) and does the following:
- creates a listener on the spirit pact's owner for tick and job ability
  - job ability `assault` and `retreat` will reset the casting cooldown and undo the "half cooldown" as well
  - tick listener will refresh the pet's casting cooldown based on the stipulations outlined in the `setMagicCastCooldown` function
- returns `-1` in `onMobMagicPrepare` when the logic determines the spirit shouldn't cast at all from `mob_controller`, which casts as `uint` and forces an early exit out of `TryCastSpell`. This is what may need to be updated in the other PR, as it results in debug warnings in the map server log. However, this behavior is done so that we can more granularly choose spell and target for the spirit pacts. The behavior is used to
  - stop the spirit from casting immediately upon spawn 
  - cast heals/buffs on particular party/alliance members
  - then return `-1` to make `TryCastSpell` not use generic `mob_spell_list` logic to cast a spell
- does some complex logic for light spirit:
  - loops over the whole alliance and finds the people who are yellow/orange/red hp, but doesn't guarantee healing them (tiered chance based on how low the hp is)
  - if a player is chosen to heal, random chance to use either highest cure or curaga
  - if no `cureTarget` is chosen, loops over the player's party to attempt buffs
  - In Buffing Mode, Light Spirit's spirit timer is halved.
    - This is done via a localvar. Utilized when pet is first summoned and after casting an enhancing magic spell
  - There are only 4 types of spells Light Spirit can cast (i.e Protect, Shell, Regen and Haste). Light Spirit cannot cast -ra buffs, bar-spells or -na spells.
  - Light Spirit will overwrite lower tiered Protect and Shell spells but will not overwrite Haste or Regen.
  - Light Spirit does not cast buffs on players who are under the Invisible status.
  - Light Spirit cannot cast buffs on itself.
  - Light Spirit can only cast Regen I and not II or III
  - Light Spirit can be encouraged to cast its buff spells on party members by having it face towards a party member.
    - This is calculated at the same time as the logic for finding a `buffTarget`
    - `buffTarget` is set to `nil` if the light spirit is in combat and the delicate positioning isn't setup

All spirit pacts are affected by the spell casting delay changes
- summon thunder spirit, aggro mob and see that it very quickly will cast a spell
- resummon thunder spirit and `assault` to see that it takes longer (twice as long) to cast a spell
- afaik none of the non-lightspirit pacts have buff spells of any kind, but if they have them defined in their `mob_spell_list` then they'll use them while idle

Light spirit testing is a bit more involved
- build a party/alliance
- set hp low on a few members. see that is will, on average, heal the lower HP player (but not always, and sometimes won't heal anyone even with multiple in yellow hp)
- position a party member between master and light spirit to see that it will always choose that player to buff (if it doesn't cure anyone)
- set the light spirit to engage a mob, see that it mostly casts nukes (same healing behavior as above though). It will also buff if a player is in the very particular position between light spirit and mob
  - > In addition to the spell repertoire for its offensive mode, the Light Spirit can still cast buffing mode spells while it is engaged.


MowFord:
Disabled cyclomatic complexity check on `xi.pets.avatar.getLightSpiritSpell`. Open to suggestions on a better course of action

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+        castingCooldown = castingCooldown - math.floor(xi.summon.getSummoningSkillOverCap(pet) / 3)
+
+        if master:hasStatusEffect(xi.effect.ASTRAL_FLOW) then
+            castingCooldown = castingCooldown - 5
+        end
+
+        local masterLegs = master:getEquipID(xi.slot.LEGS)
+        if
+            masterLegs and
+            (masterLegs == xi.item.SUMMONERS_SPATS or
```
WinterSolstice8: this should be a mod added to these items

MowFord: Was on the fence about that since it's the only item in the game with the effect (along with its +X)

I'll add the mod

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+    --     if not pet:hasStatusEffect(xi.effect.SHOCK_SPIKES) then
+    --         spellID = xi.magic.spell.SHOCK_SPIKES
+    --         spellTarget = pet
+    --     end
+    end
+
+    return spellID, spellTarget
+end
+
+-- Disable cyclomatic complexity check for this function:
```
zach2good: Don't add this exception to get around problems you're creating. We only have this to get around a couple of very problematic functions that existed before we added luacheck coverage.

https://en.wikipedia.org/wiki/Cyclomatic_complexity

The function is loo long, too nested, or some combination of both.

Break out sub-sections into their own local functions.

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+    end
+end
+
+xi.pets.avatar.getSpiritSpell = function(pet)
+    local spellID = 0
+    local spellTarget = nil
+    local petID = pet:getPetID()
+    if petID == xi.petId.LIGHT_SPIRIT then
+        spellID, spellTarget = xi.pets.avatar.getLightSpiritSpell(pet)
+    -- elseif not pet:isEngaged() then
```
zach2good: If the commented-out code isn't needed, remove it?

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
+-----------------------------------
+-- PET: Avatar
```
zach2good, line 2: If the light spirit isn't an avatar, you could move the elemental spirit logic out into their own file

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+local buffModeVar          = 'AVATAR_BUFF_MODE_OFF'
+local lastCastTimeVar      = 'AVATAR_LAST_CASTINGTIME'
+local lastCastTimeStampVar = 'AVATAR_LAST_CAST_TIMESTAMP'
+local playerListenerVar    = 'SMN_SPIRIT_CAST_DELAY'
+
+-- will determine, in seconds, the spirit's casting cooldown
+-- Called every tick for player to adjust magic casting delay in real-time
+local setMagicCastCooldown = function(pet)
+    local castingCooldown = 45
+    local master = pet:getMaster()
```
zach2good, line 28: There's a few sections in here where you nest the entire function inside `if master`, you can reduce the nesting by exiting early:

```lua
if not master then
    return
end
```

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+        spellID, spellTarget = xi.pets.avatar.getSpiritSpell(pet)
+
+        -- Final items to cast the spell and ensure cast delay is proper
+        local spell = GetSpell(spellID)
+        if spell then
+            if spell:getSkillType() == xi.skill.ENHANCING_MAGIC then
+                -- half casting delay
+                pet:setLocalVar(buffModeVar, 0)
+            end
+-- print(spellTarget:getName() .. '->' .. spellID)
```
zach2good: Can remove commented out prints, or if this is likely to need to be debugged/tweaked/tuned at some point: replace with a local debug printer, similar to here: 
https://github.com/LandSandBoat/server/blob/base/scripts/globals/chocobo_racing.lua#L19-L25

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+            for _, member in pairs(party) do
+                local tempDistance = pet:checkDistance(member)
+                if
+                    not member:hasStatusEffect(xi.effect.INVISIBLE) and
+                    (tempDistance <= distance or
+                    not buffTarget)
+                then
+                    -- Light Spirit will overwrite lower tiered Protect and Shell spells but will not overwrite Haste or Regen
+                    local tempSpellID = 0
+                    if
```
zach2good, line 451: This block looks suspiciously like it can be written as a table+driver

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+                shellTier = { 4, 2617 }
+            elseif petLvl >= 52 then
+                shellTier = { 3, 2188 }
+            elseif petLvl >= 37 then
+                shellTier = { 2, 1641 }
+            elseif petLvl >= 17 then
+                shellTier = { 1, 1055 }
+            end
+
+            local protectTier = { 0, 0 }
```
zach2good: This block looks suspiciously like it can be written as a table+driver

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+            if petLvl >= 40 then
+                hasteTier = { 1, 15 }
+            end
+
+            local regenTier = { 0, 0 }
+            if petLvl >= 21 then
+                regenTier = { 1, 1 }
+            end
+
+            local shellTier = { 0, 0 }
```
zach2good: This block looks suspiciously like it can be written as a table+driver

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+                    spellID = xi.magic.spell.CURAGA_IV
+                elseif petLvl >= 51 then
+                    spellID = xi.magic.spell.CURAGA_III
+                elseif petLvl >= 31 then
+                    spellID = xi.magic.spell.CURAGA_II
+                else
+                    spellID = xi.magic.spell.CURAGA
+                end
+            else
+                if petLvl >= 61 then
```
zach2good: This block looks suspiciously like it can be written as a table+driver

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+        end
+
+        if cureTarget then
+            -- curaga can be used on anyone in the alliance and isn't determined by how many need a heal (assumed to be based on skill over cap)
+            if
+                petLvl >= 16 and
+                math.random(100) <= 2 * xi.summon.getSummoningSkillOverCap(pet)
+            then
+                -- curaga
+                if petLvl >= 71 then
```
zach2good: This block looks suspiciously like it can be written as a table+driver

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,385 @@
...
+        local hpp       = 0
+        local alliance  = master:getAlliance()
+        local party     = master:getParty()
+        local petLvl    = pet:getMainLvl()
+
+        for _, member in pairs(alliance) do
+            -- tiered chance of healing based on how low they are
+            local tempHPP = member:getHPP()
+            if
+                pet:checkDistance(member) < 20 and
```
zach2good: This if statement is very condensed and confusing, can you break it out into sub-conditions with relevant names, or into multiple checks wrapped in a table?

scripts/globals/pets/avatar.lua
MowFord: I couldn't for the life of me think of a clean way to make this a table

scripts/globals/pets/avatar.lua
zach2good: Untested. Something like this:
```lua
local petLvl = 10

local tbl =
{
    [68] = { 4, 2617 },
    [52] = { 3, 2617 },
    [37] = { 2, 2617 },
    [17] = { 1, 2617 },
}

-- TODO: Iterate in order so you match highest or lowest, or whatever
for key, val in pairs(tbl) do
    if petLvl >= key then
        shellTier = val
        break
    end
end

```

claywar: Following up as well on this one, please if you do use the above as an example, don't use key, val as variable names

MowFord:
OK i've revamped the core logic using tables for the buff and cure data. Wouldn't be a MowFord PR without accidentally force-pushing without any new commits.

Going to go over the review notes now to comment which I think can be closed

scripts/globals/pets/avatar.lua
MowFord: Done. Hopefully you like it. I'm pretty happy of the results

scripts/globals/pets/avatar.lua
MowFord: Done. Hopefully you like it. I'm pretty happy of the results

scripts/globals/pets/avatar.lua
MowFord: Done. Hopefully you like it. I'm pretty happy of the results. 

I stored the spells in an ordered list (highest level first) using the effect as a key. Loop and break out when a spell is available to the pet.

scripts/globals/pets/avatar.lua
MowFord: Done. Hopefully you like it. I'm pretty happy of the results

scripts/globals/pets/avatar.lua
MowFord: Done. Hopefully you like it. I'm pretty happy of the results

scripts/globals/pets/avatar.lua
MowFord: Done

scripts/globals/pets/avatar.lua
MowFord: So there's quite a few places i've seen that pet type is defined/used. I think in general summoner pets are all flagged as "avatar", so I piggybacked off that structure.

scripts/globals/pets/avatar.lua
MowFord: Good point, done

scripts/globals/pets/avatar.lua
MowFord: Aye, I had done this just to get the rest of the CI to complete. TY for the suggestions below I'm pretty happy with the tidiness now

scripts/globals/pets/avatar.lua
MowFord: I like it, done

zach2good:
> stored the spells in an ordered list

A table isn't guaranteed to be iterated in the order you specify when you
use pairs() on it, I'm not sure if ipairs makes that guarantee, but it's
not a drop in replacement for how it works

scripts/globals/pets/avatar.lua
MowFord: Cleaned this up to be more obvious how the logic flows. Still works to have a chance of not healing yellow/orange/red hp players:

image

MowFord:
> A table isn't guaranteed to be iterated in the order you specify when you use pairs() on it, I'm not sure if ipairs makes that guarantee, but it's not a drop in replacement for how it works

The tables were defined without keys, which I thought auto-incremented numerical keys and ensured the order from pairs. In many spell casts I never saw it choose out of order (lower tier spell or map server indicating the spell is uncastable), but regardless it seems `ipairs` does ensure order by explicitly checking index 1, index 2, etc until it finds a nil return, so I've changed the calls to that. Good catch there as `ipairs` definitely seems to one to use when expecting order from well-defined, natural-number indexes 

Working well now:
image

zach2good:
In toy examples I can't get it to trip up, but I've absolutely seen pairs()
do things I wouldn't expect :(

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,510 @@
...
+xi.pets.avatar = {}
+
+local buffModeVar          = 'AVATAR_BUFF_MODE_OFF'
+local lastCastTimeVar      = 'AVATAR_LAST_CASTINGTIME'
+local lastCastTimeStampVar = 'AVATAR_LAST_CAST_TIMESTAMP'
+local playerListenerVar    = 'SMN_SPIRIT_CAST_DELAY'
+
+local printDebug = function(pet, textToPrint)
+    -- prints to map server if pet has local var
+    if pet:getLocalVar('debug') == 1 then
```
zach2good, line 19: good choice, this makes more sense than a script-level debug flag 👍

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,510 @@
...
+            end
+        end)
+
+        master:addListener('ABILITY_USE', playerListenerVar .. 'ABILITY', function(playerArg, target, ability, action)
+            local petArg = playerArg:getPet()
+            local abilityName = string.lower(ability:getName())
+
+            if
+                petArg and
+                (abilityName == 'assault' or
```
zach2good: Can we not use the ability IDs here? we typically don't use strings for this sort of thing

MowFord: ah we certainly can, i'm still in the old mindset of not having enums for _everything_

TeoTwawki:
@MowFord @zach2good 

each of `pairs` and `ipairs` have their own gotchas. I often forget which one is which when I need to know which problem I would face, so I looked it up before posting:

> ipairs() returns index-value pairs and is mostly used for numeric tables. **The non-numeric keys are ignored as a whole**, similar to numeric indices less than 1. **In addition, gaps in between the indexes lead to halts.** The ordering is deterministic, by numeric magnitude.
> 
> pairs() returns key-value pairs and is mostly used for associative tables. All keys are preserved, **but the order is unspecified.**

So you just have to make sure the bolded parts won't affect the place you use'em in and use the right one for your case. Though as a whole, I try to avoid needing either one as much as I can.

Also iirc didn't SOL have a problem with `pairs` not to long ago? But I think `ipairs` was ok. Hopefully I do not have that backwards and or that is no longer the case. :)

p.s.

When you don't assign keys at all, lua automatically assigns a numeric one behind your back
`a = {"one", "two", "three"},` = `a = { [1] = "one", [2] = "two", [3] = "three" }`

This is another reason why I prefer to just put them in even if it looks like they aren't needed - they are going to exist anyway I'd rather they not be invisible in my own code.

TeoTwawki:
p.s.

When you don't assign keys at all, lua automatically assigns a numeric one behind your back
`a = {"one", "two", "three"},` = `a = { [1] = "one", [2] = "two", [3] = "three" }`

This is another reason why I prefer to just put them in even if it looks like they aren't needed - they are going to exist anyway I'd rather they not be invisible in my own code.

MowFord:
Thanks @TeoTwawki for confirming what I had in my head. Yep I think in general `ipairs` only makes sense to use if you _don't_ define keys (not sure I'll ever see eye to eye with you on not defining keys when they are strictly increasing, natural-number indexes)

In this particular case i have essentially an array of spells in descending order of level so `ipairs` and not caring explicitly about the key of each entry makes perfect sense.

TeoTwawki:
@MowFord pretty much only works because of lua secretly giving them that number key, which `ipairs` requires.

MowFord:
> @MowFord pretty much only works because of lua secretly giving them that number key, which `ipairs` requires.

Yep, and I don't trust myself to properly define sequential keys the way lua would automatically

MowFord:
Pushed a much simpler method of resetting cast cooldown for retreat and assault

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,501 @@
...
+            then
+                -- Light Spirit will overwrite lower tiered Protect and Shell spells but will not overwrite Haste or Regen
+                local tempSpellID = 0
+
+                for _, spellData in ipairs(lightSpiritBuffs) do
+                    if
+                        not member:getStatusEffect(spellData.effect) or
+                        member:getStatusEffect(spellData.effect):getPower() < spellData.power
+                    then
+                        -- exit loop once a buff is found to be applied
```
TeoTwawki, line 455: aaand if there is none to apply?

ed:

`-- exit loop once a buff is found to be applied or end of table reached`

was what I was looking for, because a lot of users haven't learned tabling or loops so its better to spell it out

MowFord: then `tempSpellID` is left 0 and the function exits without setting a spellID

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,501 @@
...
+        castingCooldown = castingCooldown - 5
+    end
+
+    local masterLegs = master:getEquipID(xi.slot.LEGS)
+    if
+        masterLegs and
+        (masterLegs == xi.item.SUMMONERS_SPATS or
+        masterLegs == xi.item.SUMMONERS_SPATS_P1 or
+        masterLegs == xi.item.SUMMONERS_SPATS_P2)
+    then
```
zach2good: This needs to be an item mod

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,501 @@
...
+    end
+
+    pet:setMobMod(xi.mobMod.MAGIC_COOL, lastCastTime + castingCooldown)
+end
+
+xi.pets.avatar.onMobSpawn = function(pet)
+    local master = pet:getMaster()
+    if
+        not master or
+        master:getObjType() ~= xi.objType.PC
```
zach2good, line 101: Are there mobs that can cast light spirit and have it do things?

scripts/globals/pets/avatar.lua
```diff
@@ -0,0 +1,501 @@
...
+        master:removeListener(playerListenerVar .. 'ABILITY')
+    end
+end
+
+xi.pets.avatar.onMobMagicPrepare = function(pet)
+    -- Note that:
+    -- returning -1 in this function forces TryCastSpell to exit without choosing/casting a spell, but
+    -- will still set the m_LastMagicTime to ensure next call of this function is after the cast delay
+    -- Also, if we return nothing (or zero) TryCastSpell will default to normal mob casting behavior (nukes from spell list, etc)
+    printDebug(pet, 'onMobMagicPrepare: ' .. pet:getMobMod(xi.mobMod.MAGIC_COOL)) -- for debugging magic cooldown
```
zach2good: For all of your debug printing, can you use `string.format` or the new `fmt` helper? I _believe_ I was talking to Teo recently about Lua concat being unsafe, but I don't remember the details. Also Lua concat makes my eyes bleed

zach2good (commented):
Seems fine to me, but I don't know anything about Summoner 🤷

scripts/globals/pets/avatar.lua
MowFord: Yes there are, I am not sure how mob light spirits are supposed to behave

scripts/globals/pets/avatar.lua
MowFord: Fair enough, that was one of those temp-to-perm debug prints

MowFord: Done.

scripts/globals/pets/avatar.lua
MowFord: Done

MowFord:
Also note that i've changed the returns from `-1` to use a variable `dummySpell`. The intent here is to not flood logs with this
image

The logic there is described in `avatar.lua` but i'll paste here:
```
xi.pets.avatar.onMobMagicPrepare = function(pet)
    -- Note that:
    -- returning -1 (or a spell the spirit cannot cast) in this function forces TryCastSpell to exit without choosing/casting a spell, but
    -- will still set the m_LastMagicTime to ensure next call of this function is after the cast delay
    -- Also, if we return nothing (or zero) TryCastSpell will default to normal mob casting behavior (nukes from spell list, etc)
```

scripts/globals/pets/avatar.lua
TeoTwawki: yeah concat'ng things can lead to crashes in some cases, that fmt or `string.format()` will catch and turn into neat little "here's what you did vs what I was expecting" messages

scripts/globals/pets/avatar.lua
TeoTwawki: a typical yagudo's elemental (just for example - each mob race has a few different elements they can use, I forget who has light besides tonberry) will prioritize nuking you regardless of element - I'm not sure if they can even buff, they don't normally live long enough to find out. the ones summoned by the one tonberry NM for carby mitts have more complicated actions akin to a players light spirit (I've seen it cure). if anyone wants additional info, I can go hold some elementals to find out just let me know.

MowFord: fwiw it sounds like in both of those cases it behaves just like they currently do without my updated logic. Just randomly choosing spells from their spell list

src/map/modifier.h
```diff
@@ -994,7 +995,7 @@ enum class Mod
     // 570 through 825 used by WS DMG mods these are not spares.
     //
     // SPARE IDs:
-    // 138 to 143
+    // 141 to 143
```
zach2good, line 998: 👍

zach2good (approved):

sql/item_mods.sql
```diff
@@ -27828,6 +27830,7 @@ INSERT INTO `item_mods` VALUES (15593,363,10); -- HIGH_JUMP_ENMITY_REDUCTION: 10
 INSERT INTO `item_mods` VALUES (15594,1,30);  -- DEF: 30
 INSERT INTO `item_mods` VALUES (15594,5,25);  -- MP: 25
 INSERT INTO `item_mods` VALUES (15594,357,2); -- BP_DELAY: 2
+INSERT INTO `item_mods` VALUES (15594,140,5); -- SPIRIT_CAST_REDUCTION: 5
```
Xaver-DaRed, line 27833: Could this modifiers be placed in order?
That is, mod 140 before mod 357

MowFord: I... never noticed that pattern.

Absolutely, update coming shortly

src/map/modifier.h
MowFord: yea _somebody_ didn't increment last time (it's me, hi, I'm the problem, it's me)

sql/item_mods.sql
TeoTwawki: yep we go by primary key anytime a table has one. which in this case is
```sql
  PRIMARY KEY (`itemId`,`modId`)
```

👍

Xaver-DaRed (approved):

PR #4897, opened 2023-12-23
fix SATA stat mods not working under 100

lapislosh:
**_I affirm:_**

Fixes an integer division bug when applying Mod::TRICK_ATK_AGI and Mod::SNEAK_ATK_DEX that would truncate results before multiplying them, effectively causing the final multiplier to always be 1.0 instead of the intended value.

!additem 14895
!additem 23224
Try to trick attack or sneak attack onto mobs with and without the appropriate armlets equipped and hopefully notice a small difference in damage


TeoTwawki (approved):
the frequency with which folks forget floats in c++ need that little `f` is quite a problem

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #4901, opened 2023-12-24
[LUA] WS dual wield and H2H tp gain fixes

### LINKED ISSUE #4809: 🐛 H2H ws tp return
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

I will do more due dilligence in a bit with testing, but wanted to make this now before i forget: h2h ws appear to not return proper tp after the recent changes. Backhand blow and raging fists both return the same amount of tp on EP mobs over multiple tests

## Expected behavior

TP return should be a function of hit count (high for first hit, but more per landed hit)

MowFord:
**_I affirm:_**

Closes #4809 .

Looks like the refactor had a couple typos/oversights that caused H2H weaponskills to gain zero additional TP from multi-hits and offhand hits. 

Before these changes use backhand blow and raging fists to see you gain only the single main-hand hit TP

afterwards see that backhand blow gives 2 full hits worth of TP and raging fists gets 2 full TP hits plus 10 per additional hit (30 more)

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -666,6 +666,7 @@ xi.weaponskills.calculateRawWSDmg = function(attacker, target, wsID, tp, action,
         finaldmg = finaldmg + hitdmg
         hitsDone = hitsDone + 1
 
+        calcParams.tpHitsLanded      = calcParams.tpHitsLanded + calcParams.hitsLanded
         calcParams.offhandHitsLanded = calcParams.hitsLanded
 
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
was this tested with Jump?

MowFord:
Junp doesn't utilize either variable fixed in this PR, but no I did not explicitly test junp. I'm not even familiar with how tp gain works with jump tbh

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #4907, opened 2023-12-25
[core] Add interrupts for job abilities (pet job abilities)

WinterSolstice8:
**_I affirm:_**

Adds the ability to interrupt job abilities, such as Wyvern Breaths

Use Restoring Breath, have the target run out of range, see "Too far away" message and the wyvern stops animating.


src/map/packets/action.h
```diff
@@ -58,7 +58,8 @@ enum ACTIONTYPE : uint8
     ACTION_RANGED_INTERRUPT     = 32,
     ACTION_MOBABILITY_START     = 33,
     ACTION_MOBABILITY_INTERRUPT = 35,
-    ACTION_RAISE_MENU_SELECTION = 37
+    ACTION_RAISE_MENU_SELECTION = 37,
+    ACTION_JOBABILITY_INTERRUPT = 38,
```
Xaver-DaRed, line 62: Not really YOUR problem, or this PR problem.

However, what are this numbers?
They seem random and with jumps between them, and yet, theres a comment in there saying they aren't IDs?

WinterSolstice8: lua land has a bunch more, they're not defined in C++ for some reason. Was planning to backfill that before undrafting

WinterSolstice8: Oh and for the "not IDs thing", they don't map to real action categories but eventually do get mapped to them (such as shown in this PR in action.cpp)

WinterSolstice8:
Ready for review

src/map/packets/action.h
WinterSolstice8: Taking back that statement about "backfilling before undrafting" -- it looks like there's more here than I knew at a glance. there's some overlap in lua that doesn't match c++...

TeoTwawki (approved):
Looks legit. We're have to sort of those ID overlaps at a later time. Certain presently unused ones could be commented out for now. But dealing with the preexisting ID overlap is outside scope of the current PR.

Xaver-DaRed (approved):

src/map/ai/states/ability_state.cpp
```diff
@@ -189,5 +209,43 @@ bool CAbilityState::CanUseAbility()
         }
         return false;
     }
+    else
+    {
+        if (m_PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_AMNESIA) ||
+            (m_PEntity->StatusEffectContainer->HasStatusEffect(EFFECT_IMPAIRMENT) && (m_PEntity->StatusEffectContainer->GetStatusEffect(EFFECT_IMPAIRMENT)->GetPower() == 0x01 || m_PEntity->StatusEffectContainer->GetStatusEffect(EFFECT_IMPAIRMENT)->GetPower() == 0x03)))
+        {
+            return false;
+        }
```
zach2good: This is a bit hard to read, especially because its a combination or AND and ORs. Can you split this out a bit?

```cpp
bool hasAmnesia = ...;
bool hasImpairment = ...;
bool impairmentPower = ...;

if (...)
```

WinterSolstice8: Updated

zach2good: 👍

PR #4914, opened 2023-12-28
[CPP] Do not consume boost on counter attack

### LINKED ISSUE #4774: 🐛 Counter should not consume Boost buff
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

Boost, get petrified, set counter chance high, and wait to get a counter.

image

boost should not be consumed on counterattack

MowFord:
**_I affirm:_**

Boost should not be consumed when countering an attack. This PR resolves it in a roundabout way. Technically we could check if `isCounter` and check if the player has boost. If they do then save the effect, remove its `effectflag_attack`, then add it back after `DelStatusEffectsByFlag(EFFECTFLAG_ATTACK);`

The current PR implementation will also not consume SA, TA, sneak, invis, etc on a counter. I am not sure if those should be removed when a monk counters, would be a huge pain to test for sure. Though, intuition seems to lead me to believe that counter should work similar to a dodge (since you gain no TP for it), which doesn't remove any of those effects either.

Closes https://github.com/LandSandBoat/server/issues/4774 .

very easy to test using paralysis:
- engage mob
- give 100% paralysis: `!addeffect paralysis 100 600`
- give 100% counter chance `!setmod counter 100`
- give boost effect `!addeffect boost 1`

see before that it wears, after that it doesn't afterwards

image

image

### DIFF
--- a/src/map/attack.cpp
+++ b/src/map/attack.cpp
@@ -398,6 +398,7 @@ bool CAttack::IsCountered() const
 
 bool CAttack::CheckCounter()
 {
+    // TODO return false if boost is active (when boost gets refactored to be current retail accurate)
     if (m_attackType == PHYSICAL_ATTACK_TYPE::DAKEN)
     {
         return false;
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2330,7 +2330,7 @@ namespace battleutils
             ((CMobEntity*)PDefender)->PEnmityContainer->UpdateEnmityFromDamage(PAttacker, 0);
         }
 
-        if (PAttacker->objtype == TYPE_PC && !isRanged)
+        if (PAttacker->objtype == TYPE_PC && !isRanged && !isCounter)
         {
             PAttacker->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_ATTACK);
         }


Xaver-DaRed (dismissed):

TeoTwawki:
caveat: while boost shouldn't be consumed on counter, we probably shouldn't get the benefit of its +attack there either?

WinterSolstice8:
There's evidence to believe it works, I just can't find the testing post on ffxiah

it was something like testing counter + trick attack, can't remember

MowFord:
I'm not concerned either way and can code whichever we choose but:
- it's vastly simpler to leave boost buff and not remove the buff on counter
- there are some things in FFXI that give buffs situationally, and do not reflect in the equipment menu and in `/checkparam`. Boost shows up in both
- It is more likely that SE did a one-off change to not make monk's own ability mess up using boost for weaponskills (which is what made me notice this issue)

TeoTwawki:
> I'm not concerned either way and can code whichever we choose but:
> 
> * it's vastly simpler to leave boost buff and not remove the buff on counter
> * there are some things in FFXI that give buffs situationally, and do not reflect in the equipment menu and in `/checkparam`. Boost shows up in both
> * It is more likely that SE did a one-off change to not make monk's own ability mess up using boost for weaponskills (which is what made me notice this issue)

obviously that would be simpler, when talking about retail accuracy I give exactly zero cares about simpler. but this wouldn't really be complex. annoying, but not complex.

we'd just not have boost directly addmod attack, instead you'd check for a boost power in the places we want boosts attack to count (like in the menu, on regular autoaattacks, ws damage..) and omit it where we don't (counter)

TeoTwawki:
Mowford and I spoke on discord and I've asked him to just take a look at how rough it _would_ be if we need to do it. If there is question as to what retail does with modern boost attack during counter (we don't have yet) that's testable. As to what the older version of boost did or didn't do, we can only go by what data got recorded before SE changed it and or our best guess.

For those unaware, private servers boost effect is what retail used to do: stackable +attack. In modern, it is this awful triple your attack delay in exchange for 1 use of slightly stronger ws or slightly stronger chi blast _thing_ and the new version shall not hit the repo without a module or setting to retain the old version.

WinterSolstice8:
> caveat: while boost shouldn't be consumed on counter, we probably shouldn't get the benefit of its +attack there either?

Boost adds directly to your attack on your equip menu -- i don't see why it wouldn't

edit: or at least it DID add directly to the equip menu, i have no idea what it does now. See below comment.

WinterSolstice8:
https://wiki.ffo.jp/html/2854.html

JP wiki mentions you cannot counter when Boost is active. But LSB is not currently using retail boost.  This implies that you _do_ get some bonus that would activate during counter.

I will approve this PR contingent on a comment indicating that we must disable countering when new-retail boost is implemented

TeoTwawki:
> I will approve this PR contingent on a comment indicating that we must disable countering when new-retail boost is implemented

my own approval was just waiting on him actually checking how many places we'd need to make changes -if- we wanted to do something about it. I wasn't even requiring further change on the pr.

MowFord:
Winter, can you clarify where you want this TODO? I would assume a complete rework of counter would search for any reference to it to clear up this behavior?

WinterSolstice8:
> Winter, can you clarify where you want this TODO? I would assume a complete rework of counter would search for any reference to it to clear up this behavior?

CAttack::CheckCounter() seems logical

TeoTwawki (approved):
Assuming winter is happy post TODO addition, adding my stamp

WinterSolstice8 (approved):

Issue #207, opened 2021-05-26
Party Treasure Pool / Loot Window Display Variance


DarkTannin:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: base

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

Persistent issue with Party Treasure Pool / Loot window displaying differently for multiple players in the same party.  Some players see the standard treasure pool window, while others see the expanded treasure pool window that is only supposed to show up while in Dynamis.  This problem occurs in any and all zones, both in and out of Dynamis.  Tested multiple characters on multiple computers and issue seems to always follow the same character(s).

**Character 1 - Normal Treasure Pool**
![TreasurePool-N](https://user-images.githubusercontent.com/70767280/119745117-55b55c00-be42-11eb-953a-aeeae4095a9a.PNG)

**Character 2 - Dynamis Treasure Pool**
![TreasurePool-D](https://user-images.githubusercontent.com/70767280/119745147-6665d200-be42-11eb-9d50-2805ccaa9ad0.PNG)

Happy to provide any additional screenshots, logs, etc to help in diagnosing and addressing this issue.

zach2good:
It was recently figured out that there's a flag that controls which of these pools you see sent in some packets. I believe it's stuff in:
```
byte 0x20 in packet 0x0D
byte 0x29 in packet 0x37
```

We currently have some overlap with other character flags, so 'small' characters will get the Dyna treasure pool more often.

It also has an effect over if local mobs being fought by others turn up as purple or red.

MowFord:
As Teo mentioned, https://github.com/LandSandBoat/server/pull/4922 is related but based on my knowledge of this subject I think will fully resolve the issue in this direction.

If LSB ever decides to move forward with giving a core option to enable the zone-wide pool then there's also a packet to be set in `CCharPacket::updateWith`. In particular, setting `ref<uint8>(0x20) |= 0x08;` will push the update (as mentioned by zach above)

PR #4922, opened 2023-12-29
[CPP] Mob name color fixes for small player models

MowFord:
**_I affirm:_**

So here's a fun one that almost felt like a hoax when reported. But without this PR, small player characters (char size zero) will see others' claimed mobs with a red name. This is due to the client interpreting `ref<uint8>(0x29) == 0x02`  as old style dynamis where claim is zone-wide. Having a small player model before this PR also shows a different style of treasure pool listing:

image

This could also be fixed by changing the line to 

```
ref<uint8>(0x29) = PChar->GetGender() + (PChar->look.size > 0 ? PChar->look.size * 8 : 0);
```
or simply
```
ref<uint8>(0x29) = PChar->GetGender() + (PChar->look.size * 8);
```

but i'm opting to use the same syntax as https://github.com/airskyboat/airskyboat/pull/1100

have 2 chars next to eachother. The one that has small size character model witnesses the other fighting a mob. Without this PR the mob's name will be red, with this PR the name will be purple.

### DIFF
--- a/src/map/packets/char_update.cpp
+++ b/src/map/packets/char_update.cpp
@@ -68,7 +68,8 @@ CCharUpdatePacket::CCharUpdatePacket(CCharEntity* PChar)
         ref<uint8>(0x38) |= 0x08; // New player ?
     }
 
-    ref<uint8>(0x29) = PChar->GetGender() + (PChar->look.size > 0 ? PChar->look.size * 8 : 2); // + model sizing : 0x02 - 0; 0x08 - 1; 0x10 - 2;
+    ref<uint8>(0x29) = PChar->GetGender() + (PChar->look.size << 3);
+    // ref<uint8>(0x29) |= 0x02; // this will display the old-style zone-wide treasure pool. All claimed mobs will show red name regardless of which party has claim.
     ref<uint8>(0x2C) = PChar->GetSpeed();
     ref<uint16>(0x2E) |= PChar->speedsub << 1; // Not sure about this, it was a work around when we set speedsub incorrectly..
     ref<uint8>(0x30) = PChar->isInEvent() ? (uint8)ANIMATION_EVENT : PChar->animation;


cocosolos:
> This is due to the client interpreting ref<uint8>(0x29) == 0x02 as old style dynamis where claim is zone-wide.

Interesting! Should probably leave a comment for future development.

TeoTwawki:
https://discord.com/channels/392903136336936960/883228057517842432/979753422641324072

not the only spot that byte gets written to. (status effect)

TeoTwawki:
iirc we also discussed making the other treasure pool style an intentional option, but didn't act on this yet pending re-captures from retail and then it got forgotten in the depths of discord.

Also, so this doesn't get lost:
https://github.com/LandSandBoat/server/issues/207

MowFord:
> > This is due to the client interpreting ref(0x29) == 0x02 as old style dynamis where claim is zone-wide.
> 
> Interesting! Should probably leave a comment for future development.

Good call, added

zach2good:
From what you said in the other issue:

> If LSB ever decides to move forward with giving a core option to enable the zone-wide pool ...

In the modern retail game, are there zones that use zone-wide treasure pool? Dyna-D? Sortie? That thing in converted WoE?

zach2good:
(TRANSLATION: Give us a reason to add per-zone settings for this that aren't "muh era" and I'll literally add them tonight.)

zach2good:
Also, if you're linking and using someone else's work (Aerec & Velyn on the ASB PR), you should be co-authoring them. https://github.com/LandSandBoat/server/wiki/Development-Guide-Git#git-attribution

MowFord:
> From what you said in the other issue:
> 
> > If LSB ever decides to move forward with giving a core option to enable the zone-wide pool ...
> 
> In the modern retail game, are there zones that use zone-wide treasure pool? Dyna-D? Sortie? That thing in converted WoE?

I didn't mean to throw shade or anything. Was just extending the conversation Teo had started in this thread.

And yes sorry i'll add the co-author

zach2good:
> I didn't mean to throw shade or anything.

I'm not trying to be sassy; I'm trying to reiterate that this is a retail emulator project and the fastest way to convince us to implement something is to show us that we need it for retail emulation. If its only usage is for custom content and era throwback, it goes immediately to the back of the queue, especially if it requires us to go in and do core work to enable it.

In the case of enabling this flag in the packets, we have a ZONE_MISC flag which I don't fully understand what it does, but it might be relevant to hook up for this: MISC_TREASURE.

We also have cpp modules, where you can add a module to modify outgoing packets:

```cpp
    std::vector<uint16> globalTreasurePoolZones =
    {
        ZONE_DYNAMIS_BUBURIMU,
        ZONE_DYNAMIS_VALKURM,
    };

    void OnPushPacket(CCharEntity* PChar, CBasicPacket* packet) override
    {
        // Is this the correct packet?
        if (packet->getType() != 0x37)
        {
            return;
        }

        // Is this the correct zone?
        if (std::find(globalTreasurePoolZones.begin(), globalTreasurePoolZones.end(), PChar->getZone()) == globalTreasurePoolZones.end())
        {
            return;
        }

        // this will display the old-style zone-wide treasure pool.
        // All claimed mobs will show red name regardless of which party has claim.
        packet->ref<uint8>(0x29) |= 0x02;
    };
```

I need to make a change to the PushPacket module signature, but its all there

zach2good (approved):

PR #4921, opened 2023-12-29
Implement DRK spell Absorb-Attri

CatsEyeXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?
Implements Black Magic Spell: Absorb-Attri
Will be used in implementation of Trust: Zeid II

## Steps to test these changes
image

PR #4925, opened 2023-12-29
Additions to Trust: Zeid II

CatsEyeXI:
**_I affirm:_**

Builds off of the implementation of Zied II to add missing mob skills.
I need some feedback regarding the gambit creation for Absorb-Attri (#4921)
Typically this appears to be handled using `NOT_STATUS`, but in this case we'd need to check for many?
Do we define a set somehow? 

image
image
image

### DIFF
--- a/scripts/actions/spells/trust/zeid_ii.lua
+++ b/scripts/actions/spells/trust/zeid_ii.lua
@@ -16,6 +16,13 @@ spellObject.onMobSpawn = function(mob)
         [xi.magic.spell.LION_II] = xi.trust.messageOffset.TEAMWORK_1,
     })
 
+    mob:addListener('WEAPONSKILL_USE', 'ZEID_II_WEAPONSKILL_USE', function(mobArg, target, wsid, tp, action)
+        if wsid == 56 then -- Ground Strike
+            -- Never again will I lose sight of who I am
+            xi.trust.message(mobArg, xi.trust.messageOffset.SPECIAL_MOVE_1)
+        end
+    end)
+
     -- Stun all the things!
     mob:addSimpleGambit(ai.t.TARGET, ai.c.READYING_WS, 0,
                         ai.r.MA, ai.s.SPECIFIC, xi.magic.spell.STUN)
--- a/scripts/actions/spells/trust/zeid_ii.lua
+++ b/scripts/actions/spells/trust/zeid_ii.lua
@@ -14,6 +14,10 @@ end
 spellObject.onMobSpawn = function(mob)
     xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
 
+    xi.trust.teamworkMessage(mob, {
+        [xi.magic.spell.LION_II] = xi.trust.messageOffset.TEAMWORK_1,
+    })
+
     -- Stun all the things!
     mob:addSimpleGambit(ai.t.TARGET, ai.c.READYING_WS, 0,
                         ai.r.MA, ai.s.SPECIFIC, xi.magic.spell.STUN)
--- a/scripts/actions/spells/trust/zeid_ii.lua
+++ b/scripts/actions/spells/trust/zeid_ii.lua
@@ -12,8 +12,6 @@ spellObject.onSpellCast = function(caster, target, spell)
 end
 
 spellObject.onMobSpawn = function(mob)
-    xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
-
     xi.trust.teamworkMessage(mob, {
         [xi.magic.spell.LION_II] = xi.trust.messageOffset.TEAMWORK_1,
     })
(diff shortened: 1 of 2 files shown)


TeoTwawki:
"Zied" 
image

CatsEyeXI:
> "Zied" image

even fixed the skill list name :)

scripts/actions/spells/trust/zeid_ii.lua
```diff
@@ -14,6 +14,10 @@ end
 spellObject.onMobSpawn = function(mob)
     xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
 
+    xi.trust.teamworkMessage(mob, {
+        [xi.magic.spell.LION_II] = xi.trust.messageOffset.TEAMWORK_1,
```
zach2good, line 16: Huh, I would have thought he had more than just Lion. Get someone to check if he also does a modified line for LION_I and not just LION_II.

What do his message dumps say?

sql/mob_skill_lists.sql
```diff
@@ -3604,7 +3604,9 @@ INSERT INTO `mob_skill_lists` VALUES ('TRUST_AAGK',1111,3726); -- Tachi: Fudo
 -- INSERT INTO `mob_skill_lists` VALUES ('TRUST_Aldo_UC',1122,0);
 -- INSERT INTO `mob_skill_lists` VALUES ('TRUST_Naja_Salaheem_UC',1123,0);
 -- INSERT INTO `mob_skill_lists` VALUES ('TRUST_Lion_II',1124,0);
-INSERT INTO `mob_skill_lists` VALUES ('TRUST_Zied_II',1125,56); -- Ground Strike
+INSERT INTO `mob_skill_lists` VALUES ('TRUST_Zeid_II',1125,56);  -- Ground Strike
+INSERT INTO `mob_skill_lists` VALUES ('TRUST_Zeid_II',1125,708); -- Last Resort
+INSERT INTO `mob_skill_lists` VALUES ('TRUST_Zeid_II',1125,710); -- Souleater
```
zach2good: These are abilities, not weaponskills/mobskills?

zach2good (changes requested):
Zeid II already handles souleater and last resort as JAs: https://github.com/LandSandBoat/server/blob/base/scripts/actions/spells/trust/zeid_ii.lua#L30-L35

Did you see something in packets to indicate that these are executed as mobskills? Or does he fire these off as weaponskills/mobskills in addition to ground strike on retail?

By setting them as weaponskills, (I think?) he might dump TP to try and use them? This looks mighty fishy

scripts/actions/spells/trust/zeid_ii.lua
zach2good: (My local builds are playing up at the moment, I can't dump them for you like I did for Lion I and II)

CatsEyeXI:
Hi Zach. It's a good point and I'll check.

zach2good:
Zeid I's pages:
image

Zeid II's pages:
image

Zeid II also has another section towards the bottom which I _suspect_ are response lines, or used during a skillchain, or in response to other trusts being summoned: but we don't have plumbing for those yet so you can disregard

zach2good:
These are directly related to:
```lua
xi.trust.messageOffset =
{
    SPAWN          = 1,
    TEAMWORK_1     = 4,
    TEAMWORK_2     = 5,
    TEAMWORK_3     = 6,
    TEAMWORK_4     = 7,
    TEAMWORK_5     = 8,
    DEATH          = 9,
    DESPAWN        = 11,
    SPECIAL_MOVE_1 = 18,
    SPECIAL_MOVE_2 = 19,
}

```

scripts/actions/spells/trust/zeid_ii.lua
zach2good: Confirmed: Zeid II only responds to Lion II, not Lion I

sql/mob_skill_lists.sql
CatsEyeXI: you are right, they are abilities and not mob skills that would consume TP. I've removed that from the PR.

scripts/actions/spells/trust/zeid_ii.lua
zach2good: xi.trust.teamworkMessage replaces xi.trust.message(, otherwise you're going to get 2x spawn messages.

WinterSolstice8:
Zeid II says "Never again will I lose sight of who I am" on use of ground strike

CatsEyeXI:
> Zeid II says "Never again will I lose sight of who I am" on use of ground strike

image

zach2good:
This looks ready to go, it just needs squashing (we can do that) and https://github.com/LandSandBoat/server/pull/4921 to merge

zach2good (approved):

PR #4930, opened 2023-12-30
Trust: Tenzen II

CatsEyeXI:
**_I affirm:_**

Implements Trust: Tenzen II
Teamwork chat with Prishe
Uses only ranged attacks, acquires 250TP per hit (it is difficult getting an exact TP Bonus for this, as TP return seemed to vary slightly as the difference in mob levels grew from EP to T)
Will use Oisoya exclusively

Captured behavior from retail
Summon a tank trust, and Tenzen II, he will stand back and ranged attack, waiting to open.


scripts/actions/spells/trust/tenzen_ii.lua
```diff
@@ -12,7 +12,28 @@ spellObject.onSpellCast = function(caster, target, spell)
...
+        end
+    end)
+
+    -- Ranged Attack as much as possible (limited by 'weapon' delay)
+    mob:addSimpleGambit(ai.t.TARGET, ai.c.ALWAYS, 0, ai.r.RATTACK, 0, 0)
+
+    mob:setAutoAttackEnabled(false)
+
+    -- Gets 250 TP per hit
+    mob:addMod(xi.mod.STORETP, 210)
```
zach2good: I forget how I implemented this, but you might be able to avoid pushing this up so high by modifying his weapon stats in mob pools, which are controlled by:
```sql
  `cmbSkill` tinyint(2) unsigned NOT NULL DEFAULT 1,
  `cmbDelay` smallint(3) unsigned NOT NULL DEFAULT 240,
  `cmbDmgMult` smallint(4) unsigned NOT NULL DEFAULT 100,
```

They're still the defaults of: `3,240,100`

You'll also want to make sure these are updated on your other trust PRs. You can see Excenmille having his delay and damage boosted: `8,390,150`, and the opposite for Nanaa Mihgo: `2,190,60`

zach2good: Looks like I did this to tune Joachim's pebble throwing: `3,500,25,`

TeoTwawki: ~~Tenzen acts at 450 delay, and I would think should have the standard samurai traits to whatever degree is required after that. I thought this seemed odd last night but did not think to check his delay in `mob_pools`~~

_nvm me, I am thinking of Tenzen not Tenzen II..  Tenzen 1 acts at 450 (great Katana, derp)_

CatsEyeXI: Zach, having played around with these values for several hours, i was never able to get it quite right.
Increasing the weapon delay in the DB does infact increase the TP return, but at the cost of 20s between each arrow being shot. Even when explicitly defining the delay in the ranged attack gambit..

zach2good: You can't explicitly define the delay for attacks in Gambits? Gambits are just bags of instructions that have no bearing on what the output action does.

See here: the "delay" is just the minimum retry time for checking the gambit again:
https://github.com/LandSandBoat/server/blob/base/src/map/ai/helpers/gambits_container.cpp#L251

This is the mechanism that Kupipi uses to not spam her debuffs if an enemy is immune, since we only have the facility to check if a target has/has_not the status:
https://github.com/LandSandBoat/server/blob/base/scripts/actions/spells/trust/kupipi.lua#L51-L54

What I was getting at here was that adding TWO HUNDRED AND TEN STORE TP as a mod was not the answer. The answer is: fiddle with Tenzen's damage and delay, and then make up the difference with some STP.

CatsEyeXI: Yes, I understood the recommendation and played with the value inside the gambit. Increasing the delay in the DB to the value required that returns the correct amount of TP causes the trust to aim for a really, really long time. I tested this over the course of about 2 hours, changing the values in the database, and in the ranged attack gambit.

CatsEyeXI: https://github.com/LandSandBoat/server/assets/65316697/ee182584-2696-4cac-8015-c38034a2751d

CatsEyeXI: image

Xaver-DaRed: So Carver, since this PR is mostly correct, except for the TP issue, which is SE being SE, for the time being, I think you should comment out the store TP mod, and add a TODO, for the eventual time we understand what SE does with this trusts.

TeoTwawki: To get His TP gains retail accurate I basically need some kind of override to set a specific amount of TP per ranged attack based on his level: it changes at the exact levels he would gain the store TP trait, but retail testing (BY ME) shows there is no valid delay rating that would result in correct amount of TP when factoring in storeTP. Bets bet would be zero his TP gains and then literally just do a conditional for his level and call `addTP()`

TeoTwawki: I've removed my hold on this and pushed a new commit. Tenzen isn't perfect but he's not overpowered, and individual servers can adjust him to their taste till we get a better solution in place. He's like a gimped version of Semih at the moment, which is also how SE treats him from looking at his lower lv TP gains :(

TeoTwawki:
>  trailing whitespace in a comment

I'm surprised that got me, I normally trim trailing on-save, coz I overwrote the normal save function of my editor to trim first! I'll have to see wtf is up with that.

Xaver-DaRed (approved):

PR #4933, opened 2023-12-31
Add functionality to Trust: Prishe II

CatsEyeXI:
**_I affirm:_**

Fills out some of the logic for Trust: Prishe II
I didn't want to make too many changes to the existing mobskill scripts, other than some minor corrections.
Changed jobs from MNK/WHM to WHM/MNK.
Added items Hysteroanima and Pyschoanima
Added partial trust synergy with Ulmia

Based on two separate videos recorded from retail and corroborated with BGWiki
https://drive.google.com/file/d/10BnOcIMXn3Ea2PIM8dVFk9zczYbPCQVQ/view
https://drive.google.com/file/d/1l3Av9QiiRA3xBjIIShkDEskmQmq6zY0P/view?usp=sharing
https://www.bg-wiki.com/ffxi/Cipher:_Prishe_II
https://ffxiclopedia.fandom.com/wiki/Trust:_Prishe_II

!zone Dragons Aery
/ma "Prishe II" <me>
Engage

Watch Prishe fight


scripts/actions/spells/trust/prishe_ii.lua
```diff
@@ -12,7 +12,58 @@ spellObject.onSpellCast = function(caster, target, spell)
...
+                    mobArg:addSimpleGambit(ai.t.PARTY, ai.c.HPP_LT, 75, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.CURAGA)
+                    mobArg:removeSimpleGambit(healingGambit)
+                end
+            end
+        end
+    end)
+
+    mob:addListener('WEAPONSKILL_USE', 'PRISHE_II_WEAPONSKILL_USE', function(mobArg, target, wsid, tp, action)
+        if wsid == 3234 then -- Nullifying Dropkick
+            xi.trust.message(mobArg, xi.trust.messageOffset.SPECIAL_MOVE_1)
```
zach2good, line 40: Typically we add a comment saying what the voice line is

scripts/actions/spells/trust/prishe_ii.lua
```diff
@@ -12,7 +12,58 @@ spellObject.onSpellCast = function(caster, target, spell)
...
+        end
+    end)
+
+    mob:addListener('WEAPONSKILL_STATE_EXIT', 'ANIMA_USED', function(mobArg, wsid)
+        if wsid == psychoAnima then
+            mobArg:removeSimpleGambit(itemOneGambit)
```
zach2good, line 46: 👍

scripts/actions/spells/trust/prishe_ii.lua
```diff
@@ -12,7 +12,58 @@ spellObject.onSpellCast = function(caster, target, spell)
...
+    -- local itemTwoGambit = mob:addSimpleGambit(ai.t.TARGET, ai.c.CASTING_MA, 35, ai.r.MS, ai.s.SPECIFIC, 3540)
+    local healingGambit = mob:addSimpleGambit(ai.t.PARTY, ai.c.HPP_LT, 25, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.CURAGA)
+
+    mob:addSimpleGambit(ai.t.PARTY, ai.c.STATUS, xi.effect.SLEEP_I, ai.r.MA, ai.s.SPECIFIC, xi.magic.spell.CURAGA)
+    mob:addSimpleGambit(ai.t.PARTY, ai.c.STATUS, xi.effect.SLEEP_II, ai.r.MA, ai.s.SPECIFIC, xi.magic.spell.CURAGA)
+
+    -- Trust Synergy: If Ulmia in the party, Prishe will cast Curaga when party members are in the yellow instead of red.
+    -- TODO: Prishe will use Cure spells exclusively on Ulmia (Only Cure I-IV)
+    -- https://ffxiclopedia.fandom.com/wiki/Trust:_Prishe_II
+    mob:addListener('COMBAT_TICK', 'PRISHE_CTICK', function(mobArg)
```
zach2good: Gambit system is designed specifically around the idea of not attaching any logic to a Trust's tick. I'd rather you leave a TODO here for this and I'll add a binding for "TrustAddedToParty" or something for you to use later

scripts/actions/spells/trust/prishe_ii.lua
```diff
@@ -12,7 +12,58 @@ spellObject.onSpellCast = function(caster, target, spell)
 end
 
 spellObject.onMobSpawn = function(mob)
-    xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
+    local psychoAnima  = 3539
+
+    xi.trust.teamworkMessage(mob, {
+        -- Offsets for Lion II and Iroha II messages appear to be incorrect.
```
zach2good: Any more details on this? Whats up?

sql/mob_pool_mods.sql
```diff
@@ -705,6 +705,12 @@ INSERT INTO `mob_pool_mods` VALUES (5922,6,150,0);      -- MPP: 150
 INSERT INTO `mob_pool_mods` VALUES (5944,3,-10,0);      -- HPP: -10
 INSERT INTO `mob_pool_mods` VALUES (5944,6,35,0);       -- MPP: 35
 
+-- Trust: Prishe II
```
zach2good, line 708: I don't remember what MNKs get, but are these in addition to what she'll get by being marked as MNK/WHM in mob_pools?

sql/mob_pools.sql
```diff
@@ -6067,7 +6067,7 @@ INSERT INTO `mob_pools` VALUES (6007,'aldo_uc','Aldo',149,0x0000F90B000000000000
 INSERT INTO `mob_pools` VALUES (6008,'naja_uc','NajaSalaheem',151,0x0000FA0B00000000000000000000000000000000,6,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,417,0,0,1123,151);
 INSERT INTO `mob_pools` VALUES (6009,'lion_ii','Lion',149,0x0000090C00000000000000000000000000000000,6,13,3,240,100,0,0,0,0,0,0,32,0,3,0,0,418,0,0,1124,149);
 INSERT INTO `mob_pools` VALUES (6010,'zeid_ii','Zeid',146,0x00000E0C00000000000000000000000000000000,8,1,3,240,100,0,0,0,0,0,0,32,0,3,0,0,419,0,0,1125,146);
-INSERT INTO `mob_pools` VALUES (6011,'prishe_ii','Prishe',145,0x00000A0C00000000000000000000000000000000,2,3,3,240,100,0,0,0,0,0,0,32,0,3,0,0,420,0,0,1126,145);
+INSERT INTO `mob_pools` VALUES (6011,'prishe_ii','Prishe',145,0x00000A0C00000000000000000000000000000000,3,2,3,240,100,0,0,0,0,0,0,32,0,3,0,0,420,0,0,1126,145);
```
zach2good, line 6070: So after doing Monstrosity, I do wonder if Trusts should be treated as equal level main job and sub job, if this is to line up their traits and stats etc.

scripts/actions/spells/trust/prishe_ii.lua
zach2good: image

zach2good:
```lua
    xi.trust.teamworkMessage(mob, {
        [xi.magic.spell.TENZEN_II] = xi.trust.messageOffset.TEAMWORK_1,
        [xi.magic.spell.NASHMEIRA_II] = xi.trust.messageOffset.TEAMWORK_2,
        [xi.magic.spell.LILISETTE_II] = xi.trust.messageOffset.TEAMWORK_3,
        [xi.magic.spell.ARCIELA_II] = xi.trust.messageOffset.TEAMWORK_4,
        [xi.magic.spell.IROHA_II] = xi.trust.messageOffset.TEAMWORK_5,
    })
```

zach2good: image

scripts/actions/spells/trust/prishe_ii.lua
zach2good: You can approximate bindings for this with this diff:
```diff
diff --git a/scripts/globals/trust.lua b/scripts/globals/trust.lua
index 26746347e5..49296a8b68 100644
--- a/scripts/globals/trust.lua
+++ b/scripts/globals/trust.lua
@@ -267,6 +267,10 @@ xi.trust.onTradeCipher = function(player, trade, csid, rovCs, arkAngelCs)
     end
 end
 
+xi.trust.handleSynergyWithOtherTrusts = function(trust)
+
+end
+
 xi.trust.canCast = function(caster, spell, notAllowedTrustIds)
     -- Trusts must be enabled in settings
     if xi.settings.main.ENABLE_TRUST_CASTING == 0 then
@@ -380,13 +384,15 @@ xi.trust.canCast = function(caster, spell, notAllowedTrustIds)
 end

 xi.trust.spawn = function(caster, spell)
-    caster:spawnTrust(spell:getID())
+    local trust = caster:spawnTrust(spell:getID())

     -- Records of Eminence: Call Forth an Alter Ego
     if caster:getEminenceProgress(932) then
         xi.roe.onRecordTrigger(caster, 932)
     end

+    xi.trust.handleSynergyWithOtherTrusts(trust)
+
     return 0
 end
```

But I haven't thought of a _clean and readable_ way to lay out and track the synergies with other trusts yet, since some of them are singular, some of them are grouped, some add effects to some/all of the group, some add/remove gambits etc.

Open to suggestions

zach2good: Good examples of synergies:
https://ffxiclopedia.fandom.com/wiki/Trust:_Curilla and https://ffxiclopedia.fandom.com/wiki/Trust:_Rainemard
https://ffxiclopedia.fandom.com/wiki/Trust:_Trion and https://ffxiclopedia.fandom.com/wiki/Trust:_Pieuje_(UC)
and I forget where it mentions that some combination of Lion+Zeid+Tenzen each get different things if they're up at the same time

zach2good: Each synergy will need to be wrapped in something that can apply and remove it, since the synergy can/will be broken at any point if a trust is dismissed or dies. If each synergy is a table, then it's fine for the first entry to be a function that applies all the mods/actions/listeners whatever, and then the second entry removes them. In this layout, we'd make sure to remove all synergies every time the function is "tapped", to make sure we're starting on a clean slate. Then trying to re-apply.

CatsEyeXI: I'll try to make sense of this. I pulled this example from Mumor, here https://github.com/LandSandBoat/server/blob/17498511563b0da5175f368ae65ade0bfc703b4b/scripts/actions/spells/trust/mumor.lua#L30 and it
seems to work rather well for this use case, feel like it would be a shame to pull it out, but I can "catseye crap" it for our needs if you don't like it.

sql/mob_pool_mods.sql
CatsEyeXI: the HP and MP mods were added based on testing.
I had a naked level 99 summon her and compared her HP and MP with a naked level 99 on LSB. Added mods until we about matched. What prompted me to even look at this was this comment on BGWiki
image

The double attack is set to 100% because otherwise she will swing with 1 fist per attack round

sql/mob_pools.sql
CatsEyeXI: I could see that being the case, can potentially help research that later on

sql/mob_pool_mods.sql
TeoTwawki: Fix her weapon in her weapon instead, shes 1 hand punching because shes not set to h2h for her weapon in her pool data

scripts/actions/spells/trust/prishe_ii.lua
zach2good: Just because things exist in other parts of the codebase doesn't mean we don't want to improve on them over time. 
```lua
    -- Dynamic modifier that checks party member list on tick to apply
    mob:addListener('COMBAT_TICK', 'MUMOR_CTICK', function(mobArg)
        local sambaDurationBoost = 0
        local party = mobArg:getMaster():getPartyWithTrusts()
        for _, member in pairs(party) do
            if member:getObjType() == xi.objType.TRUST then
                if
                    member:getTrustID() == xi.magic.spell.UKA_TOTLIHN
                then
                    sambaDurationBoost = 10
                end
            end
        end

        -- Always set the boost, even if Uka wasn't found.
        -- This accounts for her being in the party and giving the boost
        -- and also if she dies and the boost goes away.
        mobArg:setMod(xi.mod.SAMBA_DURATION, sambaDurationBoost)
    end)
```

In this example, the listener is firing off every tick, going through the whole party of trusts, looking for the relevant other trust, and _always_ setting the mod. Every tick.

This doesn't have to happen any more often than "the composition of the party has changed". The code _could_ be exactly the same, but the listener could be firing when the composition of the party has changed. That could be a very easy 2-3 line C++ change for you to investigate: Add a new listener trigger when the composition of a party changes, and then hook this logic up to that (and that Mumor block too)

TeoTwawki: > Just because things exist in other parts of the codebase doesn't mean we don't want to improve on them over time.

This is also why we're so hellbent on trying to get things right the 1st time - anytime we don't, inevitably someone cites it as where they copied from. 

Some days feel like we've lost some kind of race to fix something up before someone does.

sql/mob_pool_mods.sql
CatsEyeXI: fixed weapon in mob pools, removed mods

scripts/actions/spells/trust/prishe_ii.lua
CatsEyeXI: removed this section for now and left a TODO

sql/mob_pools.sql
TeoTwawki: @CatsEyeXI are you setting her whm/mnk instead of mnk/whm? 

@zach2good the way we handle mobs traits in general isn't quite right, but they do get from both jobs at lv of main. They aren't like players where they always get the 2nd job halved - they just don't get full stats of their 2nd job instead. Certain standard job traits aren't on all mobs of that job though and at the time we decided to make jobs auto-add their traits we didn't knw this. To get it to 100% retail accurate we'll need the ability to just bolt on or "equip" if you will, traits. Just like how we can arbitrarily add spells. I ***do not*** want people compensating for missing traits by just using a bunch of `addMod` in scripts or annoying sql modifier rows because in the past that lead to people unknowingly stacking multiple sources of the same bonus. That way leads to unsleepable paladins and warriors with for more double attack than ever intended..

TeoTwawki: iirc Frank already set our monsters to have full sj levels and modified stats. Trusts likely need an adjustment to match.

CatsEyeXI: > @CatsEyeXI are you setting her whm/mnk instead of mnk/whm?
> 
> @zach2good the way we handle mobs traits in general isn't quite right, but they do get from both jobs at lv of main. They aren't like players where they always get the 2nd job halved - they just don't get full stats of their 2nd job instead. Certain standard job traits aren't on all mobs of that job though and at the time we decided to make jobs auto-add their traits we didn't knw this. To get it to 100% retail accurate we'll need the ability to just bolt on or "equip" if you will, traits. Just like how we can arbitrarily add spells. I _**do not**_ want people compensating for missing traits by just using a bunch of `addMod` in scripts or annoying sql modifier rows because in the past that lead to people unknowingly stacking multiple sources of the same bonus. That way leads to unsleepable paladins and warriors with for more double attack than ever intended..

There's conflicting information on this, But I do believe her to be a WHM/MNK based on the fact she must have a sufficiently large MP pool for Curaga without any mods.

BGWiki and FFXIClopedia do not agree on this.
image
image

TeoTwawki: she could indeed be either, since trust break the usual rules for mages who engage. But the MP issue tells me we're still stat-ing trusts like players and not mobs. even a war/blm on retail will have a ton of mp, because they aren't having their subjobs mp pool nerfed to hell like us.

zach2good (approved):

PR #4934, opened 2024-01-01
[LUA] clean up sigil buff power calculations

MowFord:
**_I affirm:_**

Currently, SIGIL doesn't give any buffs as the bit shift is coded incorrectly. This resolves both the `selectedEffects` not being always zero as well as the `sigil.lua` magic numbers to be more well-defined.

The pattern to `option` in event finish for applying sigil is 1, 4097, 8193, etc. So `bit.rshift(option, 11)` captures all the options nicely in an bit var where the bits are, in order:

none
regen
refresh
duration

i.e. if you choose regen and refresh you get a return of `6` from the `rshift`

Knowing this info, I was able to simplify the `bonusCost` calculation as well as the buff selection in `sigil.lua`

Note: This updated code supports a 4th option for decreased exp loss (assuming the pattern continues) but I don't know how to make the NPC display that option... and also the mod for that doesn't appear to be coded anyway.

Choose options from a sigil NPC in any of the past nations and see that the power of your sigil buff ranges from 1 (no bonus effects) to 14 (all 3 main effects)

### DIFF
--- a/scripts/effects/sigil.lua
+++ b/scripts/effects/sigil.lua
@@ -3,52 +3,25 @@
 -----------------------------------
 local effectObject = {}
 
--- TODO: Find a way to simplify this.  Is this a bitfield?
 effectObject.onEffectGain = function(target, effect)
     local power = effect:getPower() -- Tracks which bonus effects are in use.
 
-    if
-        power == 1 or
-        power == 3 or
-        power == 5 or
-        power == 7 or
-        power == 9 or
-        power == 11 or
-        power == 13 or
-        power == 15
-    then
+    if utils.mask.getBit(power, 1) then
         local percentage = 70 -- TODO: This should be based off of controlled areas in Campaign
         target:addLatent(xi.latent.SIGIL_REGEN_BONUS, percentage, xi.mod.REGEN, 1)
     end
 
-    if
-        power == 2 or
-        power == 3 or
-        power == 6 or
-        power == 7 or
-        power == 10 or
-        power == 11 or
-        power >= 14
-    then
+    if utils.mask.getBit(power, 2) then
         local percentage = 60 -- TODO: This should be based off of controlled areas in Campaign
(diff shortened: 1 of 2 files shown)


TeoTwawki:
> Note: This updated code supports a 4th option for decreased exp loss (assuming the pattern continues) but I don't know how to make the NPC display that option... and also the mod for that doesn't appear to be coded anyway.

https://ffxiclopedia.fandom.com/wiki/Talk:Sigil#Reduced_EXP_Loss?

its on a 2nd page and if I recall correctly was not following the pattern the last 3 did - there's probably an unused bit or 2 between the 3rd and 4th option, or maybe getting the 2nd page to exist is separate from it.

Xaver-DaRed (approved):

scripts/effects/sigil.lua
```diff
@@ -60,46 +33,21 @@ effectObject.onEffectLose = function(target, effect)
...
-        -- target:delMod(xi.mod.EXPLOSS_REDUCTION), ???)
-        -- exp loss reduction not implemented.
-    elseif effect:getPower() >= 12 then
+    if utils.mask.getBit(power, 3) then
         target:delMod(xi.mod.FOOD_DURATION, 100)
+    end
+
+    if utils.mask.getBit(power, 4) then
         -- target:delMod(xi.mod.EXPLOSS_REDUCTION), ???)
         -- exp loss reduction not implemented.
```
WinterSolstice8, line 52: Do we really not have exp loss reduction? Isn't there a moghancement that does this?

WinterSolstice8 (approved):

scripts/effects/sigil.lua
cocosolos: EXPERIENCE_RETAINED, I think.

TeoTwawki: Comment in code was written many many years ago and when mod was created, this didn't get updated. So thats another easy PR for somebody.

PR #4938, opened 2024-01-02
[HELM] Adds all item IDs, cleans up item results

loxleygames:
**_I affirm:_**

* Filled out all HELM item IDs and added missing ones to `enum/item.lua`
* Moved HELM result events to `xi.helm.result`, separating it from the main logic and allowing easy extension and overrides
* Moved Egg-Hunt event logic into the event lua, to be called from `xi.helm.result`
* Renamed `item` to `itemID` for better clarity
* Added tool type item IDs from enum

* Tested multiple different areas, was able to receive expected results
* Tested changes with Egg-Hunt event, received daily eggs correctly
* Tested function with override and super, expected regular behavior was observed
* Tested results with full inventory
* Re-tested with latest changes


scripts/events/egg_hunt_egg-stravaganza.lua
```diff
@@ -780,6 +780,23 @@ for bonusWord, rewardItem in pairs(settings.BONUS_WORDS) do
...
 end
 
+-----------------------------------
+-- HELM event handler
+-----------------------------------
+xi.events.eggHunt.helmResult = function(player, itemID)
+    if
+        itemID ~= 0 and
+        xi.events.eggHunt.enabledCheck() and
+        player:getCharVar(settings.VAR.DAILY_HELM) < VanadielUniqueDay()
```
TeoTwawki, line 789: on retail you can re-zone to get you once per day egg multiple times, so a localVar would be more appropriate

_I just gave away the secret to my egg factory, retail players. you're welcome._

Also its king of annoying when you do not want eggs, because that 1st swing of you pick/hatchet/sickle will be an egg and nothing else.

TeoTwawki: come to think of it I think I mentioned this at some point in the past but then let it slide. Was this just moved from bigger logic block elsewhere? Github doesn't like large file diffs and I haven't dug all the way in yet.

loxleygames: This seems a bit beyond the scope because it's not supposed to be a change in functionality. I think most servers would want this turned off, as it kind of ruins the event, in which case I'd need to make a setting.

scripts/globals/helm.lua
```diff
@@ -1496,6 +1496,34 @@ xi.helm.initZone = function(zone, helmType)
...
+        npcUtil.giveKeyItem(player, xi.ki.RAINBOW_BERRY)
+    end
+
+    -- Missiom: AMK04
+    if xi.settings.main.ENABLE_AMK == 1 then
+        xi.amk.helpers.helmTrade(player, helmType, broke)
+    end
+
+    -- Records of Eminence
+    player:triggerRoeEvent(xi.roeTrigger.HELM_SUCCESS, { ['skillType'] = helmType })
```
Xaver-DaRed: Wouldnt this change make it so the RoE is ALWAYS triggered no matter if you failed or not?

scripts/events/egg_hunt_egg-stravaganza.lua
TeoTwawki: a "yes" would have sufficed bro.

I think that was my own reasoning too - getting eggs when you want darksteel or adaman or w/e sucks.

scripts/globals/helm.lua
loxleygames: Good catch, I've fixed.

Xaver-DaRed (approved):

TeoTwawki (approved):

zach2good (approved):

PR #4940, opened 2024-01-02
Correct relic magian dmg augment being missing

### LINKED ISSUE #4853: 🐛 Magian DMG augments provide no benefit
**_I affirm:_**

Currently, magian augments give augment ID 740, which has no mod mapping in the DB:

https://github.com/LandSandBoat/server/blob/0ced4398f975c776801619f82a0fd81fc22ea54d/sql/augments.sql#L899

Looks like it is probably due to lack of information on how exactly the augments are supposed to behave. I believe that all the calls to read the mod `DMG_RATING` (mod id 287) do so via the _weapon_ and not the _player_, so I'd imagine this should simply be adding mod 287, value 1 to augment 740 and 746 for melee and ranged weapon augments, respectively.

I would have made a PR but I don't have retail captures to back up what I _think_ the answer should be, so putting here for direction

## Branch affected by issue 

`base`

## Steps to reproduce

Give a weapon with a dmg augment (`!additem 18270 1 740 1`), then print `player:getWeaponDmg()` to see that it isn't getting any additional weapon damage

MowFord:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/4853 .

The mod `287: DMG_RATING` applies on a weapon/ammo level, while the `376: RANGED_DMG_RATING` is applied at the end from the battle entity, so we should use the former for both main and ranged relic augments.

Can be confirmed by giving a relic weapon with the +DMG augment and using `!exec player:getRangedDmg()` and `!exec player:getWeaponDmg()`

### DIFF
--- a/sql/augments.sql
+++ b/sql/augments.sql
@@ -896,13 +896,13 @@ INSERT INTO `augments` VALUES (738,0,0,0,0,0);
 INSERT INTO `augments` VALUES (739,0,0,0,0,0);
 -- End unused block
 
-INSERT INTO `augments` VALUES (740,0,0,0,0,0); -- DMG:+1
+INSERT INTO `augments` VALUES (740,0,287,1,0,0); -- DMG:+1
 insert into `augments` values (741,0,0,0,0,0); -- Dmg:+33   Needs to work in either hand,whichever one the weapon is equipped in.
 insert into `augments` values (742,0,0,0,0,0); -- Dmg:+65   Ranged weapons use diff AugID (starts at 746) and diff ModID.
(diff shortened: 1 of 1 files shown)


TeoTwawki (approved):
As noted in the issue and somewhat in much copypasta'd file comments there were some..oddities. however nothing stops updating these later (no harm comes to extant items already aug'd) and the player wouldn't know the diff. Better to do this than have them totally useless for another 4-6 yrs.

Xaver-DaRed (approved):

Issue #2772, opened 2022-09-22
🐛 Augment ID's do not reflect accurately


N3ckB3ard:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## OS / platform the server is running (if known)

 Any

## Branch affected by issue 

`base`

## Steps to reproduce

 Adding correct Augment ID's based off client information has not worked for some time. I have looked through the Augment information and can confirm this has been a long living issue. Within the .DAT files we are able to see the ID that corresponds with the augment but using the ID's within our SQL will reflect differently. I am hoping by having this issue open here we can see what information we can collect to implement this to work in a reflective manner. I know in passing this has been looked into but I would like to have a open issue for referencing Augments and for someone looking for a major project.

Client information:

https://github.com/Windower/Resources/blob/master/resources_data/augments.lua
https://www.bg-wiki.com/ffxi/Augment_List

Notes:
* Most "Empty" augments will display 1 over the input augment ID (EX: 346->347) and some will display the same input(EX: 302->302)
* Augment 912 will increase the augment ID by the value or "Power" set(EX: Power = 8-"Occ. atk. twice", 11-"Occ. atk. 2-3 times", 12-"Occ. atk. 2-4 times")
* Augment ID = 897 - Enhances Souleater Effect +d% in reference to the client should reflect the ID of 513
* At some point our Augment ID's will shift back into prior numbers and this prevents us from using all available Augments within the client
* Adding Augments to items currently with the Value or "Power" set to 0 will reflect 1(EX: !additem itemId Qty 1(HP+v) 0(v=4) --HP+5)

TeoTwawki:
was looking at this earlier today. there are some bytes in the exData of items we haven't implemented that determine the kind of augment data that shoudl be loaded. we only have 1 type implement whioch cuts off the list of evailable IDs far lower than retail.

1st byte is flags, 2nd is  sub flags that tell it how to handle the aug. We're currently stopping at 2047 (0x7FF) because the original augments they released were bitpacked, but the list really should reach a 16 bit integer and go all the way to 0xFFxx

[...]

PR #4947, opened 2024-01-03
[SQL] Adjust multiplicative augments to behave properly

MowFord:
**_I affirm:_**

Augments have a particular mod they enact (or multiple mods). Each mod is given a power and a multiplier. I think whoever implemented augments intended for power to always be 1 or -1 when multiplier was nonzero.

Finding augments via this query:
`SELECT * FROM augments WHERE multiplier <> 0 AND ABS(VALUE) <> 1`

we can see various mismatches between what the client shows and what mods get applied:

image

image

image

This PR attempts to correct them all in one fell swoop. I am aware of an over-arching issue with augments not being packet-correct and thus sometimes giving incorrect display when augmentid/power aren't within certain parameters (i.e. [this issue](https://github.com/LandSandBoat/server/issues/2772)), but I don't believe that is related to this particular PR's resolution

Basically, the intent of this PR is to find all augments that the client interprets as counting by multiples other than 1 and fixing the SQL side to make the CPP side work in the expected way.

after changes:
image
etc

specifically, you test by giving yourself an item with the particular augment/power and see that the game client reflects the underlying mods being applied (I like to use bronze dagger myself):
- `!additem bronze_dagger 1 78 0`
  - image
- `!additem bronze_dagger 1 78 2`
  - image
- `!additem bronze_dagger 1 83 1`
  - image
etc, etc

### DIFF
--- a/sql/augments.sql
+++ b/sql/augments.sql
@@ -100,20 +100,20 @@ INSERT INTO `augments` VALUES (69,0,26,1,0,0); -- Rng.Acc. +1 Rng.Atk. +1
 INSERT INTO `augments` VALUES (69,0,24,1,0,0); -- Cont.
 INSERT INTO `augments` VALUES (70,0,30,33,0,0); -- Mag. Acc. +33 Mag.Atk.Bns +33
 INSERT INTO `augments` VALUES (70,0,28,33,0,0); -- Cont.
-INSERT INTO `augments` VALUES (71,0,160,-100,0,0); -- Damage Taken -1%
+INSERT INTO `augments` VALUES (71,100,160,-1,0,0); -- Damage Taken -1%
 INSERT INTO `augments` VALUES (72,0,0,1,0,0); -- Exp. Point +1%
 INSERT INTO `augments` VALUES (73,0,0,33,0,0); -- Exp. Point +33%
 INSERT INTO `augments` VALUES (74,0,915,1,0,0); -- Cap. Point +1%
 INSERT INTO `augments` VALUES (75,0,915,33,0,0); -- Cap. Point +33%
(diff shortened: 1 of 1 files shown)


MowFord:
Went through and confirmed each augment id, but thought i'd show the pet ones in addition to the above screenshots since they're somewhat unique:
image
image

sql/augments.sql
```diff
@@ -1427,8 +1427,8 @@ INSERT INTO `augments` VALUES (1151,0,0,0,0,0);
 INSERT INTO `augments` VALUES (1152,0,1,10,0,0); -- DEF +10
 INSERT INTO `augments` VALUES (1153,0,68,3,0,0); -- Evasion +3
 INSERT INTO `augments` VALUES (1154,0,31,3,0,0); -- Mag. Evasion +3
-INSERT INTO `augments` VALUES (1155,100,161,-2,0,0); -- Physical Damage Taken -2%
-INSERT INTO `augments` VALUES (1156,100,163,-2,0,0); -- Magic Damage Taken -2%
+INSERT INTO `augments` VALUES (1155,200,161,-1,0,0); -- Physical Damage Taken -2%
+INSERT INTO `augments` VALUES (1156,200,163,-1,0,0); -- Magic Damage Taken -2%
```
TeoTwawki, line 1431: can you take a look at aug ID 71 also? thanks.

MowFord: image

🔨

MowFord: found those 3 more augments and fixed to match the rest of what this PR is doing. Essentially some augments do have a base value then increment by 1, but _every_ mod that we (lsb) chooses to have some other base on the values has to use mutlipliers even if they increment "by one" as in these extra 3.

TL;DR - the 3 above had this behavior:
image

TeoTwawki: thanks.

TeoTwawki (approved):

Xaver-DaRed (approved):

zach2good (approved):

PR #4959, opened 2024-01-05
[LUA] Violent Flourish combat log fix

MowFord:
**_I affirm:_**

Violent Flourish always returns zero dmg in the combat log due to the helper function's return not being forwarded in the `onUseAbility` function. This fixes it

Use Violent Flourish, see that it now shows nonzero dmg


Xaver-DaRed (approved):

zach2good (approved):

zach2good:
Over time the context of the PR title is lost and all that's left in the git log is your commit messages. What you've done should be clear (including context) from just that message. I squashed your message into `return dmg for proper combat log for Violent Flourish` on the way in. Typically, I name my commits like their eventual PR titles will look.

PR #4960, opened 2024-01-05
[LUA] WS should get +100 acc on first hit

MowFord:
**_I affirm:_**

Multiple sources indicate this:
1. [bg-wiki](https://www.bg-wiki.com/ffxi/Category:Weapon_Skills)
    - > The first swing of any physical Weapon Skill receives a substantial (~+100) [accuracy](https://www.bg-wiki.com/ffxi/Accuracy) bonus, while additional swings receive no bonus.
2. [SE dev forum](https://forum.square-enix.com/ffxi/threads/49839-%E5%A4%9A%E6%AE%B5%E3%82%A6%E3%82%A7%E3%83%9D%E3%83%B3%E3%82%B9%E3%82%AD%E3%83%AB%E3%81%AE%E5%91%BD%E4%B8%AD%E3%81%AB%E3%81%A4%E3%81%84%E3%81%A6?p=571812#post571812)
    - > Additionally, physical weapon skills have a +100 hit correction added only to the first hit.
3. as well as just player knowledge from the time _before_ the +100 was added, where the meta was to use things like red lotus blade because it was guaranteed hit.

The change is implemented as a separate calculation of the first hit with an additional bonus acc of 100. That value is then used just before _only the first hit_ of a non-jump WS.

Confirmed via printing of the `calcParams.hitRate` value.

Also found that hitRate was never being capped at .95 for WS that do not vary acc with TP.

Run various ws from various categories (ranged, physical, elemental) and see that they behave the same.

add 
```
local function getSingleHitDamage(attacker, target, dmg, wsParams, calcParams)
    print(string.format("hit acc: %f", calcParams.hitRate))
``` 
to 100% confirm the hitrate is being modified only for the first hit of even dual-wield/h2h ws

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -519,6 +519,9 @@ xi.weaponskills.calculateRawWSDmg = function(attacker, target, wsID, tp, action,
     -- Recalculate accuracy if it varies with TP, applied to all hits
     if wsParams.acc100 ~= 0 then
         calcParams.hitRate = accVariesWithTP(calcParams.hitRate, calcParams.accStat, tp, wsParams.acc100, wsParams.acc200, wsParams.acc300)
+    else
+        -- clamp hitRate now if accuracy doesn't vary with TP
+        calcParams.hitRate = utils.clamp(calcParams.hitRate, 0.2, 0.95)
     end
 
     -- Calculate alpha, WSC, and our modifiers for our base per-hit damage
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
Rebase pls

MowFord:
> Rebase pls

Word. I rebased this and my other two that touch weaponskills.lua

WinterSolstice8 (approved):
Works for me

Xaver-DaRed (approved):

PR #4966, opened 2024-01-06
[Fishing] Remove era's 100 skill cap

Xaver-DaRed:
**_I affirm:_**

Removes a cap placed in fishing skill for no reason.

Set yourself at fishing skill 200 and see how you actually are able to fish anything with ease.

### DIFF
--- a/src/map/utils/fishingutils.cpp
+++ b/src/map/utils/fishingutils.cpp
@@ -1076,9 +1076,7 @@ namespace fishingutils
 
     uint8 GetFishingSkill(CCharEntity* PChar)
     {
-        uint8 rawSkill = (uint8)std::min(100, (int)std::floor(PChar->RealSkills.skill[SKILL_FISHING] / 10));
-
-        return rawSkill + PChar->getMod(Mod::FISH);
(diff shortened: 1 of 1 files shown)


src/map/utils/fishingutils.cpp
```diff
@@ -1076,9 +1076,9 @@ namespace fishingutils
 
     uint8 GetFishingSkill(CCharEntity* PChar)
     {
-        uint8 rawSkill = (uint8)std::min(100, (int)std::floor(PChar->RealSkills.skill[SKILL_FISHING] / 10));
+        uint8 rawSkill = (uint8)std::floor(PChar->RealSkills.skill[SKILL_FISHING] / 10) + PChar->getMod(Mod::FISH);
 
-        return rawSkill + PChar->getMod(Mod::FISH);
+        return rawSkill;
```
zach2good: just return the result:
```cpp
return static_cast<uint8>(std::floor(PChar->RealSkills.skill[SKILL_FISHING] / 10) + PChar->getMod(Mod::FISH));
```

TeoTwawki (approved):

PR #4866, opened 2023-12-15
Monk mob weapon damage adjustments to match PC

MowFord:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

Recreated my PR as I was doing cleanup and didn't realize my older LSB PRs were linked to the Wings-XI repo

Please see https://github.com/LandSandBoat/server/pull/4674

TL;DR - monk mobs that hit twice due to having a h2h weapon should not have the same weapon damage as mobs that only hit once.

## Steps to test these changes

Monk mobs like Lycopodiums should have reduced damage, but monk mobs like Sea Monks in Arrapago Reef should be unchanged.

PR #4968, opened 2024-01-07
[LUA] Include weapon dmg for getstats command

MowFord:
**_I affirm:_**

Investigating (_FALSE_) claims that my #4866 PR wasn't applying to garrison monks, I found this addition to be helpful. Also lots of mobskills use mob's weapon damage so I think it'll prove useful in other situations

`!getstats` and bask in its glory

image


scripts/commands/getstats.lua
```diff
@@ -41,8 +41,10 @@ commandObj.onTrigger = function(player)
     player:printToPlayer(string.format('Food Accuracy%% bonus: %i ', target:getMod(xi.mod.FOOD_ACCP)), xi.msg.channel.SYSTEM_3)
     player:printToPlayer(string.format('Accuracy Base: %i ', target:getMod(xi.mod.ACC)), xi.msg.channel.SYSTEM_3)
     player:printToPlayer(string.format('Total Accuracy: %i ', target:getStat(xi.mod.ACC)), xi.msg.channel.SYSTEM_3)
+    player:printToPlayer(string.format('Main Weap Dmg: %i ', target:getWeaponDmg()), xi.msg.channel.SYSTEM_3)
```
TeoTwawki, line 44: I'd put mainhand wdmg and ranged wdmg together and up at the top eitehr right before or right after the base stats of str/dex/etc

But really, this is getting so large we should start breaking groups of stats into subcommands soon. That was my intent to do when I originally slapped together this series of prints, which I expect is going to get redone from `string.format()` into `fmt` somewhere down the road.

Approved regardless of placement, it'll likely get re-arranged later anyhow.

TeoTwawki (approved):

Xaver-DaRed (approved):

PR #4970, opened 2024-01-08
[Combat][Spells] Refactor critical hit rate. Add SA and TA to ble spells.

Xaver-DaRed:
**_I affirm:_**

- Fixes critical hit rate from dStat being added to all WSs no matter if they could actually crit or not.

And as Mowford said in his PR: #4698 
- PHYSICAL_POTENCY now gives acc and bonus to cratio
- This ports the SATA code from CEXI for blu magic.
- It also enables TA to function for blu magic skillchain damage

End result is (using cannonball as an example)

    a SA cannonball will auto crit
    if taChar is in position
        a TA cannonball will transfer enmity
        a TA cannonball that skillchains will transfer enmity of both spell and skillchain

- Finally, some adjustments were made to support params.bonusacc and params.critchance. If these params aren't passed both will be zero by default
- Also (sorry for the late discovery, but didn't want to make a new PR since this one is still open): Found that the base multiplier is being set to 1 unless Chain Affinity is in effect. This is wrong: https://www.bg-wiki.com/ffxi/Calculating_Blue_Magic_Damage. The base multiplier is in the table as ftp0 and is only applied to the first hit.

REVIEW BY COMMIT for your sanity.

- !stun and !immortal a mob
- position for SA and compare damage before and after cannonball
- position for TA and remove !stun after to see the mob target the player
- TA for skillchain damage is a bit trickier to test, but you can use !getenmity to confirm your enmity doesn't spike after using TA (open the skillchain with the second player)

### DIFF
--- a/scripts/globals/bluemagic.lua
+++ b/scripts/globals/bluemagic.lua
@@ -4,6 +4,7 @@
 -----------------------------------
 require('scripts/globals/magic')
 require('scripts/globals/mobskills')
+require('scripts/globals/combat/physical_utilities')
 -----------------------------------
 xi = xi or {}
 xi.spells = xi.spells or {}
@@ -127,8 +128,8 @@ local function calculatefSTR(dSTR)
 end
 
 -- Get hitrate
-local function calculateHitrate(attacker, target)
-    local acc     = attacker:getACC() + attacker:getMerit(xi.merit.PHYSICAL_POTENCY) + (attacker:getMainLvl() - target:getMainLvl()) * 4
+local function calculateHitrate(attacker, target, bonusacc)
+    local acc     = attacker:getACC() + attacker:getMerit(xi.merit.PHYSICAL_POTENCY) * 2 + (attacker:getMainLvl() - target:getMainLvl()) * 4 + bonusacc
     local eva     = target:getEVA()
     local hitrate = (75 + (acc - eva) / 2) / 100
 
@@ -171,7 +172,7 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
     end
 
     -- Multiplier, bonus WSC
-    local multiplier = 1
+    local multiplier = params.multiplier or 1
     local bonusWSC   = 0
 
     -- BLU AF3 bonus (triples the base WSC when it procs)
@@ -210,8 +211,15 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
         params.offcratiomod = caster:getStat(xi.mod.ATT)
     end
 
+    params.offcratiomod = params.offcratiomod * (caster:getMerit(xi.merit.PHYSICAL_POTENCY) + 100) / 100
+    params.bonusacc     = params.bonusacc == nil and 0 or params.bonusacc
+
+    -- params.critchance will only be non-nil if base critchance is passed from spell lua
+    local nativecrit  = xi.combat.physical.calculateSwingCriticalRate(caster, target, false, 0, 0, 0)
+    params.critchance = params.critchance == nil and 0 or utils.clamp(params.critchance / 100 + nativecrit, 0.05, 0.95)
+
     local cratio  = calculatecRatio(params.offcratiomod / target:getStat(xi.mod.DEF), caster:getMainLvl(), target:getMainLvl())
-    local hitrate = calculateHitrate(caster, target)
+    local hitrate = calculateHitrate(caster, target, params.bonusacc)
 
     -------------------------
     -- Perform the attacks --
@@ -220,15 +228,34 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
     local hitsdone   = 0
     local hitslanded = 0
     local finaldmg   = 0
+    local sneakIsApplicable = caster:hasStatusEffect(xi.effect.SNEAK_ATTACK) and
+                                spell:isAoE() == 0 and
+                                params.attackType ~= xi.attackType.RANGED and
+                                (caster:isBehind(target) or caster:hasStatusEffect(xi.effect.HIDE))
+    local trickAttackTarget = (caster:hasStatusEffect(xi.effect.TRICK_ATTACK) and
+                                spell:isAoE() == 0 and
+                                params.attackType ~= xi.attackType.RANGED) and
+                                caster:getTrickAttackChar(target) or nil
 
     while hitsdone < params.numhits do
         local chance = math.random()
-        if chance <= hitrate then -- it hit
+
+        if
+            sneakIsApplicable or
+            chance <= hitrate
+        then
             -- TODO: Check for shadow absorbs. Right now the whole spell will be absorbed by one shadow before it even gets here.
 
             -- Generate a random pDIF between min and max
             local pdif = math.random(cratio[1] * 1000, cratio[2] * 1000)
             pdif       = pdif / 1000
+
+            if
+                sneakIsApplicable or
+                math.random() < params.critchance
+            then
+                pdif = pdif + 1
+            end
 
             -- Add it to our final damage
             if hitsdone == 0 then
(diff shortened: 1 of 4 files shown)


scripts/globals/bluemagic.lua
```diff
@@ -210,45 +211,77 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
...
             -- Generate a random pDIF between min and max
             local pdif = math.random(cratio[1] * 1000, cratio[2] * 1000)
-            pdif = pdif / 1000
+            pdif       = pdif / 1000
+
+            if
+                sneakIsApplicable or
+                math.random() < params.critchance
+            then
+                pdif = pdif + 0.7
```
WinterSolstice8: Crits add 1.0 to pDIF 👀

Xaver-DaRed: idk i copied whatever mow had in there

Xaver-DaRed: I was more concerned with the critical hit rate thing

scripts/globals/combat/physical_utilities.lua
```diff
@@ -522,6 +522,108 @@ xi.combat.physical.calculateRangedPDIF = function(actor, target, weaponType, wsA
...
+-- Critical rate master function.
+xi.combat.physical.calculateSwingCriticalRate = function(actor, target, isWeaponskill, TP1000, TP2000, TP3000)
+    -- See reference at https://www.bg-wiki.com/ffxi/Critical_Hit_Rate
+    local finalCriticalRate     = 0
+    local baseCriticalRate      = 0.05
+    local statBonus             = xi.combat.physical.criticalRateFromStatDiff(actor, target)
+    local inninBonus            = xi.combat.physical.criticalRateFromInnin(actor, target)
+    local fencerBonus           = xi.combat.physical.criticalRateFromFencer(actor)
+    local buildingFlourishBonus = xi.combat.physical.criticalRateFromFlourish(actor)
+    local modifierBonus         = actor:getMod(xi.mod.CRITHITRATE) / 100
```
WinterSolstice8, line 611: cpp has `ENEMYCRITRATE` mod as well which has a comment of `ENEMYCRITRATE            = 166, // Raises chance enemy will crit` which would appear to be additive to this modifierBonus

WinterSolstice8 (approved):

scripts/globals/combat/physical_utilities.lua
```diff
@@ -522,6 +522,109 @@ xi.combat.physical.calculateRangedPDIF = function(actor, target, weaponType, wsA
...
+        (subEquip == nil or subEquip:getSkillType() == xi.skill.NONE or subEquip:isShield()) -- Only shields allowed in sub.
+    then
+        fencerBonus = actor:getMod(xi.mod.FENCER_CRITHITRATE) / 100
+    end
+
+    return fencerBonus
+end
+
+-- Critical rate from Building Flourish.
+-- TODO: Study case were if we can attach modifiers to the effect itself, both this and the effect may need refactoring.
```
MowFord, line 586: You wouldn't want to put the modifers on the effect, because then BF would give crit/attack/acc on every attack before you WS

scripts/globals/weaponskills.lua
```diff
@@ -547,32 +498,14 @@ xi.weaponskills.calculateRawWSDmg = function(attacker, target, wsID, tp, action,
     local ftp = xi.weaponskills.fTP(tp, wsParams.ftp100, wsParams.ftp200, wsParams.ftp300) + calcParams.bonusfTP
 
     -- Calculate critrates
-    local critrate = calculateDEXvsAGICritRate(attacker, target)
+    local criticalRate = 0
```
MowFord, line 501: btw setting it out here wasn't giving crit rate to non-crit ws because each swing checked `wsParams.canCrit` then math.random, but def cleaner to just do everything inside the block below

MowFord:
LGTM, thanks for doing my thing and more. Excited for the continued consolidation of physical things in lua

src/map/entities/charentity.cpp
```diff
@@ -1027,11 +1027,16 @@ bool CCharEntity::OnAttack(CAttackState& state, action_t& action)
...
     TracyZoneScoped;
-    CBattleEntity::OnCastFinished(state, action);
 
     auto* PSpell  = state.GetSpell();
     auto* PTarget = static_cast<CBattleEntity*>(state.GetTarget());
 
+    // not ideal, since Trick Attack character (taChar) is also calculated on the lua side for the base spell,
+    // but should work as OnCastFinished is what removes the trick attack buff, and everything should happen serially
+    // Also taChar will return immediately if the player doesn't have Trick Attack effect
+    CBattleEntity* taChar = battleutils::getAvailableTrickAttackChar(this, PTarget);
```
zach2good: If memory serves, `getAvailableTrickAttackChar` is doing position space maths to find relevant entities to apply TA to. Is there anything we can do here to reduce how often this is called? Does it only apply to THF and BLU? We could limit this to those jobs, etc.

zach2good:
```cpp
CBattleEntity* taChar = nullptr

...
if (conditions)
{
    taChar = battleutils::getAvailableTrickAttackChar(this, PTarget);
}

...

uint16 skillChainDamage = battleutils::TakeSkillchainDamage(static_cast<CBattleEntity*>(this), PTarget, actionTarget.param, taChar);

TeoTwawki: Do we check for the status effect 1st? ~~thats the best we can do really, as any job can sub thf (only thf main gets the AGI bonus, but any job can TA)~~ _assassin trait required for crit_

TeoTwawki: actually thf main is only one to crit off ta without sa isn't it? sub would only get enmity transfer

edit: its the trait https://ffxiclopedia.fandom.com/wiki/Assassin

scripts/globals/weaponskills.lua
Xaver-DaRed: This is why weaponskills need a nuclear rewrite from scratch

src/map/entities/charentity.cpp
Xaver-DaRed: Yeah, I agree.
I'll add a check for the status effect, considering anyone can trick attack and it is needed, at the very least, for the enmity transfer.

MowFord: I'm the OP of that particular code (both the blu tie-in and the reworked TA) and there's a check for the trick_attack effect at the top of `battleutils::getAvailableTrickAttackChar`

The math is also greatly simplified in that function anyway so it shouldn't be computationally expensive to run on a few extra spells (i.e. pop TA then use a magic-based blu spell)

zach2good: In its current form, this runs on _every_ finished spell

TeoTwawki: more like every spell and has to call into the function even if you don't have the effect to check if you have the effect. uh oops?

Xaver-DaRed: EDIT: Only a few spells can actually TA

Concretely, Blue spells that act as physical Weaponskills

MowFord: > more like every spell and has to call into the function even if you don't have the effect to check if you have the effect. uh oops?

I'm definitely no cpp guru, but isn't the overhead of a function call essentially zero? We could add a check for the effect and wrap that call to the battleutil function, would that be more efficient?

image

zach2good: The cost of calling a function is negligible, but that doesn't excuse us from the premise of _doing extra work for absolutely no reason_. `getAvailableTrickAttackChar` is fine and it does exactly what it needs to, we don't need to leak the details of interacting with spells into it. That still belongs in `OnSpellCastFinished` (or w/e).

TeoTwawki: it just felt weird to me that we both jump into the function on everything and then check if we're supposed to be doing it after we're there

Xaver-DaRed: Added a check for
- spell type = blue magic
- spell doing damage
- caster having `Trick Attack`

TeoTwawki (approved):

PR #4974, opened 2024-01-09
[c++] Correct some data types in synthutils and fix HQ modifier just not working

Xaver-DaRed:
**_I affirm:_**

- HQ rate modifier was not propely casted to a float, making the bonus always 0 in the end.
Before:
imagen
After:
imagen

- Also changed success rate data type from float to uint8, couse thats how it was being used all across anyway.

Add prints to show values and craft away until your soul burns in flames and start wondering what you have ever done to deserve such a punishment.


TeoTwawki (approved):
those sneaky lil `f`'s strike again

PR #4975, opened 2024-01-09
don't send non-zero action values when spike did not proc

TeoTwawki:
**_I affirm:_**
- [ ] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [ ] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).


TeoTwawki:
god damn it github, I wasn't done typing

Xaver-DaRed:
You know, we could use those message changes at the very least

PR #4976, opened 2024-01-09
don't send non-zero action values when spike did not proc

### LINKED ISSUE #4937: 🐛 item spikes ignoring proc rate
**_I affirm:_**

n/a

## Branch affected by issue 
`base`

## Steps to reproduce
Items that give spike effects are procing regardless of the die roll for their chance. The logic allows it to get past some checks and then assumes a non proc wouldn't have made it into said die roll, so it winds up going past with the default values instead of sending back the 3 zeroes that are expected. Downstream people are making their own fixes for this **without reporting it.**

## Expected behavior
Proc rates obeyed.

TeoTwawki:
**_I affirm:_**

The entire program flow for handling spikes in general (not just equipment with spikes on them) is flat out wrong. It assumes we will be doing damage and an animation of some kind, before we even know if we've got an effect or gear that succeeded a "roll" then overwrites them with whatever values each conditional branch actually needed when its wrong, and at the end of none were entered finally overwrites with zero. this resulted in an oversight in spikes gear where a condition is always met so it never reaches those zero values at the end. The quick band-aid is to add another set of overwriting with zeroes there.

I first looked at correcting teh program flow, but its a lot of refactoring for something that is getting tirn down: we're moving how we handle damage over to lua in the future, and spikes were planned to move there even before that, so a full rewrite in cpp doesn't make sense when its going to lua later anyway.

This PR also includes some additional message enum work, and a bunch of format cleanup.

This closes #4937

Equip an item with spikes, let a bird in sky beat on you for an hr. see that sometimes they proc and sometimes they don't. Change the rate in the db on the item, restart server and retest and see that the rate MATTERS now.

Try genie weskit and see that curse message displays properly (its been missing its enum for years).

### DIFF
--- a/scripts/enum/msg.lua
+++ b/scripts/enum/msg.lua
@@ -259,8 +259,9 @@ xi.msg.basic =
     ADD_EFFECT_HP_HEAL     = 167, -- Additional effect: The <player> recovers <number> HP.
     ADD_EFFECT_DISPEL      = 168, -- Additional effect: <target>'s <Status Effect> effect disappears!
     ADD_EFFECT_WARP        = 169, -- Additional effect: Warp! (used by Halloween staves)
-    STATUS_SPIKES          = 374, -- Striking <Defender>'s armor causes <Attacker> to become <status effect>.
-    SPIKES_EFFECT_HEAL     = 383, -- <?>'s spikes restore <number> HP to the <?>.
+    SPIKES_EFFECT_RECOVER  = 373, -- <defender> recovers <number> hit points!
+    STATUS_SPIKES          = 374, -- Striking <defender>'s armor causes <attacker> to become <status effect>.
+    SPIKES_EFFECT_HEAL     = 383, -- <defenders>'s spikes restore <number> HP to the <attacker>.  (element absorbed)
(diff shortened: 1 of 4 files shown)


scripts/enum/msg.lua
```diff
@@ -405,3 +406,4 @@ xi.msg.system =
     TRUST_NO_ENMITY              = 300, -- You cannot use Trust magic while having gained enmity.
     TRUST_SOLO_OR_LEADER         = 301, -- You cannot use Trust magic unless you are solo or the party leader.
 }
+return xi.msg
```
Xaver-DaRed: I dont see returns in enum files? Is this return needed?

TeoTwawki: accidental

TeoTwawki: during testing log would whine about the cache, this was just to shut it the F up, and then I forgot it

Xaver-DaRed (approved):
CI hit you hard, but ignoring the CI style checks and the accidental return, everything looks ok

TeoTwawki:
> CI hit you hard, but ignoring the CI style checks and the accidental return, everything looks ok

misaligned stuff I didn't write in file I touched /sulk

TeoTwawki:
fixed Ci issues and the accidental return

PR #4979, opened 2024-01-10
[LUA] Add more geo data to global table and fix potency formula

MowFord:
**_I affirm:_**

This PR was sparked because potency wasn't scaling the way it's supposed to, but I think it useful to add these geo metadata to the global table for modules to manipulate them as needed.

The issue with the potency formula I think wasn't noticed at first due to the catchall clamp on the potency values, but zero combined skill was resolving way too low, meaning it took more combined skill than it should to bump past the minimum.

example:

[geo-regen](https://www.bg-wiki.com/ffxi/Geo-Regen) scales from 1 hp to 30hp (at 600 skill). This is +1 regen every 20 skill, meaning at 20 skill we should get +2. the old formula was not giving +2 until 40. 

Sidenote here: The divisors are somewhat redundant data as far as i can tell since all but regen scale linearly from 0 to 900 combined skill, but I'd rather not scope creep this PR. I plan on opening another PR where I audit all the potency min/max data. Spot checking, the data looks good, but the formula just has this oversight to make it scale appropriately

Use various indi-X or geo-X spells and observe the mod values at various combined skills.

To force a particular skill i found it easiest to have no handbell equipped, and use `!setmod geomancy_skill -50` (to reduce skill by 50 for instance). Once that mod is set you must change gear for the char skill to be recalculated (whatever is reflected in the magic skill menu is what the lua uses at time of cast)

### DIFF
--- a/scripts/globals/job_utils/geomancer.lua
+++ b/scripts/globals/job_utils/geomancer.lua
@@ -208,19 +208,22 @@ local function getEffectPotency(player, effect)
     local divisor            = xi.job_utils.geomancer.potencyData[effect].divisor
     local minPotency         = xi.job_utils.geomancer.potencyData[effect].minPotency
     local maxPotency         = xi.job_utils.geomancer.potencyData[effect].maxPotency
-    local potency            = utils.clamp(combinedSkillLevel / divisor, minPotency, maxPotency)
+    local geoModMultiplier   = xi.job_utils.geomancer.potencyData[effect].geoModMultiplier
+    -- simple linear regression between min and max potency, utlizing the pre-defined divisor
+    local potency            = utils.clamp(minPotency + math.floor(combinedSkillLevel / divisor), minPotency, maxPotency)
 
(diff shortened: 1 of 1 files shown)


MowFord:
Apologies for the clutter between two different commits, I've split them up into the two things and clarified in the OP that the goal of making global tables is to make modules easier

Spicyryan (approved):

Spicyryan:
Accident. I didn't notice when I opened the link that it was not in our Git. I apologize for hitting approve.  ^^;

zach2good:
It doesn't matter if a non-admin user submits a review - it doesn't count towards our approving reviews count. We largely ignore when people do this.

TeoTwawki:
more specifically the approval. ppl pointing out something they see wrong is encouraged.

Xaver-DaRed:
As soon as time allows, I'll get some data from retail and compare with both what we have and with the values this PR proposes.

PR #4981, opened 2024-01-11
[Core] Adds tandem blow/strike traits to sql and mods to cpp/lua

MowFord:
**_I affirm:_**

Implements the core plumbing for bst job traits [Tandem Strike](https://www.bg-wiki.com/ffxi/Tandem_Strike) and [Tandem Blow](https://www.bg-wiki.com/ffxi/Tandem_Blow)

I have a few ideas on how to implement the actual bonuses, but I think it better to put this in first to avoid merge conflicts

Test that at the appropriate job levels you get the appropriate traits and mods

### DIFF
--- a/documentation/player_traits.txt
+++ b/documentation/player_traits.txt
@@ -1,0 +1,140 @@
+Offset by 1536 from ROM\181\72.dat
+datIndex	traitID	namestring
+1537	1	Accuracy Bonus
+1538	2	Evasion Bonus
+1539	3	Attack Bonus
+1540	4	Defense Bonus
+1541	5	Magic Atk. Bonus
+1542	6	Magic Def. Bonus
+1543	7	Max HP Boost
+1544	8	Max MP Boost
+1545	9	Auto Regen
+1546	10	Auto Refresh
+1547	11	Rapid Shot
+1548	12	Fast Cast
+1549	13	Conserve MP
+1550	14	Store TP
+1551	15	Double Attack
+1552	16	Triple Attack
+1553	17	Counter
+1554	18	Dual Wield
+1555	19	Treasure Hunter
+1556	20	Gilfinder
+1557	21	Alertness
+1558	22	Stealth
+1559	23	Martial Arts
+1560	24	Clear Mind
+1561	25	Shield Mastery
+1562	26	.
+1563	27	.
+1564	28	.
+1565	29	.
+1566	30	.
+1567	31	.
+1568	32	Beast Killer
+1569	33	Plantoid Killer
+1570	34	Vermin Killer
+1571	35	Lizard Killer
+1572	36	Bird Killer
+1573	37	Amorph Killer
+1574	38	Aquan Killer
+1575	39	Undead Killer
+1576	40	.
+1577	41	Arcana Killer
+1578	42	Demon Killer
+1579	43	Dragon Killer
+1580	44	.
+1581	45	.
+1582	46	.
+1583	47	.
+1584	48	Resist Sleep
+1585	49	Resist Poison
+1586	50	Resist Paralyze
+1587	51	Resist Blind
+1588	52	Resist Silence
+1589	53	Resist Petrify
+1590	54	Resist Virus
+1591	55	Resist Curse
+1592	56	Resist Stun
+1593	57	Resist Bind
+1594	58	Resist Gravity
+1595	59	Resist Slow
+1596	60	Resist Charm
+1597	61	.
+1598	62	.
+1599	63	Resist Amnesia
+1600	64	Treasure Hunter II
+1601	65	Treas. Hunter III
+1602	66	Kick Attacks
+1603	67	Subtle Blow
+1604	68	Assassin
+1605	69	Divine Veil
+1606	70	Zanshin
+1607	71	Savagery
+1608	72	Aggressive Aim
+1609	73	Invigorate
+1610	74	Penance
+1611	75	Aura Steal
+1612	76	Ambush
+1613	77	Iron Will
+1614	78	Guardian
+1615	79	Muted Soul
+1616	80	Desperate Blows
+1617	81	Beast Affinity
+1618	82	Beast Healer
+1619	83	Snapshot
+1620	84	Recycle
+1621	85	Ikishoten
+1622	86	Overwhelm
+1623	87	Ninja Tool Expert.
+1624	88	Empathy
+1625	89	Strafe
+1626	90	Enchainment
+1627	91	Assimilation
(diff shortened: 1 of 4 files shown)


scripts/enum/mod.lua
```diff
@@ -410,6 +410,10 @@ xi.mod =
     DARK_MAGIC_DURATION             = 1072, -- Increases Dark Magic spell durations by percentage (e.g. mod value 10 = +10% duration)
     ENHANCES_DARK_SEAL              = 1073, -- Enhances "Dark Seal" effect (Increases Dark Magic spell durations by 10% per Dark Seal merit while Dark Seal active)
 
+    -- Beastmaster
+    TANDEM_STRIKE                   = 271,  -- Grants a bonus to your and your pet's accuracy and magic accuracy when you and your pet are attacking the same target.
```
Xaver-DaRed: Considering this is a modifier and not the trait itself, can we name this `TANDEM_STRIKE_BONUS`?

scripts/enum/mod.lua
```diff
@@ -410,6 +410,10 @@ xi.mod =
...
+    TANDEM_BLOW                     = 272,  -- Reduces amount of TP gained by enemies when striking them if you and your pet are attacking the same target.
```
Xaver-DaRed: Following that logic, `TANDEM_BLOW_BONUS` would apply here.

And of course, same with the ones in modifier.h

scripts/enum/mod.lua
TeoTwawki: `TANDEM_STRIKE_ACCURACY` fits better I think, bonus is usually a loaded word for things adding onto when we're naming mods

TeoTwawki: except ya know, spell better than teo..

TeoTwawki: likewise `TANDEM_BLOW_TP_REDUCTION` even if a bit wide.

Xaver-DaRed: Well, considering its a bonus to both ACC and MACC, i thought a more generic name would fit better. Not opposd to it though

Xaver-DaRed: ok then, lets go with `TANDEM_STRIKE_ACCURACY` and `TANDEM_BLOW_TP_REDUCTION`

MowFord: good point about it not being the real mod, i went with `_POWER` to indicate we're just defining how strong it is

TeoTwawki: `_POWER` sounds alright but might get changed somewhere down the road depending on future bs SE adds

Xaver-DaRed (approved):

scripts/enum/mod.lua
TeoTwawki: (if `_ACCMACC` or even `_ACC` happened I'd be gettin muh paddle)

documentation/player_traits.txt
```diff
@@ -0,0 +1,15 @@
...
+132	con anima
+133	con brio
+121	inspiration
+122	sleight of sword
+124	superior
+125	magic acc bonus
+126	magic eva bonus
+104	true shot
+135	max hp boost II
+136	shield barrier
```
TeoTwawki: Offset by 1536 from ROM\181\72.dat (so 1537 = 1)

[FFXI full trait list as of January 11th 2024.csv](https://github.com/LandSandBoat/server/files/13907756/FFXI.full.trait.list.as.of.January.11th.2024.csv)

TeoTwawki: I've been auto generating our enum for years btw. Just haven't kept up recently.

MowFord: omg so i didn't have to guess and check

MowFord: image

wow literally the last two traits eh

MowFord: added to the documentation txt in case people want to implement other missing traits

TeoTwawki (dismissed):

Xaver-DaRed:
So Teo, this trait adds a conditional value. Pretty much a "latent effect: same target"

Only when both master and pet are targeting the same entity. With separate modifiers, its easier to jus go to tp calculations, fetch actor and pet, check for same target and then just add the mod to tp gained/not gained

if we reused subtle blow II, the value would always be added... or removed... you know what i mean

TeoTwawki:
> So Teo, this trait adds a conditional value. Only when both master and pet are targeting the same mob. With separate modifiers, its easier to jus go to tp calculations, fetch actor and pet, check for same target and then just add the mod to tp gained/not gained
> 
> if we reused subtle blow II, the value would always be added... or removed... you know what i mean

I was thinking we'd be adding mods only after it becomes active in some core logic - it'd kind of hard to guess how we'd use this when its latterly a modifier we aren't even checking for (I am glad the PR didn't use the word "implements"). As I can't see how this is intended to play out, I've recused myself from review

Xaver-DaRed:
well, we can add a simple check to tp.lua in
`xi.combat.tp.calculateTPGainOnPhysicalDamage` and
`xi.combat.tp.calculateTPGainOnMagicalDamage`

```lua
local subtleBlowII = actor:getMod(xi.mod.SUBTLE_BLOW_II) -- This already exists
local pet          = actor:getPet()

if pet and actor:getTarget() == pet:getTarget() then
    subtleBlowII = subtleBlowII + actor:getMod(THE_NEW_MOD)
end
```
Everything else would be the same, I think

MowFord:
Correct this isn't the implementation PR, just getting the traits and power mods implemented. My original thought was combat tick listeners on the pet to give/remove mods but I'm undecided on an implementation method just yet

UmeboshiXI:
ASB has an implementation for [Tandem Strike](https://github.com/AirSkyBoat/AirSkyBoat/pull/3134/files) if you want to reference it for ideas on how to implement/not implement.

Edit: Tandem Blow to Tandem Strike

MowFord:
Not interested in continuing this to the second stage of implementation. If someone wants to take this over after it's merged great, otherwise might be best to just close the PR

Xaver-DaRed:
There is no need to close this PR nor any need to do any changes.
This PR implements certain missing job traits to the job trait list, adds modifiers for the job traits and adds documentation.
Thats it. Thats the scope of the PR. This PR is not intended to make this traits actuall work, and thats fne.

WinterSolstice8 (approved):

TeoTwawki:
others have stamped, so I am hittin da button. Not really a fan of placeholdering mods that are unused, but I've done it before myself to evade rebase hell.

PR #4984, opened 2024-01-12
[sql] Speed up Aria table inserts

cocosolos:
**_I affirm:_**

Sets Aria tables to not be "transactional." (In this context, transactional just means crash-safe.)

Crash-safeness doesn't really matter for these tables, and the switch to Aria _massively_ slowed down imports. Here's some numbers representing a whole DB import:

MyISAM (old):
myisam

Aria crash-safe (current) 👀:
aria transactional

Aria transactional=0 (this PR):
aria no transactional

Yeah that's over an hour to import.

I think this is a reasonable change, but open to any other ideas.

See:
https://mariadb.com/kb/en/aria-storage-engine/

Run a dbtool full update.

### DIFF
--- a/sql/abilities.sql
+++ b/sql/abilities.sql
@@ -25,7 +25,7 @@ CREATE TABLE `abilities` (
   `addType` smallint(2) NOT NULL DEFAULT '0',
   `content_tag` varchar(7) DEFAULT NULL,
   PRIMARY KEY (`abilityId`)
-) ENGINE=Aria DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci AVG_ROW_LENGTH=56;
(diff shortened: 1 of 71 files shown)


TeoTwawki:
I mistakenly thought the default settings for aria matched myisam, must be that my testbeds configuration isn't the same somewhere (I did full imports there and it didn't take long, about same as i was used to before the change).

TeoTwawki (approved):
These should all be read-only in our usage, so this should be fine

cocosolos:
> I mistakenly thought the default settings for aria matched myisam, must be that my testbeds configuration isn't the same somewhere (I did full imports there and it didn't take long, about same as i was used to before the change).

Yeah IDK my Debian server and Windows dev machine both experienced extremely slow imports. Don't think I've done any kind of tinkering with the default setups.

PR #4989, opened 2024-01-13
[LUA] Aymur listeners to enhance sic/ready with bonus tp

MowFord:
**_I affirm:_**

A bit of trickery due to the flow of abilities with sic and ready, but a works well regardless.

Enhances "Sic" and "Ready" effect gives +500 tp immediately and 500 tp is added to the ability. This is done via a listener on the player and a listener on the pet. 

Player Listener:
- gives 500 tp immediately to charmed pets upon usage of `SIC`
- adds listener to pet if not already there

Pet listener:
- just before using any skill, if it's a jug pet, give 1000 TP
- if it's a charmed pet, give the remaining 500 TP

On item unequip, remove both listeners

To test, i added prints _everywhere_.

Inside `aymur.lua`:
- i added a print at the beginning and end of the pet's listener to confirm how much TP is being added
- Added one at each level of onItemEquip to confirm listeners were being added at the appropriate times
- finally printed ability:getID and ability:getRecastID to confirm the recastID of ready abilities (note that giving tp in player's `ABILITY_USE` is too late, as the pet readies its ability before `ABILITY_USE` is called, but `ABILITY_USE` is the only place in the chain you can see that the player used an ability with `recastID == 102`. This ended up being ok as the ready abilites are for jug pets, and those are perfectly fine getting the full 1000 tp in the pet's listener, but noteworthy discovery I think)

outside of all the obvious places in `aymur.lua`, i added to `foot_kick.lua` to `print(skill:getTP())` just to confirm we were getting the tp for a jug pet actualized in the skill.

### DIFF
--- a/scripts/items/aymur.lua
+++ b/scripts/items/aymur.lua
@@ -1,0 +1,62 @@
+-----------------------------------
+-- Aymur
+-- BST mythic
+-- Enhances "Sic" and "Ready" effect
+-----------------------------------
+
+local itemObject = {}
+
+local addPetListener = function(pet)
+    pet:addListener('WEAPONSKILL_BEFORE_USE', 'AYMUR_ABILITY', function()
+        if pet:getMaster() then -- ensure we don't keep giving TP after charm wears off
+            if pet:getPetID() == 0 then -- charmed pets do not have petID
+                -- give remaining 500 tp just before ability
+                pet:addTP(500)
+            else
+                pet:addTP(1000)
+            end
+        else
+            pet:removeListener('AYMUR_ABILITY')
+        end
+    end)
+end
+
+itemObject.onItemEquip = function(target, item)
+    if item:getID() == 18979 then
+        return
+    end
+
+    local playerPet = target:getPet()
+    if playerPet then
+        addPetListener(playerPet)
+    end
+
+    target:addListener('ABILITY_USE', 'AYMUR_CREATE_PET', function(playerArg, targetArg, ability, action)
+        local pet = playerArg:getPet()
+        if pet then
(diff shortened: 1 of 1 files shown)


MowFord:
Sorry, everything should be in order now. First, i forgot to make itemObject local, then i was informed that the first aymur item doesn't have this effect. All others have identical effects though (+500 then +500 before pet skill usage)

Xaver-DaRed:
I dont know if this is the correct approach to implement the enhancement to `sic` and `ready`

For starters, there are multiple versions of the weapon, meaning, this approach would require having the same logic repeated in multiple files.

EDIT: Furthermore, there are other items with a similar (weaker) effect. Example:
- https://www.bg-wiki.com/ffxi/Nukumi_Manoplas
- https://www.bg-wiki.com/ffxi/Frn._Manoplas_%2B1

I think the correct approach here would be to implement a modifier. 1 would suffice for both Sic and Ready.
And then add some logic to sic and ready skills to account for it.

WinterSolstice8:
This feels like a hacky way to do it if you needed to module custom content. A mod that would be accounted for in the petskill/mobskill state to add TP on start is the way to go IMO

MowFord:
@Xaver-DaRed both of your examples are only giving TP_BONUS. If mobskills could be coded to respect that mod, then they could be simplified to just giving `item_pet_mods.sql` entries.

This line of weapons has a unique benefit to `sic` and `ready` in that they give 500 TP immediately, then another 500 as a TP_BONUS. Since that mod does nothing for mobskills, I added the second listener to ws prepare.

PR #4991, opened 2024-01-13
Restore missing Magicked Bones to Inner Horutoto Ruins

Thorinori:
**_I affirm:_**

This PR implements the Magicked Bones mob in Inner Horutoto Ruins. This was done with several requirements in mind to make them as retail-like as I could, and which will be detailed below.

1. There are 2 Magicked bones remaining in the zone
2. The Magicked Bones are lottery spawns off the 2 Goblins in I-9 of Map 1
3. The lottery spawn is only available from 20:00 to 4:00, and they depop once it is 4:00
4. The Magicked Bones drop Ceremonial Dagger and Bone Chips
5. The 2 Magicked Bones each use a specific model based on which placeholder it pops off of (Dagger wielding from the Weaver, Club wielding from the Thug)

Sources
[BG Wiki Ceremonial Dagger](https://www.bg-wiki.com/ffxi/Ceremonial_Dagger)
[BG Wiki Magicked Bones](https://www.bg-wiki.com/ffxi/Magicked_Bones)
[JP Wiki Ceremonial Dagger](https://wiki.ffo.jp/html/7918.html)
[JP Wiki Magicked Bones](https://wiki.ffo.jp/html/24263.html)
[FFXIDB Magicked Bones](http://ffxidb.com/zones/192/magicked-bones)

The JP wiki isn't perfectly translated by my browser, so tried to interpret it as best as I could.

1. Go to Inner Horutoto Ruins I-9 (Lily Tower entrance in Eastern Sarutabaruta to get to the right map)
2. Kill the Goblin Weaver and/or Goblin Thug in the hidden passage after 20:00
3. Repeat Step 2 until Magicked Bones pop (50% spawn rate from each placeholder, both can spawn at the same time since they are 2 distinct Magicked Bones)
4. Hopefully get loot (Ceremonial Dagger is ~5% as far as I could tell)


scripts/zones/Inner_Horutoto_Ruins/Zone.lua
```diff
@@ -129,6 +129,20 @@ zoneObject.onTriggerAreaLeave = function(player, triggerArea)
     end
 end
 
+zoneObject.onGameHour = function()
+    local totd = VanadielTOTD()
+    if totd == xi.time.NEW_DAY then
+        local bonesOne = GetMobByID(17563670)
+        local bonesTwo = GetMobByID(17563671)
```
MowFord: Might be better to add values to the zone's `mob` table so you can reference via ID.Mob.MAGICKED_BONES[1] and [2] or just _1 and _2

scripts/zones/Inner_Horutoto_Ruins/mobs/Magicked_Bones.lua
```diff
@@ -0,0 +1,27 @@
...
+-----------------------------------
+local ID = zones[xi.zone.INNER_HORUTOTO_RUINS]
+-----------------------------------
+local entity = {}
+
+local weaverId = 17563669
+local thugId = 17563668
+
+local daggerId = 17563671
+local clubId = 17563670
```
MowFord: These are redundant, they should all be references in the ID.mob table

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
```diff
@@ -2,10 +2,32 @@
...
+    if totd == xi.time.NIGHT or totd == xi.time.MIDNIGHT then
+        if math.random(1, 100) <= nmPopChance then
+            local mobId = mob:getID()
+            local nmId = ID.mob.MAGICKED_BONES_PH_DAGGER[mobId]
+            if nmId then
+                local skele = GetMobByID(nmId)
+                local respawnTime = mob:getRespawnTime()
+                DisallowRespawn(mobId, true)
+                mob:timer(respawnTime, function()
+                    skele:setModelId(1096) --Dagger Skeleton Model ID
```
MowFord: I believe you could change the name of the two magicked bones in the mob_spawn_points table, and set the modelId onMobInitialize

scripts/zones/Inner_Horutoto_Ruins/mobs/Magicked_Bones.lua
Thorinori: That is an option, I decided to opt for this rather than using the entries in IDs.lua that are already there since I didn't want to use a loop just to get a single value since the entries in ID.mobs are stored as tables and generically accessing them via pairs seemed like extra unnecessary overhead.

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
```diff
@@ -2,10 +2,32 @@
...
+-----------------------------------
 local entity = {}
+local nmPopChance = 50
 
 entity.onMobDeath = function(mob, player, optParams)
     xi.regime.checkRegime(player, mob, 647, 2, xi.regime.type.GROUNDS)
 end
 
+entity.onMobDespawn = function(mob)
+    local totd = VanadielTOTD()
```
TeoTwawki: treat them as if an NM, instead of reinventing the wheel.

```lua
-- 50% chance, pure lottery (meaning no cooldown)
xi.mob.phOnDespawn(mob, ID.mob.MAGICKED_BONES_PH_DAGGER, 50, 1)
```

scripts/zones/Inner_Horutoto_Ruins/Zone.lua
Thorinori: These are references to the mob itself rather than just the id, it didn't look like there were other actual mobs in ID

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
```diff
@@ -2,10 +2,32 @@
...
+                    skele:spawn()
```
MowFord: I assume they are supposed to spawn immediately upon killing the goblin? otherwise you could use the phMobDespawn function. It doesn't work of the PH has time-/weather-based spawn conditions, but should work fine in this direction

MowFord: Actually that function supports immediate spawn of the NM as well: 

`xi.mob.phOnDespawn(mob, ID.mob.MAGICKED_BONES_PH_DAGGER, 50, <delayTilNextNMWindow>, true)`

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
Thorinori: I was originally doing this, but it was causing problems. They were spawning on their own with the placeholders being alive still despite spawnType being set to 33. It also was causing an error when trying to call UpdateNMSpawnPoint

scripts/zones/Inner_Horutoto_Ruins/IDs.lua
```diff
@@ -45,6 +45,14 @@ zones[xi.zone.INNER_HORUTOTO_RUINS] =
...
             [17563800] = 17563801, -- -230.732 -0.025 -52.324
         },
+        MAGICKED_BONES_PH_DAGGER =
+        {
+            [17563669] = 17563671,
+        },
+        MAGICKED_BONES_PH_CLUB =
+        {
+            [17563668] = 17563670,
+        },
```
TeoTwawki:
```lua
MAGICKED_BONES_PH_CLUB = GetFirstID('Magicked_Bones')
MAGICKED_BONES_PH_DAGGER = MAGICKED_BONES_PH_CLUB +1
```

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
Thorinori: The timer was to make them spawn after a delay, otherwise yes they do spawn immediately. I wasn't sure which behavior was the desired and/or appropriate one, so I went with it spawning after a normal respawn length

scripts/zones/Inner_Horutoto_Ruins/IDs.lua
TeoTwawki: or use the other get by name function that returns a table

Thorinori: Doesn't this remove info though since the format seems to be [placeholder] = popped_mob_id?

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
TeoTwawki: you did something incorrectly for that to happen. 33 isn't a correct value either.

UpdateNMSpawnPoint error was informing you they have no positions recorded in that table.

the lotto system does also allow spawning immediately by the way.

scripts/zones/Inner_Horutoto_Ruins/IDs.lua
TeoTwawki: no it doesn't remove info. it gets the info in a more future proof manner that evades problems from ID shifts. you are copying an old pattern that's being slowly phased out.

TeoTwawki: you really only even need one of these when the others ID is just +1 from it

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
Thorinori: Yeah, I did see that it could spawn immediately normally as well, I just wasn't sure if they were supposed to be immediate or delayed. As for the value of 33, 33 should be 32 (lotto spawn) && 1 (night spawn), which is also the spawntype used for Citipati in the database, unless that is also a bug?

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
Thorinori: Potentially, I think the spawn point table is the only one where they have been distinct, otherwise both of them have a shared db entry in other tables so was trying to avoid doing the model changes via the db entries in particular

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
TeoTwawki: I'm pretty sure they are supposed to just replacing the goblin at next pop. that's how all shared spawns work all over the game world (there's several different orcs that share the spawn with Ashmaker GotBlutt's actual placeholder in Yuhgott Grotto, only 1 can be up at a time - not that we do not want to treat every case of this like an NM its just your bones here are special coz they have a drop).

scripts/zones/Inner_Horutoto_Ruins/IDs.lua
Thorinori: Ah I wasn't aware that was an old pattern, I was trying to make it fit the rest of the file. Does GetFirstID also limit it to the entries for the zone it is in, since there are Magicked bones in a bunch of other zones too?

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
TeoTwawki: sidenote: comments are `-- ` in this project not `--` you need the space.

scripts/zones/Inner_Horutoto_Ruins/Zone.lua
MowFord: yep, you gotta add them!

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
TeoTwawki: I'm gonna narrow this down for you:

> They were spawning on their own with the placeholders being alive

this specifically is user error. no matter what else. nothing else mattered if they were still spawning on their own. Solve that first. You likely gave them other incorrect values besides spawntype, its not the only element that makes them spawn/not spawn.

Thorinori: This is functionally what they are doing currently, when the goblin rolls the pop for the bones, it disables its own respawn which then gets re-enabled by the appropriate bones once the bones despawn either naturally or from being killed. This specific scenario is also the only situation I could find where a night only lotto had a placeholder that could also spawn during the day, which seemed to be causing weird behavior.

scripts/zones/Inner_Horutoto_Ruins/mobs/Magicked_Bones.lua
MowFord: yep, the first column is the lua file that gets read:

image

Scratch renaming it though just put an pair of if checks in the `onMobInitialize` function imo.

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
Thorinori: Ah alright, I can fix that easily enough. I guess I missed that when looking at the style guidelines.

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
MowFord: yep, the first column is the lua file that gets read:

image

Scratch renaming it though just put an pair of if checks in the `onMobInitialize` function imo.

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
MowFord: also while spawn_type seems like a flag, afaik everywhere in the code treats it like a value (i.e. you can't use two types of spanw type). Just use "scripted spawn" type and the xi.mob.phOnDespawn teo shared with a 1s NM repop window `xi.mob.phOnDespawn(mob, ID.mob.MAGICKED_BONES_PH_DAGGER, 50, 1, true)` to pop immediately after ph despawn (though i doubt that's how it's supposed to work)

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
MowFord: > so I went with it spawning after a normal respawn length

Ah, definitely use the ph helper function then. In the magicked_bones lua that you make to fix their model ids, you can put in `onMobSpawn` to despawn these particular ones immediately if they're outside of the valid ToTD

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
Thorinori: Ahhh so that is how it determines which file to tie it to. I was asking about that in Discord for like 2 days >:( But yeah that could definitely be done there then for sure. I wasn't 100% sure if OnInitialize happened each time or exactly how it functioned

scripts/zones/Inner_Horutoto_Ruins/IDs.lua
MowFord: image

pulls only from the zone

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
TeoTwawki: How this gets tackled depends on the answers to some question first, because we want to do what retail does. And we normally would NOT got messing with the mobs model in script. I would straight reject a pr that did that without first proving a need to.

DOES retail let BOTH of these bones pop as either model, (as in, can you get 2 of them up at the same time both wielding same weapon) or does it just do one of each and you have to hope it rolls the dagger model one to get it? I suspect the latter. The answer to this will form how I advise you and what further questions I ask.

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
Thorinori: > I'm gonna narrow this down for you:
> 
> > They were spawning on their own with the placeholders being alive
> 
> this specifically is user error. no matter what else. nothing else mattered if they were still spawning on their own. Solve that first. You likely gave them other incorrect values besides spawntype, its not the only element that makes them spawn/not spawn.

I tried a bunch of things to try to figure out what was causing it and I couldn't find it for the life of me, and couldn't find an answer for what else would cause them to spawn. spawnType was the only thing I could find or get info about, but if what MowFord says is true as well, the 33 value for Citipati is also wrong then, and that I had been given incorrect information about being able to AND them together to get multiple flags effects. I also was unaware of the phOnDespawn working right with it set to 128 for the scripted spawn, but in retrospect I am not sure why I thought that.

As for the other elements that can cause things to spawn though, could I get some extra info about that? Like I said, I hadn't been able to find anything beyond spawnType really handling that at all.

MowFord: @Thorinori spawntype_night gives sets mobs to spawnable when the zone changes to night

these are PH mobs that only spawn via a script, they need to explicitly only have spawntype_scripted

TeoTwawki:
unrelated to the review: 

no wonder I was never able to spawn that f\[beeep\]er after the updates - those bones never needed nighttime before the lv80+ mobs were added, and there are new lv80+ bones where they used to be that likewise do not require nighttime. I was legit convinced people were just repeating misinformation ever since the updates there because I've never once seen the dagger model version since. **_/sulk_**

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
Thorinori: I am not sure exactly how retail does it since I can't sub at the moment, the information that is based on is from the jp wiki (https://wiki.ffo.jp/html/7918.html https://wiki.ffo.jp/html/24263.html these pages specifically), where it stated (like I said the browser is kind of rough at translating it well so grain of salt) that each skeleton had a specific model based on which of the 2 goblins it popped from. In case your translation is different/better, here is exactly what it says for mine: "More specifically,[ Goblin Thug ](https://wiki.ffo.jp/html/10186.html)will be drawn as a one-handed club wielder, and Goblin Weaver a dagger will be drawn as wielding bone in the south hidden passageway entered from H-9"

As for  the drop, based on FFXIDB it only has a single entry for Magicked Bones for the zone, which has the dagger as part of the pool so I was operating under the assumption that they both could drop it. In general these 2 bones seem VERY unique among just about every other mob in the game for some reason.

scripts/zones/Inner_Horutoto_Ruins/IDs.lua
Thorinori: Ok cool

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
MowFord: According to the sources the mob that spawns from the weaver has one model, and the thug has another model. If we don't set the model in the lua we just need 2 different mob_pools with everything except the model the same

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Thug.lua
Thorinori: > @Thorinori spawntype_night gives sets mobs to spawnable when the zone changes to night
> 
> these are PH mobs that only spawn via a script, they need to explicitly only have spawntype_scripted

Currently they do only have 128 set, when I was initially trying with 33 they were misbehaving like I said. It could be user error apparently, but like I said I couldn't find anywhere else that would affect how they spawned, especially since it was inconsistent if they misbehaved or not. Like they would behave for a few pops, then suddenly just be there chilling with the PHs.

MowFord: that's part of why using phMobDespawn is so nice. It handles disabling respawn of the PH and enabling respawn of the NM, then flipping them when the NM dies

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
TeoTwawki: > I am not sure exactly how retail does it since I can't sub at the moment,

Not to be rude, but we're not guessing at things here or simply taking websites info at face value. I can possibly check for you myself over the weekend but if not there are folks in the captures discord who might be willing to spend some time on this. I'm a bit frazzled atm so I hope I am not coming off to much of a jerk. We need to know for sure when we implement or risk repeating the mistakes of previous iterations of the project. "I'll fix that later" and other jokes people told themselves. **I mean this in general, we have a lot of people trying to implement things based on wiki pages and guesswork, and it needs to stop.**

> As for the drop, based on FFXIDB it only has a single entry for Magicked Bones for the zone, which has the dagger as part of the pool so I was operating under the assumption that they both could drop it. In general these 2 bones seem VERY unique among just about every other mob in the game for some reason.

That site will combine them for having the same name while being in the same zone. Their method of data capture can't tell the 2 apart.

sql/mob_droplist.sql
```diff
@@ -26780,6 +26780,11 @@ INSERT INTO `mob_droplist` VALUES (3290,0,0,1000,16242,@VCOMMON); -- Dark Ixion:
 INSERT INTO `mob_droplist` VALUES (3290,0,0,1000,2654,@ALWAYS);   -- Dark Ixion: Horn
 INSERT INTO `mob_droplist` VALUES (3290,0,0,1000,2655,@ALWAYS);   -- Dark Ixion: Tail
 
+-- ZoneID: 192 - Magicked Bones
+INSERT INTO `mob_droplist` VALUES (3291,0,0,1000,880,@RARE); --Magicked Bones: Ceremonial Dagger
```
TeoTwawki: Chances are you need to edit spawn points and have 1 of the magicked bones use a different group pool and drop ID than the other. These mobs used to exist in larger quantities in the project, so one of the 2 drop lists _should_ already exists at some other number. (they were probably never separated, or share a list with another skeleton so you might possibly need _two_ for rate purposes..)

The alternate pool would use the opposite model ID.

Also that thing I mentioned about the comments, on some systems the SQL row will fail to import from it.

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
TeoTwawki: > According to the sources the mob that spawns from the weaver has one model, and the thug has another model. If we don't set the model in the lua we just need 2 different mob_pools with everything except the model the same

no you'd need 2 groups 2, in order to set 2 drop IDs. The dagger only drops from the one using the dagger model.

Thorinori: Not coming off as a jerk, I get it. I am generally trying to use as many sources as I can to corroborate stuff the best I can currently.  I will have to join the captures discord it sounds like! If you end up finding the info as well and let me know, I can get it fixed pretty quick in terms of being retail accurate. Besides that though, it seems like mostly just reverting back to the original ph code with some extras to account for having to use spawnType 128

MowFord: FYI jp wiki indicates that both types of magicked bones models drop the dagger:
image

I think i've said all I can in this PR now, though. If we still require verification from retail it sounds like it shouldn't be that hard to get the mobs to pop, and hopefully not too hard to get the dagger to drop from the club-wielder.

sql/mob_droplist.sql
Thorinori: Looks like you were right, the drop pool already existed it just wasn't assigned to the Magicked Bones at all (or anything for that matter). Well, at least I can remove the extra drop pool unless I need to have the exact same pool twice. As for the models, I was planning on switching to using onMobInitialize like Mowford suggested, assuming they both use the same dropid normally.

scripts/zones/Inner_Horutoto_Ruins/mobs/Goblin_Weaver.lua
TeoTwawki: 👍 I sometimes have a hard time gauging my own tone and also I _know_ I'm grumpy rn, so I was worried.

Spawn at night with PH being a mob that does NOT may add some extra trouble to this. Citipati's PH's are corses, who only pop at night. Multiple different values will just get treated as "scripted" and simply be overwritten (for a good migraine, take a look at what the functions toggle being allowed to spawn actually do to the spawnType field!) so combining other types with scripted doesn't work well. Past editors have made that mistake, as you found in Citipati.

I can possibly mod the NM function to handle night only as an optional thing. We probably don't want to check nighttime before rolling the chance to spawn because retail would check that at pop time - kill goblin just before night falls - as long as its "down" when night time hits, the spawn that occurs at night has the chance.

I've been considering adding a table param to it anyway and moving the force spawn option inside of that along with making the updatespawnpoint call optional.

Thorinori: Yeah, the goblins being any time and the bones being night only has definitely been a source of annoyance for sure. I will see what I can manage with the current feedback though, and at least temporarily continuing with the assumption that they both can drop it until I can find a better source than the jp wiki stating as much. 

On that note though to accelerate testing, do you have any suggestions for advancing time that doesn't completely break things? Using !time both desyncs the local clock, seems to desync more over time, and makes VanadielTOTD return 0 until a natural TOTD change happens which was breaking my TOTD checks. Otherwise to test spawning I was just disabling the totd requirement to make sure other stuff behaved properly, then enabling it when time was getting close to try to emulate proper spawn conditions.

TeoTwawki: > and at least temporarily continuing with the assumption that they both can drop it until I can find a better source than the jp wiki stating as much.

before SE added those lv80+ mobs I was the primary supplier of ceremonial daggers on my server. I know without any doubt that before they changed it, only the bones with rusty knife model coudl drop the dagger. The club model would not and the staff model would not (all 3 models existed in the zone in the past).

> On that note though to accelerate testing, do you have any suggestions for advancing time that doesn't completely break things? Using !time both desyncs the local clock, seems to desync more over time, and makes VanadielTOTD return 0 until a natural TOTD change happens which was breaking my TOTD checks.

Nope, its buggy while technically not being a bug, by its very nature. The game was never designed by SE to handle the time warpage so bad things happen unavoidably from us doing it. :( pity the in universe "time" wasn't instanced to each player/party so we coudl have timewarp spells/items, but that's a tall ask for an MMP especially back in the era FFXI was born in. Instancing wasn't even a thing yet! Some more modern games _literally every zone outside town is in an instance_ o.O

sql/mob_droplist.sql
TeoTwawki: > As for the models, I was planning on switching to using onMobInitialize like Mowford suggested, assuming they both use the same dropid normally.

pls see my comment warning against that.

_rolls up newspaper_ hey mowford, c'mere a min.

scripts/zones/Inner_Horutoto_Ruins/mobs/Magicked_Bones.lua
Xaver-DaRed: Mob IDs and any entity ID for that matter, shifts every every version update.

We can't. and refuse to, go file by file in the repository updating the IDs. It just isnt feasable. Hence why we have a zone file with mob IDs in it.

And now, we use a new pattern of `MOB_NAME = GetFirstID('Mob_Database_Name')` which allows us to just dont need to update them every version update,

This will only work from the zone.lua files.

Thorinori: Ah alright so that is the new pattern. This only works specifically in Zone.lua, or did you mean zone wide .lua files, aka IDs.lua AND Zone.lua? Since all of the IDs so far have been in IDs, that was where I would expect GetFirstID to work. Glad you sent this, I was just in the middle of making a new commit but I can tweak my ID changes again to match that pattern better if it works in IDs.lua (This specific example of extra ids in a different file is already gone though, just gotta make my current setup for ids fit the pattern better in IDs.lua)

Also, since the ids change, do the offsets change as well (for example, for this case the NMs are just ph_id + 2)

Xaver-DaRed: well, i meant in the IDs.lua files inside zone folders. Sorry, I wasnt clear.

Thorinori: Alright cool, so I will switch to using that then. I am curious how this pattern actually works though, since if I had to guess not every placeholder would be the first id of their mobtype in a zone, would they?

Thorinori:
Alright, did some revisions based on feedback here. Currently decided to leave the model changes as part of Magicked_Bones.lua until we have more info from retail. Currently still leaning towards them having the same dropid though based on the evidence provided throughout the comments here. Can still change that if necessary of course, but currently everything seems to be functioning properly with this setup.

scripts/zones/Inner_Horutoto_Ruins/IDs.lua
```diff
@@ -45,6 +45,8 @@ zones[xi.zone.INNER_HORUTOTO_RUINS] =
             [17563799] = 17563801, -- -237.426 0.5 -23.412
             [17563800] = 17563801, -- -230.732 -0.025 -52.324
         },
+        MAGICKED_BONES_PH_CLUB = GetFirstID("Magicked_Bones") - 2,
```
TeoTwawki: why do you have -2 here and then adding +2 over in the mob script?

And then minus one on the other. You don't even need both entries. The function will return the 1st magic bones ID it finds, with no +/- needed. You then "+1" to get the _second_ bones ID.

TeoTwawki: I posted the code for that last night.

scripts/zones/Inner_Horutoto_Ruins/mobs/Magicked_Bones.lua
```diff
@@ -0,0 +1,36 @@
...
+    if mob:getID() == ID.mob.MAGICKED_BONES_PH_DAGGER + 2 then
+        mob:setModelId(1096) -- Dagger Skeleton Model
+    else
+        mob:setModelId(573) -- Club Skeleton Model
+    end
+end
+
+entity.onMobSpawn = function(mob)
+    if despawn_check(mob) then
+        GetMobByID(mob:getID() - 2):spawn() -- Respawn the PH immediately if Bones were wrongly spawned outside of their window
```
TeoTwawki: that's not retail behavior. The PH spawns when its time to spawn is up, if the NM cannot/will not. Whether that is at daytime changeover or 2 and a half min after, it doesn't care.

TeoTwawki: if the bones is despawned by daytime, what you do instead if disable bones repop and enable goblin repop, both using the `DisallowRespawn()` function. You don't force pop the goblin.

scripts/zones/Inner_Horutoto_Ruins/mobs/Magicked_Bones.lua
```diff
@@ -0,0 +1,28 @@
...
+local function despawn_check(mob)
+    local totd = VanadielTOTD()
+    if totd ~= xi.time.NIGHT and totd ~= xi.time.MIDNIGHT then
+        DespawnMob(mob:getID())
+    end
+end
+
+entity.onMobInitialize = function(mob)
+    if mob:getID() == ID.mob.MAGICKED_BONES_PH_DAGGER + 2 then
+        mob:setModelId(1096) -- Dagger Skeleton Model
```
TeoTwawki: Absolutely not.

Also, I've killed them enough today that I am certain the same ID always gets the same model. just use 2 pool IDs - one for each model.

TeoTwawki (changes requested):

scripts/zones/Inner_Horutoto_Ruins/mobs/Magicked_Bones.lua
Thorinori: Will look into that then, not as versed in SQL so gotta figure out how to do that properly.

TeoTwawki:
Took me all day but confirmed these jerks can share the drop ID that already exists in the database.
image

Also confirmed:
- The same ID always has the same model, the model on the spawn is never changed

TeoTwawki (approved):
I will handle the night-only aspect of the bones soon. everything else here is ready.

Xaver-DaRed (approved):

PR #4992, opened 2024-01-13
[Magic] Raise Blind II variation range

Xaver-DaRed:
**_I affirm:_**

As per JP wiki: https://wiki-ffo-jp.translate.goog/html/3449.html?_x_tr_sl=ja&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=sc
- `Since [August 6, 2019](https://wiki-ffo-jp.translate.goog/html/37694.html?_x_tr_sl=ja&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=sc) , it has been further increased by +4 from the above.`

Cast Blind II and still have it be a massive waste of time.

### DIFF
--- a/scripts/globals/spells/enfeebling_spell.lua
+++ b/scripts/globals/spells/enfeebling_spell.lua
@@ -131,7 +131,7 @@ xi.spells.enfeebling.calculatePotency = function(caster, target, spellId, spellE
             statDiff = caster:getStat(statUsed) - target:getStat(xi.mod.MND)
 
             if spellId == xi.magic.spell.BLIND_II then
-                potency = utils.clamp(statDiff * 0.375 + 49, 15, 90) -- Values from JP wiki: http://wiki.ffo.jp/html/3449.html
+                potency = utils.clamp(statDiff * 0.375 + 49, 19, 94) -- Values from JP wiki: http://wiki.ffo.jp/html/3449.html
             else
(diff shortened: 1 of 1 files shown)


scripts/globals/spells/enfeebling_spell.lua
```diff
@@ -131,7 +131,7 @@ xi.spells.enfeebling.calculatePotency = function(caster, target, spellId, spellE
             statDiff = caster:getStat(statUsed) - target:getStat(xi.mod.MND)
 
             if spellId == xi.magic.spell.BLIND_II then
-                potency = utils.clamp(statDiff * 0.375 + 49, 15, 90) -- Values from JP wiki: http://wiki.ffo.jp/html/3449.html
+                potency = utils.clamp(statDiff * 0.375 + 49, 19, 94) -- Values from JP wiki: http://wiki.ffo.jp/html/3449.html
```
TeoTwawki, line 134: If I'm understanding the math right over on the multiple wiki's, the reason the rang is 19-94 is simply because that's the most the dStat can push (-80 to +120). I don't see why we'd need to clamp the range at this spot if our other steps are correct. /shrug

but since the clampage already existed, I'll stamp the range change.

TeoTwawki (approved):

WinterSolstice8 (approved):

PR #4994, opened 2024-01-14
Table some optional parameters for `phOnDespawn()`

TeoTwawki:
**_I affirm:_**

- Tables some optional parameters for `phOnDespawn()`
- Also adds new check for nighttime only NMs, because `disallowRespawn()` bypasses that spawnType.
- And another to not call UpdateNMSpawnPoint() because that errors on the ones that we should not be using it on in the 1st place. 

All the _required_ args still live outside the table. anything that is optional and can be fully omitted will live in the params table.

Have additionally introduced a localVar method of preventing cooldown trigger on mobs that were not actually fought. default behavior is exactly as it was before this since non existing variable return 0. An example usage lives in citipati.lua

cooldown avoidance:
- go to attohwa chasm at night, kill corses till citipati spawns. Leave him up till daytime and watch him despawn
- after he despawns note that he is still "lottery primed" to pop again.

`params.nightOnly`:
- copy the same params in the corses script over to any NM placeholder that spawns during day (corses do not), lets say rock lizards.
- crank pop rate to 100% for the test
- note the NM will not pop in daytime (4:00 to 20:00)
- retest after night, notice NM spawn right away

`params.noPosUpdate`:
- You might want to do this at the same time as the nightOnly test. Spawn the same NM repeatedly, notice its position is not being updated. In several cases this is desired behavior either because positioning is handled in script or because the mob simply doe snot HAVE multiple locations - in any case if there are no entries in the NM spawn pos table the function that updates the NMs spawn position will error. Having this parameter set simply prevents that from being attempted.

`params.immediate`:
- This used to be a stand alone argument. its merely been tabled.
- To verify its still working, go to sky and kill groundskeepers till you get Despot

### DIFF
--- a/scripts/globals/mobs.lua
+++ b/scripts/globals/mobs.lua
@@ -32,9 +32,24 @@ local function lotteryPrimed(phList)
 end
 
 -- potential lottery placeholder was killed
-xi.mob.phOnDespawn = function(ph, phList, chance, cooldown, immediate)
-    if type(immediate) ~= 'boolean' then
-        immediate = false
+xi.mob.phOnDespawn = function(ph, phList, chance, cooldown, params)
+    params = params or {}
+    --[[
+        params = { immediate   = true }    pop NM without waiting for next PH pop time
+        params = { nightOnly   = true }    spawn NM only at night time
+        params = { noPosUpdate = true }    do not run UpdateNMSpawnPoint()
+    ]]
+
+    if type(params.immediate) ~= 'boolean' then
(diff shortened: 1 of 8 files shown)


TeoTwawki:
my scope creeped a little bit here, as I found Citipati had the same issue with spawntype_night not being obeyed. adding exceptions into the core side for `DisallowRespawn()` did not appeal to me.

Xaver-DaRed (approved):
LGTM

PR #4995, opened 2024-01-14
[SQL] Oneiros grip latent mpp was greater than 100

MowFord:
**_I affirm:_**

Confirmed via code but also loaded up server to confirm. latentid 4 is real mpp, so oneiros grip was giving refresh if mp % was less than 750... so all the time

Equip oneiros grip and see `!getmod refresh` increment only after dipping below 75% mp


MowFord:
FWIW i named this branch generically because i expected more, but we good:

image

helixhamin (approved):

Xaver-DaRed (approved):

TeoTwawki (approved):

PR #4999, opened 2024-01-15
Do not consume ammo with Unlimited Shot active

MowFord:
Also do not consume Unlimited Shot if RA missed

**_I affirm:_**

Unlimited shot currently is coded to only prevent consumption of ammo on a landed Ranged Attack. The buff is preserved if the attack misses (correctly), but it should _also_ not consume the ammo if the attack misses. This resolves the latter.

https://ffxiclopedia.fandom.com/wiki/Unlimited_Shot

> Duration: 1:00 or One Successful Ranged Attack

and 

> Once activated, this ability will remain in effect until you successfully hit your mark.

Use Unlimited Shot and equip a single pebble as level 1, throw at a high level mob to see that the pebble is not consumed nor is the buff consumed. Throw until it lands (or enable godmode to speed that part up) to see that Unlimited Shot buff wears, but the pebble is _still_ not consumed on that particular throw. Throw once more to see the pebble consumed as normal

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1831,11 +1831,15 @@ void CCharEntity::OnRangedAttack(CRangeState& state, action_t& action)
 
         recycleChance += this->PJobPoints->GetJobPointValue(JP_AMMO_CONSUMPTION);
 
-        // Only remove unlimited shot on hit
-        if (hitOccured && this->StatusEffectContainer->HasStatusEffect(EFFECT_UNLIMITED_SHOT))
+        if (this->StatusEffectContainer->HasStatusEffect(EFFECT_UNLIMITED_SHOT))
         {
-            StatusEffectContainer->DelStatusEffect(EFFECT_UNLIMITED_SHOT);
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):

TeoTwawki:
ohhh it was eating the ammo. I thought you were telling me the effect wore off only.

TeoTwawki (approved):

src/map/entities/charentity.cpp
```diff
@@ -1831,11 +1831,15 @@ void CCharEntity::OnRangedAttack(CRangeState& state, action_t& action)
 
         recycleChance += this->PJobPoints->GetJobPointValue(JP_AMMO_CONSUMPTION);
 
-        // Only remove unlimited shot on hit
-        if (hitOccured && this->StatusEffectContainer->HasStatusEffect(EFFECT_UNLIMITED_SHOT))
```
TeoTwawki, line 1835: This already was checking hitOccured 🤔 

~~strange.~~ _oh because its using recycle to keep the ammo_

PR #5004, opened 2024-01-16
[Magic] Implement player Death damage

Xaver-DaRed:
**_I affirm:_**

- Separates logic for Death when used by players.
- (Player only) Implements death as a nuke when insta-kill fails.
- (Player only) Makes player death insta-kill trigger as shitty as retail. Or worse.
- (Player only) Makes spell consume all your MP, no matter what.

Get BLM job point death spell. Cast it. See it consume all your MP and either do damage or, sometimes, insta-kill an enemy.

### DIFF
--- a/scripts/actions/spells/black/death.lua
+++ b/scripts/actions/spells/black/death.lua
@@ -1,30 +1,63 @@
 -----------------------------------
 -- Spell: Death
--- Consumes all MP. Has a chance to knock out the target. If Death fails to knock out the target, it
--- will instead deal darkness damage. Ineffective against undead.
+-- Has a chance to insta-kill the target. Ineffective against undead or notorious monsters.
+-- (Player only) Consumes all MP no matter what.
+-- (Player only) If Death fails to knock out the target, it will instead deal darkness damage.
 -----------------------------------
 local spellObject = {}
 
 spellObject.onMagicCastingCheck = function(caster, target, spell)
     spell:setFlag(xi.magic.spellFlag.IGNORE_SHADOWS)
+
     return 0
 end
 
 spellObject.onSpellCast = function(caster, target, spell)
-    if
-        target:isUndead() or
-        target:hasStatusEffect(xi.effect.MAGIC_SHIELD) or
-        -- Todo: DeathRes has no place in the resistance functions so far..
-        math.random(1, 100) <= target:getMod(xi.mod.DEATHRES)
-    then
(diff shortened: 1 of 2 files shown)


scripts/actions/spells/black/death.lua
```diff
@@ -1,30 +1,70 @@
...
-        target:isUndead() or
-        target:hasStatusEffect(xi.effect.MAGIC_SHIELD) or
-        -- Todo: DeathRes has no place in the resistance functions so far..
-        math.random(1, 100) <= target:getMod(xi.mod.DEATHRES)
-    then
-        spell:setMsg(xi.msg.basic.MAGIC_NO_EFFECT)
-        return 0
-    end
+    -- Player spell.
+    if caster:isPC() or caster:isTrust() then
```
zach2good: Luckily, the check for `isTrust` isn't needed here because `isMob` _should_ return false for trusts. 

```cpp
bool CLuaBaseEntity::isMob()
{
    return m_PBaseEntity->objtype == TYPE_MOB;
}

bool CLuaBaseEntity::isTrust()
{
    return m_PBaseEntity->objtype == TYPE_TRUST;
}
```

It's checking an enum, not a flag. I think it's better to have a purely player-only branch and a purely mob-only branch, and anything else that manages to find their way here will get `return 0` or nothing (`return nil`)

scripts/actions/spells/black/death.lua
```diff
@@ -1,30 +1,70 @@
...
+
+        return result
+
+    -- Not-player/trust spell.
+    else
+        local targetResistance = target:getMod(xi.mod.DEATHRES)
+        local randomRoll       = math.random(1, 100)
+
+        if targetResistance > 0 then
+            targetResistance = targetResistance + 5 -- Assuming it works as any other trait and gets an extra 5% at first level.
```
TeoTwawki: > Assuming it works as any other trait 

Unfortunately it does not, there is no base 5% chance to escape it. You get what the gear has and nothing else. If your gear is +2 you get just 2, not 7. Some non-spell based forms of it can be evaded by reraise (and will consume the reraise) but the cast just straight up kills you unless gear procs.

scripts/actions/spells/black/death.lua
TeoTwawki: I think he was wanting to use the player logic for the trusts casting it tho

TeoTwawki (approved):

zach2good (approved):

PR #5009, opened 2024-01-17
Sacrifice fix

mmagdon:
**_I affirm:_**

The WHM spell Sacrifice had no logic in it to support its interaction with Afflatus Solace for transferring multiple (also Erasable) status effects at once. This adds it in.

Apply several status effects to a character in a party (either naturally or with !addeffect)
Have a White Mage use the job ability Afflatus Solace then cast Sacrifice on the target
Up to 7 status effects (the base effects that can be transferred without Afflatus Solace + those that can be Erased) should be transferred from the target to the caster


CatsEyeXI:
why'd you close this? seems to work well

mmagdon:
I was having trouble updating this PR through Gitlens. I am new to Github,
but not new to writing code. I was working with Mowford for a bit but
health issues stopped me in my tracks from most/all of my hobbies (this
included). More than happy to pass the code onto someone else or if someone
can just tell me what to do i'll do it when I can.

PR #5010, opened 2024-01-17
Add Afflatus Solace logic to Sacrifice

mmagdon:
**_I affirm:_**

The WHM spell Sacrifice had no logic in it to support its interaction with Afflatus Solace for transferring multiple (also Erasable) status effects at once. This adds it in.

Apply several status effects to a character in a party (either naturally or with !addeffect)
Have a White Mage use the job ability Afflatus Solace then cast Sacrifice on the target
Up to 7 status effects (the base effects that can be transferred without Afflatus Solace + those that can be Erased) should be transferred from the target to the caster

### DIFF
--- a/sql/item_mods.sql
+++ b/sql/item_mods.sql
@@ -52412,6 +52412,113 @@ INSERT INTO `item_mods` VALUES (23358,303,4);     -- TREASURE_HUNTER: 4
 INSERT INTO `item_mods` VALUES (23358,384,400);   -- HASTE_GEAR: 4%
 INSERT INTO `item_mods` VALUES (23358,896,7);     -- DESPOIL: 7
 
+-- Chevalier's sabatons +2
+INSERT INTO `item_mods` VALUES (23359,1,111);   -- DEF: 111
+INSERT INTO `item_mods` VALUES (23359,2,42);    -- HP:  42
+INSERT INTO `item_mods` VALUES (23359,5,34);    -- MP:  34
+INSERT INTO `item_mods` VALUES (23359,8,29);    -- STR: 29
+INSERT INTO `item_mods` VALUES (23359,9,23);    -- DEX: 23
+INSERT INTO `item_mods` VALUES (23359,10,35);   -- VIT: 35
+INSERT INTO `item_mods` VALUES (23359,11,38);   -- AGI: 38
+INSERT INTO `item_mods` VALUES (23359,13,17);   -- MND: 17
+INSERT INTO `item_mods` VALUES (23359,14,30);   -- CHR: 30
+INSERT INTO `item_mods` VALUES (23359,23,50);   -- ATT: 50
+INSERT INTO `item_mods` VALUES (23359,25,50);   -- ACC: 50
+INSERT INTO `item_mods` VALUES (23359,27,13);   -- ENMITY: 13
+INSERT INTO `item_mods` VALUES (23359,29,5);    -- MDEF: 5
+INSERT INTO `item_mods` VALUES (23359,30,50);   -- MACC: 50
+INSERT INTO `item_mods` VALUES (23359,31,126);  -- MEVA: 126
+INSERT INTO `item_mods` VALUES (23359,68,84);   -- EVA:  84
+INSERT INTO `item_mods` VALUES (23359,170,10);  -- FASTCAST: 10
+INSERT INTO `item_mods` VALUES (23359,384,500); -- HASTE_GEAR: 5%
+INSERT INTO `item_mods` VALUES (23359,1062,18); -- ENHANCES_DIVINE_EMBLEM: 18
+
+-- Heathen's sollerets +2
+INSERT INTO `item_mods` VALUES (23360,1,107);   -- DEF: 107
+INSERT INTO `item_mods` VALUES (23360,2,38);    -- HP:  38
+INSERT INTO `item_mods` VALUES (23360,5,34);    -- MP:  34
+INSERT INTO `item_mods` VALUES (23360,8,28);    -- STR: 28
+INSERT INTO `item_mods` VALUES (23360,9,20);    -- DEX: 20
+INSERT INTO `item_mods` VALUES (23360,10,25);   -- VIT: 25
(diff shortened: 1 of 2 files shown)


scripts/actions/spells/white/sacrifice.lua
```diff
@@ -10,25 +11,48 @@ end
...
+                    caster:addStatusEffect(effect, statusEffect:getPower(), statusEffect:getTickCount(), statusEffect:getDuration())
+                    transferCount = transferCount + 1
+                end
+
+                target:delStatusEffect(effect)
             end
+        else
+            break  -- Break the loop if the maximum number of effects are transferred
+        end
+    end
```
Xaver-DaRed: Instead of wrapping the logic inside the counter logic, we could move it to the end. Less depth = easier to read.
```lua
    -- Maximum number of effects to transfer
    local maxTransferCount = caster:hasStatusEffect(xi.effect.AFFLATUS_SOLACE) and 7 or 1
    local transferCount    = 0  -- Initialize transferCount

    -- Loop over the original effects
    for _, effect in ipairs(removables) do
        if target:hasStatusEffect(effect) then
            local statusEffect = target:getStatusEffect(effect)

            -- only add it to me if I don't have it
            if not caster:hasStatusEffect(effect) then
                caster:addStatusEffect(effect, statusEffect:getPower(), statusEffect:getTickCount(), statusEffect:getDuration())
                transferCount = transferCount + 1
            end

            target:delStatusEffect(effect)
        end

        if transferCount >= maxTransferCount then
            break
        end
    end
```
^ This is an example coded in-promptu in this editor. So dont take it as gospel

scripts/actions/spells/white/sacrifice.lua
```diff
@@ -10,25 +11,48 @@ end
...
 
-            target:delStatusEffect(effect)
-            return 1
+    -- Check if the maximum transfer count is not reached
+    while transferCount < maxTransferCount do
+        local erasableEffect = caster:stealStatusEffect(target, xi.effectFlag.ERASABLE)
+        if erasableEffect ~= 0 then
+            transferCount = transferCount + 1
+        else
+            break  -- Break the loop if no more erasable effects are available
```
Xaver-DaRed: I dont understand what this is doing or why. Could you explain it to me?

scripts/actions/spells/white/sacrifice.lua
mmagdon: I can make some edits to this, let me try and get back to you. Thank you for the suggestion.

scripts/actions/spells/white/sacrifice.lua
mmagdon: After iterating through the "base" Sacrifice enfeebling effects that it can remove, the code iterates through the erasable effects. In Afflatus Solace stance, a White Mage can transfer up to 7 effects so this will continue the loop after going through the initial list to look for anything that can be erased.

scripts/actions/spells/white/sacrifice.lua
mmagdon: Note, I'm not exactly a programmer, just a bored sysadmin who wants to contribute to fixing my favorite game! So I appreciate the patience while I dust off my programming brain

scripts/actions/spells/white/sacrifice.lua
```diff
@@ -10,25 +11,48 @@ end
...
+            break  -- Break the loop if the maximum number of effects are transferred
+        end
+    end
 
-            target:delStatusEffect(effect)
-            return 1
+    -- Check if the maximum transfer count is not reached
+    while transferCount < maxTransferCount do
+        local erasableEffect = caster:stealStatusEffect(target, xi.effectFlag.ERASABLE)
+        if erasableEffect ~= 0 then
```
WinterSolstice8:
```suggestion
        if erasableEffect then
```
lua calls return `nil` when there's no "object", not null pointers (int 0)

scripts/actions/spells/white/sacrifice.lua
```diff
@@ -10,25 +11,48 @@ end
...
+    local maxTransferCount = caster:hasStatusEffect(xi.effect.AFFLATUS_SOLACE) and 7 or 1
+
+    -- Use the helper function to transfer status effects
+    local transferCount = 0  -- Initialize transferCount
 
-            local statusEffect = target:getStatusEffect(effect)
+    -- Loop over the original effects
+    for _, effect in ipairs(removables) do
+        if transferCount < maxTransferCount then
+            if target:hasStatusEffect(effect) then
```
claywar: Wouldn't it make more sense here to iterate over the target's effects as opposed to a list of possible effects for matching purposes?

mmagdon: I can try to do that!

scripts/actions/spells/white/sacrifice.lua
mmagdon: TIL, thank you for the advice! It's much appreciated!

scripts/actions/spells/white/sacrifice.lua
mmagdon: It looks like I can use target:getStatusEffects() to iterate over the target's effects. I can try my best to implement this change if we want the solution to be best practice. I was proud to get this working in the first place!

scripts/actions/spells/white/sacrifice.lua
mmagdon: Hey! So looking more into this, stealStatusEffect actually returns uint16 so I should be checking for 0 here. I think this could be left in but I'll test out your suggested change and see what it does ingame.

scripts/actions/spells/white/sacrifice.lua
mmagdon: I'm having a little bit of trouble getting this to work, I'm gonna keep trying

mmagdon:
I think I messed up my last commit -- I tried to update my local branch from upstream and I don't know why it's showing up as a separate commit. This is the first time I've "updated" a PR, more than happy to redo things if I have to to abide by LSB process.

PR #5015, opened 2024-01-18
[SQL] set scouts beret to 25 recycle chance

lapislosh:
**_I affirm:_**

Fixes Scout's Beret and Scout's Beret +1 to have a 25% recycle chance, which matches Scout's Beret +2's current value.

https://ffxiclopedia.fandom.com/wiki/Scout's_Beret Wiki puts it at 20-30% which is likely where the 25% on Scout's Beret +2 comes from. My theory for what happened is that someone put the original 5 recycle under the assumption that each recycle merit was 5%. It might have been right a long time ago and then been refactored, but the item mod never got updated. Or someone just typed it in wrong, who knows!

!additem 15082
!additem 15255
I fired 20 arrows with only Scout's Beret +1 equipped and only used up 13.


helixhamin (approved):
Looks good based on bgwiki. Wikia also seems to line up, and FFXIAH, someone mentioned 33%, but they could just be guessing a bit high. I would say this number looks good.

Xaver-DaRed (approved):

PR #5018, opened 2024-01-21
Fixed Chocobo Shield Model Id

Thorinori:
**_I affirm:_**

Fixed the Model ID for Chocobo Shield and its +1 version. They were off by 10 from their proper value

1) Look at Chocobo Shield
2) Be happy because it now has a cute little chocobo on it instead of being Svalinn


TeoTwawki (approved):

Xaver-DaRed (approved):

Thorinori:
Is that MultiInstance check failing just for db changes normal? No clue what could have caused that

Xaver-DaRed:
It sometimes fails. Nothing you gotta worry about.

TeoTwawki:
I'm just gonna yolo it - its only a model change nothing here can harm anything and the individual builds/sql imports went fine

PR #5019, opened 2024-01-21
[Fix] Correct Behavior of common Toads

Spicyryan:
**_I affirm:_**

Toads in retail have the following behavior:
- Do not have any TP moves.
- Do not cast any spells.
- Close in to melee immediately, and never stand at a distance.

Toads in the current emulator do all of these things incorrectly. This PR sets the skill list and spell list to 0, and adds a mob_mod of NO_STANDBACK to the toads.

This was done to all standard toads encountered in the world from Sarutabaruta (S) to Abyssea to Adoulin. Changes were not applied to any non-standard Toads from Lair Reives, Fishing, etc in order to keep the scope of this change simple.

- Verified the behavior on retail.
- Changed entries.
- Tested locally on Toad in West Sarutabaruta (S). !gotoid 17166708
- Attacked Toad. It closed in to melee immediately.
- Toad never casted any spells.
- !tp 3000 to the Toad, and it did not weaponskill.
- Applied these same few changes to the toads in this PR.


Xaver-DaRed (approved):

cocosolos:
It looks like these are all using the poroggo family which is beastmen type and there is a commented out frogs family that is aquan type. I wonder if it would be preferable to change the family instead, that way these pool mods could just be a single family mod, and also fix a possible issue with frogs being marked as beastmen.

PR #5021, opened 2024-01-22
[Interaction] Fix event parameter on Windurst 8-2

Xaver-DaRed:
**_I affirm:_**

Imported from ASB: https://github.com/AirSkyBoat/AirSkyBoat/pull/3695

- Fix (adds) missing parameter.
- Uses Key item enum for parameter value.
- Makes !pos for cracked wall more convenient.

Set your mission, to Windurst 8-2, get to the event and see the KI being properly referenced in game.

### DIFF
--- a/scripts/missions/windurst/8_2_The_Jester_Whod_be_King.lua
+++ b/scripts/missions/windurst/8_2_The_Jester_Whod_be_King.lua
@@ -13,7 +13,7 @@
 -- Tosuka-Porika    : !pos -26 -6 103 238
 -- Kupipi           : !pos 2 0.1 30 242
 -- Shantotto        : !pos 122 -2 112 239
--- _5e5 (Cr. Wall)  : !pos -424.255 -1.909 619.995 194
+-- _5e5 (Cr. Wall)  : !pos -423 0 619.75 194
 -- _5cb (Gate. Drk) : !pos -228 0 99 192
 -----------------------------------
 local outerHorutotoID = zones[xi.zone.OUTER_HORUTOTO_RUINS]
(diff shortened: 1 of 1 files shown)


scripts/missions/windurst/8_2_The_Jester_Whod_be_King.lua
```diff
@@ -13,7 +13,7 @@
 -- Tosuka-Porika    : !pos -26 -6 103 238
 -- Kupipi           : !pos 2 0.1 30 242
 -- Shantotto        : !pos 122 -2 112 239
--- _5e5 (Cr. Wall)  : !pos -424.255 -1.909 619.995 194
+-- _5e5 (Cr. Wall)  : !pos -423 0 619.75 194
```
TeoTwawki, line 16:
```sql

INSERT INTO `npc_list` VALUES (17572260,'_5e5','Cracked Wall',0,-424.255,-1.909,619.995,1,40,40,9,0,0,0,3,0x0200000000000000000000000000000000000000,0,NULL,0);
```
its the exact pos in our database

Xaver-DaRed: yeah, this change puts you in front of the npc instead of inside it.

I can roll it back if you want though. It aint important in the grand scheme of the PR

TeoTwawki: we're normally just listing where the NPC is here, can just wallhack out of the npc, since if using it obviously is a GM. but its _technically not wrong_

TeoTwawki (approved):

PR #5023, opened 2024-01-22
[Magic] Make cure spells use new damage functions and misc fixes

Xaver-DaRed:
**_I affirm:_**

- Added tier II helixes to `isHelix` check.
- Removes 2 unused functions from magic.lua.
- Implements a hard cap for "stat difference" when using the old magic damage system.
- Simplifies the method to choose stat diff system.
- Adds check for Black magic when under `Ebullience` effect.
- Adds check for Divine magic when under `Divine Emblem` effect.
- Adds cure spells to damage table (for when they are used as nukes against undead).
- Adds `Divine Seal` multiplier when using healing spells as nukes.
- Changes cure spell scripts to use damage_spell.lua when using them against undead.
- Fix Horde Lullaby inflicting `LULLABY` effect instead of `SLEEP_I` effect, as weird as that sounds.
- Fix "Ice Maker" attachment giving over-the-top MAB.
  - Issue was that the description given by wikis, while correct, can be taken out of context, regarding the 100 base.
  - Its supposed to apply a multiplier to MAB in the very final step, right before MAB/MDB.

- Cast healing spells on undead and see them do damage. They already did, but now should be less made up. Use Divine Seal while you are at it.
- Cast AoE Lullaby spells. They should apply SLEEP_I effect.
- Cast black magic being OP on lvl 1 mobs WHILE using the old system in settings. BE mad.
- Use divine emblem and ebullience. Whatch it only have effect (and effect being removed) when usng the proper spells.

### DIFF
--- a/scripts/actions/abilities/pets/attachments/ice_maker.lua
+++ b/scripts/actions/abilities/pets/attachments/ice_maker.lua
@@ -9,41 +9,33 @@ attachmentObject.onEquip = function(automaton)
             return
         end
 
-        local master = pet:getMaster()
-        local maneuvers = master:countEffect(xi.effect.ICE_MANEUVER)
-        local amount = 100 + pet:getMod(xi.mod.MATT)
-        if maneuvers == 1 then
-            amount = amount * 0.2
-            pet:setLocalVar('icemakermaneuvers', 1)
-        elseif maneuvers == 2 then
-            amount = amount * 0.4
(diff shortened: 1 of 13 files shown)


Xaver-DaRed:
Tested and ready for review.

Frankie-hz:
Math lgtm

TeoTwawki (approved):

PR #5025, opened 2024-01-23
GCC 13 Clang 15 Workflow

claybie:
**_I affirm:_**

Builds the project with Clang 15 and GCC 13 in the workflow. Workarounds for Clang and MariaDB have been removed as they appear to no longer be necessary.

Run the workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,25 +19,13 @@ jobs:
   Sanity_Checks:
     runs-on: ubuntu-22.04
     steps:
-    # workaround for broken clang on ubuntu runner until fixed: https://github.com/actions/runner-images/issues/8659
-    - uses: claywar/workaround8649@4892524a133793c85e25513fae79fc968e63877c
-      with:
-        os: ubuntu-22.04
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
     - name: Install Dependencies
       run: |
-        # https://mariadb.com/docs/connect/programming-languages/c/install/#connector-c-install-repo-configure-cs
-        # Ubuntu is using an old libmariadb but python3 pip expects it to be newer. Use the mariadb provided deb packages.
-        wget -P /tmp/ https://downloads.mariadb.com/MariaDB/mariadb_repo_setup
-        # Note: there is a sha256sum command here according to the above link, but I am unsure if this file may change with time
-        chmod +x /tmp/mariadb_repo_setup
-        sudo /tmp/mariadb_repo_setup --mariadb-server-version="mariadb-10.6"
-        # Remove the above when it's no longer necessary
-
         sudo apt-get update
-        sudo apt-get install -y software-properties-common cppcheck luajit-5.1-dev luarocks mariadb-server-10.6 mariadb-client-10.6 libmariadb-dev-compat binutils-dev
+        sudo apt-get install -y gcc-13 clang-15 clang-tidy-15 clang-format-15 software-properties-common cppcheck luajit-5.1-dev luarocks mariadb-server-10.6 mariadb-client-10.6 libmariadb-dev-compat binutils-dev
         pip install -r tools/requirements.txt
         luarocks install luacheck --local
         npm install -g diff-so-fancy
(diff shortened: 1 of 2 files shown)


WinterSolstice8 (approved):

.github/workflows/build.yml
```diff
@@ -19,25 +19,13 @@ jobs:
...
-        # Ubuntu is using an old libmariadb but python3 pip expects it to be newer. Use the mariadb provided deb packages.
-        wget -P /tmp/ https://downloads.mariadb.com/MariaDB/mariadb_repo_setup
-        # Note: there is a sha256sum command here according to the above link, but I am unsure if this file may change with time
-        chmod +x /tmp/mariadb_repo_setup
-        sudo /tmp/mariadb_repo_setup --mariadb-server-version="mariadb-10.6"
-        # Remove the above when it's no longer necessary
-
         sudo apt-get update
-        sudo apt-get install -y software-properties-common cppcheck luajit-5.1-dev luarocks mariadb-server-10.6 mariadb-client-10.6 libmariadb-dev-compat binutils-dev
+        sudo apt-get install -y gcc-13 clang-15 clang-tidy-15 clang-format-15 software-properties-common cppcheck luajit-5.1-dev luarocks mariadb-server-10.6 mariadb-client-10.6 libmariadb-dev-compat binutils-dev
```
zach2good: There's no need to add these to the install section, since these are already present on the build runner machine

.github/workflows/build.yml
```diff
@@ -103,12 +91,12 @@ jobs:
     - name: CPP Formatting Checks
```
zach2good: While you're here can you rename this step "CPP Formatting Checks (clang-format-15)" so it's super clear what version we use?

.github/workflows/build.yml
```diff
@@ -193,30 +181,26 @@ jobs:
...
-      with:
-        os: ubuntu-22.04
     - uses: actions/checkout@v3
       with:
         fetch-depth: 1
     - name: Install Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y software-properties-common cmake libmariadb-dev-compat libluajit-5.1-dev libzmq3-dev zlib1g-dev libssl-dev binutils-dev
+        sudo apt-get install -y gcc-13 clang-15 clang-tidy-15 clang-format-15 software-properties-common cmake libmariadb-dev-compat libluajit-5.1-dev libzmq3-dev zlib1g-dev libssl-dev binutils-dev
```
zach2good: As mentioned before, this change isn't needed

.github/workflows/build.yml
```diff
@@ -267,30 +251,26 @@ jobs:
...
```
zach2good: As mentioned before, this change isn't needed

.github/workflows/build.yml
```diff
@@ -444,7 +420,7 @@ jobs:
     - name: Install Dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y software-properties-common cmake mariadb-client-10.6 libmariadb-dev-compat libluajit-5.1-dev libzmq3-dev zlib1g-dev libssl-dev luarocks binutils-dev
+        sudo apt-get install -y gcc-13 clang-15 clang-tidy-15 clang-format-15 software-properties-common cmake mariadb-client-10.6 libmariadb-dev-compat libluajit-5.1-dev libzmq3-dev zlib1g-dev libssl-dev luarocks binutils-dev
```
zach2good: As mentioned before, this change isn't needed

.github/workflows/build.yml
```diff
@@ -625,7 +597,7 @@ jobs:
...
```
zach2good: As mentioned before, this change isn't needed

tools/run_clang_format.sh
```diff
@@ -1,3 +1,3 @@
 # Run from repo root
-for f in $(find src/ -name '*.h' -or -name '*.cpp'); do clang-format-14 -style=file -i $f; done
-for f in $(find modules/ -name '*.h' -or -name '*.cpp'); do clang-format-14 -style=file -i $f; done
+for f in $(find src/ -name '*.h' -or -name '*.cpp'); do clang-format-15 -style=file -i $f; done
```
zach2good, line 2: 👍

zach2good (commented):
Looking good, just needs some extraneous install commands taken out 👍

.github/workflows/build.yml
claybie: I seem to be in the habit of adding packages to my Gentoo GitHub workflow where this is required :crying_cat_face:

claybie (commented):
Superfluous package downloads removed.

PR #5033, opened 2024-01-27
Add Immanence functionality

GodModeXI:
**_I affirm:_**

- This pull request adds the necessary framework to handle magic skillchains from the use of the Scholar ability "Immanence" 

1.  Apply the changed files from this pull request to your local repository.
2.  Clean and rebuild the solution.
3.  Login character and load up a scholar with a level higher than level 87.
4.  Use Dark Arts > Strategems > Addendum Black
5.  Use Strategems > Immanence
6.  Cast elemental spell (I used Blizzard V)
7.  Use Strategems > Immanence
8.  Cast elemental T1 skillchain element (I used Water V)
9.  Observe skill chain (Mine was Fragmentation.

image


Xaver-DaRed:
From what I'm seing, this could (and probably should) be handled in lua, in spell_damage.lua.

And if not:
Couldnt we use a case with getElement() instead of a elseif chain?
How does this interact with magic bursts?

I'll let others chime in

GodModeXI:
> From what I'm seing, this could (and probably should) be handled in lua, in spell_damage.lua.
> 
> And if not: Couldnt we use a case with getElement() instead of a elseif chain? How does this interact with magic bursts?
> 
> I'll let others chime in

I was attempting to convert our handling over to LSB from our server. I haven’t spent a lot of time in the LSB repo as of late, so if lua is the preferred method I can take another look and create a new pull request using that suggestion and close this one.

TeoTwawki:
> How does this interact with magic bursts?

Its description says: 

> Makes it possible for your next elemental magic spell to be used in a skillchain, but not a magic burst.

And from the bg wiki's page:
> Allows spells to interact and Skillchain as if they each had their own level 1 Skillchain properties.
Example: Immanence -> Blizzard IV -> Immanence -> Water IV -> Fragmentation
Closing a skillchain with an immanence Helix will make the magic burst window longer.

So basically your cast gets treated like a ws would have.

To my knowledge we don't have a script way to make things other than skills (weaponskills, mobskills, pet abilities) have sc properties. I'd love to see that added to spells in its binding set tho, would be pretty neat. I have no idea what that will look like - might be stupid easy, might be difficult. It _looks_ like functionality at the spell level might already exist because of blue magic, and we'd just need bindings? https://github.com/LandSandBoat/server/blob/base/src/map/spell.cpp#L595 not 100% on what we do with that without a deep dive.

I think converting that if/else into a switch case would be good enough if its not a simple conversion. I'd look at whatever we're doing with blue magic, I've a feeling most of that is actually still sql+core side, and we could just convert all of it together at a later date if so.

TeoTwawki:
I also kinda missed (till after I posted that) that we'll need to _prevent_ magic bursts during this.

I seriously have never seen this used and never leveled sch past 60 on retail, had no idea this was even a thing till today.

GodModeXI:
> > How does this interact with magic bursts?
> 
> Its description says:
> 
> > Makes it possible for your next elemental magic spell to be used in a skillchain, but not a magic burst.
> 
> And from the bg wiki's page:
> 
> > Allows spells to interact and Skillchain as if they each had their own level 1 Skillchain properties.
> > Example: Immanence -> Blizzard IV -> Immanence -> Water IV -> Fragmentation
> > Closing a skillchain with an immanence Helix will make the magic burst window longer.
> 
> So basically your cast gets treated like a ws would have.
> 
> To my knowledge we don't have a script way to make things other than skills (weaponskills, mobskills, pet abilities) have sc properties. I'd love to see that added to spells in its binding set tho, would be pretty neat. I have no idea what that will look like - might be stupid easy, might be difficult. It _looks_ like functionality at the spell level might already exist because of blue magic, and we'd just need bindings? https://github.com/LandSandBoat/server/blob/base/src/map/spell.cpp#L595 not 100% on what we do with that without a deep dive.
> 
> I think converting that if/else into a switch case would be good enough if its not a simple conversion. I'd look at whatever we're doing with blue magic, I've a feeling most of that is actually still sql+core side, and we could just convert all of it together at a later date if so.

Yea you are correct Teo, elemental spells don’t have skillchain properties on their own, they have to be added similarly to how chain affinity works with blue magic. The blue magic spells have core binding for skillchain properties that are pulled from the blue_spell_list.sql. Which this is why they have their own unique tables. This should really be handled similarly to that, but I have no idea how to modify SQL tables and how to bind them property in the core. Still learning. Adding them this way was the best alternative that didn’t alter the spell tables. 

I’ll try to build a switch case for now as suggested.

TeoTwawki:
yeah I think thats the best way forward rather than doing sql tabling - we'll eventually move this and the existing blu stuff to lua tabling most likely, and kill some static sql data that way

UmeboshiXI:
I have a similar implementation of this on my server, I based it off how the blue magic function worked. For the magic burst, I just added checks in damage_spell so it won't return the magicburstbonus if you have the effect.

src/map/entities/charentity.cpp
```diff
@@ -1077,6 +1077,80 @@ void CCharEntity::OnCastFinished(CMagicState& state, action_t& action)
...
+            if (actionTarget.param > 0 &&
+                PSpell->dealsDamage() &&
+                PSpell->getSpellGroup() == SPELLGROUP_BLACK &&
+                (StatusEffectContainer->HasStatusEffect(EFFECT_IMMANENCE)))
+            {
+                SUBEFFECT effect = SUBEFFECT_NONE;
+                switch (PSpell->getSpellFamily())
+                {
+                    case SPELLFAMILY_STONE:
+                        effect = battleutils::GetSkillChainEffect(PTarget, 4, 0, 0);
```
WinterSolstice8: Does `PSpell->getElement()` match the element that would be set in GetSkillChainEffect? If it does, you could have a less complicated case/switch with one case for spells that _can_ skillchain and a default case for ones that can't.

I don't think Bio/Drain can make a skillchain here, so checking spellfamily is probably required.

src/map/entities/charentity.cpp
```diff
@@ -1077,6 +1077,80 @@ void CCharEntity::OnCastFinished(CMagicState& state, action_t& action)
...
+                    }
+                    else
+                    {
+                        actionTarget.addEffectParam   = skillChainDamage;
+                        actionTarget.addEffectMessage = 287 + effect;
+                    }
+                    actionTarget.additionalEffect = effect;
+                }
+
+                StatusEffectContainer->DelStatusEffectSilent(EFFECT_IMMANENCE);
```
WinterSolstice8: On retail, if no skillchain is possible at all (e.g. bio/drain) does it still delete immanence?

src/map/entities/charentity.cpp
GodModeXI: I think we opted for spellfamily to specifically prevent skillchains on anything other than the elemental wheel and helixes that scholar has access to under dark Arts / Addendum black. Now that you mention it, aga spells on a BLM sub would be available to a scholar and those are excluded in this switch. 

I am going to reach out to some of the capture guys and see if I can get this confirmed.

src/map/entities/charentity.cpp
GodModeXI: I'm reaching out to some retail capture guys to try to get this detail confirmed.

CatsEyeXI: > I'm reaching out to some retail capture guys to try to get this detail confirmed.

do you still need any help with this? we're really interested in this functionality. i have access to retail. let me know.

GodModeXI: > > I'm reaching out to some retail capture guys to try to get this detail confirmed.
> 
> do you still need any help with this? we're really interested in this functionality. i have access to retail. let me know.

I requested details in captures discord but this hasn’t been confirmed yet. 
https://discord.com/channels/443544205206355968/443893540922064896/1201924153473507369

if you can confirm let me know and I’ll delete the request so siknoz doesn’t go for it.

WinterSolstice8: image
Must be elemental magic it seems.

WinterSolstice8: Whoops, cropped that badly, 
image

WinterSolstice8: In hindsight, this should've been clear (scholar guides tell you to close Gravitation with Noctohelix and not Bio which is much faster)

GodModeXI: > Whoops, cropped that badly, image

@WinterSolstice8, from what I am seeing your Immanence perpetuated through Bio, so it only deletes when there is a skillchain property triggered? I am updating this today and wanted to be sure were in agreement there.

WinterSolstice8: That is correct

WinterSolstice8:
Hmm. Your commits disappeared?

GodModeXI:
> Hmm. Your commits disappeared?

yes, Im an uber noob with git. I was trying to rebase to fix the commit message that was bugging the CI and accidentally nuked it. Ill redo this and reference this PR...

GodModeXI:
@WinterSolstice8 @TeoTwawki @CatsEyeXI @Xaver-DaRed @UmeboshiXI 
This has been resubmitted under PR #5188

PR #5041, opened 2024-01-29
[core] Filter outgoing 0x028 (action packet) by spawnlists

WinterSolstice8:
**_I affirm:_**

This should reduce excess action packets that go to a client who is not "aware" of certain entities (e.g. players above >32 that arent in SpawnPCList, entities that never got an 0x00E to the client to begin with and arent in SpawnMobList)
Free performance gain from not copying `SpawnIDList_t spawnlist` and using a reference instead

See spam in pic that would never be sent (client isnt aware of these PCs names and these packets are useless to the client):
image

Nothing should be obviously different without addons that would track 0x028s for entities they havent seen for some reason


Frankie-hz:
🙏

Xaver-DaRed (approved):

zach2good (approved):
LGTM 👍

PR #5044, opened 2024-01-30
[lua] [sql] Audit & fix Altepa Gate

WinterSolstice8:
**_I affirm:_**

Altepa Gate was broken and didn't reset properly. This PR fixes it and adds some QoL for players to see the state of the columns better.

Use Altepa Gate, see it reset properly, have other players see the changes from a long distance.

### DIFF
--- a/scripts/zones/Western_Altepa_Desert/npcs/_3h5.lua
+++ b/scripts/zones/Western_Altepa_Desert/npcs/_3h5.lua
@@ -13,8 +13,11 @@ end
 
 entity.onTrigger = function(player, npc)
     if npc:getAnimation() ~= xi.anim.OPEN_DOOR then
-        npc:setAnimation(xi.anim.OPEN_DOOR)
-        GetNPCByID(npc:getID() - 4):setAnimation(xi.anim.OPEN_DOOR)
+        npc:updateToEntireZone(xi.status.NORMAL, xi.anim.OPEN_DOOR)
+        local column = GetNPCByID(npc:getID() - 4)
+        if column then
+            column:updateToEntireZone(xi.status.NORMAL, xi.anim.OPEN_DOOR)
+        end
     else
         player:messageSpecial(ID.text.DOES_NOT_RESPOND)
     end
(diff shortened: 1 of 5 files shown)


Xaver-DaRed (approved):
LGTM

Out of curiosity, do any other places may require a similar method, off the top of your head?

WinterSolstice8:
> Out of curiosity, do any other places may require a similar method, of the top of your head?

Altepa gate is wholly unique afaik? not sure, not many switches that you can see from a distance to begin with.

Xaver-DaRed:
Oh well. better then.

Was gonna say that if this kind of logic could be used in other places, it may justify an npc utility, or something similar.

TeoTwawki (approved):

scripts/zones/Western_Altepa_Desert/npcs/_3h5.lua
```diff
@@ -25,9 +28,19 @@ entity.onTrigger = function(player, npc)
...
         GetNPCByID(ID.npc.ALTEPA_GATE + 8):getAnimation() == xi.anim.OPEN_DOOR
     then
-        local openTime = math.random(15, 30) * 60
+        local openTime = math.random(15, 30) * 1000 * 60
+
         for i = ID.npc.ALTEPA_GATE, ID.npc.ALTEPA_GATE + 8 do
-            GetNPCByID(i):openDoor(openTime)
+            local door = GetNPCByID(i)
+
+            -- openDoor used to (read: a decade ish ago before Sol) allow calls to doors already open and set a timer to close them
```
TeoTwawki, line 36: side note: want that functionality back :( 

likewise closeDoor used to work on already closed doors and then open them after the time ended.

cocosolos:
> LGTM
> 
> Out of curiosity, do any other places may require a similar method, off the top of your head?

Port Bastok bridge uses something similar I think.

WinterSolstice8:
> Port Bastok bridge uses something similar I think.

Yeah, I had updated that a while ago to send an update on how the bridge looks to the entire zone. It's not being done on zone in -- so it could still be improved.

Honestly, I think things like this should be inserted into the players SpawnNPCLists permanently so we don't need code-specific updates. There's a bunch of weird stuff if you're not in range to get updates. Depending on where you are, you sometimes don't see the airship's propellors for instance. I haven't tried tracking that one down because it's purely cosmetic and not functional (and doesn't alter _human_ behavior like the bridge or altepa gate would)

cocosolos:
> Honestly, I think things like this should be inserted into the players SpawnNPCLists permanently so we don't need code-specific updates.

This is probably the way to go. I looked at Altepa gate a few months ago and noticed the same thing you did where if you get close then far again, the column reverts. This is good enough for now though. 👍

PR #5045, opened 2024-01-30
[AMK, Lua] Mission 13 puzzle 1 - Beaucedine Glacier runaround

Flibe-XI:
**_I affirm:_**

This implements the first puzzle of mission 13 of AMK (see walkthrough [here](https://ffxiclopedia.fandom.com/wiki/A_Challenge!_You_Could_Be_a_Winner#)).  The goblin grenadier tells you to find pips, of which there are six, which based on the order of the vanadiel days of the week, draw a number between 0-9.  There is a odd bug that can occur where the args passed from the CS 507 are used to update the map, so the values sent through there can add erroneous pips to show up on the map.  This is solved by passing 0's through onEventUpdate, which leaves the map showing only what the player already discovered.  This function can also be used to change the map to show something else if desired... but unclear to what end at the moment.

`!addmission 10 12`
Zone into Castle Zvahl Bailey's and talk to shadowy pillar for cs
Go to lonely evergreen in beaucedine glacier for a cs
Talk to Goblin Grenadier in beaucedine glacier for puzzle instructions
Gather pips and determine answer according to wiki instructions
Return to goblin, ask for hints
Get wrong answer: chest should not open
Get correct answer: chest should open, get pocket mogbomb KI
- if you used 0 hints: 5 min flee, 1 hint: 3 min flee, 2 hints: no flee

Return to lonely evergreen for cs, which exchanges pocket mogbomb for trivia challenge kupon KI


Xaver-DaRed (approved):

scripts/missions/amk/13_A_Challenge_You_Could_Be_a_Winner.lua
```diff
@@ -17,28 +19,353 @@ mission.reward =
...
+            {
+                onTrigger = function(player, npc)
+                    return mission:progressEvent(100, 3)
+                end,
+            },
+
+            onEventFinish =
+            {
+                [100] = function(player, csid, option, npc)
+                    player:setCharVar('Mission[10][12]progress', 1)
```
zach2good, line 45: Don't you just call `mission.setVar('progress')` or something here, rather than making up the key by hand?

zach2good: Same question for all manual management and lookup of these keys

scripts/missions/amk/13_A_Challenge_You_Could_Be_a_Winner.lua
```diff
@@ -17,28 +19,353 @@ mission.reward =
...
+                onTrigger = function(player, npc)
+                    local pipSet = player:getLocalVar('Mission[10][12]pipSet') - 1
+                    local pos = npc:getPos()
+                    local element = xi.amk.helpers.pipSets[pipSet][1]
+
+                    return mission:progressEvent(
+                        510,
+                        pos.x * 1000,
+                        pos.z * 1000,
+                        pos.y * 1000,
```
zach2good, line 137: I don't think I've ever seen pos units put into an event like this, interesting!

zach2good (commented):
LGTM, just one question about vars

scripts/missions/amk/13_A_Challenge_You_Could_Be_a_Winner.lua
Flibe-XI: I was concerned about repeatability but now that I think about it this mission is only repeated while still on the same mission, never after you've completed it successfully. So I need to actually remove the >= check on mission ID, and yes can use mission:progress. Force of habit for me in amk missions

Flibe-XI: Actually scratch that, this mission is run while on the following mission.  So I'm not sure how it would work using the mission progress variable while on a different mission in case that one needed to use it. Thoughts?

zach2good: That's fine then, just leave a note saying that's whats happening

PR #5046, opened 2024-01-30
[lua] [sql] Implement Elel's spawn conditions

WinterSolstice8:
**_I affirm:_**

Implement's Elel's specific spawn conditions (night, dark/double dark weather)

One commit is salvaged from ASB, though I re-made it because it was one giant commit making a ton of NM updates that aren't compatible and/or I don't want to test right now.

See Elel only spawn at night during weather or when weather changes.

### DIFF
--- a/scripts/zones/Sacrarium/IDs.lua
+++ b/scripts/zones/Sacrarium/IDs.lua
@@ -46,7 +46,7 @@ zones[xi.zone.SACRARIUM] =
             { 16892074, { [xi.race.GALKA]    = true } },                            -- Indich (galka)
         },
         OLD_PROFESSOR_MARISELLE = 16891970,
-        ELEL                    = 16891948,
+        ELEL                    = GetFirstID('Elel'),
     },
     npc =
     {
--- a/scripts/zones/Sacrarium/IDs.lua
+++ b/scripts/zones/Sacrarium/IDs.lua
@@ -46,6 +46,7 @@ zones[xi.zone.SACRARIUM] =
             { 16892074, { [xi.race.GALKA]    = true } },                            -- Indich (galka)
         },
         OLD_PROFESSOR_MARISELLE = 16891970,
+        ELEL                    = 16891948,
     },
     npc =
     {
--- a/scripts/zones/Sacrarium/Zone.lua
+++ b/scripts/zones/Sacrarium/Zone.lua
@@ -62,15 +62,46 @@ end
 zoneObject.onEventFinish = function(player, csid, option, npc)
 end
 
-zone_object.onZoneWeatherChange = function(weather)
+zoneObject.onZoneWeatherChange = function(weather)
     local elel = GetMobByID(ID.mob.ELEL)
+    local vanadielHour = VanadielHour()
+
     if
-        not elel:isSpawned() and os.time() > elel:getLocalVar("cooldown") and
+        elel and
+        elel:getZone():getLocalVar('elelQueued') == 0 and -- Why doesn't onZoneWeatherChange contain the zone object...?
+        not elel:isSpawned() and os.time() > elel:getLocalVar('cooldown') and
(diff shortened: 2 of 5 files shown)


WinterSolstice8:
This is a little hard to test -- !setweather GLOOM and DARKNESS at night does the job. 
Any input on this would be nice, it seems to work but it feels hacky.
This kind of logic probably needs to hit King Vineragoon among others

TeoTwawki:
> This is a little hard to test -- !setweather GLOOM and DARKNESS at night does the job. Any input on this would be nice, it seems to work but it feels hacky. This kind of logic probably needs to hit King Vineragoon among others

All the weather pops would work better the way elementals work, if it wasn't for the function that enables/disables spawning ignoring their other restrictions. Same issue for night pops - disallow gets set false, and they are allowed to pop - even when their spawn type says not to. I had to do something hacky to some night only spawns recently so don't worry about it being hacky here.

Is vinegaroon's issue that he pops too soon after weather change? Because I've not seen anything reported or had an issue spawning/despawning him so that's all I can think of being off.

scripts/zones/Sacrarium/IDs.lua
```diff
@@ -46,6 +46,7 @@ zones[xi.zone.SACRARIUM] =
             { 16892074, { [xi.race.GALKA]    = true } },                            -- Indich (galka)
         },
         OLD_PROFESSOR_MARISELLE = 16891970,
+        ELEL                    = 16891948,
```
TeoTwawki: we have that handy get-by-name stuff now so that we don't need the static ID.

WinterSolstice8:
re KV;

If weather is active when it can spawn, it won't because it has to change _to_ single or double weather, checking it on zone tick in addition to this will cure that edge case. AFAIK KV can spawn any time weather is up, not just when it changes
```
zoneObject.onZoneWeatherChange = function(weather)
    local kvMob = GetMobByID(ID.mob.KING_VINEGARROON)

    if
        kvMob:getCurrentAction() == xi.act.DESPAWN and
        (weather == xi.weather.DUST_STORM or weather == xi.weather.SAND_STORM)
    then
        kvMob:spawn()
    elseif
        kvMob:getCurrentAction() == xi.act.ROAMING and
        weather ~= xi.weather.DUST_STORM and
        weather ~= xi.weather.SAND_STORM
    then
        DespawnMob(ID.mob.KING_VINEGARROON)
    end
end
```

TeoTwawki:
> If weather is active when it can spawn, it won't because it has to change to single or double weather, checking it on zone tick in addition to this will cure that edge case. AFAIK KV can spawn any time weather is up, not just when it changes

I actually thought we had something happening on game hr or game day (not ideal) to pseudo cover that, but I see we do not currently.

scripts/zones/Sacrarium/IDs.lua
WinterSolstice8: done

scripts/zones/Sacrarium/Zone.lua
```diff
@@ -62,4 +62,47 @@ end
 zoneObject.onEventFinish = function(player, csid, option, npc)
 end
 
+zoneObject.onZoneWeatherChange = function(weather)
+    local elel = GetMobByID(ID.mob.ELEL)
+    local vanadielHour = VanadielHour()
+
+    if
+        elel and
+        elel:getZone():getLocalVar('elelQueued') == 0 and -- Why doesn't onZoneWeatherChange contain the zone object...?
```
TeoTwawki, line 71: man if I had a nickel every time I ran into that and asked the same thing...

I'd have 2. which isn't a lot but annoying that it happened twice right?

TeoTwawki (approved):
lgtm, till we have better handling of interactions between spawn types and disallow functionality

EpicTaru (approved):

Xaver-DaRed (approved):

PR #5055, opened 2024-02-01
[c++] Fix some container issues

Xaver-DaRed:
**_I affirm:_**

As title sais. No real functionality was changed.

None.

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -2079,6 +2079,14 @@ void CCharEntity::OnItemFinish(CItemState& state, action_t& action)
     auto* PTarget = static_cast<CBattleEntity*>(state.GetTarget());
     auto* PItem   = state.GetItem();
 
+    if (!PItem->isType(ITEM_EQUIPMENT) && (PItem->getQuantity() < 1 || PItem->getReserve() > 0))
+    {
+        ShowWarning("OnItemFinish: %s attempted to use reserved/insufficient %s (%u).", this->getName(), PItem->getName(), PItem->getID());
+        this->pushPacket(new CMessageBasicPacket(this, this, PItem->getID(), 0, MSGBASIC_ITEM_FAILS_TO_ACTIVATE));
+
+        return;
+    }
+
     // TODO: I'm sure this is supposed to be in the action packet... (animation, message)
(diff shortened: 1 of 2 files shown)


TeoTwawki (approved):

claywar (approved):

src/map/packet_system.cpp
```diff
@@ -2540,7 +2540,18 @@ void SmallPacket0x04D(map_session_data_t* const PSession, CCharEntity* const PCh
...
 
-            if (quantity > 0 && PItem && PItem->getQuantity() >= quantity && PChar->UContainer->IsSlotEmpty(slotID))
+            if (quantity == 0 || !PItem)
+            {
+                return;
+            }
+
+            if (PItem->getQuantity() < quantity || PItem->getReserve() > 0)
+            {
+                ShowWarning("Delivery Box: %s attempted to send insufficient/reserved %u %s (%u).", PChar->getName(), quantity, PItem->getName(), PItem->getID());
```
Frankie-hz, line 2550: I believe `PChar->getName()` should be `PChar->GetName()`

src/map/entities/charentity.cpp
```diff
@@ -2079,6 +2079,14 @@ void CCharEntity::OnItemFinish(CItemState& state, action_t& action)
     auto* PTarget = static_cast<CBattleEntity*>(state.GetTarget());
     auto* PItem   = state.GetItem();
 
+    if (!PItem->isType(ITEM_EQUIPMENT) && (PItem->getQuantity() < 1 || PItem->getReserve() > 0))
+    {
+        ShowWarning("OnItemFinish: %s attempted to use reserved/insufficient %s (%u).", this->getName(), PItem->getName(), PItem->getID());
```
Frankie-hz, line 2084: I believe `this->getName()` should be `this->GetName()`

getName() looks like an item string

zach2good: We renamed entity and zone GetName's to getName, and we plan on doing the same for everything else so all member functions (and their bindings) start with lower case letters 👍

Frankie-hz: ahh ok, thank you!

PR #5057, opened 2024-02-02
[core, lua] Fix issues with healing dead players and reraise after moving players

TracentEden2:
**_I affirm:_**

The PR fixes several issues related to healing dead players and reraise:
1. In The CoP 6-4 battlefield helper, the player should only be healed if they are actually alive. Otherwise the player is dead (death animation and cannot move) but appears to have positive HP in the client
2. Relatedly, in places where the player is moved within the same zone (like between deck and airship in CoP 6-4), the player can die with RR and then be moved (after beating mammets for example), but the reraise prompt is lost and the player cannot be raised normally (because m_raiseSent is already set). This adds a method for players called allowSendRaisePrompt that resets the m_raiseSent variable so raise and reraise can be sent. This method is now called after players are moved in CoP 6-4 battlefield.
3. Similar to 1, adds logic for MH healing to only heal if the player actually has HP > 0, this prevents issues when players die while zoning into MH

1. Enter CoP 6-4 fight, die in fight and then have second character kill the mammets, when dead player is sent back to ship deck they should still look dead with zero HP
2. In CoP 6-4 fight, have one player put up RR and then die to mammets, then have second player kill mammets, when sent back to ship deck the player with RR should get the prompt again despite being moved
3. Use a poison potion with very little HP right outside MH and then run into moghouse as you die, you should be dead in MH with zero HP

### DIFF
--- a/scripts/zones/Sealions_Den/bcnms/one_to_be_feared_helper.lua
+++ b/scripts/zones/Sealions_Den/bcnms/one_to_be_feared_helper.lua
@@ -10,9 +10,18 @@ local oneToBeFeared = {}
 -- Not to be confused with an actual instance.
 
 local function healCharacter(player)
-    player:setHP(player:getMaxHP())
-    player:setMP(player:getMaxMP())
-    player:setTP(0)
+    -- only heal players if alive otherwise bugs out player
+    if player:isAlive() then
+        player:setHP(player:getMaxHP())
+        player:setMP(player:getMaxMP())
+        player:setTP(0)
+        if player:getPet() ~= nil then
+            local pet = player:getPet()
+            pet:setHP(pet:getMaxHP())
+            pet:setMP(pet:getMaxMP())
+            pet:setTP(0)
+        end
+    end
 end
 
 local function returnToAirship(player)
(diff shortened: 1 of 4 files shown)


TracentEden2:
These are fixes from ASB, I am having trouble with fully testing as having issues with my local LSB crashing, will un-draft in next day or two when I fix my install. Posting now for any comments.

TracentEden2:
> These are fixes from ASB, I am having trouble with fully testing as having issues with my local LSB crashing, will un-draft in next day or two when I fix my install. Posting now for any comments.

Ok, tested this and seems to be working.

TeoTwawki:
> Similar to 1, adds logic for MH healing to only heal if the player actually has HP > 0, this prevents issues when players die while zoning into MH

I just want to add that while possible, retail does LOT to make you not die after being inside, including stripping certain bad statuses while you are on the black screen transitioning in. so you basically have to have you hp hit zero just before the transition to get a dead body in their on retail to even see what happens. :) right now LSB doesn't and you can zone in while poisoned and die inside. That's a free pr for anyone with a retail account that wants to test which effects get removed!

TeoTwawki (approved):
looks good to me if attribution is correct

> These are fixes from ASB

are you the original author? if so we're good.

TracentEden:
> looks good to me if attribution is correct
> 
> > These are fixes from ASB
> 
> are you the original author? if so we're good.

Yep original author.

PR #5062, opened 2024-02-02
Simplify and format mobskill functions

claywar:
**_I affirm:_**

* Moves xi.mobskills.mobType to xi.mobType enum
* Simplifies conditionals to use math/utils functions when able
* Uses lookup tables for conditionals with fixed values
* Removes or reformats some comments in the file

Note: This is the first round of many to come, and more effort is required

No gameplay impact should be observed


TeoTwawki (approved):
Reviewing from mobile but looks good as far as I see

Xaver-DaRed (approved):

PR #5071, opened 2024-02-04
[cpp] Add destructors to CParty and CAlliance

WinterSolstice8:
* These should be temporary until Party/Alliance are rewritten

**_I affirm:_**

Dirty, ugly hack to prevent bad refs keeping garbage pointers in memory pointing to things that _could_ still be valid, causing mayhem

There should be no noticeable changes.

### DIFF
--- a/src/map/alliance.cpp
+++ b/src/map/alliance.cpp
@@ -66,6 +66,13 @@ CAlliance::CAlliance(uint32 id)
 {
 }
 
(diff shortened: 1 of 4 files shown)


claywar (approved):

TeoTwawki (approved):
_maybe its face can only be loved by its mother, but it gets the job done_

zach2good (approved):

PR #5074, opened 2024-02-04
[core] [lua] Fix hastrait in lua

WinterSolstice8:
**_I affirm:_**

hasTrait didn't work on non-PCs, even though previously it claimed to only work on mobs...

summon a DRG wyvern at 99,

`!exec player:printToPlayer(tostring(player:getPet():hasTrait(1)))` should return true now.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -15496,19 +15496,20 @@ void CLuaBaseEntity::instantiateMob(uint32 groupID)
 
 /************************************************************************
  *  Function: hasTrait()
- *  Purpose : Returns true if a Mob has an active trait
+ *  Purpose : Returns true if a battle entity has an active trait
  *  Example : if target:hasTrait(15) then -- Double Attack
  *  Notes   :
  ************************************************************************/
 
 bool CLuaBaseEntity::hasTrait(uint16 traitID)
 {
-    if (m_PBaseEntity->objtype != TYPE_PC)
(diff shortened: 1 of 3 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -15496,19 +15496,20 @@ void CLuaBaseEntity::instantiateMob(uint32 groupID)
...
 {
-    if (m_PBaseEntity->objtype != TYPE_PC)
+    CBattleEntity* PBattleEntity = dynamic_cast<CBattleEntity*>(m_PBaseEntity);
+    if (PBattleEntity)
     {
-        return false;
+        return PBattleEntity->hasTrait(traitID);
     }
 
-    return charutils::hasTrait(static_cast<CCharEntity*>(m_PBaseEntity), traitID);
```
TeoTwawki, line 15511: > Purpose : Returns true if a Mob has an active trait

hmm

> if (m_PBaseEntity->objtype != TYPE_PC)

hmmmm

> **char**utils

lawl

TeoTwawki (approved):

TeoTwawki:
and looks like it was only ever tried on players anyway https://github.com/search?q=repo%3ALandSandBoat%2Fserver%20hasTrait&type=code

WinterSolstice8:
Yeah I was looking at a crash dump and it happened to show a player wyvern having no traits and I went to use !exec that showed me it "didnt have the trait" so then I dug deeper and just found out I was being lied to. They have the traits

zach2good (approved):

PR #5080, opened 2024-02-04
Add NM: Copper Borer

CatsEyeXI:
**_I affirm:_**

Adds the spawn conditions, logic, and drop list of Copper Borer (Halvung).
Most of the implementation was based off of the JP Wiki and an online video. Some of the information on clopedia proved to be wrong (surprise).

https://www.youtube.com/watch?v=B8c-XYoPm24
https://wiki.ffo.jp/html/18803.html
https://ffxiclopedia.fandom.com/wiki/Copper_Borer?so=search

Copper Borer will spawn 2 hours after server init


TeoTwawki:
> Some of the information on clopedia proved to be wrong (surprise).

if referring to this:

> - Melee attacks for 300+ damage with 50+ Additional Effect: Fire damage.
> - Has Blaze Spikes that hit for 30+ Fire damage.

The damage of spikes and additional effect will change drastically comparing a modern lv99+ player to a lv 75 player back in era when that text was written. Resistance updates and just the fact the characters have meva and higher levels better gear etc. the factors involved aren't entirely sorted out in private server land, so I'm not making you try and match it I'm just saying that's why a lot of info regarding those will be useless from wiki's (across all NM). I mean as it is a lot of the damage in the repo is literally using `math.random()` right now which shouldn't have ever been the case but here we are. Annoyingly its not even calculated identically to players, and there's evidence that SE has quietly fiddled with the calculations on both players and mobs (except when sourced from the spells - that's remained consistent). I've legit got videos from 2008-2011 where the damage is clearly different another set of videos in 2023, with identical levels and gear and mob. /tableflip

scripts/zones/Halvung/mobs/Copper_Borer.lua
```diff
@@ -4,6 +4,20 @@
...
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
+    mob:addStatusEffect(xi.effect.BLAZE_SPIKES, 10, 0, 0)
+end
+
+entity.onMobSpawn = function(mob)
+    -- Mob needs to be in the correct state in order to use thermal pulse
+    mob:setAnimationSub(0)
```
TeoTwawki: shouldn't it just be 0 in pool? and then every spawn should start as that anyway.

scripts/zones/Halvung/mobs/Copper_Borer.lua
```diff
@@ -4,6 +4,20 @@
 -----------------------------------
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
+    mob:addStatusEffect(xi.effect.BLAZE_SPIKES, 10, 0, 0)
```
TeoTwawki: don't simply add the status effect unless its dispelable. also the video you linked shows 30-33 dmg not 10. The player takes less than his pet because he's wearing resistance/mdef while pet has none.

for perma spikes there is normally a function involved but if the damage never varies you may not need it. look at Sargas in attohwa chasm for an example. That script also shows how to set flags to prevent dispelling (since scripted spikes unfortunately still needed a status at the time it was written).

scripts/zones/Halvung/mobs/Copper_Borer.lua
```diff
@@ -4,6 +4,20 @@
...
+end
+
+entity.onMobSpawn = function(mob)
+    -- Mob needs to be in the correct state in order to use thermal pulse
+    mob:setAnimationSub(0)
+end
+
+entity.onAdditionalEffect = function(mob, target, damage)
+    return xi.mob.onAddEffect(mob, target, damage, xi.mob.ae.ENFIRE, { power = math.random(3, 7) })
```
TeoTwawki: what lv and gear were you? really don't want to add more randoming our resistance checking should provide variance.

Looking at the video I see the same problem as the spikes damage I commented on, the bat is taking damage in the 30's. its a little hard to catch on screwtube:
image

The circumstances of this battle were not controlled for a capture. This is why linking a 3rd parties video in youtube is almost always a waste of dev time here.

TeoTwawki (changes requested):
- Controlled retail testing of damage needed.
- verification of spikes removal/perma needed, and adjustment accordingly.

scripts/zones/Halvung/mobs/Copper_Borer.lua
TeoTwawki: additional inner workings info on how that power input gets used:
https://github.com/LandSandBoat/server/blob/base/scripts/globals/mobs.lua#L514

you will not get the exact damage of your input back, so you can't just random roll between lowest and highest points you see.

scripts/zones/Halvung/mobs/Copper_Borer.lua
CatsEyeXI: changed animation to 0 in db

scripts/zones/Halvung/mobs/Copper_Borer.lua
CatsEyeXI: moved this to AUTO_SPIKES as per your guidance

scripts/zones/Halvung/mobs/Copper_Borer.lua
```diff
@@ -4,6 +4,18 @@
 -----------------------------------
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
+    mob:setMobMod(xi.mobMod.AUTO_SPIKES, 1)
+
+    -- probably tops out at 50 dmg for a 75 player, max dmg seen as naked lv99 was 36
+    mob:addStatusEffect(xi.effect.BLAZE_SPIKES, 35, 0, 0)
```
TeoTwawki: You definitely didn't test this. `xi.mobMod.AUTO_SPIKES` is used to enabled scripted spikes inside `entity.onSpikesDamage()` which you have not added to this script.

*That isn't the part that prevents them from being dispeled.*

```lua
    mob:addStatusEffect(xi.effect.BLAZE_SPIKES, 36, 0, 0)
    mob:getStatusEffect(xi.effect.BLAZE_SPIKES):setEffectFlags(xi.effectFlag.DEATH)
```
Would be correct to simply give permanent blaze spikes. That wipes the flags and sets loss to be on death only (they'd be re-applied at respawn, so that's been the pattern ever since the DSP days)

without using scripted spikes, it will fall back on the damage math the core uses for the spell blaze spikes which likely won't turn out right. If you do scripted, it'll take the base damage as an optional parameter ("damage" in the function call is this). Typically we apply an INT diff that was never documented proper (I have no idea its origin) and then resistance to its element or debuff (if any) factors in.

We have numerous examples of scripted perma spikes, this was why I directed you to one.

~~This is what it looks like if we copy pasta existing examples and swap it to the power you saw and switch it over to blaze spikes/fire: 
[Copper_Borer.zip](https://github.com/LandSandBoat/server/files/14170757/Copper_Borer.zip)~~ See #5095 instead

TeoTwawki (changes requested):

scripts/zones/Halvung/mobs/Copper_Borer.lua
TeoTwawki: if you leave it just the effect and make it non-dispelable, you'll get 36 dmg every time

~~if you proper script it out, it becomes heavily INT dependent unless you remove the int Diff in the script - which you can.~~ See #5095 

how you handle that to ballpark it is up to you (I'm not looking for perfection)

scripts/zones/Halvung/Zone.lua
```diff
@@ -1,10 +1,14 @@
...
 -- Zone: Halvung (62)
 -----------------------------------
+local ID = zones[xi.zone.HALVUNG]
+-----------------------------------
 local zoneObject = {}
 
 zoneObject.onInitialize = function(zone)
     xi.helm.initZone(zone, xi.helm.type.MINING)
+
+    GetMobByID(ID.mob.COPPER_BORER):setRespawnTime(7200) -- 2 hours
```
TeoTwawki: if he's every 2hrs with no variation and no other conditions, just set him up as timed spawn in its mob group instead of scripted spawn. If there is variation but still "just timed" it should have a similar line in onMobDespawn in its script.

TeoTwawki (changes requested):

TeoTwawki (approved):

Xaver-DaRed (approved):

PR #5082, opened 2024-02-04
[core] Fix blood weapon against undead mobs

TracentEden2:
**_I affirm:_**

Fix blood weapon so it does not work against undead mobs.

Fight normal and undead mobs with blood weapon

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1207,7 +1207,7 @@ namespace battleutils
 
             uint8 enspell = (uint8)PAttacker->getMod(Mod::ENSPELL);
 
-            if (enspell == ENSPELL_BLOOD_WEAPON)
+            if (enspell == ENSPELL_BLOOD_WEAPON && PDefender->m_EcoSystem != ECOSYSTEM::UNDEAD)
             {
                 Action->additionalEffect = SUBEFFECT_HP_DRAIN;
                 Action->addEffectMessage = 161;


src/map/utils/battleutils.cpp
```diff
@@ -1207,7 +1207,7 @@ namespace battleutils
 
             uint8 enspell = (uint8)PAttacker->getMod(Mod::ENSPELL);
 
-            if (enspell == ENSPELL_BLOOD_WEAPON)
+            if (enspell == ENSPELL_BLOOD_WEAPON && PDefender->m_EcoSystem != ECOSYSTEM::UNDEAD)
```
TeoTwawki, line 1210: I wonder how this got forgotten the last time this section was re-written..

TeoTwawki: Actually shouldn't it just drain for zero in the log but still animate? its been a long time since I've seen anyone mistakenly use it vs undead on retail, so I'm not 100% sure.

TracentEden2: > Actually shouldn't it just drain for zero in the log but still animate? its been a long time since I've seen anyone mistakenly use it vs undead on retail, so I'm not 100% sure.

Yeah, that I am not sure about that either.

TeoTwawki: I guess I'll go check in a few

TeoTwawki: ok, really doesn't animate!

TeoTwawki (approved):

PR #5083, opened 2024-02-04
[core] Add interruption of fishing on zoning to prevent monsters from leaving pool

TracentEden2:
**_I affirm:_**

The fishing change prevents mobs being removed (until zone restart) from the catchable monster pool if the mob is being caught when the player zones (such as boat docking). This occurs because the mob's local variable hooked is set when initially hooked but is never set back to zero because the mob is neither caught nor unhooked. Thus the fix is calling the interrupt fishing method on zoning which unhooks the mob.

1. Edit zone to remove all fish and items except a single monster (like an NM)
2. Bring two characters to the zone including one blm with warp II
3. Start fishing on the other char and hook the monster then have the first character warp the fishing char while reeling in
4. Have the blm char start fishing and see that the mob is still in the pool (as opposed to being removed from the pool because it was never unhooked)

### DIFF
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -456,6 +456,12 @@ void CZoneEntities::DecreaseZoneCounter(CCharEntity* PChar)
     m_charList.erase(PChar->targid);
     charTargIds.erase(PChar->targid);
 
+    // Need to interupt fishing on zone out otherwise fished up mobs get stuck in hooked state
+    if (PChar->hookedFish && PChar->hookedFish->hooked == true)
+    {
+        fishingutils::InterruptFishing(PChar);
+    }
+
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

WinterSolstice8 (approved):

src/map/zone_entities.cpp
```diff
@@ -456,6 +456,12 @@ void CZoneEntities::DecreaseZoneCounter(CCharEntity* PChar)
     m_charList.erase(PChar->targid);
     charTargIds.erase(PChar->targid);
 
+    // Need to interupt fishing on zone out otherwise fished up mobs get stuck in hooked state
+    if (PChar->hookedFish && PChar->hookedFish->hooked == true)
```
zach2good, line 460: A problem for another day, but surely the non-nullness of a member called 'hookedFish' implies that it is indeed... Hooked?

PR #5089, opened 2024-02-05
[core] Fix barrage and shadows interaction message

TracentEden2:
**_I affirm:_**

This PR fixes an issue when mobs have shadows and the player uses barrage. This would sometimes result in the mob being reported as healed (though not actually healed), for example when barrage hit 4 shadows and the last shot hit the mob. The issue can be seen in the code snippet discussed below. Also in cases of blink where only some shots are blocked the damage from the previous hit shot would still be added even if blink blocked the next (since damage variable was never cleared and always added). 

In the code snippet [here](https://github.com/LandSandBoat/server/blob/defa464f7c1746277ad1e37fbe495c950d941135/src/map/entities/charentity.cpp#L1872) you can see that if the number of shadows taken is more than the number of real hits then the param will be negative and the message will be incorrectly set to absorb.

Fight a mob with shadows (like Tonberry Slasher) and use barrage and then RA. The shadows should absorb 4 ranged attacks and the final shot should show correct damage message (and not a healing message).

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1812,10 +1812,12 @@ void CCharEntity::OnRangedAttack(CRangeState& state, action_t& action)
                 {
                     charutils::TrySkillUP(this, (SKILLTYPE)PAmmo->getSkillType(), PTarget->GetMLevel());
                 }
+                totalDamage += damage;
             }
         }
         else // miss
         {
+            damage                  = 0;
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):
Absolutely mindboggling it wasn't written like this to begin with.

Xaver-DaRed (approved):
lgtm

PR #5092, opened 2024-02-05
[core, lua] Add utility methods for clearing AI queues (timer and action queues) from lua

TracentEden2:
**_I affirm:_**

This PR adds utility methods for clearing the AI queues (action and timer queues) of entities. These methods are useful, for example, when an entity (often NPCs) has some periodic timed action(s) that might need to be canceled if some other event occurs. 

For example, ASB uses the clearTimerQueue function for Morion Worm ??? behavior. Specifically the ??? uses an `npc:timer()` function to move every 30 minutes but if a player pops the Morion Worm then we need to need to clear the future movement action in the timer queue otherwise the timing of the 30 min repop and future movements will not align with the death of Morion Worm.

If there is a better and also simple way to achieve the same type of logical outcome for such situations, then I am also open to such suggestions.

Add a timer function (to move for example) to an npc and then use clearTimerQueue on that npc and notice that the npc never moves because the action was removed from the queue

### DIFF
--- a/src/map/ai/helpers/action_queue.cpp
+++ b/src/map/ai/helpers/action_queue.cpp
@@ -96,3 +96,19 @@ bool CAIActionQueue::isEmpty()
 {
     return actionQueue.empty() && timerQueue.empty();
 }
+
+void CAIActionQueue::clearActionQueue()
+{
+    while (!actionQueue.empty())
+    {
+        actionQueue.pop();
+    }
+}
+
+void CAIActionQueue::clearTimerQueue()
+{
+    while (!timerQueue.empty())
+    {
+        timerQueue.pop();
+    }
+}
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -17019,6 +17019,22 @@ bool CLuaBaseEntity::deleteRaisedChocobo()
     return true;
 }
 
+void CLuaBaseEntity::clearActionQueue()
+{
+    if (m_PBaseEntity->PAI)
+    {
+        m_PBaseEntity->PAI->ClearActionQueue();
+    }
+}
+
+void CLuaBaseEntity::clearTimerQueue()
+{
+    if (m_PBaseEntity->PAI)
+    {
+        m_PBaseEntity->PAI->ClearTimerQueue();
+    }
+}
+
 void CLuaBaseEntity::setMannequinPose(uint16 itemID, uint8 race, uint8 pose)
 {
     TracyZoneScoped;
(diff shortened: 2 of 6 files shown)


WinterSolstice8 (approved):
LGTM

src/map/lua/lua_baseentity.cpp
```diff
@@ -17019,6 +17019,31 @@ bool CLuaBaseEntity::deleteRaisedChocobo()
...
+
+    auto* PMob = static_cast<CMobEntity*>(m_PBaseEntity);
+
+    if (PMob)
+    {
+        PMob->PAI->ClearActionQueue();
+    }
+}
+
+void CLuaBaseEntity::clearTimerQueue()
```
claywar, line 17030: Have you tested this binding with a player being the target? Also, m_PBaseEntity will not be null for anything here, and that specific check is unnecessary

src/map/ai/helpers/action_queue.cpp
```diff
@@ -96,3 +96,17 @@ bool CAIActionQueue::isEmpty()
 {
     return actionQueue.empty() && timerQueue.empty();
 }
+
+void CAIActionQueue::clearActionQueue()
+{
+    while (!actionQueue.empty())
+        actionQueue.pop();
```
zach2good, line 104: Needs to be wrapped in {}

src/map/lua/lua_baseentity.cpp
zach2good: You can cover all your bases here by checking for the presence of m_PBaseEntity->PAI too

TracentEden2: > Have you tested this binding with a player being the target? Also, m_PBaseEntity will not be null for anything here, and that specific check is unnecessary

Thanks for comments, I tried on player and it seemed ok, thus I removed the monster only check. I also removed the m_PBaseEntity null check and added just a m_PBaseEntity->PAI check.

src/map/ai/helpers/action_queue.cpp
TracentEden2: > Needs to be wrapped in {}

Thanks for comments, I fixed it.

src/map/lua/lua_baseentity.cpp
TracentEden2: > You can cover all your bases here by checking for the presence of m_PBaseEntity->PAI too

Yep, I added that check now and removed the m_PBaseEntity check as per claywar.

zach2good (approved):

PR #5096, opened 2024-02-05
[core] Clear enmity of other mobs towards formerly charmed mobs

TracentEden2:
**_I affirm:_**

The PR fixes an issue wherein when charmed mobs become uncharmed they were not being removed from the enmity lists of other mobs.

Also for reference this has been tested to be retail accurate by Siknoz and can be found on captures discord under combat-mechanics section.

Fight with a charmed mob against another mob then leave and then recharm. The aggressive mob should go right after the player (not the recharmed mob) since the enmity list of the aggressive mob no longer has the charmed mob.

### DIFF
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -38,6 +38,7 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "latent_effect_container.h"
 #include "map.h"
 #include "mob_spell_list.h"
+#include "notoriety_container.h"
 #include "petutils.h"
 #include "puppetutils.h"
 #include "status_effect_container.h"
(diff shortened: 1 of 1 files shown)


claywar:
To clarify from the description, there are two expected behaviors:
1. If the mob loses charm without release (and is aggressive, this may need confirmation as its been over a decade since I soloed BST) it should head towards the previous master
2. If the mob is released, it should have its enmity list cleared, and then game on for any other state after a brief period

TracentEden:
> To clarify from the description, there are two expected behaviors:
> 
> 1. If the mob loses charm without release (and is aggressive, this may need confirmation as its been over a decade since I soloed BST) it should head towards the previous master
> 2. If the mob is released, it should have its enmity list cleared, and then game on for any other state after a brief period

This PR only deals with the enmity towards the charmed mob from other mobs, not the enmity of the charmed mob towards players. So does not impact any of those behaviors. I changed the title to be more clear and describe the current incorrect behavior below, hopefully that clarifies things:

Currently if a player is fighting a mob (called mob A) with a charmed mob (called mob B) and then releases the charmed mob (mob B) but then recharms it, the mob the player is fighting against (mob A) will still have full built up enmity (from originally fighting) towards the other mob (mob B).

WinterSolstice8 (approved):

Xaver-DaRed (approved):

Issue #847, opened 2021-10-21
NM's should build resistances to Gravity/Sleep/Stun over time


ShiyoKozuki:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: EDITME

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
https://www.bluegartr.com/threads/58566-Bind-Does-is-build
https://www.bluegartr.com/threads/62053-Gravity-resist-build-up-on-Zipacna
https://www.bluegartr.com/threads/68655-NM-and-HNM-Magic-resist-over-time

Gravity and sleep duration goes down with subsequence casts. 3-8% per cast?

Stun no visible difference in duration, mobs eventually just go "immune". Every mob has a different number of times it can be stunned before this happens.

The resistance "build up" should also fade over time if you don't continue to use said enfeebles.

TeoTwawki:
work on this kind of depends on us getting basic resistance correct first, or we won't even know if its working right.

PR #5099, opened 2024-02-07
Add magic enfeebling resist building (that decreases effect duration)

TracentEden2:
**_I affirm:_**

This PR adds an enfeebling resistance building system to the new magic framework. The resistance building is based on retail captures of Fafnir by Jimmayus and Kanryu (from 2019 and 2023). 

The building system decreases the final duration of the status effect by X seconds for each successful cast of a spell that adds that effect to the mob. The value X is set via a mod, so for fafnir for SLEEP this would be `mob:setMod(xi.mod.SLEEPRESBUILD, 3)` , as the retail captures show a decrease of 3 seconds. The system adds NAMERESBUILD mods for many enfeebles to allow res building to different status effects, and leverages local variables to track the amount of resistance building.

This built up resistance potentially should be cleared if the mob starts roaming or over some amount of time, but this is not verified so is not yet implemented. This system is also useful for custom private servers that want to prevent, for example, players from holding mobs asleep indefinitely.

Related to this bug report https://github.com/LandSandBoat/server/issues/847

This is coming from ASB to upstream.

Pop Fafnir, cast sleep, wait for it to wear off, the cast again and see the length decrease in duration by 3 seconds, repeat.

### DIFF
--- a/scripts/globals/spells/enfeebling_spell.lua
+++ b/scripts/globals/spells/enfeebling_spell.lua
@@ -13,94 +13,94 @@ xi.spells = xi.spells or {}
 xi.spells.enfeebling = xi.spells.enfeebling or {}
 -----------------------------------
 local pTable =
-{   --                                  1                             2           3                   4                    5      6    7         8       9    10        11         12
-    --                  [Spell ID ] = { Effect,                       Stat-Used,  Resist-Mod,         MEVA-Mod,            pBase, DoT, Duration, Resist, msg, immunity, pSaboteur, mAcc },
+{   --                                  1                             2           3                   4                        5                        6      7     8         9       10    11        12         13
+    --                  [Spell ID ] = { Effect,                       Stat-Used,  Resist-Mod,         Resist-Mod-Build,        MEVA-Mod,                pBase, DoT,  Duration, Resist, msg,  immunity, pSaboteur, mAcc },
     -- Black Magic
-    [xi.magic.spell.BIND          ] = { xi.effect.BIND,               xi.mod.INT, xi.mod.BINDRES,     xi.mod.BIND_MEVA,        0,   0,       60,      2,   0,        4, false,       0 },
-    [xi.magic.spell.BINDGA        ] = { xi.effect.BIND,               xi.mod.INT, xi.mod.BINDRES,     xi.mod.BIND_MEVA,        0,   0,       60,      2,   0,        4, false,       0 },
-    [xi.magic.spell.BLIND         ] = { xi.effect.BLINDNESS,          xi.mod.INT, xi.mod.BLINDRES,    xi.mod.BLIND_MEVA,       0,   0,      180,      2,   0,       64, true,        0 },
-    [xi.magic.spell.BLIND_II      ] = { xi.effect.BLINDNESS,          xi.mod.INT, xi.mod.BLINDRES,    xi.mod.BLIND_MEVA,       0,   0,      180,      2,   0,       64, true,        0 },
-    [xi.magic.spell.BLINDGA       ] = { xi.effect.BLINDNESS,          xi.mod.INT, xi.mod.BLINDRES,    xi.mod.BLIND_MEVA,       0,   0,      180,      2,   0,       64, true,        0 },
-    [xi.magic.spell.BREAK         ] = { xi.effect.PETRIFICATION,      xi.mod.INT, xi.mod.PETRIFYRES,  xi.mod.PETRIFY_MEVA,     1,   0,       30,      2,   0,        0, false,       0 },
-    [xi.magic.spell.BREAKGA       ] = { xi.effect.PETRIFICATION,      xi.mod.INT, xi.mod.PETRIFYRES,  xi.mod.PETRIFY_MEVA,     1,   0,       30,      2,   0,        0, false,       0 },
-    [xi.magic.spell.BURN          ] = { xi.effect.BURN,               xi.mod.INT, 0,                  0,                       0,   3,       90,      3,   1,        0, true,        0 },
-    [xi.magic.spell.CHOKE         ] = { xi.effect.CHOKE,              xi.mod.INT, 0,                  0,                       0,   3,       90,      3,   1,        0, true,        0 },
-    [xi.magic.spell.CURSE         ] = { xi.effect.CURSE_I,            xi.mod.INT, xi.mod.CURSERES,    xi.mod.CURSE_MEVA,      50,   0,      300,      2,   0,        0, false,       0 },
-    [xi.magic.spell.DISPEL        ] = { xi.effect.NONE,               xi.mod.INT, 0,                  0,                       0,   0,        0,      4,   0,        0, false,     175 },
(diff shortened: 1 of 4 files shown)


scripts/globals/spells/enfeebling_spell.lua
```diff
@@ -13,94 +13,94 @@ xi.spells = xi.spells or {}
...
-    [xi.magic.spell.SLEEPGA       ] = { xi.effect.SLEEP_I,            xi.mod.INT, xi.mod.SLEEPRES,    xi.mod.SLEEP_MEVA,       1,   0,       60,      2,   0,        1, false,       0 },
-    [xi.magic.spell.SLEEPGA_II    ] = { xi.effect.SLEEP_I,            xi.mod.INT, xi.mod.SLEEPRES,    xi.mod.SLEEP_MEVA,       2,   0,       90,      2,   0,        1, false,       0 },
-    [xi.magic.spell.STUN          ] = { xi.effect.STUN,               xi.mod.INT, xi.mod.STUNRES,     xi.mod.STUN_MEVA,        1,   0,        5,      4,   0,        8, false,     200 },
-    [xi.magic.spell.VIRUS         ] = { xi.effect.PLAGUE,             xi.mod.INT, xi.mod.VIRUSRES,    xi.mod.VIRUS_MEVA,       5,   3,       60,      2,   0,        0, false,       0 },
+    [xi.magic.spell.BIND          ] = { xi.effect.BIND,               xi.mod.INT, xi.mod.BINDRES,     xi.mod.BINDRESBUILD,     xi.mod.BIND_MEVA,        0,   0,       60,      2,   0,        4, false,       0 },
+    [xi.magic.spell.BINDGA        ] = { xi.effect.BIND,               xi.mod.INT, xi.mod.BINDRES,     xi.mod.BINDRESBUILD,     xi.mod.BIND_MEVA,        0,   0,       60,      2,   0,        4, false,       0 },
```
TeoTwawki, line 20: The width here has been bothering me for awhile now, and we've officially reached the point where my brain screams at me for seeing it. Some data is going to have to be broken out separately instead of continuing to expand this wider and wider.

TeoTwawki: @Xaver-DaRed I summon thee for planning

TracentEden2: Yeah, I can move into a separate table if wanted

scripts/zones/Dragons_Aery/mobs/Fafnir.lua
```diff
@@ -10,6 +10,7 @@ local entity = {}
 entity.onMobSpawn = function(mob)
     mob:setLocalVar('[rage]timer', 3600) -- 60 minutes
     mob:setMobMod(xi.mobMod.WEAPON_BONUS, 50) -- Level 90 + 50 = 140 Base Weapon Damage
+    mob:setMod(xi.mod.SLEEPRESBUILD, 3) -- decrease the duration of sleep spells 3 seconds for each successful sleep
```
Xaver-DaRed, line 13: If I'm understanding this correctly, just from a quick glance, this system relies on mobs having a modifier in there already, in order for it to work?

Xaver-DaRed: Its a good first effort, dont get me wrong, but current implementation would mean having to add a line of code, per status effect per mob script in order to fully introduce.

I think we need more data on how he system actually works in retail. Concretely:
- Which mobs build resistance? Is it a NM global thing or just certain mobs?
- Is the duration decresed in sets of 3 seconds in all cases or just in this concrete case and with some others is 5 or 2 seconds (all regarding the same status effect, `sleep` in this case)?

Depending on this, the implementation could vary DRASTICALLY.

scripts/globals/spells/enfeebling_spell.lua
Xaver-DaRed: I made some questions bellow. Depending on the answers, we may want a sepparate table linking an effect with a modifier. Something like...

```lua
local resistanceBuildTable =
{
    [xi.effect.SLEEP_I ] = xi.mod.SLEEP_RES_BUILD,
    [xi.effect.SLEEP_II] = xi.mod.SLEEP_RES_BUILD,
    [xi.effect.BLIND   ] = xi.mod.BLIND_RES_BUILD,
}

scripts/zones/Dragons_Aery/mobs/Fafnir.lua
TracentEden: > Its a good first effort, dont get me wrong, but current implementation would mean having to add a line of code, per status effect per mob script in order to fully introduce.
> 
> I think we need more data on how he system actually works in retail. Concretely:
> 
> * Which mobs build resistance? Is it a NM global thing or just certain mobs?
> * Is the duration decresed in sets of 3 seconds in all cases or just in this concrete case and with some others is 5 or 2 seconds (all regarding the same status effect, `sleep` in this case)?
> 
> Depending on this, the implementation could vary DRASTICALLY.

The reason for having the mod-based setup is because it appears to be limited to only a few NMs, currently we only know of Fafnir with Sleep. I will ask Jimmayus if they have more data.

TeoTwawki: tbh we should probably not try to create the system before we have a fair sized list of places it will be used

Xaver-DaRed: I've been asking around... Here is what I could gather (Take it with a BIG chunk of salt)
- Its a case by case system, where only certain mobs can build resistance against certain elements and effects.
- The decrease seems to happen after the resistance check.
- It looks like a multiplicative decrease more like a flat `seconds * uses` reduction, similar to nuke wall. The decrease can floor duration to 1 second.
- The resistance resets at about 5 minutes of not landing the enfeeble

With all of this in mind:
- Its seems a lua edit/modifier placement per mob is inevitable. This could be alleviated via a unique modifier that acts as a bitmask for each effect OR via sql. It could be placed in the resistances table as a bitmask, maybe.
- This would also mean that we could handle each resistance via editing the modifier directly OR use local variables and avoid defining 16+ new modifiers.

I will need to perform some reail testing and rumiate this over

TeoTwawki: > This could be alleviated via a unique modifier that acts as a bitmask for each effect or via sql.

or localVar on the monster entity to track it

TeoTwawki: > OR use local variables and avoid defining 16+ new modifiers.

hi I'm Teo and my hobbies include replying to Xaver before I have finished reading his posts 👋

zach2good:
Why is this in draft?

TracentEden2:
> Why is this in draft?

Ah, I left it as draft because @Xaver-DaRed said they were going to do some retail testing and thinking about it, so I expect some changes to be needed. If not then I can mark as ready.

Xaver-DaRed:
I agree. Please let it as a draft for the time being. And thank you^^

PR #5104, opened 2024-02-07
[Fix] [sql] Correct gigas socks item quantity result

Xaver-DaRed:
**_I affirm:_**

Fix Gigas socks desynth quantities.

https://www.bg-wiki.com/ffxi/Gigas_Socks

1. Desynth Gigas Socks.
2. Get proper amounts depending on HQ tier.
3. Collect player tears.
4. ?
5. Profit.


TeoTwawki (approved):
well there goes my illegitimate poison powder factory, I'll have to have all these taru workers deported /sulk

(kidding)

PR #5108, opened 2024-02-07
[core] Prevent special attacks from being impacted by speed modifiers for normal attacks

TracentEden2:
**_I affirm:_**

This PR prevents effects like slow and elegy from impacting the attack speed of mobs while they are using special attacks (`MOBMOD_ATTACK_SKILL_LIST`) like tiamat in the air. This has been verified by Siknoz on retail capture of Tiamat (and other wyrms) for example.

Cast slow on Tiamat while in the air and check the weapon delay before and after

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -38,6 +38,7 @@
 #include "items/item_weapon.h"
 #include "job_points.h"
 #include "lua/luautils.h"
+#include "mob_modifier.h"
 #include "notoriety_container.h"
 #include "packets/action.h"
 #include "recast_container.h"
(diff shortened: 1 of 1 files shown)


TeoTwawki (approved):

TeoTwawki:
I recall this being the case with Iron Giants as well, so its plausible to me its universal

Xaver-DaRed (approved):

TeoTwawki (approved):
re-approved after whatever it was I missed in my 1st review, I guess

TracentEden2:
> re-approved after whatever it was I missed in my 1st review, I guess

Thanks, yeah sorry about that, I had to add co-author as I forgot in my initial commit. I should focus more on individual PRs instead of juggling many.

WinterSolstice8 (approved):

PR #5110, opened 2024-02-08
Prevent getStorageItem() on non players inside criticalRateFromFencer()

TeoTwawki:
**_I affirm:_**

Some trusts (mainly Naji and Ferrous Coffin, see #2751) were tripping an error about getting inventory when they do not have one. because only players have inventories. This PR moves those inside the existing object type check and the other checks follow after. 

For Naji its a check Fencer. for Ferrous its a check in Aftermath handling.

1. Summon Naji before applying this change: watch log error happen.
2. Rebuild server with this change. Observe Naji no longer errors. 
3. Mock the "doorman" (optional)

### DIFF
--- a/scripts/globals/combat/physical_utilities.lua
+++ b/scripts/globals/combat/physical_utilities.lua
@@ -565,18 +565,19 @@ end
 -- Fencer: Critical hit rate bonus when actor is only wielding with main hand.
 xi.combat.physical.criticalRateFromFencer = function(actor)
     local fencerBonus = 0
+    -- TODO: do any Trusts or mobs ever get Fencer bonuses?
 
-    local mainEquip = actor:getStorageItem(0, 0, xi.slot.MAIN)
-    local subEquip  = actor:getStorageItem(0, 0, xi.slot.SUB)
-
-    if
-        actor:getObjType() == xi.objType.PC and
-        mainEquip and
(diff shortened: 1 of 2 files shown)


scripts/globals/combat/physical_utilities.lua
```diff
@@ -565,18 +565,21 @@ end
...
+    -- TODO: do any Trusts or mobs ever get Fencer bonuses?
 
     if
-        actor:getObjType() == xi.objType.PC and
-        mainEquip and
-        not mainEquip:isTwoHanded() and                                                      -- No 2 handed weapons.
-        not mainEquip:isHandToHand() and                                                     -- No 2 handed weapons.
-        (subEquip == nil or subEquip:getSkillType() == xi.skill.NONE or subEquip:isShield()) -- Only shields allowed in sub.
+        actor:getObjType() == xi.objType.PC
     then
```
Xaver-DaRed: if only 1 condition is left in this if, can it be all 1 line?

Other than that, lgtm

TeoTwawki: it sure can, I just moved it "up" and didn't think about it. Might look a little funny though!

TeoTwawki: change made 👍

Xaver-DaRed (approved):

PR #5113, opened 2024-02-08
[core] Fix timing of mob healing while roaming

TracentEden2:
**_I affirm:_**

This PR fixes the timing of mob healing (while the mob is roaming). 

Currently on LSB mob healing timing is based on a semi-random timer with a healing tick happening every 20-40 seconds (and each tick healing 10% of HP). However, this code is over 9 years old and does not seem to match the logic of current retail mobs. Also the reasoning for the current logic is difficult to find.

Recent retail captures by Siknoz suggest most mobs are based on a constant 10s healing tick timing (for healing 10% of HP), he checked across different mobs and NMs: Roc, Serket, Jorm for NMs, and Saruta/Bibiki/Zitah/Garlaige/Sauromugue/Ule). This PR replicates this timing.

Though there are exceptions, Siknoz checked some 99 cap zones and some are like 10% for 10s then 10% for 20s then back to 10% for 10s then 15%  for 20s. However even having the flat 10s timing of this PR is more accurate than the current system.

This is a fix from ASB moving upstream.

Fight a mob to some low HP, place a second character in the zone so zone does not sleep, then !goto yourself to clear hate, and watch the mob heal 10% every 10 seconds.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -211,7 +211,7 @@ void CMobController::TryLink()
                 if (PPartyMember->m_roamFlags & ROAMFLAG_IGNORE)
                 {
                     // force into attack action
-                    //#TODO
+                    // #TODO
                     PPartyMember->PAI->Engage(PTarget->targid);
                 }
             }
@@ -831,7 +831,7 @@ void CMobController::DoRoamTick(time_point tick)
 
         return;
     }
-    //#TODO
+    // #TODO
     else if (PMob->GetDespawnTime() > time_point::min() && PMob->GetDespawnTime() < m_Tick)
     {
         Despawn();
@@ -842,6 +842,26 @@ void CMobController::DoRoamTick(time_point tick)
     {
         // don't claim me if I ignore
         PMob->m_OwnerID.clean();
+    }
+
+    if (m_Tick >= m_mobHealTime + 10s && PMob->getMobMod(MOBMOD_NO_REST) == 0 && PMob->CanRest())
+    {
+        // recover 10% health and lose tp
+        if (PMob->Rest(0.1f))
+        {
+            // health updated
+            PMob->updatemask |= UPDATE_HP;
+        }
+
+        if (PMob->GetHPP() == 100)
+        {
+            // at max health undirty exp
+            PMob->m_HiPCLvl     = 0;
+            PMob->m_HiPartySize = 0;
+            PMob->m_giveExp     = true;
+            PMob->m_UsedSkillIds.clear();
+        }
+        m_mobHealTime = m_Tick;
     }
 
     // skip roaming if waiting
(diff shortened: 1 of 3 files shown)


TeoTwawki:
> Currently on LSB mob healing timing is based on a semi-random timer with a healing tick happening every 20-40 seconds 

Looks to me like is based on a value stored in a mobmod (which there is [a default value of 20 seconds for](https://github.com/LandSandBoat/server/blob/342aacb020b11179b434b77de15623f3e22436fd/src/map/utils/mobutils.cpp#L950)) and last action time. Basing it on the roaming cooldown seems incorrect but I'm not sure what else that mobmod may be involved in though. 

`else if (m_Tick >= m_LastActionTime + std::chrono::milliseconds(PMob->getBigMobMod(MOBMOD_ROAM_COOL)))`

if this turns out to be the only thing that is used for, we can rename the mobmod to be more obvious what its for at a glance and adjust its default value and its usage. That leaves you able to easily override or even adjust it on the fly be it in core or script, though I am not opposed to phasing it out in favor of a better system. If it is actually used for other things, those things should probably be separated from sharing a mobmod used in mob healing.A cursory glance loosk like (to me) its used ina  bunch of things that make no sense to use it in, like the de-aggro delay and pathing ticks. reeealy hope I'm misunderstanding those.

TeoTwawki:
> Steps to test these changes
Fight a mob to some low HP, then !goto yourself, and watch the mob heal 10% every 10 seconds.

you also want to make sure zone does not go empty (can use a 2nd char) when doing this, or the mob -may- full heal instantly.

TracentEden2:
> > Currently on LSB mob healing timing is based on a semi-random timer with a healing tick happening every 20-40 seconds
> 
> Looks to me like is based on a value stored in a mobmod (which there is [a default value of 20 seconds for](https://github.com/LandSandBoat/server/blob/342aacb020b11179b434b77de15623f3e22436fd/src/map/utils/mobutils.cpp#L950)) and last action time. Basing it on the roaming cooldown seems incorrect but I'm not sure what else that mobmod may be involved in though.
> 
> `else if (m_Tick >= m_LastActionTime + std::chrono::milliseconds(PMob->getBigMobMod(MOBMOD_ROAM_COOL)))`
> 
> if this turns out to be the only thing that is used for, we can rename the mobmod to be more obvious what its for at a glance and adjust its default value and its usage. That leaves you able to easily override or even adjust it on the fly be it in core or script, though I am not opposed to phasing it out in favor of a better system. If it is actually used for other things, those things should probably be separated from sharing a mobmod used in mob healing.A cursory glance loosk like (to me) its used ina bunch of things that make no sense to use it in, like the de-aggro delay and pathing ticks. reeealy hope I'm misunderstanding those.

Yeah, the logic is somewhat difficult to parse but it seems to have to do with how often the mob moves around (so pathing) and actions like buffing itself. So it does seem to do quite a bit, thus I am hesitant to repurpose it at the moment.

TracentEden2:
> > Steps to test these changes
> > Fight a mob to some low HP, then !goto yourself, and watch the mob heal 10% every 10 seconds.
> 
> you also want to make sure zone does not go empty (can use a 2nd char) when doing this, or the mob -may- full heal instantly.

Thanks, yeah I will add that instruction.

zach2good:
Why is this in draft?

WinterSolstice8 (approved):
seems reasonable

TeoTwawki (approved):
approval contingent of making a tracking issue for the details discussed in pr regarding variable timing and (now former) strange use of modmob, so that we don't forget to improve on this later down the road

Xaver-DaRed (approved):

TeoTwawki:
> approval contingent of making a tracking issue for the details discussed in pr regarding variable timing and (now former) strange use of modmob, so that we don't forget to improve on this later down the road

WHAR MUH TRAKKIN?? /sulk

PR #5115, opened 2024-02-09
[core] Fix enmity for utsusemi shadow absorb and initial hit

TracentEden2:
**_I affirm:_**

This fixes two small enmity issues. Firstly, the initial hit should give 200 CE and 900 VE (see [here](https://kanican.livejournal.com/21727.html)) however LSB currently only gives the 200 CE. Secondly, losing a Utsusemi shadow should give the player -25 CE (see same link above),however LSB currently does not give any CE.

This PR fixes these issues by adding the 900 VE and by adding a parameter to the IsAbsorbByShadow function that specifies the attacker and if the attacker is a mob then give -25 CE.

Pull a mob with, for example, a ranged attack that misses and check the initial enmity (via !getenmity). Cast Utsusemi and monitor enmity (via !getenmity) as you lose shadows. Can also test scenarios where there should be no enmity loss like losing blink shadows or ninja sange.

### DIFF
--- a/src/map/enmity_container.cpp
+++ b/src/map/enmity_container.cpp
@@ -204,6 +204,7 @@ void CEnmityContainer::UpdateEnmity(CBattleEntity* PEntity, int32 CE, int32 VE, 
         if (initial)
         {
             CE += 200;
+            VE += 900;
         }
 
         float bonus = CalculateEnmityBonus(PEntity);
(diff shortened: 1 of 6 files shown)


TeoTwawki:
While I don't think its a problem in this instance, I just want to general warn to watch the dates on enmity info of something that has produced "gotcha"s in the past. SE has changed many things including amounts on specific actions over the years, so data from 2004, 2008, 2024, etc, may not all align. But I believe this one is still valid.

WinterSolstice8 (approved):
Seems reasonable

Xaver-DaRed (approved):

PR #5117, opened 2024-02-10
[core] Fix underflow and fishing skill issue for RACC and RATT

TracentEden2:
**_I affirm:_**

This PR fixes two issues:

1. Currently the RACC and RATT functions in BattleEntity will use fishing skill to calculate RACC and RATT if the player has a fishing rod equipped. This causes issues, for example, with /checkparam. This PR adds a check to prevent this condition.

2. Secondly, currently RACC and RATT can underflow and also cause issues. This PR changes some of the variables used in the calculation from uint to int and then uses max to prevent the underflow. Also applies this max check to ACC function.

1. Equip a fishing rod with some amount of fishing skill and then use /checkparams and see correct RACC and RATT
2. With no RACC or RATT add a blind status effect and then use /checkparams and see correct RACC and RATT

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -712,7 +712,8 @@ uint16 CBattleEntity::ATT()
     {
         ATT += this->GetSkill(SKILL_AUTOMATON_MELEE);
     }
-    return ATT + (ATT * ATTP / 100) + std::min<int16>((ATT * m_modStat[Mod::FOOD_ATTP] / 100), m_modStat[Mod::FOOD_ATT_CAP]);
+    // use max to prevent underflow
+    return std::max(1, ATT + (ATT * ATTP / 100) + std::min<int16>((ATT * m_modStat[Mod::FOOD_ATTP] / 100), m_modStat[Mod::FOOD_ATT_CAP]));
 }
 
 uint16 CBattleEntity::RATT(uint8 skill, uint16 bonusSkill)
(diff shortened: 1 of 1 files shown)


TeoTwawki:
> will use fishing skill to calculate RACC and RATT if the player has a

hilarity

TeoTwawki:
I almost wanted to say we should check specifically for throwing/archery/marksmanship rather than for not being fishing, but I don't think anything else can ever get into the slot unless there is an unpatched exploit I am not aware of to pop a sword or k-club into there.

Seems we are also having a similar issue where BRD songs are trusting the skill of the slot used without checking which skill it is also.

src/map/entities/battleentity.cpp
```diff
@@ -712,7 +712,8 @@ uint16 CBattleEntity::ATT()
     {
         ATT += this->GetSkill(SKILL_AUTOMATON_MELEE);
     }
-    return ATT + (ATT * ATTP / 100) + std::min<int16>((ATT * m_modStat[Mod::FOOD_ATTP] / 100), m_modStat[Mod::FOOD_ATT_CAP]);
+    // use max to prevent underflow
+    return std::max(0, ATT + (ATT * ATTP / 100) + std::min<int16>((ATT * m_modStat[Mod::FOOD_ATTP] / 100), m_modStat[Mod::FOOD_ATT_CAP]));
```
WinterSolstice8: Shouldn't these clamp to a lower bounds of `1` to avoid divisions by zero?

TracentEden2: > Shouldn't these clamp to a lower bounds of `1` to avoid divisions by zero?

I am not sure, I took the zero lower bound from the clamping and max of other existing stats like ACC and DEF. I am not sure if on retail you can get to zero?

TracentEden2: Ok, after discussion with others, I set the lower bound of def and eva to 1 and the lower bound of acc, racc, att, and ratt to 0 (as these can be zero according to for example checkparams).

WinterSolstice8: /checkparam can report 0 ratt/racc if you have no ranged weapon, that's probably why. I don't think it can ever report 0 for your melee weapon (Specifically to avoid dividing by zero)

Xaver-DaRed (approved):

src/map/entities/battleentity.cpp
TracentEden: > /checkparam can report 0 ratt/racc if you have no ranged weapon, that's probably why. I don't think it can ever report 0 for your melee weapon (Specifically to avoid dividing by zero)

Yep, I have changed ATT to have lower bound of 1.

WinterSolstice8 (approved):

PR #5119, opened 2024-02-10
Improve functionality of Amphiptere mobs

CatsEyeXI:
**_I affirm:_**

Aims to correct certain inconsistencies in various Amphiptere mobs.
Sets up basic functionality to NM: Zirnitra
Adds missing mob skill scripts
Sets the correct animationSub for this family in mob_pools so they don't spawn grounded (they should default to in flight). 

!zone Xarcabard_S
Run forward just a bit until you encouter an Amphiptere mob, it should be flying now
You aggro it and engage, !tp 3000 several times to see the new TP moves.

!spawnmob 17338585 (Zirnitra)
!gotoid 17338585
You should be able to aggro and fight it.
Added some of the basic mechanics to this fight, including the "reaving wind knockback aura".
Mob will also print the "disappointed" message when the target wipes, however, I left a small TODO on this.

**Sources**
Silknoz' fantastic captures, both at 75 and 99: 
https://youtu.be/P0FQF4Sr4x0
https://youtu.be/DPP_ofLX6Z8
https://www.youtube.com/watch?v=P-nMxDG7XoQ
http://ffxidb.com/zones/137/zirnitra
https://www.bg-wiki.com/ffxi/Zirnitra
https://ffxiclopedia.fandom.com/wiki/Zirnitra


CatsEyeXI:
https://github.com/CatsAndBoats/catseyexi/issues/311

scripts/zones/Xarcabard_[S]/mobs/Zirnitra.lua
```diff
@@ -2,9 +2,40 @@
...
+    mob:setLocalVar('hasEngaged', 1)
+end
+
+entity.onMobRoam = function(mob)
+    -- Based on video capture, mob prints this message if the party wipes.
+    -- TODO: Message below should not originate from Zirnitra herself
+    if mob:getLocalVar('hasEngaged') == 1 then
+        mob:setLocalVar('hasEngaged', 0)
+        mob:showText(mob, ID.text.THE_MONSTER_FADES)
+        DespawnMob(mob:getID())
```
Comment (TeoTwawki) [inline scripts/zones/Xarcabard_[S]/mobs/Zirnitra.lua:29]: keep in mind I don't know this NM well and haven't looked it up yet, so this is generalized: Unless it despawns _immediately_ you probably  don't want to force despawn by script. We've a mobmod for handling this normally. You'd then have to move the message which I assume is seen by anyone nearby not just the party fighting it?

scripts/mixins/families/amphiptere.lua
```diff
@@ -23,6 +23,34 @@ g_mixins.families.amphiptere = function(amphiptereMob)
...
+        if actionId == reavingWind then
+            mobArg:setAnimationSub(2)
+            -- Zirnitra spams a knockback while aura is active
+            mobArg:setLocalVar('auraEndTime', os.time() + 20)
+        end
+    end)
+
+    amphiptereMob:addListener('WEAPONSKILL_STATE_EXIT', 'SPAM_KNOCKBACK', function(mobArg, actionId)
+        local reavingWind   = 2431
+        local reavingWindKb = 2426
```
TeoTwawki, line 40: the knockback being a hidden skill makes sense given the way knockback packets are made (atom0s gave me a rundown of it a few yrs back). I assume that ID was captured.

scripts/zones/Xarcabard_[S]/mobs/Zirnitra.lua
Comment (CatsEyeXI) [reply scripts/zones/Xarcabard_[S]/mobs/Zirnitra.lua:29]: Yeah the message would probably be a better fit under something like printToArea, but wouldn't use the exact message ID, which I felt more important to preserve. I looked through the available text functions for a closer fit, and seems we are a bit restricted with what we can do with the current tooling, happy to be proven wrong though :)
https://github.com/LandSandBoat/server/blob/base/src/map/lua/lua_baseentity.h#L53

Comment (TeoTwawki) [reply scripts/zones/Xarcabard_[S]/mobs/Zirnitra.lua:29]: we wouldn't want printToArea in upstream, but one could custom it that way at their own server.

`showText()` does do an area message as long as the base entity (part before the `:` ) is not a player.  If its seen by bystanders, that's what I mean by area message. If its only seen by the party, might be a diff message packet than what showText does and I could possibly help research that.

If the base entity is a player and that 1st parameter is an NPC it also forces the entity to face the player.

And it uses same IDs as `messageSpecial()`

There are a few other message types that may or may not share the same ID list for the messages, Some of which we even have implementations for. I have no idea which one is retail accurate but it'd be in the packets coming to client so we could find it that way.

Xaver-DaRed (approved):
This will conflict with your onMobEngage PR. One of them will probably need a rebase after the fact + an update for the new mob in here

WinterSolstice8 (approved):

scripts/actions/mobskills/feral_peck.lua
```diff
@@ -0,0 +1,32 @@
...
+-----------------------------------
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local currentHP = target:getHP()
+    local damage    = target:getHP()
```
TeoTwawki: 2 variables of the same thing. you used the 2nd one only (tho the 1st ones name makes more sense) which is causing CI to complain

> scripts/actions/mobskills/feral_peck.lua:17:11: (W211) unused variable currentHP

CatsEyeXI: sorry man, fix incoming

WinterSolstice8 (approved):

TeoTwawki:
I rebased for you

CatsEyeXI:
> I rebased for you

Thanks

PR #5120, opened 2024-02-10
[core] Convert utils.cpp "match" function to use regex for matching

WinterSolstice8:
**_I affirm:_**

Does what it says on the tin, convert utils.cpp "match" function to use regex for matching.
Also fixes a windows-only crash when using `!gotoname Rock.*` in La Theine 

Use `!gotoname Rock.*` in La Theine, don't crash.

### DIFF
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -829,14 +829,10 @@ void rtrim(std::string& s)
     // clang-format on
 }
 
-// Returns true if the given str matches the given pattern.
+// Returns true if the given str matches the given pattern using standard regex
 bool matches(std::string const& target, std::string const& pattern)
 {
-    if (std::regex_match(target, std::regex(pattern)))
-    {
-        return true;
-    }
-    return false;
+    return std::regex_match(target, std::regex(pattern));
 }
 
 bool starts_with(std::string const& target, std::string const& pattern)
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -830,60 +830,13 @@ void rtrim(std::string& s)
 }
 
 // Returns true if the given str matches the given pattern.
-// Wildcards can be used in the pattern to match "any character"
-// e.g: %anto% matches Shantotto or Canto-Ranto
-// Modification of https://www.geeksforgeeks.org/wildcard-character-matching/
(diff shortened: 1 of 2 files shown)


JRepoArchive:
Tested. +1

src/common/utils.cpp
```diff
@@ -832,58 +832,15 @@ void rtrim(std::string& s)
 // Returns true if the given str matches the given pattern.
 // Wildcards can be used in the pattern to match "any character"
 // e.g: %anto% matches Shantotto or Canto-Ranto
-// Modification of https://www.geeksforgeeks.org/wildcard-character-matching/
 bool matches(std::string const& target, std::string const& pattern, std::string const& wildcard)
 {
-    auto matchesRecur = [&](const char* target, const char* pattern, const char* wildcard, auto&& matchesRecur)
+    // Replace our wildcard with ".*", perform standard regex match.
```
zach2good: I'm in for the use of regex, but I think I'd prefer it if we used it instead of a nested replacement to swap it for the very standardized and understood `.*`. If people can't figure out .*, they shouldn't be on a computer.

WinterSolstice8: Updated to pure regex 
image

src/common/utils.cpp
```diff
@@ -830,60 +830,13 @@ void rtrim(std::string& s)
...
-            return matchesRecur(target + 1, pattern, wildcard, matchesRecur) || matchesRecur(target, pattern + 1, wildcard, matchesRecur);
-        }
-
-        return false;
-    };
-
-    return matchesRecur(target.c_str(), pattern.c_str(), wildcard.c_str(), matchesRecur);
+        return true;
+    }
+    return false;
```
zach2good: `return std::regex_match(target, std::regex(pattern));`

src/common/utils.cpp
```diff
@@ -830,60 +830,13 @@ void rtrim(std::string& s)
 }
 
 // Returns true if the given str matches the given pattern.
```
zach2good: This could probably do with being told that it matches a standard regex pattern

zach2good (approved):

PR #5121, opened 2024-02-10
Removes invalid target args from onMobRoam/onMobDisengage in scripts

CatsEyeXI:
**_I affirm:_**

Removes invalid arguments from the onMobRoam and onMobDisengage function of mob scripts. This got me a few times while working on other stuff.

I'm not sure any testing is actually required here, please let me know if otherwise.
```
map/lua/luautils.cpp:        sol::function onMobDisengage = getEntityCachedFunction(PMob, "onMobDisengage");
map/lua/luautils.cpp:        sol::function onMobRoam = getEntityCachedFunction(PMob, "onMobRoam");
```

### DIFF
--- a/scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
+++ b/scripts/zones/AlTaieu/mobs/Jailer_of_Prudence.lua
@@ -54,7 +54,7 @@ entity.onMobSpawn = function(mob)
     mob:addMod(xi.mod.LULLABY_MEVA, 30)
 end
 
-entity.onMobDisengage = function(mob, target)
+entity.onMobDisengage = function(mob)
 end
 
 entity.onMobDeath = function(mob, player, optParams)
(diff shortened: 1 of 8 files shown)


TeoTwawki (approved):
copy pasta scripting strikes again

I've seen people just copy a file with all possible functions every time they make a new one, and often they have the args wrong like this - then everything they make does too!

TeoTwawki:
p.s. does it bother anyone else that one of these is written past tense and the other present tense?
- onMobEngage**d**
- onMobDisengage

CatsEyeXI:
> p.s. does it bother anyone else that one of these is written past tense and the other present tense?
> 
> * onMobEngage**d**
> * onMobDisengage

yes this bothers me a lot and messes me up a lot as well.. they should definitely be consistent

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #5123, opened 2024-02-10
[core] Make Enmity Cap configurable via map settings.

JRepoArchive:
**_I affirm:_**

Instead of hardcoding the enmity cap, allows the setting to be both mutable and configurable.

Change the `ENMITY_CAP` setting in `settings/map.lua` then set multiple character's enmity values to the limit +1 and observe that they cannot receive any more enmity through actions.

### DIFF
--- a/settings/default/map.lua
+++ b/settings/default/map.lua
@@ -51,6 +51,12 @@ xi.settings.map =
     -- If this is the case, consider using the ah_pagination module
     AH_LIST_LIMIT = 7,
 
+    -- The total enmity cap for a given entity on the enmity table.
+    -- 30,000 is believed to be approximately current retail cap.
+    -- This directly affects a tank's ability to hold enmity over time.
+    -- The lower the value, the faster damage dealers will reach the cap and the mob will bounce.
+    ENMITY_CAP = 30000,
+
     -- Misc EXP related settings
     EXP_RATE                = 1.0,
     EXP_LOSS_RATE           = 1.0,
--- a/src/map/enmity_container.cpp
+++ b/src/map/enmity_container.cpp
@@ -41,7 +41,8 @@ along with this program.  If not, see http://www.gnu.org/licenses/
  ************************************************************************/
 
 CEnmityContainer::CEnmityContainer(CMobEntity* holder)
-: m_EnmityHolder(holder)
+: EnmityCap{ settings::get<int32>("map.ENMITY_CAP") }
+, m_EnmityHolder(holder)
 {
 }
 
--- a/src/map/enmity_container.h
+++ b/src/map/enmity_container.h
@@ -39,10 +39,10 @@ struct EnmityObject_t
 
 typedef std::unordered_map<uint32, EnmityObject_t> EnmityList_t;
 
-constexpr int32 EnmityCap = 30000;
-
 class CEnmityContainer
 {
+    int32 EnmityCap;
+
 public:
     CEnmityContainer(CMobEntity* holder);
     ~CEnmityContainer();


settings/default/map.lua
```diff
@@ -51,6 +51,9 @@ xi.settings.map =
     -- If this is the case, consider using the ah_pagination module
     AH_LIST_LIMIT = 7,
 
+    -- The total enmity cap
+    ENMITY_CAP = 30000,
```
zach2good, line 58: Presumably 30k is the retail value, but what's the use case here? What does this enable downstream?

zach2good (approved):

settings/default/map.lua
zach2good: I'd probably expand the comment to explain what this does in practice, like letting a mob "lock on harder" to a tank, etc.

Xaver-DaRed: For whatever reason, probably because we dont handle enmity as retail does, enmity gets capped really fast. You can cap it in under 3 minutes of fighting, meaning tanking jobs become useless (when it comes to holding a mob attention)

i would guess raising this cap allows for that to happen later down the line, making tank jobs remain usefull more time.

ghost: There is no verifiable data regarding enmity caps, however, it's anecdotally thought that older eras had lower caps.  

In fact, an article released in 2013 suggests that Enmity generation is dependent on the type of content you're undertaking. It's been anecdotally observed that enmity works vastly differently in SoA+ content vs older zones.  Master Trials are the most notable evidence. To support this, the enmity system would require several key changes which are out of the scope of intentions with this change.

An article regarding Enmity changes which came with SoA+. 
[Enmity System Explanation and Planned Adjustments via Akihiko Matsui](https://forum.square-enix.com/ffxi/threads/30629-Enmity-System-Explanation-and-Planned-Adjustments?p=402769&viewfull=1#post402769)



This enables downstream consumers to essentially make it what they believe it was. As it doesn't impact the functionality in any way, I think there's a strong case for configuration.

ghost: Recent push expands the commentary in the configuration generously. Closing this conversation.

Xaver-DaRed: Leave the conversation unresolved until Zach resolves it himself, since he made the question in the first place.

Gonna also open an issue linking to this PR and that article aswell. Seems like good info.

Xaver-DaRed (approved):

settings/default/map.lua
zach2good: Resolved 😆

TeoTwawki: > For whatever reason, probably because we dont handle enmity as retail does, enmity gets capped really fast. You can cap it in under 3 minutes of fighting, meaning tanking jobs become useless (when it comes to holding a mob attention)

Incorrect. Thats not a result of us capping incorrectly or handling incorrectly, most of what we do was pretty verified before retail changed it a 2nd or 3rd time.

That was the case on retail itself with the original enmity handling, caps and amounts generate by spells/actions, and pld was dead weight in abyssea because of everyone having massive buffs but there being nothing that could buff pld's hate holding since looking at a mob sideways would cap hate instantly at the 1st second. We don't have that happening but we have had a LOT of past bugs (now fixed afaik) that would result in incorrect reset/underflows/overflows/doubling of CE VE or both. Its been a bumpy ride and still is for servers running old code but right now on current LSB its not so bad if your team isn't idiots. (even on modern retail, a DD can easily peel hate with low effort till getting hit to reduce their hate again)

> i would guess raising this cap allows for that to happen later down the line, making tank jobs remain usefull more time.

Also because SE has changed caps twice already. Used to be a 10k cap, then 30k, and now nobody is sure.

TeoTwawki (approved):

Issue #4758, opened 2023-11-21
🔨 migration of mob_spawn_mods.sql table to scripts


TeoTwawki:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature
This table can almost be done away with if the modifiers are set on those mobs onMobInitialize and onMobSpawn as appropriate.

There are a very few modifiers that will not work post-server boot (max HP adjustments come to mind). Migrating all the ones that **do** work will make those easier to identify and adjust.

I lean towards using addMod in the mobs init - only happens once, and then repsawns copy that initial record, at the cost of not being able to re-edit it without restarting your map server. There is no benefit to using onmobSpawn over it unless you are testing and adjusting a newly built mob and want to edit it 30 odd times.

Remember addMod adds to whatever the monster already has, while setMod _overwrites the original value entirely._

Xaver-DaRed:
This may be a stupid question, but if I'm understanding this correctly...
Would this mean that, if I were to use addMod inside onMobSpawn, for instance, lets say add 5 STR, would it be adding 5 STR EACH spawn, making it cumulative?

TeoTwawki:
> This may be a stupid question, but if I'm understanding this correctly... Would this mean that, if I were to use addMod inside onMobSpawn, for instance, lets say add 5 STR, would it be adding 5 STR EACH spawn, making it cumulative?

it would execute each spawn but would not accumulate because that 5str is lost right before the function executes again.. ..Except for a few times in the past where that got bugged, which was what trained me to avoid it.

Its kinda wasteful but you'd prolly not even notice, even if it were every entity game wide.

MowFord:
closed by https://github.com/LandSandBoat/server/pull/6759

PR #5126, opened 2024-02-10
Replace modIDs in mob_spawn_mods.sql with vars

CatsEyeXI:
**_I affirm:_**

Replaces the mod ids in mob_spawn_mods.sql with sql variables.
This may help if I decide to take on the the task #4758 

Import the file into database.


TeoTwawki:
I had previously wraned you not to do this for mods. I guess I should have warned you away from mobmods too - this file is slated for destruction, and we'd be adding more locations to maintain an enumeration by doing this. Also your commit and title say mod ids but these are mobmod ids,

PR #5129, opened 2024-02-11
[core] Fix issue with dual wield var set to true with shield

TracentEden2:
**_I affirm:_**

This PR fixes an issue whereby for players the m_dualWield variable in battleentity is set to true when equipping a shield in the sub slot. The fix is to add a check to make sure that the item being equipped in the sub slot is a weapon before setting m_dualWield to true.

Equip a shield and use the existing Lua function (`player:isDualWielding()`) and see it return false

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -2063,13 +2063,17 @@ namespace charutils
                             case SKILL_KATANA:
                             case SKILL_CLUB:
                             {
-                                if (PItem->isType(ITEM_WEAPON) &&
-                                    (!charutils::hasTrait(PChar, TRAIT_DUAL_WIELD) || ((CItemWeapon*)PItem)->getSkillType() == SKILL_NONE))
+                                bool isWeapon = PItem->isType(ITEM_WEAPON);
+                                if (isWeapon && (!charutils::hasTrait(PChar, TRAIT_DUAL_WIELD) || ((CItemWeapon*)PItem)->getSkillType() == SKILL_NONE))
                                 {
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

src/map/utils/charutils.cpp
```diff
@@ -2063,13 +2063,17 @@ namespace charutils
                             case SKILL_KATANA:
                             case SKILL_CLUB:
                             {
-                                if (PItem->isType(ITEM_WEAPON) &&
-                                    (!charutils::hasTrait(PChar, TRAIT_DUAL_WIELD) || ((CItemWeapon*)PItem)->getSkillType() == SKILL_NONE))
+                                bool isWeapon = PItem->isType(ITEM_WEAPON);
+                                if (isWeapon && (!charutils::hasTrait(PChar, TRAIT_DUAL_WIELD) || ((CItemWeapon*)PItem)->getSkillType() == SKILL_NONE))
```
TeoTwawki, line 2067: This shouldn't be any different because `isType()` should already be returning truthy, and then the 2nd change you do below to shouldn't get hit because we should have already returned false here. Neither change should have been necessary which lead me to believe there is a further underlying problem, perhaps with the cast going on in `isType()`

TeoTwawki: nope nope, I'm reading this wrong, it has to be weapon in both places - I was reading it like "not weapon" in the 1st check. never mind me, blond moment.

TeoTwawki (approved):

PR #5131, opened 2024-02-11
[core, sql] Add damage types (such as slashing) to auto attacks of pets

TracentEden2:
**_I affirm:_**

This PR add the appropriate damage type (slashing, blunt, etc.) to auto-attacks of pets. As this varies for different pets a column is added to the petlist.sql called damageType and then the loader extracts this and sets the pets weapon to this damage type.

The types and sources for the different pet auto-attack damage types are below:

- Avatars: slashing (Jimmayus retail testing, see his testing spreadsheet)
- Jug Pets: some blunt and some slashing (SE [update](http://forum.square-enix.com/ffxi/threads/45365-Dec-10-2014-%28JST%29-Version-Update) and [ffxi wiki](https://ffxiclopedia.fandom.com/wiki/Category:Familiars))
- Wyvern: slashing ([bg-wiki](https://www.bg-wiki.com/ffxi/Wyvern_(Dragoon_Pet)))
- Pup frames: some slashing, some blunt, and piecing for RA of Sharpshot ([bg-wiki](https://www.bg-wiki.com/ffxi/Automaton))
- Luopan: unknown so kept as none (same as without this PR)

Find a mob weak to for example slashing and then parse pet (for example wyvern) damage both with and without this PR.

### DIFF
--- a/src/map/utils/petutils.h
+++ b/src/map/utils/petutils.h
@@ -82,8 +82,9 @@ struct Pet_t
     float HPscale; // HP boost percentage
     float MPscale; // MP boost percentage
 
-    uint16 cmbDelay;
-    uint8  speed;
+    uint16      cmbDelay;
+    DAMAGE_TYPE m_dmgType;
+    uint8       speed;
     // stat ranks
     uint8 strRank;
     uint8 dexRank;
@@ -144,6 +145,7 @@ struct Pet_t
     , HPscale(0.f)
     , MPscale(0.f)
     , cmbDelay(0)
+    , m_dmgType(DAMAGE_TYPE::NONE)
     , speed(0)
     , strRank(0)
     , dexRank(0)
--- a/sql/pet_list.sql
+++ b/sql/pet_list.sql
@@ -18,6 +18,7 @@ CREATE TABLE IF NOT EXISTS `pet_list` (
   `maxLevel` tinyint(2) unsigned NOT NULL DEFAULT '0',
   `time` int(10) unsigned NOT NULL DEFAULT '0',
   `element` tinyint(3) unsigned NOT NULL DEFAULT '0',
+  `damageType` tinyint(3) unsigned NOT NULL DEFAULT '0',
   PRIMARY KEY (`petid`)
 ) ENGINE=Aria TRANSACTIONAL=0 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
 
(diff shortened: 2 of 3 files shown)


TeoTwawki (approved):
In general practice this is correct, however there are some caveats to this I have found. SE spaghetti code seemingly doesn't have these always "count as" the same type in all circumstances. Its...weird. There is a specific battle in which pet melee get treated as if its magic, and another where they are typeless physical (no bonus to a mob that should have taken extra damage from blunt).

So SE seems to have a method that detects the source of the damage and then does additional things with it. In the meantime, what this pr does is as close as we get. 

We do have a field for this in pools (round about, its a dummy "weapon type" that the core then _assumes_ a damage type from e.g. all swords be slashing)  that I want to eventually retire, so I'm in favor of exactly what you did - adding it to pet list.

Other LSB peeps will chime in but that's my option.

TeoTwawki:
BTW I kid you not I was checking this just the last few weeks.

Related factoid: zero of the trusts you'd think are blunt damage did blunt damage.

TracentEden2:
> In general practice this is correct, however there are some caveats to this I have found. SE spaghetti code seemingly doesn't have these always "count as" the same time in all circumstances. Its...weird. There is a specific battle in which pet melee get treated as if its magic, and another where they are typeless physical (no bonus to a mob that should have taken extra damage from blunt).
> 
> So SE seems to have a method that detects the source of the damage and then does additional things with it. In the meantime, what this pr does is as close as we get.
> 
> We do have a field for this in pools (round about, its a dummy "weapon type" that the core then _assumes_ a damage type from e.g. all swords be slashing) that I want to eventually retire, so I'm in favor of exactly what you did - adding it to pet list.
> 
> Other LSB peeps will chime in but that's my option.

Ah, yeah interesting, the further you look the more craziness and exceptions to be found it seems :)

src/map/utils/petutils.h
```diff
@@ -144,6 +145,7 @@ struct Pet_t
     , HPscale(0.f)
     , MPscale(0.f)
     , cmbDelay(0)
+    , m_dmgType(DAMAGE_TYPE::NONE)
```
WinterSolstice8, line 148: CI is complaining that this needs to come before `cmbDelay(0)`

TracentEden2: > CI is complaining that this needs to come before `cmbDelay(0)`

Thanks should be fixed now.

Xaver-DaRed (approved):

PR #5139, opened 2024-02-13
[core] Fix several pet issues with charming and moving

TracentEden2:
**_I affirm:_**

This PR fixes several small pet issues:
1. Fixes an issue where charmed mobs would die if the player died. With this PR the mobs now just revert to normal mobs (like if the player used leave).
3. Fixes an issue where if the pet decharms while attacking another mob the pet would not attack the player and if far from home it would despawn (see a more detailed explanation at https://github.com/AirSkyBoat/AirSkyBoat/pull/2969). This PR fixes by setting the player as battle target right away when adding the enmity.
5. Fixes an issue whereby pets with a status effect such as sleep would still move around and follow their master. This PR adds a check (with `PPet->PAI->CanChangeState()`) to prevent this.

These are all ASB fixes coming upstream.

1. Charm a mob, then do !hp 0 to yourself and see that the mob does not die
2. Charm a mob and have it fight another mob, wait until it decharms due to charm timer being up, watch mob start to attack the player rather than just stand there or despawn.
3. Call wyvern, add sleep status effect to wyvern, walk away and watch wyvern stay put until it wakes up.

### DIFF
--- a/src/map/ai/controllers/pet_controller.cpp
+++ b/src/map/ai/controllers/pet_controller.cpp
@@ -31,7 +31,7 @@ CPetController::CPetController(CPetEntity* _PPet)
 : CMobController(_PPet)
 , PPet(_PPet)
 {
-    //#TODO: this probably will have to depend on pet type (automaton does WS on its own..)
+    // #TODO: this probably will have to depend on pet type (automaton does WS on its own..)
     SetWeaponSkillEnabled(false);
 }
 
@@ -50,9 +50,15 @@ void CPetController::Tick(time_point tick)
 
 void CPetController::DoRoamTick(time_point tick)
 {
-    if ((PPet->PMaster == nullptr || PPet->PMaster->isDead()) && PPet->isAlive())
+    if ((PPet->PMaster == nullptr || PPet->PMaster->isDead()) && PPet->isAlive() && PPet->objtype != TYPE_MOB)
     {
         PPet->Die();
+        return;
+    }
+
+    // if pet cannot change state (for example because pet is asleep) then just return
+    if (!PPet->PAI->CanChangeState())
+    {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):

src/map/ai/controllers/pet_controller.cpp
```diff
@@ -50,12 +50,18 @@ void CPetController::Tick(time_point tick)
 
 void CPetController::DoRoamTick(time_point tick)
 {
-    if ((PPet->PMaster == nullptr || PPet->PMaster->isDead()) && PPet->isAlive())
+    if ((PPet->PMaster == nullptr || PPet->PMaster->id == 0 || PPet->PMaster->name == "" || PPet->PMaster->isDead()) && PPet->isAlive() && PPet->objtype != TYPE_MOB)
```
zach2good: In what situations can we possibly get here where the id is 0 or the name is empty? This is a very very bad patch that sweeps a much bigger seeming problem under the rug? Can you explain when this happens, and why we aren't fixing it at the source?

zach2good (changes requested):

src/map/ai/controllers/pet_controller.cpp
TracentEden: > In what situations can we possibly get here where the id is 0 or the name is empty? This is a very very bad patch that sweeps a much bigger seeming problem under the rug? Can you explain when this happens, and why we aren't fixing it at the source?

Yep after talking with shozo I have removed that part from the PR, thanks for pointing out!

zach2good:
`[Fix several small pet issues]`

Please describe what is happening in the git commit. Messages like this aren't helpful to future readers of the git history.

TracentEden2:
> `[Fix several small pet issues]`
> 
> Please describe what is happening in the git commit. Messages like this aren't helpful to future readers of the git history.

Ok, I updated the commit message, thanks.

WinterSolstice8 (approved):

zach2good (approved):

PR #5143, opened 2024-02-15
[SQL][Additional Effects][Items] Item mods 17604-18729

helixhamin:
**_I affirm:_**

Adds additional effects to items 17604-18729. Also corrects some proc chances for thrown items.

> Proc rate is 100%. Damage remains consistent even with NMs, MDT applies (tested jorm and dmg was reduced by 40%), adding MAB/stats does not increase damage at all.

As GM, !additem 17604|17615|17632|17634|17646|17651|17769|17778|17779|18096|18210|18625|18698|18699|18700|18729
Equip item and hit things with it, making sure the additional effects proc at the given percentage.


TeoTwawki:
> Proc rate is 100%. Damage remains consistent even with NMs, MDT applies (tested Jorm and dmg was reduced by 40%), adding MAB/stats does not increase damage at all.

1. Quoting what someone tells you is not sourcing. What was the sample size, the target mobs and how much each contributed to the sampling, any raw parse or video, evidence of any kind and hopefully anything to show it was done on a retail server in the modern era...
2. I assume this is about the bomb arm type items you've set back to 95 I wanted some source for last time. 
3. 95 is not 100. And these definitely aren't 100. 
4. I just blew 50k each on 2 chars on bomb arms and grenades at the NPC in the metals works, to check it my dang self. Definitely not 100% rate, but higher than 5 for sure. looks to be level corrected (this is why 2 characters!) so assuming your test person genuinely believed it was 100% base don what they saw, even on jormy, and knowing jormy is level 95, I figured my higher level character is going to need something much stronger. So now I've sacrificed blood and accolades as well as gil by facing off on a high level unity mob to throw low level ammo in its face. It was not happy - did unmentionable things to my spleen. on my lower level character I got about a 75% rate just off bees in south gusta. on my higher level character I had to take on a high end UNM to even get any non proc at all. I think that nails down that level correction indeed happens..

The 95 you put stands, but I gotta make some changes to the global. Please hang tight while I do (and while my spleen heals)

helixhamin:
Sorry for not directly specifying. The test data came from the testing community. Links can be passed for reference, although I believe you have received the link already.

TeoTwawki:
side note: there is no way I will believe dsp's 5% per level after tonight, so level correction itself needs correction (it was carried forward when this was globalized)

> Sorry for not directly specifying. The test data came from the testing community. Links can be passed for reference, although I believe you have received the link already.

yep spoke with Siknoz and I have his vid now, getting additional data points to figure out how far lv correction reaches

TeoTwawki:
hey @helixhamin, I've made the changes I discussed with you in the global and there are 2 new modifiers in play.

1. we no longer just assume all ranged items will be int/mnd affected, a modifier controls this. I applied it to the existing bolts and arrows that deal elemental damage.
2. we no longer assume DSPs static 5% per level difference proc rate adjustment. A modifier now sets the amount to adjust by
3. Both of these modifiers can be applied or omitted on any item as needed

For now, when I added the mod for lv correction I kept the existing 5% just making it future-adjustable. We'll need to look at what the proc rates really are at several levels before deciding exactly what values to use for base rate and per-level adjusting. The function itself should probably change again in the future to use this more as some sort of value on a curve rather than a hard % but thats a problem for another day after a lot of math and samples.

Given what I and Siknoz were seeing, I would suggest using a definitely way less than 5 on those throwing items, and no int/mnd adjuster while keeping the very high base proc rate. 

My 1st hit didn't proc, but most of the rest did and I had to have quite a large lv diff to really notice.

image

For arrows/bolts, follow the examples I created.

PR #5148, opened 2024-02-16
[core] Prevent relic weapon extra damage procs on additional hits (such as DA)

TracentEden2:
**_I affirm:_**

This PR fixes an issue whereby relic weapons incorrectly had a chance to proc extra damage (2-3x depending on the relic) on every hit (on a triple attack for example). Wikis (see [here](https://www.bg-wiki.com/ffxi/Mandau_(Level_75))) and I think retail testing (by Siknoz) shows that relic weapons should only be able to proc on the first hit of a given attack round. 

The PR also moves the extra damage proc for barrage to the total barrage damage rather than on each barrage hit. This gives the same estimated total damage over time but when it does proc it shows large spikes (and more regular barrages) as seen on retail.

Use a relic weapon, set the extra damage proc rate at 100%, set player to have 100% double attack or triple attack, watch that relic procs only occur on the first hit of an attack round.

Use a relic bow, set the extra damage proc rate at like 10%, and see the large spikes followed by normal barrages (as compared to previously with more frequent medium damage barrages).

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1794,15 +1794,6 @@ void CCharEntity::OnRangedAttack(CRangeState& state, action_t& action)
 
                 if (slot == SLOT_RANGED)
                 {
-                    if (state.IsRapidShot())
-                    {
-                        damage = attackutils::CheckForDamageMultiplier(this, PItem, damage, PHYSICAL_ATTACK_TYPE::RAPID_SHOT, SLOT_RANGED);
-                    }
-                    else
-                    {
-                        damage = attackutils::CheckForDamageMultiplier(this, PItem, damage, PHYSICAL_ATTACK_TYPE::RANGED, SLOT_RANGED);
-                    }
-
                     if (PItem != nullptr)
                     {
                         charutils::TrySkillUP(this, (SKILLTYPE)PItem->getSkillType(), PTarget->GetMLevel());
@@ -1866,6 +1857,11 @@ void CCharEntity::OnRangedAttack(CRangeState& state, action_t& action)
             actionTarget.speceffect = SPECEFFECT::CRITICAL_HIT;
         }
 
+        if (slot == SLOT_RANGED)
+        {
+            auto attackType = (state.IsRapidShot()) ? PHYSICAL_ATTACK_TYPE::RAPID_SHOT : PHYSICAL_ATTACK_TYPE::RANGED;
+            totalDamage     = attackutils::CheckForDamageMultiplier(this, PItem, totalDamage, attackType, true);
+        }
         actionTarget.param =
             battleutils::TakePhysicalDamage(this, PTarget, PHYSICAL_ATTACK_TYPE::RANGED, totalDamage, false, slot, realHits, nullptr, true, true);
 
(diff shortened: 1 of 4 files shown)


WinterSolstice8 (approved):

Xaver-DaRed:
I see you didnt mark the "I tested this" box. Oversight when removing the draft status?

src/map/entities/charentity.cpp
```diff
@@ -1866,6 +1857,17 @@ void CCharEntity::OnRangedAttack(CRangeState& state, action_t& action)
...
+        if (slot == SLOT_RANGED)
+        {
+            if (state.IsRapidShot())
+            {
+                totalDamage = attackutils::CheckForDamageMultiplier(this, PItem, totalDamage, PHYSICAL_ATTACK_TYPE::RAPID_SHOT, SLOT_RANGED, true);
+            }
+            else
+            {
+                totalDamage = attackutils::CheckForDamageMultiplier(this, PItem, totalDamage, PHYSICAL_ATTACK_TYPE::RANGED, SLOT_RANGED, true);
+            }
```
zach2good:
```cpp
auto attackType = (state.IsRapidShot()) ? PHYSICAL_ATTACK_TYPE::RAPID_SHOT : PHYSICAL_ATTACK_TYPE::RANGED;
totalDamage     = attackutils::CheckForDamageMultiplier(this, PItem, totalDamage, attackType , true);
```

TracentEden2: > ```c++
> auto attackType = (state.IsRapidShot()) ? PHYSICAL_ATTACK_TYPE::RAPID_SHOT : PHYSICAL_ATTACK_TYPE::RANGED;
> totalDamage     = attackutils::CheckForDamageMultiplier(this, PItem, totalDamage, attackType , true);
> ```

Thanks for the suggestion, I updated the PR.

Xaver-DaRed (approved):

src/map/utils/attackutils.cpp
```diff
@@ -294,25 +294,28 @@ namespace attackutils
         float occ_extra_dmg        = battleutils::GetScaledItemModifier(PChar, PWeapon, Mod::OCC_DO_EXTRA_DMG) / 100.f;
         int16 occ_extra_dmg_chance = battleutils::GetScaledItemModifier(PChar, PWeapon, Mod::EXTRA_DMG_CHANCE) / 10;
 
-        if (occ_extra_dmg > 3.f && occ_extra_dmg_chance > 0 && xirand::GetRandomNumber(100) <= occ_extra_dmg_chance)
+        if (allowRelicProc)
```
TeoTwawki: we should use a different name for this bool, as its entirely plausible for some other item to someday have the same behavior and be controller in the same ,manner in the same spot.

TracentEden2: > we should use a different name for this bool, as its entirely plausible for some other item to someday have the same behavior and be controller in the same ,manner in the same spot.

I changed the bool to `allowRelicOccDamageProc`, does that work? Or do you mean something else?

TeoTwawki: I was leaning toward just not having the word "relic" in there, make it about what it does rather than what item(s) its for.

TeoTwawki: How do you feel about just allowProc or even allowOccDmgProc ? my point had been we might someday have something not a relic in that block.

PR #5150, opened 2024-02-17
Setup spawn and basic fight mechanics for Pallas in Vunkerl Inlet

CatsEyeXI:
**_I affirm:_**

Defines placeholders and spawn conditions for Pallas (S).
Pallas will spawn his 3 pets shortly after spawning.
He will periodically recall them during battle.
When he uses Familiar this will wake all of his pets and grant them a sleep resistance buff.
Logic and resistance data built off of the following sources:
Silknawz:  https://www.youtube.com/watch?v=UyW5XffUXJQ
Clopedia: https://ffxiclopedia.fandom.com/wiki/Pallas

TODO: 

- Attack power will need to be properly measured by a naked level 99 to acquire a proper baseline. Clopedia states "Power Attack can deal at least 1100 points of damage." but this was not verifiable.
- I do not believe that table `mob_pets` is setup for multiple pets. Only one entry was added. This was sufficient to allow Pallas to use Familiar regardless of whether or not the specified pet was alive.

!spawnmob 17117250
!gotoid 17117250


scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallass_Tiger.lua
```diff
@@ -0,0 +1,22 @@
+-----------------------------------
+-- Area: Vunkerl Inlet [S]
+--  Mob: Pallas's Tiger
+-----------------------------------
+local ID = zones[xi.zone.VUNKERL_INLET_S]
+-----------------------------------
+local entity = {}
+
+entity.onMobSpawn = function(mob)
+    mob:setMod(xi.mod.LULLABYRES, 100)
```
Comment (TeoTwawki) [inline scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallass_Tiger.lua]: that's not how we make things immune, which looks like what you want since you set that clear at 100.

immunity is a bit field in mob pools (add up numbers of statuses from the immunity enum that you want and stick resulting number in the pool field)

if they are just highly resistant..I'd still super doubt that should be 100

scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallas.lua
```diff
@@ -2,9 +2,111 @@
...
+                    petId:delStatusEffect(xi.effect.SLEEP_I)
+                    petId:delStatusEffect(xi.effect.SLEEP_II)
+                    petId:delStatusEffect(xi.effect.LULLABY)
+                    petId:addMod(xi.mod.SLEEPRES, 80)
+                end
+            end
+        end
+    end)
+
+    callTigerPet(mob)
```
Comment (TeoTwawki) [inline scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallas.lua:83]: pretty sure his idle pet spawning should be in roam instead of spawn (1st tick after spawn it'll happen, then if for some reason they are missing it'll re-happen as he roams)

just because thats how every other mobs with pets anywhere on retail has always worked.

scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallas.lua
```diff
@@ -2,9 +2,111 @@
...
+        isBusy = true
+    end
+
+    if
+        mob:actionQueueEmpty() and
+        not isBusy
+    then
+        if petsSpawned ~= 1 then
+            -- Begin casting animation
+            mob:timer(1000, function(mobArg)
```
Comment (TeoTwawki) [inline scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallas.lua:29]: I have not dug through packets to check. I started to and honestly bailed out with a migraine. but I really doubt this should happen via timer and forced animations instead of some actual action by the mob like other mobs with pets are doing. The problem with reusing what other mobs do is they only get 1 pet each, so I know we can't do that exactly, but I think that functionality is going to need improved by someone with the know-how for that so that you can use it.

Comment (CatsEyeXI) [reply scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallas.lua:29]: that's fair, will reach out in the morning to see if we can determine the fate of this PR. feel better.

Comment (WinterSolstice8) [reply scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallas.lua:29]: pet summoning is really jank sometimes and it can literally just be an injected action packet with no "real" action behind it and the pets just come out

scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallass_Tiger.lua
Comment (TeoTwawki) [reply scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallass_Tiger.lua]: sooo I didn't look into what he should have before and I apologize for that, I just saw that 100 res and guessed at what you intended without checking if what you intended was actually right...

from your wiki link:

> Will build up resistance against Sleep and Lullaby, but Repose seems to work very well.

This implies he is not actually immune or even really heavily resistant, its only saying he _build resistance to it_ when it lands which is a mechanic a lot of NM have

scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallas.lua
Comment (TeoTwawki) [reply scripts/zones/Vunkerl_Inlet_[S]/mobs/Pallas.lua:29]: > pet summoning is really jank sometimes and it can literally just be an injected action packet with no "real" action behind it and the pets just come out

even if that's what we're doing timer-ing it is not a long term solution we should retain/repeat for these few multi-pet mobs, so at best this is a "todo" so I had him open an issue about it. And double uh-uh to nesting or multiple timer calls to do ..anything really.

CatsEyeXI:
Seems like this might be a little too hacked together for merge?  I'm ok with it getting closed out if we're at an impasse. Thanks 🙏

PR #5159, opened 2024-02-18
Fix gilfinder to distribute gil equally to party

TracentEden2:
**_I affirm:_**

This PR fixes the gil distribution to party members when one or more members has a gilfinder trait. Currently, each player with gilfinder trait gets extra gil relative to their gilfinder level (while players without gilfinder get no extra gil). However retail testing (can also see [JP wiki](https://wiki.ffo.jp/html/1677.html)) shows that each player (even those without gilfinder) should get extra gil relative to the highest gilfinder level across the entire party (thus all party members should get equal gil).

Go to a zone with two players (one with gilfinder and one without) in a party and kill beastmen. Notice that each player gets an equal amuont of gilfinder-boosted gil. Also can kill them with players with different gilfinder levels and notice that the highest level applies to both players.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -3957,12 +3957,23 @@ namespace charutils
             // all members might not be in range
             if (!members.empty())
             {
-                // distribute gil
-                int32 gilPerPerson = static_cast<int32>(gil / members.size());
-                for (auto* PMember : members)
+                // Check for highest gilfinder tier
+                uint16 gilFinderActive = 0;
+
+                for (auto PMember : members)
                 {
-                    // Check for gilfinder
-                    gilPerPerson += gilPerPerson * PMember->getMod(Mod::GILFINDER) / 100;
+                    if (PMember->getMod(Mod::GILFINDER) > gilFinderActive)
+                    {
+                        gilFinderActive = PMember->getMod(Mod::GILFINDER);
+                    }
+                }
+
+                // Calculate gil for each party member.
+                uint32 gilPerPerson = static_cast<uint32>(gil / members.size());
+                gilPerPerson        = gilPerPerson * (100 + gilFinderActive) / 100;
+
+                for (auto PMember : members)
+                {
                     UpdateItem(PMember, LOC_INVENTORY, 0, gilPerPerson);
                     PMember->pushPacket(new CMessageBasicPacket(PMember, PMember, gilPerPerson, 0, 565));
                 }


src/map/utils/charutils.cpp
```diff
@@ -3957,14 +3957,32 @@ namespace charutils
...
+                    {
+                        int32 memberGil = gilPerPerson * (100 + gilFinderActive) / 100;
+                        UpdateItem(PMember, LOC_INVENTORY, 0, memberGil);
+                        PMember->pushPacket(new CMessageBasicPacket(PMember, PMember, memberGil, 0, 565));
+                    }
+                    else
+                    {
+                        UpdateItem(PMember, LOC_INVENTORY, 0, gilPerPerson);
+                        PMember->pushPacket(new CMessageBasicPacket(PMember, PMember, gilPerPerson, 0, 565));
+                    }
```
Xaver-DaRed: Why not just
```c++
                // Check for highest gilfinder tier
                uint16 gilFinderActive = 0;

                for (auto PMember : members)
                {
                    if (PMember->getMod(Mod::GILFINDER) > gilFinderActive)
                    {
                        gilFinderActive = PMember->getMod(Mod::GILFINDER);
                    }
                }

                // Calculate gil for each party member.
                uint32 gilPerPerson  = static_cast<uint32>(gil / members.size());
                gilPerPerson         = gilPerPerson * (100 + gilFinderActive) / 100;

                // Distribute gil.
                for (auto PMember : members)
                {
                    UpdateItem(PMember, LOC_INVENTORY, 0, gilPerPerson);
                    PMember->pushPacket(new CMessageBasicPacket(PMember, PMember, gilPerPerson, 0, 565));   
                }
            }
        }
```

Xaver-DaRed: I mean, theres no need to do the math per party member if every party member is supposed to get the same amount.

I also changed the data types, couse there is no case where any of those numbers should ever be negative.

TracentEden2: > I mean, theres no need to do the math per party member if every party member is supposed to get the same amount

Yeah, I modified the code now to avoid those unneeded calculations, thanks!

src/map/utils/charutils.cpp
```diff
@@ -3957,12 +3957,24 @@ namespace charutils
...
             {
-                // distribute gil
-                int32 gilPerPerson = static_cast<int32>(gil / members.size());
-                for (auto* PMember : members)
+                // Check for highest gilfinder tier
+                uint16 gilFinderActive = 0;
+
+                for (auto PMember : members)
+                {
+                    // Check for highest gilfinder tier
```
Xaver-DaRed: Duplicated comment. Rest looks good to me.

TracentEden2: > Duplicated comment. Rest looks good to me.

Thanks, fixed.

Xaver-DaRed (approved):

WinterSolstice8:
This reminds me to check if Gilfinder+ leaves a "permanent" mark on the mob like TH so items with gilfinder+ can actually be useful without swapping into them as the mob dies...

WinterSolstice8 (approved):

Issue #5160, opened 2024-02-18
🐛 Packet Audit - SmallPacket0x0FA


atom0s:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Windows 11

## Branch affected by issue 

`base`

## Additional Information

The furniture placement packet can be used to intentionally crash the server in multiple different ways due to a lack of validation on all of the incoming information. The packet handler also has little to no validation and error handling on any potential pointer usages it makes use of.

## Crash 1 - Invalid Item Pointer

The newer gardening abuse check happens before the `PItem` pointer is null-checked. This allows malformed packets to be sent to the server and cause a crash. 

## Crash 2 - Invalid `slotID` _(Causes crash 1.)_

The `slotID` is not validated to be within the legal range of the given container based on the clients actual unlocked storage space. It is not validated beyond a global maximum for any container either, allowing the client to send invalid slot indexes in this packet. This will cause 'Crash 1' to happen when sending bogus slot values.

## Bad / Incomplete Abuse Check

The abuse check to prevent 2nd-floor gardening pots is incomplete/incorrect in actually preventing that kind of situation from happening. Instead, it is currently setup to be a post-abuse catch. It is not checking the incoming parameters from the client to determine if they are trying to place the item onto the 2nd floor at the time of the packet usage. In its current setup, it will only catch those trying to move an item from the 2nd floor after its already been placed.

## Missing Validation On Client Input

There is no validation happening on any of the client parameters for this packet allowing the client to do malicious things and bypass potential placement limitations by using invalid `col` and `row` values. 

## Example Addon

Here is an example addon that can be used with Ashita v4 to quickly test this kind of crash/exploit abuse:

```lua
local packet = [[
FA 08 00 00 4C 00
FF
01 00 00 00 16 00 00 00 00
]];

local pkt = packet:gsub('[\r\n ]', ''):parts(2):imap(function (v) return tonumber(v, 16); end);
[...]

PR #5161, opened 2024-02-18
[core] Add more validation to SmallPacket0x0FA

zach2good:
**_I affirm:_**

Starts to address the points in https://github.com/LandSandBoat/server/issues/5160.

_I'm literally travelling internationally and I can't properly research, implement, or test this. But it's out in the open now and I have to do _something_ or someone is going to get their server's cereal pissed in.

Also sneaking in my RATE_LIMIT helper, because it was in my git tree and why not 🤷 

Use the "here's how you crash a server" scripts that were posted in the target issue.

### DIFF
--- a/src/common/utils.h
+++ b/src/common/utils.h
@@ -140,10 +140,10 @@ static mutex_guarded<std::unordered_map<std::string, time_point>> lastExecutionT
         {                                                                             \
             lastExecutionTimes[key] = currentTime;                                    \
             {                                                                         \
-                code;                                                                 \
+                code;                                                                  \
             }                                                                         \
         }                                                                             \
-    })                                                                                \
+    });                                                                               \
 }
 // clang-format on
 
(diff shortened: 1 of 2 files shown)


zach2good:
Now tested with the script in https://github.com/LandSandBoat/server/issues/5160

Also tested against:
- Placing a flower pot in the corners of every possible area on 1st and 2nd floors (treating downstairs as 3x blocks, and upstairs as 1x block)
- Bookshelf + things on it
- Flags on the walls in all the different slots that allow it
- Checking the "on 2nd floor" flag properly
- Also tested `RATE_LIMIT`, which should be how we handle logging as a response from any client input: the client shouldn't be able to trigger logging and allocations on the server willy-nilly, so rate limiting is one of the ways we can combat that.

This level of validation is what we need for every incoming packet. It's A LOT of work.

claywar (approved):

PR #5162, opened 2024-02-18
[core] Use bcrypt instead of MariaDB PASSWORD() + automatic migration

zach2good:
**_I affirm:_**

- Stops using Maria's `PASSWORD()` for hashing and storage in the db. Instead uses bcrypt.
- Offers an automatic migration path for users. New accounts will start using bcrypt right away, users logging in or changing their PW will be migrated to the new system seamlessly.
- Update CI in the one place where we make a fake account to login

:star: another PR while I'm packed into a train under the channel tunnel

### DIFF
--- a/src/login/auth_session.cpp
+++ b/src/login/auth_session.cpp
@@ -22,6 +22,33 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "auth_session.h"
 
 #include "common/socket.h" // for ref<T>
+#include "common/utils.h"
+
+#include <bcrypt/BCrypt.hpp>
+
+namespace
+{
+    constexpr int currentYear()
+    {
+        // clang-format off
+        constexpr auto charToInt = [](char c)
+        {
+            return (c >= '0' && c <= '9') ? (c - '0') : 0;
+        };
+        // clang-format on
+
+        constexpr const char* timestamp       = __TIMESTAMP__;             // "__TIMESTAMP__" is like "Mon Jan  2 15:04:05 2006"
+        constexpr size_t      timestampLength = sizeof(__TIMESTAMP__) - 1; // Length of the timestamp string
+
+        int year = 0;
+        for (size_t i = timestampLength - 4; i < timestampLength; ++i)
+        {
+            year = year * 10 + charToInt(timestamp[i]);
+        }
+
+        return year;
+    }
+} // namespace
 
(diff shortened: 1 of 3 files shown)


zach2good:
Testing:

1:
- Create character before changes, login
- Apply changes, rebuild
- Log in, see hash change to bcrypt hash in the db

2:
- With changes applied
- Create new character
- See new character's PW hash in db is bcrypt style

3:
- Roll back changes, rebuild, make new character, see they have old-style hash
- Change their password with the old system, see hash changes
- Apply changes, rebuild
- Do password reset flow
- Log in
- See hash is bcrypt style

zach2good:
^ For each of these, get your PW wrong and you should be left in your original state, not in some floating state where you need admin intervention

src/login/auth_session.cpp
```diff
@@ -309,13 +365,68 @@ void auth_session::read_func()
...
+                if (ret != SQL_ERROR && sql->NumRows() != 0 && sql->NextRow() == SQL_SUCCESS)
+                {
+                    return sql->GetStringData(0);
+                }
+                return "";
+            }();
+            // clang-format on
+
+            // OLD PASSWORD HASH MIGRATION
+            // TODO: Remove me after 2023-12-31
```
TeoTwawki: did you mean 2024?

zach2good: Yessir

src/login/auth_session.cpp
```diff
@@ -138,9 +140,65 @@ void auth_session::read_func()
...
+                if (ret != SQL_ERROR && sql->NumRows() != 0 && sql->NextRow() == SQL_SUCCESS)
+                {
+                    return sql->GetStringData(0);
+                }
+                return "";
+            }();
+            // clang-format on
+
+            // OLD PASSWORD HASH MIGRATION
+            if (std::string(__DATE__) == "Dec 31 2024")
```
WinterSolstice8: Wouldn't this only trip if your build date was specifically december 31st 2024 and not any days after?

zach2good: Quite right, I've added a constexpr check and a static_assert to blow things up if this section is still here when 2025 starts. Tested by rolling the target year to 2023 and it stops compiling producing the help message

Xaver-DaRed (approved):

PR #5163, opened 2024-02-18
Begin enumerating some of magic numbers in core

TeoTwawki:
**_I affirm:_**

Brings death to magic numbers. ⚔️ 
- Updated documentation
- Started new enum for system type messages
- Cleaned up message basic enum a little bit (CANT vs CANNOT)
- Use enum for CMessageSystemPacket calls

`!exec player:messageSystem(id here)`

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -1137,7 +1137,7 @@ void SmallPacket0x01A(map_session_data_t* const PSession, CCharEntity* const PCh
             else
             {
                 // You don't have any gysahl greens
-                PChar->pushPacket(new CMessageSystemPacket(4545, 0, 39));
+                PChar->pushPacket(new CMessageSystemPacket(4545, 0, MSGSYSTEM_YOU_DONT_HAVE_ANY));
             }
         }
         break;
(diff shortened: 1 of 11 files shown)


Xaver-DaRed (approved):
CI wants you to align some of the IDs.

But the PR is awesome. Death to magic numbers.

TeoTwawki:
somehow posted comment to wrong pr

src/map/packets/message_system.h
```diff
@@ -26,6 +26,53 @@
 
 #include "basic.h"
 
+// Found in ROM/27/76.dat
+// clang-format off
+enum MSGSYSTEM_ID : uint16
```
zach2good: `enum class` for type safety

src/map/packets/message_system.h
```diff
@@ -26,6 +26,53 @@
 
 #include "basic.h"
 
+// Found in ROM/27/76.dat
```
zach2good, line 29: You could add a comment here saying the contents of this DAT are dumped into `MessageSystemIDs.log`

src/map/packets/message_system.h
```diff
@@ -26,6 +26,53 @@
...
+// clang-format off
+enum MSGSYSTEM_ID : uint16
+{
+    // MSGSYSTEM_COULD_NOT_ENTER                  = 0,   // You could not enter the next area. (duplicate)
+    // MSGSYSTEM_COULD_NOT_ENTER                  = 1,   // You could not enter the next area. (duplicate)
+    MSGSYSTEM_COULD_NOT_ENTER                  = 2,   // You could not enter the next area.
+    // MSGSYSTEM_COULD_NOT_ENTER                  = 3,   // You could not enter the next area. (duplicate)
+    // MSGSYSTEM_COULD_NOT_ENTER                  = 4,   // You could not enter the next area. (duplicate)
+    MSGSYSTEM_COULD_NOT_ENTER_YOUR_ROOM        = 5,   // You could not enter your room.
```
zach2good: If this is documented elsewhere, then you don't need entries here saying there are duplicates. Should we be using the earliest seen entry if there are duplicates rather than one right in the middle?

zach2good: If you remove the duplicate comments, you remove the need for clang-format on/off

src/map/packet_system.cpp
```diff
@@ -4062,7 +4062,7 @@ void SmallPacket0x05E(map_session_data_t* const PSession, CCharEntity* const PCh
 
                     PChar->loc.p.rotation += 128;
 
-                    PChar->pushPacket(new CMessageSystemPacket(0, 0, 2)); // You could not enter the next area.
+                    PChar->pushPacket(new CMessageSystemPacket(0, 0, MSGSYSTEM_COULD_NOT_ENTER)); // You could not enter the next area.
```
zach2good: 👍

src/map/packets/message_system.h
TeoTwawki: captures for our currently implemented message are using the 1 value I left un-commented, and I dunno where the others are used yet.

nope clang still bitches about other lines having more than 1 space before the equal sign because it only accepts extra spaces to align with a previous line. A behavior I dislike, as these gaps are to be filled later and I want to align the entire thing now and use line breaks to highlight the gaps and not let that digesting out-of-order random groups we do in other files happen here. I'd eventually like to have the _entire_ dat here and kill the document file.

src/map/packets/message_system.h
TeoTwawki: I chose not to out of a desire to end that files life

src/map/packets/message_system.h
TeoTwawki: our other one was also not class enum. I shall attempt to adjust both of them and amend this PR.

TeoTwawki: wound up just doing the one I added, because converting MSGBASIC wound up being too large to get in one go.

WinterSolstice8 (approved):
Death to magic numbers

PR #5164, opened 2024-02-19
[core] Fix several ranged state issues

TracentEden2:
**_I affirm:_**

This PR fixes several issues with the ranged state:
1. The PR adds code to prevent players from swapping ranged weapons and ammo (if no ranged weapon) if the player is in the middle of a ranged attack. This prevents players from starting a RA with a low delay x-bow and ending with a high damage gun (thus gaining the benefit of both). This has been seen in practice on servers.
2. The PR adds a check to make sure players are waiting the appropriate time between ranged attacks and thus not gaming the system through client manipulation.
3. The PR adds a reset of the aim time in cases of RA interrupt so the player does not need to wait the entire duration of a RA until the player can try a RA again.
4. The PR adds a hasMoved function to ranged state (similar to magic state) so that there can be a micro-movement when using RA (as blinking from gear change can cause a movement of less than 0.001).

1. Start a RA with a xbox and try changing to a gun midway through aiming, you should be prevented from changing.
2. Try to use RAs very quickly
3. Interrupt an RA my some means and see that you only have to wait 2.7 seconds until trying again (even if the weapon has a very large delay like a gun).
4. Use an RA and change gear while aiming and see that you will not be interrupted

### DIFF
--- a/src/map/ai/states/range_state.cpp
+++ b/src/map/ai/states/range_state.cpp
@@ -104,7 +104,7 @@ bool CRangeState::Update(time_point tick)
         auto* PTarget = m_PEntity->IsValidTarget(m_targid, TARGET_ENEMY, m_errorMsg);
 
         CanUseRangedAttack(PTarget, true);
-        if (m_startPos.x != m_PEntity->loc.p.x || m_startPos.y != m_PEntity->loc.p.y)
+        if (HasMoved())
         {
             m_errorMsg = std::make_unique<CMessageBasicPacket>(m_PEntity, m_PEntity, 0, 0, MSGBASIC_MOVE_AND_INTERRUPT);
         }
@@ -126,6 +126,8 @@ bool CRangeState::Update(time_point tick)
             {
                 PChar->pushPacket(m_errorMsg.release());
             }
+            // reset aim time so interrupted players only have to wait the correct 2.7s until next shot
+            m_aimTime = std::chrono::seconds(0);
             m_PEntity->loc.zone->PushPacket(m_PEntity, CHAR_INRANGE_SELF, new CActionPacket(action));
         }
         else
@@ -138,7 +140,16 @@ bool CRangeState::Update(time_point tick)
         Complete();
     }
 
-    return IsCompleted() && tick > GetEntryTime() + m_aimTime + 1.5s;
+    if (IsCompleted() && tick > GetEntryTime() + m_aimTime + m_returnWeaponDelay)
+    {
+        if (auto* PChar = dynamic_cast<CCharEntity*>(m_PEntity))
+        {
+            PChar->m_LastRangedAttackTime = GetEntryTime() + m_aimTime + m_returnWeaponDelay;
+        }
+        return true;
+    }
+
+    return false;
 }
 
 void CRangeState::Cleanup(time_point tick)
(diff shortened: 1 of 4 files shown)


zach2good:
Why is this in draft?

TracentEden2:
> Why is this in draft?

Yeah I just had to check one more thing, should be ready now!

src/map/ai/states/range_state.cpp
```diff
@@ -53,7 +54,10 @@ CRangeState::CRangeState(CBattleEntity* PEntity, uint16 targid)
     }
 
     auto delay = m_PEntity->GetRangedWeaponDelay(false);
-    delay      = battleutils::GetSnapshotReduction(m_PEntity, delay);
+    // save the initial dmg and delay of weapon for checking later in the state
+    m_initialDamage = PEntity->GetRangedWeaponDmg();
+    m_initialDelay  = delay;
```
zach2good: Are players swapping between the same item? Surely storing the item ID is the correct answer for this otherwise?

Similarly, this seems remedial, rather than preventative. Would it make more sense to check where we do gear changes and interrupt the user's state if they change the relevant weapon - rather than in the state itself?

src/map/ai/states/range_state.cpp
```diff
@@ -103,8 +107,21 @@ bool CRangeState::Update(time_point tick)
...
+                if (auto PChar = dynamic_cast<CCharEntity*>(m_PEntity))
+                {
+                    m_errorMsg = std::make_unique<CMessageBasicPacket>(PChar, PChar, 0, 0, MSGBASIC_NO_RANGED_WEAPON);
+                }
+            }
+        }
+
         CanUseRangedAttack(PTarget, true);
-        if (m_startPos.x != m_PEntity->loc.p.x || m_startPos.y != m_PEntity->loc.p.y)
+        if (HasMoved())
```
zach2good, line 107: 👍

src/map/ai/states/range_state.cpp
```diff
@@ -126,6 +143,8 @@ bool CRangeState::Update(time_point tick)
             {
                 PChar->pushPacket(m_errorMsg.release());
             }
+            // reset aim time so interrupted players only have to wait the correct 2.7s until next shot
+            m_aimTime = std::chrono::seconds(0);
```
zach2good, line 130: 👍

src/map/ai/states/range_state.cpp
zach2good: Also, a sufficiently quick player with a sufficiently slow weapon will be able to swap from set A, to set B, back to set A and have the ranged attack continue - I don't have any proof, but I can only assume that isn't what retail does, so this should probably be handled where we handle gear changing.

TeoTwawki: Hey guys. I tried this and retail will actually prevent the swap while mid-ranged attack and sometimes do the swap ***after*** if you did it late enough. A similar thing happens if you try to change your ring equip in the middle of using a warp ring - it'll ignore you unless it was near completion, then it'll just do it _after_ instead of during. I presume because the packet then arrived after server side "unlocked" it. You just plain can't swap till server deems it safe.

TracentEden2: > Hey guys. I tried this and retail will actually prevent the swap while mid-ranged attack and sometimes do the swap _**after**_ if you did it late enough. A similar thing happens if you try to change your ring equip in the middle of using a warp ring - it'll ignore you unless it was near completion, then it'll just do it _after_ instead of during. I presume because the packet then arrived after server side "unlocked" it. You just plain can't swap till server deems it safe.

Thanks for retail info, I have updated PR to prevent such gear changing if the player is in a range state, thus should be fairly retail accurate.

src/map/ai/states/range_state.cpp
TeoTwawki: guessing someone copied from magic casting logic. because you can star a magic cast run all over creation and as long as you manage to stop at the spot you started before casting completes, it'll go off 😂 have never heard the same about ranged and I haven't tried because I am not good enough at stopping in the right spot for that

WinterSolstice8 (approved):
Good stuff, thanks.

Xaver-DaRed (approved):

PR #5146, opened 2024-02-15
Atkorkamuy Implemented

Thorinori:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

This implements Atkorkamuy's (Sea Monk NM in Qufim) retail functionality and drops. Note, values are approximated base on comparing fighting him on retail personally.

-Gave 10% DA
-Gave Enaero
-Gave sTP that scales up to 40 based on his missing hp
-Limited his WS pool to Whirlwind
-Gave him loot based on bg-wiki, ffxidb, ffo.jp, and personally getting them while testing
-Gave several spawn points I witnessed him pop at on retail

In addition to the above retail spots, I added a spot that he seems to always pop at on server restart regardless of if it is in the db or not, and I was unable to find what was causing it. It was slightly further away than most of the other spawns, but still within the area that he is supposed to spawn in as best as I could tell. 

## Steps to test these changes

1. Go to Qufim
2. Hit Atkorkamuy
3. Get hit by Atkorkamuy, maybe even DAed
4. Get Whirlwinded ~~noob~~
5. Get him lower hp, then get Whirlwinded a lot more often
6. Get loot maybe

PR #5170, opened 2024-02-20
Improve functionality of Wivres, add some logic to NM:Gharial

CatsEyeXI:
**_I affirm:_**

Sets Wivre 'behavior' to NO_TURN (0x400) or 1024(dec)
Improve the accuracy of the implementation Gharial by adding logic to always use Boiling Blood after Demoralizing Roar.
He ONLY uses these two mobskills. Therefore a new skill list has been created for him.
Add check to Demoralizing Roar to only be used if target is beside the mob.
https://www.youtube.com/watch?v=CyTY32jS9pU
https://www.bg-wiki.com/ffxi/Demoralizing_Roar

!gotoid 16986320


scripts/zones/Wajaom_Woodlands/mobs/Gharial.lua
```diff
@@ -9,6 +9,24 @@ entity.onMobInitialize = function(mob)
     mob:setMod(xi.mod.MOVE_SPEED_STACKABLE, 12)
 end
 
+entity.onMobSpawn = function(mob)
+    mob:addListener('WEAPONSKILL_STATE_EXIT', 'USE_BOILING_BLOOD', function(mobArg, skillId)
+        local roarAbility  = 2101
+        local boilingBlood = 2102
+
+        if
+            mob:getLocalVar('abilityUsed') == 0 and
```
Xaver-DaRed: I dont think this variable is needed at all, considering its always set to 0 on ability use.

WinterSolstice8: It looks like.. this is logic to force use of Boiling Blood after Demoralizing Roar immediately, looks used to me. It might not act correctly after Stuns and such but it's better than not working at all.

WinterSolstice8 (approved):

scripts/zones/Wajaom_Woodlands/mobs/Gharial.lua
Xaver-DaRed: Correct. This mob has only 1 skill in their mobskill list, forbidding it from using the second skill randomly.

After its only skill is used, the mob immediately uses the 2nd one.

However, precisely becouse of that, the use of a local variable seems unnecesary. There is no case I can see in which that variable will be 1 in this step. So the whole local variable logic can and should be removed IMO.

Xaver-DaRed (approved):

scripts/zones/Wajaom_Woodlands/IDs.lua
```diff
@@ -39,6 +39,7 @@ zones[xi.zone.WAJAOM_WOODLANDS] =
         {
             ['Marid'] = GetTableOfIDs('Chigoe', 5),
         },
+        GHARIAL                = GetFirstID('Gharial'),
```
TeoTwawki: sorry but where is this being used? if we're not using it, don't add it. if its being used I don't see it in the pr.

Xaver-DaRed: It was used before my change request. But yeah, it needs to be removed now.

Oopsie

CatsEyeXI: removed it

sql/mob_skill_lists.sql
```diff
@@ -3849,7 +3849,9 @@ INSERT INTO `mob_skill_lists` VALUES ('Zmey_Gorynych',2001,380); -- scythe_tail
 
 INSERT INTO `mob_skill_lists` VALUES ('Copper_Borer',2002,1817); -- thermal_pulse
 
--- Next available ID: 2003
+INSERT INTO `mob_skill_lists` VALUES ('Gharial',2003,2101); -- demoralizing_roar
```
TeoTwawki: 2003 got used by a pr merged earlier today I think.

TeoTwawki: correction: I'd forgotten to hit the button https://github.com/LandSandBoat/server/pull/5146/files#diff-28d0cbb5981fc0487c120848c5354b354cf55784208e97a2483deb3acbc64c79R3852

PR #5178, opened 2024-02-20
Fix hide with sound-based mobs and rain weather with scent-tracking mobs

TracentEden2:
**_I affirm:_**

This PR fixes two issues:
1. The PR adds a missing logical not so that scent tracking mobs lose the ability to pursue by scent in certain weather conditions like rain (which sets `m_disableScent`).
2. The PR adds to the conditional in the `CheckHide` function so that hide does not work on sound-based mobs (like Quadav). This brings hide closer to retail accuracy (though the system is complicated so potentially not fully accurate, see [JP Wiki](https://wiki.ffo.jp/html/1620.html))

1. Find a Kindred mob (sight-based and scent tracking), aggro it, then try to use hide with and without rain weather. You should be able to use hide with rain weather and not without rain weather (as Hide checks indirectly `m_disableScent`).
2. Find a Quadav mob (hearing-based and scent tracking), aggro it, and try to use hide. Hide should fail to shed aggro.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -108,7 +108,7 @@ bool CMobController::CanPursueTarget(CBattleEntity* PTarget)
         if (!PMob->PAI->PathFind->InWater() && PTarget && !PTarget->StatusEffectContainer->HasStatusEffect(EFFECT_DEODORIZE))
         {
             // certain weather / deodorize will turn on time deaggro
-            return PMob->m_disableScent;
+            return !PMob->m_disableScent;
         }
     }
     return false;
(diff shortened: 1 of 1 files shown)


src/map/ai/controllers/mob_controller.cpp
```diff
@@ -108,7 +108,7 @@ bool CMobController::CanPursueTarget(CBattleEntity* PTarget)
         if (!PMob->PAI->PathFind->InWater() && PTarget && !PTarget->StatusEffectContainer->HasStatusEffect(EFFECT_DEODORIZE))
         {
             // certain weather / deodorize will turn on time deaggro
-            return PMob->m_disableScent;
+            return !PMob->m_disableScent;
```
TeoTwawki, line 111: hmm. `m_disableScent` should already be a bool that we're setting elsewhere. feels odd to return "not thisBool" here instead of setting it to be true/false as needed. (its as if the whole condition is bassackward, with all the `!`ing)

Rain has a few less documented "features" on retail as well, such as making worms more active in la theine and decreasing detection range (on specific mobs) for agro/linking separately from being able to track _after_  ..Gonna be a fun thing to try and measure on retail for somebody not named Teo or Tracent, right? 😄

TeoTwawki: This isn't a request to do anything specific on my part. just an observation.

Xaver-DaRed (approved):

zach2good (approved):

Issue #4796, opened 2023-12-02
Sea mob stats


ShiyoKozuki:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

Retail testing and https://wiki.ffo.jp/

Ghrah:
Spider form has double normal mob weapon damage.
All Ghrah are BLM/WAR, but the forms are different. Bird form grants evasion and triple attack equal to appropriate level thief
human form gives defense bonus equal to the defense bonus traits of a paladin of that level AND a bonus 100% total defense. 
spider form gives attack bonus and defense equal to warrior of that level. (Since Ghrah are already BLM/WAR, this basically means their trait bonuses are doubled).
Change form every 60s

Euvhi:
Bloomed/open no def% bonus normal weapon damage * 1.5. Euvhi and 3s delay
Petal / closed +100% def bonus normal wep damage and 4s delay
Will go into closed mode when it takes a hit of at least 350 damage, and will unclose after 80 seconds. 
Do not change form while roamning

Aern:
Bracelets: around a 30% attack boost ~3s delay, around a +40 accuracy bonus.
Bracelets: Last 30 seconds. Aerns will either automatically use them after 80 seconds, or after they have sustained 300 damage while outside of bracelets. No internal cooldown.

to be continued..

demosthene84:
Hey, I am considering to work on this ticket (at least start with the aern part).

After watching a few video like:
- https://www.youtube.com/watch?v=NdhnOGmtGmQ
- https://www.youtube.com/watch?v=Ug2Y95qvG8Q

I can make the following changes:
Bracelets:
- 30% attack boost
- Delay reduce by 50% > to follow the changes made on IxMNK
- +40 accuracy bonus.
- Last 30 seconds. 
- Aerns will automatically use bracelet after 80 seconds in non bracelet mode

But I can find any information about this, does anyone have more details about this ?
"after they have sustained 300 damage while outside of bracelets"

[...]

PR #5181, opened 2024-02-21
[Lua] Move Aweuvhi logic to family mixin, add animation/aggressive relation

CatsEyeXI:
**_I affirm:_**

Aweuvhi's do not currently change their form while roaming like retail does.
Additionally, we have no existing logic to support these mobs' aggressive states based on their current animationSub.
They should be aggressive while open, and non-aggressive while closed.

I've moved the logic from the mob script into a family mixin and added some new logic for ROAM_TICK to randomly alternate between available forms and set aggressive state based on the animationSub chosen.

I've also added an additional zone check to ensure that the Aweuvhis only become aggressive in The Garden of Ruhmet.

AlTaieu:
!gotoid 16912811
Follow it around and see it will change forms every periodically, but never aggro you.

Ruhmet:
!gotoid 16920628 with !togglegm on
Take GM status off, and if flower is open, it will aggro, otherwise, it will not.

### DIFF
--- a/scripts/mixins/families/aweuvhi.lua
+++ b/scripts/mixins/families/aweuvhi.lua
@@ -1,0 +1,91 @@
+require('scripts/globals/mixins')
+
+g_mixins = g_mixins or {}
+g_mixins.families = g_mixins.families or {}
+
+g_mixins.families.aweuvhi = function(aweuvhiMob)
+    aweuvhiMob:addListener('SPAWN', 'AWEUVHI_SPAWN', function(mob)
+        -- Set a random animation when it spawns
+        mob:setAnimationSub(math.random(0, 3))
+        mob:setLocalVar('roamTime', os.time())
+    end)
+
+    aweuvhiMob:addListener('COMBAT_TICK', 'AWEUVHI_COMBAT', function(mob)
+        -- Forms: 0 = Closed  1 = Closed  2 = Open 3 = Closed
+        -- According to http://wiki.ffxiclopedia.org/wiki/Category:Euvhi
+        -- ..when attacked will change states every minute or so..
+        local randomTime = math.random(50, 75)
+        local changeTime = mob:getLocalVar('changeTime')
+
+        if mob:getBattleTime() - changeTime > randomTime then
+            if mob:getAnimationSub() == 2 then
+                mob:setAnimationSub(1)
+            else
+                mob:setAnimationSub(2)
+            end
+
+            mob:setLocalVar('changeTime', mob:getBattleTime())
+
+            -- According to http://wiki.ffxiclopedia.org/wiki/Category:Euvhi
+            -- When in an open state, damage taken by the Euvhi is doubled. Inflicting a large amount of damage to an Euvhi in an open state will cause it to close.
+            -- Make everything do double
+            if mob:getAnimationSub() == 2 then
+                mob:setMod(xi.mod.HTH_SDT, 2000)
+                mob:setMod(xi.mod.SLASH_SDT, 2000)
+                mob:setMod(xi.mod.PIERCE_SDT, 2000)
+                mob:setMod(xi.mod.IMPACT_SDT, 2000)
+                for n = 1, #xi.magic.resistMod, 1 do
+                    mob:setMod(xi.magic.resistMod[n], 2000)
+                end
+
+                for n = 1, #xi.magic.specificDmgTakenMod, 1 do
+                    mob:setMod(xi.magic.specificDmgTakenMod[n], -10000)
+                end
+            else -- Reset all damage types
+                mob:setMod(xi.mod.HTH_SDT, 1000)
+                mob:setMod(xi.mod.SLASH_SDT, 1000)
+                mob:setMod(xi.mod.PIERCE_SDT, 1000)
+                mob:setMod(xi.mod.IMPACT_SDT, 1000)
+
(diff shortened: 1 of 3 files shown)


almuth150:
According to https://github.com/LandSandBoat/server/issues/4796, euvhi aren't supposed to change forms while not engaged, and form change is based off of damage taken to open, and time to close. I don't know what it's supposed to be, but it's worth bringing up the existing issue about this mob type.

Also, while it's been an incredibly long time since I've been on retail, but I swore that the Palace ones were always closed while idle, and the Garden ones were always open while idle.

CatsEyeXI:
> According to #4796, euvhi aren't supposed to change forms while not engaged, and form change is based off of damage taken to open, and time to close. I don't know what it's supposed to be, but it's worth bringing up the existing issue about this mob type.
> 
> Also, while it's been an incredibly long time since I've been on retail, but I swore that the Palace ones were always closed while idle, and the Garden ones were always open while idle.

thanks for raising this, i went to retail to verify and it seems everything that was said is true. i'm going to let this marinate for now.

TeoTwawki:
I'm unsure which ones switch while roaming and which don't, its def not universal behavior even in same zone (eschan ones I've had to wait for them to change to sneak by before)

DiscipleOfEris:
I actually just did a bunch of research on retail and did a PR already for my server on this, but the basics are:

* Specific Euvhis (by mob id, so it is *not* random) spawn and roam in either bud or bloom form.
  * Some Euvhis in both the Garden and the Palace start in bud. Some in both the Garden and the Palace start in bloom.
* Available mob skills change depending on form (three unique to each form, one shared).
* Regardless of whether which form they start/roam in, their in-combat behavior is basically identical.
  * While a mob is in bloom form, it will *never* transition to bud form based on time. It will *always* transition if it takes >=10% of its max HP in a single hit (can be regular attack, spell, WS, anything).
  * While a mob is in bud form, it will transition to bloom form after some combat time between 60 - 90 seconds. I'm not *entirely* sure if this is random or if it varies by zone. It will *never* transition because of damage.
* Mobs will return to their spawn form when they return to roaming and (iirc) reach full HP.

TeoTwawki:
> I actually just did a bunch of research on retail and did a PR already for my server on this, but the basics are:
> 
> * Specific Euvhis (by mob id, so it is _not_ random) spawn and roam in either bud or bloom form.
>   
>   * Some Euvhis in both the Garden and the Palace start in bud. Some in both the Garden and the Palace start in bloom.
> * Available mob skills change depending on form (three unique to each form, one shared).
> * Regardless of whether which form they start/roam in, their in-combat behavior is basically identical.
>   
>   * While a mob is in bloom form, it will _never_ transition to bud form based on time. It will _always_ transition if it takes >=10% of its max HP in a single hit (can be regular attack, spell, WS, anything).
>   * While a mob is in bud form, it will transition to bloom form after some combat time between 60 - 90 seconds. I'm not _entirely_ sure if this is random or if it varies by zone. It will _never_ transition because of damage.
> * Mobs will return to their spawn form when they return to roaming and (iirc) reach full HP.

excellent. if need be we can use 2 names (append an identifier) in the db to avoid having to maintain a list of which mob ids are which kind if they ever exist in same zone, but sounds like to cop zones at least they -never- swap while roaming except to return to their stock state? So we can just start them in the right state  in the db (theres a subanim field in pools I think) then keep track of it after spawn through out changes to return to it.

CatsEyeXI:
> I actually just did a bunch of research on retail and did a PR already for my server on this, but the basics are:
> 
>     * Specific Euvhis (by mob id, so it is _not_ random) spawn and roam in either bud or bloom form.
>       
>       * Some Euvhis in both the Garden and the Palace start in bud. Some in both the Garden and the Palace start in bloom.
> 
>     * Available mob skills change depending on form (three unique to each form, one shared).
> 
>     * Regardless of whether which form they start/roam in, their in-combat behavior is basically identical.
>       
>       * While a mob is in bloom form, it will _never_ transition to bud form based on time. It will _always_ transition if it takes >=10% of its max HP in a single hit (can be regular attack, spell, WS, anything).
>       * While a mob is in bud form, it will transition to bloom form after some combat time between 60 - 90 seconds. I'm not _entirely_ sure if this is random or if it varies by zone. It will _never_ transition because of damage.
> 
>     * Mobs will return to their spawn form when they return to roaming and (iirc) reach full HP.

if you want to contribute what you have, im happy to either close this, or feel free to pitch me your changes on discord and i'll do my best to implement them here

DiscipleOfEris:
> excellent. if need be we can use 2 names (append an identifier) in the db to avoid having to maintain a list of which mob ids are which kind if they ever exist in same zone, but sounds like to cop zones at least they -never- swap while roaming except to return to their stock state? So we can just start them in the right state in the db (theres a subanim field in pools I think) then keep track of it after spawn through out changes to return to it.

I went and retested just now 'cause I couldn't remember, but the euvhi returns to its spawn state the very first time it paths while roaming, regardless of its HP or current bud timer. And yes, the same zone (both Grand Palace and the Garden) can have both default-bloomed and default-bud euvhis. In some cases sitting literally right next to eachother.

> if you want to contribute what you have, im happy to either close this, or feel free to pitch me your changes on discord and i'll do my best to implement them here

Sure I'll DM you.

One thing I forgot to mention was that the *very first hit* against a default-bloomed Euvhi will never cause the euvhi to go into bud form, regardless of how much damage it does.

PR #5185, opened 2024-02-21
[core] Remove usage of `new`, remove `CStateInitException`, template `pushPacket`

zach2good:
**_I affirm:_**

- Removes the pattern of using `new` to create objects, in favour of `std::make_unique` and some templated helpers.
- Removes `CStateInitException` and the new warning-disable we have in MSVC, and rewrites parts of the state system to accommodate this (no more throwing in constructors to signify an error, or holding onto error packets in the state, etc.)
- I've expanded the scope of this to be more than just packets. This'll apply to StatusEffects and anywhere else I can sniff out calls to `new`, where we're creating shared objects and not closely managing the lifetimes of them.

*TODO* 
- Rewriting state system
- The other new calls
- Extensive testing

Everything works as normal, but will need extensive testing of all systems and before/after monitoring with Tracy for both performance and memory usage.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -18,7 +18,6 @@ function(set_project_warnings project_name)
       /wd4457 # declaration of 'var' hides function parameter
       /wd4458 # declaration of 'var' hides class member
       /wd4459 # declaration of 'var' hides global declaration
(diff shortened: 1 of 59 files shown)


zach2good:
**Stretch Goals**
- What using this templated helper might allow; is creating a pool of blank packets that we can use, wipe, and return to the pool, so we're not constantly allocating and freeing memory at runtime. Potential gains!

PR #5186, opened 2024-02-22
[core] Fix SMN player avatar auto attack issue (that can steal claim)

TracentEden2:
**_I affirm:_**

This PR fixes an issue whereby SMN player avatars do not respect mob ownership, thus allowing a SMN or /SMN to steam claim of a mob (such as HNM). The SMN can do this by, for example, spamming cure on another player that has a mob claimed (to gain mob attention), then the avatar will auto-attack the mob (to defend the SMN) and the attack will transfer mob ownership to the SMN (going from purple mob directly to red).

This PR fixes the issue by checking the CanAttack function of the avatar before having the avatar auto-attack to defend the SMN. The PR implements the CanAttack function of PetEntity which checks the mob ownership of the master and then calls the parent CanAttack function of BattleEntity. Also the IsMobOwner function of CharEntity is moved from protected to public so the CanAttack function can call.

This is a PR from ASB coming upstream.

With two characters, have player A claim a mob and then !hp 1 themselves, have player B with an avatar out then spam cure on Player A, and notice that even if the mob turns and starts attacking player B the avatar will not attack the mob and will transfer ownership to Player A.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -183,7 +183,18 @@ void CMobController::TryLink()
     {
         if (PTarget->PPet->objtype == TYPE_PET && ((CPetEntity*)PTarget->PPet)->getPetType() == PET_TYPE::AVATAR)
         {
-            petutils::AttackTarget(PTarget, PMob);
+            if (PTarget->objtype == TYPE_PC)
+            {
+                std::unique_ptr<CBasicPacket> errMsg;
+                if (PTarget->PPet->CanAttack(PMob, errMsg))
+                {
+                    petutils::AttackTarget(PTarget, PMob);
+                }
+            }
+            else
+            {
+                petutils::AttackTarget(PTarget, PMob);
+            }
         }
     }
 
@@ -211,7 +222,7 @@ void CMobController::TryLink()
                 if (PPartyMember->m_roamFlags & ROAMFLAG_IGNORE)
                 {
                     // force into attack action
-                    //#TODO
+                    // #TODO
                     PPartyMember->PAI->Engage(PTarget->targid);
                 }
             }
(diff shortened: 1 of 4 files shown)


CatsEyeXI:
We've known this issue to exist with Automatons as well. Does this address that?

zach2good:
> [Automatons] Does this address that?

This is the updated block:
```cpp
        if (PTarget->PPet->objtype == TYPE_PET && ((CPetEntity*)PTarget->PPet)->getPetType() == PET_TYPE::AVATAR)
        {
            if (PTarget->objtype == TYPE_PC)
            {
                std::unique_ptr<CBasicPacket> errMsg;
                if (PTarget->PPet->CanAttack(PMob, errMsg))
                {
                    petutils::AttackTarget(PTarget, PMob);
                }
            }
```

Without testing I'd wager probably not

TracentEden2:
> We've known this issue to exist with Automatons as well. Does this address that?

What is the strategy for having an automation attack an already claimed mob? At least when I try to use activate it does not work and says already claimed. Do you mean like if the automation uses an AoE move? If the issue is AoE move then potentially needs change in findtarget or something.

The code zach pointed out is only for the avatars auto-defending their master, and thus should not apply to puppets as far as I know.

CatsEyeXI:
I'm going to try to dig up the video I have and get back to you. 

But I was just being inquisitive. If this is out of scope then don't worry about it. I had just occurred to me that this was still happening.

TracentEden2:
> I'm going to try to dig up the video I have and get back to you. 
> 
> But I was just being inquisitive. If this is out of scope then don't worry about it. I had just occurred to me that this was still happening. 

No worries, if it needs a fix would be good to add it here as well, no pup yet on Horizon so no opportunity to have it reported there yet.

CatsEyeXI:
> > I'm going to try to dig up the video I have and get back to you.
> > But I was just being inquisitive. If this is out of scope then don't worry about it. I had just occurred to me that this was still happening.
> 
> No worries, if it needs a fix would be good to add it here as well, no pup yet on Horizon so no opportunity to have it reported there yet.

after speaking with @Xaver-DaRed on this, I don't want to drag the scope out of your PR by making it more complex. This is a nice improvement. We can look at the PUP stuff another time. Thanks for being receptive

Xaver-DaRed (approved):

zach2good (approved):

PR #5191, opened 2024-02-25
[Quest] Elderly Pursuits

jamesbradleym:
**_I affirm:_**

Closes: https://discord.com/channels/933423693848260678/1204689258175139890

Allows players to accept and complete the quest Elderly Pursuits.

Capture: https://www.youtube.com/watch?v=beJ9ca8aZYM

Update your db with dbtool to import mob spawn changes.

!completequest 4 73 (prerequisites)

Speak to Despachiaire : !pos 108 -40 -83 26
Speak to Rouva             : !pos -16 2 11 230
Check ???                      : !pos -412 0 -362 2
Kill spawned Para
Speak to Rouva             : !pos -16 2 11 230
Speak to Despachiaire : !pos 108 -40 -83 26

## NOTE: 

~~I need to update my .dat files to match LSB - I've been able to test everything here except for the text IDs coming from checking ???. The IDs between ASB and LSB in this regard are different. Will update when I've managed to get those dat files sorted.. 😅~~
Confirmed the text IDs are valid based on current LSB dats from retail

### DIFF
--- a/scripts/quests/otherAreas/Elderly_Pursuits.lua
+++ b/scripts/quests/otherAreas/Elderly_Pursuits.lua
@@ -1,0 +1,161 @@
+-----------------------------------
+-- Elderly Pursuits
+-----------------------------------
+-- Log ID: 4, Quest ID: 75
+-- Despachiaire : !pos 108 -40 -83 26
+-- Rouva        : !pos -16 2 11 230
+-- ???          : !pos -412 0 -362 2
+-----------------------------------
+-- !addquest 4 75
+-----------------------------------
+local carpentersLandingID = zones[xi.zone.CARPENTERS_LANDING]
+-----------------------------------
+
+local quest = Quest:new(xi.questLog.OTHER_AREAS, xi.quest.id.otherAreas.ELDERLY_PURSUITS)
+
+quest.reward =
+{
+    item = xi.item.ELEGANT_RIBBON,
+}
+
+local function AllParaDead()
+    for i = 0, 4 do
+        local mob = GetMobByID(carpentersLandingID.mob.PARA_OFFSET + i)
+        if not mob:isDead() then
+            return false
+        end
+    end
+
+    return true
+end
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_AVAILABLE and
+            player:hasCompletedQuest(xi.questLog.OTHER_AREAS, xi.quest.id.otherAreas.SECRETS_OF_OVENS_LOST)
+        end,
+
+        [xi.zone.TAVNAZIAN_SAFEHOLD] =
+        {
+            ['Despachiaire'] =
+            {
+                onTrigger = function(player, npc)
+                    return quest:progressEvent(517)
+                end,
+            },
+
+            onEventFinish =
+            {
+                [517] = function(player, csid, option, npc)
+                    quest:begin(player) -- begin quest
+                    npcUtil.giveKeyItem(player, xi.ki.ANTIQUE_AMULET)
+                end,
+            },
+        },
+    },
+
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_ACCEPTED
+        end,
+
+        [xi.zone.TAVNAZIAN_SAFEHOLD] =
+        {
+            ['Despachiaire'] =
+            {
+                onTrigger = function(player, npc)
+                    if quest:getVar(player, 'Prog') == 3 then
+                        return quest:progressEvent(518)
+                    end
+                end,
+            },
+
+            onEventFinish =
+            {
+                [518] = function(player, csid, option, npc)
+                    if quest:complete(player) then
+                        player:delKeyItem(xi.ki.CATHEDRAL_MEDALLION)
+                    end
+                end,
+            },
+        },
+
+        [xi.zone.SOUTHERN_SAN_DORIA] =
+        {
+            ['Rouva'] =
+            {
+                onTrigger = function(player, npc)
+                    if
+                        quest:getVar(player, 'Prog') == 0 and
+                        player:hasKeyItem(xi.ki.ANTIQUE_AMULET)
+                    then
+                        return quest:progressEvent(747)
+                    elseif
+                        quest:getVar(player, 'Prog') == 2 and
+                        player:hasKeyItem(xi.ki.CATHEDRAL_MEDALLION)
+                    then
+                        return quest:progressEvent(748)
+                    end
+                end,
+            },
+
+            onEventFinish =
+            {
+                [747] = function(player, csid, option, npc)
+                    quest:setVar(player, 'Prog', 1)
+                end,
+
+                [748] = function(player, csid, option, npc)
+                    quest:setVar(player, 'Prog', 3)
+                end,
+            },
+        },
+
+        [xi.zone.CARPENTERS_LANDING] =
+        {
+            ['qm_para'] =
+            {
+                onTrigger = function(player, npc)
+                    if quest:getVar(player, 'Prog') == 1 then
+                        if quest:getLocalVar(player, 'ParaKilled') == 1 then
+                            player:messageSpecial(carpentersLandingID.text.POLISH, xi.ki.ANTIQUE_AMULET)
+                            player:delKeyItem(xi.ki.ANTIQUE_AMULET)
+                            quest:setVar(player, 'Prog', 2)
+                            quest:setLocalVar(player, 'ParaKilled', 0)
+                            return quest:keyItem(xi.ki.CATHEDRAL_MEDALLION)
+                        elseif
+                            GetNPCByID(carpentersLandingID.npc.QM_PARA):getLocalVar('engaged') < os.time() and
+                            GetNPCByID(carpentersLandingID.npc.QM_PARA):getLocalVar('engaged') ~= 1 and
+                            npcUtil.popFromQM(player, npc, { carpentersLandingID.mob.PARA_OFFSET }, { claim = true, hide = 0 })
+                        then
+                            GetNPCByID(carpentersLandingID.npc.QM_PARA):setLocalVar('engaged', 1)
+                            GetNPCByID(carpentersLandingID.npc.QM_PARA):setLocalVar('nextMob', carpentersLandingID.mob.PARA_OFFSET + 1)
+                            return quest:messageSpecial(carpentersLandingID.text.STENCH_OF_DECAY)
+                        end
+                    end
+                end,
+            },
+
+            ['Para'] =
+            {
+                onMobDeath = function(mob, player, optParams)
+                    if AllParaDead() then
+                        quest:setLocalVar(player, 'ParaKilled', 1)
+                        GetNPCByID(carpentersLandingID.npc.QM_PARA):setLocalVar('nextMob', 0)
+                    end
+                end,
+            },
+
+            onZoneOut =
+            {
+                function(player, prevZone)
+                    quest:setLocalVar(player, 'ParaKilled', 0) -- zoning resets kill progress
+                end
+            },
+        }
+    },
+}
+
+return quest
--- a/scripts/zones/Carpenters_Landing/DefaultActions.lua
+++ b/scripts/zones/Carpenters_Landing/DefaultActions.lua
@@ -3,4 +3,5 @@ local ID = zones[xi.zone.CARPENTERS_LANDING]
 return {
     ['Anguenet']          = { event = 21 },
     ['qm_cryptonberries'] = { messageSpecial = ID.text.CRYPTONBERRY_FALLEN_TREE },
+    ['qm_para']           = { messageSpecial = ID.text.NOTHING_OUT_OF_ORDINARY },
 }
--- a/scripts/zones/Carpenters_Landing/IDs.lua
+++ b/scripts/zones/Carpenters_Landing/IDs.lua
@@ -28,9 +28,11 @@ zones[xi.zone.CARPENTERS_LANDING] =
         FISHING_MESSAGE_OFFSET        = 7285, -- You can't fish here.
         LOGGING_IS_POSSIBLE_HERE      = 7402, -- Logging is possible here if you have <item>.
         BEUGUNGEL_SHOP_DIALOG         = 7434, -- Hello, [sir/ma'am]! I'm selling goods direct from the Carpenters' Guild.
+        STENCH_OF_DECAY               = 7484, -- You are overwhelmed by the putrid stench of decay!
         CRYPTONBERRY_EXECUTOR_DIE     = 7486, -- ...Cleave our foesss with barren hate.
         CRYPTONBERRY_ASSASSIN_2HR     = 7487, -- ..Take up thy lanternsss. The truth we shall illuminate.
         CRYPTONBERRY_EXECUTOR_2HR     = 7488, -- Through this we ssseek our just reward...
+        POLISH                        = 7489, -- You polish the <item> with the glowing mushroom.
         MYCOPHILE_MUSHROOM            = 7505, -- There is a rotten mushroom here. There are 3 openings in its cap.
         HERCULES_TREE_NOTHING_YET     = 7507, -- There is nothing here yet. Check again in the morning.
         UNITY_WANTED_BATTLE_INTERACT  = 7550, -- Those who have accepted % must pay # Unity accolades to participate. The content for this Wanted battle is #. [Ready to begin?/You do not have the appropriate object set, so your rewards will be limited.]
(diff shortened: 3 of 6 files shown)


scripts/zones/Carpenters_Landing/mobs/Para.lua
```diff
@@ -0,0 +1,81 @@
...
+        spawnNextMob(target)
+        mob:setLocalVar('spawn', 0)
+    end
+end
+
+entity.onMobWeaponSkill = function(target, mob, skill)
+    local triggerSkills = { 310, 311, 312 }
+    local skillID = skill:getID()
+    if utils.contains(skillID, triggerSkills) then
+        if math.random(0, 100) <= 90 then
```
Xaver-DaRed: math.random should be between 1 and 100.
(0, 100) has 101 possible values

scripts/zones/Carpenters_Landing/mobs/Para.lua
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    return true
+end
+
+local entity = {}
+
+entity.onMobInitialize = function(mob)
+    mob:setMobMod(xi.mobMod.IDLE_DESPAWN, 180) -- 3 min despawn
+    --mob:addImmunity(xi.immunity.LIGHT_SLEEP) immunity does not exist in status.lua
```
Xaver-DaRed: There is no "Light sleep" effect. There is sleep effect from a light elemental source.

If the mob never gets slept by lullabies or repose, just set its light resistance rank to 11.

EDIT: I have watched the video. The mob is immune to sleep. I saw you used both light and dark based sleeps and the mob `completely resisted` both.

To expand:
An actual immunity would return the message `Para completely resists the effect of Sleep.`

The effect not landing, always returning `Para resists the effect of sleep` means it isnt actually immune, it just wont ever land, becouse the resistance rank makes the macc vs meva check always fail
- Resistance rank 10 floors magic hit rate to 5%
- Resistance rank 11 forces a 1/8, combined with the forced 1/2, resulting in a 1/16. In the case of status effects, this means an insta-fail.

scripts/zones/Carpenters_Landing/mobs/Para.lua
```diff
@@ -0,0 +1,81 @@
...
+    mob:setLocalVar('died', 1)
+    if AllParaDead() then
+        GetNPCByID(ID.npc.qm_para):setLocalVar('engaged', os.time()) -- no time gate after death
+    end
+end
+
+entity.onMobDespawn = function(mob)
+    if
+        AllParaDead() and
+        mob:getLocalVar('died') ~= 1
```
Xaver-DaRed, line 75: I dont understand the purpose of this local variable. Could you explain it?

scripts/quests/otherAreas/Elderly_Pursuits.lua
```diff
@@ -0,0 +1,183 @@
...
+-- Elderly Pursuits
+-----------------------------------
+-- Log ID: 4, Quest ID: 75
+-- Despachiaire : !pos 108 -40 -83 26
+-- Rouva        : !pos -16 2 11 230
+-- ???          : !pos -412 0 -362 2
+-----------------------------------
+-- !addquest 4 75
+-----------------------------------
+require('scripts/globals/interaction/quest')
```
claywar: These globals are preloaded, and not necessary

scripts/quests/otherAreas/Elderly_Pursuits.lua
```diff
@@ -0,0 +1,183 @@
...
+        local mob = GetMobByID(carpentersLandingID.mob.PARA_OFFSET + i)
+        if not mob:isDead() then
+            return false
+        end
+    end
+
+    return true
+end
+
+quest.reward =
```
claywar: Minor, but typically reward in all other scripts precedes any other functions or definitions (right after the quest is declared)

scripts/quests/otherAreas/Elderly_Pursuits.lua
```diff
@@ -0,0 +1,183 @@
...
+{
+    item = xi.item.ELEGANT_RIBBON,
+}
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            return status == QUEST_AVAILABLE and
+                player:getQuestStatus(xi.quest.log_id.OTHER_AREAS, xi.quest.id.otherAreas.SECRETS_OF_OVENS_LOST) == QUEST_COMPLETED
```
claywar: `player:hasCompletedQuest`

scripts/quests/otherAreas/Elderly_Pursuits.lua
```diff
@@ -0,0 +1,183 @@
...
+                onTrigger = function(player, npc)
+                    return quest:progressEvent(517)
+                end,
+            },
+
+            onEventFinish =
+            {
+                [517] = function(player, csid, option, npc)
+                    quest:begin(player) -- begin quest
+                    quest:setVar(player, 'Prog', 1)
```
claywar: The quest is beginning, and therefore the state has changed.  There's no need to start with a Prog value of 1 here

scripts/quests/otherAreas/Elderly_Pursuits.lua
```diff
@@ -0,0 +1,183 @@
...
+                        return quest:progressEvent(518)
+                    end
+                end,
+            },
+
+            onEventFinish =
+            {
+                [518] = function(player, csid, option, npc)
+                    player:delKeyItem(xi.ki.CATHEDRAL_MEDALLION)
+                    quest:complete(player)
```
claywar: delKeyItem should be wrapped here, since the reward is an item and that can fail:
```
    player:delKeyItem
end
```

scripts/quests/otherAreas/Elderly_Pursuits.lua
```diff
@@ -0,0 +1,183 @@
...
+        },
+
+        [xi.zone.CARPENTERS_LANDING] =
+        {
+            ['qm_para'] =
+            {
+                onTrigger = function(player, npc)
+                    if quest:getVar(player, 'Prog') == 2 then
+                        if quest:getVar(player, 'ParaKilled') == 1 then
+                            player:messageSpecial(carpentersLandingID.text.POLISH, xi.ki.ANTIQUE_AMULET)
```
claywar, line 123: Does this messageSpecial replace the generic Obtained: <keyitem> or are both displayed?  The returned object will display the latter

scripts/quests/otherAreas/Elderly_Pursuits.lua
```diff
@@ -0,0 +1,183 @@
...
+                        end
+                    end
+                end,
+            },
+
+            ['Para'] =
+            {
+                onMobDeath = function(mob, player, optParams)
+                    if AllParaDead() then
+                        quest:setVar(player, 'ParaKilled', 1)
```
claywar: Two questions: Do the kills still count if the player zones, if they do, why not increment the Prog value instead of introducing another variable?

scripts/zones/Carpenters_Landing/DefaultActions.lua
```diff
@@ -3,4 +3,5 @@ local ID = zones[xi.zone.CARPENTERS_LANDING]
 return {
     ['Anguenet']          = { event = 21 },
     ['qm_cryptonberries'] = { messageSpecial = ID.text.CRYPTONBERRY_FALLEN_TREE },
+    ['qm_para'] = { messageSpecial = ID.text.NOTHING_OUT_OF_ORDINARY },
```
claywar: Alignment

scripts/zones/Carpenters_Landing/IDs.lua
```diff
@@ -61,6 +64,8 @@ zones[xi.zone.CARPENTERS_LANDING] =
         },
 
         LOGGING = GetTableOfIDs('Logging_Point'),
+
+        qm_para = 16785780,
```
claywar: case

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: Zoning after killing Para resets the kill progress: the capture shows the player killing Para -> Zoning -> checking qm = spawn new Para.

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: Switched to hasCompletedQuest, had originally seen this in ASB and saw the same used in LSB (in a few quests but mainly in status == QUEST_COMPLETED, would it be helpful at all to make a pr switching to the hasCompletedquest pattern in these cases?)

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: Works for me!

scripts/zones/Carpenters_Landing/mobs/Para.lua
jamesbradleym: The local variable 'died' allowed me to distinguish whether the mob had despawned due to death or due to idle timeout and thus setting the qm spawn time check - would there be a better way to track this rather than adding a new local variable?

Edit: I should clarify I don't have confirmation that a death clear (vs idle despawn) changes whether or not the qnm should be available to spawn immediately after a clear or if that should also require a 3 min time gate to trigger.. 😅

scripts/zones/Carpenters_Landing/mobs/Para.lua
jamesbradleym: Whoops, fixed. Thanks!

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: In making this change, should ine 74 be modified to:

`if quest:getVar(player, 'Prog') == 0 then`?

claywar: Yup, all undefined variables will return 0

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: Gotcha, done.

scripts/quests/otherAreas/Elderly_Pursuits.lua
claywar: I'd say focus on converting first and replace with that pattern as you go.  There's a lot of overly complex conditions in some NPC scripts, so its probably best to handle one thing at a time

scripts/quests/otherAreas/Elderly_Pursuits.lua
claywar: In that case, 'paraKilled' should be using a local var instead of one that is persisted in the db.  `quest:setLocalVar(player, ...)`

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: Both are displayed, the POLISH text and Obtained text were both shown from my testing - the reason I added the return was because the default text was displaying without it.

Capture:
Screenshot 2024-02-27 at 12 15 12 PM
Local:
Screenshot 2024-02-27 at 12 59 44 PM

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: Removed 👌

Edit: Looks like I still needed `require('scripts/globals/npc_util')`, otherwise I could not get the qm_para to interact after killing Para (confirmed player vars were set correctly).

scripts/zones/Carpenters_Landing/mobs/Para.lua
jamesbradleym: At 18:04 Soporific applies sleep to Para so it looks like it is succeptible to Dark sleep

https://youtu.be/beJ9ca8aZYM?si=Sq1YFiczwdajb5v7&t=1084

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: I see, I'll need to take a look at the difference between setLocalVar vs setVar as used here. Assuming setLocalVar handles cleanup but will take a closer look.

Thanks for the direction here(and everywhere else 😁)!

scripts/zones/Carpenters_Landing/mobs/Para.lua
Xaver-DaRed: I dunnow how I missed that.

In any case, now this is an issue.
This is the first documented case (that ive seen) of a mob being ACTUALLY immune to one sleep kind but not the other. Regularly, as I said before, its just a case of one of the 2 never landing, giving the regular "resists" message (not completely resists)

Now we will have to go to retail and check that those spells actually apply the same effect id or not.

For the time being, this can stay as is, however, I would change the comment to "TODO: Mob is immune to light based sleep, but not dark"

scripts/quests/otherAreas/Elderly_Pursuits.lua
jamesbradleym: Perfect, thanks!

scripts/zones/Carpenters_Landing/mobs/Para.lua
jamesbradleym: added TODO comment 👍🏻

scripts/quests/otherAreas/Elderly_Pursuits.lua
```diff
@@ -0,0 +1,181 @@
...
+                onTrigger = function(player, npc)
+                    return quest:progressEvent(517)
+                end,
+            },
+
+            onEventFinish =
+            {
+                [517] = function(player, csid, option, npc)
+                    quest:begin(player) -- begin quest
+                    return quest:keyItem(xi.ki.ANTIQUE_AMULET)
```
Xaver-DaRed: Dont quote me on this, but i dont think onEventFinish needs a `return`
so
```lua
            onEventFinish =
            {
                [517] = function(player, csid, option, npc)
                    quest:begin(player) -- begin quest
                    quest:keyItem(xi.ki.ANTIQUE_AMULET)
                end,
            },

jamesbradleym: Seems safe to quote you, doesn't look like it's necessary - removed 👍🏻

jamesbradleym: I wasn't having luck without the return on quest:keyItem (and had seen it used elsewhere in a return, albeit not within onEventFinish).. I could use `npcUtil.giveKeyItem(player, ...` which works without a return

claywar: So in the event finish now that I look at it, they don't require a returned value, so the npcUtil function would be the correct path here

scripts/zones/Carpenters_Landing/mobs/Para.lua
Frankie-hz: Well then…that’s interesting…

TeoTwawki: related: immunities are normally added via `mob_pools.sql` and not in script. being able to handle immunities in script is a very recent thing

the full list of values that can be used in the bit mask lives in [battleentity.h](https://github.com/LandSandBoat/server/blob/62f01f3f490833aad7c356e4c5f6918b6816d23d/src/map/entities/battleentity.h#L457) and _does_ have an entry for element sleeps but we don't want that used just yet as it was originally allowed for different reasons and we're not sure that's how we want to handle this going forward.

jamesbradleym: Thanks for the insight @TeoTwawki ! Will keep that in mind, for this particular case I'll keep the TODO pending direction on implementation

jamesbradleym:
@Xaver-DaRed @claywar Hoping this doesn't get lost in the noise, anything else required here?

scripts/zones/Carpenters_Landing/DefaultActions.lua
jamesbradleym: fixed!

scripts/zones/Carpenters_Landing/IDs.lua
jamesbradleym: fixed!

jamesbradleym:
@claywar @Xaver-DaRed Hopefully not being annoying in asking this, what would the next steps be here for this PR?

Xaver-DaRed:
We've been busy lately. Sorry.

We'll get to this as soon as we can.

jamesbradleym:
Understood! No worries :)

jamesbradleym:
> We've been busy lately. Sorry.
> 
> We'll get to this as soon as we can.

Hey @Xaver-DaRed , Just wanted to bump this again to see if anything else needs doing, thanks!

jamesbradleym:
Hey @Xaver-DaRed  / @claywar, anything else needed here?

jamesbradleym:

PR #5194, opened 2024-02-25
[core] [lua] [sql] Nuke nameflags, rework 0x037 and 0x00D (and anything else in the way of destruction)

WinterSolstice8:
**_I affirm:_**

Rework 2SC 0x037 based on https://github.com/atom0s/XiPackets/tree/main/world/server/0x0037
Rework 2SC 0x00D based on https://github.com/atom0s/XiPackets/tree/main/world/server/0x000D
Add sanity to codebase by removing devilish nameflags
Bunch of other stuff, like optimizing the size of 0x00D to generally use less bandwidth

Play the game and see nothing unusual --
Use all the old "nameflags" like d/c, mentor, /inv etc in search and see it work


src/map/packets/char_update.cpp
```diff
@@ -30,53 +30,169 @@
...
+    flags1.FreezeFlag   = 0;                      // Freeze client in place. Is this used?
+    flags1.unknown_1_14 = 0;                      // Unknown.
+    flags1.InvisFlag    = PChar->StatusEffectContainer->HasStatusEffectByFlag(EFFECTFLAG_INVISIBLE);
+    flags1.unknown_2_16 = 0; // Unknown.
+    flags1.SpeedBase    = PChar->speedsub;
+    flags1.unknown_3_25 = 0; // Unknown
+    flags1.BazaarFlag   = PChar->hasBazaar();
+    flags1.CharmFlag    = PChar->isCharmed;
+    flags1.GmIconFlag   = false; // If set, allows other icon flags to be shown in addition to the GM red/pink text for the name.
+
```
zach2good: image

WinterSolstice8:
It's time for reviews, make sure to look for things that maybe don't belong

WinterSolstice8:
looks like !togglegm doesnt persist between zones, which I assume people would want it to?

src/map/lua/luautils.cpp
```diff
@@ -311,7 +311,7 @@ namespace luautils
             }
         }
 
-        if (gLoadAllLua) // Load all lua files (for sanity testing, no need for during regular use)
+        if (true) // Load all lua files (for sanity testing, no need for during regular use)
```
zach2good: Debug testing left in

src/map/packet_system.cpp
```diff
@@ -6178,32 +6176,53 @@ void SmallPacket0x0DB(map_session_data_t* const PSession, CCharEntity* const PCh
...
+        uint32_t ConfigSys[3]; // The players current PTR_pGlobalNowZone->ConfigSys values.
+        uint32_t padding01[4]; // Padding; unused. (Space for future information?)
+        uint32_t Param;        // Packet parameter.
+    } packet = {};
 
-    // Extract the system filter bits and update MenuConfig
-    const uint8 systemFilterMask = (NFLAG_SYSTEM_FILTER_H | NFLAG_SYSTEM_FILTER_L) >> 8;
-    PChar->menuConfigFlags.byte2 &= ~systemFilterMask;
-    PChar->menuConfigFlags.byte2 |= data.ref<uint8>(0x09) & systemFilterMask;
+    std::memcpy(&packet, data, sizeof(packet_c2s_0DB_t));
```
zach2good, line 6194: I like this pattern a lot.

src/map/packet_system.cpp
```diff
@@ -6178,32 +6176,53 @@ void SmallPacket0x0DB(map_session_data_t* const PSession, CCharEntity* const PCh
...
+    uint32_t oldChatFilter1  = {};
+    uint32_t oldChatFilter2  = {};
 
-    PChar->search.language = data.ref<uint8>(0x24);
+    std::memcpy(&oldPlayerConfig, &PChar->playerConfig, sizeof(uint32_t));
+    std::memcpy(&oldChatFilter1, &PChar->playerConfig.MassageFilter, sizeof(uint32_t));
+    std::memcpy(&oldChatFilter2, &PChar->playerConfig.MassageFilter2, sizeof(uint32_t));
 
-    if (oldMenuConfigFlags != PChar->menuConfigFlags.flags)
+    if (packet.Kind == 1)
```
zach2good, line 6205: Can these magic numbers get enum'd, or some comments?

src/map/packet_system.cpp
```diff
@@ -6220,132 +6239,131 @@ void SmallPacket0x0DC(map_session_data_t* const PSession, CCharEntity* const PCh
     TracyZoneScoped;
     switch (data.ref<uint32>(0x04))
     {
-        case NFLAG_INVITE:
+        case 0x01:
```
zach2good, line 6243: Can these magic numbers get enum'd, or some comments?

src/map/packet_system.cpp
```diff
@@ -7978,22 +7996,11 @@ void SmallPacket0x106(map_session_data_t* const PSession, CCharEntity* const PCh
...
 {
     TracyZoneScoped;
-    CItemContainer* PStorage = PChar->getStorage(LOC_INVENTORY);
-    if (PStorage == nullptr)
-    {
-        return;
-    }
 
-    for (uint8 slotID = 1; slotID <= PStorage->GetSize(); ++slotID)
+    if (PChar->isSettingBazaarPrices)
```
zach2good, line 8001: It seems like there are const methods to expose the flags for all of these new bools, can we do the same here please? Maybe making the internal flags private?

src/map/packets/basic.h
```diff
@@ -29,7 +29,8 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include <stdio.h>
 #include <string.h>
 
-#define PACKET_SIZE 0x104
+// Max packet size
+#define PACKET_SIZE 0x1FF
```
zach2good, line 33: Is this true across the board? This means every packet we allocate will be bigger. Just wanted to scrutinise a bit

src/map/packets/basic.h
```diff
@@ -134,7 +135,7 @@ class CBasicPacket
 
     std::size_t getSize()
     {
-        return (size_t)2 * (ref<uint8>(1) & ~1);
+        return std::min<std::size_t>((size_t)2 * (ref<uint8>(1) & ~1), PACKET_SIZE);
```
zach2good: While you're here can we kill this c-style cast? You can probably change the 2's and 1's to 2U and 1U to appease the compiler. Or make a helper so you can call 2_size, 1_size, etc.

src/map/packets/basic.h
```diff
@@ -180,6 +181,16 @@ class CBasicPacket
     {
         return reinterpret_cast<int8*>(data) + index;
     }
+
+    // used for setting "proper" packet sizes rounded to the nearest four away from zero
+    uint32 roundUpToNearestFour(uint32 input)
+    {
+        int remainder = input % 4;
+        if (remainder == 0)
```
zach2good, line 189: Needs some {}'s, but good helper.
Can we get the 2, 4, and 8 siblings of this please?

zach2good: Maybe move to utils

src/map/packets/char.cpp
```diff
@@ -29,186 +29,407 @@
 #include "status_effect_container.h"
 #include "utils/itemutils.h"
 
-CCharPacket::CCharPacket(CCharEntity* PChar, ENTITYUPDATE type, uint8 updatemask)
+// Namespace to avoid compilation units using the same structs twice with different definitions
+namespace charflags
+{
+    // PS2: (Non-defined bitfield type.)
+    struct sendflags_t
```
zach2good, line 36: I am very much a fan of just bringing in the structs as they exist in xipackets. Maybe we should talk to atom0s about making xipackets suitable as a build-time dependency and have these defined in there?

src/map/packets/char.cpp
```diff
@@ -29,186 +29,407 @@
...
+            // This seems wrong, but displays correctly?
+            packet.Flags2.r = (LSColor.R << 4) + 15;
+            packet.Flags2.g = (LSColor.G << 4) + 15;
+            packet.Flags2.b = (LSColor.B << 4) + 15;
         }
-        [[fallthrough]];
-        // clang-format on
-        case ENTITY_UPDATE:
+
+        packet.Flags2.PvPFlag       = 0; // Gate breach status in ballista, apparently.
```
zach2good, line 322: This whole section is :chefs_kiss:

src/map/packets/char.cpp
```diff
@@ -29,186 +29,407 @@
...
+        if (PChar->m_Costume2 != 0)
+        {
+            packet.GrapIDTbl[0] = look->race << 8 | PChar->m_Costume2;
+            packet.GrapIDTbl[8] = 0xFFFF;
+        }
 
-            if (PChar->getMod(Mod::SUPERIOR_LEVEL) == 5 && PChar->m_jobMasterDisplay)
+        if (PChar->m_PMonstrosity != nullptr)
+        {
+            // NOTE: Changing this 0x8000 to 0xC000 will hide the species name.
```
zach2good, line 391: Yeah, this whole system is wacky

zach2good (approved):
Great work here!

I think the benefits of this outweigh any risks that might come from merging it, so apart from my style requests etc. I'd call this ready to go whenever you're happy with it.

It might be worth us saying somewhere in the style guide that "something something while these chunks of code don't explicitly match our style and naming conventions, they're just _too good_ for us not to take wholesale. So, deal with it scrubs"

Very much the exception, rather than the rule.

I am very on board for using these structs, and memcpying into structs at the start of each incoming packet handler. Thats a nice pattern. A+

src/common/mmo.h
```diff
@@ -264,6 +168,122 @@ constexpr auto msgTypeToStr = [](uint8 msgtype)
...
+
+    uint8_t MentorUnlockedFlg : 1;
+    uint8_t MentorFlg : 1;
+    uint8_t NewAdventurerOffFlg : 1;
+    uint8_t DisplayHeadOffFlg : 1;
+    uint8_t unknown28 : 1;
+    uint8_t RecruitFlg : 1;
+    uint8_t unused : 2;
+
+    filters1_t MassageFilter;
```
cocosolos, line 271: Typo?

src/map/packets/char_update.cpp
```diff
@@ -30,119 +30,342 @@
...
+        uint32_t AnonymousFlag : 1;
+        uint32_t CfhFlag : 1;
+        uint32_t AwayFlag : 1;
+        uint32_t Gender : 1;
+        uint32_t unknown_1_9 : 1;
+        uint32_t unknown_1_10 : 1;
+        uint32_t GraphSize : 2;
+        uint32_t Chocobo_Index : 3;
+        uint32_t hpp : 8;
+        uint32_t PlayOnelineFlag : 1;
```
cocosolos, line 58: Typo?

TeoTwawki:
> looks like !togglegm doesnt persist between zones, which I assume people would want it to?

def assume correct here, it always has to now

src/map/packets/basic.h
WinterSolstice8: yes, client can accept up to 0x1FF bytes

src/common/mmo.h
WinterSolstice8: Typo by Square Enix -- this is how the PS2 debug symbols call that variable.

src/map/packets/char_update.cpp
WinterSolstice8: Typo by Square Enix -- this is how the PS2 debug symbols call that variable.

src/map/packets/basic.h
WinterSolstice8: Rounding up to the nearest 4 is a feature of how SE aligns packets, so I thought it made sense to put it in the packet class

src/map/packets/char_update.cpp
TeoTwawki: should we leave their typo and comment that or should we adjust it? lol

src/map/lua/luautils.cpp
WinterSolstice8: Yup this is why I asked to look for stuff that doesn't belong. I was stuck on this branch for so long I had to do development in it too, i knew i'd miss something

src/map/packet_system.cpp
WinterSolstice8: Enums for things like `kind` and `param` aren't really knowable because they got compiled into numbers, but comments aren't a problem

src/map/packet_system.cpp
WinterSolstice8: There's already comments on these 1 line below the cases. Is that sufficient? I could move them up I suppose.

src/map/packet_system.cpp
WinterSolstice8: Those const methods were required to access single bits of a struct iirc
here we would get some absolutely insane naming like

`setIsSettingBazaarPrices(true)`
and `getIsSettingBazaarPrices()`
Not sure if we want that, there really isn't a better name for that variable because that's whats happening. The bazaar flag needs to away while they are _setting_ prices, because they still technically have items in their bazaar in the db because the prices haven't updated yet

src/map/packets/basic.h
WinterSolstice8: Should be done, waiting on CI to see if it hates what I did.

src/map/packets/basic.h
WinterSolstice8: Added the brackets

zach2good: Sure, if this was a member method. But since this takes in and outputs a uint32, it isn't specific to packet logic, it's just a number thing

src/map/packets/basic.h
zach2good: This might be better renamed as MAX_PACKET_SIZE, provided we actually truncate these based on size and aren't always sending out 0x1FF sized packets. As always, maybe a comment explaining this to passers-by, and those of us who forget things

src/map/packet_system.cpp
zach2good: Makes sense, carry on without changes 👍

src/map/packet_system.cpp
zach2good: Meh, leave it as is

src/map/packets/basic.h
WinterSolstice8: It already is a member method of CBasicPacket and won't be used anywhere else but in packets? I'm not really following

src/map/packets/basic.h
WinterSolstice8: GetSize() is called in map.cpp to determine how much of the buffer to copy already, it was just named this way the whole time

zach2good:
My point was that it's simple enough that it looks like it should be moved
to utils, with a 2 and 8 sibling. Not important, leave it as is

zach2good:
YEET

src/map/packets/char_update.cpp
cocosolos: TBH I think it's a bad idea to leave this kind of thing as is. Whatever SE did internally is irrelevant to the project, and in the future if someone tries to "fix" this it will have to be explained to them that "if you decompile the client you'll find this is actually a real typo." In the context of LSB, it just makes things more confusing and harder to search.

TeoTwawki: @cocosolos I can see us changing it, then someone else looking at the debug symbols it came from and then trying to "fix it" to match that again. I don't think it at all matters which we use we'll just want to have an explanation visible either way to head off repeatedly explaining in chat or repeatedly swapping back and forth.

PR #5205, opened 2024-02-29
[Magic] Implement Innin Ninjutsu damage multiplier

Xaver-DaRed:
**_I affirm:_**

- Implements damage multiplier from Innin
- Applies "Koga Hatsuburi" type gear bonuses to magic damage.
- Adds modifier to Innin effect.
- Makes Innin effect decay a little better, making it respect minimum values.
- Found **A** data source for Ninjutsu spells, so updated parameters to refect them as opposed to using made-up data.
- Corrected Ninjutsu skill multiplier not using maximum skill level dependant on spell tier.

Review by commit.

Use innin.
Use a ninjutsu spell facing a mob. See damage.
Now do it from his back. See damage be greater.


WinterSolstice8:
I thought Koga head was just +5 MAB?

Xaver-DaRed:
There's conficting information on the matter but my (limited) retail testing suggests that it isnt a mere MAB addition, but a different (separate) multiplier, additive to Innin's multiplier.

Need to update this PR to account for the fact this bonus (Kogas) is not limited by player position.

zach2good (approved):

PR #5213, opened 2024-03-02
[core] Change dynamic targids to linearly ramp

WinterSolstice8:
**_I affirm:_**

On retail, IDs are not recycled as soon as possible -- They appear to increase until they wrap.
This may fix issues with stale flags/looks.

Adjust long form ID to match retail:

Retail in The Colosseum (yes I managed to get the lowest ID!):
image

LSB:
image

!fafnir, summon pets, trusts, see nothing different (to the player)

### DIFF
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -521,13 +521,7 @@ void CZoneEntities::AssignDynamicTargIDandLongID(CBaseEntity* PEntity)
 
     lastDynamicTargID = targid;
 
-    auto id = 0x1000000 + (m_zone->GetID() << 12) + targid;
-
-    // Add 0x100 if targid is >= 0x800 -- observed on retail.
-    if (targid >= 0x800)
-    {
-        id += 0x100;
-    }
+    auto id = 0x01000000 | (m_zone->GetID() << 0x0C) | (targid + 0x0100);
 
     dynamicTargIds.insert(targid);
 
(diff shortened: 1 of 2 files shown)


src/map/zone_entities.cpp
```diff
@@ -73,7 +73,8 @@ namespace
 typedef std::pair<float, CCharEntity*> CharScorePair;
 
 CZoneEntities::CZoneEntities(CZone* zone)
-: m_zone(zone)
+: lastDynamicTargID(0x6FF) // Start at 0x6FF so the first one used is 0x700. See AssignDynamicTargIDandLongID
```
zach2good: This is a big confusing, start at 0x700 and then increment after assignment

src/map/zone_entities.cpp
```diff
@@ -488,23 +489,39 @@ uint16 CZoneEntities::GetNewCharTargID()
 void CZoneEntities::AssignDynamicTargIDandLongID(CBaseEntity* PEntity)
 {
     // NOTE: 0x0E (entity_update) entity updates are valid for 0 to 1023 and 1792 to 2303
-    uint16 targid = 0x700;
-    for (auto it : dynamicTargIds)
+    // Step targid up linearly from 0x700 one by one to 0x8FF unless that ID is already occupied.
+    uint16 targid = lastDynamicTargID + 1;
```
zach2good: You can use postfix ++ here, or two lines to assign and then increment

src/map/zone_entities.cpp
WinterSolstice8: dead and rewritten

src/map/zone_entities.cpp
WinterSolstice8: Updated the algorithm to be "next" dynamic targID so there's no jank like this, the search will prove what the end result targid is anyway

zach2good (approved):

PR #5214, opened 2024-03-02
[core] [lua] Add ranged attack and accuracy distance correction

TracentEden2:
**_I affirm:_**

This PR adds distance correction to ranged attacks, I am posting here as draft for any initial feedback, as probably some potential improvements that could be made.

 This is done through several different steps
1. Add a new subskill to SUBSKILLTYPE to differentiate between short bows and long bows which have different distance corrections. The types after this change include `SUBSKILL_XBOW_SHORTBOW, SUBSKILL_GUN, SUBSKILL_CNN, SUBSKILL_SHURIKEN, SUBSKILL_LONGBOW`. Also update the item_weapon.sql file to indicate longbows with this new subskill.

2. Add an exception to the equip code in charutils.cpp to account for the fact that now long bows will have a different subskill than arrows (which will still use `SUBSKILL_XBOW_SHORTBOW`). This prevents arrows from unequipping when equipping long bows and vice versa.

3. Update the `RACC` and `RATT` functions of battleentity.cpp to take a distance parameter and a bool parameter for indicating if the RACC or RATT calculation should include a distance correction.

4. Add a `GetRangedDistanceCorrection` function to battleutils.cpp that calculates a fraction (value between 0 and 1) to scale RACC and RATT for some given distance. The actual calculation is delegated to a function in a distance_correction.lua file that uses curves fitted to values from [wiki](https://ffxiclopedia.fandom.com/wiki/Distance#Distance_and_Ranged_Attack). You can see the curve fitting vs. wiki values for the main weapon types in the graphs below (wiki values above and fittings below). This will make it easier to leverage when damage moves to Lua space. Also note that distance correction also applies to mobs, but not to automations (as of 2010), thus the correction is not used for automations (though the automation distance formula is kept in the Lua for completeness sake).

Also several other small improvements were made such as also having the SUBSKILLTYPE enum in lua space to avoid the use of several magic numbers. Changing getRACC in lua_baseentity.cpp to use the battleentity.cpp RACC function rather than recalculating RACC. 

image

This is a PR from ASB coming upstream.


scripts/enum/skill.lua
```diff
@@ -68,3 +68,63 @@ xi.skill =
     -- 60~63 unused
     -- MAX_SKILLTYPE = 64
 }
+
+xi.subskill =
+{
+    XBOW_SHORTBOW = 0,
+    GUN           = 1,
+    CNN           = 2,
```
zach2good: While you're here, can you kill CNN as an abbreviation and use CANON please?

scripts/enum/skill.lua
```diff
@@ -68,3 +68,63 @@ xi.skill =
     -- 60~63 unused
     -- MAX_SKILLTYPE = 64
 }
+
+xi.subskill =
```
zach2good, line 72: If this is specific to ranged subskill, the name could reflect that better (even if there aren't subskills for main etc.)

scripts/globals/automaton.lua
```diff
@@ -297,8 +297,8 @@ end
 
 local function hasAnimatorEquipped(player)
     if
-        player:getWeaponSubSkillType(xi.slot.RANGED) == 10 or
-        player:getWeaponSubSkillType(xi.slot.RANGED) == 11
+        player:getWeaponSubSkillType(xi.slot.RANGED) == xi.subskill.ANIMATOR or
```
zach2good, line 300: 👍

scripts/globals/combat/distance_correction.lua
```diff
@@ -0,0 +1,183 @@
...
+xi.combat.distanceCorrection = xi.combat.distanceCorrection or {}
+-----------------------------------
+
+-- Distance formulas based on piecewise polynomial curve fitting to values from
+-- https://ffxiclopedia.fandom.com/wiki/Distance#Distance_and_Ranged_Attack
+xi.combat.distanceCorrection.table =
+{
+    LONGBOW =
+    {
+        { range =  1.00,
```
zach2good: Range needs to be on the next line for these

src/map/utils/battleutils.cpp
```diff
@@ -298,6 +298,14 @@ namespace battleutils
         }
     }
 
+    /************************************************************************
+     *  Load Distance Correction                                            *
+     ************************************************************************/
+    void LoadDistanceCorrection()
```
zach2good: This isn't needed if the correction code is in a global, we load them all at startup

src/map/utils/battleutils.cpp
```diff
@@ -7060,6 +7078,29 @@ namespace battleutils
         }
     }
 
+    float GetRangedDistanceCorrection(CBattleEntity* PBattleEntity, float distance, bool atk)
+    {
+        TracyZoneScoped;
+
+        auto distanceCorrectionGetValue = lua["xi"]["combat"]["distanceCorrection"]["getValue"];
```
zach2good: Typically, I think we like to interact with lua only from luautils

zach2good (commented):
I like it, seems like a thing that Xaver should review though

zach2good:
Also, I didn't see if there was a setting or module to disable this, seems like an easy add that lots of people will want

scripts/globals/combat/distance_correction.lua
```diff
@@ -0,0 +1,182 @@
...
+    return 1.00
+end
+
+xi.combat.distanceCorrection.throwing = function(distance)
+    if distance <= 1.00 then
+        return 1.00 -- Sweet Spot Under or at 1'
+    end
+
+    return 1.00 - ((1.00 * distance) / 100) -- Lose 1%/yalm
+end
```
Xaver-DaRed, line 139: `return 1.00 - ((1.00 * distance) / 100)`
vs
`return 1 - distance / 100`

Any difference between both?
Also, in general, any reason to have floats all around in lua?

scripts/globals/combat/distance_correction.lua
Xaver-DaRed: To expand:
```lua
xi.combat.distanceCorrection.table =
{
    LONGBOW =
    {
        {
            range =  1.00,
            func = function(dist)
                return 65  -- <=1'
            end
        },
```

scripts/enum/skill.lua
TracentEden2: > While you're here, can you kill CNN as an abbreviation and use CANON please?

yep, fixed.

scripts/globals/combat/distance_correction.lua
TracentEden2: > Range needs to be on the next line for these

ok, fixed.

src/map/utils/battleutils.cpp
TracentEden2: > This isn't needed if the correction code is in a global, we load them all at startup

ah, good to know, thanks, I removed.

scripts/globals/combat/distance_correction.lua
TracentEden2: > `return 1.00 - ((1.00 * distance) / 100)` vs `return 1 - distance / 100`
> 
> Any difference between both? Also, in general, any reason to have floats all around in lua?

Ah, I think that was translated from c++ code originally so probably had leftover float logic from there. I cleaned up the file now so fewer floats. Thanks!

TracentEden2:
> Also, I didn't see if there was a setting or module to disable this, seems like an easy add that lots of people will want

Added a setting, thanks!

scripts/enum/skill.lua
TeoTwawki: Cannon.

3 n for a weapon, 2 n for the writing term

TeoTwawki:
> Also, I didn't see if there was a setting or module to disable this, seems like an easy add that lots of people will want

definitely, a lot of folks including myself see the "sweet spot" things as a nerf and one of SE's poorer decisions. Since its an actual thing SE changed like move speed, I lean towards setting on this one.

scripts/enum/skill.lua
TeoTwawki: seconded

scripts/enum/skill.lua
```diff
@@ -68,3 +68,63 @@ xi.skill =
...
+{
+    XBOW_SHORTBOW = 0,
+    GUN           = 1,
+    CANNON        = 2,
+    SHURIKEN      = 3,
```
TeoTwawki, line 77: I think boomerangs and chakrams also have the same "sweet spot" as shuriken in the distance calcs SE uses. We can probably just rename this "thrown"

...unless that's used to remove ammo or something else weird

src/map/utils/battleutils.cpp
TracentEden2: > Typically, I think we like to interact with lua only from luautils

Yep, moved that function to luautils, thanks.

TracentEden:
For this PR I am waiting on more retail testing from Kanryu as to specific types of weapons and sweet spots (might need to separate out also shortbows and xbows, and RACC and RATT)

zach2good:
I just remembered that this existed, and what needs to be done:
- If you're waiting on more retail testing, that
- Instead of a setting that's essentially enabling this, it should be enabled by default, there shouldn't be a setting, but we should have an era module that replaces the `xi.combat.distanceCorrection` function with a function that returns `1.0f`
- Will need rebase for the conflicts
- Will need Xaver and Winter to look at it again, and decide where it should live in regards to how they want to lay out combat damage in Lua etc.

zach2good:
Such a module should also list some of the information here: https://www.bg-wiki.com/ffxi/Distance_Correction

```
Introduced in the July 19, 2005 version update, it applies to [Ranged Attack](https://www.bg-wiki.com/ffxi/Ranged_Attack) and depends on the distance between the target and the person firing the [Ranged Attack](https://www.bg-wiki.com/ffxi/Ranged_Attack). You receive a penalty to ranged damage and ranged accuracy for not being the proper distance.
```

TracentEden2:
> I just remembered that this existed, and what needs to be done:
> 
> * If you're waiting on more retail testing, that

Yeah some horizon people have been doing a large testing effort on retail but there seems to be a lot of things to test because have to check RACC and RATT for many different levels (as it seems to vary by level) and also different weapons have different sweet spots, and so on. I will work on the other issues you mentioned as well! Thanks

scripts/globals/combat/distance_correction.lua
```diff
@@ -0,0 +1,184 @@
...
+
+xi.combat.distanceCorrection.getValue = function(attacker, distance)
+    if attacker == nil then
+        return 1 -- Just in case BattleEntity is null, then we will just return full damage.
+    end
+
+    -- Automaton
+    if not attacker:isPC() then
+        return xi.damage.distanceCorrection.automaton(distance)
```
UmeboshiXI, line 158: xi.combat.distanceCorrection.automaton(distance)

zach2good:
Any movement on this? I thought this had already been merged

TracentEden2:
> Any movement on this? I thought this had already been merged

Apologies for the delay on this, there is a group testing a lot on this and I am waiting for their summary document with the testing results, then I can update the PR.

scripts/globals/combat/distance_correction.lua
```diff
@@ -0,0 +1,184 @@
...
+    -- Fallback
+    return 1
+end
+
+xi.combat.distanceCorrection.throwing = function(distance)
+    if distance <= 1 then
+        return 1 -- Sweet Spot Under or at 1'
+    end
+
+    return 1 - distance / 100 -- Lose 1%/yalm
```
zach2good, line 138: This needs a clamp on it

scripts/globals/combat/distance_correction.lua
```diff
@@ -0,0 +1,184 @@
...
+xi.combat.distanceCorrection.applyCurve = function(distance, curve)
+    for i = 1, #curve do
+        if distance <= curve[i].range then
+            -- the curve table is out of 100 thus clamp and divide to get [0,1]
+            return utils.clamp(curve[i].func(distance), 65, 100) / 100
+        end
+    end
+
+    -- Fallback
+    return 1
```
zach2good, line 130: I'm not sure whether or not the global fallback for everything should be 100% - I'd rather the failure condition leads to obscenely less damage, to make it a visibly different game experience so people will complain and make the bug visible

settings/default/map.lua
```diff
@@ -277,4 +277,7 @@ xi.settings.map =
     -- The max GM level is 5, so setting this to 6 disables it
     -- for everone. Setting it to 0 enables for everyone.
     REPORT_LUA_ERRORS_TO_PLAYER_LEVEL = 6,
+
+    -- Enable/disable distance correction to ranged attack and accuracy
+    RANGED_DISTANCE_CORRECTION = true,
```
zach2good, line 282: I don't think this setting is needed anymore, with the advent of modules. Just replace this with an `era` module that overrides `lua["xi"]["combat"]["distanceCorrection"]["getValue"];` with a function that returns 1.0, or whatever it is in era

src/map/entities/battleentity.cpp
```diff
@@ -748,11 +748,17 @@ uint16 CBattleEntity::RATT(uint8 skill, uint16 bonusSkill)
     // make sure to not use fishing skill
     uint16 baseSkill = skill == SKILL_FISHING ? 0 : GetSkill(skill);
     int32  RATT      = 8 + baseSkill + bonusSkill + m_modStat[Mod::RATT] + battleutils::GetRangedAttackBonuses(this) + (STR() * 3) / 4;
+
+    if (withDistanceCorrection && settings::get<bool>("map.RANGED_DISTANCE_CORRECTION"))
+    {
+        RATT = int32((float)RATT * luautils::GetRangedDistanceCorrection(this, distance));
```
zach2good, line 754: If we get rid of this setting, we can just use this directly

src/map/entities/battleentity.cpp
```diff
@@ -772,6 +778,13 @@ uint16 CBattleEntity::RACC(uint8 skill, uint16 bonusSkill)
     RACC += getMod(Mod::RACC);
     RACC += battleutils::GetRangedAccuracyBonuses(this);
     RACC += (AGI() * 3) / 4;
+
+    if (withDistanceCorrection && settings::get<bool>("map.RANGED_DISTANCE_CORRECTION") &&
+        !this->StatusEffectContainer->HasStatusEffect(EFFECT_SHARPSHOT))
+    {
+        RACC = int16((float)RACC * luautils::GetRangedDistanceCorrection(this, distance));
```
zach2good, line 785: If we get rid of this setting, we can just use this directly

zach2good:
> Apologies for the delay on this, there is a group testing a lot on this and I am waiting for their summary document with the testing results, then I can update the PR.

Since this does my favourite thing of "making things more retail-accurate, and also stopping people face-rolling things", I want to get this merged and if you have further testing results the numbers can be tweaked later

TracentEden:
> > Apologies for the delay on this, there is a group testing a lot on this and I am waiting for their summary document with the testing results, then I can update the PR.
> 
> Since this does my favourite thing of "making things more retail-accurate, and also stopping people face-rolling things", I want to get this merged and if you have further testing results the numbers can be tweaked later

Yep, I understand, I will work on making this current version work, I have to make some adjustments to the core code of this PR due to changes since I originally posted it.

Xaver-DaRed:
#7673 was merged, making this one obsolete. Linking both PRs just in case.

PR #5218, opened 2024-03-04
[core] Fix scaling of range critical hit rate from AGI

TracentEden2:
**_I affirm:_**

This fixes a small issue related to critical hit rate calculation from AGI. Essentially, instead of using abs(dAGI), we should be using max(0, dAGI) so that negative dAGI do not have any impact on crit hit rate (this completes the fix started in my [previous PR](https://github.com/LandSandBoat/server/pull/5142), and thus should have been there, my apologies). Also no source that I could find denotes a known cap (unlike for dDex and critical hit rate), thus I removed the cap (see [bg-wiki here](https://www.bg-wiki.com/ffxi/Critical_Hit_Rate)).

Fight mobs with higher AGI and lower AGI (than player) and use RAs and track critical hit rate

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -2822,7 +2822,7 @@ namespace battleutils
             critRate = 1;
         }
 
-        // Crit rate delta from stats caps at +-15
+        // Crit rate delta from stats caps at 15
         return std::min(critRate, 15);
     }
 
@@ -2871,15 +2871,13 @@ namespace battleutils
         // https://www.bg-wiki.com/bg/Critical_Hit_Rate
         int32 attackerAgi = PAttacker->AGI();
         int32 defenderAgi = PDefender->AGI();
-        int32 dAGI        = attackerAgi - defenderAgi;
-        int32 dAgiAbs     = std::abs(dAGI);
+        int32 dAgi        = attackerAgi - defenderAgi;
+        // Only consider positive dAgi
+        int32 dAgiPositive = std::max(0, dAgi);
 
-        // Default to +0 crit rate
-        int32 critRate = 0;
+        int32 critRate = dAgiPositive / 10;
 
-        critRate = dAgiAbs / 10;
-
-        return std::min(critRate, 15);
+        return critRate;
     }
 
     /************************************************************************


src/map/utils/battleutils.cpp
```diff
@@ -2871,15 +2871,13 @@ namespace battleutils
...
+        int32 dAgiPositive = std::max(0, dAgi);
 
-        // Default to +0 crit rate
-        int32 critRate = 0;
-
-        critRate = dAgiAbs / 10;
+        int32 critRate = dAgiPositive / 10;
 
-        return std::min(critRate, 15);
+        return critRate;
```
Xaver-DaRed, line 2880: Doesnt this remove the 15 cap?

attackerAgi = 200
defenderAgi = 1
dAgi = 200 - 1 = 199
dAgiPositive = std::max(0, 199) = 199
critRate = 199 / 10 = 19

return 19

EDIT: Re-read your PR description

Xaver-DaRed: Comment in code should be updated to reflect why theres no cap.
Or added, in this case.

However, I think the cap should remain until proven otherwise. Just my opinion though.

Frankie-hz: I believe it does say there is a cap unless I am misunderstanding this line:

`you can see that there is an extremely significant jump in critical hit rate when dDEX is increased from 40 to 50 (from 4% at dDEX=39 to 15% at dDEX=50), but no further boost to critical hit rate once dDEX is greater than 50.`

TracentEden2: > I believe it does say there is a cap unless I am misunderstanding this line:
> 
> `you can see that there is an extremely significant jump in critical hit rate when dDEX is increased from 40 to 50 (from 4% at dDEX=39 to 15% at dDEX=50), but no further boost to critical hit rate once dDEX is greater than 50.`

That is for dDex to normal crits, not dAgi to ranged crit hits. This PR only deals with ranged crit hits.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5220, opened 2024-03-05
[core] [lua] Fix Fomor hate clearing and aggro

TracentEden2:
**_I affirm:_**

This PR fixes two fomor hate issues: 

Firstly, the PR allows clearing fomor hate by killing certain mobs (like gigas and beastmen in misareaux coast). This was previously not working because those mobs had code attempting to set local vars to negative numbers which is not possible (because local vars are stored as uint in c++ land). This PR fixes this by setting the local vars to positive numbers and then multiplying these numbers by -1 in the fomor hate mixin if the mob is not a fomor.

Secondly, the PR prevents aggro if the player has the lowest level of fomor hate. This is done by adding a check in `CMobController::CanAggroTarget`.

This is an ASB PR coming upstream.

Check the char variable FOMOR_HATE after killing, for example, fomor in sacrarium and after killing gigas and beastmen in misareaux coast. Also check if fomor aggro in, for example, sacrarium with <8 FOMOR_HATE and >= 8 FOMOR_HATE.

### DIFF
--- a/scripts/mixins/fomor_hate.lua
+++ b/scripts/mixins/fomor_hate.lua
@@ -15,6 +15,12 @@ g_mixins.fomor_hate = function(fomorMob)
                         adj = 2 -- default: most fomor add 2 hate
                     end
 
+                    -- if not a fomor then decrease hate instead of increase
+                    -- Note cannot use negatives in fomorHateAdj because local vars can only be positive
+                    if mob:getFamily() ~= 115 and mob:getFamily() ~= 359 then
+                        adj = -adj
+                    end
+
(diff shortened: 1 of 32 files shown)


TeoTwawki:
> (because local vars are stored as uint in c++ land)

🤦‍♂️ this frustrates me so much, EVERY other variable type uses signed values to allow negatives.

no complaint about your fix, just hating the past atm.

Xaver-DaRed (approved):
This system needs a fire treatment, but that isn't your fault.

claywar (approved):

PR #5223, opened 2024-03-07
[core] [lua] Fix mob meikyo shisui and mob skill tp gain

TracentEden2:
**_I affirm:_**

This PR fixes several issues with mobskills and TP:
1. Currently, mobs using Meikyo Shisui get 3000 TP but do not use multiple mob skills in a row (the first mobskill just wipes the 3000 tp). This PR fixes this by giving mobs 3000 TP and then allowing them to use 3 mob skills in a row (each consumes 1000 TP, so first skill is at 3000, second at 2000 and so on). This leverages a local var on the mob which is set upon using Meikyo Shisui.

2. Currently, mobs using mob skills do not gain TP from these moves. This PR fixes this by adding code in mobskill.lua to add the correct TP for different types of mob skills. The formula is essentially similar to players.

1. Fight a SAM mob and watch the mob use three mob skills in a row after using Meikyo Shisui.

2. Fight mobs and monitor mobs TP with for example !exec print(target:getTP()) before and after mob skills.

### DIFF
--- a/scripts/actions/mobskills/meikyo_shisui.lua
+++ b/scripts/actions/mobskills/meikyo_shisui.lua
@@ -14,6 +14,9 @@ mobskillObject.onMobWeaponSkill = function(target, mob, skill)
 
     mob:addTP(3000)
 
+    -- use local var counter to ensure only three mobskills count for Meikyo Shisui
+    mob:setLocalVar('[MeikyoShisui]MobSkillCount', 3)
+
     return xi.effect.MEIKYO_SHISUI
 end
 
--- a/scripts/globals/mobskills.lua
+++ b/scripts/globals/mobskills.lua
@@ -230,6 +230,11 @@ xi.mobskills.mobPhysicalMove = function(mob, target, skill, numHits, accMod, dmg
         finaldmg   = 0
         hitslanded = 0
         skill:setMsg(xi.msg.basic.SKILL_MISS)
+    -- calculate tp return of mob skill
+    else
+        local tpReturn = xi.combat.tp.getSingleMeleeHitTPReturn(mob, target)
+        tpReturn = tpReturn + 10 * (hitslanded - 1) -- extra hits give 10 TP each
+        mob:addTP(tpReturn)
     end
 
     returninfo.dmg        = finaldmg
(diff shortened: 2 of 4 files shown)


TeoTwawki:
> Currently, mobs using mob skill attacks like tiamat fire attack while flying do not gain TP from these moves. This PR fixes this by adding a new mobskill flag the denotes skills that give the mob normal TP and the associated plumbing that enables this in mobskill state. 

A flag already exists for ensuring their "skill use" when using them as substitutes for actual melee don't consume the tp. their tp gains should be happening regardless of weather melee is fake or not, so we can just check that flag without making a new one **Edit: actually, there should still be tp gains on -all- skills just like you get tp for the "hits" of your weapon skill, shouldn't there? does that not happen for retail mobs?**

as for meikyo shisui I was unaware someone had broken this - for years I've been beating up a certain yagudo who uses it and he always went 3x after popping it. there should have been a core side check ensuring that during the status, only 1000 TP was consumed each time. I'm almost certain there used to be.

TracentEden2:
> > Currently, mobs using mob skill attacks like tiamat fire attack while flying do not gain TP from these moves. This PR fixes this by adding a new mobskill flag the denotes skills that give the mob normal TP and the associated plumbing that enables this in mobskill state.
> 
> A flag already exists for ensuring their "skill use" when using them as substitutes for actual melee don't consume the tp. their tp gains should be happening regardless of weather melee is fake or not, so we can just check that flag without making a new one **Edit: actually, there should still be tp gains on -all- skills just like you get tp for the "hits" of your weapon skill, shouldn't there? does that not happen for retail mobs?**
> 
> as for meikyo shisui I was unaware someone had broken this - for years I've been beating up a certain yagudo who uses it and he always went 3x after popping it. there should have been a core side check ensuring that during the status, only 1000 TP was consumed each time. I'm almost certain there used to be.

Yeah I think you are correct that there should be tp gains on all skills for mobs. I will look into how this should be accomplished. A solution for this would then not require the special flag or any changes on that front.

TracentEden2:
@TeoTwawki I updated the PR so that mobskills.lua now leverages `xi.combat.tp.getSingleMeleeHitTPReturn` to calculate the TP return for different types of mob skills (phy, rng, magic, breath). From what I see all magic and breath mob skills are single hit (or at least there is no number of hits field) so I use the TP return of single melee hit.

WinterSolstice8 (approved):

TeoTwawki (approved):

PR #5224, opened 2024-03-07
[Proposal] [Enum] Enumerate FOURCC animation codes

Xaver-DaRed:
**_I affirm:_**

Enumarates some of those random FOURCC codes we use all around, at times.
Adds some minor documentation to the enum file.

The purpose/hope with this is it will ease and streamline their usage when actually needed.

No changes should be notized.


scripts/enum/animation_strings.lua
```diff
@@ -0,0 +1,107 @@
...
+-----------------------------------
+SendEntityVisualPacket(ID.npc.GEYSER_OFFSET, 'kkj2')
+-----------------------------------
+This one is a luautils binding, sending a packet to the entire zone (of whoever is in visual range of the ID specified) on demand.
+
+This are FOURCC codes.
+For most things you can get away with using the wrong one, and they still just work.
+--]]
+
+xi.animationString =
```
claywar, line 29: Seems reasonable to me.  My only question is should we nest tables on this with these logical groupings instead of having them all in one?

claywar (approved):

scripts/enum/animation_strings.lua
TeoTwawki: I think we can kick the grouping idea down the road for now, because if we group we're going to have to settle on how we should group them and right now I don't see a way to do so that we won't regret later. These are "kind of sort of" model specific, yet multiple models can have the same one. And some are part of the zone instead of a specific npc id. (Or a mob in rarer cases that are usually a hacky work around.) One big list is the lesser evil at this time.

scripts/enum/animation_strings.lua
```diff
@@ -0,0 +1,107 @@
...
+-----------------------------------
+player:entityVisualPacket('saa1')
+-----------------------------------
+In this there is no mob  or npc at all.
+Tt's a part of the zone itself animating, so player is used as base entity and any player that zones in gets sent the packet to light up that part of the zone.
+So the thing animating it's really the zone itself, sort of.
+(There is technically "an object"  or "entity" but its not ID'd like an NPC, and technically speaking a zone dat is laid out like NPCs are part of it anyway)
+
+-----------------------------------
+SendEntityVisualPacket(ID.npc.GEYSER_OFFSET, 'kkj2')
```
TeoTwawki: you probably want to use your shiny new enum instead of the actual string in these examples. Should probably also paraphrase rather than copy that whole discord message, because ol Teo is a long winded pain in the rear that wrote you a novel to point out 3 function calls ^^;;

scripts/enum/animation_strings.lua
```diff
@@ -0,0 +1,107 @@
...
+    OPEN_CRATE_GLOW           = 'open',
+    OPEN_CRATE_SMOKE          = 'ope1',
+    OPEN_CRATE                = 'ope2',
+
+    -- Doors
+    OPEN_DOOR                 = 'smin',
+    CLOSE_DOOR                = 'kmin',
+
+    -- Special animations
+    MOB_SPECIAL_0             = 'sp00', -- Prishe Raise, Selh'teus Lancet, Al'tieu Euvhi (Flowers) sticking to ground, Zdei vertical barriers on, etc...
```
TeoTwawki: suggest dropping the prepend of `mob_` because there is a metric ton of NPCs that use these as well. Including some of the chest objects.

image

I'd go with `SPECIAL_00`

TeoTwawki: more "specials" for synergy furnaces:
`sp12` = start
`sp21`  = finish success
`sp31` = finish fail
might do diff thing on another model, no idea. probably are actually more generic kind of "open" animations, since the furnace lid opens similarly to a chest

scripts/enum/animation_strings.lua
```diff
@@ -0,0 +1,107 @@
...
+    CAST_BLACK_MAGIC_STOP     = 'shbk',
+    CAST_BLUE_MAGIC_STOP      = 'shbl',
+    CAST_ITEM_STOP            = 'shit',
+    CAST_NINJA_STOP           = 'shnj',
+    CAST_SUMMONER_STOP        = 'shsm',
+    CAST_SONG_STOP            = 'shso',
+    CAST_WHITE_MAGIC_STOP     = 'shwh',
+
+    -- Actions
+    JUMP_1                    = 'jmp0',
```
TeoTwawki: I think its less confusing to stick to the same number SE used, like you did with the specials where you started at 0

scripts/enum/animation_strings.lua
Xaver-DaRed: But I liked uncle Teo's wording....

I'll use the enums though and try to make the explanations shorter

TeoTwawki (approved):

PR #5225, opened 2024-03-08
Fix PH entries for Antican Magister

CatsEyeXI:
**_I affirm:_**

FIx the placeholder entry for Antican Magister.

https://www.bg-wiki.com/ffxi/Antican_Magister
https://ffxiclopedia.fandom.com/wiki/Antican_Magister

Set the spawn chance in each of the mob Lua's to 100% and kill each placeholder in the area of C-6 QSC

### DIFF
--- a/scripts/zones/Quicksand_Caves/IDs.lua
+++ b/scripts/zones/Quicksand_Caves/IDs.lua
@@ -57,7 +57,15 @@ zones[xi.zone.QUICKSAND_CAVES] =
         },
         ANTICAN_MAGISTER_PH     =
         {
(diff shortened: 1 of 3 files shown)


WinterSolstice8 (approved):
Approved, but note that there's an update coming soon where a bunch of IDs shifted (nearly 700 files last patch, mostly from zones!). This might need to be adjusted.

claywar:
Holding on this, as I'm currently touching every NPC in the game to adapt for the latest update

WinterSolstice8:
The patch PRs came in, this can be revisited now

PR #5226, opened 2024-03-08
[LUA] re-apply mantra to the target

MowFord:
**_I affirm:_**

Removes the mantra buff from the target of the ability instead of the player. These are the same objects when solo but when in a party the monk was losing mantra immediately after gaining it

Use mantra in a party and see that all party members in range have hp boost afterwards

### DIFF
--- a/scripts/globals/job_utils/monk.lua
+++ b/scripts/globals/job_utils/monk.lua
@@ -132,7 +132,7 @@ end
 xi.job_utils.monk.useMantra = function(player, target, ability)
     local merits = player:getMerit(xi.merit.MANTRA)
 
-    player:delStatusEffect(xi.effect.MAX_HP_BOOST)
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
https://wiki.ffo.jp/html/2834.html
image

Interesting. Can you put a TODO to verify if mantra can overwrite itself?

TeoTwawki:
> https://wiki.ffo.jp/html/2834.html image
> 
> Interesting. Can you put a TODO to verify if mantra can overwrite itself?

if jp wiki proves correct, we'll need to have max HP boost overwrite by greater tier (not greater or equal like most effects) only and make mantra the highest tier - then it replaces any other except itself.

WinterSolstice8 (approved):

TeoTwawki (approved):

PR #5227, opened 2024-03-09
[LUA] Ensure player EES gets 5x multiplier

MowFord:
**_I affirm:_**

Player Eagle Eye Shot has all the right parameters, but `xi.weaponskills.ftp` requires tp between 1k and 3k, afaict there is nothing that varies with TP in the chain of logic (i.e we could pass 1k or 3k tp and get the same results in this PR).

Do regular ranged attack, then EES... see they are about the same dmg

apply patch and see EES doing about 5x dmg

### DIFF
--- a/scripts/actions/abilities/eagle_eye_shot.lua
+++ b/scripts/actions/abilities/eagle_eye_shot.lua
@@ -44,6 +44,7 @@ abilityObject.onUseAbility = function(player, target, ability, action)
     params.numHits = 1
 
     -- TP params.
+    local tp          = 1000 -- to ensure ftp multiplier is applied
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

TeoTwawki (approved):
huh..I would have expected any tp < 1000 to get treated as if it was 1000, since under normal circumstances either tp can't be utilized less than that or doesn't actually influence the damage (to date, monstrosity is the only way to see a tp attack go off using less than 1k tp that isn't just a trick of lag and timing)

I was sure we did that somewhere for melee ws, but don't see it now. I guess a refactor must have killed that undocumented feature. We should probably consider if we want to do a clamp check again in the future.

PR #5229, opened 2024-03-09
[Lua] Correct 'Smite of rage' shadow behavior and fTP

CatsEyeXI:
**_I affirm:_**

Smite of Rage is currently consuming between 2-3 shadows. It is a single hit weaponskill. 
This change removes the random removal of 2-3 shadows as well as updates the mobskill script to align with current LSB standard.
Tested as Lv75 SAM on Aura Weapons.

image

Go to your favorite 'weapon' mob and spam !exec target:useMobAbility(513)

### DIFF
--- a/scripts/actions/mobskills/smite_of_rage.lua
+++ b/scripts/actions/mobskills/smite_of_rage.lua
@@ -10,11 +10,14 @@ end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
-    local accmod = 1
-    local dmgmod = 2.7
+    local accmod  = 1
+    local dmgmod  = mob:getWeaponDmg() * 0.03
+
     local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
-    local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.PHYSICAL, xi.damageType.SLASHING, info.hitslanded * math.random(2, 3))
+    local dmg  = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.PHYSICAL, xi.damageType.SLASHING, xi.mobskills.shadowBehavior.NUMSHADOWS_1)
+
     target:takeDamage(dmg, mob, xi.attackType.PHYSICAL, xi.damageType.SLASHING)
+
     return dmg
 end
 
--- a/scripts/actions/mobskills/smite_of_rage.lua
+++ b/scripts/actions/mobskills/smite_of_rage.lua
@@ -11,10 +11,10 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
     local accmod  = 1
-    local dmgmod  = mob:getWeaponDmg() * 0.03
+    local dmgmod  = 3
 
     local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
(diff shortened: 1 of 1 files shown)


scripts/actions/mobskills/smite_of_rage.lua
```diff
@@ -9,11 +9,10 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    local numhits = 1
     local accmod = 1
-    local dmgmod = 2.7
-    local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
-    local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.PHYSICAL, xi.damageType.SLASHING, info.hitslanded * math.random(2, 3))
+    local dmgmod = .03
+    local info = xi.mobskills.mobPhysicalMove(mob, target, skill, mob:getWeaponDmg(), accmod, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
```
Xaver-DaRed: I don't understand why `numhits` is being replaced by `mob:getWeaponDmg()`
imagen

CatsEyeXI: I recall a previous discussion (i can't seem to find the PR where) we would be getting rid of numhits in favor of mob:getWeaponDmg() * multiplier...

If you look at mobskills zarbzan, wind_blade, or stormwind, for example, you'll see numhits is not present.

CatsEyeXI: i can certainly add it back, but its unused right?

Xaver-DaRed: Those are magical moves, not physical ones.
They use `xi.mobskills.mobMagicalMove` instead of `xi.mobskills.mobPhysicalMove`

numhits is certanly used in the physical mobskill function
imagen

Xaver-DaRed: That's the confusion here.
`xi.mobskills.mobMagicalMove` does indeed use `mob:getWeaponDmg() * multiplier` in the place of `numhits`, because magical moves are always single hit, iirc.

scripts/actions/mobskills/smite_of_rage.lua
```diff
@@ -10,11 +10,14 @@ end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
-    local accmod = 1
-    local dmgmod = 2.7
+    local accmod  = 1
+    local dmgmod  = mob:getWeaponDmg() * 0.03
```
Frankie-hz: This is already being calculated inside the `mobPhysicalMove`

dmgmod is basically a flat FTP right now the way its coded. 

It will hit this: `local base = math.max(1, mob:getWeaponDmg() + dSTR)` which the damage will return `level + 2` which will then calculate the hitdamage after it with dlvl here:

```
--work out the base damage for a single hit
local hitdamage = math.max(1, base + lvldiff) * dmgMod
```

It appears the only thing that needs to change in this PR would be the 2.7 to 3 as per FTP and shadows to hits landed per retail testing shown here: https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit#gid=57955395

Xaver-DaRed (approved):

scripts/actions/mobskills/smite_of_rage.lua
TeoTwawki: yep #5062 baked in the multiplier for us. older scripts need an audit soon to prevent copy pasting `mob:getWeaponDmg()` into new ones

edit: I'm told it was actually another pr before that, but git blamed 5062 for it /shrug

TeoTwawki (approved):
looks like smite began its life as copy paste of whirl.

PR #5239, opened 2024-03-11
[lua] snake eye to automatic 11 only when previous roll is 5+

MowFord:
**_I affirm:_**

Snake eye effect should give a chance to automatic roll 11 only when the roll _power_ is 5+, old code was checking the previous roll's effectID

Fully merit snake eye and see that you only ever roll 11 when previous roll power is 5+


MowFord:
Correction it wasn't comparing the old effect id, it was comparing the current roll power, i.e. Chaos power 6 is, etc. Regardless, it wasn't using the roll count

Xaver-DaRed (approved):

WinterSolstice8 (approved):
Ah, this one sure was a subtle one huh. thanks.

PR #5241, opened 2024-03-13
[LUA] Shijin Spiral plague duration

MowFord:
**_I affirm:_**

[Shijin Spiral](https://www.bg-wiki.com/ffxi/Shijin_Spiral) should give 5-8 _ticks_ of plague, not 4-7 seconds. I suspect "5-8" means 30s with resists lowering duration but there's no good way to verify that I imagine.

Shijin mobs, see duration of effect. Here's godmode at 3k tp for full duration

image

### DIFF
--- a/scripts/actions/weaponskills/shijin_spiral.lua
+++ b/scripts/actions/weaponskills/shijin_spiral.lua
@@ -29,7 +29,7 @@ weaponskillObject.onUseWeaponSkill = function(player, target, wsID, tp, primary,
     local damage, criticalHit, tpHits, extraHits = xi.weaponskills.doPhysicalWeaponskill(player, target, wsID, params, tp, action, primary, taChar)
 
     if damage > 0 then
-        local duration = (tp / 1000) + 4
+        local duration = 3 * ((tp / 1000) + 5) -- Plague effect is -50TP/tick and lasts for 5-8 ticks.
         if not target:hasStatusEffect(xi.effect.PLAGUE) then
             target:addStatusEffect(xi.effect.PLAGUE, 5, 0, duration)
         end


WinterSolstice8 (approved):

scripts/actions/weaponskills/shijin_spiral.lua
```diff
@@ -29,7 +29,7 @@ weaponskillObject.onUseWeaponSkill = function(player, target, wsID, tp, primary,
     local damage, criticalHit, tpHits, extraHits = xi.weaponskills.doPhysicalWeaponskill(player, target, wsID, params, tp, action, primary, taChar)
 
     if damage > 0 then
-        local duration = (tp / 1000) + 4
+        local duration = 3 * ((tp / 1000) + 5) -- Plague effect is -50TP/tick and lasts for 5-8 ticks.
         if not target:hasStatusEffect(xi.effect.PLAGUE) then
             target:addStatusEffect(xi.effect.PLAGUE, 5, 0, duration)
```
TeoTwawki, line 34: It's odd to me that the (preexisting) tick rate here is 5 seconds. SE tends to re-use several standard values, 5 might be among the rarer ones if used on retail at all. worth investigating.

TeoTwawki (approved):

scripts/actions/weaponskills/shijin_spiral.lua
MowFord: Power is 5, it's a normal 3-second tick afaik

TeoTwawki: i yustacood read mor betterer

I was looking at wrong field

but now that you said that I realize that line uses 0 where the 3 tick rate should be. we're only getting 3 because of fallback code introduced that prevents anything with a tickrate from having infinity duration\* 

###### \*0 is intentionally used for infinity for non-ticking effects. but ever since dsp changes the global ticker to 2.4s and the statuses minimum is 3s, that introduced a small chance of an underflow for tickable effects

TeoTwawki: note to self: audit all tickables later

PR #5254, opened 2024-03-14
[core] Remove "Invisible" status effect when using items

TracentEden2:
**_I affirm:_**

This fixes an issue whereby using an item did not remove the status effect invisible.

Cast invisible and use an item and see that invisible is removed

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -2580,6 +2580,15 @@ namespace luautils
             return -1;
         }
 
+        // using an item removes invisible status effect
+        if (auto* PBattleEntity = dynamic_cast<CBattleEntity*>(PUser))
+        {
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):
LGTM

I tried several items in retail to confirm, including Prism Powder.

Obviously, its unreasonable to check them all.

WinterSolstice8:
Does this still apply invisible over itself when using prism powders?

TracentEden2:
> Does this still apply invisible over itself when using prism powders?

Ah, good catch, I moved the code up slightly so this also now fully works, thanks!

WinterSolstice8 (approved):

TeoTwawki:
> * I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

you should be testing things related to the change, not just that invis got removed.

PR #5259, opened 2024-03-15
[core] Add function to send item(s) to a delivery box

TracentEden2:
**_I affirm:_**

This PR adds a lua function called SendItemToDeliveryBox that sends a specified item to a players delivery box. The function takes params including name of player, item id, quantity, and message to display as the sender (such as "Server Staff" or "GM"). And it leverages an enum of return codes to indicate the outcome of the request.

The main purpose of the function is to allow lua GM commands to be built that leverage the function. For example a GM command might be !deliveritem that would allow GMs to refund (due to a game bug) some item(s) even if the player is not online when the GM is processing the ticket.

This is an ASB addition coming upstream.

Can call the command with !exec and try sending different items to players delivery boxes

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -224,6 +224,7 @@ namespace luautils
         lua.set_function("GetContainerFilenamesList", &luautils::GetContainerFilenamesList);
         lua.set_function("GetCachedInstanceScript", &luautils::GetCachedInstanceScript);
         lua.set_function("GetItemIDByName", &luautils::GetItemIDByName);
+        lua.set_function("SendItemToDeliveryBox", &luautils::SendItemToDeliveryBox);
         lua.set_function("SendLuaFuncStringToZone", &luautils::SendLuaFuncStringToZone);
         lua.set_function("RoeParseRecords", &roeutils::ParseRecords);
         lua.set_function("RoeParseTimed", &roeutils::ParseTimedSchedule);
@@ -5590,6 +5591,84 @@ namespace luautils
         }
 
         customMenuContext.erase(PChar->id);
+    }
+
+    SendToDBoxReturnCode SendItemToDeliveryBox(std::string const& playerName, uint16 itemId, uint32 quantity, std::string senderText)
+    {
+        const char* getPlayerIDQuery = "SELECT charid FROM chars WHERE charname = '%s'";
+        int32       queryRet         = _sql->Query(getPlayerIDQuery, playerName);
+        uint32      playerID         = 0;
+
+        if (queryRet != SQL_ERROR && _sql->NumRows() != 0 && _sql->NextRow() == SQL_SUCCESS)
+        {
+            playerID = _sql->GetIntData(0);
+        }
+        else
+        {
+            return SendToDBoxReturnCode::PLAYER_NOT_FOUND;
+        }
+
+        auto isGil = itemId == 65535;
+
+        // Check to confirm that the item legitimately exists
+        // exclude gil as gil does not have an item pointer
+        auto* PItem = itemutils::GetItemPointer(itemId);
+        if (PItem == nullptr && !isGil)
+        {
+            return SendToDBoxReturnCode::ITEM_NOT_FOUND;
+        }
+
+        // default stack size of gil
+        uint32 stackSize = 999999999;
+        // if not gil then get the actual stack size
+        if (!isGil)
+        {
+            stackSize = PItem->getStackSize();
+        }
+
+        bool quantityMoreThanStackSize = quantity > stackSize;
+
+        // limit the quantity to the stack size of the item
+        quantity = std::clamp<uint32>(quantity, 1, stackSize);
+
+        bool isAutoCommitOn = _sql->GetAutoCommit();
+
+        if (_sql->SetAutoCommit(false) && _sql->TransactionStart())
+        {
+            const char* Query = "INSERT INTO delivery_box (charid, box, itemid, quantity, senderid, sender) VALUES ("
+                                "%u, "     // Player ID
+                                "1, "      // Box ID == 1
+                                "%u, "     // Item ID
+                                "%u, "     // Quantity
+                                "%u, "     // Sender ID ( =Player ID )
+                                "'%s'); "; // Sender Text
+            int32 ret = _sql->Query(Query, playerID, itemId, quantity, playerID, senderText);
+
+            if (ret == SQL_ERROR)
+            {
+                _sql->TransactionRollback();
+                _sql->SetAutoCommit(isAutoCommitOn);
+                return SendToDBoxReturnCode::QUERY_ERROR;
+            }
+            else
+            {
+                _sql->TransactionCommit();
+                _sql->SetAutoCommit(isAutoCommitOn);
+            }
+
+            if (quantityMoreThanStackSize)
+            {
+                return SendToDBoxReturnCode::SUCCESS_LIMITED_TO_STACK_SIZE;
+            }
+            else
+            {
+                return SendToDBoxReturnCode::SUCCESS;
+            }
+        }
+        else
+        {
+            return SendToDBoxReturnCode::QUERY_ERROR;
+        }
     }
 
     uint16 GetItemIDByName(std::string const& name)
--- a/src/map/lua/luautils.h
+++ b/src/map/lua/luautils.h
@@ -63,6 +63,15 @@ extern sol::state lua;
 #include "lua_trigger_area.h"
 #include "lua_zone.h"
 
+enum class SendToDBoxReturnCode : uint8
+{
+    SUCCESS                       = 0,
+    SUCCESS_LIMITED_TO_STACK_SIZE = 1,
+    PLAYER_NOT_FOUND              = 2,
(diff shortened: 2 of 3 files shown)


src/map/lua/luautils.cpp
```diff
@@ -5581,6 +5582,51 @@ namespace luautils
         customMenuContext.erase(PChar->id);
     }
 
+    int SendItemToDeliveryBox(std::string const& playerName, uint16 itemId, uint32 quantity, std::string senderText)
```
claywar: Should be a uint return, given the results, also, since this lands in lua, the ret results should most likely be enum'd

src/map/lua/luautils.cpp
```diff
@@ -5581,6 +5582,51 @@ namespace luautils
...
+        {
+            return 1; // Player Not Found (return code 1)
+        }
+
+        auto isGil = itemId == 65535;
+
+        // For anything other than gil, clamp to a 99 maximum
+        if (!isGil)
+        {
+            quantity = std::clamp<uint32>(quantity, 1, 99);
```
claywar: What is the reasoning for clamping gil to 99?

TracentEden2: > What is the reasoning for clamping gil to 99?

Ah, the logic is to clamp everything except gil to 99

TeoTwawki: get stacksize, then clamp to stacksize instead. clamp gil as well but to the lesser of max gil or the unint32 size (max gil 9 digits of 9 I think?) consider possibly using a ternary, might not need any bool variable.

TeoTwawki: update: turns out gil has a defined stacksize too, so we can just stacksize check everything 👍

TeoTwawki: pseudo-code, lacking safety or proper casts:
```cpp
auto PItem     = itemutils::GetItemPointer(itemID);
auto stackSize = PItem->getStackSize();
```

TracentEden2: > pseudo-code, lacking safety or proper casts:
> 
> ```c++
> auto PItem     = itemutils::GetItemPointer(itemID);
> auto stackSize = PItem->getStackSize();
> ```

Yeah, the issue is that using getItemPointer with the gil itemid of 65535 gives a null pointer, so would need some other way to get a gil item object, or keep in the gil check.

TeoTwawki: > Yeah, the issue is that using getItemPointer with the gil itemid of 65535 gives a null pointer, so would need some other way to get a gil item object, or keep in the gil check.

oh that needs fixed, but for now I guess ternary or conditional so that its 999999999 when gil (there's a DEFINE setting gils stacksize at this number, which was why I thought we could use stackSize for gil)

TeoTwawki: *guess not get

src/map/lua/luautils.cpp
TracentEden2: > Should be a uint return, given the results, also, since this lands in lua, the ret results should most likely be enum'd

For the enum'd, do you mean like an enum class in luautils and in a lua enum file under scripts/enum? Also do I need an enum lua object then that is passed between c++ and lua? or can I just use like uint8?

WinterSolstice8: > For the enum'd, do you mean like an enum class in luautils and in a lua enum file under scripts/enum?

Yes. You can still return a uint8 with `static_cast<uint8>(your::enum::val)`, in lua its all numbers so equivalency isn't a problem

TracentEden2: > > For the enum'd, do you mean like an enum class in luautils and in a lua enum file under scripts/enum?
> 
> Yes. You can still return a uint8 with `static_cast<uint8>(your::enum::val)`, in lua its all numbers so equivalency isn't a problem

Thanks! I updated the PR now to leverage an enum.

src/map/lua/luautils.cpp
TracentEden2: > > Yeah, the issue is that using getItemPointer with the gil itemid of 65535 gives a null pointer, so would need some other way to get a gil item object, or keep in the gil check.
> 
> oh that needs fixed, but for now I guess ternary or conditional so that its 999999999 when gil (there's a DEFINE setting gils stacksize at this number, which was why I thought we could use stackSize for gil)

Thanks I updated the PR now so it also limits gil to the gil stack size.

WinterSolstice8 (approved):

src/map/lua/luautils.h
```diff
@@ -354,6 +363,7 @@ namespace luautils
 
     // Retrive the first itemId that matches a name
     uint16 GetItemIDByName(std::string const& name);
+    uint8  SendItemToDeliveryBox(std::string const& playerName, uint16 itemId, uint32 quantity, std::string senderText);
```
zach2good: If you've created an enum type to return here, return that type:
```
auto SendItemToDeliveryBox(std::string const& playerName, uint16 itemId, uint32 quantity, std::string senderText) -> SendToDBoxReturnCode;
```

src/map/lua/luautils.h
```diff
@@ -63,6 +63,15 @@ extern sol::state lua;
 #include "lua_trigger_area.h"
 #include "lua_zone.h"
 
+enum SendToDBoxReturnCode : uint8
```
zach2good: If this doesn't need to be decay'd down into it's underlying type, make it an `enum class`

src/map/lua/luautils.cpp
```diff
@@ -5581,6 +5582,69 @@ namespace luautils
...
+
+        // default stack size of gil
+        uint32 stackSize = 999999999;
+        // if not gil then get the actual stack size
+        if (!isGil)
+        {
+            stackSize = PItem->getStackSize();
+        }
+
+        bool quantityMoreThanStackSize = false;
```
zach2good:
```
bool quantityMoreThanStackSize = quantity > stackSize;
```

zach2good (approved):
Code seems fine, just some style/sanity bits

TracentEden2:
> Code seems fine, just some style/sanity bits

Thanks, I updated given your suggestions!

WinterSolstice8:
you've got a space missing according to CI

TracentEden2:
> you've got a space missing according to CI

Thanks, I added the space

WinterSolstice8:
Small conflict from the huge sql pr, looks like its adding an underscore?

src/map/lua/luautils.cpp
```diff
@@ -5592,6 +5593,65 @@ namespace luautils
...
+                            "'%s'); "; // Sender Text
+        int32 ret = _sql->Query(Query, playerID, itemId, quantity, playerID, senderText);
+
+        if (ret == SQL_ERROR)
+        {
+            return SendToDBoxReturnCode::QUERY_ERROR;
+        }
+        else
+        {
+            _sql->TransactionCommit();
```
zach2good: Presumably, this should also be paired with `TransactionStart` and potentially `TransactionRollback`, or there's no point in using it?

zach2good (changes requested):

src/map/lua/luautils.cpp
zach2good: In this case, if the query fired, it's fired. There's no transaction to commit

TracentEden2: > In this case, if the query fired, it's fired. There's no transaction to commit

Yep, I will add the full transaction logic, thanks! 

Also one question is related to user input escaping, I see existing examples for player name (fixed length of 16) escaping however not for escaping variable length strings (like the sender text). 

I see this closed [PR](https://github.com/LandSandBoat/server/pull/5326/files) that had a fix for the EscapeString function for strings (as it currently does not work). Will the fix be put in another PR? @WinterSolstice8  Thanks!

WinterSolstice8: > I see this closed [PR](https://github.com/LandSandBoat/server/pull/5326/files) that had a fix for the EscapeString function for strings (as it currently does not work). Will the fix be put in another PR? @WinterSolstice8 Thanks!

That PR was obsoleted by the new SQL system Zach put in, I believe the escaping is already done for you now.

WinterSolstice8 (approved):
Approved contingent on Zach's request

src/map/lua/luautils.cpp
TracentEden: > > I see this closed [PR](https://github.com/LandSandBoat/server/pull/5326/files) that had a fix for the EscapeString function for strings (as it currently does not work). Will the fix be put in another PR? @WinterSolstice8 Thanks!
> 
> That PR was obsoleted by the new SQL system Zach put in, I believe the escaping is already done for you now.

Ah, thanks for the info!

TracentEden: > In this case, if the query fired, it's fired. There's no transaction to commit

Ok, I updated the PR to leverage all the appropriate transaction functions now. Also I no longer using `getCharIdFromName` because it requires the name to be in quotes which is inconsistent with other commands that take a player name. Maybe this function and other areas in the code that leverage with similar functionality can be refactored, but out of scope for this PR.

zach2good (approved):

PR #5266, opened 2024-03-16
[core] Remove destructor on CInstance, as the parent performs cleanup.

### LINKED ISSUE #5263: 🐛 Instace cleanup crashes server
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce
It appears that if mobs remain in an instance, a cleanup of the instance crashes the server due to this deconstructor:

image

image

We've seen it for instance id 6300 as well. Here's a crashdump

https://cdn.discordapp.com/attachments/1152236166817521785/1217858332748939284/toau.zip?ex=66058df5&is=65f318f5&hm=3057548938a7edf2f81505808831f3d9488bc6b3325ed711c289b4565b65492f&

WinterSolstice8:
**_I affirm:_**

As it says on the tin, the CInstance destructor is obsoleted by the parent class, preventing a double free.

Fixes #5263 

see #5263

### DIFF
--- a/src/map/instance.cpp
+++ b/src/map/instance.cpp
@@ -35,6 +35,7 @@ CInstance::CInstance(CZone* zone, uint16 instanceid)
 , m_instanceid(instanceid)
 , m_zone(zone)
 {
+    TracyZoneScoped;
     LoadInstance();
 
     m_startTime = server_clock::now();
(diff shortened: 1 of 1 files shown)


MowFord:
Confirmed this resolved the linked issue

src/map/instance.cpp
```diff
@@ -41,25 +41,7 @@ CInstance::CInstance(CZone* zone, uint16 instanceid)
...
-    for (auto entity : m_petList)
-    {
-        destroy(entity.second);
-    }
-    for (auto entity : m_trustList)
-    {
-        destroy(entity.second);
-    }
-}
+CInstance::~CInstance() = default;
```
zach2good: Instead of default, can you slap a Tracy Zone in here, and in the constructor? Then it's good to go 👍

zach2good (approved):

src/map/instance.cpp
WinterSolstice8: Updated, this -- with `TracyZoneScoped;` ?

zach2good: Yessir, looks good 👍

PR #5272, opened 2024-03-18
[cpp] Ensure player smn avatars assist from any range

MowFord:
**_I affirm:_**

Started as a bug report, but in testing to create the report I ended up figuring out how to fix it

SMN avatar should attack a mob that aggros the SMN even with no action taken by the smn

[This PR](https://github.com/LandSandBoat/server/pull/5186/files) seems to have made it not work until the mob has a battletarget (which means the mob has to get close enough to act/attack the player), is within melee range of the avatar, and is claimed, since the `battleutils::CanAttack` function requires those last 2 things

This PR doesn't revert the requirement to have the mob be actionable, but it doesn't go all the way to requiring everything from `battleutils::CanAttack`

Summon an avatar and aggro a mob without acting on it, see your pet assists. Similarly, act on it with say smn/rdm and casting dia.

Before this change the first would not attack at all, second would wait to assist until the mob got within melee range of the pet.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -185,8 +185,8 @@ void CMobController::TryLink()
         {
             if (PTarget->objtype == TYPE_PC)
             {
-                std::unique_ptr<CBasicPacket> errMsg;
-                if (PTarget->PPet->CanAttack(PMob, errMsg))
+                auto* PChar = dynamic_cast<CCharEntity*>(PTarget);
+                if (PChar && PChar->IsMobOwner(PMob))
                 {
                     petutils::AttackTarget(PTarget, PMob);
                 }


TracentEden2:
Thanks for this PR, one thing is I think it might still need the `IsUntargetable` check from the `CanAttack` function? Otherwise it might be possible to act on mobs that should be untargetable at certain stages?

MowFord:
> Thanks for this PR, one thing is I think it might still need the `IsUntargetable` check from the `CanAttack` function? Otherwise it might be possible to act on mobs that should be untargetable at certain stages?

Oh perhaps, yea. This would catch issues like a worm going underground. Feels like that should be added to the base `battleutils::CanAttack`?

TracentEden2:
> > Thanks for this PR, one thing is I think it might still need the `IsUntargetable` check from the `CanAttack` function? Otherwise it might be possible to act on mobs that should be untargetable at certain stages?
> 
> Oh perhaps, yea. This would catch issues like a worm going underground. Feels like that should be added to the base `battleutils::CanAttack`?

yeah, I think it is there, but just not sure if it is checked in this case of the PR. Also I think it should be `PChar->IsMobOwner(PMob)` ? As now I think it is currently checking if the player owns themselves

MowFord:
In the context of this PR it would probably engage the mob, chase it, then first action the overloaded CanAttack would cause it to disengage.

You're absolutely right about checking ownership of the wrong entity, updated and tested

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5277, opened 2024-03-18
[core] [lua] Add GM command to print all local vars of a target entity

TracentEden2:
**_I affirm:_**

This PR adds a GM command to print out all local vars of an entity. This command is very useful for GMs when a player reports some mob or player issue and the GM wants to collect information to later pass on to devs (who are often not online for real time debugging of the mob or player).

This is from ASB and coming upstream.

Target a mob or player as a GM (level 1+) and use !getlocalvars

### DIFF
--- a/scripts/commands/getlocalvars.lua
+++ b/scripts/commands/getlocalvars.lua
@@ -1,0 +1,33 @@
+-----------------------------------
+-- Func: getlocalvars
+-- Desc: Gets all local vars of a target
+-----------------------------------
+local commandObj = {}
+
+commandObj.cmdprops =
+{
+    permission = 1,
+    parameters = ''
+}
+
+commandObj.onTrigger = function(player)
+    local targ = player:getCursorTarget()
+    if targ == nil then
+        return
+    end
+
+    local vars = {}
+    vars = targ:getLocalVars()
+
+    if #vars > 0 then
+        player:printToPlayer(string.format('Printing local vars for entity: %s', targ:getName()), xi.msg.channel.SYSTEM_3)
+        player:printToPlayer('----------------------------------', xi.msg.channel.SYSTEM_3)
+        for _, var in pairs(vars) do
+            player:printToPlayer(string.format('"%s" : %u', var['varname'], var['value']), xi.msg.channel.SYSTEM_3)
+        end
+    else
+        player:printToPlayer(string.format('No local vars for entity: %s', targ:getName()), xi.msg.channel.SYSTEM_3)
+    end
+end
+
+return commandObj
--- a/src/map/entities/baseentity.cpp
+++ b/src/map/entities/baseentity.cpp
@@ -202,6 +202,11 @@ uint32 CBaseEntity::GetLocalVar(std::string var)
     return m_localVars[var];
 }
 
+std::map<std::string, uint32>& CBaseEntity::GetLocalVars()
+{
+    return m_localVars;
+}
+
 void CBaseEntity::SetLocalVar(std::string var, uint32 val)
 {
     m_localVars[var] = val;
--- a/src/map/entities/baseentity.h
+++ b/src/map/entities/baseentity.h
@@ -283,6 +283,7 @@ public:
     void   ResetLocalVars();
     uint32 GetLocalVar(std::string var);
     void   SetLocalVar(std::string var, uint32 val);
+    auto   GetLocalVars() -> std::map<std::string, uint32>&;
 
     // pre-tick update
     virtual void Tick(time_point) = 0;
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -665,6 +665,28 @@ void CLuaBaseEntity::setVolatileCharVar(std::string const& varName, int32 value,
 }
 
 /************************************************************************
+ *  Function: getLocalVars()
+ *  Purpose : Returns all variables assigned locally to an entity
+ *  Example : local localVars = KingArthro:getLocalVars()
+ *  Notes   :
+ ************************************************************************/
+
+auto CLuaBaseEntity::getLocalVars() -> sol::table
+{
+    auto  table     = lua.create_table();
+    auto& localVars = m_PBaseEntity->GetLocalVars();
+
+    for (auto const& [varName, value] : localVars)
+    {
+        auto subtable       = lua.create_table();
+        subtable["varname"] = varName;
+        subtable["value"]   = value;
+        table.add(subtable);
+    }
+    return table;
+}
+
+/************************************************************************
  *  Function: getLocalVar()
  *  Purpose : Returns a variable assigned locally to an entity
  *  Example : if KingArthro:getLocalVar("[POP]King_Arthro") > 0 then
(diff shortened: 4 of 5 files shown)


scripts/commands/getlocalvars.lua
```diff
@@ -0,0 +1,33 @@
+-----------------------------------
+-- Func: getlocalvars
+-- Desc: Gets all local vars of a target
+-----------------------------------
+local commandObj = {}
+
+commandObj.cmdprops =
+{
+    permission = 3,
```
TeoTwawki: Except for extremely dangerous actions, we normally start at `1` and let servers adjust them if they need more restriction

This doesn't seem dangerous.

TeoTwawki:
was this tested on targets that have nothing?

TracentEden2:
> was this tested on targets that have nothing?

Yep, it gives the message "No local vars for entity", can be tested by adding `targ:resetLocalVars()` right above `vars = targ:getAllLocalVars()` in the command script. Otherwise difficult to test because entities seem to get `pauseNPCPathing` set to 0 every tick (I am not sure of the reason).

scripts/commands/getlocalvars.lua
TracentEden2: > Except for extremely dangerous actions, we normally start at `1` and let servers adjust them if they need more restriction
> 
> This doesn't seem dangerous.

Ah, ok, thanks for info, yeah I think not very dangerous in itself. I adjusted the script to permission 1.

TeoTwawki (dismissed):

src/map/entities/baseentity.cpp
```diff
@@ -202,6 +202,11 @@ uint32 CBaseEntity::GetLocalVar(std::string var)
     return m_localVars[var];
 }
 
+std::map<std::string, uint32> CBaseEntity::GetAllLocalVars()
+{
+    return m_localVars;
```
zach2good, line 207: This is returning a full copy of `m_localVars`, no reason you can't return a `&` to it

src/map/entities/baseentity.h
```diff
@@ -284,6 +284,8 @@ class CBaseEntity
     uint32 GetLocalVar(std::string var);
     void   SetLocalVar(std::string var, uint32 val);
 
+    std::map<std::string, uint32> GetAllLocalVars();
```
zach2good: The naming here isn't particularly helpful, `GetLocalVars` is fine, unless there are other functions that return a filtered version of the map

src/map/lua/lua_baseentity.cpp
```diff
@@ -664,6 +664,28 @@ void CLuaBaseEntity::setVolatileCharVar(std::string const& varName, int32 value,
...
+ *  Example :
+ *  Notes   :
+ ************************************************************************/
+
+auto CLuaBaseEntity::getAllLocalVars() -> sol::table
+{
+    auto                          table     = lua.create_table();
+    std::map<std::string, uint32> localVars = m_PBaseEntity->GetAllLocalVars();
+
+    for (auto var : localVars)
```
zach2good: This creates copies and throws them away on every iteration of this loop.

`for (auto const& [varname, value] : ...`

src/map/lua/lua_baseentity.cpp
```diff
@@ -664,6 +664,28 @@ void CLuaBaseEntity::setVolatileCharVar(std::string const& varName, int32 value,
...
+ *  Function: getAllLocalVars()
+ *  Purpose : Returns all variables assigned locally to an entity
+ *  Example :
+ *  Notes   :
+ ************************************************************************/
+
+auto CLuaBaseEntity::getAllLocalVars() -> sol::table
+{
+    auto                          table     = lua.create_table();
+    std::map<std::string, uint32> localVars = m_PBaseEntity->GetAllLocalVars();
```
zach2good: Use `auto&` here

src/map/entities/baseentity.h
zach2good: Also, the return type is quite long, you can use `auto GetLocalVars() -> std::map<std::string, uint32>;`

src/map/entities/baseentity.cpp
zach2good: The reason I recommend passing back a reference here is because you have to make copies when you create the eventual Lua table. If you don't do this then you'll create 2x essentially full copies on the way over - just to chuck them away

src/map/entities/baseentity.h
TracentEden2: > The naming here isn't particularly helpful, `GetLocalVars` is fine, unless there are other functions that return a filtered version of the map

Thanks, I updated the PR, hopefully this is ok now.

src/map/lua/lua_baseentity.cpp
TracentEden2: > This creates copies and throws them away on every iteration of this loop.
> 
> `for (auto const& [varname, value] : ...`

Thanks, I updated the PR, hopefully this is ok now.

src/map/lua/lua_baseentity.cpp
TracentEden2: > Use `auto&` here

Thanks, I updated the PR, hopefully this is ok now.

src/map/entities/baseentity.cpp
TracentEden2: > This is returning a full copy of `m_localVars`, no reason you can't return a `&` to it

Thanks, I updated the PR, hopefully this is ok now.

src/map/lua/lua_baseentity.cpp
```diff
@@ -664,6 +664,28 @@ void CLuaBaseEntity::setVolatileCharVar(std::string const& varName, int32 value,
...
 }
 
+/************************************************************************
+ *  Function: getAllLocalVars()
+ *  Purpose : Returns all variables assigned locally to an entity
+ *  Example :
+ *  Notes   :
+ ************************************************************************/
+
+auto CLuaBaseEntity::getAllLocalVars() -> sol::table
```
zach2good: Make these function calls consistant with GetLocalVars

src/map/entities/baseentity.h
```diff
@@ -284,6 +284,8 @@ class CBaseEntity
     uint32 GetLocalVar(std::string var);
     void   SetLocalVar(std::string var, uint32 val);
 
+    std::map<std::string, uint32>& GetLocalVars();
```
zach2good: auto GetLocalVars() -> std::map<std::string, uint32>&;

src/map/lua/lua_baseentity.cpp
TracentEden2: > Make these function calls consistant with GetLocalVars

Thanks, I updated the PR!

src/map/entities/baseentity.h
TracentEden2: > auto GetLocalVars() -> std::map<std::string, uint32>&;

Thanks, I updated the PR!

zach2good (approved):
Core side is all fine 👍

Xaver-DaRed (approved):

PR #5281, opened 2024-03-19
[cpp] Allow samba to work if party leader is on another process

MowFord:
**_I affirm:_**

Sambas store the targid of the source, and anyone hitting a mob with a daze does a quick loop of their party members to see if they get the daze benefit.

Before this PR, the party leader had to be in the same zone cluster, or the `dynamic_cast` would fail and nobody (including the originator of the samba) would get a benefit

Use drain samba with multiple people in the party in a multi cluster setup. Have party leader `!zone` to somewhere in the other cluster and see samba effects continue working

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1381,26 +1381,32 @@ namespace battleutils
 
                 if (PAttacker->objtype == TYPE_PC && PAttacker->PParty != nullptr)
                 {
-                    if (auto* PLeader = dynamic_cast<CCharEntity*>(PAttacker->PParty->GetLeader()))
+                    if (auto* PChar = dynamic_cast<CCharEntity*>(PAttacker))
                     {
-                        PLeader->ForPartyWithTrusts([&](CBattleEntity* PMember)
-                                                    {
+                        // clang-format off
+                        PChar->ForPartyWithTrusts([&](CBattleEntity* PMember)
+                        {
                             if (attackerID == PMember->id)
                             {
                                 power = PDefender->StatusEffectContainer->GetStatusEffect(daze)->GetPower();
-                            } });
+                            }
+                        });
+                        // clang-format on
                     }
                 }
                 else if (PAttacker->objtype == TYPE_TRUST)
                 {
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):

MowFord:
CI final boss always gets me, should pass now

src/map/utils/battleutils.cpp
```diff
@@ -1381,10 +1381,10 @@ namespace battleutils
 
                 if (PAttacker->objtype == TYPE_PC && PAttacker->PParty != nullptr)
                 {
-                    if (auto* PLeader = dynamic_cast<CCharEntity*>(PAttacker->PParty->GetLeader()))
+                    if (auto* PChar = dynamic_cast<CCharEntity*>(PAttacker))
                     {
-                        PLeader->ForPartyWithTrusts([&](CBattleEntity* PMember)
-                                                    {
+                        PChar->ForPartyWithTrusts([&](CBattleEntity* PMember)
+                                                  {
```
zach2good: While you're here, can you wrap this in // clang-format on/off and align this lambda please?

src/map/utils/battleutils.cpp
```diff
@@ -1393,10 +1393,10 @@ namespace battleutils
                 }
                 else if (PAttacker->objtype == TYPE_TRUST)
                 {
-                    if (auto* PMaster = dynamic_cast<CCharEntity*>(PAttacker->PMaster))
+                    if (auto* PChar = dynamic_cast<CCharEntity*>(PAttacker->PMaster))
                     {
-                        PMaster->ForPartyWithTrusts([&](CBattleEntity* PMember)
-                                                    {
+                        PChar->ForPartyWithTrusts([&](CBattleEntity* PMember)
+                                                  {
```
zach2good: Same here

zach2good (approved):

Xaver-DaRed (approved):

zach2good:
Renamed cluster -> process on the way in

zach2good:
https://github.com/LandSandBoat/server/wiki/Miscellaneous-Server#what-is-multi-process

PR #5286, opened 2024-03-19
Stratagem time left correction at low levels

MowFord:
**_I affirm:_**

2 things
- low level scholar has stratagem charges, but only 1, so they function like a normal job ability
  - Before change a scholar with stratagems on cooldown would underflow the time left:
  
image

- when you have more stratagems the time reported by the chat log ended up being the time to a full set of stratagems (i think?)
  - This "Time Left" message is only ever given when you have zero charges, so we can infer the time until the next charge. This (incorrect time) was with 75 sch:
  
image

The new logic will 
- do unchanged behavior for JA that don't have charges
- treat abilities that have maxcharges == 1 the same as above (instead of underflowing)
- report "time left" for abilities with maxcharges > 1 to be the time until the next charge is available

Use stratagems at various levels of scholar and see that the time left reported matches the game client UI:
image


MowFord:
CI framed me, sigh.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

Xaver-DaRed:
Needs rebase

PR #5288, opened 2024-03-20
[core] Fix mob hundred fists delay

TracentEden2:
**_I affirm:_**

This PR changes the delay of mobs using hundred fists to be more accurate to actual retail captures. The captures by Siknoz are from a variety of mobs including from RoZ, CoP, ToAU, and WoTG (see captures [here](https://discord.com/channels/443544205206355968/674750598939279400/1220077549347016724)). 

(monk mobs are back!)

Fight mobs that use hundred fists and compare speed before and after this PR. The mobs should attack much faster.

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -311,10 +311,6 @@ bool CBattleEntity::Rest(float rate)
 int16 CBattleEntity::GetWeaponDelay(bool tp)
 {
     TracyZoneScoped;
-    if (StatusEffectContainer->HasStatusEffect(EFFECT_HUNDRED_FISTS) && !tp)
-    {
-        return 1700;
-    }
     uint16 WeaponDelay = 9999;
     if (auto* weapon = dynamic_cast<CItemWeapon*>(m_Weapons[SLOT_MAIN]))
     {
@@ -369,6 +365,13 @@ int16 CBattleEntity::GetWeaponDelay(bool tp)
         // This should be enforced on -delay equipment, martial arts, dual wield, and haste, hence MinimumDelay * 0.2.
         // TODO: Could be converted to value/1024 if the exact cap is ever determined.
         MinimumDelay -= (uint16)(MinimumDelay * 0.8);
+
+        // if hundred fists then use the min delay (as hundred fists also reduces base delay by 80%
+        if (StatusEffectContainer->HasStatusEffect(EFFECT_HUNDRED_FISTS) && !tp)
+        {
+            WeaponDelay = MinimumDelay;
+        }
+
         WeaponDelay = (WeaponDelay < MinimumDelay) ? MinimumDelay : WeaponDelay;
     }
     return WeaponDelay;


src/map/entities/battleentity.cpp
```diff
@@ -313,6 +313,12 @@ int16 CBattleEntity::GetWeaponDelay(bool tp)
     TracyZoneScoped;
     if (StatusEffectContainer->HasStatusEffect(EFFECT_HUNDRED_FISTS) && !tp)
     {
+        if (this->objtype == ENTITYTYPE::TYPE_MOB)
+        {
+            // Captures show mobs swing around 700 delay under hundered fists
```
TeoTwawki: Hundred fists doesn't actually set a specific delay value, it causes a reduction that some sources say is 75% and other say is 80%

To the best of my knowledge this is true for both players and mobs, the difference being mobs don't have added delay from a h2h weapon and are -usually- using the basic unarmed h2h delay as their basis (there exist weird cases like non monk non h2h mobs using it). So in theory, whatever delay the mob already had looked like the values seen after hundred fists delay reduction applied. But static setting a value is going to give edge cases down the road even though you are improving it over the previous wrong static value, and so it might be better to either work out the correct math for the reduction -or- prove and document mobs work differently right now to justify the static value once and for all.

...Also the fact players are getting a static value when the reduction is pretty well known kills me.

TeoTwawki: ## Note, these quotes are for players (highlighting that the 1700 isn't correct for players either)

old clopedia wiki : 
> Upon usage, character's melee attack delay is greatly reduced, similar to a massive Haste effect. The effect does not kick in until the next melee strike is made, as attack delay is calculated immediately following each attack round.
> - Delay is not reduced to zero, or any other set amount. Attack rate is still based on the delay of your weapon. Higher delay weapons will still attack slower than lower delay weapons will.
>
> Gives the effect of 80% haste, not affected by magical hastes (i.e. marches, haste spell) and also not affected by gear haste.
> - During Hundred Fists, it is recommended to replace Haste gear with Accuracy / Attack / STR gear, to maximize damage output.
> - Completely unaffected by all forms of slow, including raise weakness

BG wiki :
> Reduces delay by 75% after calculated delay
> - Calculated delay is delay after including Weapon Delay and Martial Arts traits and gear.
> - Because your post-Arts delay is used, it is possible to break the normal 80% Attack Speed cap with this ability when using Hand-to-Hand weapons.

wiki.ffo.jp :
> マーシャルアーツ による隔の減算をした後の攻撃間隔が1/4になる。この時の攻撃間隔]はヘイストキャップ80%を超えて短縮される
マーシャルアーツ効果アップなどのような得TP減少は無い（隔マイナスではない）
全てのヘイスト・スロウ効果が無視される

(translated) :
> After subtracting the distance from martial arts, the attack interval becomes 1/4. At this time, the attack interval] will be reduced by more than 80% of the haste cap.
There is no reduction in TP gain like when martial arts effects are increased (not a negative difference)
All haste and slow effects are ignored.

TeoTwawki: PS, pretty sure this is milliseconds and not actual delay, because 1700 delay is insanely long 

1700 ms is massive too, but gets divided to something more reasonable

I only mention this because the comment says "700 delay". Even 700 delay would be huge.  You may want to amend that to "about 700ms".

Really kills me that the core has multiple cases of using the word delay without obvious distinguishing between weapon delay units and milliseconds. (see issue 2502...)

TracentEden2: > Hundred fists doesn't actually set a specific delay value, it causes a reduction that some sources say is 75% and other say is 80%
> 
> To the best of my knowledge this is true for both players and mobs, the difference being mobs don't have added delay from a h2h weapon and are -usually- using the basic unarmed h2h delay as their basis (there exist weird cases like non monk non h2h mobs using it). So in theory, whatever delay the mob already had looked like the values seen after hundred fists delay reduction applied. But static setting a value is going to give edge cases down the road even though you are improving it over the previous wrong static value, and so it might be better to either work out the correct math for the reduction -or- prove and document mobs work differently right now to justify the static value once and for all.
> 
> ...Also the fact players are getting a static value when the reduction is pretty well known kills me.

Yeah the mob captures suggest that the HF value is either a fixed delay of 85 or if you assume the base delay of monk mobs is 480 (similar to players) then 85 would be about an 80% reduction in delay. 

The captures also suggest that (at least for mobs) the base delay used does not take martial arts into account. For example Hippolytos (with MA3) showed pre-HF delay of 364 and a HF delay of  84. While Tieholtsodi (with MA5-6) has pre-HF delay of 213 and also a HF-delay of 85.

I would be ok with using an 80% delay reduction from the base delay for mobs instead of fixed value.

TeoTwawki: > The captures also suggest that (at least for mobs) the base delay used does not take martial arts into account.

thats how it is for players 2 - all other forms of delay reduction get ignored, hf becomes the only reduction in play, but its massive

edit: I misread part of this the 1st time, trait matters for players but the way traits happen on mobs isn't the same on retail.  all reductions are simply trumped by hf tho to the point there is zero benefit from haste and -delay from augments simply gimps your tp.

LacunaXI: > Hundred fists doesn't actually set a specific delay value, it causes a reduction that some sources say is 75% and other say is 80%
> To the best of my knowledge this is true for both players and mobs, the difference being mobs don't have added delay from a h2h weapon and are -usually- using the basic unarmed h2h delay as their basis (there exist weird cases like non monk non h2h mobs using it). So in theory, whatever delay the mob already had looked like the values seen after hundred fists delay reduction applied. But static setting a value is going to give edge cases down the road even though you are improving it over the previous wrong static value, and so it might be better to either work out the correct math for the reduction -or- prove and document mobs work differently right now to justify the static value once and for all.

Is this video of any use?

https://www.youtube.com/watch?v=oW8sbhyEhLg

TeoTwawki: Youtube videos are practically useless for determining delays. Even if I trusted doing some kind of frame counting or clock watching and assumed nothing happened to the playback rate or frame rate, all we'd get is an approximate time between hits and not how that time was determined. 

I do not doubt that the 700ms was about or even exactly the time between punches here in the examples the PR author was working with, we're discussing the methods it got to that value from the normal delay before hundred fists was activated.

TeoTwawki: @TracentEden 
> I would be ok with using an 80% delay reduction from the base delay for mobs instead of fixed value.

I would be ok merging that, for both players and monsters.

TracentEden2: > @TracentEden
> 
> > I would be ok with using an 80% delay reduction from the base delay for mobs instead of fixed value.
> 
> I would be ok merging that, for both players and monsters.

Yeah, I am asking capture person to check about martial arts and player HF (as that is mentioned in the wikis you linked), thus can make sure both player and mob logic is correct.

zach2good:
Any movement on this, are we waiting for a confirmation of a % reduction in base delay for both mobs and players?

TracentEden2:
> Any movement on this, are we waiting for a confirmation of a % reduction in base delay for both mobs and players?

Ok, Siknoz has now validated that martial arts does not impact player HF (see this [capture](https://youtu.be/ZRbXEuDulHY) for example), so this PR should be ready to go. Thanks!

WinterSolstice8:
Seems like I can't kick CI to restart for some reason. you might need to force push an amended commit

src/map/entities/battleentity.cpp
WinterSolstice8: There's a typo here, `hundered ` -> `hundred`
that should be a justfiable commit amend to kick CI haha

TracentEden2: > There's a typo here, `hundered ` -> `hundred` that should be a justfiable commit amend to kick CI haha

Ok, should be good now. Thanks!

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5291, opened 2024-03-21
[core] Allow interruption of certain player events when being attacked

TracentEden2:
**_I affirm:_**

This PR adds interruptibility to certain events when the player is the target of an aggressive action (like an attack or mob ability). On retail such events are primarily menu-based events like when confirming the use of a teleporter in sky or sea. The retail behavior (and event interruption) can be seen for example in this [retail video](https://www.youtube.com/watch?v=qWCVSUmPX1Q). 

Specifically the PR moves release function from lua base entity to char entity so it can be easily called in battle entity code (such as onAttack). The battle entity code itself checks in several places for aggressive actions (at the same locations where detectable status effects (like invis) are removed from targets of aggressive actions), and if the player is not locked in a CS then it releases the player from the event.

The current PR is a draft to get comments as I am not an expert on events.

Attempt to use different teleporters and NPCs while being attacked by mobs.

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -1757,6 +1757,14 @@ void CBattleEntity::OnCastFinished(CMagicState& state, action_t& action)
         // Should not be removed by AoE effects that don't target the player or
         // buffs cast by other players or mobs.
         PActionTarget->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DETECTABLE);
+
+        if (auto PChar = dynamic_cast<CCharEntity*>(PActionTarget))
+        {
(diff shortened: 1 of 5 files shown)


zach2good:
> certain events

I can't see any logic for _which_ events are interruptable? In terms of sending packets to the client, there doesn't seem to be a difference between an event that's just a menu and one that's a "cutscene", so we need to have the logic to handle it.

It's been a long time since I looked, but this is what `startCutscene` and `startOptionalCutscene` are for.

zach2good:
We're very close to actually already having this working, and it's all based around those bindings I mentioned. I'll port the "recent" changes Eden has to this system (maybe 2 years old) around character locking and non-"Cutscene" skipping

PR #5292, opened 2024-03-21
[cpp] Add mobskill bindings for skillchain properties

MowFord:
**_I affirm:_**

Currently, mobskills can skillchain, but only if they're defined in the db to do so. This binding would allow the ability to set skillchain properties for a mobskill via lua. The way that mobskill states are coded though, we have to take care when these two new bindings are called
- if they are called in a check function, before the mob goes into the mobskill state, it would adjust globally all mobskills
  - i.e. `final_sting.lua` inside `mobskillObject.onMobSkillCheck` calling `skill:setSkillchainProps(xi.skillchainType.FUSION, xi.skillchainType.COMPRESSION, xi.skillchainType.SCISSION)` (even with a conditional) would make every future final sting have those sc properties. 
  - Or more specifically if we did something like create a `GetMobskillByID` function
- Some plumbing in `luautils` is adjusted to make the cloned mobskill be passed through to each call of `damage = luautils::OnMobWeaponSkill(...`, allowing the adjustments of a mobskill to be reflected in calls to `lua["xi"]["zones"][zone]["mobs"][name]["onMobWeaponSkill"]` and `lua["xi"]["actions"]["mobskills"][mobskill_name]["onMobWeaponSkill"]`
  - This means that for particular mobs, we can do the following:

something like this in their mob lua:
```
entity.onMobWeaponSkill = function(target, mob, skill, action)
    if skill:getID() == 336 then
        skill:setSkillchainProps(xi.skillchainType.FUSION, xi.skillchainType.COMPRESSION, xi.skillchainType.SCISSION)
    end
end
```
or this in the final_sting lua
```
    if mob:getName() == 'Digger_Wasp' then -- or perhaps if the mob has a certain buff mob:getStatusEffect?
        skill:setSkillchainProps(xi.skillchainType.FUSION, xi.skillchainType.COMPRESSION, xi.skillchainType.SCISSION)
    end
```
and make final sting create fusion for Digger Wasps, for example

I am not well-versed in partciular sc properties and how they chain together, so the above example is purely theoretical.

The primary use-case would be for bst jug pets. Alternative uses would be for partciular battlefield encounters where the NMs should skillchain using regular mobskills (ark angels, Bozzetto Enceladus, Bozzetto Shade, certain tonberries), or just modules for generic custom content.

Digger Wasp in action self-skillchaining at low hp using only lua inside the Digger_Wasp lua file:

image

The end result is that all existing functionality is untouched, but the above workflows can be utilized to change sc properties on a mobskill

`print(fmt("{}-{}-{}", skill:getSkillchainProps()))` in the `onMobWeaponskill` function of any particular skill now shows the active sc properties that will apply to the target. This can be used to verify mob/ws level lua changes.

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -1961,7 +1961,7 @@ void CBattleEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
         }
         else
         {
-            damage = luautils::OnMobWeaponSkill(PTargetFound, this, PSkill, &action);
+            damage = luautils::OnMobWeaponSkill(PTargetFound, this, &PSkill, &action);
             this->PAI->EventHandler.triggerListener("WEAPONSKILL_USE", CLuaBaseEntity(this), CLuaBaseEntity(PTargetFound), PSkill->getID(), state.GetSpentTP(), CLuaAction(&action), damage);
             PTarget->PAI->EventHandler.triggerListener("WEAPONSKILL_TAKE", CLuaBaseEntity(PTargetFound), CLuaBaseEntity(this), PSkill->getID(), state.GetSpentTP(), CLuaAction(&action));
         }
--- a/src/map/lua/lua_mobskill.cpp
+++ b/src/map/lua/lua_mobskill.cpp
@@ -50,6 +50,27 @@ void CLuaMobSkill::setMsg(uint16 message)
     m_PLuaMobSkill->setMsg(message);
 }
 
+std::tuple<uint8, uint8, uint8> CLuaMobSkill::getSkillchainProps()
+{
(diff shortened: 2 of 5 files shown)


TeoTwawki:
we previously had bindings for this and they were removed.

MowFord:
Do you recall why?

TeoTwawki:
> Do you recall why?

lots of things core side already, was redundant and simpler to have everything pre-written into the that to execute additional calls every time we ran the skill. now if we start moving those core side things to lua..it starts to be a different picture. but right now all the cases where a skill would only cause a chain "sometimes" has a condition paired with some status effect or JA etc. in core. (think blue magic - I could see utility there)

TeoTwawki:
I will say I do like that you put all 3 sc props in one call. previous form had a separate call per tier, very annoying

MowFord:
This is working code, but I fully expect to need some protections in the `CLuaMobSkill::setSkillchainProps` function. I wouldn't want a simple mistake in calling the function cause a map crash

MowFord:
rebased the branch

PR #5293, opened 2024-03-21
[ENUMS] Add entries to item.lua

hooksta4:
**_I affirm:_**

Adds entries into item.lua in preperation for future quests to be converted to IF and subsequently, called by xi.item.XXX vice itemID. 

Changed shoalweed to clump_of_shoalweed in item.lua to align with naming conventions.
Adjusted quest [scripts/quests/windurst/SOB7_The_Promise.lua] reference to shoalweed to clump_of_shoalweed

None required.


TeoTwawki:
You need 2 items for your script, but you've added far more in the enum. generally we were not trying to put every item into it - just what we needed. If that changes in the future we'll just automate it.

hooksta4:
@TeoTwawki  totally tracking - this is in preparation to push quite a few quests that are converted to IF along with some mission calls.

TeoTwawki:
> totally tracking - this is in preparation to push quite a few quests that are converted to IF along with some mission calls.

issue with that is the diff is to large - maybe others will but I'm not manual reviewing 512 new rows in that lua table
image

hooksta4:
Totally get it. Let me know what you all would like me to do in the mean time.

claywar:
If this were just updating shoalweed to use its proper long name, I'd approve in a heartbeat since it definitely doesn't match expected long name; however, there's a lot of other unrelated items here that don't conform with the intent.  I have shared a script to auto-generate entries and recreate that enum table given added itemids, and if you do not have it, please let me know.

scripts/enum/item.lua
```diff
@@ -1992,16 +2114,20 @@ xi.item =
...
     HI_POTION_P3                        = 4119,
     ETHER                               = 4128,
     ETHER_P1                            = 4129,
+    ETHER_P2                            = 4130,
+    ETHER_P3                            = 4131,
     HI_ETHER                            = 4132,
-    HI_ETHER_P1                         = 4133,
-    HI_ETHER_P2                         = 4134,
-    HI_ETHER_P3                         = 4135,
+    HI_ETHER_I                          = 4133,
```
claywar: Hard no.  This scheme exists for a reason, and we do not use roman numerals when the db is +N

claywar (changes requested):

scripts/enum/item.lua
hooksta4: Fair enough - and I get it but these enums were pulled from quests, missions, loot tables, dialog etc that will be submitted in the following weeks. prepping the battlefield so to speak so there are not 5 quests that all inevitably have merge conflicts because they sit for a while.

TeoTwawki: > Fair enough - and I get it but these enums were pulled from quests, missions, loot tables, dialog etc that will be submitted in the following weeks. prepping the battlefield so to speak so there are not 5 quests that all inevitably have merge conflicts because they sit for a while.

That type of merge conflict should be trivial to deal with. Even editing the same file doesn't guarantee a conflict though. Usually its when 2 commits edit the same line. There is a good chance that if each quest PR contained the enumerations that specific quest needed that they'd actually make it through without a conflict, or trigger only the "made you look" type of conflict where git just makes you open it and tell it "this is fine" and there was no fire, save and resolve and go on your way. 

Keep this in mind going forward, even if another reviewer lets it pass this time. 

----

Also:

While we don't like preemptive adding to this particular enum, in other cases a way to avoid giant nearly unreviewable diffs is to break it into multiple commits (we're able to look at a specific commit). Its still better to just do it in multiple PRs but some cases may justify just packing 4-5 commits onto top each other (see the many many past item sql prs that have been happening - if it were all in a single commit in a single pr I'd flat out reject).

PR #5297, opened 2024-03-23
[core] Prevent sending bazaar browsing packets for hidden GMs

TracentEden2:
**_I affirm:_**

This PR prevents sending bazaar packets to players if the person checking the bazaar is a hidden GM. This change was requested by GMs on Horizon as they wanted to be able to check the players bazaar without letting the player know there is a hidden GM observing them. Note importantly if the GM purchases an item from the bazaar the player is still notified.

This is a PR from ASB/Horizon coming upstream.

The change can be tested by having a hidden GM try to interact with the bazaar of a non-GM character or a GM char of an equal or lower level.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -7767,7 +7767,10 @@ void SmallPacket0x104(map_session_data_t* const PSession, CCharEntity* const PCh
                 PTarget->BazaarCustomers.erase(PTarget->BazaarCustomers.begin() + i--);
             }
         }
-        PTarget->pushPacket(new CBazaarCheckPacket(PChar, BAZAAR_LEAVE));
+        if (!PChar->m_isGMHidden || (PChar->m_isGMHidden && PTarget->m_GMlevel >= PChar->m_GMlevel))
+        {
+            PTarget->pushPacket(new CBazaarCheckPacket(PChar, BAZAAR_LEAVE));
+        }
     }
     PChar->BazaarID.clean();
 }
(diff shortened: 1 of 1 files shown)


TeoTwawki:
Does the PR eliminate just the browsing messages or the buying ones as well?

eliminating the "name is looking at you/your bazaar" is probably a good idea, but likely we'd better keep the actual purchase message to avoid mysteriously vanishing item confusion on the player side. Hidden GMs shouldn't perform actions they -know- are going to reveal themselves and I would expect buying items to do so plus..uh...why is a GM buying items? yeah.

TracentEden:
> Does the PR eliminate just the browsing messages or the buying ones as well?
> 
> eliminating the "name is looking at you/your bazaar" is probably a good idea, but likely we'd better keep the actual purchase message to avoid mysteriously vanishing item confusion on the player side. Hidden GMs shouldn't perform actions they -know- are going to reveal themselves and I would expect buying items to do so plus..uh...why is a GM buying items? yeah.

Yeah it eliminates also the buying notification, I am will ask GMs about what type of situation that would be and let you know.

TracentEden:
> Does the PR eliminate just the browsing messages or the buying ones as well?
> 
> eliminating the "name is looking at you/your bazaar" is probably a good idea, but likely we'd better keep the actual purchase message to avoid mysteriously vanishing item confusion on the player side. Hidden GMs shouldn't perform actions they -know- are going to reveal themselves and I would expect buying items to do so plus..uh...why is a GM buying items? yeah.

Ok, I checked and GMs say they generally do not need this hidden buying functionality as just browsing bazaar gives the info they need. I removed that part from the PR. Thanks!

WinterSolstice8 (approved):

Xaver-DaRed (approved):

zach2good (approved):

PR #5298, opened 2024-03-23
[lua]Fix merit stat enum offsets

MowFord:
**_I affirm:_**

I didn't full audit, but spot checked and other categories seemed good. I suspect this was never noticed since most of the time these categories aren't being looked at in lua. There was a missing category in the HP/MP group and the offsets for attributes were not offset properly. Used [this as reference](https://github.com/LandSandBoat/server/blob/8ca07a25d92b820a5b1985402f8f670ca8d8c893/src/map/merit.h#L115), and it matched up perfectly (of course)

Update the enum in this screenshot instead of using raw offsets:

image

### DIFF
--- a/scripts/enum/merit.lua
+++ b/scripts/enum/merit.lua
@@ -74,15 +74,16 @@ xi.merit =
     -- HP
     MAX_HP                      = meritCategory.HP_MP + 0x00,
     MAX_MP                      = meritCategory.HP_MP + 0x02,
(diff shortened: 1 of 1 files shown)


TeoTwawki (approved):
looks like src and lua got out of sync, presumably someone updated one without realizing the opposite existed

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5299, opened 2024-03-23
small change to Vrtra to remove full localvar reset

MowFord:
**_I affirm:_**

Adjusts Vrtra's engage function to not wipe all localvars. There's only 2 used in onMobFight to track progress of the fight, a full localvar reset clears many unrelated things.

Specifically, our custom claimshield and rage timer mechanics on CEXI.

### DIFF
--- a/scripts/zones/King_Ranperres_Tomb/mobs/Vrtra.lua
+++ b/scripts/zones/King_Ranperres_Tomb/mobs/Vrtra.lua
@@ -7,7 +7,9 @@ local entity = {}
 local offsets = { 1, 3, 5, 2, 4, 6 }
 
 entity.onMobEngage = function(mob, target)
-    mob:resetLocalVars()
+    -- Reset the onMobFight variables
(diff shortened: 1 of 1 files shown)


Frankie-hz:
If you want I believe Tiamat and Jorm should have the same issues in case of a wipe/re-engage.

MowFord:
> If you want I believe Tiamat and Jorm should have the same issues in case of a wipe/re-engage.

they don't resetlocalvars, but are you saying they don't reset their behavior on wipe?

Frankie-hz:
> > If you want I believe Tiamat and Jorm should have the same issues in case of a wipe/re-engage.
> 
> they don't resetlocalvars, but are you saying they don't reset their behavior on wipe?

Sorry yes. We had to add the reset twohour/changetime on those two as well because of the same reason.

MowFord:
Simple enough, out of scope for this one imo though

WinterSolstice8 (approved):

PR #5303, opened 2024-03-25
[core] Cleanup allies of battlefields even if they are not alive

TracentEden2:
**_I affirm:_**

This PR removes the condition that allies must be alive to be cleaned up at the clean up stage of battlefields. Currently this condition causes issues with allies that can be dead but do not despawn (such as Prishe in Dawn fight), for example causing players to see multiple extra dead Prishes (from previous groups).

Enter Dawn fight and engage Promathia, then !hp 0 Prishe and leave the battlefield. Then reenter and see that there is not an extra dead Prishe on the ground.

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -628,7 +628,9 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
                 {
                     if (std::find(m_AllyList.begin(), m_AllyList.end(), PMobEntity) != m_AllyList.end())
                     {
-                        if (PMobEntity->isAlive() && PMobEntity->PAI->IsSpawned())
+                        // We should not put an isAlive check here because some ally can be dead at cleanup
+                        // but not despawned (for example Prishe in Dawn fight)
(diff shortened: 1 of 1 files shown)


TeoTwawki:
> causing players to see multiple extra dead Prishes

Prishe: these parallel universe incidents are ticking me off! _3 taru's run and scatter_

src/map/battlefield.cpp
```diff
@@ -628,7 +628,9 @@ bool CBattlefield::RemoveEntity(CBaseEntity* PEntity, uint8 leavecode)
                 {
                     if (std::find(m_AllyList.begin(), m_AllyList.end(), PMobEntity) != m_AllyList.end())
                     {
-                        if (PMobEntity->isAlive() && PMobEntity->PAI->IsSpawned())
+                        // We should not put an isAlive check here because some ally can be dead at cleanup
+                        // but not despawned (for example Prise in Dawn fight)
```
TeoTwawki: missed the h in Prishe

TracentEden2: > missed the h in Prishe

Thanks, fixed!

WinterSolstice8 (approved):

PR #5310, opened 2024-03-27
Implementation of nightmare sleep mechanics

### LINKED ISSUE #851: Missing support for "Nightmare" status effect
- [x] I have specified what branch this happens on branch: EDITME

https://ffxiclopedia.fandom.com/wiki/Nightmare
https://www.bg-wiki.com/ffxi/Nightmare
https://www.bg-wiki.com/ffxi/Waking_Dreams

MowFord:
**_I affirm:_**

Resolves #851 .

player avatar using nightmare on a mob:
image

the `no_effect` mob is one that already had nightmare sleep.

image

after nightmare applied, see that there's the attp reduction as well as hp reduction (hp goes from 240 to 238), which implies the bio effect is successfully applied

sleep not being broken by helix:

image

Finally, a really hacky way to test the mobskill side of things:

image

Force a mob to use nightmare and see that the player gets sleep and bio and doesn't get woken by attacks

Onto the details:
- Nightmare abilities apply a sleep and a bio effect: evidence in the linked issue #851
- Nightmare sleep is unbroken by any DoT
  - NM/Mob version also is unbroken by _all_ damage sources: https://www.bg-wiki.com/ffxi/Promathia_Mission_3-5
- The Bio effect from Nightmare does not break any sleep effects

Run through the steps above. Code for waking a slept player is unchanged but confirming that nightmare sleep is woken up by Cures would be a plus

### DIFF
--- a/scripts/actions/abilities/pets/nightmare.lua
+++ b/scripts/actions/abilities/pets/nightmare.lua
@@ -1,0 +1,45 @@
+-----------------------------------
+-- Nightmare - Player's Avatar
+-- AoE Sleep
+-- Sleep that is not broken from DoT effects (any dmg source that doesn't break bind).
+-- This version of it is from a player's avatar. The sleep is broken by most damage sources except other DoTs
+--
+-- see mobskills/nightmare.lua for full explanation
+-----------------------------------
+local abilityObject = {}
+
+abilityObject.onAbilityCheck = function(player, target, ability)
+    return 0, 0
+end
+
+abilityObject.onPetAbility = function(target, pet, skill)
+    local duration = 90
+    local dotdamage = 2
+    local sleepTier = 1
+    local dINT = pet:getStat(xi.mod.INT) - target:getStat(xi.mod.INT)
+    local bonus = xi.summon.getSummoningSkillOverCap(pet)
+    local resm = xi.mobskills.applyPlayerResistance(pet, -1, target, dINT, bonus, xi.element.DARK)
+    if resm < 0.5 then
+        skill:setMsg(xi.msg.basic.JA_MISS_2) -- resist message
+        return xi.effect.SLEEP_I
+    end
+
+    duration = duration * resm
+    if
+        target:hasImmunity(1) or
+        target:hasStatusEffect(xi.effect.SLEEP_I) or
+        target:hasStatusEffect(xi.effect.SLEEP_II) or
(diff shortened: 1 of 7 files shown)


scripts/actions/abilities/pets/nightmare.lua
```diff
@@ -0,0 +1,45 @@
...
+    local bonus = xi.summon.getSummoningSkillOverCap(pet)
+    local resm = xi.mobskills.applyPlayerResistance(pet, -1, target, dINT, bonus, xi.element.DARK)
+    if resm < 0.5 then
+        skill:setMsg(xi.msg.basic.JA_MISS_2) -- resist message
+        return xi.effect.SLEEP_I
+    end
+
+    duration = duration * resm
+    if
+        target:hasImmunity(1) or
```
WinterSolstice8, line 29: `1`? is there no enum for this?

TeoTwawki: one exists in [battleentity.h](#L457) but becaus eit hasn't been added in scripts/enum yet there are both magic numbers as well commented out lua that probably shoudln't even exists https://github.com/search?q=repo%3ALandSandBoat%2Fserver+immunity+language%3ALua&type=code&l=Lua because those mobs pools should be doing the deed 

tl;dr we need to add a new file to the enum folder or start coming up with methods that share C++ enums to lau or lua enums to C++ so we stop having things in 2 places

WinterSolstice8: We can do a pass to fix immunity enum later

src/map/status_effect_container.cpp
```diff
@@ -792,6 +792,16 @@ void CStatusEffectContainer::DelStatusEffectsByFlag(uint32 flag, bool silent)
     {
         if (PStatusEffect->HasEffectFlag(flag))
         {
+            // If this is an NM/Mob Nightmare sleep, it can be removed explictly by a cure
+            // see mobskills/nightmare.lua for full explanation
```
WinterSolstice8, line 796: What is the mechanism that cure uses to bypass this check?

MowFord: This block is only for deleting by `damage` flag. A cure would wake via a different means

WinterSolstice8 (approved):

PR #5311, opened 2024-03-27
Ensure mobs don't feedback loop and spam TP moves at low hp

MowFord:
**_I affirm:_**

Mobs can get into a feedback loop where they will spam TP moves without even trying to do other things (include move at all)

example of this would be king behemoth below 25% hp with a small amount of regain. Essentially he spams flame armor no matter what his target is.

The main issue lies in the timestamp `m_LastMobSkillTime` never coming into play. We create a new public function in `mob_controller` to update that timestamp. Simply updating it won't actually change behavior though since we always compare `m_Tick` with that timestamp before checking TPUseChance, ~so we also add 0-5s RNG. This timestamp is then tapped when entering the mobskill state, as well as exiting the mobskill state (the former in case the mobskill gets interrupted, TP is consumed so we should resume the same behavior).~

Edit: Updated to use a flat mobmod, default to 1s. After discussing with a few people this mobmod should _probably_ be set to the attack delay of the mob, but 1s works well and being a mobmod allows specific mobs to be tuned separately. 

Finally, added the condition that TPUseChance is 100% if below 50% HP and above 2k tp

This has _always_ been an issue with DSP, etc... but the recent update to allow mobs to gain TP from their mobskills has made it even more apparent.

~KB is a very extreme example, but with before the chnages this same setup he would literally stand in place:~

https://imgur.com/Gkhaco3

~I suspect there is some hard minimum for mobskills, but random 1s-6s delay before even considering TPUseChance seems reasonable~

Note that these changes only affect the `mob_controller` automatic selection of mobskills. `mob:useMobAbility` is unaffected

`!getmobmod 78` to see the default behavior being used, set to different values with high regain to see the limit in action

### DIFF
--- a/scripts/enum/mob_mod.lua
+++ b/scripts/enum/mob_mod.lua
@@ -85,4 +85,5 @@ xi.mobMod =
     CAN_PARRY              = 75, -- Check if a mob is allowed to have parry rank (Rank Value 1-5)
     NO_WIDESCAN            = 76, -- Disables widescan for a specific mob
     TRUST_DISTANCE         = 77, -- TRUSTS ONLY: Set movement type/distance. See trust.lua for details.
+    MOBSKILL_DELAY         = 78, -- Delay (in ms) before using a mobskill after using a mobskill.
 }
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -1202,6 +1202,13 @@ void CMobController::TapDeclaimTime()
     m_DeclaimTime = m_Tick;
 }
 
+// To ensure mobskills don't get spammed with high regain/low hp.
+// This affects the automatic mobskill selection to give 0% chance until after this time
+void CMobController::TapLastMobSkillTime()
+{
+    m_LastMobSkillTime = m_Tick + std::chrono::milliseconds(PMob->getMobMod(MOBMOD_MOBSKILL_DELAY));
+}
+
 bool CMobController::Cast(uint16 targid, SpellID spellid)
 {
     TracyZoneScoped;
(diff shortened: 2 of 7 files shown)


TeoTwawki:
> This has _always_ been an issue with DSP, etc... but the recent update to allow mobs to gain TP from their mobskills has made it even more apparent.

Great so theres an issue report somewhere right? Right? >.>

> so we also add 0-5s RNG

A made up random delay of up to 5 whole seconds (which is HUGE) is def not the direction I want to see a fix take.

MowFord:
> A made up random delay of up to 5

Correct, I did this as a stop-gap to get the conversation started with a PoC. I spent a bit of time researching last night and found no definitive answer. I think perhaps the simpler solution is to not have the guaranteed mobskill for 25%/50% hp within a timeframe after a mobskill finishes.

i.e. 

- `TPUseChance` check becomes (pseudo-code)
```
if (health.tp == 3000 || (m_Tick > m_LastMobSkillTime + 5s && (GetHPP() <= 25 && health.tp >= 1000) || (GetHPP() <= 50 && health.tp >= 2000)))
{ ...
```
- and `TapLastMobSkillTime()` becomes just like the other Tap functions in this controller and updates to the current time

This would make the time within 5s of a mobskill reliant on only the mobmod for TP use-chance. I'm just not sure how one would properly verify this, but I do know that there's an issue here.

WinterSolstice8:
`example of this would be king behemoth below 25% hp with a small amount of regain.`
how small are we talking? the imgur video looks like it has a ton of regain?

MowFord:
> `example of this would be king behemoth below 25% hp with a small amount of regain.` how small are we talking? the imgur video looks like it has a ton of regain?

I didn't do the math but I would guess 500. At any rate the recent change to give tp from AOE moves means it didn't need and regain, this was just an example of when it would happen. 

There's likely an internal cooldown that exists instead of the rand(1,6) I have in this PR. I think players are 1s so we could adjust to that but currently the last mob skill timepoint is completely unused lol

MowFord:
Note that i've updated the OP with a rework of this fix

Issue #863, opened 2021-10-25
Worm shouldn't cast in melee range


ShiyoKozuki:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: EDITME

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 

family Worm mixin

```
        local targ = mob:getTarget()
        if targ then
            if mob:checkDistance(targ) < 3 then
               mob:SetMagicCastingEnabled(false)
            else 
                mob:SetMagicCastingEnabled(true)       
            end
          end
                 

```

TeoTwawki:
For desert worms, your assertion is correct. But its not every worm game wide. Worm behaviors aren't universally consistent in retail: Many will still cast, but less often.

Related is an issue about caster mobs opening their aggro with an immediate cast. 

I had started to document it back in an old topaz issue before someones meltdown demotivated me. topaz next imported that issue but I could not longer edit my post and had already been away from the task too long. https://github.com/topaz-next/topaz/issues/2222#issuecomment-747261721

Worms in Koroloka and Qufim will point blank nuke (or rasp) me if they are alive long enough to do so. 
_(note to self: go time it and see if timing is constant too)_

Xaver-DaRed:
So, this is only true for certain worms, is what I got from digging arround Teo's comments in different PRs and chats.

PR #5313, opened 2024-03-27
[core] Fix two worm casting issues

TracentEden2:
**_I affirm:_**

This PR fixes two issues with worms casting. Firstly, worms should only cast when their target is outside of melee range (see [here](https://www.bg-wiki.com/ffxi/Category:Worm)). Secondly, worms should not go underground during the middle of casting (as this interrupts their casting).

To test the first change go fight any worm and move into and out of melee range to see casting behavior. To test the second change go to for example the Phantom Worm spawn area (which has a lot of worms close together) and watch to see that they no longer interrupt their own casting.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -498,6 +498,15 @@ bool CMobController::CanCastSpells()
         }
     }
 
+    // worms only cast if target is outside of melee range
+    if (PMob->m_Family == 258)
+    {
+        if (PTarget && distance(PMob->loc.p, PTarget->loc.p) <= PMob->GetMeleeRange())
+        {
+            return false;
+        }
+    }
+
     return IsMagicCastingEnabled();
 }
 
@@ -960,7 +969,7 @@ void CMobController::DoRoamTick(time_point tick)
                                                                             (uint8)PMob->getMobMod(MOBMOD_ROAM_TURNS), PMob->m_roamFlags))
                 {
                     // #TODO: #AIToScript (event probably)
-                    if (PMob->m_roamFlags & ROAMFLAG_WORM)
+                    if (PMob->m_roamFlags & ROAMFLAG_WORM && !PMob->PAI->IsCurrentState<CMagicState>())
                     {
                         // move down
                         PMob->animationsub = 1;


TeoTwawki:
> This PR fixes two issues with worms casting. Firstly, worms should only cast when their target is outside of melee range (see [here](https://www.bg-wiki.com/ffxi/Category:Worm)). 

this has come up before and its not strictly true, last time I created video proof of the fact not all worms use identical behavior

they can in fact, point blank you in the face or buff/debuff depending which worms/which area

This needs handled in a less global way, if you just do something hardcoded to the entire family you force all worms to behave the same and that's not retail correct thats just shit wiki says.

> Secondly, worms should not go underground during the middle of casting (as this interrupts their casting).

they shouldn't be going underground while still having a target they are engaged to that the casting requires in the 1st place

TeoTwawki:
took me a minute to find, the issue for tracking it got destroyed 2 project iterations ago.

https://github.com/topaz-next/topaz/issues/2222#issuecomment-747261721

https://github.com/LandSandBoat/server/issues/863#issuecomment-950519943

> Worms in Koroloka and Qufim will point blank nuke (or rasp) me if they are alive long enough to do so.

TracentEden2:
> > This PR fixes two issues with worms casting. Firstly, worms should only cast when their target is outside of melee range (see [here](https://www.bg-wiki.com/ffxi/Category:Worm)).
> 
> this has come up before and its not strictly true, last time I created video proof of the fact not all worms use identical behavior
> 
> they can in fact, point blank you in the face or buff/debuff depending which worms/which area
> 
> This needs handled in a less global way, if you just do something hardcoded to the entire family you force all worms to behave the same and that's not retail correct thats just shit wiki says.
> 
> > Secondly, worms should not go underground during the middle of casting (as this interrupts their casting).
> 
> they shouldn't be going underground while still having a target they are engaged to that the casting requires in the 1st place

Ah, ok, yeah I will try to make it into a less global solution. 

For the second issue, they are not going underground while engaged to a target but while they are casting buffs on themselves like stoneskin.

TeoTwawki:
> For the second issue, they are not going underground while engaged to a target but while they are casting buffs on themselves like stoneskin.

ah ok yeah, didn't realize that was happening. they need to wait till cast stops to resume their "roaming". thats true of other mobs as well they don't have that step of changing state before stepping/moving tho.

PR #5314, opened 2024-03-28
[sql] "??? Needles" mobskill name adjustment.

UmeboshiXI:
**_I affirm:_**

Other references to "questionmark_needles" needles were renamed to "random_needles" but this one got over looked at some point causing mobskills/amigo sabotender's "??? Needles" to not activate.

"!exec target:useMobAbility(3867)" on a cactus or "!additem sun_water" to summon Amigo Sabotender. See that the skill activates.

Note: This skill does not seem a have a valid animation attached currently. Animation ID for 1000 Needles is 66 and Animation ID for 10000 needles is 69. Will need a capture to verify proper ID but one of those two would be visually similar.


TeoTwawki (approved):
for a sec there I thought "oh no not again" when I saw the notification for this. You can't imagine my relief that this is correct now.

PR #5315, opened 2024-03-28
[core] [lua] Fix mobskill TP return issue for AoE skills

TracentEden2:
**_I affirm:_**

This PR fixes an issue with mobskills whereby a mob is gaining TP for each player hit by an AoE mobskill. The fix leverages a similar pattern to the Spell object by adding PrimaryTargetID functions to the MobSkill object. Also adds the skill object as a parameter to `xi.mobskills.mobBreathMove` (similar to `mobPhysicalMove` and `mobMagicMove` which already passes the skill), so that `mobBreathMove` can also leverage the `getPrimaryTargetID` function.

Fight different mobs with AoE mobskills of different types (magic, physical, and breath) and check mob TP before and after the skill to see that mob should gain small amount of TP (like 80 TP) if the skill hit the primary target and should not add more tp for non-primary targets hit.

### DIFF
--- a/scripts/actions/mobskills/acid_breath.lua
+++ b/scripts/actions/mobskills/acid_breath.lua
@@ -19,7 +19,7 @@ mobskillObject.onMobWeaponSkill = function(target, mob, skill)
 
     xi.mobskills.mobStatusEffectMove(mob, target, xi.effect.STR_DOWN, power, tick, duration)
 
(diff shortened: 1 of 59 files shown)


WinterSolstice8 (approved):
Good catch, thanks.

Xaver-DaRed (approved):

PR #5316, opened 2024-03-28
[LUA] WS crit cratio max should be 1 higher than normal cratio max

MowFord:
**_I affirm:_**

The issue fixed a couple weeks ago related to crit ws always critting made me wonder why we didn't notice sooner, and i found that cratio is being heavily limited for crit attacks

The cratio clamp is not properly extending the max, artificially limiting the dmg output of weaponskill hits that crit. We need to adjust two things:
- clamp the cratio at 1 more than the max for crits
- Adjust the `getMeleePDifRange` function to not hard cap at 3, now that we have a pdif weapon table, we can copy the logic from the `xi.combat.physical.calculateMeleePDIF` function

Edit: i've added a `printDebug` local function [similar to here](https://github.com/LandSandBoat/server/blob/2436717ba375711d9c494f510854c4507293e3cd/scripts/globals/pets/avatar.lua#L17). Set the localvar `debug` equal to 1 on a player and they get pdif prints for each ws hit

image

print of the cratio values inside of the ws function affected (`xi.weaponskills.cMeleeRatio`)

print(fmt("{}: cratio-{} min-{} max-{} critmin-{} critmax-{}", attacker:getName(), cratio, pdif[1], pdif[2], pdifcrit[1], pdifcrit[2]))

Then use ws for various weapon types after setting the target's def high/low to see the ranges scaling across pdif max and attack vs def of attacker/defender
- `!setmod def -500`
- `!setmod def 1000`

example of club vs h2h, then 3rd is h2h with defender having very high def:
image

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -1137,34 +1137,35 @@ xi.weaponskills.calculatedIgnoredDef = function(tp, def, ignoredDefenseTable)
     return 0
 end
 
-local function getMeleePDifRange(correctedRatio)
-    -- pDifMax
-    local pDifMax = 3
-    if correctedRatio < 0.5 then
-        pDifMax = correctedRatio + 0.5
-    elseif correctedRatio < 0.7 then
-        pDifMax = 1
-    elseif correctedRatio < 1.2 then
-        pDifMax = correctedRatio + 0.3
-    elseif correctedRatio < 1.5 then
-        pDifMax = correctedRatio * 0.25 + correctedRatio
-    elseif correctedRatio < 2.625 then
-        pDifMax = correctedRatio + 0.375
+local function getMeleePDifRange(wRatio, pDifFinalCap)
+    local pDifUpperCap       = 0
+    local pDifLowerCap       = 0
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
Is there anything (other than your sanity) from stopping weaponskill.lua using the `xi.combat.physical.calculateMeleePDIF` function instead?

If it's just your sanity, understandable. we all hate that file.

WinterSolstice8 (dismissed):
stamped anyway, an improvement is an improvement.

MowFord:
> Is there anything (other than your sanity) from stopping weaponskill.lua using the `xi.combat.physical.calculateMeleePDIF` function instead?
> 
> If it's just your sanity, understandable. we all hate that file.

Yea I looked at the flow and there's so much that would need to change to use the physical util

MowFord:
I've added some debug prints that have been helpful in troubleshooting all this. Edited the OP about it

Edit: slight update to the debug print since I assume this is waiting re-approval for `fmt` exclusion to be merged and this branch rebased

MowFord:
Rebased so we'd pass lua styling checks btw (from using `fmt`)

PR #5317, opened 2024-03-28
[LUA] building flourish is a percentage increase in attack

MowFord:
**_I affirm:_**

The new physical util used for cpp attack rounds has considerations for weaponskills, but isn't used for them yet. It was brought to my attention that building flourish isn't treated properly here, so might as well adjust before it goes into use

remove the `isWeaponskill` restriction on these two sections of code and print pdif numbers while autoattacking before and after having building flourish to see that the attack is indeed being increased by 25%

```
    pDif = math.random(pDifLowerCap * 1000, pDifUpperCap * 1000) / 1000
print(fmt("{}: {}", actor:getName(), pDifLowerCap))
```

image

before and after using building flourish

### DIFF
--- a/scripts/globals/combat/physical_utilities.lua
+++ b/scripts/globals/combat/physical_utilities.lua
@@ -342,7 +342,7 @@ xi.combat.physical.calculateMeleePDIF = function(actor, target, weaponType, wsAt
             local flourishEffect = actor:getStatusEffect(xi.effect.BUILDING_FLOURISH)
 
             if flourishEffect:getPower() >= 2 then -- 2 or more Finishing Moves used.
-                actorAttack = actorAttack + 25 + flourishEffect:getSubPower()
+                actorAttack = math.floor(actorAttack * (125 + flourishEffect:getSubPower()) / 100)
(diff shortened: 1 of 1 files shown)


scripts/globals/combat/physical_utilities.lua
```diff
@@ -342,7 +342,7 @@ xi.combat.physical.calculateMeleePDIF = function(actor, target, weaponType, wsAt
             local flourishEffect = actor:getStatusEffect(xi.effect.BUILDING_FLOURISH)
 
             if flourishEffect:getPower() >= 2 then -- 2 or more Finishing Moves used.
-                actorAttack = actorAttack + 25 + flourishEffect:getSubPower()
+                actorAttack = actorAttack * (100 + 25 + flourishEffect:getSubPower()) / 100
```
Xaver-DaRed: `actorAttack = math.floor(actorAttack * (125 + flourishEffect:getSubPower()) / 100)`

Same below

Xaver-DaRed (approved):

PR #5322, opened 2024-03-29
North and South Gustaberg sheep have same drop table

MowFord:
**_I affirm:_**

https://www.bg-wiki.com/ffxi/Ornery_Sheep

image

🏀 
🧺


WinterSolstice8 (approved):

TeoTwawki (approved):

TeoTwawki:
same thing was a problem in the past for..virtually every other mob in north gusta. surprised the sheep got missed. I wonder if maybe some of the corrections wound up rolled back around the time someone foolishly tried to re-sequence some of the drop ids (which wound up breaking a ton of drop tables)

PR #5323, opened 2024-03-29
Sic recast time should be 1:30

MowFord:
**_I affirm:_**

https://www.bg-wiki.com/ffxi/Sic

Ready charges were correctly updated to 3x 30s per charge, sic was never updated


MowFord:
Also sidenote, i'm not sure sic should be attached to recastid 102. It doesn't have charges and messes with the reporting of time left, but I don't know enough to say for sure

WinterSolstice8 (approved):

TeoTwawki (approved):

PR #5326, opened 2024-03-30
[core] Escape more user input

WinterSolstice8:
**_I affirm:_**

More user inputs are escaped

Use commands that have strings in them, such as /tell, /pcmd kick, /pcmd leader, create a linkshell, blacklist a player, use alliance commands, kick a player from party/linkshell/alliance, and see them unchanged.

### DIFF
--- a/src/common/sql.cpp
+++ b/src/common/sql.cpp
@@ -285,10 +285,15 @@ size_t SqlConnection::EscapeString(char* out_to, const char* from)
 std::string SqlConnection::EscapeString(std::string const& input)
 {
     TracyZoneScoped;
-    std::string escaped_full_string;
-    escaped_full_string.reserve(input.size() * 2 + 1);
-    EscapeString(escaped_full_string.data(), input.data());
-    return escaped_full_string;
+    const char*  inputCString = input.c_str();
+    const size_t len          = strlen(inputCString);
+    char*        buffer       = new char[len * 2 + 1];
+    EscapeStringLen(buffer, inputCString, len);
+
+    std::string escapedString = buffer;
+    destroy(buffer);
+
+    return escapedString;
 }
 
 int32 SqlConnection::QueryStr(const char* query)
--- a/src/common/sql.cpp
+++ b/src/common/sql.cpp
@@ -308,6 +308,8 @@ int32 SqlConnection::QueryStr(const char* query)
         ShowError(query);
     }
 
+    // False positive, it is detecting ";" as a macro
+    // cppcheck-suppress unknownMacro
     DebugSQL(query);
 
     if (self == nullptr)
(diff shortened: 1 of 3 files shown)


Xaver-DaRed (approved):

src/common/sql.cpp
```diff
@@ -303,6 +308,8 @@ int32 SqlConnection::QueryStr(const char* query)
         ShowError(query);
     }
 
+    // False positive, it is detecting ";" as a macro
+    // cppcheck-suppress unknownMacro
```
zach2good, line 312: Gross, can you make an issue  for this and assign it to me and I'll fix this

src/common/sql.cpp
```diff
@@ -285,10 +285,15 @@ size_t SqlConnection::EscapeString(char* out_to, const char* from)
 std::string SqlConnection::EscapeString(std::string const& input)
 {
     TracyZoneScoped;
-    std::string escaped_full_string;
-    escaped_full_string.reserve(input.size() * 2 + 1);
-    EscapeString(escaped_full_string.data(), input.data());
-    return escaped_full_string;
+    const char*  inputCString = input.c_str();
+    const size_t len          = strlen(inputCString);
+    char*        buffer       = new char[len * 2 + 1];
```
zach2good, line 290: Death to raw new! Can you make this a unique_ptr[] please?

src/map/packet_system.cpp
```diff
@@ -2554,7 +2554,14 @@ void SmallPacket0x04D(map_session_data_t* const PSession, CCharEntity* const PCh
 
             if (PChar->UContainer->IsSlotEmpty(slotID))
             {
-                int32 ret = sql->Query("SELECT charid, accid FROM chars WHERE charname = '%s' LIMIT 1;", str(data[0x10]));
+                std::string charName = str(data[0x10]);
+                if (!std::regex_match(charName, std::regex("^([A-Z]{1}[a-z]{2,14})$")))
```
zach2good, line 2558: Remind me, this isn't one of the places where M.H.Whatever.Whatever can send you things?

TeoTwawki: Delivery box/sendbox AH records can possibly have non alphanumeric characters in the name (in the AH's case, because everyone tends to name their AH bot something invalid intentional so that players can't take its name)

TeoTwawki: retail has mailed things from `M.H.M.U.` (mog house management union) on occasion

WinterSolstice8: This isn't relevant here, this is packets recieved from the client about where to send it to, they can only send to real players, aka `^([A-Z]{1}[a-z]{2,14})$`

TeoTwawki: cools then

PR #5327, opened 2024-03-30
[core] Fix offhand damage ratio and attack reporting

TracentEden2:
**_I affirm:_**

This PR fixes three related issues:
1. Currently on LSB there is no differentiation between mainhand and offhand attack when calculating the damage ratio (for PDIF), both mainhand and offhand use the mainhand attack value. This causes incorrect damage ratios for offhand hits.
2. The above issue also causes incorrect offhand attack to be reported with /checkparam (as it gives mainhand attack for both)
3. Mobs that dual wield (currently only ark angel HM) do very weak damage with their offhand because SUB slot weapon damage is never set

This PR fixes these issues by adding an optional parameter to `getStat` to specify the slot (main or sub), and by adding a slot param to the functions involved in the calculation of the damage ratio. Also passing in the correct slot to getStat fixes the /checkparam issue. Finally, for mobs with dual wield the offhand weapon damage is set to the same as the mainhand weapon damage (as videos of ark angel HM show attacks from both hands to do roughly the same damage). 

Fight as a ninja with mainhand as fully skilled katana and offhand as low skill sword and notice the damage difference due to PDIF as compared to without the PR, also use /checkparams and see correct attack for main and offhand reported. Also Fight ark Angel HM and see that both mainhand and offhand attack do significant damage, can also check the damage without this PR.

### DIFF
--- a/scripts/globals/combat/physical_utilities.lua
+++ b/scripts/globals/combat/physical_utilities.lua
@@ -325,14 +325,14 @@ end
 
 -- WARNING: This function is used in src/utils/battleutils.cpp "GetDamageRatio" function.
 -- If you update this parameters, update them there aswell.
-xi.combat.physical.calculateMeleePDIF = function(actor, target, weaponType, wsAttackMod, isCritical, applyLevelCorrection, tpIgnoresDefense, tpFactor, isWeaponskill)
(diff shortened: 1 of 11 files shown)


WinterSolstice8 (approved):

WinterSolstice8:
Small conflict from another PR that fixed an attack bonus issue

TeoTwawki (approved):

PR #5328, opened 2024-03-31
[SMN] Correct Fenrir animations & Blood Pact abilities

N3ckB3ard:
**_I affirm:_**

Updates all of Fenrir's animations to reflect new pet_skills sql table
Compared all Blood pact msg's to retail captures
Added new msg enums to reflect retail captures
Removes Fenrir's animation shifting from core
Added SQL updates for future Fenrir Blood Pact abilities

Use Fenrir's BP Ward to see proper MP consumption and messages displayed
Use Fenrir's BP Rage to see proper MP consumption and messages displayed
Tested animations of future Blood Pact abilities


scripts/actions/abilities/pets/lunar_roar.lua
```diff
@@ -1,16 +1,24 @@
...
+abilityObject.onPetAbility = function(target, pet, petskill, summoner, action)
+    xi.job_utils.summoner.onUseBloodPact(target, petskill, summoner, action)
+
+    local effect = target:dispelStatusEffect()
+    if effect == xi.effect.NONE then
+        petskill:setMsg(xi.msg.basic.JA_NO_EFFECT_2)
+    else
+        target:dispelStatusEffect()
+        target:dispelStatusEffect()
+        petskill:setMsg(xi.msg.basic.NONE)
```
WinterSolstice8, line 18: Wouldn't this be dispelling 3 status effects (wiki says two effects) and then not reporting a dispel?

N3ckB3ard: Retail doesn't report when a effect is dispelled based on captures but I see what you mean. I can't remember if I intended to remove the additional dispel or left it from testing. I will have to report back on this

N3ckB3ard: I corrected the issue of the effects. This is a reference to retail with no display message of removal 
image

When nothing can be removed, the client is sent "no effect" message
image

WinterSolstice8: Interesting. The only difference is the "reaction" other than the message but I don't know if that actually produces a different message to the client. Thanks for the caps.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5330, opened 2024-04-02
[LUA] DNC step accuracy (of behavior and acc)

MowFord:
**_I affirm:_**

Har har, get it... accuracy

Steps should have an innate +10 accuracy.

Also, sub dancer can only apply up to level 5 steps, the steps have no effect if a more powerful effect is active. If the max step level is greater or equal to the existing power, the JA refreshes the duration 

Some screenshots to show testing of the second change:

image

Not shown in this second screenshot is starting as only sub dnc, re-applying a level 5 daze

image


TeoTwawki:
> the JA refreshes the duration

more specifically, duration stacks with remaining duration until cap. that's already happening I'm just clarifying for skim readers.

claywar (approved):

PR #5332, opened 2024-04-02
[LUA]Usable Items now display count and item on use.

hooksta4:
**_I affirm:_**

After speaking with Zach and Teo, it seems that addItem no longer has the message "Obtained [Quantity] Item) "baked" in. npcUtil however does. 
Modified all usable items with target:addItem to npcUtil.giveItem(target, {  { Item, [Quantity]  }  } )

Now will display the useable items Obtained: [Quantity] Item.  

locate usable item in scripts/items/USABLE ITEM (ie acid_bolt_quiver, woodsy_top_+1)
!additem 
equip item 
use item.

image

### DIFF
--- a/scripts/items/sheep_cap_+1.lua
+++ b/scripts/items/sheep_cap_+1.lua
@@ -12,9 +12,22 @@ itemObject.onItemCheck = function(target)
 end
 
 itemObject.onItemUse = function(target)
-    local giftList = { 4363, 4505, 1845, 4366, 919, 2213, 4571, 4504, 8929, 4367, 833 }
+    local giftList =
+    {
+        xi.item.FAERIE_APPLE,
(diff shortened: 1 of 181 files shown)


scripts/items/sheep_cap_+1.lua
```diff
@@ -12,9 +12,22 @@ itemObject.onItemCheck = function(target)
 end
 
 itemObject.onItemUse = function(target)
-    local giftList = { 4363, 4505, 1845, 4366, 919, 2213, 4571, 4504, 8929, 4367, 833 }
+    local giftList =
+    {
```
TeoTwawki, line 16: 👍

TeoTwawki (approved):

PR #5335, opened 2024-04-02
[core] [lua] Fix player BST charm calculations and move to lua space

TracentEden2:
**_I affirm:_**

This PR does several things:
1. This PR adjusts the calculations of players charming (chance and duration) mobs with the beastmaster charm ability to match retail testing. The charm duration calculations are from this [bg wiki](https://www.bg-wiki.com/ffxi/Charm_Duration) which includes a link to an alla page with testing data. While the calculations for charm chance are based on retail testing by Monti, Carnap, and Jimmayus (see [page](https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit#gid=151930160) of Jimmy sheet) of around 10000 charms. The previous charm calculations were based on an old ffxicyclopedia page with simple logic and unclear data origins. 
2. The PR also moves this charm logic into Lua space via a new xi.job_utils.beastmaster file, while removing several unneeded c++ functions and adding a duration param to the luabaseentity charm function.
3. The PR also fixes some small issues like light staff double counting for charm chance and some messaging for the charm ability.
5. Adjusts Maiden's Virelai (when cast by players) to have a base duration of 30 seconds with potential resistance to the song (similar to the current calculation when mobs cast it). The dependence of Maiden's Virelai on any bst charm-ability logic or calculations is removed as there is no evidence to support such dependence as far as I could find.

Note that this does not impact how mobs charm players as that is done via a different mechanism based on status resistance of the charm status effect.

This is a partly an ASB PR coming upstream.

Charm various mobs with varying level diff and CHR, results should match retail charm rates.


src/map/lua/lua_baseentity.cpp
```diff
@@ -13189,11 +13189,15 @@ bool CLuaBaseEntity::addBardSong(CLuaBaseEntity* PEntity, uint16 effectID, uint1
  *  Function: charm()
  *  Purpose : Charms an entity target
  *  Example : mob:charm(target)
+ *  Notes: optional parameter specifying the duration in seconds
+ *         (default duration of 0 implies duration is controlled
+ *         by the status effect charm, which is the case for charmed players)
  ************************************************************************/
 
-void CLuaBaseEntity::charm(CLuaBaseEntity const* target)
+void CLuaBaseEntity::charm(CLuaBaseEntity const* target, sol::variadic_args va)
```
zach2good: If you are only going to have (always target, optional duration) you don't need to use variadic_args here, you can use a sol object I think? or a sol optional, or something like that. It makes the function signature say what the intended use is, rather than a catch-all

scripts/globals/job_utils/beastmaster.lua
```diff
@@ -0,0 +1,132 @@
...
+
+    -- Calculate base duration and dLvl
+    local baseCharmDuration = math.floor(1.25 * charmer:getStat(xi.mod.CHR) + 150)
+    local dLvl = charmer:getMainLvl() - target:getMainLvl()
+
+    -- Default multiplier for dLvl -6 or lower
+    local dLvlCharmMult = 1 / 24
+
+    if dLvl >= -6 and dLvl < 9 then
+        -- Quintic fitting of duration multiplier as function of dLvl (r^2 > 0.999)
```
zach2good: Very fancy, where did you get the information for this? Might be worth linking to a site that shows a visual graphing of this so it can be easily compared to what's in the wikis etc.

scripts/globals/job_utils/beastmaster.lua
```diff
@@ -0,0 +1,132 @@
+-----------------------------------
+-- beastmaster Job Utilities
+-----------------------------------
+
+-----------------------------------
+xi = xi or {}
+xi.job_utils = xi.job_utils or {}
+xi.job_utils.beastmaster = xi.job_utils.beastmaster or {}
```
zach2good: No need for the extra gaps in here, unless you have some requires:

```lua
-- Warrior Job Utilities
xi = xi or {}
xi.job_utils = xi.job_utils or {}
xi.job_utils.warrior = xi.job_utils.warrior or {}
```

```lua
-- Thief Job Utilities
require('scripts/globals/quests')
require('scripts/globals/utils')
xi = xi or {}
xi.job_utils = xi.job_utils or {}
xi.job_utils.thief = xi.job_utils.thief or {}

```

zach2good (commented):
I don't know anything about BST, but this seems to be a sane general pattern of moving things up into Lua.

src/map/lua/lua_baseentity.cpp
zach2good: Yeah, we use ` sol::object const& p0` for things that might be nil, but the function signature is mainly the same:
```
uint32 param0 = (p0 != sol::lua_nil) ? p0.as<uint32>() : 0;
```

scripts/globals/job_utils/beastmaster.lua
```diff
@@ -0,0 +1,132 @@
...
+
+    return math.floor(charmDuration)
+end
+
+xi.job_utils.beastmaster.getCharmChance = function(charmer, target, includeMods)
+    if
+        not charmer or                                -- Invalid charmer
+        not target or                                 -- Invalid target
+        not charmer:isPC() or                         -- Charmer not a player
+        not target:isMob() or                         -- Target not a mob
```
Xaver-DaRed, line 562: Cant charm be used in PvP? Or does it never charm players?

If so, please comment it so no1 else asks it.

scripts/globals/job_utils/beastmaster.lua
```diff
@@ -0,0 +1,132 @@
...
+        target:getMobMod(xi.mobMod.CHARMABLE) == 0 or -- Not charmable
+        (target:getMaster() ~= nil)                   -- Someone else's pet
```
Xaver-DaRed: unneeded parenthesis

scripts/globals/job_utils/beastmaster.lua
```diff
@@ -0,0 +1,132 @@
...
+        charmChance = charmChance * 1.5
+    elseif rank <= -2 then
+        charmChance = charmChance * 1.4
+    elseif rank <= -1 then
+        charmChance = charmChance * 1.2
+    elseif rank <= 0 then
+        charmChance = charmChance
+    else
+        charmChance = charmChance / 2
+    end
```
Xaver-DaRed, line 599: This seems weird to me.
If resistance rank is...
- 10, shouldnt MHR be floored to 5%?
- 11, shouldnt it auto-fail?

All of this looks like it COULD use the same functions used in magic enfeebles, resistance wise.

scripts/globals/job_utils/beastmaster.lua
```diff
@@ -0,0 +1,132 @@
...
+        not target or         -- Invalid target
+        not charmer:isPC() or -- Charmer not a player
+        not target:isMob()    -- Target not a mob
+    then
+        return xi.msg.basic.JA_MISS
+    elseif
+        -- Not charmable so apply bind
+        target:getMobMod(xi.mobMod.CHARMABLE) == 0
+    then
+        local resist = applyResistanceAddEffect(charmer, target, xi.element.ICE, 0)
```
Xaver-DaRed, line 628: Styling
```lua
    if
        not charmer or        -- Invalid charmer
        not target or         -- Invalid target
        not charmer:isPC() or -- Charmer not a player
        not target:isMob()    -- Target not a mob
    then
        return xi.msg.basic.JA_MISS

    -- Not charmable so apply bind
    elseif target:getMobMod(xi.mobMod.CHARMABLE) == 0 then
        local resist = applyResistanceAddEffect(charmer, target, xi.element.ICE, 0)
```

scripts/globals/job_utils/beastmaster.lua
```diff
@@ -0,0 +1,132 @@
...
+        if not target:hasStatusEffect(xi.effect.BIND) and resist >= 0.5 then
+            target:addStatusEffect(xi.effect.BIND, 1, 0, math.random(1, 5))
+            return 127
```
Xaver-DaRed: 127? We should add this msg entry to msg enum

Xaver-DaRed (commented):
I like what I'm seing in general. Please do take a look at magic-hit-rate calculations in case that logic can be used.

scripts/globals/job_utils/beastmaster.lua
TracentEden2: dlvl_to_charm_duration_func_fitting

Yeah, this is the fitting of the function to the data points from the testing, including fitting metrics. Let me know if you anything else, thanks!

scripts/globals/job_utils/beastmaster.lua
TeoTwawki: > Cant charm be used in PvP? Or does it never charm players?
> 
> If so, please comment it so no1 else asks it.

in pvp it binds instead

scripts/globals/job_utils/beastmaster.lua
TracentEden2: > This seems weird to me. If resistance rank is...
> 
> * 10, shouldnt MHR be floored to 5%?
> * 11, shouldnt it auto-fail?
> 
> All of this looks like it COULD use the same functions used in magic enfeebles, resistance wise.

I think the highest resist rank validated on retail was a mob with rank of 2 (this [mob](https://www.bg-wiki.com/ffxi/Category:Twitherym) via [this post](http://blog.livedoor.jp/yotsuba_yudaya/archives/2007204.html#more)), it might be possible to check a mob with rank of 4 (Sabotender) but I think most mobs with higher ranks than that are not susceptible to charm for testing. Maybe there is another way of testing or creating high rank mobs?

src/map/lua/lua_baseentity.cpp
TracentEden2: > Yeah, we use ` sol::object const& p0` for things that might be nil, but the function signature is mainly the same:
> 
> ```
> uint32 param0 = (p0 != sol::lua_nil) ? p0.as<uint32>() : 0;
> ```

Thanks for the feedback, I have fixed the PR.

scripts/globals/job_utils/beastmaster.lua
TracentEden2: Thanks for the feedback, I have fixed the PR.

scripts/globals/job_utils/beastmaster.lua
TracentEden2: Yep, I have fixed the logic so players should now be bound similar to non-charmable mobs

scripts/globals/job_utils/beastmaster.lua
TracentEden2: Thanks for the feedback, I have fixed the PR.

scripts/globals/job_utils/beastmaster.lua
TracentEden2: Thanks, after further checking I found there already is a 127 msg enum so I fixed the PR to use that now.

scripts/globals/job_utils/beastmaster.lua
TracentEden2: Thanks for the feedback, I have fixed the PR.

TracentEden2:
I am also awaiting some small retail checks from Jimmayus before marking the PR as ready to go (non-draft status).

scripts/globals/job_utils/beastmaster.lua
TracentEden: > > This seems weird to me. If resistance rank is...
> > 
> > * 10, shouldnt MHR be floored to 5%?
> > * 11, shouldnt it auto-fail?
> > 
> > All of this looks like it COULD use the same functions used in magic enfeebles, resistance wise.
> 
> I think the highest resist rank validated on retail was a mob with rank of 2 (this [mob](https://www.bg-wiki.com/ffxi/Category:Twitherym) via [this post](http://blog.livedoor.jp/yotsuba_yudaya/archives/2007204.html#more)), it might be possible to check a mob with rank of 4 (Sabotender) but I think most mobs with higher ranks than that are not susceptible to charm for testing. Maybe there is another way of testing or creating high rank mobs?

After talking with Jimmayus, if there is some kind of flooring for charm it is not possible to determine (on retail) because from our understanding there are not charmable mobs with light EEM lower than 50%. So makes sense to stick with the PR values as those have been validated in captures.

TracentEden2: > > This seems weird to me. If resistance rank is...
> > 
> > * 10, shouldnt MHR be floored to 5%?
> > * 11, shouldnt it auto-fail?
> > 
> > All of this looks like it COULD use the same functions used in magic enfeebles, resistance wise.
> 
> I think the highest resist rank validated on retail was a mob with rank of 2 (this [mob](https://www.bg-wiki.com/ffxi/Category:Twitherym) via [this post](http://blog.livedoor.jp/yotsuba_yudaya/archives/2007204.html#more)), it might be possible to check a mob with rank of 4 (Sabotender) but I think most mobs with higher ranks than that are not susceptible to charm for testing. Maybe there is another way of testing or creating high rank mobs?

After talking with Jimmayus, if there is some kind of flooring for charm it is not possible to determine (on retail) because from our understanding there are not charmable mobs with light EEM lower than 50%. So makes sense to stick with the PR values as those have been validated in captures. Let me know if anything needs more clarification on this point thanks!

TracentEden2:
> I am also awaiting some small retail checks from Jimmayus before marking the PR as ready to go (non-draft status).

We got the captures from several more mobs and validated the formula so this PR should be ready to go. Let me know if anything else that would be needed, thanks!

TracentEden2:
> I like what I'm seing in general. Please do take a look at magic-hit-rate calculations in case that logic can be used.

Yeah from my checking the magic hit rate cannot be used because for example the multipliers and calculations are different.

Xaver-DaRed (approved):

scripts/globals/job_utils/beastmaster.lua
zach2good: You've generated this, how about a link to the site where you did this so people can look it up themselves? Once the context of this PR is gone these are just numberwang

zach2good (commented):
Seems fine to me, I just want that link so people can see how the maths was mathed

scripts/globals/job_utils/beastmaster.lua
TracentEden: > You've generated this, how about a link to the site where you did this so people can look it up themselves? Once the context of this PR is gone these are just numberwang

Yeah the data points are from the table [here](https://www.bg-wiki.com/ffxi/Charm_Duration) which was validated by retail testing by Monti on 11/2022. The fitting was done by quintic least squares fitting and can be seen [here](https://mycurvefit.com/index.html?action=openshare&id=358a5d99-4499-4a6a-bbfe-0a667739335c) via mycurvefit. Do you want me to put the links in the code itself? Thanks for the comments so far!

zach2good: Yeah, a comment in the code please. Like I say, once the context of the PR is gone, there's nothing else to go on

TracentEden: > Yeah, a comment in the code please. Like I say, once the context of the PR is gone, there's nothing else to go on



> Yeah, a comment in the code please. Like I say, once the context of the PR is gone, there's nothing else to go on

Yep, ok, I added comments.

zach2good (approved):

PR #5336, opened 2024-04-03
[lua] Aura steal should be possible even on mob with steal items

MowFord:
**_I affirm:_**

Aura steal was coded to only try if `mob:getStealItem()` returned zero, but it should also attempt to steal a buff also if a mob has stealable items still but the steal failed.

no aura steal merits, attempt steal on dynamis mob with a buff (whm or blm) repeatedly (mob at full hp should be very low chance to actually steal an item). See that you still get regular "steal fails" messages
merit aura steal to max to get capped 95% chance to steal a buff when steal fails and basic resist check passes.

### DIFF
--- a/scripts/globals/job_utils/thief.lua
+++ b/scripts/globals/job_utils/thief.lua
@@ -411,7 +411,7 @@ xi.job_utils.thief.useSteal = function(player, target, ability, action)
 
     -- Attempt Aura steal
     -- local effect = xi.effect.NONE
-    if stolen == 0 and player:hasTrait(75) then
+    if player:hasTrait(75) then
         local resist = applyResistanceAbility(player, target, xi.element.NONE, 0, 0)
         -- local effectStealSuccess = false
         if resist > 0.0625 then


scripts/globals/job_utils/thief.lua
```diff
@@ -411,7 +411,7 @@ xi.job_utils.thief.useSteal = function(player, target, ability, action)
 
     -- Attempt Aura steal
     -- local effect = xi.effect.NONE
-    if stolen == 0 and player:hasTrait(75) then
```
TeoTwawki, line 414: you need to make sure you don't get an item -and- steal a buff via aura steal in the same attempt though. flow should be : try to get an item, get or fail, if fail then try aura steal, if STILL fail then do the fail to steal thing.

From the current code flow it doesn't look like that happens looking at lines [400 to 414](https://github.com/LandSandBoat/server/pull/5336/files#diff-5b0ce0ef3da04a17a4c39488728e3ddb14d1154486f68991269cf268cfe88852R400-R414), even after your change removing that check for a variable that likely mislead prior authors. 

Likely past authors saw it and thought the getStealItem returned zero for fails rather than returning what steal items exist

MowFord: Yea a more ideal flow would probably be to not return early when you steal and reset the steal variable to zero on failed steal. But this code does work properly and I figured the comment near the early return was sufficient

TeoTwawki: if I've missed something and this actually functions retail accurately with only removing that variable in the condition say so (I ain't perfect) and making the flow make more sense is a future me problem.

But if steal can both proc and still aura steal in the same use (non retail behavior that) that's a nope and should be fixed rather than just breaking this a different way than it was before.

MowFord: If this code steals an item it returns early. If it fails streal then it sets the message and animation for a failed steal, then it checks if it should steal a buff, and if that is successful it updates the message etc. Afaik it functions perfectly

Xaver-DaRed (approved):

scripts/globals/job_utils/thief.lua
Xaver-DaRed: This change works properly. The cleanup for the skill can come after both this PR and the enum PR #5338 go in.

TeoTwawki (approved):

PR #5338, opened 2024-04-04
[Chore] Create job trait enum and use when appropiate

Xaver-DaRed:
**_I affirm:_**

As title sais.

Play regularly. Use a high lvl THF with ambush, assassin, etc and WS. Or use rampart with Iron will trait.


scripts/enum/trait.lua
```diff
@@ -0,0 +1,136 @@
+xi = xi or {}
+
+xi.trait =
+{
+    -- IF YOU ADD ANY NEW ENTRY HERE, ADD IT IN src/map/trait.h ASWELL!
+
```
TeoTwawki, line 11: request: add comment from the [doc](https://github.com/LandSandBoat/server/blob/base/documentation/player_traits.txt)
> Offset by 1536 from ROM\181\72.dat

in both lua and the .h file

and then nuke doc coz it'll be redundant

src/map/trait.h
```diff
@@ -29,6 +29,7 @@
 
 enum TRAITTYPE
 {
+    // IF YOU ADD ANY NEW ENTRY HERE, ADD IT IN scripts/enum/trait.lua ASWELL!
```
TeoTwawki: `as well` is two words. suggest `ALSO`

scripts/enum/trait.lua
Xaver-DaRed: Rebased, all comments adressed

WinterSolstice8 (approved):

TeoTwawki (approved):

PR #5343, opened 2024-04-05
[core] Don't update destination in blob extraction if blob is null

### LINKED ISSUE #5342: 🐛 Server Crash upon New Char Create
**_I affirm:_**

Windows 11

## Branch affected by issue 

`base`

## Steps to reproduce

With the newest update from the game. When creating a new char and attempt log in after picking starting area, server will crash. Char does save in the Database, When attempt log in again after selecting that char server will crash again. Can log into the Char on Older update just fine and no issues. I have tested this on Two different Window PC's with both updated on latest. I can log in on my main Char on the newest update fine without a server crash, it's just for anyone that is new to the server.

## Expected behavior

Not have the server crash upon logging in a new created char.

WinterSolstice8:
**_I affirm:_**

Fixes #5342

Create a new character, login without crash

### DIFF
--- a/src/common/database.h
+++ b/src/common/database.h
@@ -270,10 +270,13 @@ namespace db
 
         TracyZoneScoped;
 
-        if (!rset->isNull(blobKey.c_str())) // This can do very bad things if the blob is NULL.
+        std::memset(&destination, 0x00, sizeof(T));
+
+        // If we use getString on a null blob we will get back garbage data.
+        // This will introduce difficult to track down crashes.
+        if (!rset->isNull(blobKey.c_str()))
         {
             auto blobStr = rset->getString(blobKey.c_str());
-            std::memset(&destination, 0x00, sizeof(T));
             std::memcpy(&destination, blobStr.c_str(), sizeof(T));
         }
     }
--- a/src/common/database.h
+++ b/src/common/database.h
@@ -270,9 +270,12 @@ namespace db
 
         TracyZoneScoped;
 
-        auto blobStr = rset->getString(blobKey.c_str());
-        std::memset(&destination, 0x00, sizeof(T));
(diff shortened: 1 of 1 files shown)


zach2good (approved):
:thinking: huh, but I literally tested new character creation and log in / log out / log in at least 4 times when I was testing alliance and cross process stuff

Even if we're not copying the contents of the null string (which I assume is the issue?), should we be memsetting the target struct, because we've asked to load the contents of the db into that struct anyway? To me this is never additive, it should always be destructive/replacement

zach2good:
To the effect of:

```cpp
        std::memset(&destination, 0x00, sizeof(T));
        if (!rset->isNull(blobKey.c_str())) // This can do very bad things if the blob is NULL.
        {
            auto blobStr = rset->getString(blobKey.c_str());
            std::memcpy(&destination, blobStr.c_str(), sizeof(T));
        }
```

I should imagine

zach2good:
And maybe reword the comment to be clearer about what will happen: something something reading in uninitialised memory which will populate `destination` with garbage, leading to difficult to diagnose crashes etc.

I like using the terminology "difficult to diagnose crashes" because it puts the fear of god into anyone who touches compilers.

zach2good:
Assuming it was before/after, and that the eminence blob was zero'd out before use and didn't start as uninit'd garbage

zach2good:
I think cpp changes from web mess up CI, can you fix this for me when you get the chance 🙏

WinterSolstice8:
> I think cpp changes from web mess up CI, can you fix this for me when you get the chance 🙏

Seems to be an extra newline, I'm out for a bit but will do

Xaver-DaRed (approved):

zach2good (approved):

PR #5355, opened 2024-04-06
[core] Remove UB with charming

WinterSolstice8:
**_I affirm:_**

Does what it says on the tin - removes a lot of undefined behavior with charm.
Previously, there were rare crashes where a CMobEntity was getting calls to CPetEntity and crashing.

Use pets of all kinds and see no changes

### DIFF
--- a/src/map/ai/controllers/pet_controller.cpp
+++ b/src/map/ai/controllers/pet_controller.cpp
@@ -27,7 +27,7 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "status_effect_container.h"
 #include "utils/petutils.h"
 
-CPetController::CPetController(CPetEntity* _PPet)
+CPetController::CPetController(CMobEntity* _PPet)
 : CMobController(_PPet)
 , PPet(_PPet)
 {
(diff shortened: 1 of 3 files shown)


src/map/utils/battleutils.cpp
```diff
@@ -4611,8 +4612,10 @@ namespace battleutils
 
         if (PVictim->objtype == TYPE_MOB)
```
zach2good, line 4613: For completion's sake, can you add a check here for `PCharmer type is CCharEntity`, since we're imminently going to cast it into one

src/map/ai/controllers/pet_controller.cpp
```diff
@@ -27,7 +27,7 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "status_effect_container.h"
 #include "utils/petutils.h"
 
-CPetController::CPetController(CPetEntity* _PPet)
+CPetController::CPetController(CMobEntity* _PPet)
```
zach2good, line 30: So, a mob's pet is still a mob entity, but is using a pet controller? But a player's pet is a pet entity that uses a pet controller?

zach2good (approved):

src/map/utils/battleutils.cpp
WinterSolstice8: There is a dynamic cast (and check at line 4637) if the charmer is a PChar, should not be a problem
All other functionality isn't needed on a non-pchar

src/map/ai/controllers/pet_controller.cpp
WinterSolstice8: And a players pet can also be a mob entity that uses a pet controller, yeah.

PR #5363, opened 2024-04-08
[Quest] Paradise, Salvation, and Maps

jamesbradleym:
**_I affirm:_**

Implements quest Paradise, Salvation, and Maps, adapted from [ASB Commit](https://github.com/hookstar/AirSkyBoat/commit/c1d8348202af3852cde7f9527ab6a1e9bdea4c3d)
Additional: Added new cycle dialog shown in capture post quest completion. Added additional check so player could not turn in a second floorplan prior to getting the results (We could take this one step further and not treasure reward player with floorplans if there is a current turn in for this quest)

[Capture](https://www.youtube.com/watch?v=psWsc_1o6RY)

- Speak with Nivorajean in Tavnazia with The Savage (Promathia 4-2) mission complete (!gotoid 16883781)
- Unlock Sacrarium Chest !gotoid 16892183 (IMPORTANT: You must get this KI from a chest as its index is part of the pop-quiz)
- Speak with Nivorajean with Ripped Floorplans Key Item from chest and select the correct chest location
- ??
- Profit

## Notes
The original ASB commit had code which would inform the player to contact a GM if their chest index was 0. This could seemingly only happen if the chest locations are not within 1 coordinate position of the chest spawn locations. If it seems like a good safety measure I can put this safeguard back in or we could wide up the coordinate threshold.


jamesbradleym:

PR #5364, opened 2024-04-08
[Quest] Fix Wild Card soft lock

ampitere:
**_I affirm:_**

Fixes an issue with the Wild Card quest that causes a soft lock making players unable to progress if they receive the cutscene with Apururu instead of Joker (if they are below rank 9). This would put their WildCard progress at 3, but the coffer only checks if their progress is at 2.

1. !completequest 2 42 name
2. !completequest 2 76 name
3. !addquest 2 77 name
4. !setplayervar name [2][77]Prog 3 (should work with 2 or 3)
5. !additem 1057
6. !pos 219 16 -49 169 name Coffer should give you Joker Card KI
7. !pos -26 -13 259 239 name to receive the cutscene
8. !pos -194 -10 -119 238 name to complete the quest

### DIFF
--- a/scripts/globals/treasure.lua
+++ b/scripts/globals/treasure.lua
@@ -1107,12 +1107,12 @@ xi.treasure.treasureInfo =
                 {
                     {
                         test = function(player)
-                            return player:getCharVar('WildCard') == 2
+                            return not player:hasKeyItem(xi.ki.JOKER_CARD) and
+                                (player:getCharVar('[2][77]Prog') == 2 or player:getCharVar('[2][77]Prog') == 3)
                         end,
 
                         code = function(player)
                             npcUtil.giveKeyItem(player, xi.ki.JOKER_CARD)
-                            player:setCharVar('WildCard', 3)
                         end,
                     },
                 },


zach2good (dismissed):

scripts/globals/treasure.lua
```diff
@@ -1107,7 +1107,7 @@ xi.treasure.treasureInfo =
                 {
                     {
                         test = function(player)
-                            return player:getCharVar('WildCard') == 2
+                            return player:getCharVar('WildCard') == 2 or player:getCharVar('WildCard') == 3
```
Xaver-DaRed: This variable is never set to 2 in the first place. Its just used in this file.

Variable being checked should be `'[2][77]Prog'`

So the fix would be in the end...
```lua
    misc =
    {
        {
            -- Players is on "Wild Card" quest (after meeting with Joker or Apururu) and doesn't have KI already.
            test = function(player)
                return not player:hasKeyItem(xi.ki.JOKER_CARD) and
                    (player:getCharVar('[2][77]Prog') == 2 or player:getCharVar('[2][77]Prog') == 3)
            end,

            code = function(player)
                npcUtil.giveKeyItem(player, xi.ki.JOKER_CARD)
            end,
        },
    },
```

Note its similar to your fix, except we change the proper variable and we remove setting it when getting the KI, because its unneeded

Xaver-DaRed (changes requested):
Please change the variable name, add the check for KI and remove the variable progression

scripts/globals/treasure.lua
ampitere: That makes a lot more sense (in regards to the char var). Appreciate the suggestion, refactors have been made and pushed.

ampitere:
> Please change the variable name, add the check for KI and remove the variable progression

Refactored per your suggestion.

Xaver-DaRed (approved):

scripts/globals/treasure.lua
Xaver-DaRed: For future cases, please dont resolve conversations yourself. Wait for the original reviewer to close it, sowe dont miss stuff on accident.

zach2good (approved):

PR #5369, opened 2024-04-08
[SQL] Fix invalid position entry for Jormungand

Xaver-DaRed:
**_I affirm:_**

Fixes incorrect sign in Y pos.

Prevents Jormungand from spawning here...
imagen

And instead makes it spawn in a valid position

None


almuth150:
Curious, wouldn't changing the `pos_y` of that entry to be negative yield a spawn point in the area it should be? It looks like it might be a typo that dropped the negative sign.

Xaver-DaRed:
You are right, changing the sign makes it a valid position

WinterSolstice8 (approved):

PR #5370, opened 2024-04-08
[cpp] Ensure proper mapping from db to player rank array

MowFord:
**_I affirm:_**

image

Recent changes to db handler had a little mistake in renaming indexes to column names. Result is a player allied with windurst shows their rank for bastok, and a player aligned with sandoria reports rank 0 (depending on nation rank points i assume)

Edit: worth mentioning that this was tested with multiple player nation allegiances and confirmed profile, search linkshell, search party, and generic search returns the proper nation

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -472,9 +472,9 @@ namespace charutils
         {
             PChar->profile.rankpoints = rset->getUInt("rank_points");
 
-            PChar->profile.rank[0] = (uint8)rset->getInt("rank_points");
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):

MowFord:
BTW i found the search server having a similar issue, I will update this PR with that fix as well

image

sandorians are nation 0, so this needs to be +1... but i'll just make the 3 references a switch statement

zach2good (approved):

WinterSolstice8 (approved):

PR #5383, opened 2024-04-10
[LUA] Erle NM

MowFord:
**_I affirm:_**

Includes both commits, but #5382 will absorb the first, sql, commit

[Erle](https://www.bg-wiki.com/ffxi/Erle) 
- Captures found that he could double attack about 10% of the time and triple attack about 35% of the time, which matches with him being a warrior (like other bees) with high chance to triple attack
image
image

Update `Death_Jacket.lua` with `xi.mob.phOnDespawn(mob, ID.mob.ERLE_PH, 100, 5400, { immediate = true }) -- 1.5 hour` to make 100% to spawn Erle immediately
Kill single PH: 17150075
See Erle link with and link other bees as well as use no mobskills and cast the relevant spells

### DIFF
--- a/scripts/zones/Rolanberry_Fields_[S]/IDs.lua
+++ b/scripts/zones/Rolanberry_Fields_[S]/IDs.lua
@@ -3,6 +3,8 @@
 -----------------------------------
 zones = zones or {}
 
+local erleID = GetFirstID('Erle')
+
 zones[xi.zone.ROLANBERRY_FIELDS_S] =
 {
     text =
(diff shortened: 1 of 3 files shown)


MowFord:
rebased after sql got merged. Strange github didn't detect that commit is already present in `base`

scripts/zones/Rolanberry_Fields_[S]/IDs.lua
```diff
@@ -37,6 +39,11 @@ zones[xi.zone.ROLANBERRY_FIELDS_S] =
             [17150279] = 17150280, -- -484.535 -23.756 -467.462
         },
 
+        ERLE_PH =
+        {
+            [erleID - 6] = erleID, -- -298.487 3.637 54.586
```
Comment (WinterSolstice8) [inline scripts/zones/Rolanberry_Fields_[S]/IDs.lua:44]: I assume you've tried, but there isn't better way to do this because of the order of operations when the table get created, yes?

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/IDs.lua:44]: Correct best I can tell there not a better way to do it that keeps the dynamic use of `GetFirstID`

WinterSolstice8 (approved):

PR #5391, opened 2024-04-11
[Quest]Sin Hunting RNG AF1 to IF

hooksta4:
**_I affirm:_**

Converts Sin Hunting quest to Interaction Framework.

- Implemented several DefaultActions that were missing for associated NPC text.
- Added message ID for accurate messages
- Removed NPC lua functions related to Sin Hunting
There is a dialog event 542 that will show up during dialog on complete due to it checking for var < unbridledPassionCS which is always true. This will resolve when that quest is converted. At this point it is more effort to fix that just to be converted shortly. 

Converts quest Sin Hunting. 
With RNG set as main job and over level 40, talk to Perih Vashai in Windurst Woods. Dialogs change for guards Muhk and Kapel. Their default messages also change after quest completion.


hooksta4:
@Xaver-DaRed what do you think about these changes?

scripts/quests/windurst/RNG_AF1_Sin_Hunting.lua
```diff
@@ -0,0 +1,179 @@
...
+            ['qm2'] =
+            {
+                onTrigger = function(player, npc)
+                    if
+                        quest:getVar(player, 'Prog') == 3 --and
+                        --IsMoonFull()--disable comment for era time
+                    then
+                        return quest:progressEvent(13, 0, xi.item.PINCH_OF_GLITTERSAND)
+                    end
+                end,
```
Xaver-DaRed, line 132:
```lua
                    -- Full moon requirement was removed.
                    if quest:getVar(player, 'Prog') == 3 then
                        return quest:progressEvent(13, 0, xi.item.PINCH_OF_GLITTERSAND)
                    end
                end,
```

scripts/zones/Ranguemont_Pass/npcs/Perchond.lua
```diff
@@ -3,45 +3,20 @@
 --  NPC: Perchond
 -- !pos -182.844 4 -164.948 166
 -----------------------------------
-local ID = zones[xi.zone.RANGUEMONT_PASS]
+--local ID = zones[xi.zone.RANGUEMONT_PASS]
```
Xaver-DaRed: Remove this line if it isn't needed anymore

scripts/zones/Windurst_Woods/npcs/Perih_Vashai.lua
```diff
@@ -4,7 +4,7 @@
 -- Starts and Finishes Quest: The Fanged One, From Saplings Grow
 -- !pos 117 -3 92 241
 -----------------------------------
-local ID = zones[xi.zone.WINDURST_WOODS]
+--local ID = zones[xi.zone.WINDURST_WOODS]
```
Xaver-DaRed: Same here

Xaver-DaRed (approved):

scripts/quests/windurst/RNG_AF1_Sin_Hunting.lua
```diff
@@ -0,0 +1,176 @@
...
+                    else
+                        return quest:progressEvent(527)
+                    end
+                end,
+            },
+
+            ['Kapeh_Myohrye'] =
+            {
+                onTrigger = function(player, npc)
+                    if quest:getVar(player, 'Prog') <= 4 then
```
claywar: Looks like 4 is the maximum value used in this script.  Since this will always be true, the condition is not necessary here or below with Muhk Johldy

hookstar: So you are saying remove the if and just put return event ()?

claywar: Since there is no other logic for these NPCs, in this example, 
```
            ['Kapeh_Myohrye'] =
            {
                    if quest:getVar(player, 'Prog') <= 4 then
                        return quest:event(526)
                    end
                end,
            },
```
can be simplified to:
`['Kapeh_Myohrye'] = quest:event(526),`

hookstar: Ok - thats what I was trying to say above. Got it - will fix shortly.

hooksta4: @claywar Thanks- complete.

hooksta4:
Just waiting for [The Fanged One](https://github.com/LandSandBoat/server/pull/5530) to be merged and then I can fix this one.

PR #5392, opened 2024-04-11
[SQL] Fix Sneaking Boots missing enhance Stealth mod

ampitere:
**_I affirm:_**

Sneaking Boots are missing an item mod that enhances the effect of Stealth (a NIN trait). A 75 NIN has a Stealth mod value of 3 so 1 seems to make sense here. No information on the wikis regarding the actual value so may require retail verification.

1. !additem 15698
2. Equip the boots
3. !getmod 358

### DIFF
--- a/sql/item_mods.sql
+++ b/sql/item_mods.sql
@@ -28459,7 +28459,8 @@ INSERT INTO `item_mods` VALUES (15697,20,6); -- WATER_RES: 6
 INSERT INTO `item_mods` VALUES (15697,26,2); -- RACC: 2
 
 -- Sneaking Boots
-INSERT INTO `item_mods` VALUES (15698,1,9); -- DEF: 9
+INSERT INTO `item_mods` VALUES (15698,1,9);   -- DEF: 9
+INSERT INTO `item_mods` VALUES (15698,358,1); -- STEALTH: 1 -- TODO: Stealth trait needs retail verification, real value unknown
 
 -- Templar Sabatons
 INSERT INTO `item_mods` VALUES (15699,1,11);  -- DEF: 11


sql/item_mods.sql
```diff
@@ -28459,7 +28459,8 @@ INSERT INTO `item_mods` VALUES (15697,20,6); -- WATER_RES: 6
 INSERT INTO `item_mods` VALUES (15697,26,2); -- RACC: 2
 
 -- Sneaking Boots
-INSERT INTO `item_mods` VALUES (15698,1,9); -- DEF: 9
+INSERT INTO `item_mods` VALUES (15698,1,9);   -- DEF: 9
+INSERT INTO `item_mods` VALUES (15698,358,1); -- STEALTH: 1
```
TeoTwawki, line 28463: well, I'm glad you went with the minimum value since I can't find any data about how much it enhances it by. Also this mod  works even if you don't have the trait already. I don't know if that's also how retail treats it. There is just no info about it. So you are right about some retail checking being needed.

ampitere: The only spot it seems to be used currently is inside of mob_controller.cpp on line 259 where it's calculating currentDistance and adding the mod value to it.

TeoTwawki: right. fun fact: you can add negative amounts to people to make them aggro from shorter distances 😈

WinterSolstice8 (approved):
I see no reason to not add this, but please add a TODO indicating we don't actually know the real value. At some point I will probably test Stealth to figure out how it actually works, a value of 1 is pretty conservative imo.

ampitere:
> I see no reason to not add this, but please add a TODO indicating we don't actually know the real value. At some point I will probably test Stealth to figure out how it actually works, a value of 1 is pretty conservative imo.

Added a TODO indicating Stealth trait needs retail verification.

TeoTwawki (approved):

PR #5401, opened 2024-04-12
[SQL] Fix Nightfall missing Terror additional effect

ampitere:
**_I affirm:_**

Nightfall is missing the Terror additional effect item mod. Terror is an improved Stun and extremely strong so setting the chance and duration pretty low on this. Also uses the Stun sub effect since that seems to make the most sense.

1. !changejob 1 99
2. !additem 19163
3. Equip Nightfall
4. !godmode
5. Go somewhere with enemies that have high HP (I chose Reisenjima, !zone Reisenjima)
6. Attack a monster that isn't immune to Terror

image


sql/item_mods.sql
```diff
@@ -39555,6 +39555,12 @@ INSERT INTO `item_mods` VALUES (19160,501,5); -- ITEM_ADDEFFECT_CHANCE: 5
 
 -- Nightfall
 INSERT INTO `item_mods` VALUES (19163,368,10); -- REGAIN: 10
+INSERT INTO `item_mods` VALUES (19163,431,2);  -- ITEM_ADDEFFECT_TYPE: DEBUFF
+INSERT INTO `item_mods` VALUES (19163,499,16); -- ITEM_SUBEFFECT: 16
+INSERT INTO `item_mods` VALUES (19163,501,5);  -- ITEM_ADDEFFECT_CHANCE: 5
+INSERT INTO `item_mods` VALUES (19163,951,28); -- ITEM_ADDEFFECT_STATUS: TERROR: 28
+INSERT INTO `item_mods` VALUES (19163,952,1);  -- ITEM_ADDEFFECT_POWER: 1
+INSERT INTO `item_mods` VALUES (19163,953,5);  -- ITEM_ADDEFFECT_DURATION: 5
```
TeoTwawki, line 39563: 5 seconds is pretty long. has retails max duration before resist been measured?

ampitere: Couldn't find any info on the wikis. Used Jettatura as reference but that also has a 2 minute cooldown and a resist threshold of 0.5, so 5 seconds may be a bit long for this I agree. Can tweak it to whatever seems best.

TeoTwawki: gonna hold for a bit and see if we can find out its peak before either asking for a change or merging as is

WinterSolstice8: image
10~ seconds?
going to get some more data

WinterSolstice8: image
image
image
For sure 10s base

TeoTwawki: mind blown

got any idea how many total swings and total procs?

WinterSolstice8: Low sample size, about 168 swings, 9 procs, 5.35%ish

TeoTwawki: even for low sample size on a single mob, bigger rate than expected.

so I guess pr need to keep its 5% proc rate and _increase_ the base duration. And someday we'll figure out why the F dark elemental seem to have crazy gobs of terror resist vs this, but nothing else tested did.

ampitere: Updated to 10 second duration.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5403, opened 2024-04-12
[lua] Entering empyrean paradox softlocks player

MowFord:
**_I affirm:_**

~~Changes the default position in the Empyrean Paradox to stop~~ zoning in and immediately triggering a triggerarea creates a mismatch between the server and client player state:
- server thinks the player is in event 100
- client has no idea and can run around

when this happens the player cannot change equipment (packet 0x050 is equipitem) 
image

and if they interact with the BCNM npc the client softlocks. `!release` will release the player and proper functionality resumes.

Edit: reworked the PR to be a more general solution

Zone into Empyrean Paradox via gm command to put you at 0,0,0. Default location puts you on the elevator inside of a trigger area

### DIFF
--- a/scripts/globals/player.lua
+++ b/scripts/globals/player.lua
@@ -229,9 +229,11 @@ xi.player.onGameIn = function(player, firstLogin, zoning)
 
     -- remember time player zoned in (e.g., to support zone-in delays)
     player:setLocalVar('ZoneInTime', os.time())
+    player:setLocalVar('ZoningIn', 1)
 
     -- Slight delay to ensure player is fully logged in
     player:timer(2500, function(playerArg)
+        player:setLocalVar('ZoningIn', 0)
         -- Login Campaign rewards points once daily
         xi.events.loginCampaign.onGameIn(playerArg)
     end)
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -1887,6 +1887,12 @@ namespace luautils
     {
         TracyZoneScoped;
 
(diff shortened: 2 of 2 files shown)


claywar:
Is this a state that should be handled by onZoneIn, or is this something handled separately?  I would prefer we handle these conflicts in a more generic way though as opposed to moving someone

TeoTwawki:
I was thinking there should be another condition besides "I'm in the trigger area" before playing the event. Something that makes sure the client is ready to receive it.

TeoTwawki:
further thought: set localvar on arrival, clear localvar on region leave. on retail the menu will not pop back up until I walk off the platform and then back onto it again.

MowFord:
The triggering of triggerarea happens in `zone.cpp` and `zone_instances.cpp`

can possibly check `PChar->m_Substate` and only activate trigger areas if zero

TeoTwawki:
> The triggering of triggerarea happens in `zone.cpp` and `zone_instances.cpp`
> 
> can possibly check `PChar->m_Substate` and only activate trigger areas if zero

worth trying I think, see if it results in the right behavior

MowFord:
Found `ZoneInTime` localvar, but that adds complexity on doing time comparisons every trigger area entrance. Added a second localvar that gets reset in the 2.5s timer in `xi.player.onGameIn`

works perfectly for this particular scenario.

WinterSolstice8:
This is the cause of a general defect of the server sending packets to the client before they are ready (see Empyreal Paradox teleporter)

WinterSolstice8 (approved):
Seems simple enough

PR #5405, opened 2024-04-12
[cpp] Load bazaar message from the db

MowFord:
**_I affirm:_**

The db rewrite commented this out. Uncommented it crashed the zone but I believe it was due to index mismatch with the new library. This code confirmed to make the bazaar message work with setting, resetting, and preserving on zone.

open menu, scroll right, select "bazaar", then "set message"
enter any message, hit esc
See the message when you check yourself
zone and see that the message persists

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -672,20 +672,17 @@ namespace charutils
             PChar->profile.mhflag = (uint16)rset->getInt("mhflag");
             PChar->profile.title  = (uint16)rset->getInt("title");
 
-            // TODO:
-            /*
             std::array<uint8, 512> bazaarMessageArray;
             db::extractFromBlob(rset, "bazaar_message", bazaarMessageArray);
             const char* bazaarMessageStr = reinterpret_cast<const char*>(bazaarMessageArray.data());
             if (bazaarMessageStr != nullptr)
             {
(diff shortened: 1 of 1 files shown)


src/map/utils/charutils.cpp
```diff
@@ -672,20 +672,17 @@ namespace charutils
...
 
-            // TODO:
-            /*
             std::array<uint8, 512> bazaarMessageArray;
             db::extractFromBlob(rset, "bazaar_message", bazaarMessageArray);
             const char* bazaarMessageStr = reinterpret_cast<const char*>(bazaarMessageArray.data());
             if (bazaarMessageStr != nullptr)
             {
-                PChar->bazaar.message.insert(0, (char*)_sql->GetData(7));
+                PChar->bazaar.message.insert(0, bazaarMessageStr);
```
WinterSolstice8, line 680: I'm not sure why it used Insert to begin with, probably not necessary, but approving anyway.

WinterSolstice8 (approved):

src/map/utils/charutils.cpp
zach2good: Yeah, I hate this pattern, but it's been there forever so it's probably OK

zach2good (approved):

PR #5406, opened 2024-04-12
[core] Random deal functionality

MowFord:
**_I affirm:_**

Ports random deal from WingsXI

~~This is a draft as I'm going to get a capture of the message IDs~~

All the core functionality appears to be retail accurate:
- loaded deck and winning streak traits only show up when merited
- https://www.bg-wiki.com/ffxi/Random_Deal bonuses to appropriate gear
- Checks for loaded deck and give chance to fail if only one ability is on cooldown
- If loaded deck isn't present then get list of all abilities with recast and choose up to 2 abilities

TODO:
- ~~Return true/false properly if abilities are actually reset~~
- ~~get message id captures to ensure we are returning the right messages to each target of the ability (I doubt we should be "evading" if there are no abilities to reset~~

Captures showed that random deal gives message id 324 on a miss and 441 on successful reset. 

image

Test via `!reset`, using JAs then using random deal over and over


src/map/modifier.h
```diff
@@ -601,6 +601,7 @@ enum class Mod
     EXP_BONUS         = 382,  //
     ROLL_RANGE        = 528,  // Additional range for COR roll abilities.
     JOB_BONUS_CHANCE  = 542,  // Chance to apply job bonus to COR roll without having the job in the party.
+    RANDOM_DEAL_BONUS = 573,  // % chance to reset 2 abilities
```
Xaver-DaRed: We have a problem

Down in the file
`// 570 through 825 used by WS DMG mods these are not spares.`

Worst part is we have some of those IDs used when we shouldnt

MowFord: Oops, I'll find a new id

Xaver-DaRed: Theres a list of spare Ids down bellow

MowFord: Updated, nice catch

src/map/lua/lua_baseentity.cpp
```diff
@@ -13063,6 +13063,23 @@ void CLuaBaseEntity::doWildCard(CLuaBaseEntity* PEntity, uint8 total)
...
+ ************************************************************************/
+bool CLuaBaseEntity::doRandomDeal(CLuaBaseEntity* PTarget)
+{
+    if (m_PBaseEntity->objtype != TYPE_PC)
+    {
+        ShowWarning("Invalid entity type calling function (%s).", m_PBaseEntity->getName());
+        return false;
+    }
+
+    return battleutils::DoRandomDealToEntity(static_cast<CCharEntity*>(m_PBaseEntity), static_cast<CCharEntity*>(PTarget->m_PBaseEntity));
```
zach2good, line 13086: PTarget needs to be checked for validity before here

src/map/utils/battleutils.cpp
```diff
@@ -5864,6 +5865,106 @@ namespace battleutils
...
 
+    /************************************************************************
+     *                                                                       *
+     *   Does the random deal effect to a specific character (reset ability) *
+     *                                                                       *
+     ************************************************************************/
+    bool DoRandomDealToEntity(CCharEntity* PChar, CCharEntity* PTarget)
+    {
+        std::vector<uint16> ResetCandidateList;
+        std::vector<uint16> ActiveCooldownList;
```
zach2good: https://github.com/LandSandBoat/server/wiki/Development-Guide#naming--misc

zach2good: ^ lowerCamelCase

zach2good (commented):
Generally seems fine, just needs one style change and one more null check

src/map/lua/lua_baseentity.cpp
MowFord: The battleutil function returns false if it's null

Could probably remove the above check as well?

zach2good: If you try to pass in `PTarget->m_PBaseEntity` when `PTarget` is nullptr, you'll dereference a null ptr and crash.

zach2good: https://www.youtube.com/watch?v=bLHL75H_VEM

zach2good (approved):

PR #5407, opened 2024-04-13
[SQL] Fix Deae Gratia missing drain effect

ampitere:
**_I affirm:_**

Daea Gratia should have an additional HP drain effect but it's currently missing from the items mods. There isn't a whole lot of information on exact proc chance or how much damage it does in retail but there are other similar weapons with the same additional effect implemented (Bloody Rapier, Bloody Sword, Carnage Rapier, Carnage Blade, etc.).

1. !additem 18856
2. !changejob 1 75
3. Equip club
4. Attack a mob


TeoTwawki:
I actually have one of these on retail and this is close, but the highest proc I got out of it was 14 most swings getting 10 or 12 (resists).

ampitere:
> I actually have one of these on retail and this is close, but the highest proc I got out of it was 14 most swings getting 10 or 12 (resists).

Tweaked down to 12 for now.

TeoTwawki:
> Tweaked down to 12 for now.

Sorry you didn't know but we want to use peak value and then resist functions should take it wrong there. _Hopefully_ that is why that one sword already in the mod list has 20 (most drain items got 5 or 10, from a quick looksie). We've got some issues to work out with how these get mathed but if we use the upper limit of what it can do we should be able to fix the proper at some point.

ampitere:
> > Tweaked down to 12 for now.
> 
> Sorry you didn't know but we want to use peak value and then resist functions should take it wrong there. _Hopefully_ that is why that one sword already in the mod list has 20 (most drain items got 5 or 10, from a quick looksie). We've got some issues to work out with how these get mathed but if we use the upper limit of what it can do we should be able to fix the proper at some point.

Tweaked to 14!

TeoTwawki (approved):

EpicTaru (approved):

Xaver-DaRed (approved):

PR #5408, opened 2024-04-13
Fix extractFromBlob overwriting initialized values

cocosolos:
**_I affirm:_**

`db::extractFromBlob` overwrites initialized values. Previously this was done like this:
```cpp

            size_t length = 0;
            char* buf     = nullptr;
            sql->GetData(13, &buf, &length);
            memcpy(&PChar->m_something, buf, (length > sizeof(PChar->m_something) ? sizeof(PChar->m_something) : length));
```
I think this change is equivalent. I also removed the `memset` only because it wasn't used before and I'm not sure if it would cause any unintended behavior with existing data, idk. I think I can add that back without issue if it's wanted, wrapped inside the null check though because we don't want to overwrite initialized data.

Make a new character, check that mission log has initialized (valid) data. I think everything else that uses this used the same pattern before, except maybe monstrosity.

### DIFF
--- a/src/common/database.h
+++ b/src/common/database.h
@@ -275,6 +275,9 @@ namespace db
         if (!rset->isNull(blobKey.c_str()))
         {
             auto blobStr = rset->getString(blobKey.c_str());
+            // Login server creates new chars with null blobs. Map server then initializes.
+            // We don't want to overwrite the initialized map data with null blobs / 0 values.
+            // See: login_helpers.cpp saveCharacter() and charutils::LoadChar
             std::memset(&destination, 0x00, sizeof(T));
             std::memcpy(&destination, blobStr.c_str(), std::min(sizeof(T), blobStr.length()));
         }
--- a/src/common/database.h
+++ b/src/common/database.h
@@ -270,14 +270,13 @@ namespace db
 
         TracyZoneScoped;
 
-        std::memset(&destination, 0x00, sizeof(T));
-
         // If we use getString on a null blob we will get back garbage data.
         // This will introduce difficult to track down crashes.
         if (!rset->isNull(blobKey.c_str()))
         {
             auto blobStr = rset->getString(blobKey.c_str());
-            std::memcpy(&destination, blobStr.c_str(), sizeof(T));
+            std::memset(&destination, 0x00, sizeof(T));
+            std::memcpy(&destination, blobStr.c_str(), std::min(sizeof(T), blobStr.length()));
         }
     }
 


WinterSolstice8 (approved):

cocosolos:
Sorry the ternary was really bothering me.

zach2good:
I put the memset there for two reasons:
- if you call this on an uninitialised struct, it'll make sure it's zerod out
- if you call this on anything and the db result is empty, the destination struct will end up empty.

I don't think there's any situation where we want the original content of the destination struct to remain intact if we've called this on it

cocosolos:
This is used for a lot of stuff that lives in memory and just gets saved to the database, so it gets initialized in memory, then checks the database and if the database is null it just uses the initial values until it needs to save a change.

cocosolos:
I can add it back inside the null check to address your first point. I just don't think it should be clearing the memory if it has nothing to set.

cocosolos:
> I don't think there's any situation where we want the original content of the destination struct to remain intact if we've called this on it

I should note that the login server does the initial char insert with info you pick during creation, then the map server initializes the char object, pulling whatever extra info that exists from the db, it doesn't differentiate between new and existing chars. That's why we have this situation where char info exists in the db before the char is ever actually initialized for the first time. I don't think my first reply was very clear on that.

zach2good:
This isn't a criticism of your proposed change, this is a criticism of the original design.

This still doesn't make sense. The signature of the function is explicitly: `void extractFromBlob(std::unique_ptr<sql::ResultSet>& rset, std::string const& blobKey, T& destination)`, where `destination` is an out variable. It shouldn't be making any guarantees about preserving anything that's passed in. This is enforced by that `memset`. It should be equivalent to `operator=` and not `operator=|`.

If we're relying on logic where calling something on an out variable may allow the contents of it to be preserved, we need to change that logic and not this function.

cocosolos:
I was looking at this function as basically a wrapper for memset. If it's gonna force a write then we could wrap all calls to it with a null check, but at that point the function doesn't really save any work and the logic could just be put inline.

WinterSolstice8:
> If we're relying on logic where calling something on an out variable may allow the contents of it to be preserved, we need to change that logic and not this function.

We have a chicken and egg problem:
https://github.com/LandSandBoat/server/blob/4bae8f2b67b2bae45616696a210b9483685123e2/src/map/entities/charentity.cpp#L153-L165
This is what the mission log looks like on a char that has done (almost) nothing sql update. forgive the formatting

```
FF  FF  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  FF  FF  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  FF  FF  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  FF  FF  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  65  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  02  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  03  00  00  00  01  00  01  00  01  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  00  
```

tl;dr: memsetting this to 0 isn't what the blob "should" look like by default (I am assuming this is some guessed/hacked up value that is close enough to work)

char creation in login doesn't populate the blobs, and CCharEntity() populates this one, which as it currently is written would be overridden by 0s then written to the DB thus causing #5388

WinterSolstice8:
My stance on the memset is that it should be unnecessary -- CCharEntity (and other constructors) are responsible for filling their own data structs with valid defaults. If a db query returns with a "we dont have that field with data, sorry" memsetting that region to 0 is superfluous 

Though, apparently, blobs can have default values now? https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html

zach2good:
For compatibility with old code and unbreaking things for people, let's whip out the memset and make everything look more like the old code internally. I'll go back to the drawing board for fancy helpers.

I'm out tonight, so someone else please approve and merge when ready

zach2good:
Maybe leave comments with these findings, etc

cocosolos:
Added some comments explaining things. Left the memset since it doesn't harm anything where it is and will pad out the struct in the weird case where the db has _some_ information that is smaller than the struct for whatever reason, just to make sure there's no sensitive garbage data left in memory.

> I am assuming this is some guessed/hacked up value that is close enough to work

```
struct missionlog_t
{
    uint16 current;
    uint16 statusUpper;
    uint16 statusLower;
    bool   complete[64];

    missionlog_t()
    {
        current     = 0;
        statusUpper = 0;
        statusLower = 0;

        std::memset(&complete, 0, sizeof(complete));
    }
};
```
This is (edit: part of the) blob struct, so taking the first "line" of the blob, from FFFF to the next FFFF:

`FFFF` = nation and zilart default 'none' mission id is 65535 (client uses this value)
`0000 0000` = 32-bit MissionStatus var, generally made up but uses retail gathered values for CoP, split in 2 to reduce padding/alignment
`[00 ... 00]` = bool for mission[id] completion

x15, 1 for each mission area.

zach2good (approved):

PR #5409, opened 2024-04-13
[cpp] Set model slot on non-pc entities

MowFord:
**_I affirm:_**

Extends the `setModelId()` binding to support setting the model of a slot on a humanoid NPC/MOB

No change if a second param isn't passed:

image

but if a slot is passed, it works to adjust just the slot for a humanoid entity
image

image

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5145,12 +5145,13 @@ uint16 CLuaBaseEntity::getModelId()
  *  Function: setModelId()
  *  Purpose : Updates the Model ID of the entity
  *  Example : mob:setModelId(1168)
- *  Notes   :
+ *  Notes   : Humanoid entities can be passed a slot to change the modelid of that equipment
+ *              npc:setModelId(47, 5) -- Vermillion cloak
  ************************************************************************/
 
 void CLuaBaseEntity::setModelId(uint16 modelId, sol::object const& slotObj)
 {
-    if (m_PBaseEntity->objtype == TYPE_PC)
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -5145,12 +5145,13 @@ uint16 CLuaBaseEntity::getModelId()
...
  *  Example : mob:setModelId(1168)
- *  Notes   :
+ *  Notes   : Humanoid entities can be passed a slot to change the modelid of that equipment
+ *              npc:setModelId(47, 5) -- Vermillion cloak
  ************************************************************************/
 
 void CLuaBaseEntity::setModelId(uint16 modelId, sol::object const& slotObj)
 {
-    if (m_PBaseEntity->objtype == TYPE_PC)
+    if (m_PBaseEntity->objtype == TYPE_PC || slotObj.is<uint8>())
```
claywar, line 5154: This feels pretty unsafe, unless all entities can have these, which then the PC check is unnecessary.

MowFord: I suppose the PC check is unnecessary here. I just didn't want to inadvertantly break something that was sending a model update to the main weapon of a player

The goal was to preserve existing functionality 100%

Essentially mobs that aren't humanoid wouldn't play well with this change, so perhaps we could change this to find a model flag that implies they're humanoid (player/mob/npc), I'm just not familiar with what that would be

WinterSolstice8 (approved):

PR #5416, opened 2024-04-14
convert quest An Explorers Footsteps to new quest framework and fix seveal bugs

Grahf0085:
**_I affirm:_**

This will convert the quest an Explorers Footsteps to the new quest framework. Also, the quest had several bugs where Abelard would stop working if you turned the tablets in to him in several certain orders. There are bug reports in ASB/Horizons Issue tracker related to this. This fixes those bugs.

I'm not sure how ampitere's commits got into my PR. I can close this PR and make a new one. Also, although this quest doesn't include a capture from retail it does have comments in the code where I'm not sure if it matches retail exactly. But it is an improvement over the current code in several ways: doesn't break the NPC Abelard, uses cutscene 45 when a player tries to turn in the same tablet twice (old code doesn't appear to handle that situation with a cutscene), and Abelard wont tell the player to go to a stone monument they have already turned in a tablet for.

Complete the quest An Explorers Footsteps


hooksta4:
https://www.youtube.com/watch?v=XAqdcjoYmuA

Grahf0085:
I'm going to be opening a new cleaner PR tomorrow. It will have a few changes from the capture @hooksta4  mentioned :)

Grahf0085:
I'm going to be opening a new cleaner PR tomorrow. It will have a few changes from the capture @hooksta4 mentioned :)

PR #5421, opened 2024-04-14
[core] Consume mana

MowFord:
**_I affirm:_**

Consume mana should apply to the base weapon damage for both auto attacks and weaponskills. Consensus is pretty set that it's applying to base weapon damage for the first hit only of a weaponskill (similar to souleater applying to each hit individually as well)

This PR does 2 things in cpp:
- fix the naming of an attack-round attack parameter to be generic "increase to base damage", and rework the logic to accurately apply "augments SA" and "augments TA" separately (before this they would have double-dipped)
- update the location of the consume-mana bonus to be base damage before pdif, etc

This PR also adds a small change to weaponskills to apply consume mana bonus to base damage:
- small local helper function that would likely be converted to a global `xi.combat` function when attack rounds get moved to lua
- adjust `getSingleHitDmg` function to pass ftp so we can appropriately boost base damage before ftp

buff doesn't wear on missed ws
image

buff doesn't wear on missed melee, but still works on hits
image

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -117,6 +117,18 @@ local scarletDeliriumBonus = function(attacker)
     return bonus
 end
 
+local consumeManaBonus = function(attacker)
+    local bonus = 0
+
+    if attacker:hasStatusEffect(xi.effect.CONSUME_MANA) then
+        bonus = math.floor(attacker:getMP() / 10)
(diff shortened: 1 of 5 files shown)


WinterSolstice8 (approved):
looks like the Vajra bonuses are wrong, but thats out of scope of this PR.

Issue #5419, opened 2024-04-14
🐛 Conquest weekly calculations incorrect


MowFord:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base`

## Steps to reproduce

Our live server is running the db rewrite and weekly calculations assigned everything to beastmen control with a 3-way tie for first:

![image](https://github.com/LandSandBoat/server/assets/131182600/08de563c-4080-4282-b38c-378b84f4ae5f)

we can pull backups to give existing conquest data, but it is hard/impossible to imagine there weren't _some_ winning nations

based on [this commit](https://github.com/LandSandBoat/server/commit/508ae02ef338837baae1ec85389aeaa98a4fb6b2) everything looks like it was mapped over properly, though one concern I have is the slight change to the logic in the `ConquestSystem::updateInfluencePoints` function, which might mean that influence never accumulates in regions?

I'm not sure. I'm very much not acquainted with the conquest system in detail but can test/provide whatever is needed to get this sorted

MowFord:
Update: 

gaining influence resets to zero across teh board:

![image](https://github.com/LandSandBoat/server/assets/131182600/434e9c50-ab8a-4ac9-9ab8-a3de0aadec6f)

went into ronfaure as windurst citizen and killed a bit, region 0 updated to zero across th board 

![image](https://github.com/LandSandBoat/server/assets/131182600/368e281d-6d64-4b21-ba43-13d72f60225c)

same for dunes:

![image](https://github.com/LandSandBoat/server/assets/131182600/bbe43ec8-60f4-4dd0-8d26-5a40503701c0)

MowFord:
PR merged

PR #5425, opened 2024-04-14
[world] Update influence in xi_world to not reset values on gain

MowFord:
**_I affirm:_**

Decided to dig into #5419 myself.

Turning off world server stopped the influence reset, which led me to trace the logic from a `MSG_MAP2WORLD_REGIONAL_EVENT` message.

Found in [this commit](https://github.com/LandSandBoat/server/commit/07a9e09c1643054b507788922f9f932291636a47) the code that changed and noticed a duplicate definition of `int influences[4]`. The issue in #5419 is due to a scope inception where the code that runs is ignored and influences[4] is used with default values.

It's not clear to me why this wouldn't show nonzero influence for the nation gaining points with this bug, but this PR appears to fix it. I'm assuming the scheduled calculations ensure influence is not zero across the board?

Reset region_id 0's values in `conquest_system` table to zeroes in all 3 Nationcolumns and 5000 in beastmen
change job to level 1, give signet, and kill a mob in ronfaure
see influence update:
image

### DIFF
--- a/src/world/conquest_system.cpp
+++ b/src/world/conquest_system.cpp
@@ -186,28 +186,24 @@ bool ConquestSystem::updateInfluencePoints(int points, unsigned int nation, REGI
         return false;
     }
 
-    int influences[4]{};
+    std::string query = "SELECT sandoria_influence, bastok_influence, windurst_influence, beastmen_influence FROM conquest_system WHERE region_id = %d";
 
+    auto rset = db::query(fmt::sprintf(query.c_str(), static_cast<uint8>(region)));
+    if (!rset || rset->rowsCount() == 0 || !rset->next())
     {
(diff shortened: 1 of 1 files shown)


MowFord:
also update to those wanting to import this change:
- you can recompile only the world server to get the fix
- run this sql to fix the influence values: `UPDATE conquest_system SET beastmen_influence = 5000 WHERE sandoria_influence = 0 AND bastok_influence = 0 AND windurst_influence = 0`
- let players collect conquest influence for X amount of time
- have a gm run `!updateconquest 0` to initiate a weekly calculation

WinterSolstice8:
Interesting. I would have expected CI to catch shadowing. hmm...

WinterSolstice8 (approved):

PR #5427, opened 2024-04-14
[LUA] Fix players appear in Aht Urhgan fountain during ship arrival cutscenes

ampitere:
**_I affirm:_**

When arriving into Aht Urhgan from a ship players initially appear at the default location (0, 0, 0) which coincidentally is inside of the fountain in the center of the city. While here they are able to be targeted by other players.

Instead of keeping them here we will instead place them underneath their intended arrival point so that they are not targetable by other players and load the general area they're supposed to first. Also cleans up a bit of the if/else logic since they use the same cutscene id.

The positions used are nearly identical to the actual position after the cutscene, however they are underneath the ground so they are not seen/targetable by other players. If we place them at the actual intended arrival position then they appear during the cutscene which looks a bit strange. If there is a more optimal position for this please let me know.

1. !zone Silver Sea route to Al Zahbi or !zone Open Sea route to Al Zahbi
2. !zone Aht Urhgan Whitegate or wait for the ship to dock

### DIFF
--- a/scripts/zones/Aht_Urhgan_Whitegate/Zone.lua
+++ b/scripts/zones/Aht_Urhgan_Whitegate/Zone.lua
@@ -26,10 +26,13 @@ zoneObject.onZoneIn = function(player, prevZone)
         player:getZPos() == 0
     then
         if prevZone == xi.zone.OPEN_SEA_ROUTE_TO_AL_ZAHBI then
+            player:setPos(-11, 5, -142, 192)
             cs = 201
-        elseif prevZone == xi.zone.SILVER_SEA_ROUTE_TO_AL_ZAHBI then
-            cs = 204
-        elseif prevZone == xi.zone.SILVER_SEA_ROUTE_TO_NASHMAU then
+        elseif
+            prevZone == xi.zone.SILVER_SEA_ROUTE_TO_AL_ZAHBI or
+            prevZone == xi.zone.SILVER_SEA_ROUTE_TO_NASHMAU
+        then
+            player:setPos(11, 5, 142, 64)
             cs = 204
         else
             -- MOG HOUSE EXIT


WinterSolstice8:
you probably will want to add the cutscene flags if retail uses them:

https://github.com/LandSandBoat/server/blob/0829c7e4ae8e1704a2499ed2d40fe6e1b45922ff/scripts/enum/cutscene_flag.lua#L10
https://github.com/LandSandBoat/server/blob/0829c7e4ae8e1704a2499ed2d40fe6e1b45922ff/scripts/quests/hiddenQuests/New_Character_Cutscenes.lua#L22

(I am not sure there is a good mechanism for this yet?)

WinterSolstice8 (approved):

scripts/zones/Aht_Urhgan_Whitegate/Zone.lua
```diff
@@ -26,10 +26,10 @@ zoneObject.onZoneIn = function(player, prevZone)
         player:getZPos() == 0
     then
         if prevZone == xi.zone.OPEN_SEA_ROUTE_TO_AL_ZAHBI then
+            player:setPos(-11, 5, -142, 192)
             cs = 201
-        elseif prevZone == xi.zone.SILVER_SEA_ROUTE_TO_AL_ZAHBI then
-            cs = 204
-        elseif prevZone == xi.zone.SILVER_SEA_ROUTE_TO_NASHMAU then
+        elseif prevZone == xi.zone.SILVER_SEA_ROUTE_TO_AL_ZAHBI or prevZone == xi.zone.SILVER_SEA_ROUTE_TO_NASHMAU then
```
WinterSolstice8: This is what CI is asking you to do
```suggestion
        elseif
            prevZone == xi.zone.SILVER_SEA_ROUTE_TO_AL_ZAHBI or
            prevZone == xi.zone.SILVER_SEA_ROUTE_TO_NASHMAU
        then
```

ampitere:
> you probably will want to add the cutscene flags if retail uses them:
> 
> https://github.com/LandSandBoat/server/blob/0829c7e4ae8e1704a2499ed2d40fe6e1b45922ff/scripts/enum/cutscene_flag.lua#L10
> 
> 
> https://github.com/LandSandBoat/server/blob/0829c7e4ae8e1704a2499ed2d40fe6e1b45922ff/scripts/quests/hiddenQuests/New_Character_Cutscenes.lua#L22
> 
> (I am not sure there is a good mechanism for this yet?)

I was wondering if there was a way to place the players at the correct position but just not show them during the cutscene and it seems like this is a good direction to go with. I'll look into this more.

scripts/zones/Aht_Urhgan_Whitegate/Zone.lua
ampitere: Formatting updated.

EpicTaru (approved):

PR #5331, opened 2024-04-02
[sql] Mhaura moogle visible underneath Laughing Bison

MowFord:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

Unsure what this moogle's intended use is, but it's definitely not supposed to be visible. There's no references to the NPC id anywhere, nor is there a lua file for it. Probably an asset for some cutscene. This PR simply sets the flag to hide it.

image

using `!npchere` to move laughing bison:

image

PR #5428, opened 2024-04-14
[SQL] Remove random Moogle inside Aht Urhgan fountain

ampitere:
**_I affirm:_**

There's a random Moogle hanging out inside the fountain in Aht Urhgan for some reason. You are able to interact with him but he does not do anything. Seems really similar to: https://github.com/LandSandBoat/server/pull/5331

1. !pos 0 0 1 50 (if you use 0 0 0 it will place you at the mog house exit)

image


sql/npc_list.sql
```diff
@@ -3988,7 +3988,6 @@ INSERT INTO `npc_list` VALUES (16982044,'_1es','',0,-11.692,1.690,-144.423,1,40,
 INSERT INTO `npc_list` VALUES (16982045,'_1et','',0,12.066,1.385,144.424,1,40,40,9,0,0,0,6147,0x0200000000000000000000000000000000000000,0,NULL,0);
 INSERT INTO `npc_list` VALUES (16982046,0x14,'',0,-20.726,3.389,-154.231,98311,40,40,18,0,0,2,2051,0x0400000000000000000000000000000000000000,32,NULL,1);
 INSERT INTO `npc_list` VALUES (16982047,'Zhamwaa','Zhamwaa',0,-103.323,0.001,-76.504,8,50,50,0,0,0,0,27,0x01000301AE10AB20AB309F409F50006000700000,32,'TOAU',1);
-INSERT INTO `npc_list` VALUES (16982048,'Moogle','Moogle',64,0.000,0.000,1.500,32769,50,50,0,1,0,0,3,0x0000520000000000000000000000000000000000,0,'TOAU',1);
```
WinterSolstice8, line 3991: Is this the right way to do this? A re-cap in the zone in the future will just place them right back there.

Xaver-DaRed:
I have a request. Check if the Moghouse moogle is still visible after this change

ampitere:
> I have a request. Check if the Moghouse moogle is still visible after this change

Interesting it does seem to be the Moghouse moogle, that's super weird.

sql/npc_list.sql
ampitere: Actually seems this is the Moghouse moogle so definitely don't want to remove it.

Xaver-DaRed:
You see, the mog house is NOT a different zone than the zone you enter it from. Its just a concrete position of the map with a different zone model applied

ampitere:
> You see, the mog house is NOT a different zone than the zone you enter it from. Its just a concrete position of the map with a different zone model applied

Would it instead be better to move him to a different location or do you think that would have adverse effects as well?

Xaver-DaRed:
Moving it would also be counter-productive. you wouldnt be able to access it from MH

ampitere:
> Moving it would also be counter-productive. you wouldnt be able to access it from MH

Well I'm definitely out of ideas then haha. I'm fine with closing this if there isn't something else you'd like me to try.

Xaver-DaRed:
Nothing that can be done unless we find a way to dynamically load npcs per player in non-instanced zones. So yeah, close it. Stuff happens, it's alright.

ampitere:
MowFord pointed out it might be missing a flag, and it was missing a status set to 2 which the other Moghouse moogle's have. Adding that caused it to not appear in the fountain but still appear inside the Moghouse so I will open a separate PR for that fix.

TeoTwawki:
ok Teo story time.

The way the mog house moogles have always worked is there is this bit of code in the core that goes looking for an NPC with the face/head (because model type index 0 only uses 2 bytes) - and i mena like, its not even a specific Moogle NPC ID used, its looking for ANY moogle it can find with that name+face combo - and then very quickly sends a hide+plus unhide in an order and pacing that result sin the player inside the moghouse only getting the packet to SHOW the npc, while for everyone else outside its re-hidden before the client can even display it. Unless some freak accident has them standing exactly in the right spot at the right time or with the right amount of lag OR if that specific moogle entry was never flagged correctly.

This is trash, and I always meant to replace it with a flag for "instance this to the mog house". We need the same thing for the Symphonic Curator anyways. Every time I was going to make the change, I wound up in rebase hell because another months updates to the client landed. SE has stopped ripping very single zone to shreds now and monthly changes are minimal, but my irl is so messed up I just don't have the time anymore. 

My original intent was to DROP an unused flag field and repurpose it as a real flag field in the same spot (so zero that column, redefine whats being saved there and make get/set for it). On retail that field we have now is a step counter we aren't using - watch Ramenel in South sandy, every time he's moving that number is climbing. We get this filled in all the time from caps but it means nothing for use because we aren't using anything like what SE does for our pathing. So I was gonna drop that completely so ppl stop bothering to fill it in. Then in that same spot I'd have out now field for controlling if that npc is only ever visible in special circumstance. 

At the time, that would have been mog house only and could have been a simple on/off for it but now mog garden is a thing too which has a TON of objects that have conditions before appearing. So a mask here is beneficial.

**Deleting 1 moogle out of NPC list will not fix this issue. Flag it to unspawned/hidden in its "status" in SQL if it is not already. If that's not fixing it you can't fix this with just SQL and are going to have to live with it longer.**

**_P.S. Sometimes when you leave the moghouse, you can see SE's own mog house moogle in position x 0 y 0 z0 also. As in, they have a similar problem and then send you a packet to hide it after! sometimes the client never even gets that hide packet at all and then the lil buggers menu's partially work! I can check my delivery box and mail stuff even. job change doesn't function though, neither does checking plants or re-arranging furniture._**

PR #5430, opened 2024-04-15
fix content tags of several COP NPCs

Grahf0085:
**_I affirm:_**

Puts COP content tag on several COP era NPCs.

Tesch Garanjy stores artifact armor and here (https://www.bg-wiki.com/ffxi/The_History_of_Final_Fantasy_XI/2005) it says the "The Artifact Armor Storage System was implemented." February of 2005.

Venessa is a ENM quest activator - ENMs were added during COP.

Wobke gives the quest A Chocobo's Tale and a prerequisite for that quest is PM 3-2. So it's likely she was added around the same time PM3-2 was added.

Run npc_list.sql


Grahf0085:
Oh and Wilhelm takes drops from Ultima and Omega and gives you gear in return.  Ultima and Omega and both in Limbus - a COP zone.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

Issue #4789, opened 2023-12-01
🐛 Teleport in a zone with a mog house can steal party members from the Mog House


MowFord:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base`

## Steps to reproduce

Go to south sandy as a whm on the ramp in front of the AH (position 0, 0, 0)
Get a party member to go into the mog house in south sandy and also stand at 0,0,0

WHM teleport will grab the player in the mog house

This also works for teleport items. I believe the proper fix would be to `targetfind.cpp`, though it could be fixed with lua bindings in each file as well

## Expected behavior

Player in MH is excluded from teleport

MowFord:
was closed by the referenced PR

PR #5431, opened 2024-04-15
[CPP] Fix party members area targetable in Mog House

ampitere:
**_I affirm:_**

When 2 characters are in a group together when one is inside a Mog House there is no check to prevent that character from being buffed or teleported by the other party member. This adds a checks and the necessary method to CBattleEntity to check if the current player character entity is inside of a Mog House. Fixes issue: https://github.com/LandSandBoat/server/issues/4789

1. Log in with 2 characters
2. Move one character as close to 0,0,0 as possible (easiest place is probably Aht Urhgan Whitegate next to the fountain !pos 10 0 10 50)
3. Move the other character inside the Mog House (!pos -100 0 -80 50 will put you close)
4. Both characters need to be in a party together
5. Use an area targetable ability or spell on the character outside (any -ra spells, teleports, job abilities, etc.)

### DIFF
--- a/src/map/ai/helpers/targetfind.cpp
+++ b/src/map/ai/helpers/targetfind.cpp
@@ -288,7 +288,10 @@ void CTargetFind::addAllInParty(CBattleEntity* PTarget, bool withPet)
     {
         static_cast<CCharEntity*>(PTarget)->ForPartyWithTrusts([this, withPet](CBattleEntity* PMember)
         {
-            addEntity(PMember, withPet);
+            if (!PMember->isInMogHouse())
+            {
+                addEntity(PMember, withPet);
+            }
         });
     }
     else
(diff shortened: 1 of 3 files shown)


WinterSolstice8:
I feel like you can't cast in the moghouse, and in addition to that there would need to be a check if they are in the same moghouse (in a city for instance)

WinterSolstice8:
I reread and understand now, this is anti-griefing for people who stand near 0,0,0

ampitere:
> I reread and understand now, this is anti-griefing for people who stand near 0,0,0

Right the person who is outside is able to affect the person inside the Mog House. The person inside the Mog House is not doing any casting/abilities.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5436, opened 2024-04-15
[QUEST] An Explorers Footsteps Interaction Framework

Grahf0085:
**_I affirm:_**

This PR translates the quest "An Explorers Footsteps" to the new quest framework.  It also fixes several bugs that have been found in the way the quest is currently implemented. Those bugs are mentioned in the ASB/Horizon issue trackers. 

Hookstar provided this capture from someone named Wiggo: https://www.youtube.com/watch?v=XAqdcjoYmuA&list=LL&index=1
He turned in three tablets and I mostly mimic what retail does with a couple exceptions.... 

For example, when Wiggo turns at stone tablet at 58:54 in the video it's from Konschat Highlands but the Mayor request a tablet from La Theine. The mayor  gives the same dialog as every other time Wiggo turned in a stone tablet. I take a different approach and use a cutscene(46) where the Mayor tells you that you gave him the wrong tablet. 

Wiggo doesn't capture turning in he same tablet twice. I utilize cutscene 45 if a player tries to turn in the same tablet twice. I believe this is correct.

Fixes these issues: 
https://github.com/HorizonFFXI/HorizonXI-Issues/issues/814
https://github.com/AirSkyBoat/AirSkyBoat/issues/2591
https://github.com/AirSkyBoat/AirSkyBoat/issues/2532

Accept the quest from Abelard and complete it.


scripts/globals/interaction/container.lua
```diff
@@ -145,6 +145,20 @@ function Container:unsetVarBit(player, name, bitNum)
...
+    while currentValue > 0 do
+        if bit.band(currentValue, 1) == 1 then
+            count = count + 1
+        end
+
+        currentValue = bit.rshift(currentValue, 1)
+    end
+
+    return count
+end
```
zach2good, line 160: Pretty nifty. For passers by: This is shifting and truncating the target number until it comes back zero, which means we've hit and counted the last possible set bit - so we can stop iterating.

scripts/quests/otherAreas/An_Explorers_Footsteps.lua
```diff
@@ -0,0 +1,358 @@
...
+        npcUtil.giveItem(player, xi.item.CLAY_TABLET)
+        quest:setVar(player, 'currentTablet', currentTablet)
+    end
+end
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            return status == QUEST_AVAILABLE and player:getFameLevel(xi.quest.fame_area.SELBINA_RABAO) >= 1 --unsure where they are getting rabao fame from
```
zach2good: What is this comment meant to mean?

scripts/quests/otherAreas/An_Explorers_Footsteps.lua
```diff
@@ -0,0 +1,358 @@
...
+    800, -- South Gustaberg -- !gotoid 17216174
+    1000, -- Konschtat Highlands -- !gotoid 17220150
+    1000, -- Pashhow Marshlands -- !gotoid 17224302
+    3000, -- Rolanberry Fields -- !gotoid 17228353
+    800, -- West Sarutabaruta -- !gotoid 17248818
+    800, -- East Sarutabaruta -- !gotoid 17253064
+    1000, -- Tahrongi Canyon -- !gotoid 17257053
+    1000, -- Buburimu Peninsula -- !gotoid 17261140
+    1000, -- Meriphataud Mountains -- !gotoid 17265242
+    10000, -- Sauromugue Champaign -- !gotoid 17269222
```
zach2good: If a server is set up as multi-process, gotoid won't work here. These could/should be !pos with x, y, z, zone, etc.

scripts/quests/otherAreas/An_Explorers_Footsteps.lua
```diff
@@ -0,0 +1,358 @@
+-----------------------------------
+-- An Explorer's Footsteps
+-----------------------------------
+-- Log ID: 4, Quest ID: 19
+-- This quest was changed to require a minimum amount of fame to combat RMTs POS-Hacking around to
+-- quickly earn gil. However, as this is not a legitimate concern on private servers players may
+-- complete this quest even with no fame.
```
zach2good: So, what is the actual retail implementation? We want that. 

If you want to apply popular modifications on top of this, you can also supply a module alongside it

zach2good (commented):
Seems fine to me, but I'm not a quest/IF guy

scripts/quests/otherAreas/An_Explorers_Footsteps.lua
gweivyth: I just did this quest on retail with a fairly new character and didn't encounter any sort of fame requirements.  The wiki mentions neeing Selbina fame level 1 but I have done nothing but my subjob quest in Selbina and was able to complete this.

scripts/globals/interaction/container.lua
Grahf0085: I'll remove this

scripts/quests/otherAreas/An_Explorers_Footsteps.lua
Grahf0085: I'll remove this

scripts/quests/otherAreas/An_Explorers_Footsteps.lua
Grahf0085: Will update soon

scripts/quests/otherAreas/An_Explorers_Footsteps.lua
Grahf0085: > So, what is the actual retail implementation? We want that.
> 
> If you want to apply popular modifications on top of this, you can also supply a module alongside it

This comment was at the top of Abelard.lua (the NPC who gives the quest). I moved it from Abelard.lua to An_Explorers_Footsteps.lua but I'm not sure what it means. I've checked a few websites and can't find anything about this quest having a fame requirement added due to RMT. I think that corroborates  @gweivyth  comment about doing the quest with no fame. Can't find anything about a frame requirement for this quest anywhere so I'm not sure where the original author of that comment got their info from. 

I can remove the comment since I think it's superfluous.

PR #5444, opened 2024-04-16
[Magic] Banish Spell Functionality

UmeboshiXI:
Implements Banish SDT Debuffs when used vs undead mobs. Implements Afflatus Misery functionality to Banish spells. Implements WHM Merit bonuses for Banish Spells.
Implements gear bonuses for "Potency of Banish vs Undead"

**_I affirm:_**

Implements various missing functionality for Banish series spells.
See link for more details on specific SDT Values.
https://www.bg-wiki.com/ffxi/Category:Banish_Spells

1. Adds missing SDT debuff vs undead type mobs.
2. Adds Afflatus Misery bonus for Banish series spells.
3. Adds WHM Banish Merit bonuses.
4. Adds gear modifiers for "Potency of Banish vs Undead" (BANISH_POTENCY)

**_Things that still need to be added:_** Duration resistance. Duration resistance should be independent of spell damage resists. (Currently unsure what all influences this)

Things to note:
- The SDT reduction can not go above 100% (Will never give bonus damage, only reduces resistance value like Tomahawk).
- Gear modifiers are calculated multiplicatively. Example: Banish 1 = -50% reduction to SDT resistance. Gear Mod = 10. Final Value would be -55%, not -60%.

1. Find Skeleton.
2. Use `!getmod pierce_sdt` on mob. Note value as 500 (50% damage reduction).
3. Cast banish spell on mob.
4. Use `!getmod pierce_sdt` on mob. See that the value has changed.
5. Wait for effect to wear off, check once more and see that the value goes back to 500.
6. Use `!additem clerics_mitts` , equip item. Cast Banish again.
7. Use `!getmod pierce_sdt` on mob. See that the value has changed from step 4.

8. Banish Merits give +2 MATT to Banish spells and +2 seconds to the duration.

9. Afflatus Misery: Spell damage increases based on last damage instance.
See: https://wiki.ffo.jp/html/1963.html


scripts/globals/spells/damage_spell.lua
```diff
@@ -616,6 +661,14 @@ xi.spells.damage.calculateMagicBonusDiff = function(caster, target, spellId, ski
...
     end
 
+    -- WHM Catagory 1 Merits (Banish)
+    if
+        (spellId >= xi.magic.spell.BANISH and spellId <= xi.magic.spell.BANISH_V) or
+        (spellId >= xi.magic.spell.BANISHGA and spellId <= xi.magic.spell.BANISHGA_V)
+    then
+        mab = mab + (caster:getMerit(xi.merit.BANISH_EFFECT) * 2) -- +2 Magic Attack Bonus for each merit level.
+    end
+
```
Xaver-DaRed: Banish merits do not give MAB. It just adds 2 to base dmg per merit

MowFord: also keep in mind that merits have a value, which defines what you get per merit point:

image

5 points in banish effect would return `caster:getMerit(xi.merit.BANISH_EFFECT) == 10`

Xaver-DaRed:
Do we really need a new status effect? Do tomahawk and this stack? Couldnt we just reuse/rework tomahawk status effect to work with this? The amount of status effects is limited, we cant use them willy nilly.

Also, logic should not under any circustance be present in those spell scripts.

UmeboshiXI:
> Do we really need a new status effect? Do tomahawk and this stack? Couldnt we just reuse/rework tomahawk status effect to work with this? The amount of status effects is limited, we cant use them willy nilly.
> 
> Also, logic should not under any circustance be present in those spell scripts.

I'll try folding it into the Tomahawk effect.

For spells that add special enfeebling effects like Banish, Ancient Magic, BLM -Ja spells etc, where/how would you recommend we handle that logic going forward?

Xaver-DaRed:
Plan was to create a function for this kind of cases (Non-resistable enfeebles) inside enfeeble_spell.lua and add a callback in the spell script, so no actual logic ever resides inside the spell scripts.

Besides, its the same piece of code repeated anyway.

UmeboshiXI:
Made adjustments. If you're okay with the function, I'll also port related spells to use it in future (Ninja Nukes, AM etc).

Also do you have any recommendations of ways/plugins for finding errors (Specifically whitespace). I'm always overlooking them.

Xaver-DaRed:
> I'll try folding it into the Tomahawk effect.
> 
> For spells that add special enfeebling effects like Banish, Ancient Magic, BLM -Ja spells etc, where/how would you recommend we handle that logic going forward?

So, my question was if it stacked or not. We need to go to retail and test if they do stack or not.
If they do stach we do need a different status effect, after all.

The new function in enfeebling spells file seems fine.

The added modifier to gear looks fine aswell. Maybe could have been a different PR altoguether to unclutter this one.

As for the style checks, if whitespace is your issue, most coding programs, like VS and Notepad++ have a setting to display them in screen, so you can actually see them.
In any case, CI catches us all. When that hapens, it tells you where the issue is.

UmeboshiXI:
Tomahawk does not stack with Banish. Banish can be overridden by higher tiers of banish (Banish 2 will overwrite Banish 1 but not the other way around)

Xaver-DaRed:
I havent forgotten about this. Sorry.

MowFord:
Womp womp

Butsuon:
I realize this is closed, but banish still doesn't work in 2026.

UmeboshiXI:
> I realize this is closed, but banish still doesn't work in 2026.

Been a long time. Will get around to making a new updated PR some time soon once I have less on my plate.

PR #5445, opened 2024-04-16
[lua] Instance Entry logic cleanup

MowFord:
**_I affirm:_**

### Issue 1
First off, instance entry requirement checks and join logic wasn't properly filtering on the zoneid of the initiator:
- a player could be in another zone in the same map process and if all players met requirements (key item check, mission progress, etc), the instance would start and all players in the party (within the map process) would be entered into the instance

### Fix 1
This PR fixes that by adding proper zoneid filters in both the `xi.instance.onEventUpdate` and `xi.instance.onInstanceCreatedCallback` player-party loops

### Issue 2
Second, and closely related, was the fact that the initiator wasn't always getting the proper event packets to get sent into the instance. This caused everyone in the party to be stuck at a black screen:
- initiator stuck at the runic portal in a black screen
- rest of the party stuck at a black screen inside of nyzul

image

### Fix 2
This PR "fixes" that by adding a catchall timer on every player that meets entry requirements after instance is confirmed created (inside of `xi.instance.onInstanceCreatedCallback`) to print to the map server and use `setPos` to send the player into the instance. This ensures that even parties with players getting incorrect event timings (or i assume client version issues that cause the events to not function properly) will eventually get into the instance and start the encounter

Also, small change was made to the _order_ of calling `setInstance` on players to always make the initiator (the `player` in the callback) the instance commander and also send the instance join event last to them. This appeasr to make the process work more cleanly to not cause `player:getEventTarget()` to return nil due to initiator being done with the initial cutscene. This issue is actually caught by the catchall 35s timer, but best to not let situations exist that can hit that

### Issue 3
Currently, there's commented code in the nyzul isle runic portal entrance (position `125, 0, 20` in alzadaal undersea ruins) that blocks the call to cancel the event. It is listed as having messed up entry to Path of Darkness instance. 

This causes the initiator's game to stay in the event forever while party members exist in the zone that don't meet the entry requirements

I started digging and found that the issue lies in `xi.instance.onEventUpdate` returning false if a player doesn't meet the entry requirements, but returns `player:getInstance() ~= nil` if the party does meet the requirements. This is a fundamental flaw as the `player:createInstance` call adds the instance to a queue to register the player into an instance. This is not instant.

### Fix 3
This PR fixes that by adding a counter to the logic in `xi.instance.onEventUpdate` to allow it to check a few times before failing out and returning false, which causes the initiator to drop out of the event, and updated nyzul runic portal to support the softer return logic in `xi.instance.onEventUpdate` (Note: every other call to this doesn't even check the requirements properly

### Notes
Note that the choice of 35s fallback timer and 9 checks for `player:getInstance() ~= nil` is arbitrary and best-effort to make a clean server-client interaction. These values could be updated of course. 

There would be a slight issue if 9 tries exited out, but then the instance eventually did the callback. ultimately the instance would load and bring all players in, but depending on how long that time between `createInstance` and the callback was, the player could run out of range, etc. The catchall that brings the player into the zone happens on the callback, so if the timing was too bad then a player could zone in and be included in the fight.

Testing that the entry logic to limit to same zone:
- create party of two members, edit `scripts\zones\Nyzul_Isle\instances\path_of_darkness.lua` to return true for `registryRequirements`
  - have one player stand at entrance to alz, other enter the runic portal. You'll get message that player is too far and cleanly exit event
  - have one player stand in aht urgan, other player can enter Path of Darkness no problem

Testing updated event timing to get all players into zone
- Have two members stand at the runic portal and enter the instance
  - have one or both run `!release` after the teleport animation starts
  - see that after 35s they get sent into instance anyway with a map server log

Testing the cleaner returns of `xi.instance.onEventUpdate`
- edit `scripts\globals\instance.lua` to comment out `player:createInstance(instanceId)`
  - This will cause the logic to work exactly the same, but we'll never get the callback from instance creation: `xi.instance.onInstanceCreatedCallback`
- enter like normal and see that after a bit of a wait, the player is returned to normal and entry event exits
  - adding prints to where we disabled `createInstance` as well as inside the `onEventUpdate` function, you'll see this in the map server log
image
image

A proper flow of entering with 2 members can hit this `xi.instance.onEventUpdate` function 3 times with a pretty beefy server setup:
image
Confirming the need to let this `player:getInstance() ~= nil` check be a little softer

### DIFF
--- a/scripts/globals/instance.lua
+++ b/scripts/globals/instance.lua
@@ -367,7 +367,10 @@ xi.instance.onEventUpdate = function(player, csid, option, npc)
 
     if party ~= nil then
         for _, v in pairs(party) do
-            if v:getID() ~= player:getID() then
+            if
+                v:getID() ~= player:getID() and
+                v:getZoneID() == player:getZoneID()
+            then
                 -- Check entry requirements for party
                 if not checkEntryReqs(v, instanceId) then
                     player:messageText(npc, ID.text.MEMBER_NO_REQS, false)
@@ -377,7 +380,7 @@ xi.instance.onEventUpdate = function(player, csid, option, npc)
                 end
 
                 -- Check everyone is in range
-                if v:getZoneID() == player:getZoneID() and v:checkDistance(player) > 50 then
+                if v:checkDistance(player) > 50 then
                     player:messageText(npc, ID.text.MEMBER_TOO_FAR, false)
                     player:instanceEntry(npc, 1)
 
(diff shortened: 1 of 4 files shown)


MowFord:
I believe `fmt` needs to be added [here](https://github.com/LandSandBoat/server/blob/38b6371c988b5d9dc879828b813a8c5e3ec0375b/tools/ci/lua.sh#L46)

Edit: my weaponskills PR is also failing CI due to this

scripts/globals/instance.lua
```diff
@@ -414,20 +428,49 @@ xi.instance.onInstanceCreatedCallback = function(player, instance)
...
-                v:instanceEntry(npc, 4)
+                v:setInstance(instance)
+                local npc = player:getEventTarget()
+                if npc ~= nil then
+                    v:instanceEntry(npc, 4)
+                end
+
+                v:timer(35000, function(playerArg)
+                    -- failsafe to bring all players into instance
+                    -- taking care not to shadow any existing variables
```
zach2good: You don't need to leave comments here saying you're not shadowing vars - CI will fail you if you do

scripts/globals/instance.lua
```diff
@@ -414,20 +428,49 @@ xi.instance.onInstanceCreatedCallback = function(player, instance)
...
+                    -- if a player doesn't receive the event packet, the whole party will be stuck in blackscreen
+                    -- timer is destroyed if onEventFinish works properly and player gets zoned
+                    -- a properly-functioning event loop will take 20s to zoning into the instance
+                    -- this _should_ be completely unnecessary due to the looping logic with INSTANCE_REQUESTED, but just in case
+                    local instanceArg = playerArg:getInstance()
+                    if instanceArg then
+                        print(fmt('Player {} failed to cleanly transition into instance event, forcing entry via setPos.', playerArg:getName()))
```
zach2good, line 460: Like you mentioned in your comment, `fmt` needs a global exception - you can add it as a commit in the PR if you like

scripts/zones/Alzadaal_Undersea_Ruins/npcs/_20m.lua
```diff
@@ -21,10 +21,11 @@ entity.onEventUpdate = function(player, csid, option, npc)
     -- Force the Nyzul Isle loop to bail out
     if xi.instance.onEventUpdate(player, csid, option, npc) then
         player:setLocalVar('NYZUL_INSTANCE', 1)
-    -- This was causing entry to fail out, and 116 to not be hit.  Handled inside
-    -- Path of Darkness instance script.
-    -- else
-    --     v:updateEvent(405, 3, 3, 3, 3, 3, 3, 3)
+    else
+        -- stops the loop of checking all party member requirements
```
zach2good, line 25: 👍

zach2good (approved):
It's been a long time since I've looked at/rewrote these systems - thanks for giving them some love

Add `fmt` and remove that redundant comment and we can go ahead with merge

PR #5448, opened 2024-04-17
[sql/lua] Delicieuse Delphine NM

MowFord:
**_I affirm:_**

SQL changes are small in this one, bundling together.

https://www.bg-wiki.com/ffxi/Delicieuse_Delphine

NM was already coded to spawn, just updated fight mechanics and stats

Immunities: 39 = 0x27
- paralyze
- bind
- gravity
- sleep

Edit Ochu.lua: `xi.mob.phOnDespawn(mob, ID.mob.DELICIEUSE_DELPHINE_PH, 100, 3600, { immediate = true })`

zone into rolanberry_s, `!gotoid 17150279`, kill ochu, on despawn `!gotoid 17150280`, fight mob

Main thing it has is triple impale moves:

image


scripts/zones/Rolanberry_Fields_[S]/mobs/Delicieuse_Delphine.lua
```diff
@@ -1,9 +1,36 @@
...
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:setMod(xi.mod.DOUBLE_ATTACK, 5)
+    mob:setMod(xi.mod.TRIPLE_ATTACK, 50)
+    mob:addMod(xi.mod.MOVE_SPEED_STACKABLE, 18)
+    mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
+end
+
+entity.onMobWeaponSkillPrepare = function(mob, target)
```
Comment (WinterSolstice8) [inline scripts/zones/Rolanberry_Fields_[S]/mobs/Delicieuse_Delphine.lua:17]: prepare is only called when it readies a TP move on TP, and isn't queued up from useMobAbility?

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Delicieuse_Delphine.lua:17]: Correct, it's only called when the controller decides which skill to use

So `mob:useMobAbility()` would call it, but passing a parameter would not

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #5450, opened 2024-04-17
[lua] Mountain worm nm lua

MowFord:
**_I affirm:_**

Partner to https://github.com/LandSandBoat/server/pull/5449

[Mountain Worm NM](https://www.bg-wiki.com/ffxi/Mountain_Worm_(NM)) is fairly simple

Has the standard thing that quite a few worms should have, where it only casts if out of melee range. Coded it manually since it's an NM (based on the same logic in `mob_controller.cpp`)

If this is approved can just close my sql PR. 

`!spawnmob 16798031`, `!gotoid 16798031`, witness fight

### DIFF
--- a/scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
+++ b/scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
@@ -2,7 +2,7 @@
 -- Area: Uleguerand Range
 --  Mob: Mountain Worm NM
 --  https://www.bg-wiki.com/ffxi/Mountain_Worm_(NM)
--- TODO allow deaggro based on distance (core CMobEntity::CanDeaggro() forces NM and Battlefield mobs to never stop chasing)
+-- TODO allow deaggro (core CMobEntity::CanDeaggro() forces NM and Battlefield mobs to never stop chasing)
 -----------------------------------
 local entity = {}
 
--- a/scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
+++ b/scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
@@ -1,0 +1,23 @@
+-----------------------------------
+-- Area: Uleguerand Range
+--  Mob: Mountain Worm NM
+--  https://www.bg-wiki.com/ffxi/Mountain_Worm_(NM)
+-- TODO allow deaggro based on distance (core CMobEntity::CanDeaggro() forces NM and Battlefield mobs to never stop chasing)
+-----------------------------------
+local entity = {}
+
+entity.onMobSpawn = function(mob)
+    mob:setMod(xi.mod.REGEN, mob:getMaxHP() / 100)
+end
+
+entity.onMobFight = function(mob, target)
+    -- TODO should only cast if out of melee range, but this PR should resolve that https://github.com/LandSandBoat/server/pull/5313
+end
+
+entity.onMobDeath = function(mob, player, optParams)
+end
+
+entity.onMobDespawn = function(mob)
+end
+
+return entity
--- a/scripts/zones/Uleguerand_Range/mobs/Mountain_Worm.lua
+++ b/scripts/zones/Uleguerand_Range/mobs/Mountain_Worm.lua
@@ -1,0 +1,11 @@
+-----------------------------------
+-- Area: Uleguerand Range
+--  Mob: Mountain Worm
+--  Not to be confused with the NM, which has a different mobname in mob_spawn_points
+-----------------------------------
+local entity = {}
+
+entity.onMobDeath = function(mob, player, optParams)
+end
+
+return entity


scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
```diff
@@ -0,0 +1,29 @@
+-----------------------------------
+-- Area: Uleguerand Range
+--  Mob: Mountain Worm NM
+--  https://www.bg-wiki.com/ffxi/Mountain_Worm_(NM)
+-- TODO allow deaggro based on distance (core CMobEntity::CanDeaggro() forces NM and Battlefield mobs to never stop chasing)
+-- TODO confirm if despawns on lost aggro?
```
TeoTwawki: I can help here!

He doesn't depop, but he will very likely move (just standard random worm movement behavior) to a diff part of the cave system (little jerk...) and be really difficult to find again before he full heals.

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
```diff
@@ -0,0 +1,29 @@
...
+-----------------------------------
+local entity = {}
+
+entity.onMobSpawn = function(mob)
+    mob:setMagicCastingEnabled(false)
+    mob:setMod(xi.mod.REGEN, mob:getMaxHP() / 100)
+end
+
+entity.onMobFight = function(mob, target)
+    if mob:checkDistance(target) > mob:getModelSize() + 3 then
```
TeoTwawki: kinda surprised the attempt at checking size wasn't baked into distance checks already (it'd be wrong, but its something i expect the mistake to be made in). 

`getModelSize()` really needs either renamed or how we determine size needs reworked for mobs, because the value stored in `m_ModelRadius` is NOT the physical size of the model on screen: its the "melee range" of the monster stored in `mob_family_system` that currently is not related to its actual size. you can set crazy values there and have a tiny little chigoe that is able to melee hit you and be hit in return from 30 yalms away. And that's the value you are adding 3 to and including here.

The mobs actual size is set in a few bits of a mask that have different purposes depending on which type of entity model it is. In the case of this worm, there are 3 single bits that influence how large it appears in the client. These are currently in the 1st byte of the entityFlags value stored in `mob_pools`

tl;dr You should probably add a comment here about how because he shares a family with the normal mountain worm but is a wee bit larger you had to expand this a bit. At least if I'm correct about that being the case anyway. If I'm not, prolly should drop the +3 then.

TeoTwawki: side thought: we could also bake this into a mob mod or a mixin to avoid writing the same code for future mobs, but I suggest crossing that bridge when we get to another mob we want that behavior on.

sql/mob_spawn_points.sql
```diff
@@ -1101,7 +1101,7 @@ INSERT INTO `mob_spawn_points` VALUES (16798027,'Nightmare_Bats','Nightmare Bats
 INSERT INTO `mob_spawn_points` VALUES (16798028,'Nightmare_Bats','Nightmare Bats',29,-232.594,-36.283,-140.618,135);
 INSERT INTO `mob_spawn_points` VALUES (16798029,'Mountain_Worm','Mountain Worm',42,-267.339,-37.646,-141.238,1);
 INSERT INTO `mob_spawn_points` VALUES (16798030,'Mountain_Worm','Mountain Worm',42,-202.226,-30.116,-179.173,0);
-INSERT INTO `mob_spawn_points` VALUES (16798031,'Mountain_Worm','Mountain Worm',48,-269.816,-79.023,18.257,3);
+INSERT INTO `mob_spawn_points` VALUES (16798031,'Mountain_Worm_NM','Mountain Worm',48,-269.816,-79.023,18.257,3);
```
TeoTwawki, line 1104: 👍

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
MowFord: Ah so he can deaggro (unlike most NM) but doesn't despawn just hard to find again

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
MowFord: Btw I didn't add this because of his size, I did it because that's the exact logic in core to determine if a mob will move or auto attack

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
TeoTwawki: All non battlefield NM on retail can deaggro. It just they have to go a long while without interacting with you, for worms thats easy - be out of range. for other NMs they chase your \*ss down and slap you :)

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
TeoTwawki: so you remade logic related to movement for a mob that _doesn't move._ 🤔 

I know that spot in the core, I've touched it myself. its...less than optimal and fudges on the side of making sure players can actually hit things they are standing in front of despite how fat king behemoths rear end may be. Now I'm thinking again about the mobmod line of thought I was having, maybe that bridge _should_ be crossed sooner rather than later, instead of running a distance check a and mob radius check then attempting to set its casting ability every 240ms.


I think you really wanted "can I hit the player?". I really hope the logic in the **draw in** code that works on that very premise doesn't do the same thing.

_looks_

```cpp
if (currentDistance >= PMob->GetMeleeRange() * 2 && battleutils::DrawIn(PTarget, PMob, PMob->GetMeleeRange() - 0.2f))
```

```cpp
float CBattleEntity::GetMeleeRange() const
{
    return m_ModelRadius + 3.0f;
}
```

_**"Son of a...**_

😭 we really can't win can we. We may as well lua bind `getMeleeRange()` and save some future steps at least, because _some day_ those size bits are going to be correctly used and then melee range will get updated to account for them.

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
MowFord: https://github.com/LandSandBoat/server/blob/f0d23cacf13df0a8cc91f8a5342b3351b26fd86a/src/map/entities/mobentity.cpp#L359

👀

TeoTwawki: 🤦‍♂️

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
MowFord: If we want that binding I can add it, sure

TeoTwawki: > If we want that binding I can add it, sure

yeah lets do that for now theres bound to be other times we want to know melee range in lua

MowFord: Or perhaps I could remove this logic completely and let [this PR](https://github.com/LandSandBoat/server/pull/5313/files) handle it

TeoTwawki: I'd forgotten about that, last I recall it needed to sort out which worms did and didn't do this behavior. But if you wnat to leave the casting behavior to that, that's find by me as well.

MowFord: I'll remove the mob fight logic altogether and watch for that PR to include this NM. Ezpz

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
```diff
@@ -0,0 +1,24 @@
+-----------------------------------
+-- Area: Uleguerand Range
+--  Mob: Mountain Worm NM
+--  https://www.bg-wiki.com/ffxi/Mountain_Worm_(NM)
+-- TODO allow deaggro based on distance (core CMobEntity::CanDeaggro() forces NM and Battlefield mobs to never stop chasing)
+-----------------------------------
+local entity = {}
+
+entity.onMobSpawn = function(mob)
+    mob:setMagicCastingEnabled(false)
```
TeoTwawki: missed a spot when removing the onMobFight part

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
```diff
@@ -0,0 +1,24 @@
+-----------------------------------
+-- Area: Uleguerand Range
+--  Mob: Mountain Worm NM
+--  https://www.bg-wiki.com/ffxi/Mountain_Worm_(NM)
+-- TODO allow deaggro based on distance (core CMobEntity::CanDeaggro() forces NM and Battlefield mobs to never stop chasing)
```
TeoTwawki: Its technically not the distance (I think) but just how long it goes without being able to target/attack you in any way. which of course would mean you are out of casting range.

TeoTwawki (changes requested):

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
MowFord: Oops, thank you

WinterSolstice8 (approved):

scripts/zones/Uleguerand_Range/mobs/Mountain_Worm_NM.lua
MowFord: updated

TeoTwawki (approved):
I amended your TODO comment

PR #5456, opened 2024-04-18
[LUA] Modified Ranged Trust, Rahal, and Tenzen behavior. Added Stealth Shot, Empyreal Arrow (Najelith)

Thayos:
**_I affirm:_**

Modifies Ranged trust behavior as well as Rahal and Tenzen to be close to retail like. Also, added Stealth Shot into abilities and Empyreal Arrow for Najelith.

Elvira requires database mob_pools ID 5941 change for weapon skill type and delay to match her gun delay.

### DIFF
--- a/src/map/utils/fishingutils.cpp
+++ b/src/map/utils/fishingutils.cpp
@@ -1361,17 +1361,24 @@ namespace fishingutils
                 return false;
             }
 
-            if (PChar->hookedFish->successtype != FISHINGSUCCESSTYPE_CATCHITEM)
+            if (PChar->hookedFish == nullptr)
             {
-                if (PBait->getQuantity() == 1)
+                ShowWarning("PChar->hookedFish was null.");
+            }
+            else
+            {
+                if (PChar->hookedFish->successtype != FISHINGSUCCESSTYPE_CATCHITEM)
                 {
-                    charutils::UnequipItem(PChar, SLOT_AMMO, false);
-                }
-                charutils::UpdateItem(PChar, PBait->getLocationID(), PBait->getSlotID(), -1);
+                    if (PBait->getQuantity() == 1)
+                    {
+                        charutils::UnequipItem(PChar, SLOT_AMMO, false);
+                    }
+                    charutils::UpdateItem(PChar, PBait->getLocationID(), PBait->getSlotID(), -1);
 
-                if (SendUpdate)
-                {
-                    PChar->pushPacket<CInventoryFinishPacket>();
+                    if (SendUpdate)
+                    {
+                        PChar->pushPacket<CInventoryFinishPacket>();
+                    }
                 }
             }
         }


scripts/actions/spells/trust/tenzen_ii.lua
```diff
@@ -13,6 +13,16 @@ end
 
 spellObject.onMobSpawn = function(mob)
     xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
+	
+	mob:addSimpleGambit(ai.t.TARGET, ai.c.ALWAYS, 0, ai.r.RATTACK, 0, 0)
+
+    mob:setAutoAttackEnabled(true)
+	
+	mob:addMod(xi.mod.STORETP, 95)
```
TeoTwawki: No. Absolutely no. See #4930 which I took over from Carver because SE pulled BS with his STP gains. it is NOT a single static value to fix it.

TeoTwawki: His Tp per shot changes at exactly the levels that a samurai woudl gain the next store tp Trait, hoever there is no delay value that is consistent with his TP gains when combined with Store TP. Somebody at SE made it up. This ironically penalizes his TP gains before the final tier of STP, because Semih Lefana will get that exact same value at lower levels. Its like they punished him for being a Samurai.

Xaver-DaRed: Comparing both Tenzen II scripts, from both PRs, they look nothing like each other.
yours -> them
Auto attack true -> Auto attack false
Trust distance mid range -> Trust distance long range
TP opener random -> TP opener Highest

I have to ask, did you get the info from retail, Thayos?
Couse some1has to be wrong here

TeoTwawki: I was working off carver's and comparing with my live retail account. H e was mostly behaving right but teh TOP gains per ranged shot were just wrong.

To get His TP gains retail accurate I basically need some kind of override to set a specific amount of TP per ranged attack based on his level: it changes at the exact levels he would gain the store TP trait, but retail testing (BY ME) shows there is no valid delay rating that would result in correct amount of TP when factoring in storeTP. Bets bet would be zero his TP gains and then literally just do a conditional for his level and call `addTP()`

TeoTwawki: He and Semi also have much shorter delays than their TP gains would indicate, btw. Our implementation of Joachim's throw is also wrong, which i think was the basis for carver's Tenzen II script.

tl;dr ranged attacker trusts need some special handling and a total do-over

scripts/actions/spells/trust/tenzen_ii.lua
```diff
@@ -13,6 +13,16 @@ end
 
 spellObject.onMobSpawn = function(mob)
     xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
+	
+	mob:addSimpleGambit(ai.t.TARGET, ai.c.ALWAYS, 0, ai.r.RATTACK, 0, 0)
```
TeoTwawki: this should (usually) be off for ranged-only attackers or else they will additionally do an un-animated melee attack when in range for it.

scripts/actions/spells/trust/tenzen_ii.lua
Thayos: > Comparing both Tenzen II scripts, from both PRs, they look nothing like each other. yours -> them Auto attack true -> Auto attack false Trust distance mid range -> Trust distance long range TP opener random -> TP opener Highest
> 
> I have to ask, did you get the info from retail, Thayos? Couse some1has to be wrong here

I pulled all my information from Bg-wiki trust page. https://www.bg-wiki.com/ffxi/BGWiki:Trusts#Tenzen_II

TeoTwawki: > Only uses his weaponskill to open skillchains when another party member has 1000 TP. Will wait at 3000 TP indefinitely if no other party members gain TP.

TeoTwawki: BG doesn't specify what rules for -who- he opens for, just that somebody has to be > 1000

Thayos: > I was working off carver's and comparing with my live retail account. H e was mostly behaving right but teh TOP gains per ranged shot were just wrong.
> 
> To get His TP gains retail accurate I basically need some kind of override to set a specific amount of TP per ranged attack based on his level: it changes at the exact levels he would gain the store TP trait, but retail testing (BY ME) shows there is no valid delay rating that would result in correct amount of TP when factoring in storeTP. Bets bet would be zero his TP gains and then literally just do a conditional for his level and call `addTP()`

I'll look into what I can do specifically for this without affecting other ranged attackers.

TeoTwawki: sorry for multi-posting:

presumably, as soon as somebody hits 1k he's gonna go immediately, but we can double check that easy enough

TeoTwawki: > I'll look into what I can do specifically for this without affecting other ranged attackers.

if you can manage getting him retail accurate, awesome. if not just don't even try to fix his TP gain for now, leave out STP mods in script he already would be getting traits from job  and SE did something crazy with trust TP gains on ranged attackers in general - none of them match their actual delays as seen in game. they all get far more tp than their delay indicated.

I spent a solid week testing him on retail and various levels and trying to math out how much STP it'd take at his observed delay, along with all the other ranged trusts. every last one.

Thayos: > His Tp per shot changes at exactly the levels that a samurai woudl gain the next store tp Trait, hoever there is no delay value that is consistent with his TP gains when combined with Store TP. Somebody at SE made it up. This ironically penalizes his TP gains before the final tier of STP, because Semih Lefana will get that exact same value at lower levels. Its like they punished him for being a Samurai.

This was an oversight as I didn't test at incremental levels. Will adjust testing practices prior to future PRs.

TeoTwawki: at least Tenzen is the only one with variable TP gains. If he were real he'd probably be opening support tickets asking why he was being nerfed for no reason and wishing he leveled war instead.

Thayos: The short answer would be, make him not samurai but that's not the right answer lol

scripts/actions/spells/trust/elivira.lua
```diff
@@ -13,6 +13,26 @@ end
...
     xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
+	
+		mob:addSimpleGambit(ai.t.SELF, ai.c.NOT_STATUS, xi.effect.BARRAGE,
+                        ai.r.JA, ai.s.SPECIFIC, xi.ja.BARRAGE)
+						
+		mob:addSimpleGambit(ai.t.SELF, ai.c.NOT_STATUS, xi.effect.BERSERK,
+                        ai.r.JA, ai.s.SPECIFIC, xi.ja.BERSERK)
+	
+		mob:addSimpleGambit(ai.t.SELF, ai.c.NOT_STATUS, xi.effect.DOUBLE_SHOT,
+                        ai.r.JA, ai.s.SPECIFIC, xi.ja.DOUBLE_SHOT)
```
zach2good: Not correctly alligned

scripts/actions/spells/trust/elivira.lua
```diff
@@ -13,6 +13,26 @@ end
...
+	
+	mob:addSimpleGambit(ai.t.SELF, ai.c.NOT_STATUS, xi.effect.DECOY_SHOT,
+                        ai.r.JA, ai.s.SPECIFIC, xi.ja.DECOY_SHOT)
+
+	mob:addSimpleGambit(ai.t.TARGET, ai.c.ALWAYS, 0, ai.r.RATTACK, 0, 0)
+
+    mob:setAutoAttackEnabled(true)
+	
+	mob:setMobMod(xi.mobMod.TRUST_DISTANCE, xi.trust.movementType.NO_MOVE)
```
zach2good: So I'm reading this right: She'll spawn, she'll spam RA, she won't move, and if a mob gets close to her she won't reposition and she'll start meleeing? 

What happens on retail? I've never used her I don't think

scripts/actions/spells/trust/makki-chebukki.lua
```diff
@@ -13,7 +13,33 @@ end
 
 spellObject.onMobSpawn = function(mob)
     xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
```
zach2good: This trust has chat lines: https://ffxiclopedia.fandom.com/wiki/Trust:_Makki-Chebukki

scripts/actions/spells/trust/rahal.lua
```diff
@@ -12,7 +12,30 @@ spellObject.onSpellCast = function(caster, target, spell)
 end
 
 spellObject.onMobSpawn = function(mob)
-    xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
+    xi.trust.teamworkMessage(mob, {
```
zach2good: This trust has more chat lines than this: https://ffxiclopedia.fandom.com/wiki/Trust:_Rahal

scripts/actions/spells/trust/tenzen_ii.lua
```diff
@@ -13,6 +13,16 @@ end
 
 spellObject.onMobSpawn = function(mob)
     xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
```
zach2good: This trust has chat lines: https://ffxiclopedia.fandom.com/wiki/Trust:_Tenzen_II

zach2good (commented):
It's been a long time, but you can use this: https://github.com/LandSandBoat/server/blob/base/scripts/globals/trust.lua#L439

To dump a trust's possible messages to see if they have additional synergy voice lines etc. it's a better resource than just using the wikis or exhaustively testing on retail

PR #5271, opened 2024-03-17
[lua] Added pathing for LB2 NMs

Nemisis1987:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

Closes #2850 by adding a patrol path with delays along the paths.

## Steps to test these changes

Zone into Xarcabard and go to LB2 NMs
`!gotoname Boreal_Hound`
Watch pathing of the NMs

PR #5457, opened 2024-04-18
[LUA] Lb2 mob pathing

### LINKED ISSUE #2850: 🐛 Atop the Highest Mountains(level 55 limit break quest)

Window 11

## Branch affected by issue 

`base`

## Steps to reproduce
Start level 55 quest from Maat
go to Xarcabard
Those 3 NM's defending frigicite from ???'s
Boreal hound, Boreal Tiger, and Boreal Coeurl not roaming far enough for me to grab frigicite without engaging.
I remember that in retail I used to use mount to grab item without fighting.

## Expected behavior
Should able to grab frigicite without fighting any NM

MowFord:
**_I affirm:_**

Continues the work of Nemisis in [this PR](https://github.com/LandSandBoat/server/pull/5271). He is still getting accustomed to LSB processes, learning git, and we discussed it would be better to collab and have me make the PR this time.

Closes https://github.com/LandSandBoat/server/issues/2850 .

Packet captures we collected while mounted, then analyzed.

All 3 NMs have this same basic behavior:
- path between a pretty tight area, to well-defined points that, while not perfectly repeating, repeat close enough to say that we should just cycle through them.
- Sometimes they ran between points, and sometimes they walked (seen by his speed in the packet cap)
- When arriving at a pathpoint
  - Sometimes stay for one packet, then move on. 
  - If they stay, they stay for 4-8 seconds while looking around, turning a little at a time

The `pathNodes` were generated by exporting each NM's packet capture to excel, splitting out the position data, and using formula trickery to find when the mobs arrived at a pathNode:

image

Filtered on "eq" and dumped that into a list of pathNodes. Then looked for the trend to repeat and snipped off the rest of the list.

The "look around" mechanic was witnessed as such while a mob's position didn't move:

image

So we came up with the following pathing flow:
- can't use a patrol path with waits, due to
  - `onPathPoint` being called _after_ a mob has waited
  - and because the captures show that waiting at a point is not pre-determined
- Pathing within `pathNodes` could be random, but since the list of points is so small and waiting at a point is random, we chose to simplify the rest of the logic
- When reaching a path node, it's decided if they should wait or not
  - If they wait, a timer is set for 4-8s to have them continue their path
  - while waiting, `rotateMob` is called every 600 ms to rotate a little bit
    - a small chance to not rotate and switch directions is there to match the retail behavior of mostly turning in the same direction, but changing on occasion
    - if we did truly random rotate direction then it would dart back and forth and look unnatural
- we set his base speed higher _sometimes_ when going to the next pathNode
  - `xi.pathflag.RUN` made him clip through things, so we used `xi.pathflag.SLIDE` so they didn't overshoot

Go to xarc, then go and witness each mob:
- `!gotoname Boreal_Tiger`
- `!gotoname Boreal_Coeurl`
- `!gotoname Boreal_Hound`

https://imgur.com/Fh4xc5S

### DIFF
--- a/scripts/zones/Xarcabard/mobs/Boreal_Coeurl.lua
+++ b/scripts/zones/Xarcabard/mobs/Boreal_Coeurl.lua
@@ -8,43 +8,87 @@ local ID = zones[xi.zone.XARCABARD]
 -----------------------------------
 local entity = {}
 
-local pathNodes = 
+local pathNodes =
 {
-    { x = 580.23, y = -9.68,  z = 290.94, }, -- 1 (Origin)
-    { x = 578.95, y = -10.00, z = 285.33, }, -- 2
-    { x = 581.13, y = -10.00, z = 274.86, }, -- 3
-    { x = 578.93, y = -9.52,  z = 268.83, }, -- 4
-    { x = 580.00, y = -10.00, z = 260.17, }, -- 5
-    { x = 578.93, y = -9.52,  z = 268.83, }, -- 4
-    { x = 581.13, y = -10.00, z = 274.86, }, -- 3
-    { x = 578.95, y = -10.00, z = 285.33, }, -- 2
-    { x = 580.23, y = -9.68,  z = 290.94, }, -- 1 (Origin)
+    { x = 578.95, y = -10,   z = 285.33 },
+    { x = 581.13, y = -10,   z = 274.86 },
+    { x = 581.11, y = -9.65, z = 290.31 },
+    { x = 578.95, y = -10,   z = 285.33 },
+    { x = 581.11, y = -9.65, z = 290.31 },
+    { x = 578.93, y = -9.52, z = 268.83 },
+    { x = 581.11, y = -9.65, z = 290.31 },
+    { x = 578.93, y = -9.52, z = 268.83 },
+    { x = 580,    y = -10,   z = 280.17 },
+    { x = 581.13, y = -10,   z = 274.86 },
 }
 
-entity.onMobSpawn = function(mob)
-    -- Failsafe to make sure NPC is down when NM is up
-    if xi.settings.main.OLDSCHOOL_G2 then
-        GetNPCByID(ID.npc.BOREAL_COEURL_QM):showNPC(0)
+local baseSpeed = 60
+local lookDelay = 600
+
+-- recursive if mob is waiting at path point
+local function rotateMob(mob)
+    if mob:getSpeed() == 0 then
+        local rotationDirection = mob:getLocalVar('rotationDirection')
+        local rotationChange = 6
+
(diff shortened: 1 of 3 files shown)


WinterSolstice8:
Base speed of 60? is that from caps?

TeoTwawki:
> Base speed of 60? is that from caps?

SE does some really strange shenanigans whereby the speed is adjusted at every "step" in their movement pathing. (sometimes its "set" to a value it already was even)

Visually to me they look like they are going at the same speed as mobs in our DB we have set to 75, but this is what their in memory values in the client alternate between:

screenshot

ignore "rel" speed, I was experimenting and comparing something.

Atom0s was helping me with an NPC some months ago that had jank values and changing values like this, packet said speed 50, in memory of client said it would be 40 one second and 13.000001 a millisecond after that. Its from How SE's eventing works and from the old compiler they used

> [6:43 PM]atom0s: The client will adjust the speeds as needed when doing the move calculations and lerp junk.
> [6:46 PM]atom0s: If he's also scripted to run an 'event' constantly for any kind of movement related junk that will also affect his speed data potentially if its being updated/set during the event.

...

> 8:01 PM]atom0s: As for why you get the wonky 1.3xxxxxx decimal junk, it's due to the fact that FFXI is compiled with VC6. So the floating point math is different than modern compilers due to modern stuff using newer x86 instructions.

MowFord:
> Base speed of 60? is that from caps?

You can see in my screenshot above that one particular NM was 40 (don't recall which) but the other two were 60 base and I figured it was better to have consistent logic between all three

Also setting their speed to 3x or more caused them to jump around a lot 

Iirc the base 60 mobs had 222 speed when running

TeoTwawki:
The other 2 having higher values is a result of their monster race bonus (hounds = normal, coerurl and tiger = boosted) but in reality none of it matters because the scripted positioning puts them where it does with the same timing.

SEs pathing system is jank, ours is completely unlike theirs but to emulate their jank we make different jank, so no matter what we do its all jank if we want it to look like retail.

Important things I asked on previous PR that got answered here: 
* "why are we randoming?" pr notes gave a good description of what and why
* "sourced from retail capture?" wasn't before, is now
* "have we got all the features or do we need some ToDo's (round about, paraphrased)" didn't before, now it does

I believe on retail SE isn't doing any of it randomly but just has several patterns chained back to back for one very long repeating cycle to make it look random, **but** if we did that we would hit what I call the "node count problem" : if you make a very long set of path nodes, the system lags up HARD and that mobs/npc will _crawl_ in slowmo. If you go back in the history you can see this with Cactrot not-so-Rapido (which still needs new nodes capped btw, he runs through the underground at a few spots).

This is never going to be perfect. It is passable tho.

TeoTwawki (approved):

PR #5458, opened 2024-04-18
[lua] Player avatar BPs give proper miss message and only physical miss from PD

MowFord:
Don't miss when magical abilities are against a PD target

**_I affirm:_**

Player avatars simply do zero dmg when they miss from reasons other than simple accuracy checks. This resolves that with setting the proper miss message.

Also, Magical abilites were missing during PD due to a logic mistake with parenthesis

force a mob to use PD: `!exec player:getTarget():useMobAbility(xi.jsa.PERFECT_DODGE)`
image

and when tiamat goes in the air
image

### DIFF
--- a/scripts/globals/summon.lua
+++ b/scripts/globals/summon.lua
@@ -261,11 +261,18 @@ local attackTypeShields =
 }
 
 xi.summon.avatarFinalAdjustments = function(dmg, mob, skill, target, skilltype, damagetype, shadowbehav)
+    local missMessage = xi.msg.basic.SKILL_MISS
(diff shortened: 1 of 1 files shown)


MowFord:
Also note about the messages above... lol

image

Someone is working on a separate PR to fix that. I suspect it's from the work moving avatar BPs into the petskills sql table

MowFord:
Actually, looking into this it's related to the miss message in use. Petskills need to use the job-ability miss message:

using `xi.msg.basic.SKILL_MISS`
image

using `JA_MISS_2`
image


PR should be ready for review now

Xaver-DaRed (approved):

PR #5459, opened 2024-04-18
Fix logic for determining skill type for damage ratio

TracentEden2:
**_I affirm:_**

This PR fixes an issue with the logic that determines the correct skill for use when calculating the damage ratio. Specifically, the current code has a transpositional error such that `m_attackDirection == PHYSICAL_ATTACK_DIRECTION::RIGHTATTACK` implies `SUB` slot while the `GetWeaponSlot` method implies the opposite. Thus currently on LSB when dual wielding the skill used for each hand is reversed and when with no sub weapon (no dual wield) the default `SKILL_NONE` is used for the main weapon rather than the actual skill.

This PR changes the code of `SetCritical` to directly use `GetWeaponSlot` so the logic is the same and thus the correct skill is used.

A bit difficult to test directly, one way is to add a showdebug statement to print out the skill type in `SetCritical` and then try different weapon types with both dual wielding and without, and notice that the correct skill type is used

### DIFF
--- a/src/map/attack.cpp
+++ b/src/map/attack.cpp
@@ -108,33 +108,23 @@ void CAttack::SetCritical(bool value)
             }
         }
 
-        SKILLTYPE skilltype = SKILLTYPE::SKILL_NONE;
+        SKILLTYPE skilltype  = SKILLTYPE::SKILL_NONE;
+        SLOTTYPE  weaponSlot = static_cast<SLOTTYPE>(GetWeaponSlot());
 
         if (m_attacker->objtype == TYPE_PC)
         {
-            SLOTTYPE slot = SLOT_MAIN;
-
-            if (m_attackDirection == PHYSICAL_ATTACK_DIRECTION::RIGHTATTACK)
+            if (auto* weapon = dynamic_cast<CItemWeapon*>(m_attacker->m_Weapons[weaponSlot]))
             {
-                slot = SLOT_SUB;
+                skilltype = static_cast<SKILLTYPE>(weapon->getSkillType());
             }
-
-            if (m_attacker->objtype == TYPE_PC)
+            else
             {
-                if (auto* weapon = dynamic_cast<CItemWeapon*>(m_attacker->m_Weapons[slot]))
-                {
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):

src/map/attack.cpp
```diff
@@ -108,33 +108,23 @@ void CAttack::SetCritical(bool value)
...
-        SKILLTYPE skilltype = SKILLTYPE::SKILL_NONE;
+        SKILLTYPE skilltype  = SKILLTYPE::SKILL_NONE;
+        SLOTTYPE  weaponSlot = static_cast<SLOTTYPE>(GetWeaponSlot());
 
         if (m_attacker->objtype == TYPE_PC)
         {
-            SLOTTYPE slot = SLOT_MAIN;
-
-            if (m_attackDirection == PHYSICAL_ATTACK_DIRECTION::RIGHTATTACK)
+            if (auto* weapon = dynamic_cast<CItemWeapon*>(m_attacker->m_Weapons[weaponSlot]))
```
zach2good, line 116: weaponSlot is always guaranteed to be sane and within bounds for m_Weapons?

src/map/attack.cpp
```diff
@@ -108,33 +108,23 @@ void CAttack::SetCritical(bool value)
...
             {
-                slot = SLOT_SUB;
+                skilltype = static_cast<SKILLTYPE>(weapon->getSkillType());
```
zach2good, line 118: Why is getSkillType not returning SKILLTYPE anyway?

src/map/attack.cpp
TracentEden2: > weaponSlot is always guaranteed to be sane and within bounds for m_Weapons?

Yeah It can only return either MAIN or SUB in this context, there is a check above for Daken so it will not return AMMO in this context. Overall it can only return one of those three.

src/map/attack.cpp
TracentEden2: > Why is getSkillType not returning SKILLTYPE anyway?

Yeah I am not sure, `GetWeaponSlot` also returns a uint8 for some reason. I can fix the functions in a separate PR if we want?

PR #5464, opened 2024-04-19
[login] Don't close file if file pointer is null

WinterSolstice8:
**_I affirm:_**

Fixes a potential crash on fclose if file is nullptr

Have expired cert, don't crash

### DIFF
--- a/src/login/cert_helpers.h
+++ b/src/login/cert_helpers.h
@@ -37,8 +37,8 @@ namespace certificateHelpers
                 {
                     ShowInfo(fmt::format("Found existing login.key"));
                 }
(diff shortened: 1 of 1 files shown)


src/login/cert_helpers.h
```diff
@@ -37,8 +37,8 @@ namespace certificateHelpers
                 {
                     ShowInfo(fmt::format("Found existing login.key"));
                 }
+                fclose(fileHandle);
```
zach2good, line 40: Use `utils::openFile` to open a file and get back a `unique_ptr` that'll close itself when it falls out of scope

PR #5468, opened 2024-04-19
[lua] Pet magic burst message and accbonus

MowFord:
**_I affirm:_**

Implements "Magic Burst!" messages for smn petskills

The `JA_MAGIC_BURST` message is only valid for petskill-type actions, so as the rest of the smn abilities get moved over they will all get the message.

example of MB on the primary target
image

example of MB on secondary target
image

Nether blast message unchanged as it doesn't use the petskill action type yet (Note that it still gets bonus acc/damage)
image

Confirmed wyvern breaths already have the MB message, so the final thing would be jug pets for bst. None of these are moved to petskills as far as I know, but as they do this logic will catch them due to applying when `petID() >0` and `actionType== xi.action.PET_MOBABILITY_FINISH`

Bst Jug pets currently use mobskill lua files, so they (correctly) do not get the call to `xi.mobskills.mobAddBonuses`, but when they get reworked their lua should mimic avatar blood pact lua files and it should "just work"

Confirmation that bst jugs should magic burst from [this retail update](https://forum.square-enix.com/ffxi/threads/46068-Feb-19-2015-(JST)-Version-Update)
image

Note: this also fixes the default damage message for petskills, as they should be using `xi.msg.basic.USES_JA_TAKE_DAMAGE` instead of the mobskill/weaponskill `xi.msg.basic.DAMAGE` message id.

create skillchains and use various abilities to see you get magic burst if the elements match.

### DIFF
--- a/scripts/globals/mobskills.lua
+++ b/scripts/globals/mobskills.lua
@@ -288,15 +288,23 @@ xi.mobskills.mobMagicalMove = function(mob, target, skill, damage, element, dmgm
     local finaldmg = damage * mab * dmgmod
 
     -- get resistance
-    local avatarAccBonus = 0
+    local petAccBonus = 0
     if mob:isPet() and mob:getMaster() ~= nil then
         local master = mob:getMaster()
         if mob:isAvatar() then
-            avatarAccBonus = utils.clamp(master:getSkillLevel(xi.skill.SUMMONING_MAGIC) - master:getMaxSkillLevel(mob:getMainLvl(), xi.job.SMN, xi.skill.SUMMONING_MAGIC), 0, 200)
+            petAccBonus = utils.clamp(master:getSkillLevel(xi.skill.SUMMONING_MAGIC) - master:getMaxSkillLevel(mob:getMainLvl(), xi.job.SMN, xi.skill.SUMMONING_MAGIC), 0, 200)
+        end
+
+        local skillchainTier, _ = xi.magicburst.formMagicBurst(element, target)
+        if
+            mob:getPetID() > 0 and
+            skillchainTier > 0
+        then
+            petAccBonus = petAccBonus + 30
         end
     end
 
-    local resist       = xi.mobskills.applyPlayerResistance(mob, nil, target, mob:getStat(xi.mod.INT)-target:getStat(xi.mod.INT), avatarAccBonus, element)
+    local resist       = xi.mobskills.applyPlayerResistance(mob, nil, target, mob:getStat(xi.mod.INT)-target:getStat(xi.mod.INT), petAccBonus, element)
     local magicDefense = getElementalDamageReduction(target, element)
 
     finaldmg       = finaldmg * resist * magicDefense
@@ -337,7 +345,7 @@ xi.mobskills.applyPlayerResistance = function(mob, effect, target, diff, bonus, 
     return getMagicResist(p)
 end
 
-xi.mobskills.mobAddBonuses = function(caster, target, dmg, ele) -- used for SMN magical bloodpacts, despite the name.
+xi.mobskills.mobAddBonuses = function(caster, target, dmg, ele, skill) -- used for SMN magical bloodpacts, despite the name.
     local magicDefense = getElementalDamageReduction(target, ele)
     dmg = math.floor(dmg * magicDefense)
 
@@ -375,6 +383,14 @@ xi.mobskills.mobAddBonuses = function(caster, target, dmg, ele) -- used for SMN 
     dmg             = math.floor(dmg * dayWeatherBonus)
 
     local burst = calculateMobMagicBurst(caster, ele, target)
+    if
+        skill and
+        burst > 1.0 and
+        caster:getPetID() > 0 -- all pets except charmed pets can get magic burst message, but only with petskill action
+    then
+        skill:setMsg(xi.msg.basic.JA_MAGIC_BURST)
+    end
+
     dmg         = math.floor(dmg * burst)
 
     local mdefBarBonus = 0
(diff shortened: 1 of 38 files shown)


scripts/globals/mobskills.lua
```diff
@@ -375,6 +383,15 @@ xi.mobskills.mobAddBonuses = function(caster, target, dmg, ele) -- used for SMN
     dmg             = math.floor(dmg * dayWeatherBonus)
 
     local burst = calculateMobMagicBurst(caster, ele, target)
+    if
+        burst > 1.0 and
+        caster:getPetID() >= 0 -- all pets except charmed pets can get magic burst message, but only with petskill action
```
WinterSolstice8: `>=`?

MowFord: Multiplier is default 1

Oh you mean pet id, whoops

MowFord: Fixed, nice catch

scripts/globals/mobskills.lua
```diff
@@ -288,15 +288,23 @@ xi.mobskills.mobMagicalMove = function(mob, target, skill, damage, element, dmgm
...
-            avatarAccBonus = utils.clamp(master:getSkillLevel(xi.skill.SUMMONING_MAGIC) - master:getMaxSkillLevel(mob:getMainLvl(), xi.job.SMN, xi.skill.SUMMONING_MAGIC), 0, 200)
+            petAccBonus = utils.clamp(master:getSkillLevel(xi.skill.SUMMONING_MAGIC) - master:getMaxSkillLevel(mob:getMainLvl(), xi.job.SMN, xi.skill.SUMMONING_MAGIC), 0, 200)
+        end
+
+        local skillchainTier, _ = xi.magicburst.formMagicBurst(element, target)
+        if
+            mob:getPetID() > 0 and
+            skillchainTier > 0
+        then
+            petAccBonus = petAccBonus + 25
```
WinterSolstice8, line 303: Where does this 25 bonus macc come from? Any source?

scripts/globals/mobskills.lua
```diff
@@ -375,6 +383,15 @@ xi.mobskills.mobAddBonuses = function(caster, target, dmg, ele) -- used for SMN
     dmg             = math.floor(dmg * dayWeatherBonus)
 
     local burst = calculateMobMagicBurst(caster, ele, target)
+    if
+        burst > 1.0 and
+        caster:getPetID() > 0 -- all pets except charmed pets can get magic burst message, but only with petskill action
+    then
+        target:setLocalVar('abilityMagicBurst', 1)
```
WinterSolstice8: Not a fan of using local vars unless absolutely necessary. the two kinds of `skill` we feed in could have something like... `isMagicBurst` and `setMagicBurst`, what do you think?

scripts/globals/mobskills.lua
MowFord: Pulled this from an xi.magic function. I assumed it was a global bonus during a magic burst

WinterSolstice8: yeah Jimmy seems to think it should be 30? he's looking into it so standby

scripts/globals/mobskills.lua
MowFord: This function does not pass in the skill. If we pass the skill we could set the message directly.

I assumed deciding the message in a single place (final adjustments function) was better

WinterSolstice8: I agree that setting the message in final adjustments makes sense, but I think passing in `skill` here and setting the flag to be read later would be ideal

MowFord: I have an idea of what to do to keep the simplicity this PR currently has. Will work on it later today

scripts/globals/mobskills.lua
MowFord: I've updated to +30, lmk if it ends up being something else

scripts/globals/mobskills.lua
MowFord: Updated all petskill lua to pass the respective skill object and the magic burst message is set if the skill is there

then in the "final adjustment" functions, I check the message only if it's a pet_finish action.

Spot checked that it still works as expected

MowFord: Touching all these files made me notice that Ifrit's abilities are getting bonus dmg from skillchains:

image

I assume this isn't correct

Xaver-DaRed:
Needs rebase

scripts/globals/mobskills.lua
WinterSolstice8: Looks like it's 25, there's some ancient blog that says that

scripts/globals/mobskills.lua
WinterSolstice8: I don't recall this being a thing, yeah

WinterSolstice8: I think we can leave that alone for now and fix it later unless it's egregious levels of extra damage?

scripts/globals/summon.lua
```diff
@@ -278,7 +278,15 @@ xi.summon.avatarFinalAdjustments = function(dmg, mob, skill, target, skilltype,
 
     -- set message to damage
     -- this is for AoE because its only set once
-    skill:setMsg(xi.msg.basic.DAMAGE)
+    if
+        mob:getCurrentAction() == xi.action.PET_MOBABILITY_FINISH
+    then
```
Xaver-DaRed: Theres only one condition here. It can be one line

scripts/globals/mobskills.lua
```diff
@@ -485,7 +501,15 @@ xi.mobskills.mobFinalAdjustments = function(dmg, mob, skill, target, attackType,
 
     -- set message to damage
     -- this is for AoE because its only set once
-    skill:setMsg(xi.msg.basic.DAMAGE)
+    if
+        mob:getCurrentAction() == xi.action.PET_MOBABILITY_FINISH
+    then
```
Xaver-DaRed: This is also only one condition. Could be just one line.

Xaver-DaRed (approved):
Approving anyway. This file needs a cleanup anyway.

scripts/globals/mobskills.lua
MowFord: snip snap. Updated back to +25. Thanks for verifying

scripts/globals/summon.lua
MowFord: Done

scripts/globals/mobskills.lua
MowFord: Done

WinterSolstice8 (approved):

PR #5469, opened 2024-04-20
[LUA] Fix BLU physical magic displays misses as 0 dmg

ampitere:
**_I affirm:_**

Currently when BLU use physical based magic and it misses it will display as doing 0 damage to a mob. Instead we'll display that the magic had no effect on the target to better indicate that the spell missed.

1. !changejob 16 99
2. !addallspells
3. Set some physical spells in town
4. Use a physical BLU spell on a mob

### DIFF
--- a/scripts/globals/bluemagic.lua
+++ b/scripts/globals/bluemagic.lua
@@ -285,6 +285,10 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
         hitsdone = hitsdone + 1
     end
 
+    if finaldmg <= 0 then
+        spell:setMsg(xi.msg.basic.MAGIC_NO_EFFECT)
+    end
+
     return xi.spells.blue.applySpellDamage(caster, target, spell, finaldmg, params, trickAttackTarget)
 end
 


scripts/globals/bluemagic.lua
```diff
@@ -285,6 +285,10 @@ xi.spells.blue.usePhysicalSpell = function(caster, target, spell, params)
         hitsdone = hitsdone + 1
     end
 
+    if finaldmg <= 0 then
```
WinterSolstice8, line 288: Too simple, stoneskin and phalanx can make stuff do 0 dmg without "no effect"

ampitere: Stoneskin and Phalanx modifiers are applied after this by applySpellDamage so if the value calculated here is 0 it will be sent to applySpellDamage as 0. When the spell successfully hits and does damage it will display a 0 damage message correctly since the spell damage was calculated higher than 0 but applySpellDamage then makes the Phalanx/Stoneskin adjustments.

I tested on a mob with Stoneskin and it did properly display both the no effect message as well as the 0 damage message. Please let me know if there's something I might be missing.

WinterSolstice8: great, that's all I needed to know

WinterSolstice8 (approved):

PR #5471, opened 2024-04-20
[core] - Fixed a bug where player couldn't heal escort npc's

sarca571ca:
**_I affirm:_**

#### Fixes an issue when a player healed an escort npc. 
The npc would take the heal as an attack and try to attack the player. Since the npc is allied to the player this would lock the npc in place unable to take any actions.

!zone Grand Palace of Hu'Xzoi
!gotoid 16916928
Select the Cermet Alcove to spawn the Quasilumin.
Cast Cure on the Quasilumin and see that it continues as normal instead of freezing in place.

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -4832,7 +4832,9 @@ namespace battleutils
     {
         TracyZoneScoped;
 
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):
Seems reasonable

sarca571ca:
Sigh formatting really wants the comments line up with each other.

sarca571ca:
Yea I've got no idea why this is failing now.

PR #5475, opened 2024-04-21
Core: Change how entities are loaded for instances

zach2good:
**_I affirm:_**

`instance_entities.sql` is not shift-proof, and apparently a few things have broken at some point due to shifts in the instance entities.

All that `instance_entities` does is say "hey, you need to load this specific entity when you create whatever instance". We can achieve this same thing by scraping entries from `ID.lua`, or the relevant instance's Lua file. I'm probably going to add the entities to the `ID.lua` file, and then have a section in the instance Luas that lists all the relevant `ID` entries for loading. 

This will make all instances as shift proof as anything else we have, and make it a little more obvious how entities are declared for use with instances - pushing more things into the individual instance files.


zach2good:
Will come back to this from scratch at some point - probably when I rewrite instances again

PR #5477, opened 2024-04-21
[lua] Minuet potency

MowFord:
**_I affirm:_**

Was noticed that minuet potency was not scaling well with combined skill. At 500 combined skill minuet III was still not max potency. Found 3 things:
- the table has a column for "Smax" to convert to `skillMax` variable, but this is a misnomer, as it's really the minimum skill required to get potency increases past the base power.
  - Renamed appropriately
- Ranged skill was not properly limiting on skill type, meaning a rng/brd would have high combined skill when wielding a gun, etc
  - only add ranged item skill if of proper skill type
- Minuet `divisor` was not a sane value based on multiple sources
  - Minuet III at 500 combined skill: 74
  - Minuet IV at 500 combined skill: 64
  - using divisor 4, you get 96 (max potency) and 81, respectively
  - This is the most complex part, so i'll continue below

Original divisor of 8, 6... results in the following skill caps for minuet tiers:
image

Doing some math results in the following exact divisors to match bgwiki data (they do not list the skill cap for minuet II):
image

but using divisor of 4 across the board seems to do a reasonable fit
image

This likely matches with reality (1 extra potency per 4 skill past "minimum required skill per tier"), and definitely sits better than the current values.

Getting the actual values here (for all brd enhancement songs) will be a very tedious process, I hope this can be taken as "good enough" until someone does the research from retail.

Edit: I've adjusted the potency values to only the ones explicitly defined in bgwiki: 1 - 166, 3 - 435, 5 - 874

Apply singing skill to test various potency values at different tiers of minuet:
- `!setmod singing 100` to give extra 100 singing skill, but you must change gear for the "magic skill" menu to reflect the change
- then cast a song
- and extract the power into a map server print: `!exec print(player:getStatusEffect(xi.effect.MINUET):getPower())`

### DIFF
--- a/scripts/globals/spells/enhancing_song.lua
+++ b/scripts/globals/spells/enhancing_song.lua
@@ -109,12 +109,22 @@ xi.spells.enhancing.calculateSongPower = function(caster, target, spell, spellId
     local power       = pTable[spellId][7] -- The variable we want to calculate.
     local meritEffect = pTable[spellId][5]
     local jpEffect    = pTable[spellId][6]
-    local skillReq    = pTable[spellId][8]
+    local skillNeeded = pTable[spellId][8]
     local potencyCap  = pTable[spellId][9]
     local multiplier  = pTable[spellId][10]
     local divisor     = pTable[spellId][11]
+    local singingLvl  = caster:getSkillLevel(xi.skill.SINGING)
+    local rangedLvl   = caster:getWeaponSkillLevel(xi.slot.RANGED)
 
-    local singingLvl  = caster:getSkillLevel(xi.skill.SINGING) + caster:getWeaponSkillLevel(xi.slot.RANGED)
+    -- Add ranged skill level ONLY if it's an instrument.
+    local rangeType = caster:getWeaponSkillType(xi.slot.RANGED)
+
+    if
+        rangeType == xi.skill.STRING_INSTRUMENT or
+        rangeType == xi.skill.WIND_INSTRUMENT
+    then
+        singingLvl = singingLvl + rangedLvl
+    end
 
     -- Get Potency bonuses from Singing Skill and Instrument Skill. TODO: Investigate JP-Wiki. Most of this makes no sense.
     -- NOTE: Tier 1 Etudes.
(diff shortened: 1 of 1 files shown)


scripts/globals/spells/enhancing_song.lua
```diff
@@ -109,12 +109,19 @@ xi.spells.enhancing.calculateSongPower = function(caster, target, spell, spellId
     local power       = pTable[spellId][7] -- The variable we want to calculate.
     local meritEffect = pTable[spellId][5]
     local jpEffect    = pTable[spellId][6]
-    local skillCap    = pTable[spellId][8]
+    local skillReq    = pTable[spellId][8]
```
Xaver-DaRed: Please change to `skillRequirement` or `skillNeeded`

scripts/globals/spells/enhancing_song.lua
```diff
@@ -109,12 +109,19 @@ xi.spells.enhancing.calculateSongPower = function(caster, target, spell, spellId
...
-    local singingLvl  = caster:getSkillLevel(xi.skill.SINGING) + caster:getWeaponSkillLevel(xi.slot.RANGED)
+    local rangeSkill  = caster:getWeaponSkillLevel(xi.slot.RANGED)
+    local rangeType   = caster:getWeaponSkillType(xi.slot.RANGED)
+    local singingLvl  = caster:getSkillLevel(xi.skill.SINGING)
+    if
+        rangeType == xi.skill.STRING_INSTRUMENT or
+        rangeType == xi.skill.WIND_INSTRUMENT
+    then
+        singingLvl = singingLvl + rangeSkill
+    end
```
Xaver-DaRed, line 127:
```lua
    local singingLvl = caster:getSkillLevel(xi.skill.SINGING)
    local rangedLvl  = caster:getWeaponSkillLevel(xi.slot.RANGED)

    -- Add ranged skill level ONLY if it's an instrument.
    local rangeType = caster:getWeaponSkillType(xi.slot.RANGED)

    if
        rangeType == xi.skill.STRING_INSTRUMENT or
        rangeType == xi.skill.WIND_INSTRUMENT
    then
        singingLvl = singingLvl + rangedLvl
    end
```

Xaver-DaRed (approved):

EpicTaru (approved):

PR #5479, opened 2024-04-21
[Fix] Correct Iridescence bonus based on staff quality.

Xaver-DaRed:
**_I affirm:_**

Makes Iridescence give the apropiate bonus (5% * staff quality), instead of a flat 10% no matter what.

Cast, preferably with an obi that forces weather bonuses, and an iridial or Chatoyant staff.
See potency actually vary.

### DIFF
--- a/scripts/globals/magic.lua
+++ b/scripts/globals/magic.lua
@@ -390,47 +390,38 @@ function getCureFinal(caster, spell, basecure, minCure, isBlueMagic)
     end
 
     local dayWeatherBonus = 1
-    local ele = spell:getElement()
+    local spellElement    = spell:getElement()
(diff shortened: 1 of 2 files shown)


scripts/globals/magic.lua
```diff
@@ -439,7 +428,13 @@ function getCureFinal(caster, spell, basecure, minCure, isBlueMagic)
         dayWeatherBonus = 1.4
     end
 
-    local final = math.floor(math.floor(math.floor(math.floor(basecure) * potency) * dayWeatherBonus) * rapture) * dSeal
```
WinterSolstice8, line 442: much better

WinterSolstice8 (approved):

PR #5480, opened 2024-04-21
[core] Use recast container for mob spells as well

WinterSolstice8:
**_I affirm:_**

Enables use of spell recasts for mobs like they should

No slow 2:
image
Slow 2:
image

Slow/Elegy/Addle mobs and see them take longer to cast spells

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1001,8 +1001,7 @@ bool CCharEntity::ValidTarget(CBattleEntity* PInitiator, uint16 targetFlags)
 bool CCharEntity::CanUseSpell(CSpell* PSpell)
 {
     TracyZoneScoped;
-    return charutils::hasSpell(this, static_cast<uint16>(PSpell->getID())) && CBattleEntity::CanUseSpell(PSpell) &&
-           !PRecastContainer->Has(RECAST_MAGIC, static_cast<uint16>(PSpell->getID()));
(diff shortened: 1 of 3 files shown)


src/map/entities/charentity.cpp
```diff
@@ -216,7 +216,6 @@ CCharEntity::CCharEntity()
 
     PAutomaton             = nullptr;
     PClaimedMob            = nullptr;
-    PRecastContainer       = std::make_unique<CCharRecastContainer>(this);
```
WinterSolstice8: This was removed because CBattleEntity is already doing this.

TeoTwawki (approved):

src/map/entities/charentity.cpp
cocosolos: Correct me if I'm wrong, but battle entity uses the generic recast container whereas the char recast container writes recasts to the DB.

WinterSolstice8: ah, you're right, I didn't notice it was a different class

TeoTwawki: > Correct me if I'm wrong, but battle entity uses the generic recast container whereas the char recast container writes recasts to the DB.

I knew we were for item and ability, I hadn't thought we do that for spells too. you might be right.

WinterSolstice8: Fixed

TeoTwawki (approved):

Issue #3627, opened 2023-02-25
🔨 Latent Effects that grant Pet mods


DiscipleOfEris:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature

Some items have latent effects that grant pet mods. For example, [Affinity Earring](https://ffxiclopedia.fandom.com/wiki/Affinity_Earring) which provides +Attack to your pet if a party member (not including the wearer) is a Puppetmaster. Or [Fidelity Earring](https://ffxiclopedia.fandom.com/wiki/Fidelity_Earring) which provides +pet Accuracy if there is a Beastmaster in the party.

To my knowledge, there is no simple way to grant these effects with the current codebase.

Ideally, there wouldn't really be a big difference between granting a pet mod and granting a standard mod, but the codebase already separates `item_mods` and `item_mods_pet`. That said, I think adding two columns to `item_latents` for `isPetMod` and `petType` would be preferable to adding a new table for `item_latents_pet`.

TeoTwawki:
there was a past discussion for this https://github.com/LandSandBoat/server/discussions/2672#discussioncomment-3588049 you should take a look at.

TeoTwawki:
the short of it is if going extra columns instead of extra tables, I'd want to see the same with mods. I don't like the inconsistency of doing things 2 ways, and thats why the suggested table was the shortest point a to point b for that problem.

DiscipleOfEris:
Ah, I searched issues but not Discussions. My mistake.

I actually think item pet mods should also be in the same table as item mods, but obviously that's work someone has to do.

TeoTwawki:
> Ah, I searched issues but not Discussions. My mistake.
> 
> I actually think item pet mods should also be in the same table as item mods, but obviously that's work someone has to do.

I kinda agree with you, but its a larger effort to do one table for each of mods and latents, rather than 4 tables.

MowFord:
This was resolved by the linked PR

PR #5487, opened 2024-04-22
[sql/lua] Avatar conditional mods from gear

MowFord:
**_I affirm:_**

The first commit is cleanup of a few simple latents based on a particular avatar being out
- [Carbie Cap +1](https://www.bg-wiki.com/ffxi/Carbie_Cap_%2B1)
- [Diabolos's Rope](https://www.bg-wiki.com/ffxi/Diabolos%27s_Rope)
- [Accord Hat](https://www.bg-wiki.com/ffxi/Accord_Hat)

In auditing these pet-specific items, I ran across the issue mentioned in https://github.com/LandSandBoat/server/issues/3627 , this PR attempts to make a method of applying pet-based mods via arbitrary conditionals using the item lua file

A working example of [Fervor Ring](https://www.bg-wiki.com/ffxi/Fervor_Ring) is in the 2nd commit

Remaining items would be:
- [Evoker's Gages](https://www.bg-wiki.com/ffxi/Evoker%27s_Gages)
- [Karura Hachigane](https://www.bg-wiki.com/ffxi/Karura_Hachigane)
- [Morana Pigaches](https://www.bg-wiki.com/ffxi/Morana_Pigaches)
- [Fenrir's Crown](https://www.bg-wiki.com/ffxi/Fenrir%27s_Crown)
- [Affinity](https://www.bg-wiki.com/ffxi/Affinity_Earring)/[Fidelity](https://www.bg-wiki.com/ffxi/Fidelity_Earring) Earring
  - I imagine the function in `xi.itemUtils` would not be needed here, since it applies to all pets. Just need a simple `player:addPetMod(` and `player:delPetMod(` when the conditions are met/lost. The trick would be in determining the conditions. a `TICK` listener with a localvar could definitely do it
  - These would be a little more complex, and I think out of scope for this PR

There are very few pieces of gear with these restrictive conditions and this lua method works. Gauging responses before I code the rest of the items

### DIFF
--- a/scripts/globals/item_utils.lua
+++ b/scripts/globals/item_utils.lua
@@ -265,3 +265,26 @@ xi.itemUtils.removeMultipleEffects = function(target, effects, count, random)
         return removed
     end
 end
+
+-- for applying pet mods based on arbitrary conditions
+-- I.E. avatar attack only for a particular pet
+-- example usage in fervor_ring.lua
+-- can pass nil for petId to apply to all pets
+xi.itemUtils.handlePetLatentMods = function(owner, petId, mods, give)
+    local pet = owner:getPet()
+
+    if
+        pet and
+        (not petId or pet:getPetID() == petId)
+    then
+        if give then
+            for _, mod in ipairs(mods) do
+                pet:addMod(mod[1], mod[2])
+            end
+        else
+            for _, mod in ipairs(mods) do
+                pet:delMod(mod[1], mod[2])
+            end
+        end
+    end
+end
--- a/scripts/items/fervor_ring.lua
+++ b/scripts/items/fervor_ring.lua
@@ -1,0 +1,36 @@
+-----------------------------------
+-- ID: 11675
+-- Fervor Ring
+-- Pet mod via latent effect
+-----------------------------------
+local itemObject = {}
+local listenerName = 'PET_MOD_LATENT_11675'
+local latentPetId = xi.petId.IFRIT
+local latentMods =
+{
+    { xi.mod.MATT, 2 },
+}
+
+itemObject.onItemCheck = function(target)
+    return 0
+end
+
+itemObject.onItemEquip  = function(user, item)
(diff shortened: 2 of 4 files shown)


WinterSolstice8 (approved):
I don't really see anything wrong here

Xaver-DaRed (approved):
While this method doesnt really convince me, I dont think its worth adding a whole new column to a table for ... 9? items in the whole game that use an specific pet ID, as oposed to a pet type

TeoTwawki:
This is fine for now. For years DSP had all the pet stuff in onEquip in individual items until `item_mods_pet.sql was born`


> While this method doesnt really convince me, I dont think its worth adding a whole new column to a table for ... 9? items in the whole game that use an specific pet ID, as oposed to a pet type

specific pet ID could easily be a whole new latent ID -if- we actually had a sane way to apply latents. Spoiler, we don't. The earlier discussion was that the code that lived since dsp didn't have a consistent way of applying any latents for them _**at all**_ : when they did redid pet modifiers they opted to use a whole separate table and delete all the existing pet modifier IDs (those applied to player and then the core did things to pets base don what was on the player, the change applied any modifier directly to pets so then there was new need to have both a player str mod and a pet str mod both on the player). We have 2 tables for mods, but only 1 for latents : no pet specific latent system was ever made. I would not want to see that done entirely in lua and set in stone. (edit to add exception: if the entire latent system for all entities became lua, as long as it performed and was maintaiable wouldn't care)

"a whole new column" is a not a big deal (this is not a wide table, 1 column isn't tanking anyone performance, it has literally no negatives), but I wouldn't want to do that when thats not what we do for player stuff either and I like consistency. That makes `item_latents_pet.sql` my ideal long term solution. Things could easily go the other way, and we wouldn't need 2 mod tables. But then somebody would be making a lot of new rows in the mod table to combine all the existing items to it.

Its also not just 9 items. its 9 items _and every future pet latent and every pet latent not specific to a singular pet id_

th-2392184863

TeoTwawki (approved):

PR #5488, opened 2024-04-22
[lua] Fix incorrect zone ID table

MowFord:
**_I affirm:_**

My previous PR to clean up the sch af indescript markings left this typo to the table:

image

Apologies for that. It was the last one I fixed and obviously didn't test after putting that final restriction due to there being a second Markings in the zone


Xaver-DaRed (approved):
Shit happens

PR #5492, opened 2024-04-23
[lua] Chi blast is now breath damage

WinterSolstice8:
**_I affirm:_**

Chi blast is now breath damage
Chi blast also now adjusts for Phalanx

https://www.bg-wiki.com/ffxi/Chi_Blast (spirit damage aka breath damage)
https://wiki.ffo.jp/html/3268.html
image

Use chi blast against things with BDT, see them take reduced damage

### DIFF
--- a/scripts/globals/ability.lua
+++ b/scripts/globals/ability.lua
@@ -4,7 +4,7 @@
 xi = xi or {}
 xi.ability = xi.ability or {}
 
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):
omg, that xi.ability function sucks.

Out of scope of this PR, approving.

PR #5499, opened 2024-04-24
[sql/cpp] Blu traits

MowFord:
**_I affirm:_**

- A handful of Blue Traits were giving the wrong mod
  - trait 11 (rapid shot) should be modid 359
  - trait id 2 (evasion bonus) should be modid 68
  - trait id 110 (mag. burst bonus) should be modid 487
  - trait id 20 (gilfinder) should be modid 897 with power 50
- Blue Traits with multiple mods (ratt/att) and (racc/acc) were only applying one of the mods
  - This PR changes it to loop over all eligible traits and find the highest power for each traitid/modid before adding to the player
- The logic didn't explicitly check for this, but somehow worked reliably: Blue Trait categories can have higher power based on more set points
- There was some extraneous logic since this function is only ever called when rebuilding the character traits from scratch
  
New logic 
  - gets the total points per category
  - loops over all eligible blue traits based on category and total points matching
  - loops over all job-related traits (remember no blue traits are added yet) and skips if the player already has the trait id
  - if it's determined that this blue trait is eligible
    - it checks if the same traitid/modid is staged to be added
    - if the new trait is higher power, it replaces the other
    - if the new trait is lower power it skips the new trait

To test getting all the mods for a Trait Category
 - Set spells Uppercut and Battle Dance to get "Attack Bonus" trait (lists once in the menu), which adds RATT and ATT

To test getting the highest tier of a Trait
 - Set spells from [this page](https://www.bg-wiki.com/ffxi/Clear_Mind) to get increasing power of clear_mind as your set points total more


sql/blue_traits.sql
```diff
@@ -53,6 +53,6 @@ INSERT INTO `blue_traits` VALUES (25,2,18,259,10);
 INSERT INTO `blue_traits` VALUES (25,4,18,259,15);
 INSERT INTO `blue_traits` VALUES (25,6,18,259,25);
 INSERT INTO `blue_traits` VALUES (26,2,70,306,15);
-INSERT INTO `blue_traits` VALUES (27,2,110,407,5);
-INSERT INTO `blue_traits` VALUES (28,2,20,0,0);
```
WinterSolstice8, line 57: So strange they'd just insert zeroes without a comment...

src/map/utils/blueutils.cpp
```diff
@@ -438,57 +445,70 @@ namespace blueutils
...
+                        // Check if any existing player Traits conflict
                         for (std::size_t j = 0; j < PChar->TraitList.size(); ++j)
                         {
                             CTrait* PExistingTrait = PChar->TraitList.at(j);
 
                             if (PExistingTrait->getID() == PTrait->getID())
                             {
-                                if (PExistingTrait->getLevel() == 0 && ((CBlueTrait*)PExistingTrait)->getCategory() == PTrait->getCategory())
+                                // Player has the real job trait, making them ineligible
+                                if (PExistingTrait->getLevel() > 0)
```
WinterSolstice8, line 462: hmm. BLU is capable of getting traits above the strength of main job traits, and easily able to get above the strength of subs. Is this true of job mastered BLU? I can't say I know if it's true or not, however.

It was my understanding people set AoE BLU spells on retail that give them more MAB than /RDM does, so this could be pretty easily testable.

WinterSolstice8: Will approve anyway because this is a good improvement, and a bit beyond the scope of this PR

WinterSolstice8 (approved):

src/map/utils/blueutils.cpp
MowFord: Maybe it only overwrites if the blue trait is stronger? News to me that Blu traits would apply if the sj gave the trait

WinterSolstice8: https://www.bg-wiki.com/ffxi/Blue_Mage_Job_Traits -> `In the event that a trait is granted through both spells and one's support job, whichever grants the higher level of the trait will take priority.`

sql/blue_traits.sql
Xaver-DaRed: Back in the dark ages, gilfinder wasnt even coded

So its a placeholder entry that never got updated

Xaver-DaRed (approved):

src/map/utils/blueutils.cpp
MowFord: Updated some comments and added a TODO

PR #5501, opened 2024-04-24
[core] Don't load npc script if the npc name is unprintable

WinterSolstice8:
**_I affirm:_**

Skip loading NPCs that have a nonprintable (and therefore invalid filepath) name

Check the evil box here, don't crash on map load
image

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -807,6 +807,28 @@ namespace luautils
         std::string filename;
         if (PEntity->objtype == TYPE_NPC)
         {
+            // clang-format off
+            auto isNamePrintable = [](const std::string& name)
+            {
+                // Match non-printable ASCII
+                for (const char& c : name)
+                {
+                    if ((c >= 0 && c <= 0x20) || c >= 0x7F)
+                    {
+                        return false;
+                    }
+                }
+                return true;
+            };
+            // clang-format on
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

zach2good:
I jiggled around the top of `CacheLuaObjectFromFile` for a similar effect to this. My concern is why this only explodes on non-EN systems, or if you've enabled the `"Beta: Use Unicode UTF-8 for worldwide language support"` flag, and all of our other systems go through this section fine with the binary-garbage NPC names...

```cpp
    void CacheLuaObjectFromFile(std::string const& filename, bool overwriteCurrentEntry /* = false*/)
    {
        TracyZoneScoped;
        TracyZoneString(filename);

        std::filesystem::path path;
        try
        {
            path = std::filesystem::path(filename);
        }
        catch (std::exception& e)
        {
            ShowErrorFmt("Invalid filename encountered, check your user/system locale/language settings! {} ({})", filename, e.what());
        }

        if (path.empty() || path.extension() == "")
        {
            return;
        }

```

zach2good:
I think I'd rather have logging here if this fails since I'm fairly sure there are some other bad things that can happen if your system is set up so that it fails here. It's currently the only repro/early warning that we have - so we should shout loudly instead of continuing silently

PR #5503, opened 2024-04-24
[LUA] Fix Aura Steal not applying shadows stolen

ampitere:
**_I affirm:_**

When a THF with Aura Steal merits for Steal attempts to take a mob's Utsusemi (of any kind) the effect looks like it is stolen but the modifier for Utsusemi is lost which is what is used in takeShadows. Hits from enemies will not clear the buff nor will abilities that should wipe shadows since it thinks you have 0. It will eventually just expire over time.

To fix this we will preserve the Utsusemi mod in case it does exist, then check to see if the stolen effect was copy images. If it is we will take the preserved mod value and set that to the player so that takeShadows can take these into account properly.

1. !changejob 6 99
2. Ensure you have 5/5 Aura Steal
3. !addeffect 66
4. !setmod 307 4
5. Steal
6. !reset if it misses

### DIFF
--- a/scripts/globals/job_utils/thief.lua
+++ b/scripts/globals/job_utils/thief.lua
@@ -417,10 +417,18 @@ xi.job_utils.thief.useSteal = function(player, target, ability, action)
         if resist > 0.0625 then
             local auraStealChance = math.min(player:getMerit(xi.merit.AURA_STEAL), 95)
             if math.random(100) < auraStealChance then
+                local targetShadows = target:getMod(xi.mod.UTSUSEMI)
+
                 stolen = player:stealStatusEffect(target)
                 if stolen ~= 0 then
                     ability:setMsg(xi.msg.basic.STEAL_EFFECT)
                     action:setAnimation(target:getID(), 181)
+
+                    if stolen == xi.effect.COPY_IMAGE then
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):
Good find, thanks

scripts/globals/job_utils/thief.lua
```diff
@@ -417,10 +417,18 @@ xi.job_utils.thief.useSteal = function(player, target, ability, action)
...
                 stolen = player:stealStatusEffect(target)
                 if stolen ~= 0 then
                     ability:setMsg(xi.msg.basic.STEAL_EFFECT)
                     action:setAnimation(target:getID(), 181)
+
+                    if stolen == xi.effect.COPY_IMAGE then
+                        if targetShadows > 0 then
+                            player:setMod(xi.mod.UTSUSEMI, targetShadows)
+                        end
+                    end
```
Xaver-DaRed, line 431: The way we handle utsusemi could be improved, in general. I dont see a reason why we couldnt just use power or subpower of the status effect, instead of a modifier.

Xaver-DaRed (approved):
Of course, approving anyway. Changing how utsusemi works is out fo scope of this PR

PR #5507, opened 2024-04-24
[LUA] Fix Avatar Blood Pacts eats actions when avatar is asleep, stunned, petrified, or already acting

ampitere:
**_I affirm:_**

Currently when a SMN attempts to quickly use 2 Blood Pacts (with Apogee or one Rage and one Ward) the second action will get eaten, the Rage/Ward will go on cooldown, and MP will be consumed. This change will add checks to ensure the avatar is not under a status effect that would prevent them from acting or already using an action in canUseBloodPact to prevent the second action from getting eaten.

Definitely requires retail verification, also quite a few TODOs for the message verification in canUseBloodPact.

Action
1. !changejob 15 99
2. Summon an Avatar
3. Use Apogee
4. Try to use 2 BPs in quick succession

Status effects
1. !changejob 15 99
2. Summon an Avatar
3. Target the avatar and !exec target:addStatusEffect(2, 1, 0, 100)
4. Try to use a BP

Petrify - 7
Sleep 1 - 2
Sleep 2 - 19
Lullaby - 193
Stun - 10

### DIFF
--- a/scripts/globals/job_utils/summoner.lua
+++ b/scripts/globals/job_utils/summoner.lua
@@ -175,6 +175,18 @@ xi.job_utils.summoner.canUseBloodPact = function(player, pet, target, petAbility
             return xi.msg.basic.TARG_OUT_OF_RANGE, 0
         end
 
+        local petAction = pet:getCurrentAction()
+
+        -- check if avatar is under status effect
+        if petAction == xi.action.SLEEP or petAction == xi.action.STUN then
+            return xi.msg.basic.PET_CANNOT_DO_ACTION, 0 -- TODO: verify exact message in packet.
+        end
+
+        -- check if avatar is using a move already
+        if petAction == xi.action.PET_MOBABILITY_FINISH then
+            return xi.msg.basic.PET_CANNOT_DO_ACTION, 0 -- TODO: verify exact message in packet.
+        end
+
         local baseMPCost = getBaseMPCost(player, petAbility)
 
         if player:getMP() < baseMPCost then


scripts/globals/job_utils/summoner.lua
```diff
@@ -175,6 +175,16 @@ xi.job_utils.summoner.canUseBloodPact = function(player, pet, target, petAbility
             return xi.msg.basic.TARG_OUT_OF_RANGE, 0
         end
 
+        -- check if avatar is asleep
+        if pet:hasStatusEffect(xi.effect.SLEEP_I) or pet:hasStatusEffect(xi.effect.SLEEP_II) then
```
Xaver-DaRed: Lullaby, pretrify and surely other status effects also impede action.

There is an status effect called lullaby, btw. It's just not used by lullabies, as weird as it sounds.

MowFord: pet:getCurrentAction() == xi.action.SLEEP should catch them all

i.e. 
```
pet:getCurrentAction() == xi.action.SLEEP or pet:getCurrentAction() == xi.action.STUN
```

Xaver-DaRed: Big should. Test that before pushing changes

ampitere: Good call Mow that works for Petrify, both Sleeps, Lullaby, and Stun. There's also a separate issue that was noted where a mob moves too far out of range as the BP is being activated which results in the BP going on cooldown and it eating MP. It does give the out of range message when it happens but it probably shouldn't go on cooldown and consume MP. Will keep that separate from this but wanted to bring it up in case anybody had ideas on a good fix.

WinterSolstice8: > There's also a separate issue that was noted where a mob moves too far out of range as the BP is being activated which results in the BP going on cooldown and it eating MP. Will keep that separate from this but wanted to bring it up in case anybody had ideas on a good fix.

Depending on the age of the info, and the avatar, that may be different now. All avatars are converted now so it would be worth double checking and comparing with retail

RAIST5150: Don't known if I'll catch it later, so commenting now while it is on the mind:

Will be important to track when retail processes the ability/mp consumption now for the out of range issue.  Seem to recall it being consumed regardless in retail back in the day... like the consumption/cooldown was bound to the player action and not the avatar's response to it.  Same thing could happen if someone killed the claimed target before the avatar completed the last BP called (even with wards)--the animation and all couid get interrupted but the timer was already set and MP taken. Wasn't like casting a spell where it waits until it hits the completion point .. fired off when the player called the ability and then the avatar was sort of a seperate action on its own after that?

WinterSolstice8 (approved):

PR #5508, opened 2024-04-24
Remove invalid pets in zone petList

### LINKED ISSUE #5439: 🐛 [Map.exe Crash] PUP Automaton in Instanced Zone
**_I affirm:_**

Windows 10

## Branch affected by issue 

`base` (Unmodified LSB)

## Steps to reproduce

Go into an instanced zone (Nyzul if enabled,  or any assault) as a PUP. Spawn your automaton, then despawn it, then spawn it again right away. Notice that the grey name of your despawned automaton might stay after you respawn the next. The grey name does not go away. You will likely experience a map crash as well after respawning the automaton once or after a couple tries.

image

[xi_map.exe_15-4_19-4-6.log](https://github.com/LandSandBoat/server/files/14986067/xi_map.exe_15-4_19-4-6.log)

## Expected behavior

Be able to summon Automaton without crashes.

MowFord:
**_I affirm:_**

Resolves the crash condition to close https://github.com/LandSandBoat/server/issues/5439 .

The issue seems to lie in deactivating a pet tearing down the pet entity without removing it from the zone's petlist

The steps are reproducible with this macro:
image

This PR seems to fully resolve the crash condition, but there's still the situation with spawning automatons too quickly leaving grey names around. I believe it's due to the same underlying issue, so perhaps someone more knowledgeable can help me update this PR (or make a new one and close this) to fix both.

Pets aren't removed from the player's `SpawnPETList` unless the it can be found via `((CPetEntity*)it->second)->id`, which as we see in this PR's solution, is bunk data.

- go into nyzul
- use macro above to spawn/despawn automatons
- go back to alz undersea ruins before the next zone tick: `!pos 131 0 20 72`

see these entries instead of a zone crash:
image

### DIFF
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -1595,6 +1595,15 @@ void CZoneEntities::ZoneServer(time_point tick)
         //     : this way, but we need to do this to keep allies working (for now).
         if (auto* PPet = static_cast<CPetEntity*>(it->second))
         {
+            if (PPet->m_PetID > PETID::MAX_PETID)
+            {
+                // Automatons don't get destroyed in this loop, clean up invalid pointer to avoid crashes
+                // static_cast seemingly makes up properties for this PPet object
+                ShowDebug("CZoneEntities::ZoneServer: Pet: Removing stale Automaton from petlist");
+                m_petList.erase(it++);
+                continue;
+            }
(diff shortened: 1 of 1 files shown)


claywar:
Superseded by #5534

PR #5510, opened 2024-04-24
[sql] Dyinyanga NM

MowFord:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

SQL side of [Dyinyinga](https://www.bg-wiki.com/ffxi/Dyinyinga)

## Steps to test these changes

PH-style nm with no lua code yet, but as with our other recent NMs I'm doing it in 2 stages to allow for review time of the lua without getting potential merge conflicts in the sql

PR #5512, opened 2024-04-24
[sql/lua] Dyinyinga NM

MowFord:
**_I affirm:_**

~~LUA side of [this pr](https://github.com/LandSandBoat/server/pull/5510)~~

[Dyinyinga](https://www.bg-wiki.com/ffxi/Dyinyinga) has a neat aura mechanic when the AE isn't applied to its target

https://imgur.com/1o5UQVS

Update the slug's lua to spawn the NM immediately on despawn: `xi.mob.phOnDespawn(mob, ID.mob.DYINYINGA_PH, 100, 3600, { immediate = true }) -- 1 hour`

Kill the PH

Fight NM, see it does the things

kill and see that it has drops


scripts/zones/Rolanberry_Fields_[S]/mobs/Scabrous_Slug.lua
```diff
@@ -0,0 +1,17 @@
...
+local ID = zones[xi.zone.ROLANBERRY_FIELDS_S]
+-----------------------------------
+local entity = {}
+
+entity.onMobDeath = function(mob, player, optParams)
+end
+
+entity.onMobDespawn = function(mob)
+    xi.mob.phOnDespawn(mob, ID.mob.DYINYINGA_PH, 10, 3600) -- 1 hour
+end
```
Comment (Xaver-DaRed) [inline scripts/zones/Rolanberry_Fields_[S]/mobs/Scabrous_Slug.lua:20]: You could define the PH table as a local here in the mob script instead of setting a local variable in the IDs file.

Like so...
In the IDs file
```lua
    mob =
    {
        DYINYINGA = GetFirstID('Dyinyinga')
    }
```
And in this file
```lua
local dyinyingaPHTable =
{
    [ID.mob.DYINYINGA - 1] = ID.mob.DYINYINGA 
}

end

    xi.mob.phOnDespawn(mob, dyinyingaPHTable, 10, 3600) -- 1 hour
end
```

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Scabrous_Slug.lua:20]: Is that how we'd like to do it? I thought we liked having the pH table in the zone table?

There's also the nms that have phs of different mob types

Comment (Xaver-DaRed) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Scabrous_Slug.lua:20]: I'll let others chime in, but I personally dont like local variables defined in the IDs files.

As an example, I do the same thing with Despot in this PR #5515

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Scabrous_Slug.lua:20]: I'll do whatever lsb team decides, but imo it's no different in terms of code cleanliness, but it does make the values part of the global table my way whereas your method doesn't

Comment (Xaver-DaRed) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Scabrous_Slug.lua:20]: Correct. Something that doesnt need to be global, shouldnt be global.
My way still makes the NM global anyway, which is all you need for it's placeholder.

scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua
```diff
@@ -1,11 +1,62 @@
...
+                end
+
+                mob:pathTo(newX, newY, newZ)
+            end
+        end
+    else
+        -- Aura and closes the gap
+        mob:setMobMod(xi.mobMod.TARGET_DISTANCE_OFFSET, 50)
+
+        for _, entry in pairs(mob:getEnmityList()) do
```
Comment (Xaver-DaRed) [inline scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: The aura only affects players in its hate list?

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: Such a small AOE I'd be surprised if it were anyone

Not sure we know 100%

Comment (WinterSolstice8) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: > The aura only affects players in its hate list?

This is generally true

scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua
```diff
@@ -1,11 +1,62 @@
...
```
Comment (WinterSolstice8) [inline scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: Is there any reason the real aura effect was not used?
Probably something like... (size is made up from an example)
```
mob:setMod(xi.mod.AURA_SIZE, 320) -- 6.25 * 320 / 100 = 20'
mob:addStatusEffectEx(xi.effect.COLURE_ACTIVE, xi.effect.COLURE_ACTIVE, 6, 3, 0, xi.effect.SLOW, 50, xi.auraTarget.ENEMIES, xi.effectFlag.AURA)
```
this is assuming that the aura doesn't have a visible bubble, not sure how to do that offhand, you can see if it uses a real aura in a capture if the status effect for SLOW has its duration constantly update.

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: There is no visible bubble. 

I had no idea we had actual aura code 

I'll check the capture for that, too

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: Well captures confirm it's an aura.

I tried what you suggested and got only the second addEffect to stick:

image

Edit: here's the code
```
        mob:addStatusEffectEx(xi.effect.COLURE_ACTIVE, xi.effect.COLURE_ACTIVE, 6, 3, 0, xi.effect.SILENCE, 50, xi.auraTarget.ENEMIES, xi.effectFlag.AURA)
        mob:addStatusEffectEx(xi.effect.COLURE_ACTIVE, xi.effect.COLURE_ACTIVE, 6, 3, 0, xi.effect.AMNESIA, 50, xi.auraTarget.ENEMIES, xi.effectFlag.AURA)
```

Comment (WinterSolstice8) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: currently you can only have one COLURE_ACTIVE effect, auras were designed with GEO in mind (and I dont think anyone was aware a mob could have _two_)... hmmm

Comment (TeoTwawki) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: auras and geo spells are slightly different while performing similar tasks. A GEO's spell can normally only applies one to one target: one on itself, one on a loupan, one in a 2nd party member via use of a JA...


While a mobs aura and throw a ton of crap at you at SE's whim

Comment (WinterSolstice8) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: Set one to `mob:addStatusEffectEx(xi.effect.NONE xi.effect.NONE,`
?

We can't see them anyway and you can mark that effect as undispelable, etc

edit: they are by default not dispellable with this method

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: How would it know to be a colour effect of we set it to none

Comment (MowFord) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Dyinyinga.lua]: Sorry for questioning, it works!

I guess the aura effect stuff is just global for all effects, neat

scripts/zones/Rolanberry_Fields_[S]/mobs/Scabrous_Slug.lua
Comment (Xaver-DaRed) [reply scripts/zones/Rolanberry_Fields_[S]/mobs/Scabrous_Slug.lua:20]: We have discussed it among us.

Use the method I described

Xaver-DaRed (changes requested):
Don't define local variables in IDs file.

Move Erle out aswell while you are at it.

MowFord:
> Don't define local variables in IDs file.
> 
> Move Erle out aswell while you are at it.

Updated and tested that they both work (by setting ph chance to 100% and immediate spawn)

Xaver-DaRed (approved):

PR #5514, opened 2024-04-24
[sql/lua] Nommo NM

MowFord:
**_I affirm:_**

Implements [Nommo](https://www.bg-wiki.com/ffxi/Nommo)

- Updates the behavior for the NM Nommo. Verified in retail.
- Updates the HP of Nommo. Verified in retail.
- Updates the drop rate of Nommo, Lyft Lance to something more realistic and standard.
- Fixes the drop from Tikbalang. There was a typo where the Lyft Lance was duplicated by accident. It should be Lyft Pole.
- Adds another mob AE for enamnesia

set PH to spawn him 100% of the time: `xi.mob.phOnDespawn(mob, ID.mob.NOMMO_PH, 100, 3600, { immediate = true }) -- 1 hour`

kill PH

fight Nommo, see it does stuff
image


scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua
```diff
@@ -1,9 +1,20 @@
 -----------------------------------
 -- Area: Pashhow Marshlands [S]
 --   NM: Nommo
+-- https://www.bg-wiki.com/ffxi/Nommo
 -----------------------------------
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:setMod(xi.mod.PETRIFYRES, 10000)
```
Comment (TeoTwawki) [inline scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: is it supposed to be super resistant or immune? if immune, edit mob pool, if resistant, this is set greatly higher than usual. unless the scale was change din another pr that escaped my notice this would be well beyond 100%

scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua
```diff
@@ -1,9 +1,20 @@
...
+    mob:addMod(xi.mod.MOVE_SPEED_STACKABLE, 25)
```
Comment (TeoTwawki) [inline scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: mobs should never have movement modifiers as a form of permanent speed boost. set his speed instead with either his mob family or the lua function to set his speed (the latter only if the family edit negatively impacts mobs that need a different speed)

TeoTwawki (changes requested):

scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua
Comment (MowFord) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: There is no petrify immunity defined in core, unfortunately

scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua
Comment (MowFord) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: understood, i'll grab the speed from the capture

scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua
Comment (MowFord) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: And you are right, this is a raw percent increase in resistance so 100 is what we want

image

Comment (TeoTwawki) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: if he were immune I'd say add it to the enum, but looks like he's not. that looks like a trait proc. which means bg is wrong when they say immune to petrify.

Comment (MowFord) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: That screenshot is from my local after changing to 100.

I'm pretty confident he is literally immune, but I'll verify when I confirm the speed

Comment (TeoTwawki) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: if on retail if it says "completely resists the <whatever>" he truly immune

Comment (MowFord) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: Confirmed he's truly immune

scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua
Comment (MowFord) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: Apologies, Dyinyinga has speed 100 via packet capture, Nommo does not

scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua
Comment (TeoTwawki) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: #5526

merry xmas in april

Comment (MowFord) [reply scripts/zones/Pashhow_Marshlands_[S]/mobs/Nommo.lua]: Updated to use the immunity value

TeoTwawki (approved):

PR #5515, opened 2024-04-25
[Cleanup] Make Ru'Aun Gardens zone shift-proof

Xaver-DaRed:
**_I affirm:_**

- Removes all tables from IDs file
- Use `GetFirstID()` with all npcs and mobs in the file.
- Correct portal event being wrongly assigned. The end-result is the same though, since that event does exactly the same as the new one, but now its retail-accurate.
- Document properly all regions.
- Accidentally added a text ID to full moon fountain, but thats gonna be needed anyway.


scripts/zones/RuAun_Gardens/Zone.lua
```diff
@@ -5,10 +5,78 @@ local ID = zones[xi.zone.RUAUN_GARDENS]
...
-    for k, v in pairs(ID.npc.PORTALS) do
-        zone:registerTriggerArea(k, unpack(v['coords']))
+local function handleClosingPortal(player, eventId, npcId)
+    if GetNPCByID(npcId):getAnimation() == xi.anim.OPEN_DOOR then
+        player:startOptionalCutscene(eventId)
+    end
+end
+
+local function handleGreenPortal(player)
+    if player:getCharVar('skyShortcut') == 1 then
```
TeoTwawki, line 15: later (not today) we should move this to the unlocks table, like so many other one off forever vars we've yet to deal with.

TeoTwawki: no wait I forgot, you can actually -lose- this one after gaining it, never mind! I forgot it breaks.

TeoTwawki (approved):

TeoTwawki:
CI griping about newlines in zone file, but looks good aside from Ci issues

(its the switch cases I think)

scripts/zones/RuAun_Gardens/mobs/Groundskeeper.lua
```diff
@@ -7,6 +7,26 @@ local ID = zones[xi.zone.RUAUN_GARDENS]
 -----------------------------------
 local entity = {}
 
+local despotPHTable =
+{
+    [ID.mob.DESPOT - 16] = ID.mob.DESPOT,
```
Frankie-hz, line 12: Edit: nvm

PR #5518, opened 2024-04-25
[cpp/lua] Bp recast

MowFord:
**_I affirm:_**

4 things
- Deactivate was right above the bp delay code and I noticed it has leftover code from before the lua file reset the recast.
  - The code is superfluous, so I removed it
- the MP cost of Cait Sith abilities at some point was broken due to the lack of a call to `onUseBloodpact`, this is fixed (and all BP tested)
- The recast IDs for various groups of JA were never defined as an enum
- The meat of this PR:
  - Blood Pact JA recast is only set on the successful completion of a petskill (similar to the MP consumption)
  - However (here's the tricky part) it is confirmed that bp delay gear is considered on use of the player JA, not when the pet performs the action
  - I've made this functionality work properly for all smn pets that use `onUseBloodpact` in lua, and left it the same for the fallback (which I believe is only Diabolos at this point)

Activate an automaton, make it full hp and deactivate to see activate timer reset still

Summon various avatars, give yourself `AVATAR_PERPETUATION` to negate the perp cost, and use various blood pacts to see proper mp consumption, but also no recast packet until the skill goes off (for all but diabolos)

set `BP_DELAY` and `BP_DELAY_II` mods to reduce bp recast:
image

https://imgur.com/qWpsnqg
(Note the mp consumption and bp ward recast happening when the skill goes off, with 45s due to having -15 bp_delay mod)

use diabolos skills and see he functions the same as before, consuming mp and recast on player JA use


src/map/entities/charentity.cpp
```diff
@@ -1655,6 +1645,10 @@ void CCharEntity::OnAbility(CAbilityState& state, action_t& action)
                     }
 
                     addMP((int32)-mpCost);
+                    if (this->GetLocalVar("bpRecastTime") > 0) // This will go away when all smn petskills are handled via jobutils/summoner.lua
```
WinterSolstice8, line 1656: `// This will go away when all smn petskills are handled via jobutils/summoner.lua`
So why isn't this PR adding it there? This C++ block is not necessary anymore. Perhaps this should be removed from this PR so the rest of the PR can be approved and merged sooner

scripts/globals/job_utils/summoner.lua
```diff
@@ -111,10 +118,12 @@ local function getBaseMPCost(player, ability)
 
     local baseMPCost = nil
 
-    if ability:getAddType() == xi.addType.ADDTYPE_ASTRAL_FLOW then
-        baseMPCost = player:getMainLvl() * 2
-    elseif ability ~= nil then
-        baseMPCost = baseMPCostMap[ability:getID()]
+    if ability then
```
WinterSolstice8, line 132: When/how is ability `nil` ? hmm...

src/map/entities/charentity.cpp
MowFord: Diablos doesn't use the new petskill db table yet

scripts/globals/job_utils/summoner.lua
MowFord: Hopefully never, but I just cleaned up the existing logic

src/map/entities/charentity.cpp
WinterSolstice8: ah... interesting that the pet animation hack in c++ wasn't necessary for diabolos

WinterSolstice8 (approved):

MowFord:
rebased

MowFord:
Oops, did not consider apogee for this rework

Edit: [fixed](https://imgur.com/rXbzqp0)

Xaver-DaRed (approved):

Xaver-DaRed:
Rebase pls

MowFord:
> Rebase pls

Sigh. I can't lol

Xaver-DaRed:
Rebased it for you

PR #5522, opened 2024-04-25
Resolve 2 players zoning in within 500ms and not rendering the other

### LINKED ISSUE #5519: 🐛 Players don't always render for other players on zone-in
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

takes some work to get it to happen, but basically both players have to zone in at approximately the same time:
image

https://imgur.com/Uo7sGID
Notice the minimap showing no player as well, mowford2 never changed positions during the running back and forth

MowFord:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/5519 .

The original find i made in the issue seemed inefficient packet-wise, but this workaround should work just fine.

The described issue with leaving a trial by avatar fight makes sense as all players get zoned at the same time, so very likely to trigger the condition described in the issue

see https://github.com/LandSandBoat/server/issues/5519

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -1180,7 +1180,14 @@ void SmallPacket0x01A(map_session_data_t* const PSession, CCharEntity* const PCh
             }
             else
             {
-                PChar->requestedInfoSync = true;
+                PChar->loc.zone->ForEachChar([&](CCharEntity* tempChar)
+                {
+                    // Avoids two players zoning in at slightly different times and missing each-other's update packet
+                    if (PChar == tempChar || distance(PChar->loc.p, tempChar->loc.p) < 50)
+                    {
+                        tempChar->requestedInfoSync = true;
+                    } });
+
                 PChar->loc.zone->SpawnNPCs(PChar);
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
This works but it's not the right fix, the root cause appears to be that the client isn't yet ready to receive spawn packets. I need to do some quick caps to verify when the server starts to send spawn packets, but i should have something better tonight (assuming I don't die from having worked too much this week already)

Basically, my assumption is that 0x01A with the 0x14 action is when the client is ready for some kinds of packets, some info via atom0s (his name is cut off due to my crappy screen res, sorry atom0s):

image

This doesn't constitute what the server does

tl;dr: need to see if 0x01A send corresponds to SPAWN 0x00D/0x00Es happening after it's sent.  Then we need a latch to prevent zone entities from being spawned in the loop until the server has received that

MowFord:
Cool deal, Ill just close this one out. We applied it as a workaround for now

WinterSolstice8:
Yeah it's not a bad workaround for sure, it will get the job done while I work on this

PR #5525, opened 2024-04-26
[cpp] Immanence logic cleanup

MowFord:
**_I affirm:_**

Simple fix to Immanence JA not being usable if already have the effect and giving a message that the effect wears off when casting a spell (just like all other stratagems)

Meat of the PR:
- Someone reported helixes weren't skillchaining or consuming the effect. I believe they dealing zero damage and correctly not applying skillchain effect. This is a common retail tactic
- magic numbers and duplicate calls to `GetSkillChainEffect` bothered me enough to refactor it
- Also made me read up a bit more and found that closing a skillchain with a helix boosts the skillchain effect duration by 2s

Edit: had someone log into retail and -ga, -ja, AM do not consume immanence or apply skillchain effect

image

I added a print to `effects/skillchain.lua` on effect gain and lose. first start/end was a simple nuke, second group was closing a skillchain with a helix:
image

image

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1131,70 +1131,69 @@ void CCharEntity::OnCastFinished(CMagicState& state, action_t& action)
                 StatusEffectContainer->DelStatusEffectSilent(EFFECT_CHAIN_AFFINITY);
             }
 
-            if (actionTarget.param > 0 &&
+            // Immanence will create or extend a skillchain for elemental spells
+            if (actionTarget.param >= 0 &&
                 PSpell->dealsDamage() &&
                 PSpell->getSpellGroup() == SPELLGROUP_BLACK &&
                 (StatusEffectContainer->HasStatusEffect(EFFECT_IMMANENCE)))
             {
-                SUBEFFECT effect = SUBEFFECT_NONE;
+                auto      immanenceApplies = true;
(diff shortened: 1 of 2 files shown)


MowFord:
Final testing after last commit: 
image

MowFord:
do i need to put explicit `[[fallthrough]];` on each?

WinterSolstice8 (approved):
I feel like the case/switch could be simplified, but it does the job

src/map/entities/charentity.cpp
```diff
@@ -1212,6 +1211,13 @@ void CCharEntity::OnCastFinished(CMagicState& state, action_t& action)
                         actionTarget.addEffectMessage = 287 + effect;
                     }
                     actionTarget.additionalEffect = effect;
+
+                    // Closing a skillchain with an immanence Helix will make the magic burst window longer
+                    auto scEffect = PTarget->StatusEffectContainer->GetStatusEffect(EFFECT_SKILLCHAIN, 0);
+                    if (isHelix && scEffect)
+                    {
+                        scEffect->SetDuration(12000);
```
WinterSolstice8, line 1219: Second thought: should this not get the duration then add 2000? Skillchains are supposed to have reduced duration at least for following skillchains.

WinterSolstice8 (changes requested):

PR #5526, opened 2024-04-26
Expand immunity enum to 32 bits and add petrify

TeoTwawki:
**_I affirm:_**

Expands immunity enum to 32 bits and adds petrify to it
- We maxed out the 16 bits we had already and needed 1 more to add petrify
- [Functions and bindings already expected it to be 32bits](https://github.com/LandSandBoat/server/blob/c4c83018c7d756b8ab55ade22f89f34bbf89f946/src/map/entities/battleentity.cpp#L142)
- Added enum to lua
- I also padded out the leading zeroes

**Future suggestion:** Drop the column in pools entirely and migrate to handling this in lua. Most mobs with immunities are NMs, the ones that aren't are likely candidates for mixins or mixin-like code. We already have the bindings.
```cpp
    bool hasImmunity(uint32 immunityID); // Check if the mob has immunity for a type of spell (immunity list in mobentity.h)
    void addImmunity(uint32 immunityID);
    void delImmunity(uint32 immunityID);
```
The comment saying mobentity.h is incorrect. probably got moved at some point. Also says spell not effect, but I see nothing for being immune to any spell, just status effects. ¯\\\_(ツ)\_/¯

pick on some mobs with immunities, see that they still function as intended


Xaver-DaRed (approved):

Xaver-DaRed (approved):

Xaver-DaRed:
Why github, why did you do that?

PR #5527, opened 2024-04-26
Failsafe to mob roaming home logic

MowFord:
**_I affirm:_**

Similar to [this PR](https://github.com/LandSandBoat/server/pull/5257), mob's that roam home get stuck on literally anything nonlinear in the navmesh

This mob is set to NO_DESPAWN and pulled far away from despot area. before this change he would stand still
image


MowFord:
Edit: he made it home

image

Xaver-DaRed (approved):

WinterSolstice8:
What problem is this solving? the change looks good but I'm not 100% sure what im looking at

MowFord:
> What problem is this solving? the change looks good but I'm not 100% sure what im looking at

Mobs do not roam home if the navmesh isn't perfect, similar to how pets used to get stuck following their owner

WinterSolstice8 (approved):
Thanks for the explanation, better than nothing I suppose.

Issue #5441, opened 2024-04-16
🐛 Multiple Pet Bug


UmeboshiXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Windows 10

## Branch affected by issue 

`base` (Unmodified LSB)

## Steps to reproduce
Set job to DRG/PUP, Use Call Wyvern. Use !changejob DRG 75. See that your pet commands have disappeared. Use Call Wyvern, see that you now have two wyverns following you.

Note: This works with any pet job with pup subbed. Only the most recently spawned pet has any logic/commands besides following the player.

Note: I have had players that were also able to trigger this intermittently by spamming abilities that spawn pets after zoning. Though the GM commands with job change is the easiest way to reproduce this. After zoning with multiple pets out, the extra pets seem to get cleared.

MowFord:
I actually ran into this bug while testing the fix in [this](https://github.com/LandSandBoat/server/pull/5531)

The issue here appears to be due to `LoadAutomaton(` calling `LoadAutomatonStats(` and resetting PPet, but most pets don't check if their master has null PPet so the old wyvern is just hanging around.

![image](https://github.com/LandSandBoat/server/assets/131182600/a94320b9-0f86-4226-9ad9-57f097fea2b4)

The changeJob and changeSJob bindings have completely separate logic from the packet_handler that receives a job change packet. But I believe both have the oversight that they don't handle puppetmaster specifically very well and assume they don't have a pet.

Simplest solution might be to just cleanly remove a pet before changing jobs in both places.

WinterSolstice8:
about ready to refactor automatons because of this spaghetti. I spent some time looking at the instance crash and it's related to something similar to this

tl;dr: 'ate PAutomaton

FFXILevelDown:
[...]

PR #5531, opened 2024-04-26
[cpp] Preserve ALL pet zoning info after loadautomaton

MowFord:
**_I affirm:_**

Continuation of [this PR](https://github.com/LandSandBoat/server/pull/5521).

After the second PR, zoning with an automaton would summon a fire spirit after going into the new zone. This would be in the log:

image

I could have sworn i tested both permutations, but apparently not.

This new change only messes with the `petZoningInfo` if the pet type is not an automaton and the `petZoningInfo` was such that we needed to preserve the pet

I've tried smn/pup, pup/smn, drg/pup, pup/drg and zoning with and without a pet seems to all work like it should

Edit: changed to draft because i got a crash re-doing the steps below:

image

I'm guessing that's the wyvern listener... 

- change job to pup, zone and see that your automaton setup menu loads (no active pet)
- change sub job to smn, summon avatar, zone and see that your avatar persists
- change main job to drg, call wyvern, zone and see that your wyvern persists
- change subjob to pup, zone and see wyvern persists
  - (note that the automaton setup UI takes a few ticks to show up before zoning, as the changesjob lua binding doesn't immediately send the CCharJobsPacket... might need to open the equipment screen)

### DIFF
--- a/src/map/utils/puppetutils.cpp
+++ b/src/map/utils/puppetutils.cpp
@@ -566,8 +566,7 @@ namespace puppetutils
 
     void LoadAutomatonStats(CCharEntity* PChar)
     {
-        // Save this since LoadPet() below changes it, but we don't want this changed
-        auto origPetID = PChar->petZoningInfo.petID;
+        auto origPetZoningInfo = PChar->petZoningInfo;
         switch (PChar->PAutomaton->getFrame())
         {
             default: // case FRAME_HARLEQUIN:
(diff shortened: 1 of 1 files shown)


MowFord:
i believe the steps to crash are:
- change drg/thf
- call wyvern
- zone in place
- changesjob pup
- zone in place
- you'll see your old wyvern still there dead and you won't have a pet
- call wyvern and zone in place

Sorry i have to step out for the afternoon, i'll keep cracking at this later tonight

MowFord:
The crash condition i ran into appears to be uniquely related to using `!changejob` or `!changesjob` with a pet out. More details in https://github.com/LandSandBoat/server/issues/5441

PR #5532, opened 2024-04-27
[LUA]Move BST JA to jobUtils file.

hooksta4:
**_I affirm:_**

- Creates Beastmaster jobUtils
- Moves all BST job abilities into jobUtils
- Removes all logic from individual job abilities and puts them in jobUtils
- Capture shows beast under sync are not able to call jug pet greater than sync'd level (currently allowed to do so) with @Xaver-DaRed 
    - two beastmasters sync'd with pets, used JA leave and tried to call 28 level jub while sync'd at 23. Retail returned message CANT_BE_USED_IN_AREA
 

Moves all job ability logic into centralized location (jobUtils)


scripts/actions/abilities/gauge.lua
```diff
@@ -8,27 +8,11 @@
 local abilityObject = {}
 
 abilityObject.onAbilityCheck = function(player, target, ability)
-    if player:getPet() ~= nil then
-        return xi.msg.basic.ALREADY_HAS_A_PET, 0
-    end
-
-    return 0, 0
+    return xi.job_utils.beastmaster.onAbilityCheckGuage(player, target, ability)
```
WinterSolstice8: Gauge is typoed here and in the job_utils lua

scripts/globals/job_utils/beastmaster.lua
```diff
@@ -0,0 +1,512 @@
...
+    return 0, 0
+end
+
+-- On Ability Use Unleash
+xi.job_utils.beastmaster.onUseAbilityUnleash = function(player, target, ability)
+    player:addStatusEffect(xi.effect.UNLEASH, 9, 0, 60)
+end
+
+-- On Ability Check For Leave, Heel and Stay.
+xi.job_utils.beastmaster.onAbilityCheckNilPet = function(player, target, ability)
```
WinterSolstice8, line 420: This function needs to check if your pet is a jug pet (or charmed pet). Leave/Heel/Stay shouldn't work on drg wyverns, automatons, summoner avatars, etc.

WinterSolstice8: It looks like other functions need that check too, might need a small refactor.

scripts/actions/abilities/gauge.lua
hooksta4: least i was consistent ;)

WinterSolstice8: Yup haha

scripts/globals/job_utils/beastmaster.lua
hooksta4: I think I understand what you are saying, for this PR, I didn't change anything that was already in current code base *except* for the level sync change. Not sure if that is in the scope of this PR?

hooksta4: BUT, would it be something like this?

`        if player:isJugPet() and
        pet:getObjType() ~= xi.objType.MOB`

hooksta4: although arent all the commands different for each like leave/release/dismiss/deactivate all from main Job?

WinterSolstice8: For generic ability check for "can use ability"
```
if
    player:isJugPet() or
    pet:getObjType() == xi.objType.MOB
then
```
I think the one you pasted is from familiar, which does behave differently for jug vs non-jug(?)

WinterSolstice8: p.s. I hate "isJugPet", but thats not our problem here :)

WinterSolstice8: > although arent all the commands different for each like leave/release/dismiss/deactivate all from main Job?

They are, but for some reason there's conditions where they show up when they should not. I recently saw Dismiss as DRG/PUP or PUP/DRG with an automation out. It's just a safety thing until I can find time to fix that

hooksta4: Interesting - i just went on a bst/pup and didnt see it in menu - not saying its not there but

WinterSolstice8: It's pure paranoia, not gonna lie
absolutely not me when I see a SMN/BST

hooksta4: haha love it!

hooksta4:
Quick update: Made a TODO list in On Ability Check Reward: It is not returning the correct message.

WinterSolstice8 (approved):

PR #5539, opened 2024-04-28
[lua] Twinkbrix

MowFord:
**_I affirm:_**

Cleans up the logic for [Twinkbrix](https://www.bg-wiki.com/ffxi/Twinkbrix). Started with a cherry-pick from ASB then fell down the rabbit hole of getting retail captures and changed basically everything

- The logic of giving key items based on event id is unchanged from ASB (in `onEventFinish`)
  - Note that the cooldown technically needs to be started when the battle is entered, but afaik all other ENMs do it in the npc lua so that is left as a TODO (esp. since the fights aren't implemented yet)
- The trade logic was mostly correct except that trading 2k gil should only teleport you if you have the gate dial.
  - sub logic exists that if you have the operating lever, prompt if you are sure:
  - image
- A message is added to the trade logic to revert to the `onTrigger` informational event if you trade the requirements for either key item
- The onTrigger function only ever gives information about the two ENM key items: `SHAFT_GATE_OPERATING_DIAL` and `SHAFT_2716_OPERATING_LEVER`
  - logic has so many possible combinations it felt impossible but after many captures I noticed the pattern
  - Event ID 50 is a generic "you got what you need" message
  - Event ID 53 is a generic "you can't get the operating lever until X time"
  - Event ID 52 has many possible combinations that vary depending on which key items you have/are on cooldown

Final note about trading 2716 gil:
- SE unveiled on a stream that this NPC had a special combination of gil and day to get a higher chance of winning. 
  - It is assumed to be 2716 on lightsday, due to the special message for trading 2716
  - It is highly likely the day is lightsday, due to player reports of winning often. Takes a bit more testing but we've yet to trade 2716 on lightsday and not win (current code in this PR makes the winning number <=13 in the range 2-20 so decent chance of losing still)
  - The 2716 on lightsday still says you need to roll between 2 and 13 so the higher chance is given (in this code) by changing the range of the random number. This matches what we've seen on retail:
    - image
    - image

trade the following things and notice the behavior depending on if you have the right key item/the ENM cooldown is in the future 
- 2k gil
  - if you have the key item: 
image

- 2716 gil
  - if you don't have key item and aren't on cooldown:  
image

- between 1k and 10k gil
  - same as above, without the special message
- Sylvan Stone item
  - if you don't have key item and aren't on cooldown:
image

### DIFF
--- a/scripts/zones/Oldton_Movalpolos/npcs/Twinkbrix.lua
+++ b/scripts/zones/Oldton_Movalpolos/npcs/Twinkbrix.lua
@@ -4,48 +4,89 @@
 -- Type: Warp NPC
 -- !pos -292.779 6.999 -263.153 11
 -----------------------------------
+local ID = require("scripts/zones/Oldton_Movalpolos/IDs")
+require("scripts/globals/teleports")
+require("scripts/globals/npc_util")
+-----------------------------------
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
+    local operatingLeverCD = player:getCharVar("[ENM]OperatingLever")
+    local gateDialCD = player:getCharVar("[ENM]GateDial")
     local mineShaftWarpCost = 2000
     local tradeGil = trade:getGil()
 
     if
+        player:hasKeyItem(xi.ki.SHAFT_2716_OPERATING_LEVER) and
+        npcUtil.tradeHasExactly(trade, { { "gil", mineShaftWarpCost } })
+    then
+        player:startEvent(56, 1781, 23, 1757, 177552692, 8, 17407, 15, 0)
+
+    elseif
         player:hasKeyItem(xi.ki.SHAFT_GATE_OPERATING_DIAL) and
-        npcUtil.tradeHas(trade, { { 'gil', mineShaftWarpCost } })
+        npcUtil.tradeHasExactly(trade, { { "gil", mineShaftWarpCost } })
     then
         player:startEvent(56)
+
     elseif
         not player:hasKeyItem(xi.ki.SHAFT_GATE_OPERATING_DIAL) and
-        tradeGil > 0 and
-        tradeGil <= 10000 and
-        npcUtil.tradeHas(trade, { { 'gil', tradeGil } })
+        tradeGil > 0 and tradeGil <= 10000 and
+        npcUtil.tradeHasExactly(trade, { { "gil", tradeGil } }) and
+        gateDialCD < VanadielTime()
     then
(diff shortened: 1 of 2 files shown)


scripts/zones/Oldton_Movalpolos/npcs/Twinkbrix.lua
```diff
@@ -6,48 +6,117 @@
...
     if
         player:hasKeyItem(xi.ki.SHAFT_GATE_OPERATING_DIAL) and
-        npcUtil.tradeHas(trade, { { 'gil', mineShaftWarpCost } })
+        npcUtil.tradeHasExactly(trade, { { 'gil', mineShaftWarpCost } })
     then
-        player:startEvent(56)
+        -- teleport for mineShaftWarpCost relies on having SHAFT_GATE_OPERATING_DIAL
+        -- but consumes SHAFT_2716_OPERATING_LEVER (after confirming with you)
+        if player:hasKeyItem(xi.ki.SHAFT_2716_OPERATING_LEVER) then
+            player:startEvent(56, 1781, 23, 1757, 177552692, 8, 17407, 15, 0)
```
WinterSolstice8: This looks like an NPC ID, can you confirm it's either you and or/ the goblin? `177552692`

MowFord: npc id is 16822535

image

Looks like it works fine without the rest of the params. Want to nix them?

Xaver-DaRed: The first parameter after event ID is an item id. Concretely, the item you defined.

MowFord: Sure seems that way.

Xaver-DaRed (changes requested):
Fist paramter IS the item ID. Use the enum you already defined instead of the raw ID

MowFord:
> Fist paramter IS the item ID. Use the enum you already defined instead of the raw ID

Understood, would you like to remove the extra params from event 56 since they don't seem required? I can leave a comment with the info from the capture

scripts/zones/Oldton_Movalpolos/npcs/Twinkbrix.lua
MowFord: Updated

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #5547, opened 2024-04-29
[cpp] Update to allow sneak to be silent

MowFord:
**_I affirm:_**

Recent packet rewrite made sneak not quiet a player's footsteps. Going by what was previously in `char_update.cpp` we can see that it was just before the new player flag:

image

https://imgur.com/Olx9GAz

### DIFF
--- a/src/map/packets/char_update.cpp
+++ b/src/map/packets/char_update.cpp
@@ -94,7 +94,7 @@ namespace charUpateFlags
     {
         uint32_t LfgMasterFlag : 1;
         uint32_t TrialFlag : 1;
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
That isn't just one bit before 0x10

edit: the order of flags was different, disregard

MowFord:
Sorry I neglected to include the video after the changes. Updated

WinterSolstice8 (approved):
Interesting. Are we really sending 0x037s past the first char update...? I believe that is not correct. Not your problem here, though

Xaver-DaRed (approved):

PR #5552, opened 2024-04-29
[SQL/LUA] Fix jug Antlion unable to use ready abilities

ampitere:
**_I affirm:_**

The abilities Sandblast and Sandpit in the BST jug Antlion pet (ChopsueyChucky) are named slightly different than the actual mob skills (which are Sand Blast and Sand Pit). This causes the jug pet to not be able to ready these abilities. To fix this we will add the space to match up the mob skill names. In game it still displays as Sandblast and Sandpit in the menu and messages.

1. !changejob 1 99
2. !additem 17892 12
3. Equip the jug and use Call Beast
4. Use Fight on a mob
5. Attempt to use Sandblast or Sandpit
6. !reset if necessary

image


Xaver-DaRed:
They are named differently because they are different skills that probably need a different script and a different database definition

ampitere:
> They are named differently because they are different skills that probably need a different script and a different database definition

There are commented out records in mob_skills.sql I was originally using and creating separate mob skills that do not have spaces but this change seems to work. Happy to go down that road again if you'd like.

TeoTwawki:
I wanna know what ones captures say happen for both jug and not-jug, and naming of each: space vs no space.

That dictates the path we should take. If I need to dig out my bst and buy a jug, say the word.

ampitere:
> I wanna know what ones captures say happen for both jug and not-jug, and naming of each: space vs no space.
> 
> That dictates the path we should take. If I need to dig out my bst and buy a jug, say the word.

I definitely don't have retail access currently so any retail captures that puts me on the right path would be appreciated!

TeoTwawki:
> > I wanna know what ones captures say happen for both jug and not-jug, and naming of each: space vs no space.
> > That dictates the path we should take. If I need to dig out my bst and buy a jug, say the word.
> 
> I definitely don't have retail access currently so any retail captures that puts me on the right path would be appreciated!

alright, in a bit I'll go confirm which IDs are jug and which are mob, and which naming convention is correct. I'm getting unwanted OT atm coz some guy at work clicked a virus banner

Xaver-DaRed:
This PR seems abandoned and the retail data still needs to be confirmed. I'll close it and open a new one (if needed) once I get the captures

PR #5553, opened 2024-04-29
[cpp] Allow changeJob binding to used on mobs

CatsEyeXI:
**_I affirm:_**

Allows changeJob() to be used on mobs. 
This functionality is needed to properly implement ENM's like "Pulling the Strings".

!zone Dragon's Aery
!getstats
Target a Spider
!exec target:changeJob(xi.job.PLD)
!getstats

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5919,69 +5919,157 @@ uint8 CLuaBaseEntity::getSubJob()
 
 void CLuaBaseEntity::changeJob(uint8 newJob)
 {
-    auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity);
-
-    if (!PChar)
+    if (m_PBaseEntity && m_PBaseEntity->objtype == TYPE_PC)
     {
-        ShowWarning("CLuaBaseEntity::changeJob() - m_pBaseEntity is not of type Char.");
-        return;
-    }
+        auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity);
 
-    JOBTYPE prevjob = PChar->GetMJob();
-
-    PChar->resetPetZoningInfo();
-
-    PChar->jobs.unlocked |= (1 << newJob);
-    PChar->SetMJob(newJob);
-
-    if (newJob == JOB_BLU)
-    {
-        if (prevjob != JOB_BLU)
+        if (!PChar)
         {
-            blueutils::LoadSetSpells(PChar);
+            ShowWarning("CLuaBaseEntity::changeJob() - m_pBaseEntity is not of type Char.");
+            return;
         }
+
+        JOBTYPE prevjob = PChar->GetMJob();
+
+        PChar->resetPetZoningInfo();
+
+        PChar->jobs.unlocked |= (1 << newJob);
+        PChar->SetMJob(newJob);
+
+        if (newJob == JOB_BLU)
+        {
+            if (prevjob != JOB_BLU)
+            {
+                blueutils::LoadSetSpells(PChar);
+            }
+        }
+        else if (PChar->GetSJob() != JOB_BLU)
+        {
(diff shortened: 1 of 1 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -5919,69 +5919,157 @@ uint8 CLuaBaseEntity::getSubJob()
 
 void CLuaBaseEntity::changeJob(uint8 newJob)
 {
-    auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity);
-
-    if (!PChar)
+    if (m_PBaseEntity && m_PBaseEntity->objtype == TYPE_PC)
```
zach2good: `if (auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity)) ...`

src/map/lua/lua_baseentity.cpp
```diff
@@ -5919,69 +5919,157 @@ uint8 CLuaBaseEntity::getSubJob()
...
+        PChar->pushPacket(new CCharRecastPacket(PChar));
+        PChar->pushPacket(new CCharAbilitiesPacket(PChar));
+        PChar->pushPacket(new CCharUpdatePacket(PChar));
+        PChar->pushPacket(new CMenuMeritPacket(PChar));
+        PChar->pushPacket(new CMonipulatorPacket1(PChar));
+        PChar->pushPacket(new CMonipulatorPacket2(PChar));
+        PChar->pushPacket(new CCharSyncPacket(PChar));
     }
-    else if (PChar->GetSJob() != JOB_BLU)
+    else
```
zach2good: `else if (auto* PMob = dynamic_cast<CMobEntity*>(m_PBaseEntity)) ...`

src/map/lua/lua_baseentity.cpp
```diff
@@ -5919,69 +5919,157 @@ uint8 CLuaBaseEntity::getSubJob()
...
+                PWeapon->setSkillType(SKILL_STAFF);
+                PWeapon->setDelay(8000);
+                PWeapon->setBaseDelay(8000);
+                break;
+            default:
+                break;
+        }
+
+        mobutils::CalculateMobStats(PMob);
+    }
```
zach2good, line 6051: `else { Error condition logging here }`

src/map/lua/lua_baseentity.cpp
```diff
@@ -5919,69 +5919,157 @@ uint8 CLuaBaseEntity::getSubJob()
...
+        PChar->pushPacket(new CCharJobsPacket(PChar));
+        PChar->pushPacket(new CCharStatsPacket(PChar));
+        PChar->pushPacket(new CCharSkillsPacket(PChar));
+        PChar->pushPacket(new CCharRecastPacket(PChar));
+        PChar->pushPacket(new CCharAbilitiesPacket(PChar));
+        PChar->pushPacket(new CCharUpdatePacket(PChar));
+        PChar->pushPacket(new CMenuMeritPacket(PChar));
+        PChar->pushPacket(new CMonipulatorPacket1(PChar));
+        PChar->pushPacket(new CMonipulatorPacket2(PChar));
+        PChar->pushPacket(new CCharSyncPacket(PChar));
```
zach2good: Presumably these big chunks of logic are replicated from somewhere, so it makes sense to extract them into a `charutils` function. Same with the chunk below into `mobutils`

CatsEyeXI: please bear with me as I'm learning. To be clear im just adding to the existing binding, so it isn't exactly duplicative, right? is the recommendation to move this entire logic into charutils?

zach2good: Don't worry about it. Just do the recommended styling changes.

CatsEyeXI: I think that's everything. Please let me know if I failed to make any of the changes as requested.

zach2good (approved):

PR #5554, opened 2024-04-29
[LUA]Clears enmity when using Staging Point doors.

hooksta4:
**_I affirm:_**

- Changed startEvent() to startOptionalEvent() which will clear enmity.

Mob should not follow you into staging point. 
Go to staging point, aggro a mob, reenter staging point via door. Enmity will be lost for player and pet as per retail (No mobs in the champagne room!)

### DIFF
--- a/scripts/zones/Arrapago_Reef/npcs/_jib.lua
+++ b/scripts/zones/Arrapago_Reef/npcs/_jib.lua
@@ -15,7 +15,7 @@ entity.onTrigger = function(player, npc)
         if player:getXPos() < 8 then
             player:messageSpecial(ID.text.STAGING_GATE_ILRUSI)
             player:messageSpecial(ID.text.STAGING_GATE_INTERACT)
-            player:startEvent(106)
+            player:startOptionalCutscene(106)
         elseif not player:hasKeyItem(xi.ki.ILRUSI_ASSAULT_ORDERS) then
             player:messageSpecial(ID.text.STAGING_GATE_ILRUSI)
             player:messageSpecial(ID.text.STAGING_GATE_INTERACT)
--- a/scripts/zones/Bhaflau_Thickets/npcs/_1g1.lua
+++ b/scripts/zones/Bhaflau_Thickets/npcs/_1g1.lua
@@ -15,7 +15,7 @@ entity.onTrigger = function(player, npc)
         if player:getZPos() > -761 then
             player:messageSpecial(ID.text.STAGING_GATE_MAMOOL)
             player:messageSpecial(ID.text.STAGING_GATE_INTERACT)
-            player:startEvent(106)
+            player:startOptionalCutscene(106)
         elseif not player:hasKeyItem(xi.ki.MAMOOL_JA_ASSAULT_ORDERS) then
             player:messageSpecial(ID.text.STAGING_GATE_MAMOOL)
             player:messageSpecial(ID.text.STAGING_GATE_INTERACT)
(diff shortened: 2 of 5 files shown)


Xaver-DaRed:
This functionality already exists
startCutscene()
startOptionalCutscene()

Both do that

hooksta4:
fair enough.

hooksta4:
There might be some other instances where we would want to clearEnmity from a function at some point maybe?

src/map/lua/lua_baseentity.h
```diff
@@ -631,6 +631,7 @@ class CLuaBaseEntity
     void  setClaimable(bool claimable);
     bool  getClaimable();
     void  clearEnmityForEntity(CLuaBaseEntity* PEntity);
+    void  clearEnmity(CLuaBaseEntity* PEntity); // clears player enmity from notoriety list
```
WinterSolstice8: This is actually clearing `EnmityContainer`, I assume this was an old comment (they have similar names but different functions)

Xaver-DaRed:
Definately. Hence why I dont opose to implementing that function.
But I dont think we need to use it in those concrete places (having an actual event)

In situations where we dont have an event to trigger? definately

src/map/lua/lua_baseentity.h
hooksta4: fixed.

hooksta4:
> Definately. Hence why I dont opose to implementing that function. But I dont think we need to use it in those concrete places (having an actual event)
> 
> In situations where we dont have an event to trigger? definately

Updated the SP with the startCutscene but left the clearEnmity() in.

scripts/zones/Arrapago_Reef/npcs/_jib.lua
```diff
@@ -15,7 +15,7 @@ entity.onTrigger = function(player, npc)
         if player:getXPos() < 8 then
             player:messageSpecial(ID.text.STAGING_GATE_ILRUSI)
             player:messageSpecial(ID.text.STAGING_GATE_INTERACT)
-            player:startEvent(106)
+            player:startCutscene(106)
```
Xaver-DaRed: Ok, so, when an event returns an option ("wanna open this door?" "wanna use the portal?" etc) the function to use is `:startOptionalCutscene()`

The stagging point does ask you.

src/map/lua/lua_baseentity.cpp
```diff
@@ -12362,6 +12362,31 @@ sol::table CLuaBaseEntity::getNotorietyList()
...
 
+/************************************************************************
+ *  Function: clearEnmity()
+ *  Purpose : Clears all enmity towards target
+ *  Example : mob:clearEnmity(target)
+ *  Notes   :
+ ************************************************************************/
+void CLuaBaseEntity::clearEnmity(CLuaBaseEntity* PEntity)
+{
+    (m_PBaseEntity->objtype != TYPE_MOB);
```
Xaver-DaRed: ?

Was this a check to return at some point?

hooksta4:
I am not sure why the CI is blowing me up? I don't get any errors on load.

src/map/lua/lua_baseentity.cpp
Xaver-DaRed: From ASB
imagen

hooksta4: XI_DEBUG_BREAK oh that whole line should be gone then eh

Xaver-DaRed: Well, you probably want to check for entity type, insead of deleting the check entirely

Xaver-DaRed (approved):
Repurposed to make stagging points doors clear enmity

PR #5557, opened 2024-04-30
[cpp/lua] Sacrifice Torque

MowFord:
**_I affirm:_**

This PR does 3 major things
- Adds a player lua binding to check if an item in a certain slot has its latents active
- Fixes the latent effect for "Avatar in the party" to properly understand latent condition 21 to mean _any_ smn pet
- Adds final Sacrifice Torque latent effect to follow logic from [this PR](https://github.com/LandSandBoat/server/pull/5487) for Ferver Ring
  - [Sacrifice torque](https://www.bg-wiki.com/ffxi/Sacrifice_Torque) already has the regen and refresh latents, which should stay as regular latents of course
  - the pet attack latent should be in lua as it only matters/applies if the smn has a pet out, which can only happen if they cast it

for testing the player binding, equip any item with a latent and check map server output from running: `!exec player:allLatentsActive(xi.slot.BLAH)`

here it is before and after summoning ifrit with the sacrifice torque: `!exec player:allLatentsActive(xi.slot.NECK)`
image

For testing the updated latent, again sacrifice torque is a simple way to go
image

and finally, testing this updated logic to add the last latent to sacrifice torque:
image

### DIFF
--- a/scripts/items/sacrifice_torque.lua
+++ b/scripts/items/sacrifice_torque.lua
@@ -19,7 +19,7 @@ end
 
 itemObject.onItemEquip  = function(user, item)
     local listenerName = fmt('{}_{}', listenerPrefix, item:getID())
-    if user:allLatentsActive(xi.slot.NECK) then
+    if user:hasAllLatentsActive(xi.slot.NECK) then
         user:setLocalVar(listenerName, 1)
         xi.itemUtils.handlePetLatentMods(user, latentPetId, latentMods, true)
     end
@@ -38,6 +38,7 @@ itemObject.onItemUnequip = function(user, item)
         xi.itemUtils.handlePetLatentMods(user, latentPetId, latentMods, false)
     end
 
+    user:setLocalVar(listenerName, 0)
     user:removeListener(listenerName)
 end
 
--- a/scripts/items/sacrifice_torque.lua
+++ b/scripts/items/sacrifice_torque.lua
@@ -1,14 +1,16 @@
 -----------------------------------
--- ID: 11675
--- Fervor Ring
+-- ID: 15528
+-- Sacrifice Torque
 -- Pet mod via latent effect
+-- Latents are added to player before OnItemEquip is called
+-- Latents are removed from player before OnItemUnequip is called
 -----------------------------------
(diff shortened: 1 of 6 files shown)


src/map/latent_effect_container.h
```diff
@@ -66,6 +66,7 @@ class CLatentEffectContainer
 
     void AddLatentEffects(std::vector<CItemEquipment::itemLatent>& latentList, uint8 reqLvl, uint8 slot);
     void DelLatentEffects(uint8 reqLvl, uint8 slot);
+    bool AllLatentsActive(uint8 slot);
```
Xaver-DaRed: i'd propose `HasAllLatentsActive()`
We do use Add, Del, Get and Has elsewhere

scripts/items/sacrifice_torque.lua
```diff
@@ -0,0 +1,44 @@
...
+}
+
+itemObject.onItemCheck = function(target)
+    return 0
+end
+
+itemObject.onItemEquip  = function(user, item)
+    local listenerName = fmt('{}_{}', listenerPrefix, item:getID())
+    if user:hasAllLatentsActive(xi.slot.NECK) then
+        user:setLocalVar(listenerName, 1)
```
WinterSolstice8, line 23: when does this get removed?

WinterSolstice8: (or set to 0)

MowFord: Oops it should get reset on unequip regardless if you have a pet, for consistency

MowFord: Fixed.

FWIW i'm not sure there's a scenario where that would have caused a problem, since this new logic for the handful of "specific-avatar latent mods" applies mods to pets. If this latent were inactive (not setting the localvar), then you wouldn't have a pet for the localvar to matter.

Good catch regardless. Better to be explicit than make assumptions

WinterSolstice8 (approved):

EpicTaru (approved):

Xaver-DaRed (approved):

PR #5563, opened 2024-04-30
[sql] Shadow ring has 25% deathres

MowFord:
**_I affirm:_**

Perhaps the units of `DEATHRES` used to be a different scale, but currently it's only used [here](https://github.com/search?q=repo%3ALandSandBoat%2Fserver%20deathres&type=code), the rate is out of 100, and [the rate](https://www.bg-wiki.com/ffxi/Shadow_Ring) is well-known to be 25%

### DIFF
--- a/sql/item_mods.sql
+++ b/sql/item_mods.sql
@@ -23062,7 +23062,7 @@ INSERT INTO `item_mods` VALUES (14645,5,30);   -- MP: 30
 INSERT INTO `item_mods` VALUES (14645,21,-30); -- LIGHT_RES: -30
 
 -- Shadow Ring
-INSERT INTO `item_mods` VALUES (14646,255,5);  -- DEATHRES: 5
+INSERT INTO `item_mods` VALUES (14646,255,20); -- DEATHRES: 20
 INSERT INTO `item_mods` VALUES (14646,476,13); -- MAGIC_NULL: 13
 
 -- Telluric Ring


claywar (dismissed):

Xaver-DaRed:
Considering how "Resist Status" traits work, I'm gonna ask for this value to be changed to 20.

Since the enfeebling system already adds an extra 5% just because of the trait existing (modifier being > 0)

This spell will be converted to enfeebling magic system in a different PR

MowFord:
> Considering how "Resist Status" traits work, I'm gonna ask for this value to be changed to 20.
> 
> Since the enfeebling system already adds an extra 5% just because of the trait existing (modifier being > 0)
> 
> This spell will be converted to enfeebling magic system in a different PR

Since it's only used in the spell lua what if we make it 20 in SQL. Then give base 5 in death.lua, since that will be rewritten

Xaver-DaRed:
No, couse then i will have to clean that up aswell and I may forget. Just the sql, ty.

TeoTwawki:
"well known" eh?

well lessee...
> Resist Death proc rate appears to be about 25%.

"appears". ok but there are citations below lets check those

> After a small discussion on Shadow Ring proc rate last night, someone went back to the parsed logs on the last 5 Tiamats and extracted some Shadow Ring data. I have to say we were pretti suprised. It turned out to be around 13% proc rate over 1000+ hits. Special is our LS Taru PLD~ (Data from Daxam~)
>
> [http://img162.imageshack.us/img162/6...atsringhd5.gif](http://img162.imageshack.us/img162/6199/tiamatsringhd5.gif)

hum. Ok how about the other.

> 121 deaths cast.
32 resists.
96 dead.
26.4% Proc rate.
> 
> This is a small sample size, so the real value could like +/- 8%

Sigh. #shitWikiSays 25 is a guess leaning toward a high result from a low sample size. At best, I'd guess 20 without more data. The larger sample parse showed only 13% but thats an odd number to land at and luck is involved.

LacunaXI:
The magic nullification should be 12%.

Greater testing showed it to be 12% and not 13%.

https://www.ffxiah.com/forum/topic/45755/does-occ-absorbs-magic-damage-stack#3697946

WinterSolstice8 (approved):
https://www.ffxiah.com/forum/topic/52250/resist-testing/6/#3681433

```
Angra Mainyu (NM) -- Resist Death +48 (Shadow Ring, Warder's Ring, Odium)
"Resist!": 84
DEAD: 100
Total: 184
Percentage: 45.7%
```
Warders ring and odium are known values, Shadow ring may be in fact 25%. This kind of testing is especially grueling and Sagasinger has a similar power level (50% chance of resist terror)
At the very least, it's >=20% with the limited sample size. 5% is far too low given the data we have.

TeoTwawki:
It may well be ABOVE 25 but sample sizes weren't large enough to get past the tester own luck being a factor. Or it could be more complex and death itself has a power value involved making it more or less usefull vs certain sources of death.

I simply dispute the phrasing : this isn't knowm, this an unknown best guess based on limited data. 

I'm not even asking for changes. I'm calling a  spade a spade : wiki is bs that had link that didn't even corroborate its own statement, largesy sample in their own links was lower. A smaller sample showed it even higher. Obviously its > 5. But if someome wants to change it layer (up or dowm) its going to take more than this.

sql/item_mods.sql
```diff
@@ -23062,7 +23062,7 @@ INSERT INTO `item_mods` VALUES (14645,5,30);   -- MP: 30
 INSERT INTO `item_mods` VALUES (14645,21,-30); -- LIGHT_RES: -30
 
 -- Shadow Ring
-INSERT INTO `item_mods` VALUES (14646,255,5);  -- DEATHRES: 5
+INSERT INTO `item_mods` VALUES (14646,255,20);  -- DEATHRES: 20
```
Xaver-DaRed, line 23065: Please correct the alignment of the comment.

Xaver-DaRed (changes requested):
I dont want this to be missed in the comments.

sql/item_mods.sql
MowFord: Done

Xaver-DaRed (approved):

PR #5570, opened 2024-04-30
[Chore] Add more battlefield text IDs

Xaver-DaRed:
**_I affirm:_**

Adds text IDs needed for the battlefield rewrite.

None


scripts/zones/Ghelsba_Outpost/IDs.lua
```diff
@@ -23,7 +30,8 @@ zones[xi.zone.GHELSBA_OUTPOST] =
         LOGIN_NUMBER                     = 7535, -- In celebration of your most recent login (login no. <number>), we have provided you with <number> points! You currently have a total of <number> points.
         MEMBERS_LEVELS_ARE_RESTRICTED    = 7555, -- Your party is unable to participate because certain members' levels are restricted.
         FISHING_MESSAGE_OFFSET           = 7593, -- You can't fish here.
-        HUT_DOOR                         = 7727, -- This looks like an Orcish dwelling. The door is firmly shut.
```
claywar, line 26: It amazes me that this was defined but unused

claywar (approved):

scripts/zones/Ghelsba_Outpost/IDs.lua
Xaver-DaRed: yep. It may have been used at some point but removed for unknown reasons

TeoTwawki: I think that line is supposed to be used at a step in the missions that use the BCNM entrance, but when they cannot enter because they aren't on that step, and are not doing any bcnm there. But iirc, that should be a cs event anyway and not a message special. Age old code often had messages done that way. I'd need to see a cap to know for sure but if we ever need/use it, we can always add it then.

TeoTwawki: oh, I didn't realize you were renaming, thought it was being deleted. Should have scrolled down more!

PR #5572, opened 2024-05-01
[Cleanup] Shift-proof West Saru [S], West Saru, Xarcabard, Yughott Grotto

Frankie-hz:
**_I affirm:_**

Shift-proof the following zones: West Saru [S], West Saru, Xarcabard, Yughott Grotto

**West Sarutabaruta**
image
The two PHs nearest I-7 are these two IDs: 
```sql
INSERT INTO `mob_spawn_points` VALUES (17248466,'Mandragora','Mandragora',15,77.509,-20.719,434.757,127);
INSERT INTO `mob_spawn_points` VALUES (17248467,'Mandragora','Mandragora',15,31.149,-20.045,358.265,61);
INSERT INTO `mob_spawn_points` VALUES (17248468,'Tom_Tit_Tat','Tom Tit Tat',25,139.943,-35.719,299.835,127);
```
This follows SE's normal setup on how to determine what mob is an actual PH. During Horizons massive zone audit this pattern can be seen throughout the entire codebase. PH mobs are usually in ID order. A lot of the IDs and positions on the LSB spawn points table are incorrect. I only know this from capturing every single zone and their mob positions with 0-1 players in them. The same goes for Nunyenunc
```sql
INSERT INTO `mob_spawn_points` VALUES (17248514,'Carrion_Crow','Carrion Crow',11,159.501,-20.117,485.528,31);
INSERT INTO `mob_spawn_points` VALUES (17248515,'Carrion_Crow','Carrion Crow',11,-7.194,-17.288,431.604,127);
INSERT INTO `mob_spawn_points` VALUES (17248516,'Carrion_Crow','Carrion Crow',11,53.159,-24.540,554.652,93);
INSERT INTO `mob_spawn_points` VALUES (17248517,'Nunyenunc','Nunyenunc',26,-25.000,-15.000,385.000,127);
```

**Xarcabard**
Updated the obscure PH ID from Shadow Eye and corrected mobs positions based on retail capture.

Each commit has its separate zone for review.


Frankie-hz:
Note: Voidwalker needs a refractor to remove all the tables from the ID files so I left it as is until we figure out how to do all the IDs at once.

cocosolos:
> A lot of the IDs and positions on the LSB spawn points table are incorrect.

IIRC capture tools will record the "spawn" position as wherever the mob is when it appears to the player, whether it's by zoning in, getting within 50', or actually spawning. Unless I'm mistaken, the only way to actually get accurate spawn points would be to kill mobs and stay within 50' until they respawn.

You can see in some areas where mobs spawn in unnaturally tight groups, as if they were being AoE'd by another player when the capturer came into range.

Xaver-DaRed (approved):

TeoTwawki:
17248466	Mandragora    <- not PH
17248467	Mandragora    <- PH
17248468	Tom Tit Tat

17248484	Mandragora    <- not PH
17248485	Mandragora    <- PH
17248486	Tom Tit Tat


Sorry I am late t get here. Both Tom's can be spawned back to back, similarly to Leaping Lizzy. They share spawn positions too, even tho one of them's PH is way north of the other. He can also spawn way north. They can use any spawn node any of their PH's could have used basically.

Frankie-hz:
> 17248466 Mandragora <- not PH 17248467 Mandragora <- PH 17248468 Tom Tit Tat
> 
> 17248484 Mandragora <- not PH 17248485 Mandragora <- PH 17248486 Tom Tit Tat
> 
> Sorry I am late t get here. Both Tom's can be spawned back to back, similarly to Leaping Lizzy. They share spawn positions too, even tho one of them's PH is way north of the other. He can also spawn way north. They can use any spawn node any of their PH's could have used basically.

I must of missed that 2nd NM due to the IDs of the NM all being the same. I will fix it when I get home tonight.

cocosolos:
> 17248466 Mandragora <- not PH 17248467 Mandragora <- PH 17248468 Tom Tit Tat
> 
> 17248484 Mandragora <- not PH 17248485 Mandragora <- PH 17248486 Tom Tit Tat
> 
> Sorry I am late t get here. Both Tom's can be spawned back to back, similarly to Leaping Lizzy. They share spawn positions too, even tho one of them's PH is way north of the other. He can also spawn way north. They can use any spawn node any of their PH's could have used basically.

I'm pretty sure that all of these 2-ID NMs are also able to be their own PH. I've had LL pop from itself and a couple others while doing Magian trials.

image

TeoTwawki:
@cocosolos I'm unsure if all of them have zero cooldown but I've seen some that don't quite fit the pattern also pop off their own death. I've had 2 of Jaggedy Eared Jack back to back, but he only has one placeholder. I've had 3 leaping lizzzy in 15 min before, and the PH repop time is 5 min so clearly no cooldown there. Panzer percival (discussed more below) also has 2 PH but only one NM, and one day on a lucky streak I got FIVE of him back to back before seeing a PH pop again.

But as far as pop system itself. it is the same pattern for all of them that have 2 copies.

There are few outliers in the system: Panzer Percival has only 1 NM copy, but 2 PH with a restriction that only 1 of his PH is up at a time. Despot in sky is unique in that his PH can be any one of several of those wall-dolls. Everything else falls into this group pattern of 1 PH per group per NM copy and can pop anytime their PH could have repopped. I suspect that unlike us were we decide the pop on mobs death/depop, SE is deciding at pop time which one to do so its "nm not in cooldown and ph not alive and nm not alive, so roll for nm spawn chance" and most of them just don't _have_ a cooldown. Some provably do (but i can't personally prove the example I am about to use). Yags in giddeus def have cooldowns and a sort of mini window during which they WILL pop if you have the timing down. if you miss that window then you are just rolling dice till they do.

I got interrupted like 9 times trying to compose this by my day-job suddenly having insane emergencies, I hope its actually coherent. Also take my words with a grain of salt I can't prove any of this and we don't implement based on shitTeoSays anymore than we do shitWikiSays 😉

cocosolos:
>  I suspect that unlike us were we decide the pop on mobs death/depop, SE is deciding at pop time

This makes perfect sense especially if you consider "spawn points" as their own entity (which I think is the goal of #3938, though I haven't looked too closely).

TeoTwawki:
Interview with an SE guy confirmed they have node points and that he had to place all the mobs points manually one at a time. Sounds like a hellish experience.

PR #5573, opened 2024-05-01
[LUA] Fix RSE ammo drop chances are incorrect

ampitere:
**_I affirm:_**

Currently the RSE ammo NMs are set with a drop rate of 5 (UNCOMMON in the drop_rate enum) but the method it's using to add the item to the pool is addItemFixed which should be using the values in RATE_PERCENTAGES (lua_loot.cpp). It's also noted in the example for the addItemFixed method: `loot:addItemFixed(xi.items.ANCIENT_BEASTCOIN, 500, 2);`.

1. Reference https://www.pyogenes.com/ffxi/timer/v2.html to figure out rotation
2. Login with the proper character race and gender
3. Zone into the proper RSE zone (Gusgen Mines, Maze of Shakhrami, or Ordelle's Caves)
4. !npchere (17580347 Mines, 17588711 Maze, 17568147 Caves)
5. Repeat as necessary or temporarily set the chance to 1000 (guaranteed)

If there's a better way to do this please let me know!

image

### DIFF
--- a/scripts/zones/Gusgen_Mines/npcs/qm1.lua
+++ b/scripts/zones/Gusgen_Mines/npcs/qm1.lua
@@ -27,7 +27,7 @@ entity.onTrigger = function(player, npc)
 
         local item = 18246 + playerRace - raceOffset
         GetMobByID(ID.mob.AROMA_FLY):addListener('ITEM_DROPS', 'ITEM_DROPS_RSE', function(mob, loot)
-            loot:addItemFixed(item, xi.drop_rate.UNCOMMON)
+            loot:addItem(item, xi.drop_rate.UNCOMMON)
         end)
 
         local newSpawn = math.random(1, 3) -- determine new spawn point for ???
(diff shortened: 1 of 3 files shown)


lapislosh:
This was actually fixed in ASB but never made it back. It used a different function rather than changing the rate enum: https://github.com/AirSkyBoat/AirSkyBoat/pull/3539/files

ampitere:
> This was actually fixed in ASB but never made it back. It used a different function rather than changing the rate enum: https://github.com/AirSkyBoat/AirSkyBoat/pull/3539/files

That works for me, would be nice to convert Golden-Tongued Culberry's drop as well but it doesn't seem like the drop_rate enum supports 50% chance drop rates.

TeoTwawki:
> That works for me, would be nice to convert Golden-Tongued Culberry's drop as well but it doesn't seem like the drop_rate enum supports 50% chance drop rates.

he gets his drop rate out of a localvar set by the trade anyway, no reason to convert

WinterSolstice8 (approved):
approved; I wonder if we should make a warning if you use AddItemFixed and happen to use the xi.drop_rate enums that you may not want to use that function?

Xaver-DaRed (approved):

PR #5576, opened 2024-05-01
[PoC] [Rework] Promyvion-Dem zone

Xaver-DaRed:
**_I affirm:_**

Zone:
- Detach Promyvion-Dem from promyvion global.
- Shift-proof zone.

Portals:
- Make floor 4 return portal account for player original entrance.
- Properly choose portals, by groups.
- Make portals open on memory receptacles APPARENT despawn.

Memory receptacle:
- Add logic for stray spawn when idle.
- Add logic for hide/spawn decorations, and ensure they fade-out and fade-in properly.
- Correct Receptacle subAnimations.
- Correct and time properly stray-spawn animation from Receptacle.
- Give regain to memory receptacles. The math beind it comes from the fact The timing between each mobskill when left alone varied in minutes From every 1 to 3 minutes with very minor second variations.
- Change memory receptacle status to 6 (used in retail) 2 seconds after death, mimicing retails "fast despawn" of them.

Strays:
- Strays will now link PROPERLY to they Memory Receptacle.
  - Strays poped while roaming, will only link if in range.
  - Strays poped mid-battle will spawn agressive.
- Strays will no longer respawn. Now they will only pop from receptacles.
- Strays spawn points have been updated to avoid pos changes shennanigans.

## What this is/isn't
This is a Proof of Concept.

This is not a final product. The final intention isn't to de-globalize promyvion, abd having a bunch of repeated logic all around, but rather, have a zone to test before globalizing this changes.

As you can see from the list of changes, the mechanics were lacking, at best. This is an attempt to make it more retail accurate, based on my own retail time-wasting.

Get to the Spire of Dem walking, with no GM powers.


Xaver-DaRed:
I consider this work mostly finished. I would ask for some1 to pull it and play with it.

Remember the intention is:
- Make promyvion more retail-like.
- To use the logic introduced here to replace the global promyvion currently used in a different PR.

The reason this wasnt done in this PR is so:
- Only this one zone is affected.
- One can compare promyvion dem behavior with any other zone, like a before/after kind of thing.

Xaver-DaRed:
OmicronXI server operator has agreed to test this PR. TY Deme.

jeffnavy14:
I have completed a few runs so far and it feels noticeably more retail-like, 

- Teleports are not insta-spawning onMobDeath.
- Stray's are linking as soon as you strike the Receptacle.
- Animations appear to be correct.

The only issue is some rare occasions the receptacles will not use any abilities my testing method was a naked blm punching it to death slowly. 

After multiple runs so far on different toons and Party setting I and fellow testers have not seen any issues with teleporters not spawning or being able to advance or return to a previous floor. I am going to continue testing and follow up.

claywar:
Superseded by #5671

PR #5581, opened 2024-05-01
[LUA] Add Dirge and Sirvente spells/effects

ampitere:
**_I affirm:_**

Adds Adventurer's Dirge and Foe Sirvente as well as their applicable effects.

1. !changejob 10 99
2. !addallspells
3. Cast Adventurer's Dirge or Foe Sirvente
4. !getmod enmity for Dirge or !getmod enmity_loss_reduction for Sirvente

image

Example above was with +4 equipment bonus. Base enmity is -1.

### DIFF
--- a/scripts/actions/spells/songs/adventurers_dirge.lua
+++ b/scripts/actions/spells/songs/adventurers_dirge.lua
@@ -1,0 +1,15 @@
+-----------------------------------
+-- Spell: Adventurer's Dirge
+-- Reduces target party member's enmity.
+-----------------------------------
+local spellObject = {}
+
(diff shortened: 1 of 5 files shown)


UmeboshiXI:
Just adding a heads up since I recently had a bug report related to this on my server. These two spells also need an item_usable entry so players can use the scrolls to learn the spells in the first place.

ampitere:
> Just adding a heads up since I recently had a bug report related to this on my server. These two spells also need an item_usable entry so players can use the scrolls to learn the spells in the first place.

Good call I didn't think to try that, updated the PR with some item_usable records.

image

WinterSolstice8 (approved):

PR #5582, opened 2024-05-01
[Fix] Alchemy and Goldsmith advanced support npc messages.

Xaver-DaRed:
**_I affirm:_**

- Fixes special cases with alchemy and goldsmithing advanced image support NPCs.
- Documents bitmask used in most guild events (guilds joined, npc conversations and unknowns)
- Discovered what first parameter in aht uhrgan image support npcs actually does (menu options displayed). Adjust accordingly.

Speak with those npcs after joining their guilds. Be able to get image support after the first try.

### DIFF
--- a/scripts/globals/crafting/crafting_image_support.lua
+++ b/scripts/globals/crafting/crafting_image_support.lua
@@ -2,6 +2,7 @@
 -- Image Support NPCs
 -----------------------------------
 require('scripts/globals/crafting/crafting_utils')
+require('scripts/globals/utils')
 -----------------------------------
 xi = xi or {}
 xi.crafting = xi.crafting or {}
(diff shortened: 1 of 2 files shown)


claywar (approved):

scripts/globals/crafting/crafting_image_support.lua
```diff
@@ -114,10 +115,27 @@ xi.crafting.oldImageSupportOnEventFinish = function(player, csid, option, npc)
...
+        -- Calculate Gil cost.
         gilCost = (player:getSkillRank(skillId) + 1) * 30
+
+        -- Handle special initial messages.
+        if
+            guildId == xi.guild.GOLDSMITHING and                          -- Goldsmithing guild.
+            utils.mask.getBit(joinedGuildMask, xi.guild.GOLDSMITHING) and -- Has joined goldsnmithing guild.
+            not utils.mask.getBit(joinedGuildMask, 24)                    -- Has not spoken to npc after joining guild.
+        then
+            player:incrementCharVar('Guild_Member', bit.lshift(1, 24))
```
claywar, line 131: Not a fan of forever charvars, and guild membership could most likely move to a set of uniques in the future.  This is also a bit dangerous to add the bit if anyone follows this pattern without checking first (no issue here though)

PR #5587, opened 2024-05-02
[Fix] Update Slow II and Paralyze II potency

Xaver-DaRed:
**_I affirm:_**

Adds what would have been the old merit bonuses to the spell potency.
Slow II: https://wiki.ffo.jp/html/3454.html
Paralyze II: https://wiki.ffo.jp/html/3453.html

None. Spells slightly stronger.


TeoTwawki (approved):
I hadn't realized this got missed back when they stopped being merit spells

PR #5589, opened 2024-05-02
[cpp] Do not turn entities that have NO_TURN

MowFord:
**_I affirm:_**

Recent change to make entities turn towards their target when readying a skill that has a readytime > 0 neglected to consider the mob behavior `NO_TURN`. Fafnir is a good example (funnier example is [Dark Ixion](https://github.com/LandSandBoat/server/pull/4868), where the logic turns him away to do a rear kick...)

This PR has no change if the entity performing a JA or mobskill is not a mob, and if it is a mob, make sure their behavior allows them to turn

Feed tp to fafnir and watch him not turn on mobskills

~~set a wyvern to behavior no turn `!exec target:setBehaviour(xi.behavior.NO_TURN)` and watch it _not_ turn doing breaths after a weaponskill~~

Wyvern can't have behavior set, and i can't find a way to make an ai-controlled entity use a job ability, but I assume it works just as well as the mobskill fix

### DIFF
--- a/src/map/ai/states/ability_state.cpp
+++ b/src/map/ai/states/ability_state.cpp
@@ -69,9 +69,13 @@ CAbilityState::CAbilityState(CBattleEntity* PEntity, uint16 targid, uint16 abili
         m_PEntity->PAI->EventHandler.triggerListener("ABILITY_START", CLuaBaseEntity(m_PEntity), CLuaAbility(PAbility));
 
         // face toward target
-        m_PEntity->loc.p.rotation = worldAngle(m_PEntity->loc.p, PTarget->loc.p);
-        m_PEntity->updatemask |= UPDATE_POS;
-        m_PEntity->loc.zone->UpdateEntityPacket(m_PEntity, ENTITY_UPDATE, UPDATE_POS);
+        auto* PMob = dynamic_cast<CMobEntity*>(m_PEntity);
+        if (!PMob || !(PMob->m_Behaviour & BEHAVIOUR_NO_TURN))
+        {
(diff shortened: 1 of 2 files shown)


TeoTwawki:
lsb team was actually just talking about this before this opened

I think we actually do want to just respect that flag for now, but at some point we're going to have to check which individual mobskills do/don't. An example I just used was a dragon (not wyrm) doing petro eyes - you can be behind it the whole fight and it'll face you to do that gaze attack, then resume its no-turn behavior.

MowFord:
> lsb team was actually just talking about this before this opened
> 
> I think we actually do want to just respect that flag for now, but at some point we're going to have to check which individual mobskills do/don't. An example I just used was a dragon (not wyrm) doing petro eyes - you can be behind it the whole fight and it'll face you to do that gaze attack, then resume its no-turn behavior.

Fair enough, a flag on the skill I guess

TeoTwawki:
> > lsb team was actually just talking about this before this opened
> > I think we actually do want to just respect that flag for now, but at some point we're going to have to check which individual mobskills do/don't. An example I just used was a dragon (not wyrm) doing petro eyes - you can be behind it the whole fight and it'll face you to do that gaze attack, then resume its no-turn behavior.
> 
> Fair enough, a flag on the skill I guess

eventually thats likely to happen. in the meantime we'll at least respect `NO_TURN`

WinterSolstice8:
I have a plan to fix this cleaner (we clearly have too much copy pasted logic)
i'll fix it after I get home from work

MowFord:
I'll let you close it out when you supersede in case someone comes around looking for a solution

WinterSolstice8:
Superceded by #5591

PR #5592, opened 2024-05-03
[sql/lua] fix A_THIEF_IN_NORG entry req and completion

MowFord:
**_I affirm:_**

First, simple error in the entry req. can easily check the function logic works by running the exec manually when on the quest:
`xi.quest.getVar(player, xi.questLog.OUTLANDS, xi.quest.id.outlands.A_THIEF_IN_NORG, 'Prog')`

Next, the `Gaki` mob is a smn, with `Ayakashi` as an astral flow pet and `Onibi` as a spirit pact. The spawn positions made battlefield 1 impossible to complete due to the spirit being far far away. updated spawn point of spirit and carbuncle to be near the smn

Updated the zone table mob entry to be the table of Gaki mobs to clean up the look of the battlefield lua. 

This made things all wacky. I realized that the order of the table was bad, due to no sort on the query:

image

updated that so that index 1 is reliably the lowest id

`!togglegm` to get right into the battlefield, or do the legwork to test the entry req fix by changing job to sam, completiong prereq, bouncing around to get the 6 required actions (talk to npcs, zone in, etc)

confirm 3 mobs and spirit spawn, use `!exec target:useMobAbility(xi.jsa.ASTRAL_FLOW)` to see the carbuncle join the fray


Frankie-hz:
Table IDs look good 👍 

Weird that query orders them correctly for me. We have the same issue with the auction house sell dates its backwards for us on Horizon but it works for everyone else and locally as well.

MowFord:
> Table IDs look good 👍
> 
> Weird that query orders them correctly for me. We have the same issue with the auction house sell dates its backwards for us on Horizon but it works for everyone else and locally as well.

Likely an issue with modules delete/inserting objects. Mariadb gives no guarantee of order, we just get lucky that it returns in the order we insert :)

scripts/battlefields/Waughroon_Shrine/thief_in_norg.lua
```diff
@@ -30,24 +30,24 @@ content.groups =
         mobIds =
         {
             {
-                waughroonID.mob.GAKI,
-                waughroonID.mob.GAKI + 1,
-                waughroonID.mob.GAKI + 2,
-                waughroonID.mob.GAKI + 3,
+                waughroonID.mob.GAKI[1],
```
claywar, line 33: I'm not understanding what you're trying to do here by storing all IDs for Gaki, and this pattern won't solve clarity for other BCNMs (See: A Worm's Turn, Steamed Sprouts SKCNM's)
 
INSERT INTO `mob_spawn_points` VALUES (17367059,'Gaki','Gaki',7,-177.000,60.000,-144.000,130);
INSERT INTO `mob_spawn_points` VALUES (17367060,'Rasetsu','Rasetsu',8,-177.000,60.000,-142.000,130);
INSERT INTO `mob_spawn_points` VALUES (17367061,'Onki','Onki',9,-177.000,60.000,-140.000,130);
INSERT INTO `mob_spawn_points` VALUES (17367062,'Onibi','Onibi',10,1.000,1.000,1.000,130);
INSERT INTO `mob_spawn_points` VALUES (17367063,'Ayakashi','Ayakashi',11,0.000,0.000,0.000,0);

The issue here is that I had the wrong offset in the pet group, so just need to update the order

sql/mob_spawn_points.sql
```diff
@@ -51631,18 +51631,18 @@ INSERT INTO `mob_spawn_points` VALUES (17367058,'GoBha_Slaughterer','Go\'Bha Sla
...
-INSERT INTO `mob_spawn_points` VALUES (17367062,'Onibi','Onibi',10,1.000,1.000,1.000,130);
-INSERT INTO `mob_spawn_points` VALUES (17367063,'Ayakashi','Ayakashi',11,0.000,0.000,0.000,0);
+INSERT INTO `mob_spawn_points` VALUES (17367062,'Onibi','Onibi',10,-177.000,60.000,-140.000,130);
+INSERT INTO `mob_spawn_points` VALUES (17367063,'Ayakashi','Ayakashi',11,-177.000,60.000,-140.000,130);
 INSERT INTO `mob_spawn_points` VALUES (17367064,'Gaki','Gaki',7,23.274,0.000,15.964,130);
 INSERT INTO `mob_spawn_points` VALUES (17367065,'Rasetsu','Rasetsu',8,23.274,0.000,17.964,130);
 INSERT INTO `mob_spawn_points` VALUES (17367066,'Onki','Onki',9,23.274,0.000,19.964,130);
-INSERT INTO `mob_spawn_points` VALUES (17367067,'Onibi','Onibi',10,201.274,-59.000,160.964,130);
-INSERT INTO `mob_spawn_points` VALUES (17367068,'Ayakashi','Ayakashi',11,0.000,0.000,0.000,0);
+INSERT INTO `mob_spawn_points` VALUES (17367067,'Onibi','Onibi',10,23.274,0.000,19.964,130);
```
claywar, line 51639: Your PR seems to mention only battlefield 1, but it looks like here all of them are missing entries? Are there any reference caps?

MowFord: No I just noticed they were zero but were listed mobids in the battlefield

These are the avatars for onki, they only come out when he uses astral flow which is I assume why they didn't have positions

scripts/battlefields/Waughroon_Shrine/thief_in_norg.lua
MowFord: I thought it was simpler to read/audit having the table of initial mob IDs for each battlefield, I don't think the offsets were wrong at all. 

The bigger issue I was having is that get first id for gaki was returning the third mob, throwing everything off

MowFord:
Sorry for the confusion, but i'm breaking out the cpp fix into a separate PR. Didn't mean to have so many things going on in this PR... was a bit of tunnel vision trying to fix waughroon shrine last night

scripts/battlefields/Waughroon_Shrine/thief_in_norg.lua
claywar: So this really doesn't fix the problem.  The issue is that the pet is in position 4 out of 5 mobs, as opposed to the end, where I made the assumption and was incorrect.  This solution (and I'm not a fan of using gettable here), would make the spirit a required mob to kill, but one of the others not.

I have a branch up that corrects the BCNM itself, but am working to capture the actual mob positions for all three areas currently.

MowFord: If your branch fixes the entryreq I'm good to just close this PR and let you handle it.

MowFord: also fyi making the spirit required is fine because it despawns when the owner dies

claywar: Sounds good, but just a note if converting future BCNMs: Not all pets are really "pets" with the way we currently handle things.  In this case it still is and spawned, but mob_pets always attempts to spawn the pet on master spawn, which is not always correct.  In those cases, we currently need to separate that relationship and handle the accordingly (see: AAGK, AAMR, Divine Might currently)

PR #5593, opened 2024-05-03
[cpp] Ensure sequential order for getfirstID etc

MowFord:
**_I affirm:_**

As mentioned [in this PR](https://github.com/LandSandBoat/server/pull/5592), the IDs.lua helper functions rely on pulling the list of sequential npc and mobs from a zone to do shift-proof mappings. The query in luautils was not ensuring proper order, and relying on the results to always be in the order of insertion. There are scenarios where mariadb does not return those results in order of smallest to greatest ID (modules doing inserts/deletes would be the most obvious, but i'm sure there are others)

Compile with this change, see that `GetFirstID` and `GetTableOfIDs` return what you would expect in terms of the objects being in order of smallest IDs first

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -872,7 +872,7 @@ namespace luautils
         // Mobs
         {
             auto query = fmt::sprintf("SELECT mobname, mobid FROM mob_spawn_points "
-                                      "WHERE ((mobid >> 12) & 0xFFF) = %i",
+                                      "WHERE ((mobid >> 12) & 0xFFF) = %i ORDER BY mobid ASC",
                                       zoneId);
             auto ret   = _sql->Query(query.c_str());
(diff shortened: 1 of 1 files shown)


zach2good (commented):
SQL keywords in caps please, to help differentiate them from args like `mobid`

zach2good:
Also, add `ASC` or `DESC` to the `ORDER BY` for clarity

MowFord:
> Also, add `ASC` or `DESC` to the `ORDER BY` for clarity

aww no fair I was doing that anyway!

zach2good (approved):

zach2good:
https://learn.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver16

`The order in which rows are returned in a result set are not guaranteed unless an ORDER BY clause is specified.`

Huh, TIL, makes sense

Xaver-DaRed (approved):

PR #5595, opened 2024-05-03
[lua/sql] Correct Diabolos avatar animations and migrate to petskills

MowFord:
**_I affirm:_**

Finishing the work of @N3ckB3ard ([re: these PRs](https://github.com/LandSandBoat/server/pull/5478/files)), This is the last smn avatar that is coded I believe

Commits by NB:
- Petskill animations, etc all pulled from captures
- abilities updated to fit the new format for petskill consumption

Commits by me:
- updated all diabolos pet lua files to use new helper functions
- confirmed existing petskills still work
- put notes in `pet_skills.sql` for ones that don't have a lua (and made [an issue](https://github.com/LandSandBoat/server/issues/5594))
- confirmed ranges of player abilities and petskills using the color of the reticle to determine max range. Examples:
  - Matched buffs to other petskills to have range of 18
  - stun pet, dream shroud, selecting myself for DS or noctoshield gives a red reticle at 20 yalms
  - unstun pet, select nether blast, move around to see the color change based on the pet's distance from the mob
    - camisodo has slightly longer range than somnolence
- Note that the abilities have a range of 15, which means the player must get within 15 of the target, then the range of the pet to the target is checked when the skill goes off

Load it up, use the abilities, see proper behavior/messages:

Dream shroud has a special set of message ids:
image
image

Somnolence wasn't getting the final adjustments (which apply magic burst message now), so that was added
image


MowFord:
Sorry about the multiple pushes. Had reworked the enum name for Dream Shroud and found that i had a few issues in the dream_shroud.lua during final testing. 

PR is ready now

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5598, opened 2024-05-03
Fix units of spell interruption check

MowFord:
**_I affirm:_**

Spell Interruption units for the check are out of 100, then the final comparison assumes it's float from 0 to 1

This PR fixes the units to be able to avoid spell interruption again

Before change:
image

After change (with `!setmod spellinterrupt 50`):
an interrupt
image

also getting hit it's possible to not be interrupted


Xaver-DaRed:
This was superseeded by #5601

PR #5603, opened 2024-05-03
Fix underflow for slow mobs and negative mobspeedmod

MowFord:
**_I affirm:_**

Troubleshooting Royal Jelly, i ran across a pretty silly bug. My local has `MOB_SPEED_MOD = -10` and the princess jellies have speed of 2... so they were underflow zooming to the center

Set map setting for mob_speed_mod negative, give a mob a speed such that this would be negative.

before:
https://imgur.com/f5eao3Q

after:
https://imgur.com/OmgTfXv

### DIFF
--- a/settings/default/map.lua
+++ b/settings/default/map.lua
@@ -112,7 +112,7 @@ xi.settings.map =
     -- Note retail treats the mounted speed as double what it actually is.
     MOUNT_SPEED_MOD = 0,
 
-    -- Modifier to apply to agro'd monster speed. 0 is the retail accurate default. Negative numbers will reduce it.
+    -- This is an integer percentage. Modifier to apply to agro'd monster speed (i.e. while engaged in combat). 0 is the retail accurate default. Negative numbers will reduce ALL mobs's speed.
     MOB_SPEED_MOD = 0,
 
     -- Allows you to manipulate the constant multiplier in the skill-up rate formulas, having a potent effect on skill-up rates.
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -580,7 +580,13 @@ float CPathFind::GetRealSpeed()
         }
         else if (m_POwner->animation == ANIMATION_ATTACK)
         {
-            realSpeed = realSpeed + settings::get<int8>("map.MOB_SPEED_MOD");
+            auto speedMod = settings::get<int8>("map.MOB_SPEED_MOD");
+            if (speedMod < -90)
+            {
+                speedMod = -90;
+            }
+
+            realSpeed *= 1.0f + speedMod / 100.0f;
         }
     }
 


src/map/ai/helpers/pathfind.cpp
```diff
@@ -580,7 +580,16 @@ float CPathFind::GetRealSpeed()
...
+            auto speedMod = settings::get<int8>("map.MOB_SPEED_MOD");
+            if (speedMod + realSpeed > 0)
+            {
+                realSpeed = realSpeed + speedMod;
+            }
+            else
+            {
+                // don't let speed underflow due to extreme map setting or low base mob speed
+                realSpeed = 1;
+            }
```
TeoTwawki: speed can be exactly zero in some legitimate cases. your conditionals handled > than zero and then in all other cases sets to exactly one.

MowFord: I want to say speed zero would never get to this function, but I'll do some more testing

MowFord: After discussing with Teo, I decided to convert this map setting to a %. The logic is so much simpler and it honestly makes more sense since the global `SPEED_MOD` is raw for all entities. 

Cool thing I learned: map settings file reloads go into effect real-time. You can set `MOB_SPEED_MOD` to `-100`, pull a mob, then save different values to watch the mob change run speed

settings/default/map.lua
```diff
@@ -105,14 +105,14 @@ xi.settings.map =
     -- Enable/disable jobs other than BST and RNG having widescan
     ALL_JOBS_WIDESCAN = true,
 
-    -- Modifier to apply to player speed. 0 is the retail accurate default. Negative numbers will reduce it.
+    -- Modifier to apply to player speed. 0 is the retail accurate default. Negative numbers will reduce it. This affects all entity base speed as well as max speed.
```
TeoTwawki: your addition to the comment technically true but we overwrite that almost instantly, so in the end this will not influence the final result of objects that have their own separate speed settings here, and I think also will fail to influence pets and trusts. I left some commenting about how bizarrely set up that was in the core code, that everything "begins" with the same base entity speed and is assumed a player until the sql loads happen for mobs and and NPCs and then poof that number is gone. As much as I wanted that to not be the case, I don't think Xaver's speed rewrite tackled that aspect and if it had, this would then be fully not true.

tldr this is just going to confuse people about something that isn't supposed to happen.

settings/default/map.lua
```diff
@@ -105,14 +105,14 @@ xi.settings.map =
...
     SPEED_MOD = 0,
 
     -- Modifier to apply to mount speed. 0 is the retail accurate default. Negative numbers will reduce it.
     -- Note retail treats the mounted speed as double what it actually is.
     MOUNT_SPEED_MOD = 0,
 
-    -- Modifier to apply to agro'd monster speed. 0 is the retail accurate default. Negative numbers will reduce it.
+    -- Modifier to apply to agro'd monster speed (i.e. while engaged in combat). 0 is the retail accurate default. This is a percentage.
```
TeoTwawki: if its a percent we should put a float value here and just give it a range starting at 1.00, making 0.5 half speed and 2.0 double speed. I just say that because that is consistent with other % adjusters in the settings and help visually differentiate it from the 2 speed settings that do not work that way.

settings/default/map.lua
MowFord: oh interesting, good point. I wasn't thinkg that yea mob speeds are defined in the db explicitly

settings/default/map.lua
TeoTwawki: for github readers: mow and I keep being naughty and discussing things in discord and not github. revolving this current discussion.

TeoTwawki (approved):
good 'nuff methinks

PR #5615, opened 2024-05-04
[lua] Correct isValidEntry logic to trade for registration

### LINKED ISSUE #5613: 🐛 Central Apollyon entry
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

I don't see what would be causing it, but trading to get entry into central apollyon just spits out the default onTrigger message:

image

I'm not very well-versed in troubleshooting these new battlefield files if the error isn't obvious in the code, sorry. Any help would be appreciated.

Best i can tell the requirements are correct, battlefield index, etc

MowFord:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/5613 .

The `isValidEntry` logic was referencing the battlefield `entryNpc` value (which never explicitly gets set, it gets passed and used to build a table of one entry for `entryNpcs`) instead of the traded npc's name

Give yourself all the key items and items and trade the 4 chips to the swirling vortex

### DIFF
--- a/scripts/battlefields/Apollyon/central_apollyon.lua
+++ b/scripts/battlefields/Apollyon/central_apollyon.lua
@@ -19,15 +19,11 @@ local content = Limbus:new({
     timeLimit        = utils.minutes(30),
     index            = 4,
     area             = 5,
-    entryNpc         = '_12i',
+    entryNpcs        = { '_12i', '_127' },
     requiredKeyItems = { xi.ki.COSMO_CLEANSE, { xi.ki.RED_CARD, xi.ki.BLACK_CARD }, message = ID.text.YOU_INSERT_THE_CARD_POLISHED },
     requiredItems    = { xi.item.SMALT_CHIP, xi.item.SMOKY_CHIP, xi.item.CHARCOAL_CHIP, xi.item.MAGENTA_CHIP },
     name             = 'CENTRAL_APOLLYON',
 })
-
-function content:isValidEntry(player, npc)
-    return self.entryNpc == '_12i' or self.entryNpc == '_127'
-end
 
 function content:onEntryEventUpdate(player, csid, option, npc)
     if Battlefield.onEntryEventUpdate(self, player, csid, option, npc) then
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (dismissed):

WinterSolstice8 (approved):
Makes sense, apollyon has multiple entrances for this

scripts/battlefields/Apollyon/central_apollyon.lua
```diff
@@ -26,7 +26,7 @@ local content = Limbus:new({
 })
 
 function content:isValidEntry(player, npc)
-    return self.entryNpc == '_12i' or self.entryNpc == '_127'
+    return npc and (npc:getName() == '_12i' or npc:getName() == '_127')
```
claywar:
```
for _, npcName in self.entryNpcs do
    if npc:getName == npcName then
        return true
    end

    return false
end
```

claywar: entryNpcs is a table, entryNpc on construct is now converted to a table, so this was a miss on my part

claywar: Also set entryNpcs in the constructor to { '_12i, '_127' }

claywar: Most likely now that I think of it, you won't even need the entry function in this BCNM

MowFord: Got it, so for both we just need to remove the isvalidentry override and defined entryNpcs in their constructors?

claywar: I believe so.  Sorry for the mixed info as I was thinking through things, but yeah, that should work on its own.  Added support for multiple entrances with Divine Might, so the same logic should work here I think

MowFord: Yep, simpler logic and it works.

Really love the new battlefield content structure. Easy to module changes, cleaner format, more things in one place

WinterSolstice8 (approved):
A lot more elegant

PR #5621, opened 2024-05-05
[LUA] Fix Stegotaur and Teratotaur resetting fomor hate

ampitere:
**_I affirm:_**

Local vars cannot be stored in negative numbers, -4 currently translates to 4294967292 when checked in the fomor hate mixin. The mixin handles non-family types and makes the adjustment negative already so we just need to remove the negative from the local var set on the Stegotaur and Teratotaur.

1. !pos 100 -2 -41 28
2. !setplayervar name FOMOR_HATE 60
3. Kill Teratotaur or Stegotaur
4. !checkvar FOMOR_HATE

image

### DIFF
--- a/scripts/zones/Sacrarium/mobs/Stegotaur.lua
+++ b/scripts/zones/Sacrarium/mobs/Stegotaur.lua
@@ -7,7 +7,7 @@ mixins = { require('scripts/mixins/fomor_hate') }
 local entity = {}
 
 entity.onMobSpawn = function(mob)
-    mob:setLocalVar('fomorHateAdj', -4)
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (commented):
The way we handle this makes no sense at all.

In any case, could you add a comment next to those lines, to indicate those values are substracted elsewhere?

ampitere:
> The way we handle this makes no sense at all.
> 
> In any case, could you add a comment next to those lines, to indicate those values are substracted elsewhere?

Added comments that the values are treated as negative in the fomor_hate mixin.

Xaver-DaRed (approved):
TY sir

EpicTaru (approved):

PR #5625, opened 2024-05-05
[lua/sql] Sarcopsylla NM

MowFord:
**_I affirm:_**

Implements retail accuracy for [Sarcopsylla](https://www.bg-wiki.com/ffxi/Sarcopsylla)

Goto the mob in grauberg, fight it and see that `!getstats` shows proper data, casts the right set of spells, and it doesn't instagib from JA/crit/ws


scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua
```diff
@@ -1,17 +1,19 @@
...
 -- Area: Grauberg [S]
 --   NM: Sarcopsylla
+-- https://www.bg-wiki.com/ffxi/Sarcopsylla
 -----------------------------------
-mixins = { require('scripts/mixins/families/chigoe') }
+mixins = { require('scripts/mixins/families/chigoe_nm') }
 -----------------------------------
 local entity = {}
 
 entity.onMobInitialize = function(mob)
```
Comment (WinterSolstice8) [inline scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: can you add a `TODO: check immunites`
bg wiki claims it has some

Comment (MowFord) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: We added this nm's immunities to the mob pool, I'll confirm when I rebase tomorrow

Comment (WinterSolstice8) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: Oh right, I forgot you can do it through that too

Comment (MowFord) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: Yea group defines: 65559 = 65536 + 16 + 4 + 2 + 1 (sleep, grav, bind, silence, petrify)

Would you prefer we use the `addImmunity` binding in the mob init moving forward (or convert this one)? Definitely seems cleaner and easier to review

Comment (Xaver-DaRed) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: I dunnow if that binding works properly. But if it does, I would definatelly prefer to use that. It would at least make it clear whats going on

Comment (MowFord) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: Seems like a clean solution for NMs only. Mob groups for regular mobs should probably keep the immunity flags in the db

I could add to the `!getstats` command to list the mob's immunities though

Comment (TeoTwawki) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: I don't remember where I was talking about it but i too would like to move forward with doing immunities lua side where possible/non painful to do so and tackle any issues in the way of that. I hope to someday retire the column altogether, but to do that we need ways of auto applying across multiple mobs in some cases and we ain't there yet.

Regarding `!getstats`, its got a lot in there and a lot more we'd all like to see in there. I'd like to see _which_ stats we're getting broken up into separate parameters passed into the command so that we aren't throwing multiple screen-full amounts of information at the client and potentially r0'ing the user or scrolling off useful log information before it can be read. 

Just thinking aloud here but: `!getstats basestats` producing hp/mp/str/dex/etc `!getstats derived` producing attack defense etc `!getstats special` for things like treasure hunter tier currently applied, battle timer value, rage timer, etc `!getstats resist` for ifno about its current resistances and immunities. There could be diff organization or parameter names, just some way of breaking up the info is desired before the amount of it it becomes a problem.

Comment (MowFord) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: > !getstats basestats...

I like it. I'll make a PR to clean that command up. It's very nice having one command to get details about a mob, but it does have a gigantic output

As to this mob. It has the right immunities in the db but afaik the immunity bindings work. I suggest a process of setting immunity to 0 in the db and explicitly defining them in the mob init for NMs

Comment (TeoTwawki) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: as long as we have lua functions for adding and removing immunities (there actually are a few legit cases for remove, in addition to testing ability) and those function do in fact work as expected, I am open to pr's that start converting existing pools to 0 and moving the immunity to the mob lua, as well as a policy of using 0 in the pools going forward. once there are no non-zero left in the pool table, the column can die.

@WinterSolstice8 @Xaver-DaRed  @claywar  @zach2good we all in agreement?

Comment (Xaver-DaRed) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: I agree.

I would add to that one more thing related to this:

Right now we have
- SLEEP
- SLEEP_LIGHT
- SLEEP_DARK

We dont need all 3, at all. SLEEP_LIGHT + SLEEP_DARK == SLEEP.

I would propose to also start normalizing using DARK and LIGHT and 0ing regular sleep, for a future removal/substitution. We are missing ADDLE, for instance, iirc.

I know the cases where one mob is trully immune to 1 without the other can be counted with one hand, but still, I dont think that justifies having 3 sleeps.

Comment (TeoTwawki) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: @Xaver-DaRed 

`SLEEP_LIGHT` and `SLEEP_DARK` were added for server custom bs purposes so they could separate them without polluting upstream while we upstream only ever used `SLEEP`, because at that time upstream did not belive any retail mob could be immune rather than highly resistant to just one of the 2

Comment (Xaver-DaRed) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: I remember. Still, now that we know the cases exist, we should just embrace those 2 and say bye bye to the old one, couse, again, it aint needed

Comment (MowFord) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: Updated to set immunities via mob init and confirmed it works:
image

Also Xaver requested I go ahead and split out sleep immunities now:
image

Comment (TeoTwawki) [reply scripts/zones/Grauberg_[S]/mobs/Sarcopsylla.lua:10]: > Still, now that we know the cases exist

murder it is then!

TeoTwawki (approved):

PR #5628, opened 2024-05-05
[core] Allow targetfind to reject matching allegiances for ENEMY search

### LINKED ISSUE #5618: 🐛 [CRASH] Server crashes when a player gets charmed while a mob is readying Sweep
**_I affirm:_**

Windows 11

## Branch affected by issue 

`base`

## Steps to reproduce
I have not found any other skill or combination this works with. We have been getting occasional silent crashes in Dynamis for some time now and its due to a player getting charmed during a readying sweep skill. The best way to reproduce this is this:

You can either go into Dynamis Windurst or Castle O.
Aggro a bunch of yagudo
For a guaranteed crash you need to turn off all the other mobskills other than sweep.
`!setmod regain 3000` on all the mobs that are attacking you except one
Use `!exec target:useMobAbility(710)` on the mob that does not have 3k regain. This will replicate a charmed 2hr from a beastmaster that happens in Dynamis Windurst.
See crash.

image

## Expected behavior
Not crash

WinterSolstice8:
**_I affirm:_**

This fixes an issue where due to some odd conditions Yagudo could sweep themselves, stun themselves, and break their states causing a crash.
Fixes #5618 

See #5618
Also test bomb toss suicide `!exec target:useMobAbility(592)` and see it works
Test moonlight, curaga, firaga, see them all work.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -531,7 +531,7 @@ void CMobController::CastSpell(SpellID spellid)
                 {
                     // chance to target party
                     PMob->PAI->TargetFind->reset();
-                    PMob->PAI->TargetFind->findWithinArea(PMob, AOE_RADIUS::ATTACKER, PSpell->getRange());
+                    PMob->PAI->TargetFind->findWithinArea(PMob, AOE_RADIUS::ATTACKER, PSpell->getRange(), FINDFLAGS_NONE, TARGET_NONE);
 
                     if (!PMob->PAI->TargetFind->m_targets.empty())
                     {
--- a/src/map/ai/helpers/targetfind.cpp
+++ b/src/map/ai/helpers/targetfind.cpp
@@ -70,9 +70,9 @@ void CTargetFind::reset()
     m_PMasterTarget = nullptr;
 }
 
-void CTargetFind::findSingleTarget(CBattleEntity* PTarget, uint8 flags)
+void CTargetFind::findSingleTarget(CBattleEntity* PTarget, uint8 findFlags, uint16 targetFlags)
 {
-    m_findFlags     = flags;
+    m_findFlags     = findFlags;
     m_zone          = m_PBattleEntity->getZone();
     m_PTarget       = nullptr;
     m_PRadiusAround = &PTarget->loc.p;
@@ -80,12 +80,13 @@ void CTargetFind::findSingleTarget(CBattleEntity* PTarget, uint8 flags)
     addEntity(PTarget, false);
 }
 
-void CTargetFind::findWithinArea(CBattleEntity* PTarget, AOE_RADIUS radiusType, float radius, uint8 flags)
+void CTargetFind::findWithinArea(CBattleEntity* PTarget, AOE_RADIUS radiusType, float radius, uint8 findFlags, uint16 targetFlags)
 {
     TracyZoneScoped;
-    m_findFlags = flags;
-    m_radius    = radius;
-    m_zone      = m_PBattleEntity->getZone();
+    m_findFlags   = findFlags;
+    m_targetFlags = targetFlags;
+    m_radius      = radius;
+    m_zone        = m_PBattleEntity->getZone();
 
     if (radiusType == AOE_RADIUS::ATTACKER)
     {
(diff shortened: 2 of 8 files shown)


Xaver-DaRed (dismissed):

WinterSolstice8:
This touches a lot of stuff, gotta make sure it doesn't break anything, added hold label

zach2good (dismissed):

WinterSolstice8:
Yup looks like some stuff broke because targetflags != findflags. it fixed my exact bug but some AoEs stopped working. Time to add targetflags into targetfind

We already do some jank conversions from targetflags -> findflags, so makes me wonder if it just should have been this way to begin with...

WinterSolstice8:
Alright, go ahead and re-review. had to touch more stuff

src/map/ai/helpers/targetfind.cpp
```diff
@@ -70,22 +70,23 @@ void CTargetFind::reset()
     m_PMasterTarget = nullptr;
 }
 
-void CTargetFind::findSingleTarget(CBattleEntity* PTarget, uint8 flags)
```
TeoTwawki, line 73: 👍 
HATE it when old code doesn't say what _WHAT_ flags are being passed along

TeoTwawki (approved):

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -531,7 +531,7 @@ void CMobController::CastSpell(SpellID spellid)
                 {
                     // chance to target party
                     PMob->PAI->TargetFind->reset();
-                    PMob->PAI->TargetFind->findWithinArea(PMob, AOE_RADIUS::ATTACKER, PSpell->getRange());
+                    PMob->PAI->TargetFind->findWithinArea(PMob, AOE_RADIUS::ATTACKER, PSpell->getRange(), 0, 0);
```
zach2good: Can we use the enum values here instead of 0s please?

FINDFLAGS_NONE etc

zach2good (approved):

src/map/ai/controllers/mob_controller.cpp
WinterSolstice8: there isnt an enum for targettype value 0

WinterSolstice8:
```
enum TARGETTYPE
{
    TARGET_SELF                    = 0x01,
    TARGET_PLAYER_PARTY            = 0x02,
    TARGET_ENEMY                   = 0x04,
    TARGET_PLAYER_ALLIANCE         = 0x08,
    TARGET_PLAYER                  = 0x10,
    TARGET_PLAYER_DEAD             = 0x20,
    TARGET_NPC                     = 0x40, // an npc is a mob that looks like an npc and fights on the side of the character
    TARGET_PLAYER_PARTY_PIANISSIMO = 0x80,
    TARGET_PET                     = 0x100,
    TARGET_PLAYER_PARTY_ENTRUST    = 0x200,
    TARGET_IGNORE_BATTLEID         = 0x400, // Can hit targets that do not have the same battle ID
};
```

zach2good: No reason you can't add `FINDFLAGS_NONE`, rather than piping in a raw number that means the same thing

zach2good: or `TARGET_NONE`, whatever

zach2good: Then the storage and the argument types become the enum:

```
    FIND_TYPE m_findType;
    FINDFLAGS    m_findFlags;   // what to search for
    TARGETTYPE    m_targetFlags; // targetflags to reject potentially bad targets
```

WinterSolstice8: > Then the storage and the argument types become the enum:
> 
> ```
>     FIND_TYPE m_findType;
>     FINDFLAGS    m_findFlags;   // what to search for
>     TARGETTYPE    m_targetFlags; // targetflags to reject potentially bad targets
> ```
Apparently implementing this is editing a _ton_ of stuff and I don't really have time to test every single permutation of spell/ability/mobskill/petskill right now

I've made the other changes asked of, though.

zach2good: Yeah, that one was a "we should probably do this at some point" thing, currently looks good 👍

PR #5632, opened 2024-05-06
[core] Implement supertanking

### LINKED ISSUE #5542: 🔨 Implement supertanking
**_I affirm:_**

## Describe the feature

Supertanking doesn't exist currently because any aggro adds to the hate list WITH initial hate (this is wrong, an actual aggressive action needs to be taken)

WinterSolstice8:
**_I affirm:_**

Implements supertanking
Closes #5542

Do supertank things, like aggro/spawn stuff and see that supertanking works, for example:

[2024-05-05 21-01-54.webm](https://github.com/LandSandBoat/server/assets/60417494/5fca4178-2d89-4673-9049-af2266abb8e4)

### DIFF
--- a/src/map/entities/mobentity.cpp
+++ b/src/map/entities/mobentity.cpp
@@ -1105,6 +1105,10 @@ void CMobEntity::OnEngage(CAttackState& state)
             PPet    = state.GetTarget();
             PTarget = ((CPetEntity*)PTarget)->PMaster;
         }
+
+        // TODO: Supertanking might be effected by this block when we don't want it to be.
+        // Things like Ambuscade "don't have" supertanking, though.
+        // This block apparently only effects rare things like NW apollyon, so might be ok for now.
         if (PTarget->objtype == TYPE_PC)
         {
             // clang-format off
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -202,7 +202,7 @@ void CMobController::TryLink()
     // my pet should help as well
     if (PMob->PPet != nullptr && PMob->PPet->PAI->IsRoaming())
     {
-        ((CMobEntity*)PMob->PPet)->PEnmityContainer->AddBaseEnmity(PTarget);
+        PMob->PPet->PAI->Engage(PTarget->id);
     }
 
     // Handle monster linking if they are close enough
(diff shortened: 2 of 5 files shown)


src/map/entities/mobentity.cpp
```diff
@@ -1105,6 +1105,10 @@ void CMobEntity::OnEngage(CAttackState& state)
             PPet    = state.GetTarget();
             PTarget = ((CPetEntity*)PTarget)->PMaster;
         }
+
+        // TODO: Supertanking might be effected by this block when we don't want it to be.
+        // Things like Ambuscade "don't have" supertanking, though.
+        // This block apparently only effects rare things like NW apollyon, so might be ok for now.
```
cocosolos, line 1111: So this is a pretty bad implementation of alliance hate, where mobs will aggro the rest of the party if its target dies even if they haven't acted at all. I think an example of something that has this _and_ supertanking is Divine Might. The implementation is bad and will need to be changed, but there should be a way to add a player to a mobs "I'm gonna mess you up when I'm done with this guy" list while still allowing supertanking.

WinterSolstice8: In LSB this mod is only used in Apollyon NW on the last floor

cocosolos: It never got widely implemented, mostly because it wasn't a great implementation but worked well enough in Limbus. I did think it was being used in more than just the 1 battlefield though.

WinterSolstice8: Hmm. Maybe I didn't look hard enough?

cocosolos: It looks like it was part of the battlefield mob setup functions in mobutils but was apparently removed probably when that Limbus rewrite went in. I'm surprised NW is still left over. Either way there should exist a way for a mob to have a backup list of targets while still allowing supertanking. Not making any request here and if it was up to me I'd say just rip out any remnants of the existing alli_hate_mobmod. It's kind of obscure anyway and having working supertanking is more important right now.

WinterSolstice8: There's plans in the works to get better enmity systems, stay tuned

cocosolos: Yeah so I looked and this mod isn't even being used uniformly for all the mobs in the zone, just the last floor in NW and the first floor in NE. Weird, it should all just be removed.

Xaver-DaRed (approved):

PR #5638, opened 2024-05-06
[QUEST]Foreman Best Friend xp OR map

hooksta4:
**_I affirm:_**

Updates A foremans best friends quest in Bastok to give EITHER map of gusgen mines OR 2k exp if already has map.

!addquest 1 9
!giveitem dog_collar
If player doesnt have map then will give map. Otherwise will give 2k xp.

### DIFF
--- a/scripts/quests/bastok/A_Foremans_Best_Friend.lua
+++ b/scripts/quests/bastok/A_Foremans_Best_Friend.lua
@@ -9,10 +9,8 @@ local quest = Quest:new(xi.questLog.BASTOK, xi.quest.id.bastok.A_FOREMANS_BEST_F
 
 quest.reward =
 {
-    exp      = 2000,
     fame     = 60,
     fameArea = xi.fameArea.BASTOK,
-    keyItem  = xi.ki.MAP_OF_THE_GUSGEN_MINES,
 }
 
 quest.sections =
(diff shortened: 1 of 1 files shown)


scripts/quests/bastok/A_Foremans_Best_Friend.lua
```diff
@@ -56,8 +54,11 @@ quest.sections =
             onEventFinish =
             {
                 [112] = function(player, csid, option, npc)
-                    if quest:complete(player) then
-                        player:confirmTrade()
+                    player:confirmTrade()
+                    if player:hasKeyItem(xi.ki.MAP_OF_THE_GUSGEN_MINES) then
+                        player:addExp(2000)
+                    else
+                        npcUtil.giveKeyItem(player, xi.ki.MAP_OF_THE_GUSGEN_MINES)
```
Xaver-DaRed: We still want the quest:complete() check, I would think

Or else, the quest would never complete

hooksta4: Solid point.

hooksta4: Tis fixed my friend.

Xaver-DaRed (approved):

PR #5639, opened 2024-05-07
[lua] Adjust accVaries to reduce accuracy not hitrate

MowFord:
**_I affirm:_**

I know there are comments that accVaries is possibly not implemented properly, but the current intention of the param is to reduce accuracy by a percent based on TP:
image

The current code reduces `hitRate` by the percent of `accuracy` (ranged or physical depending on the ws), but it ignores the evasion. I think someone was trying to be clever by utilizing the same divisor as the getHitRate functions `(acc - eva)/200`, but the result is that `hitRate` reduction is only a function of accuracy, meaning the explicit `hitRate` penalty is increased as player accuracy increases.

This PR rearranges the `accVaries` consideration to before `hitRate` is calculated and adjusts the `calcParams.bonusAcc` value.

Note that previous code had a param to not cap `hitRate` at 95% so `accVaries` could be considered later, then clamped `hitRate`. So this new code is functionally the same there.

Ensure dancer steps, desperate flourish, and violent flourish work, as they utilize `xi.weaponskills.getHitRate`

add a print in the `getRangedHitRate` and `xi.weaponskills.getHitRate` functions to view the hitrate and bonusacc to verify things are working as expected.

image
^^ first is the physical first hit's hitrate due to +100 acc. Second is regular hitrate

### DIFF
--- a/scripts/globals/job_utils/dancer.lua
+++ b/scripts/globals/job_utils/dancer.lua
@@ -246,7 +246,7 @@ xi.job_utils.dancer.useStepAbility = function(player, target, ability, action, s
         player:delTP(100 + player:getMod(xi.mod.STEP_TP_CONSUMED))
     end
 
-    if math.random() <= xi.weaponskills.getHitRate(player, target, true, 10 + player:getMod(xi.mod.STEP_ACCURACY)) then
+    if math.random() <= xi.weaponskills.getHitRate(player, target, 10 + player:getMod(xi.mod.STEP_ACCURACY)) then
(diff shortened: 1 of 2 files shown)


WinterSolstice8 (approved):
this acc varies with tp nonsense needs to get figured out. one day it will activate my autism hard enough that i'll research every single one I can
I simply do not believe all of these are accuracy penalties. there just isn't any evidence for it.

WinterSolstice8 (approved):

PR #5640, opened 2024-05-07
[Cleanup] Shift-proof "R" Zones

Frankie-hz:
**_I affirm:_**

Shift-proof the following zones:
Ruhotz Silvermines 
Rolanberry Fields and [S]
Riverne Site A 01 and B 01
Ranguemont Pass
Rala Waterways U
RoMaeve

All lotto NMs, one ID change to a quest.

### DIFF
--- a/scripts/missions/bastok/7_1_The_Final_Image.lua
+++ b/scripts/missions/bastok/7_1_The_Final_Image.lua
@@ -134,7 +134,7 @@ mission.sections =
                             return mission:keyItem(xi.ki.REINFORCED_CERMET)
                         elseif
                             -- TODO: xi.settings default (300s) is used to hide this QM on spawn.  Verify if this is accurate since we're moving the QM
-                            npcUtil.popFromQM(player, npc, { romaeveID.mob.MOKKURKALFI_I, romaeveID.mob.MOKKURKALFI_II }, { claim = false, look = true, radius = 2 })
+                            npcUtil.popFromQM(player, npc, { romaeveID.mob.MOKKURKALFI, romaeveID.mob.MOKKURKALFI + 1 }, { claim = false, look = true, radius = 2 })
                         then
                             local newPosition = npcUtil.pickNewPosition(npc:getID(), romaeveID.npc.BASTOK_7_1_QM_POS, true)
                             npcUtil.queueMove(npc, newPosition)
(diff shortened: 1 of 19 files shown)


Frankie-hz:
For Ranguemont Pass I wasn't sure how to do Taisaijin without making a global.
Same for the qm positions in RoMaeve.

I can make a global file unless anyone has any other ideas I am not seeing.

TeoTwawki:
> For Ranguemont Pass I wasn't sure how to do Taisaijin without making a global. Same for the qm positions in RoMaeve.
> 
> I can make a global file unless anyone has any other ideas I am not seeing.

In the case of Taisaijin ..we should probably have a standardized global way of handling that type of spawn: its a "don't kill my PH for this duration" deal, and there are several NM like that, some with 2 tiers (HQ NM of the NQ NM is left alone even longer) we have lotto pop mechanics globalized in `mobs.lua` I think it'd be worth giving these a similar treatment. That's outside the scope of your current PR'ings of shift proofing though. What you have atm is fine.



🤔 
For "moving" QM (these mostly needed renamed btw, `qm_what_am_I_for_exactly`) NPC we should have a utils thing for both the kind where its just switching one off and another on, and the kind where the same npc id actually is `setPos()`'d ..I think we may have something for the latter style already, not sure if it would fit here. Obv that still needs fed id info though.

It's also fine for some of these to stay in IDs.lua, we just won't keep things there unless we really need to - if its all self contained in one script, by by IDs entry. if we need it in several places? it can stay.

TeoTwawki (approved):

cocosolos:
Idk if there's a util but misareaux coast has a similar qm. It doesn't move around but it does execute some logic at set hours. This one looks like it's supposed to move every 6 hours. If it moves on the hour then a similar approach could be used.

Edit:
Rabbit Footprint in Uleguerand range is probably a better example.

Frankie-hz:
> Idk if there's a util but misareaux coast has a similar qm. It doesn't move around but it does execute some logic at set hours. This one looks like it's supposed to move every 6 hours. If it moves on the hour then a similar approach could be used.
> 
> Edit: Rabbit Footprint in Uleguerand range is probably a better example.

Ok I will make a note for these cases and submit a PR at the end with commits separating them.

PR #5641, opened 2024-05-07
[Cleanup] Shift-proof all "Q" Zones

Frankie-hz:
**_I affirm:_**

Shift-proof the following zones:
Qulun Dome
Quicksand Caves
Qufim Island
QuBia Arena

All Lotto NMs


Frankie-hz:
Same question for the `ANTICAN_TAG_POSITIONS` in quicksands. Do we want a global file?

Edit: Out of scope will fix at the end

Xaver-DaRed (approved):

PR #5653, opened 2024-05-08
[lua] Update !getstats to partition reports

MowFord:
**_I affirm:_**

Per [this thread](https://github.com/LandSandBoat/server/pull/5625#discussion_r1594326764), I've updated the `!getstats` command

~~It now collects all the messages to print into a table and spits them out at the end, to reduce duplicated code~~

Went through a lot of iterations, but suffice to say the help text shows up when you don't choose a proper stat group, and it still prints all the same data, as well as a whole new group of "elements" with resistance data as well as immunities.

image

image

image

### DIFF
--- a/scripts/commands/getstats.lua
+++ b/scripts/commands/getstats.lua
@@ -4,14 +4,21 @@
 -----------------------------------
 local commandObj = {}
 
+local options =
+{
+    [1] = 'base',
+    [2] = 'defensive',
+    [3] = 'offensive',
+    [4] = 'elements',
+}
+
 commandObj.cmdprops =
 {
     permission = 1,
-    parameters = ''
+    parameters = 's'
 }
 
--- function onTrigger(player, extendedMode)
-commandObj.onTrigger = function(player)
+commandObj.onTrigger = function(player, option)
     local target = player:getCursorTarget()
     if target == nil then
         player:printToPlayer('Target something first.')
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

MowFord:
rebased after enum changes

image

scripts/commands/getstats.lua
```diff
@@ -4,14 +4,22 @@
...
+    'defensive',
+    'offensive',
+    'elements',
+}
+
 commandObj.cmdprops =
 {
     permission = 1,
-    parameters = ''
+    parameters = 'i'
```
TeoTwawki: why not just use string parameter and make default no-param result be to tell the user what the potential parameters are and return out (other command use an error handler wrapper) instead of looping your table to print them?

scripts/commands/getstats.lua
```diff
@@ -4,14 +4,22 @@
...
 }
 
 -- function onTrigger(player, extendedMode)
```
TeoTwawki: this line is a remnant of my original intention to have a string param for extended modes (spit out params other than the default) and can now be removed

scripts/commands/getstats.lua
```diff
@@ -25,59 +33,121 @@ commandObj.onTrigger = function(player)
...
+    if options[option] == 'base' then
+        table.insert(messages, string.format('MainJob(jID: %s) LV: %i / SubJob(jID: %s) LV: %i ',
+            target:getMainJob(), target:getMainLvl(), target:getSubJob(), target:getSubLvl()))
+
+        table.insert(messages, string.format('HP: %i/%i  MP: %i/%i (current/max) ',
+            target:getHP(), target:getMaxHP(), target:getMP(), target:getMaxMP()))
+
+        table.insert(messages, string.format('Total STR: %i ', target:getStat(xi.mod.STR)))
+        table.insert(messages, string.format('Total DEX: %i ', target:getStat(xi.mod.DEX)))
+        table.insert(messages, string.format('Total VIT: %i ', target:getStat(xi.mod.VIT)))
```
TeoTwawki: I am not a fan of all this table inserting just to spit out prints while still using the if/else anyway.

scripts/commands/getstats.lua
MowFord: simpler to remember an integer rather than a stat group name

scripts/commands/getstats.lua
MowFord: seemed cleaner than a million printToPlayer calls

TeoTwawki: suggested restructure:
- string param
- table the type instead of the prints (so reverse your options table: `["string"] = number` so you can `get[input]` it)
- on matching conditions, do prints for each of the type print as normal
- but on anything that has to be specific to an entity type, wrap in a check for that entity type. no more `elseif`
- print example of command use and optional params when used without params or invalid params. and return out

scripts/commands/getstats.lua
TeoTwawki: `['string'] = number,` / `[number] = 'string',` and you can have it both ways

scripts/commands/getstats.lua
TeoTwawki: switch statement also acceptable to kill if/else-ness

scripts/commands/getstats.lua
MowFord: Removed

scripts/commands/getstats.lua
MowFord: table `allthethings.gif`

TeoTwawki (approved):

MowFord:
switch:caseof works with a default block, neat

TeoTwawki:
> switch:caseof works with a default block, neat

whats even neater is this is probably the 1st time anyone has used it. which was why I stamped before hand lol.

MowFord:
When cleaning up goes too far, sorry about that. Accidentally deleted the branch

Frankie-hz:
Love this idea

PR #5655, opened 2024-05-08
Remove 'Reraise' on Job Change

CatsEyeXI:
**_I affirm:_**

Properly removes Reraise when changing jobs.
Verified on retail by casting reraise on self as WHM, then using Nomad Moogle in Selbina to change jobs to RDM, effect was removed.

!changejob WHM 99
/ma "Reraise" <me>
Use Nomad Moogle in Selbina to change jobs
Effect will disappear.


TeoTwawki:
wow seriously we were leaving rr on this whole time? I wonder how far back that mistake started.

claywar (approved):

PR #5659, opened 2024-05-08
[lua] Scitalis NM

MowFord:
**_I affirm:_**

Implements [Scitalis](https://www.bg-wiki.com/ffxi/Scitalis)

Most data in sql was fine, so leaving it as one commit.

Simple NM with _brutal_ AE damage

Adjust PH's lua to have
`xi.mob.phOnDespawn(mob, scitalisPHTable, 100, 3600, {immediate=true}) -- 1 hour`

kill it and see NM spawn immediately.

image

### DIFF
--- a/scripts/zones/Grauberg_[S]/mobs/Scitalis.lua
+++ b/scripts/zones/Grauberg_[S]/mobs/Scitalis.lua
@@ -1,8 +1,25 @@
 -----------------------------------
 -- Area: Grauberg [S]
 --   NM: Scitalis
+-- https://www.bg-wiki.com/ffxi/Scitalis
 -----------------------------------
 local entity = {}
+
+entity.onMobInitialize = function(mob)
+    mob:addImmunity(xi.immunity.GRAVITY)
+    mob:addImmunity(xi.immunity.SILENCE)
+    mob:addImmunity(xi.immunity.SLOW)
+    mob:addImmunity(xi.immunity.DARK_SLEEP)
+    mob:addImmunity(xi.immunity.PETRIFY)
+
+    mob:setMod(xi.mod.DOUBLE_ATTACK, 50)
+    mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
+end
+
+entity.onAdditionalEffect = function(mob, target, damage)
+    -- Captures show unresisted damage between 120 and 200. TODO find what causes full power AE to vary so greatly
+    return xi.mob.onAddEffect(mob, target, damage, xi.mob.ae.ENAERO, { power = math.random(165, 190) })
+end
 
 entity.onMobDeath = function(mob, player, optParams)
     xi.hunts.checkHunt(mob, player, 503)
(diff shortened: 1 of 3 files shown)


scripts/zones/Grauberg_[S]/mobs/Scitalis.lua
```diff
@@ -1,9 +1,25 @@
...
+    mob:addImmunity(xi.immunity.SLOW)
+    mob:addImmunity(xi.immunity.DARK_SLEEP)
+    mob:addImmunity(xi.immunity.PETRIFY)
+
+    mob:setMod(xi.mod.DOUBLE_ATTACK, 50)
+    mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
+end
+
+entity.onAdditionalEffect = function(mob, target, damage)
+    return xi.mob.onAddEffect(mob, target, damage, xi.mob.ae.ENAERO, { power = math.random(165, 190) })
```
Comment (TeoTwawki) [inline scripts/zones/Grauberg_[S]/mobs/Scitalis.lua:21]: sources vary greatly on the power of its wind effect, I've seen as low as 120 reported. Put a TODO here to verify the real base power later (we want generally to let resistance functions handle the changes, but right now all the mob AEs are using randomness and haven't been reworked to do this accurately)

TeoTwawki (changes requested):

scripts/zones/Grauberg_[S]/mobs/Scitalis.lua
Comment (MowFord) [reply scripts/zones/Grauberg_[S]/mobs/Scitalis.lua:21]: This was verified via capture last week in retail

Comment (TeoTwawki) [reply scripts/zones/Grauberg_[S]/mobs/Scitalis.lua:21]: values seen one exactly one cap with one set of gear does not equal knowing its exact correct base power unmodified. if you had it down, there would not be any math.random() happening

I went out there my self on 2 diff chars and one got walloped for 200 and then other for less than 140. you  have a random between 165 and 190. SOMETHING is influencing the hit, we dunno what is involved yet. Add the TODO.

Comment (MowFord) [reply scripts/zones/Grauberg_[S]/mobs/Scitalis.lua:21]: fair enough, here's some data that i just gathered from the cap:

image

raw:

image

Comment (TeoTwawki) [reply scripts/zones/Grauberg_[S]/mobs/Scitalis.lua:21]: usually these things are heavily influenced by elemental resist and I thought I had it all figured out especially on items before item levels came along and dumped gobs of meva and magic damage stats onto everything and started resulting in wildly different numbers than 1-75.  _cries in math_

Comment (MowFord) [reply scripts/zones/Grauberg_[S]/mobs/Scitalis.lua:21]: TODO added, btw

TeoTwawki (approved):

claywar:
Looks like a rebase is needed on this one due to recent shift changes

MowFord:
> Looks like a rebase is needed on this one due to recent shift changes

Understood. I'll catch up Monday

Xaver-DaRed (approved):

PR #5661, opened 2024-05-09
Cleanly fail the GetFirstID lookup

### LINKED ISSUE #5567: 🐛 GetFirstID() does not fail gracefully when invalid name is provided
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce
1. Add GetFirstID() Mob entry for "Dark_Dragon" in Horlais Peak (mob does not exist in zone)
2. Start map server

## Expected behavior
1. Invalid mob name should fail gracefully, but not crash

MowFord:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/5567 .

We cleanly exit and return 0 when the lookup doesn't exist in the `unordered_map`

Do the steps in the issue:
image

the table's value ends up being zero:
image

and if we do a name that has multiple entries, see we still return the first:
image

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -901,9 +901,18 @@ namespace luautils
 
         // Update GetFirstID to use this new lookup
         // clang-format off
-        lua.set_function("GetFirstID", [&](std::string const& name) -> uint32
+        lua.set_function("GetFirstID", [&](std::string const& name) -> std::optional<uint32>
         {
-            return lookup[name].front();
+            if (lookup.find(name) != lookup.end())
+            {
+                return lookup[name].front();
+            }
+            else
+            {
+                ShowWarning(fmt::format("GetFirstID({}) in zone {}: Returning nil", name, zoneName));
+
+                return std::nullopt;
+            }
         });
 
         std::unordered_map<std::string, sol::table> idLuaTables;
@@ -926,7 +935,7 @@ namespace luautils
                 // If we have no entries, bail out and return nil
                 if (lookup.find(name) == lookup.end())
                 {
-                    ShowWarning(fmt::format("GetTableOfIDs({}): Returning nil", name));
+                    ShowWarning(fmt::format("GetTableOfIDs({}) in zone {}: Returning nil", name, zoneName));
                     return sol::lua_nil;
                 }
 
@@ -934,7 +943,7 @@ namespace luautils
 
                 if (entriesVec.empty())
                 {
-                    ShowWarning(fmt::format("GetTableOfIDs({}): Returning empty table", name));
+                    ShowWarning(fmt::format("GetTableOfIDs({}) in zone {}: Returning empty table", name, zoneName));
                     return table;
                 }
 
@@ -964,7 +973,7 @@ namespace luautils
 
             if (table.empty())
             {
-                ShowWarning(fmt::format("Tried to look do ID lookup for {}, but found nothing.", name));
+                ShowWarning(fmt::format("GetTableOfIDs({}) in zone {}: Tried to look do ID lookup, but found nothing.", name, zoneName));
             }
 
             // Add to cache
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):

Xaver-DaRed (approved):

src/map/lua/luautils.cpp
```diff
@@ -903,7 +903,16 @@ namespace luautils
...
-            return lookup[name].front();
+            if (lookup.find(name) != lookup.end())
+            {
+                return lookup[name].front();
+            }
+            else
+            {
+                ShowWarning(fmt::format("GetFirstID({}): Returning nil", name));
+
+                return 0;
```
claywar: Nil here should be `sol::lua_nil` if the intent is to return a nil and not '0' value

src/map/lua/luautils.cpp
```diff
@@ -903,7 +903,16 @@ namespace luautils
...
         lua.set_function("GetFirstID", [&](std::string const& name) -> uint32
         {
-            return lookup[name].front();
+            if (lookup.find(name) != lookup.end())
+            {
+                return lookup[name].front();
+            }
+            else
+            {
+                ShowWarning(fmt::format("GetFirstID({}): Returning nil", name));
```
zach2good: This error message could have some more context. Also, it ends up returning 0, not nil.

zach2good: Useful to remember that the only things in Lua that are falsey are `false` and `nil` (`if 0 then print("zero") end` will print, etc.)

src/map/lua/luautils.cpp
MowFord: sorry, copied the debug print from the GetTableOfIDs, this returns zero as the function is expected to return a number. If we return nil it would still prevent a full crash, but break more things like `ID.mob.DEATH_JACKET + 5` (which may be a good thing)

src/map/lua/luautils.cpp
MowFord: Yea probably a good idea to provide the path of the `IDs.lua` that is being processed for both this and `GetTableOfIDs`

src/map/lua/luautils.cpp
claywar:
```
        // Update GetFirstID to use this new lookup
        // clang-format off
        lua.set_function("GetFirstID", [&](std::string const& name) -> std::optional<uint32>
        {
            if (lookup.find(name) != lookup.end())
            {
                return lookup[name].front();
            }
            else
            {
                ShowWarning(fmt::format("GetFirstID({}): Returning nil", name));

                return std::nullopt;
            }
        });
```

Untested, but this is what I would try

zach2good: I'm 95% sure this'll work, since all types in Lua are nullable. Or rather,  they're not complex enough to not be nullable. 

In case it doesn't work: since this is bound directly into Lua with `set_function`, you won't need to stuff the stuff into `std::optional<uint32>`. Instead, you can just use `sol::object`, so it'll be able to take both a number (maybe?) and `sol::lua_nil`

MowFord: This indeed worked:
image

claywar: I would suggest using ShowError instead of warn for GetFirstID as well, since we'd want to catch that in CI on startup, even if doesn't fail later on

src/map/lua/luautils.cpp
MowFord: Updated warning messages with the zone name in all permutations of failures for both functions:

image

I can't actually figure out how to trigger this warning, but updated as well:

image

zach2good (approved):

src/map/lua/luautils.cpp
MowFord: By that logic shouldn't GetTableOfIDs also show error? I would think either of these functions being used and returning nothing should be the same cause for concern

MowFord: Did the thing:
image

PR #5662, opened 2024-05-09
[core] Fix mobs turning while casting with Manafont up

WinterSolstice8:
**_I affirm:_**

Fixes an issue where mobs will rotate when manafont is up while casting (which they shouldn't do)
[2024-05-08 19-51-22.webm](https://github.com/LandSandBoat/server/assets/60417494/2a6bb024-d7ce-4936-a48c-6f1849f820f8)

`!exec target:addEffect(xi.effect.MANAFONT, -1, -1, -1)`
spin a blm and see it not spin with you while its casting

Sidenote: chainspell/manafont isnt as brutal on retail. Mobs do not instantly start recasting as soon as possible. Other minor nitpick is that it seems our spellcasting state ends too late, and mobs should start to rotate sooner, but also stay in place.. so not sure what that's about.

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -582,7 +582,7 @@ void CMobController::DoCombatTick(time_point tick)
         PMob->PAI->EventHandler.triggerListener("COMBAT_TICK", CLuaBaseEntity(PMob));
         luautils::OnMobFight(PMob, PTarget);
 
-        if (PMob->PAI->IsCurrentState<CInactiveState>())
+        if (PMob->PAI->IsCurrentState<CInactiveState>() || !PMob->PAI->CanChangeState())
         {
             return;
         }


WinterSolstice8:
P.S. there is a second way to fix this, that's more "laser precise" using the same added if check but only for IsSpellReady:

```
bool CMobController::IsSpellReady(float currentDistance)
{
    TracyZoneScoped;
    int32 bonusTime = 0;
    if (currentDistance > 5)
    {
        // Mobs use ranged attacks quicker when standing back
        bonusTime = PMob->getBigMobMod(MOBMOD_STANDBACK_COOL);
    }

    if (PMob->StatusEffectContainer->HasStatusEffect({ EFFECT_CHAINSPELL, EFFECT_MANAFONT }))
    {
        return true;
    }

    return (m_Tick >= m_LastMagicTime + std::chrono::milliseconds(PMob->getBigMobMod(MOBMOD_MAGIC_COOL) - bonusTime));
}
```

Xaver-DaRed (approved):

PR #5665, opened 2024-05-09
[LUA] Fix Shattering Stars BCNMs should restrict subjob

ampitere:
**_I affirm:_**

The Shattering Stars BCNMs are currently being converted into the new battlefield framework but are missing the parameter to disable subjobs. This causes players to be able to use their subjob during the BCNM which should not be possible.

1. !changejob 11 99
2. !additem 1436
3. !zone Horlais Peak
4. Trade the testimony to the Burning Circle


claywar (approved):
Yep, this was a miss on my part.  I was looking for rulemask settings for a more generic conversion, so expect this method of defining will change in the future

Though since this is a logical or, there may be a greater issue of first not starting with the fully-defined ruleset

PR #5673, opened 2024-05-10
[LUA] Fix typeless magic damage causes error in calculateResistRate

ampitere:
**_I affirm:_**

When using blue magic that does not have an elemental nature (physical for example) currently will throw an error in `xi.combat.magicHitRate.calculateResistRate` due to the element being nil when matching. To handle this we will default `targetResistRank` to 0 and only set it if the element is available. Example error log:

```
[05/09/24 18:33:06:346][map][error] luautils::onSpellCast: stack index 2, expected number, received nil: not a numeric type (bad argument into 'short(unsigned short)')
stack traceback:
	[C]: in function 'getMod'
	.\scripts/globals/combat/magic_hit_rate.lua:393: in function 'applyResistanceEffect'
	./scripts/globals/bluemagic.lua:556: in function 'usePhysicalSpellAddedEffect'
	./scripts/actions/spells/blue/disseverment.lua:55: in function <./scripts/actions/spells/blue/disseverment.lua:21> (luautils::OnSpellCast:2751)
```

1. !changejob 16 99
2. !addallspells
3. Set Disseverment in Blue Magic
4. Use Disseverment on a mob

### DIFF
--- a/scripts/globals/combat/magic_hit_rate.lua
+++ b/scripts/globals/combat/magic_hit_rate.lua
@@ -390,7 +390,11 @@ end
 
 xi.combat.magicHitRate.calculateResistRate = function(actor, target, skillType, spellElement, magicHitRate, rankModifier)
     local targetResistRate = 0 -- The variable we return.
-    local targetResistRank = target:getMod(xi.combat.element.resistRankMod[spellElement])
+    local targetResistRank = 0
+
+    if spellElement ~= nil and spellElement ~= xi.element.NONE then
+        targetResistRank = target:getMod(xi.combat.element.resistRankMod[spellElement])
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
That error makes absolutely 0 sense at all.
`.\scripts/globals/combat/magic_hit_rate.lua:393: in function 'applyResistanceEffect'`
"applyResistanceEffect" is a global function located in magic.lua, and NEVER used in combat/magic_hit_rate

Xaver-DaRed:
This will need more work, couse that error should not be happening at all when using a blue spell.
For the time being, I've converted your change in this PR #5679 and co-authored you.

PR #5677, opened 2024-05-10
[Cleaup] Shift-proof "G Zones

Frankie-hz:
**_I affirm:_**

Shift-proof more zones
Gustav Tunnel
Gusgen
Grauberg
Grand Palace
Giddeus
Ghelsba Outpost
Garlaige Citadel

More lottos and hand hurting


Xaver-DaRed (approved):
You know, I know it's not your fault, but, now that we use offsets, a lot of this PH tables seem... made up

Frankie-hz:
Yeah I did my best to remove the very obscure ones when I see them and make sure the mob in the offset is actually in that area. Surprisingly most were. Some NMs just dont use the mobs near its mob ID and I have no explanation for it

I also realized that all the lotto NMs have this math.random which most likely does not belong there. From what I understand the nm “opens” after the min time and its just chance until it spawns.

PR #5680, opened 2024-05-10
[Magic] Cleanup magic.lua, first pass

Xaver-DaRed:
**_I affirm:_**

- Deletes functions. Special grave offenders: functions which sole purpose was to redirect to a different function.
- Standarices, a little, magic hit rate calculations. Lots of wrong in there. But it was already there. If anything, I **accidentally** made them better.

Play the game. Fight against mobs. This affects a lot of skills, abilities and such.

This started because of an error reported by PR #5673 which made absolutely no sense.
Even when not used, magic.lua is still a bane screwing everything up.

### DIFF
--- a/scripts/globals/combat/magic_hit_rate.lua
+++ b/scripts/globals/combat/magic_hit_rate.lua
@@ -284,7 +284,7 @@ local function magicAccuracyFromFood(actor, magicAccPreFood)
     return magicAcc
 end
 
--- Global function to calculate magicc accuracy for addiional effects. (No stat, no magic burst bonuses)
+-- Global function to calculate magic accuracy for some non-spell magic actions. (No stat diff, no magic burst bonuses).
 xi.combat.magicHitRate.calculateNonSpellMagicAccuracy = function(actor, target, spellGroup, skillType, spellElement, bonusMacc)
     local finalMagicAcc = 0
 
--- a/scripts/globals/combat/magic_hit_rate.lua
+++ b/scripts/globals/combat/magic_hit_rate.lua
@@ -284,6 +284,29 @@ local function magicAccuracyFromFood(actor, magicAccPreFood)
     return magicAcc
 end
 
+-- Global function to calculate magicc accuracy for addiional effects. (No stat, no magic burst bonuses)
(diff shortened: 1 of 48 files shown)


scripts/globals/magic.lua
```diff
@@ -468,141 +317,66 @@ params.skillType = $3
...
+    return resistRate
 end
 
 -- Returns the amount of resistance the target has to the given effect
 local effectToResistanceMod =
 {
     [xi.effect.SLEEP_I      ] = xi.mod.SLEEPRES,
     [xi.effect.SLEEP_II     ] = xi.mod.SLEEPRES,
-    [xi.effect.LULLABY      ] = xi.mod.LULLABYRES,
+    [xi.effect.LULLABY      ] = xi.mod.SLEEPRES,
```
Frankie-hz, line 379: I know we found an NM that was resistant to lullaby and not sleep, do we want to make a TODO or a note here that it needs another look into at some point?

Xaver-DaRed: By resist, I assume you mean immunity, of a hight dark resistance and not a job trait trigger.
(This modifiers are specifically used for that. PLD "Resist Sleep" works for all)

Moving forward, we are going to use the immunity system for light and dark sleeps, instead of the regular sleep immunity.

As for this, all of this logic here in magic.lua is going to be nuked in favor of the new magic system.
This is just not being handled properly AND is seemingly affecting other stuff.

Our magic system is very robust. I believe it's robust enough to handle all magic things from spells, abilities, mobskills, etc...

scripts/globals/combat/magic_hit_rate.lua
```diff
@@ -284,6 +284,29 @@ local function magicAccuracyFromFood(actor, magicAccPreFood)
     return magicAcc
 end
 
+-- Global function to calculate magicc accuracy for addiional effects. (No stat, no magic burst bonuses)
```
UmeboshiXI: Just a small typo on 'magicc' and 'addiinal' in comment.

Xaver-DaRed: Corrected

claywar (approved):

PR #5686, opened 2024-05-11
[core] Add utility method to get highest job level of char

TracentEden2:
**_I affirm:_**

This PR adds a utility function to get the highest level of all jobs of a char. This is useful for implementing retail logic such as fishing fatigue being different if chars highest level is less than 20 (see [retail patch notes](http://forum.square-enix.com/ffxi/threads/41117)). This is also useful for private servers that can use the function to implement restrictions to avoid spammers and bots. For example Horizon implements a min highest level of 20 for using yell and for fishing.

There is no user facing LSB implementation as of yet (as fishing data for example is not fully known). Thus one way to test is to add code to like zoning logic that prints to the player their highest level, the test by zoning.

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -2442,6 +2442,21 @@ int32 CCharEntity::GetTimeCreated()
     return 0;
 }
 
+uint8 CCharEntity::getHighestJobLevel()
+{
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):
I dont see why not. Simple loop.

PR #5687, opened 2024-05-11
[Cleanup]Standardize AF Quest Names

hooksta4:
**_I affirm:_**

Standardizes all current Artifact Quests into [JOB]AF[1-3]Quest_Name
Adjusted quests that were not using player:getMainLvl() >= xi.settings.main.AF[X]_QUEST_LEVEL

Standardizes naming Convention
None should be noted. Renamed files.


Xaver-DaRed (approved):
I'm gonna approve couse I was the one who started this trend.

However, from this point on, I believe we should stick to just "Those questlines everyone does". Aka, Limit breaks and AFs

claywar:
Apologies, but a rebase will be required here due to corrections in that quest script

hooksta4:
> Apologies, but a rebase will be required here due to corrections in that quest script

Did I finally do it correctly?

hooksta4:
> I'm gonna approve couse I was the one who started this trend.
> 
> However, from this point on, I believe we should stick to just "Those questlines everyone does". Aka, Limit breaks and AFs

Yea I don't think there are any more quest series that are more than a few long that would really warrant these. The ones that come to mind are already done, The Onions series, Rycharde chef... I mean maybe the ToAU mythic questline?

PR #5695, opened 2024-05-11
[LUA] Fix Ripper Fang doesn't match wikis

ampitere:
**_I affirm:_**

Noticed some comment issues on Ripper Fang which led me down the rabbit hole. Wikis seems to have conflicting information but from what I can gather it should be closer to 200% modifier (JP wiki says 150%?) and slashing damage, the additional effect may no longer be a thing but would require retail verification so I've left it in for now.

https://www.bg-wiki.com/ffxi/Category:Raptor
https://www.bg-wiki.com/ffxi/Category:Jug_Pets
https://wiki.ffo.jp/html/23464.html

1. Find a mob
2. !exec target:useMobAbility(374)

### DIFF
--- a/scripts/actions/mobskills/ripper_fang.lua
+++ b/scripts/actions/mobskills/ripper_fang.lua
@@ -1,7 +1,7 @@
 -----------------------------------
---  Sprout Smack
---  Description: Additional effect: Slow.  Duration of effect varies with TP.
---  Type: Physical (Blunt)
+--  Ripper Fang
+--  Description: Deals physical damage. Damage varies with TP.
+--  Type: Physical (Slashing)
 -----------------------------------
 local mobskillObject = {}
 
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
Jimmy's spreadsheet says 1.5 with some attack bonus

https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit#gid=57955395&range=A410

ampitere:
> Jimmy's spreadsheet says 1.5 with some attack bonus
> 
> https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit#gid=57955395&range=A410

Definitely bookmarking that doc. 

Removed the status effect since it doesn't appear to have one. Updated dmgmod value to 1.8 (30% seems fair) as that seems to be how these attack bonuses are calculated for other skills with attack bonuses. Please let me know if there's a better way to add the attack bonus though as I didn't see one.

scripts/actions/mobskills/ripper_fang.lua
```diff
@@ -12,12 +12,11 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
     local accmod = 1
-    local dmgmod = 2.5
+    local dmgmod = 1.8
```
WinterSolstice8:
```suggestion
    local dmgmod = 1.5
    local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.physicalTpBonus.ATK_VARIES, 1.3, 1.3, 1.3)

```

ampitere: Refactored, thanks.

WinterSolstice8 (approved):

Issue #669, opened 2021-08-29
Flurry issues


almuth150:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: base

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
1) The effect does not display the duration when using the in-game status duration display, despite it not being set to a duration of infinite. I haven't seen any other status effect do this.

2) The spell appears to have a mismatch with valid targets. The client lets me target people outside of party/alliance when selecting to cast it from the menu, unlike other restricted spells such as Phalanx II. However, I cannot actually cast the spell on anyone except a party member. Neither wiki gives information about valid targets, unfortunately.

3) Not sure what's going on here, but the FFXIclopedia entry for the spell says the effect is mutually-exclusive with Haste effect (ie. both overwrites and is overwritten by Haste). However, the page for the effect on that same wiki says it *can* co-exist with the Haste effect. BG wiki does not say anything about this.

I do not know if any of these issues applies to Flurry II because the server I play on does not have access to that spell.

PR #5697, opened 2024-05-12
[Magic] Fix Flurry and Haste overwrites and Flurry self stack

ampitere:
**_I affirm:_**

Currently players can stack Flurry and Haste but this should not be possible as they overwrite each other.

https://wiki.ffo.jp/html/644.html
https://wiki.ffo.jp/html/31588.html

It also seems Flurry is using the wrong effect as there are 2 types of Flurry (at least in English, in Japanese Flurry is called Snap). The 265 effect actually seems to be the Nyzul specific version of Flurry which does not have any duration and I believe is the reason the duration does not properly display for Flurry. It also cannot be removed if you attempt to do so. As a result Flurry and Flurry II will no longer be stackable, I couldn't find documentation on whether they do stack but since it works similarly to Haste I can only assume that you cannot have both effects at the same time.

https://wiki.ffo.jp/html/11145.html

Before (Flurry and Flurry II stacking)
image

After
image

This fixes the 1st and 3rd issue on https://github.com/LandSandBoat/server/issues/669. The 2nd issue should already be resolved from what I can tell.

1. !changejob 5 99
2. !addallspells
3. Cast Haste
4. Cast Flurry, it should overwrite Haste (!getmod snap_shot should give 15)
5. Cast Haste, it should overwrite Flurry
6. Cast Flurry II, it should overwrite Haste (!getmod snap_shot should give 30)
7. Cast Flurry, it should have no effect (!getmod snap_shot should still give 30)


Xaver-DaRed:
This will need retail captures.

Not asking you to do it.

PR #5708, opened 2024-05-12
[Magic] Add Shield Barrier trait functionality

ampitere:
**_I affirm:_**

Adds functionality for the PLD Shield Barrier trait as described on the wiki. Adds the defense value of the equipped shield to the power of all Protect spells cast by self.

https://www.bg-wiki.com/ffxi/Shield_Barrier

1. !changejob 7 99
2. !getmod shield_barrier should be 1
3. !addallspells
4. !additem 15070
5. Cast Protect with shield on, bonus should apply
6. !changejob 7 69
7. !getmod shield_barrier should be 0
8. Cast Protect with shield on, bonus shouldn't apply

Protect base 20
Protect with shield 60
Protect II base 50
Protect II with shield 90
Protect III base 90
Protect III with shield 130
Protect IV base 140
Protect IV with shield 180
Protect V base 220
Protect V with shield 260

### DIFF
--- a/scripts/globals/spells/enhancing_spell.lua
+++ b/scripts/globals/spells/enhancing_spell.lua
@@ -318,6 +318,14 @@ xi.spells.enhancing.calculateEnhancingFinalPower = function(caster, target, spel
             finalPower = finalPower + (tier * 2)
         end
 
+        -- Handle "Shield Barrier" Job Trait.
+        if
+            caster:isPC() and
+            caster:getMod(xi.mod.SHIELD_BARRIER) > 0
+        then
+            finalPower = finalPower + caster:getShieldDefense()
+        end
+
     -- Refresh
     elseif spellEffect == xi.effect.REFRESH then
         finalPower = finalPower + caster:getMod(xi.mod.ENHANCES_REFRESH)
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -602,6 +602,18 @@ int8 CCharEntity::getShieldSize()
     return PItem->getShieldSize();
 }
 
+int16 CCharEntity::getShieldDefense()
+{
+    CItemEquipment* PItem = getEquip(SLOT_SUB);
+
+    if (PItem && PItem->IsShield())
+    {
+        return PItem->getModifier(Mod::DEF);
+    }
+
+    return 0;
+}
+
 bool CCharEntity::hasBazaar()
 {
     if (isSettingBazaarPrices)
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -4726,6 +4726,27 @@ int8 CLuaBaseEntity::getShieldSize()
 }
 
 /************************************************************************
+ *  Function: getShieldDefense()
+ *  Purpose : Return the defense of the equipped shield
+ *  Example : player:getShieldDefense()
+ *  Notes   : Returns 0 if player does not have shield equipped
+ ************************************************************************/
+
+int16 CLuaBaseEntity::getShieldDefense()
+{
+    if (m_PBaseEntity->objtype == TYPE_PC)
+    {
+        return static_cast<CCharEntity*>(m_PBaseEntity)->getShieldDefense();
+    }
+    else
+    {
+        ShowWarning("Entity is not a Player: (%s).", m_PBaseEntity->getName());
+    }
+
+    return 0;
+}
+
+/************************************************************************
  *  Function: addGearSetMod()
  *  Purpose : Need to research functionality more to provide description
  *  Example : player:addGearSetMod(setId, modId, modValue)
@@ -17611,6 +17632,7 @@ void CLuaBaseEntity::Register()
     SOL_REGISTER("unlockEquipSlot", CLuaBaseEntity::unlockEquipSlot);
 
     SOL_REGISTER("getShieldSize", CLuaBaseEntity::getShieldSize);
+    SOL_REGISTER("getShieldDefense", CLuaBaseEntity::getShieldDefense);
 
     SOL_REGISTER("addGearSetMod", CLuaBaseEntity::addGearSetMod);
     SOL_REGISTER("clearGearSetMods", CLuaBaseEntity::clearGearSetMods);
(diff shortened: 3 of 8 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -4725,6 +4725,29 @@ int8 CLuaBaseEntity::getShieldSize()
...
+        return 0;
+    }
+
+    if (m_PBaseEntity->objtype == TYPE_PC)
+    {
+        return static_cast<CCharEntity*>(m_PBaseEntity)->getShieldDefense();
+    }
+
+    return 0;
+}
```
Xaver-DaRed, line 4747:
```c++
int16 CLuaBaseEntity::getShieldDefense()
{
    if (m_PBaseEntity->objtype == TYPE_PC)
    {
        return static_cast<CCharEntity*>(m_PBaseEntity)->getShieldDefense();
    }
    else
    {
        ShowWarning("Entity is not a Player: (%s).", m_PBaseEntity->getName());
    }

    return 0;
}
```

src/map/entities/charentity.cpp
```diff
@@ -602,6 +602,23 @@ int8 CCharEntity::getShieldSize()
...
+        return 0;
+    }
+
+    if (!PItem->IsShield())
+    {
+        return 0;
+    }
+
+    return PItem->getModifier(Mod::DEF);
+}
```
Xaver-DaRed, line 615:
```c++
int16 CCharEntity::getShieldDefense()
{
    CItemEquipment* PItem = getEquip(SLOT_SUB);

    if (PItem && PItem->IsShield())
    {
        return PItem->getModifier(Mod::DEF);
    }

    return 0;
}
```

scripts/globals/spells/enhancing_spell.lua
```diff
@@ -318,6 +318,14 @@ xi.spells.enhancing.calculateEnhancingFinalPower = function(caster, target, spel
...
         end
 
+        if caster:getMod(xi.mod.SHIELD_BARRIER) > 0 then
+            local shieldDefense = caster:getShieldDefense()
+
+            if shieldDefense > 0 then
+                finalPower = finalPower + shieldDefense
+            end
+        end
+
```
Xaver-DaRed, line 328:
```lua
        -- Handle "Shield Barrier" Job Trait.
        if
            caster:isPC() and
            caster:getMod(xi.mod.SHIELD_BARRIER) > 0
        then
            finalPower = finalPower + caster:getShieldDefense()
        end

Xaver-DaRed: We need the PC check, so we dont clutter the console log with errors when mobs start casting protect wile roaming.

The core functions will already make sure to check for gear and return a non nil value, so we can simplify the logic

Xaver-DaRed: Also, here's some homework: What happens if a PLD subs /whm and casts protectra?
Does the other party members ALSO get the DEF bonus? Or just the PLD?

Xaver-DaRed (changes requested):
Apply the minor style corrections I pointed out.
We also need to knpw what happens with the edge case.

scripts/globals/spells/enhancing_spell.lua
ampitere: Refactored. They do get the bonus and they should (I think):

> A level 70+ Paladin can cast Protect V to all members of the party within range and bonus applies to all party members. 

https://www.bg-wiki.com/ffxi/Category:Protect_Spell

It should say 90+ for Protect V but it applies to all types of Protect. It also applies to the party when they use Majesty so Protectra should also get it (at least in my mind).

src/map/entities/charentity.cpp
ampitere: Refactored.

src/map/lua/lua_baseentity.cpp
ampitere: Refactored.

ampitere:
> Apply the minor style corrections I pointed out. We also need to knpw what happens with the edge case.

Refactors completed, it does work with Protectra with WHM sub which makes sense since it works with Majesty as well. Funny enough it actually looks like Majesty was changed to work with Protect in the same update that Shield Barrier was added:

https://forum.square-enix.com/ffxi/threads/57008-July-8-2020-%28JST%29-Version-Update?p=628917&viewfull=1#post628917

scripts/globals/spells/enhancing_spell.lua
Xaver-DaRed: Yeah, however, i did ask for protectra concretely, for a reason.

Retail treats protect and protectra as different spell families, which means the possibility is still there.

Retail is just as spaguetti as we are.

src/map/entities/charentity.cpp
Xaver-DaRed: I think you swaped both functions. This is the charentity file and that you have defined is the lua base entity function

scripts/globals/spells/enhancing_spell.lua
ampitere: It's free week and I'm pretty sure I have PLD over 70 let me go try to check on retail.

src/map/lua/lua_baseentity.cpp
Xaver-DaRed: and same here: this is the lua base entity file and that is the charentity function

scripts/globals/spells/enhancing_spell.lua
ampitere: Damn nope only 66.

src/map/entities/charentity.cpp
ampitere: I definitely did, fixed.

src/map/lua/lua_baseentity.cpp
ampitere: Fixed.

scripts/globals/spells/enhancing_spell.lua
Xaver-DaRed: I can check it in retail tonight.

Better safe than sorry. One has to be nitpicky with this stuff.

ampitere: I totally understand and appreciate it!

Xaver-DaRed (approved):

PR #5709, opened 2024-05-13
[CPP] Fix error on zone when naked

### LINKED ISSUE #5702: 🐛 Loading/zoning on a naked character throws an error
**_I affirm:_**

Windows 10

## Branch affected by issue 

`base`

## Steps to reproduce

1. Unequip everything
2. Zone

`[05/12/24 10:01:36:317][map][error] Loading error from char_equip (charutils::LoadEquip:1118)`

## Expected behavior

An error shouldn't be thrown just because a character is naked.

ampitere:
**_I affirm:_**

Adds a check to prevent reporting an error if a character is naked and has no equipment to load. Resolves https://github.com/LandSandBoat/server/issues/5702

1. Unequip everything
2. Zone
3. Error shouldn't appear

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -1020,7 +1020,7 @@ namespace charutils
                             "WHERE charid = (?)";
 
         auto rset = db::preparedStmt(Query, PChar->id);
-        if (rset && rset->rowsCount())
+        if (rset)
         {
             // equipSlotData[equipSlotId] = { slotId, containerId }
             std::unordered_map<uint8, std::pair<uint8, uint8>> equipSlotData;


WinterSolstice8:
checking if you're naked here really is a no-op, this happens on CCharEntity construction when being put into the world, the intent of that error message appears to be checking for if the query failed, but apparently literally zero equipment will trip the false positive here

ampitere:
> checking if you're naked here really is a no-op, this happens on CCharEntity construction when being put into the world, the intent of that error message appears to be checking for if the query failed, but apparently literally zero equipment will trip the false positive here

Do we need this error if `db::preparedStmt` is already reporting errors if it runs into them? It will show if the query failed and give the query so having both errors seems a bit redundant to me.

I'm not sure what else we can check for if `db::preparedStmt` fails gracefully and returns no results but also returns no results when a player is naked.

WinterSolstice8:
Will wait for some input on Zach but I think if that is the case then we just don't need to log the error at all.

zach2good:
What did the code look like before my refactor?

claywar:
Gut feeling is that log message added with the nuking of `XI_DEBUG_BREAK_IF` (and oops to editing Zach's comment instead of adding this)

ampitere:
> What did the code look like before my refactor?

The if was previously `if (ret != SQL_ERROR)`. The error was touched 3 years ago but was just cleanup.

zach2good:
In which case none of this is really needed to match what was there before. This:
```cpp
auto rset = db::preparedStmt(Query, PChar->id);
if (rset && rset->rowsCount())
```

Can/should become:
```cpp
auto rset = db::preparedStmt(Query, PChar->id);
if (rset)
```

ie. "The query didn't fail" rather than "The query didn't fail AND there are rows of results" etc.

ampitere:
> In which case none of this is really needed to match what was there before. This:
> 
> ```c++
> auto rset = db::preparedStmt(Query, PChar->id);
> if (rset && rset->rowsCount())
> ```
> 
> Can/should become:
> 
> ```c++
> auto rset = db::preparedStmt(Query, PChar->id);
> if (rset)
> ```
> 
> ie. "The query didn't fail" rather than "The query didn't fail AND there are rows of results" etc.

Changed to `if (rset)`.

zach2good:
Presumably you've tested it and it works?

ampitere:
> Presumably you've tested it and it works?

Works perfectly, yes.

zach2good (approved):

PR #5710, opened 2024-05-13
[SQL] Level correct Gyre-Carlin

### LINKED ISSUE #5700: 🐛 Gyre-Carlin is probably set to too high of a level
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

Spawn Gyre-Carlin, attempt to fight it as a group of level 40 players as was once intended by SE when the area once capped players at level 40, and get annihilated by the level 75-80 NM.

## Expected behavior

Looking at some old testimonials, I see one saying that they lost only 2 shadows as a level 74 THF, and another saying they got 53 EXP as a level 64 MNK in Sept. 2010.

ampitere:
**_I affirm:_**

[Gyre-Carlin](https://www.bg-wiki.com/ffxi/Gyre-Carlin) is currently set the same min/max as the Nyzul version which isn't appropriate for Pso'Xja. The wiki lists his level as 50, and a capture video also shows the same so. Referencing a similar NM in the same zone ([Nunyunuwi](https://www.bg-wiki.com/ffxi/Nunyunuwi)) we'll tweak the min to 56 and max to 58. HP is estimated to 5883 in the capture. Resolves https://github.com/LandSandBoat/server/issues/5700

https://www.bg-wiki.com/ffxi/Gyre-Carlin
https://www.youtube.com/watch?v=UYTrBzyYaXI

I am not entirely how to glean the real level from captures (if that's even possible). If there is a way please let me know!

1. !zone Pso'Xja
2. !mobhere 16814331

### DIFF
--- a/sql/mob_groups.sql
+++ b/sql/mob_groups.sql
@@ -429,7 +429,7 @@ INSERT INTO `mob_groups` VALUES (52,3992,9,'Treasure_Chest',0,128,0,0,0,55,60,0)
 INSERT INTO `mob_groups` VALUES (53,6387,9,'Gazer',924,0,315,0,0,42,44,0);
 INSERT INTO `mob_groups` VALUES (54,1040,9,'Diremite',924,0,657,0,0,42,46,0);
 INSERT INTO `mob_groups` VALUES (55,3685,9,'Snowball',924,0,2286,0,0,43,46,0);
-INSERT INTO `mob_groups` VALUES (56,6963,9,'Gyre-Carlin',0,32,1260,0,0,75,80,0);
+INSERT INTO `mob_groups` VALUES (56,6963,9,'Gyre-Carlin',0,32,1260,4883,0,50,50,0);
 INSERT INTO `mob_groups` VALUES (57,2922,9,'Nunyunuwi',0,128,0,3500,0,56,58,0);
 INSERT INTO `mob_groups` VALUES (58,292,9,'Avalanche',960,0,30,0,0,75,75,0);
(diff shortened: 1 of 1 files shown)


almuth150:
Oh, I actually managed to find an old EXP gain table dating back to 2010 [here](https://ffxiclopedia.fandom.com/wiki/Experience_Points?oldid=1258732) to go along with the [old testimonial](https://ffxiclopedia.fandom.com/wiki/Talk:Gyre-Carlin) of getting 53 EXP at level 64, To gain 53 EXP from a kill at that level, the mob would have to be 14 levels lower than the player, which would put that particular spawn at level 50, though that doesn't cover the full potential level range. Which makes sense for a NM in an area that, at the time it was added, was in an area that restricted the level of players to 40.

That being said, I don't know if anything about the NM was changed over the years. I'd be surprised if there were any changes to the NM itself though. Also, I had to dig up an old EXP gain table because it seems that EXP gain was changed in 2011?

Edit: Nunyunuwi is in the part of Pso'Xja that wasn't under level restriction back when it was originally added, unlike Gyre-Carlin.

WinterSolstice8:
The widescan packet, for some reason, has mob levels in it. the windower capture suite captures that in a table

however, I'm not convinced its always correct. JoL doesn't give enough exp to match it's widescan level for instance.

ampitere:
It does look like the widescan capture also shows 50. What do y'all think? Keep it 56-58 or tune down to 50-50?

WinterSolstice8:
well, given the two data points we have 50 makes sense

ampitere:
Tweaked to 50-50.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5711, opened 2024-05-13
[CPP] Fix non-equipment items can cause map crash

### LINKED ISSUE #5703: 🐛 map server crash while calling canEquipItem()
**_I affirm:_**

Windows10

## Branch affected by issue 

`base`

## Steps to reproduce

I've not been able to reproduce this reliably despite trying to track down the responsible player(s).
Full dump available upon request. I looked through recently merged PR's and one that may be relevant is #5622 

image

## Expected behavior
Condition should be handled gracefully as to not result in a crash.

ampitere:
**_I affirm:_**

Currently since `CLuaBaseEntity::canEquipItem` only takes an item ID as a param then casts it, it will sometimes be a non-equipment item. With the addition of `isEquippableByRace` it attempts to check a modifier that doesn't exist on non-equipment items. To fix this we'll add a return false if the item is not equipment. Resolves https://github.com/LandSandBoat/server/issues/5703

1. Go to a conquest NPC
2. Keep opening the "Are you sure you want to purchase the X?" menu on a non-equipment item (Scroll of Instant Warp for example)

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -4557,6 +4557,11 @@ bool CLuaBaseEntity::canEquipItem(uint16 itemID, sol::object const& chkLevel)
         return false;
     }
 
+    if (!PItem->isType(ITEM_EQUIPMENT))
+    {
+        return false;
+    }
+
     if (!(PItem->getJobs() & (1 << (PChar->GetMJob() - 1))))
     {
         return false;


WinterSolstice8 (approved):

WinterSolstice8:
Just to be sure -- you also checked ranged weapons, ammo, shields, things that aren't weapons but are equippable in ammo, etc?

I'm sure it works based on the set type/get type functions
```
void CItem::setType(uint8 type)
{
    m_type |= type;
}

bool CItem::isType(ITEM_TYPE type) const
{
    return (m_type & type);
}
```
all of those things should be under "CItemEquipment" but this GetType junk is a bit unknown to me

ampitere:
> Just to be sure -- you also checked ranged weapons, ammo, shields, things that aren't weapons but are equippable in ammo, etc?
> 
> I'm sure it works based on the set type/get type functions
> 
> ```
> void CItem::setType(uint8 type)
> {
>     m_type |= type;
> }
> 
> bool CItem::isType(ITEM_TYPE type) const
> {
>     return (m_type & type);
> }
> ```
> 
> all of those things should be under "CItemEquipment" but this GetType junk is a bit unknown to me

The cast to `CItemEquipment` changes the type to `EQUIPMENT` so if it's not that type then it failed the cast and still has it's previous type as far as I can tell. Shields, ammo, fishing rods, bait, lures, guns, bows, etc. are all in `item_equipment.sql` so those all should properly cast to equipment.

Xaver-DaRed (approved):

PR #5712, opened 2024-05-13
[CPP/LUA] Add synthesis high quality chance configurable setting

ampitere:
**_I affirm:_**

Adds synthesis high quality chance configurable setting requested in https://github.com/LandSandBoat/server/discussions/5453

1. Tweak the setting
2. Attempt to synth or desynth

### DIFF
--- a/settings/default/map.lua
+++ b/settings/default/map.lua
@@ -138,6 +138,9 @@ xi.settings.map =
     -- Amount of points allowed in crafts over the level defined above. Points are shared across all crafting skills. (Retail = 400; All skills can go to max = 3200)
     CRAFT_SPECIALIZATION_POINTS = 400,
 
+    -- Multiplier applied to high quality chance
+    CRAFT_HQ_CHANCE_MULTIPLIER = 1.0,
+
     -- Enables fishing. 0 = Disabled. 1 = Enable. ENABLE AT YOUR OWN RISK.
     FISHING_ENABLE = false,
 
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -364,6 +364,7 @@ namespace synthutils
             // HQ success rate modifier.
             // See: https://www.bluegartr.com/threads/130586-CraftyMath-v2-Post-September-2017-Update page 3.
             chanceHQ = chanceHQ + 100.0f * PChar->getMod(Mod::SYNTH_HQ_RATE) / 512.0f;
+            chanceHQ = chanceHQ * settings::get<float>("map.CRAFT_HQ_CHANCE_MULTIPLIER"); // server configured additional HQ multiplier (default 1.0)
 
             // limit max hq chance
             if (chanceHQ > maxChanceHQ)


ampitere:
Will do some more testing on this tomorrow, created as a draft for now.

Xaver-DaRed:
Several things here:
- Synthesis configurations should all remain together. This should go bellow the existing ones.
- They should all follow the same naming convention. This should all start with `CRAFT_`
- I think this is absolutely overkill. Current configurations are there because once upon a time, the system worked with different caps and different rates (era vs modern). HQ rates have never changed, I dont think EVERY single number needs a configuration.

I would limit it to a single HQ chance multiplier. That would seem more reasonable.

settings/default/map.lua
```diff
@@ -123,6 +123,36 @@ xi.settings.map =
...
+    SYNTHESIS_DEFAULT_HQ_CHANCE = 0.0,
+
+    -- T1 high quality chance (0-10 levels over recipe)
+    SYNTHESIS_TIER_1_HQ_CHANCE = 1.0,
+
+    -- T2 high quality chance (11-30 levels over recipe)
+    SYNTHESIS_TIER_2_HQ_CHANCE = 5.0,
+
+    -- T3 high quality chance (31-50 levels over recipe)
+    SYNTHESIS_TIER_3_HQ_CHANCE = 20.0,
```
Frankie-hz: 25 not 20

src/map/utils/synthutils.cpp
```diff
@@ -340,25 +340,25 @@ namespace synthutils
...
-                    chanceHQ = 50.0;
+                    chanceHQ = settings::get<float>("map.SYNTHESIS_TIER_4_HQ_CHANCE");
                     break;
                 case 3: // 1 in 4
-                    chanceHQ = 25.0;
+                    chanceHQ = settings::get<float>("map.SYNTHESIS_TIER_3_HQ_CHANCE");
                     break;
                 case 2: // 1 in 20
-                    chanceHQ = 5.0;
+                    chanceHQ = settings::get<float>("map.SYNTHESIS_TIER_2_HQ_CHANCE");
```
Frankie-hz: This might be a little out of scope for this PR but I will leave this here anyway so its at least somewhere. I believe this crafting rate should be 6.25 (1/16) instead of (1/20). Most likely the same for T0 is 1/64 instead of 1/100 but thats been very difficult to prove. There are very large sample sizes from Byrth here: https://www.bluegartr.com/threads/120352-CraftyMath

The 5% came from the JP wiki which is only based off 1000 crafts in each tier which has an HQ T1 rate of 5.82%.  https://wiki.ffo.jp/html/843.html

SE said the HQ rates were increased and people always guessed T1 was changed from 1/16 to 1/10 but from all the sample sizes it was never proven as far as I know and SE never came out to say what the rates actually were. 
image

Xaver-DaRed: Remember that when in doubt, we always go with a "in players detriment" choice.

Boosting a rate is ok for a server OP.
Having to nerf it, however, causes drama and the damage has already be done.

Frankie-hz: You don't have to tell me that ;) haha

Xaver-DaRed: Well, it was for everyone, not specifically for you.

This kind of stuff has to be said on occasions. New ppl come and dont know the reasoning behind it.

ampitere: I can certainly tweak this to 6.25, would y'all prefer a separate PR for that? Or are we fine with keeping it at 5.0 for now?

settings/default/map.lua
ampitere: Ended up getting rid of this.

ampitere:
> Several things here:
> 
> * Synthesis configurations should all remain together. This should go bellow the existing ones.
> * They should all follow the same naming convention. This should all start with `CRAFT_`
> * I think this is absolutely overkill. Current configurations are there because once upon a time, the system worked with different caps and different rates (era vs modern). HQ rates have never changed, I dont think EVERY single number needs a configuration.
> 
> I would limit it to a single HQ chance multiplier. That would seem more reasonable.

Refactored into a single setting that is an additional multiplier applied to `chanceHQ`.

src/map/utils/synthutils.cpp
```diff
@@ -345,8 +345,8 @@ namespace synthutils
                 case 3: // 1 in 4
                     chanceHQ = 25.0;
                     break;
-                case 2: // 1 in 20
-                    chanceHQ = 5.0;
+                case 2: // 1 in 16
+                    chanceHQ = 6.25;
```
ampitere: More than happy to split this into a different PR or keep it the same as it has been.

Frankie-hz: I would personally change it because the data shows it closer to the rates SE use for other things but if xaver wants to keep it 5 we can open a discussion post and gather all the data in one place before making a change like this one. (if at all)

src/map/utils/synthutils.cpp
Xaver-DaRed: For the time being, all values should be kept as they are.

src/map/utils/synthutils.cpp
Xaver-DaRed: Lets not mix stuff. I would focus this PR on the single multiplier setting.

As for the rates, feel free to open a discussion, but as I said up there, without really really solid evidence, they should stay against player's odds, for server OPs to judge.

src/map/utils/synthutils.cpp
```diff
@@ -361,9 +361,11 @@ namespace synthutils
                 chanceHQ = chanceHQ * 1.5f;
             }
 
+            float chanceHQMultiplier = settings::get<float>("map.CRAFT_HQ_CHANCE_MULTIPLIER"); // server configured additional HQ multiplier (default 1.0)
+
             // HQ success rate modifier.
             // See: https://www.bluegartr.com/threads/130586-CraftyMath-v2-Post-September-2017-Update page 3.
-            chanceHQ = chanceHQ + 100.0f * PChar->getMod(Mod::SYNTH_HQ_RATE) / 512.0f;
+            chanceHQ = (chanceHQ * chanceHQMultiplier) + 100.0f * PChar->getMod(Mod::SYNTH_HQ_RATE) / 512.0f;
```
Xaver-DaRed: The setting should come last. Affecting both gear and base, etc...
So...
```c++
chanceHQ = chanceHQ + 100.0f * PChar->getMod(Mod::SYNTH_HQ_RATE) / 512.0f;
chanceHQ = chanceHQ * settings::get<float>("map.CRAFT_HQ_CHANCE_MULTIPLIER");
```
No need to define a variable for the setting if the setting is only used once.

src/map/utils/synthutils.cpp
ampitere: Removed.

src/map/utils/synthutils.cpp
ampitere: Refactored.

src/map/utils/synthutils.cpp
ampitere: Reverted.

Xaver-DaRed (approved):
Beautiful

src/map/utils/synthutils.cpp
ampitere: Added a discussion: https://github.com/LandSandBoat/server/discussions/5720

Frankie-hz (approved):

PR #5713, opened 2024-05-13
[core, lua] Add ability to jail offline chars

TracentEden2:
**_I affirm:_**

This PR adds the ability to jail offline chars to the !jail command. This saves admins a lot of time because they do not have to run manual database queries to jail chars that are offline. The PR also adds utility functions PlayerHasValidSession and GetPlayerIDByName that !jail and SendItemToDeliveryBox can use.

Attempt to jail characters that are online (on same map process as GM), online (on different map process as GM, will notify that does not work), and offline.

### DIFF
--- a/scripts/commands/jail.lua
+++ b/scripts/commands/jail.lua
@@ -1,6 +1,7 @@
 -----------------------------------
 -- func: jail
 -- desc: Sends the target player to jail. (Mordion Gaol)
+-- note: Only works if the target is on the same map process (cluster) as the GM or the target is offline
 -----------------------------------
 local commandObj = {}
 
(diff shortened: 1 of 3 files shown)


scripts/commands/jail.lua
```diff
@@ -26,31 +27,60 @@ commandObj.onTrigger = function(player, target, cellId, reason)
...
+    -- did not find the target (so either invalid target name, offline target, or online target in a different map process (cluster) than GM)
+    if targetObj == nil then
+        -- check if the target name is in the database
+        targetID = GetPlayerIDByName(target)
+        -- if name in the database then will have valid ID > 0
+        if targetID > 0 and targetID < 0xFFFFFFFF then
+            -- next check if the target has a valid session (and thus online in a different map process (cluster))
+            if PlayerHasValidSession(targetID) then
+                -- target online in different map process (cluster) than GM so inform the GM and exit
+                player:printToPlayer(string.format("Player '%s' is online but in a different zone. Go to that zone to use !jail", target))
```
Xaver-DaRed: I think you meant to say cluster instead of zone.

TracentEden2: > I think you meant to say cluster instead of zone.

I did not want to confuse GMs as some private servers have many clusters and the mapping of zones to clusters can be non-intuitive (so GMs will not actually know the mapping) so I thought easier to just say they need to be in the same zone.

Xaver-DaRed: The wording here would confuse or missinform of the actual command usage in the other directiion.

GMs SHOULD know how the server is setup

I guess you could change it to "zone group" (since cluster would also be technically incorrect from what i've been told)

TracentEden2: > The wording here would confuse or missinform of the actual command usage in the other directiion.
> 
> GMs SHOULD know how the server is setup
> 
> I guess you could change it to "zone group" (since cluster would also be technically incorrect from what i've been told)

Yeah, I will change to zone group or something, thanks for feedback! 

Yeah ideally GMs would know but for example there are over 50 "clusters" on Horizon so I think GMs would need to consult a table and the clustering is changing over time for performance reasons. In any case zone group is a better term.

TracentEden2: Ok updated and changed the terms.

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #5717, opened 2024-05-13
[cpp] Immanence logic cleanup

MowFord:
**_I affirm:_**

image

Apologies for the PR re-creation that will be happening, but any of my older PRs that need changes will have to be re-made as I've lost access to the source branches.

I will try to do my best to re-create the comments to avoid extra work on LSB's side.

(I won't be putting this disclaimer on every new PR, just a link to the old)

https://github.com/LandSandBoat/server/pull/5525

image

Good call on extending the duration, as the skillchain tier does affect the remaining duration of the skillchain effect.

Same testing behavior is useful: add prints to `effects/skillchain.lua` to get the explicit start/stop of the effect (Note that the effect isn't removed when a skillchain is created, so this below screenshot is the logs from the above skillchains)
image

and helixes
image

image

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -1131,70 +1131,69 @@ void CCharEntity::OnCastFinished(CMagicState& state, action_t& action)
                 StatusEffectContainer->DelStatusEffectSilent(EFFECT_CHAIN_AFFINITY);
             }
 
-            if (actionTarget.param > 0 &&
+            // Immanence will create or extend a skillchain for elemental spells
+            if (actionTarget.param >= 0 &&
                 PSpell->dealsDamage() &&
                 PSpell->getSpellGroup() == SPELLGROUP_BLACK &&
                 (StatusEffectContainer->HasStatusEffect(EFFECT_IMMANENCE)))
             {
-                SUBEFFECT effect = SUBEFFECT_NONE;
+                auto      immanenceApplies = true;
+                auto      isHelix          = false;
+                SUBEFFECT effect           = SUBEFFECT_NONE;
                 switch (PSpell->getSpellFamily())
                 {
-                    case SPELLFAMILY_STONE:
-                        effect = battleutils::GetSkillChainEffect(PTarget, 4, 0, 0);
-                        break;
                     case SPELLFAMILY_GEOHELIX:
-                        effect = battleutils::GetSkillChainEffect(PTarget, 4, 0, 0);
-                        break;
-                    case SPELLFAMILY_WATER:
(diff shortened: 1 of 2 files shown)


src/map/entities/charentity.cpp
```diff
@@ -1212,6 +1211,13 @@ void CCharEntity::OnCastFinished(CMagicState& state, action_t& action)
                         actionTarget.addEffectMessage = 287 + effect;
                     }
                     actionTarget.additionalEffect = effect;
+
+                    // Closing a skillchain with an immanence Helix will make the magic burst window longer
+                    auto scEffect = PTarget->StatusEffectContainer->GetStatusEffect(EFFECT_SKILLCHAIN, 0);
+                    if (isHelix && scEffect)
+                    {
+                        scEffect->SetDuration(scEffect->GetDuration() + 2000);
```
MowFord, line 1219: > [WinterSolstice8](https://github.com/WinterSolstice8) [2 weeks ago](https://github.com/LandSandBoat/server/pull/5525#discussion_r1581879360) • 
Second thought: should this not get the duration then add 2000? Skillchains are supposed to have reduced duration at least for following skillchains.

CatsEyeXI:
Sorry for the inconvenience, I tried reaching out to you to discuss this, but never got a reply.
image

MowFord:
Even without being deleted, I wouldn't have been able to make changes and push them. 

Suppose I should have put this screenshot instead:
image

Anyways... apologies again to LSB for having to deal with the duplicated effort

zach2good:
Previous PRs for reference:
https://github.com/LandSandBoat/server/pull/5525
https://github.com/LandSandBoat/server/pull/5273
https://github.com/LandSandBoat/server/pull/5188
https://github.com/LandSandBoat/server/pull/5033

src/map/entities/charentity.cpp
zach2good: Is this a completely flat 2s bonus, or does it scale with stats/gear/something else? I know _nothing_ about SCH, and this isn't listed obviously when I google `ffxi immanence`

zach2good (approved):
Code seems fine, just asking about the game logic 👍

src/map/entities/charentity.cpp
MowFord: Yep. Had it confirmed when I made the PR

https://www.bg-wiki.com/index.php?title=Immanence&diff=prev&oldid=698295

zach2good: Delightful, thanks 👍

zach2good (approved):

PR #5728, opened 2024-05-13
[Cleanup] Shift-proof "C" Zones

Frankie-hz:
**_I affirm:_**

Shift proof
Crawlers Nest & [S]
Castle Zvahl Keep
Castle Zvahl Baileys
Castle Oztroja [S]
Castle Oztroja
Carpenters Landing
Cape Terriggan
Caedarva Mire

**Caedarva Mire**
I made a PEALLAIDH_PH_OFFSET call because the IDs were -270 which seemed weird to put there in case an ID shifts. I decided to make the offset the first wild kara which is the PH group instead.

Start server, lotto NMs

### DIFF
--- a/scripts/zones/Caedarva_Mire/IDs.lua
+++ b/scripts/zones/Caedarva_Mire/IDs.lua
@@ -45,37 +45,29 @@ zones[xi.zone.CAEDARVA_MIRE] =
     },
     mob =
     {
+        AYNU_KAYSEY           = GetFirstID('Aynu-kaysey'),
+        CAEDARVA_TOAD         = GetFirstID('Caedarva_Toad'),
         CHIGOES =
         {
             ['Wild_Karakul'] = GetTableOfIDs('Chigoe', 5),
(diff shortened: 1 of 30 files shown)


scripts/zones/Caedarva_Mire/IDs.lua
```diff
@@ -50,32 +50,24 @@ zones[xi.zone.CAEDARVA_MIRE] =
...
-        CAEDARVA_TOAD         = 17101145,
-        JAZARAAT              = 17101146,
-        LAMIA_NO27            = 17101148,
-        MOSHDAHN              = 17101149,
-        KHIMAIRA              = 17101197,
-        VERDELET              = 17101198,
-        TYGER                 = 17101199,
-        MAHJLAEF_THE_PAINTORN = 17101200,
-        EXPERIMENTAL_LAMIA    = 17101201,
+        PEALLAIDH             = GetFirstID('Peallaidh'),
```
claywar: Only request here would be to sort by alpha

Xaver-DaRed (approved):

PR #5731, opened 2024-05-14
[LUA] Add Frozen Mist and Hydro Wave mobskills

### LINKED ISSUE #201: Implement `Ruszor` mob skills, because BLU spell `Sub-Zero Smash` is not learnable
The skill exists in the sql tables, but there is no Lua implementation, so the mob never casts it.

The other skills for the mob, so that the `sub-zero smash` is not cast 100% of the time.

ampitere:
**_I affirm:_**

Adds Frozen Mist and Hydro Wave mobskills. Requires implementing some unique types of Stoneskin, one which absorbs only physical and the other which only absorbs magic, breath, and non-elemental damage (to be done in https://github.com/LandSandBoat/server/pull/5734) Resolves https://github.com/LandSandBoat/server/issues/201

https://www.bg-wiki.com/ffxi/Category:Ruszor

1. !zone Beaucedine Glacier [S]
2. !mobhere 17334289

**Frozen Mist** - https://wiki.ffo.jp/html/21510.html

1. !exec target:useMobAbility(2438)
2. !getmod ICE_ABSORB
3. Effect should absorb 1000 physical damage, while active mob will absorb Ice damage (JP wiki says 1000, bg-wiki says 1500) - will be implemented in https://github.com/LandSandBoat/server/pull/5734

**Hydro Wave** - https://wiki.ffo.jp/html/24705.html

1. !exec target:useMobAbility(2439)
2. !getmod WATER_ABSORB
3. Effect should absorb 1500 magic, breath, and non-elemental damage, while active mob will absorb Water damage - will be implemented in https://github.com/LandSandBoat/server/pull/5734

### DIFF
--- a/scripts/actions/mobskills/frozen_mist.lua
+++ b/scripts/actions/mobskills/frozen_mist.lua
@@ -1,0 +1,28 @@
+-----------------------------------
+--  Frozen Mist
+--  Description: Deals ice damage to enemies around the caster. Additional effect: Terror
+--  Type: Magical (Ice)
+-----------------------------------
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local power    = 30
+    local resist   = xi.mobskills.applyPlayerResistance(mob, xi.effect.TERROR, target, mob:getStat(xi.mod.INT) - target:getStat(xi.mod.INT), 0, 0)
+    local duration = 30 * resist
+
+    xi.mobskills.mobStatusEffectMove(mob, target, xi.effect.TERROR, power, 0, duration)
+
+    local dmgmod = 1.5
+    local info   = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg(), xi.element.ICE, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
+    local dmg    = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.ICE, xi.mobskills.shadowBehavior.WIPE_SHADOWS)
+
+    target:takeDamage(dmg, mob, xi.attackType.MAGICAL, xi.damageType.ICE)
+
+    return dmg
+end
+
+return mobskillObject
--- a/scripts/actions/mobskills/hydro_wave.lua
+++ b/scripts/actions/mobskills/hydro_wave.lua
@@ -1,0 +1,30 @@
+-----------------------------------
+--  Hydro Wave
+--  Description: Deals water damage to enemies around the caster.
+--  Type: Magical (Water)
+-----------------------------------
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local power    = math.random(1, 16)
+    local resist   = xi.mobskills.applyPlayerResistance(mob, xi.effect.ENCUMBRANCE_II, target, mob:getStat(xi.mod.INT) - target:getStat(xi.mod.INT), 0, 0)
+    local duration = 30 * resist
+
+    xi.mobskills.mobStatusEffectMove(mob, target, xi.effect.ENCUMBRANCE_II, power, 0, duration)
+
+    local dmgmod = 2.5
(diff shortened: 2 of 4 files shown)


UmeboshiXI:
The special effects for the Ruszor mobskills are a result of changing the mob's AnimationSub.

AnimationSub:
0 = "Normal"
1 = "Bubbly Effect"
2 = "Frosty/Ice Effect"

scripts/actions/mobskills/hydro_wave.lua
```diff
@@ -0,0 +1,24 @@
...
+-----------------------------------
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local dmgmod = 2.5
+    local info   = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg(), xi.element.ICE, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
```
Frankie-hz: dmgmod looks correct. Element should be water instead of ice

scripts/actions/mobskills/frozen_mist.lua
```diff
@@ -0,0 +1,26 @@
...
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    xi.mobskills.mobStatusEffectMove(mob, target, xi.effect.TERROR, 30, 0, 20)
+
+    local dmgmod = 2
```
Frankie-hz: dmgmod should be 1.5 per Jimmayus retail testing here: https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit#gid=57955395

Duration is most likely 30 seconds base and the difference in duration is a resist state.

image

scripts/actions/mobskills/hydro_wave.lua
ampitere: Refactored.

scripts/actions/mobskills/frozen_mist.lua
ampitere: Refactored the dmgmod. It looks like Terror is always 20 seconds at least from this capture video: https://www.youtube.com/watch?v=mbfAnJNJdoU. I tried testing with `applyResistanceAddEffect` but it always seems to return 1 so the duration ends up always being 30.

src/map/utils/battleutils.cpp
```diff
@@ -5289,7 +5289,7 @@ namespace battleutils
         return damage;
     }
 
-    int32 HandleStoneskin(CBattleEntity* PDefender, int32 damage)
+    int32 HandleStoneskin(CBattleEntity* PDefender, int32 damage, ATTACK_TYPE attackType)
```
ampitere, line 5292: Any preference on attack type vs damage type? Attack type requires less logic but isn't as specific.

scripts/actions/mobskills/hydro_wave.lua
```diff
@@ -0,0 +1,36 @@
...
+-----------------------------------
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local duration = 30
+    local resist = applyResistanceAddEffect(mob, target, xi.element.WATER, 0)
```
ampitere, line 14: This always seems to return 1, resulting in a max duration every mobskill. Am I doing something wrong here?

scripts/actions/mobskills/frozen_mist.lua
```diff
@@ -0,0 +1,36 @@
...
+-----------------------------------
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local duration = 30
+    local resist = applyResistanceAddEffect(mob, target, xi.element.ICE, 0)
```
ampitere, line 14: This always seems to return 1, resulting in a max duration every mobskill. Am I doing something wrong here?

scripts/actions/mobskills/frozen_mist.lua
ampitere: Tested with more magic evasion and was able to get resist values, swapped to `xi.mobskills.applyPlayerResistance` as well for calculating resist.

Xaver-DaRed (approved):

ampitere:
Will need to change these to target self once https://github.com/LandSandBoat/server/pull/5745 is merged.

PR #5732, opened 2024-05-14
Fix Treasure Chest NPC ID reference in Pso'Xja IDs

claywar:
**_I affirm:_**

image

Pso'Xja has 6 mobs named treasure chest, and the 7th and last one is the NPC, in which GetFirstID would return the first mob.  Pulls the entire table and sets to the last value for the expected NPC value, and to resolve the startup warning

No warning, treasure chest NPC appears and works as expected

### DIFF
--- a/scripts/zones/PsoXja/IDs.lua
+++ b/scripts/zones/PsoXja/IDs.lua
@@ -41,7 +41,7 @@ zones[xi.zone.PSOXJA] =
     npc =
     {
         STONE_DOOR_OFFSET       = GetFirstID('_090'),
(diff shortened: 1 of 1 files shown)


Frankie-hz:
I was bound to miss one, thank you! I can go over all the treasure chest to see if I made that mistake somewhere else tonight.

claywar:
At least for mobs, it appears that Pso'Xja is the only zone that has some named Treasure_Chest.  There's a chance that other NPCs might be present though

Xaver-DaRed (approved):

PR #5735, opened 2024-05-14
[LUA] Add Fuscous Ooze's encumbrance effect

ampitere:
**_I affirm:_**

Adds the encumbrance effect to Fuscous Ooze. Duration seems like it should be random between 30 and 45 seconds according to https://wiki.ffo.jp/html/24648.html, not much other information available.

1. !zone Pashhow Marshlands [S]
2. !mobhere 17145979
3. !exec target:useMobAbility(2183)

### DIFF
--- a/scripts/actions/mobskills/fuscous_ooze.lua
+++ b/scripts/actions/mobskills/fuscous_ooze.lua
@@ -1,7 +1,7 @@
 -----------------------------------
 --  Fuscous Ooze
 --  Family: Slugs
---  Description: A dusky slime inflicts encumberance and weight.
+--  Description: A dusky slime inflicts encumbrance and weight.
 --  Type: Magical
 --  Utsusemi/Blink absorb: Ignores shadows
 --  Range: Cone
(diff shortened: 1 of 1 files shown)


scripts/actions/mobskills/fuscous_ooze.lua
```diff
@@ -13,8 +13,11 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    -- TODO: Encumberance seems to do nothing?
-    xi.mobskills.mobStatusEffectMove(mob, target, xi.effect.WEIGHT, 50, 0, 45)
+    local power = math.random(1,16)
```
Xaver-DaRed: `local power = math.random(1, 16)`

Need that space. Also, allign the power and duration while here.

ampitere: Formatted.

Xaver-DaRed (approved):

PR #5737, opened 2024-05-15
Updates httplib dependency to 0.15.3

clanofartisans:
**_I affirm:_**

Updates the `yhirose/cpp-httplib` dependency to the most recent release.

Enable the world HTTP server in `network.lua`, start the world server, hit the endpoints provided and make sure everything works as expected.
* `/api`
* `/api/zones`
* `/api/zones/230`
* `/api/settings`
* `/api/bad_endpoint` (should throw 404)

### DIFF
--- a/ext/CMakeLists.txt
+++ b/ext/CMakeLists.txt
@@ -137,7 +137,7 @@ CPMAddPackage(
 CPMAddPackage(
     NAME cpp-httplib
     GITHUB_REPOSITORY yhirose/cpp-httplib
-    GIT_TAG v0.11.3
+    GIT_TAG v0.15.3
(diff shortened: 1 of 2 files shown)


zach2good:
I'm not against bumping deps versions, but was there a specific reason for this?

clanofartisans:
> I'm not against bumping deps versions, but was there a specific reason for this?

Honestly, we just had something downstream using a different version and I wanted to clean that up. Updating both Horizon and LSB seemed the most expedient way to get everything lined up. I completely understand if that's not a good enough reason to update LSB and won't be surprised if you just close this :)

zach2good (approved):

PR #5745, opened 2024-05-16
New TARGETTYPE flag for wotg-style mobskills

### LINKED ISSUE #4695: 🔨 Forced Use Mobskill Mechanic
**_I affirm:_**

## Describe the feature
Couldn't think of a good descriptor for title.

Some mobs have a mechanic where if a target is out of range, they can still animate and use their Mobskill/Skill Attacks instead of just cancelling when out of range. Other players in range can still get hit if they are in AOE/Conal range.

The video below has a pretty clear demonstration of this.

https://www.youtube.com/watch?v=amQCt-tgEKU&t=146s

MowFord:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/4695 .

Adds a new valid target flag to allow aoe or conal mobskills to hit all entities in range
- When combined with `TARGET_SELF`
  - mob will target itself, then include all enemies in range based on skillflag
  - This allows a skill to not rely in the primary target staying in range
- When combined with `TARGET_ENEMY`
  - Mob will target its battle target, then include all _entities_ in range (mobs and players)
- skillflag 8 includes all entities (mobs and players), new logic handles allegiance filtering based on above choice
- skillflag 0 includes only includes party/alliance/enmity entities, so it doesn't quite work for the `TARGET_ENEMY` option

I was careful to leave all behavior unaffected if the validTargets flag doesn't contain the new flag, `2048`.

Examples from retail:
Ruszor family targets itself so skills can't be out-ranged, the readying action type is different and actiontarget is itself:
image
and of course it doesn't target itself when the mobskill completes
image

combat log of some Ruszor skills being used
image
image

Dark Ixion as well:
image

Finally, Rafflesia have skills that can hit all mobs/players in range:
image

Basically, WoTG mobs have wacky targetfinding for aoe and conals, but it all seems to boil down to either:
- ability targets self to not allow player 1 running away and saving player 2 still in range. But self is excluded from processing the actual ability (mobs don't hit themselves with their offensive abilities)
- ability targets mobs and players in range

In both cases, the targetfinding is mixing allegiances of found targets, hence calling the new flag `TARGET_MOB_AND_PLAYER`

Give a mobskill's `mob_skill_flag` a value of 8 and `mob_valid_targets` either 2049 (ruszor style) or 2052 (Rafflesia style)

Examples from LSB:
Updated some mobskills with completely fake options to show the new behavior
image

image
With the above settings, diamondhide only hits player's partymembers

image
Hard to tell what's going on, but Hydro wave was used with a target out of range, it still went off and hit "Mowtwo" close to it

image
Floral bouquet hitting the players and mobs in range
(Note Floral Bouquet is supposed to "charm" Bees, but I adjusted for testing this PR)

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -2131,7 +2131,19 @@ void CBattleEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
         return;
     }
 
-    uint16 targets = static_cast<uint16>(PAI->TargetFind->m_targets.size());
+    uint16 targets  = static_cast<uint16>(PAI->TargetFind->m_targets.size());
+    auto   skipSelf = false;
+
+    if ((PSkill->getValidTargets() & TARGET_MOB_AND_PLAYER) && (PSkill->getValidTargets() & TARGET_SELF))
+    {
+        // This ability targets self for aoe skills (such as Frozen Mist)
+        // Should be impossible for self to not be in target list, but just in case
+        if (targets > 0)
+        {
+            targets -= 1;
+        }
+        skipSelf = true;
+    }
 
     // No targets, perhaps something like Super Jump or otherwise untargetable
     if (targets == 0)
@@ -2161,6 +2173,14 @@ void CBattleEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
     bool first{ true };
     for (auto&& PTargetFound : PAI->TargetFind->m_targets)
     {
+        if (PTarget == PTargetFound && skipSelf)
+        {
+            // This ability targets self for aoe skills (such as Frozen Mist)
+            // Ignore self completely
+
+            continue;
+        }
+
         actionList_t& list = action.getNewActionList();
 
         list.ActionTargetID = PTargetFound->id;
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -2134,7 +2134,7 @@ void CBattleEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
     uint16 targets  = static_cast<uint16>(PAI->TargetFind->m_targets.size());
     auto   skipSelf = false;
 
-    if ((PSkill->getValidTargets() & TARGET_MOB_AND_PLAYER) && (PSkill->getValidTargets() & TARGET_SELF))
+    if ((PSkill->getValidTargets() & TARGET_ANY_ALLEGIANCE) && (PSkill->getValidTargets() & TARGET_SELF))
     {
         // This ability targets self for aoe skills (such as Frozen Mist)
         // Should be impossible for self to not be in target list, but just in case
@@ -2152,6 +2152,7 @@ void CBattleEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
         actionList.ActionTargetID = id;
 
         actionTarget_t& actionTarget = actionList.getNewActionTarget();
+        actionTarget.messageID       = 0;
 
         if (skipSelf)
         {
@@ -2159,7 +2160,6 @@ void CBattleEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
             // And it found no valid targets in range, the skill and animation should still trigger
             // action.actiontype unchanged
             actionTarget.animation  = PSkill->getAnimationID();
-            actionTarget.messageID  = 0;
             actionTarget.reaction   = REACTION::MISS | REACTION::HIT | REACTION::GUARDED;
             actionTarget.speceffect = SPECEFFECT::SELFAOE_MISS;
         }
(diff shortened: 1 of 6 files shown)


MowFord:
One more change incoming, was researching what should happen if there's no targets in range of the "target self and don't get interupted by range"... turns out it still goes off with no target

image

and the actual packet data for one such move with no targets in range:
image


One more change incoming

Edit: Change complete and tested
https://imgur.com/i8Vi9X0

Xaver-DaRed:
This could be used for flying wyrms landing skills, correct?

MowFord:
> This could be used for flying wyrms landing skills, correct?

Absolutely. Any skill that should trigger even if no targets are in range

UmeboshiXI:
Nice work. This actually solves something I brought up before that can be closed when this gets merged. No more hack methods needed after this. 

https://github.com/LandSandBoat/server/issues/4695

Xaver-DaRed:
Re-opening it to revisit this.

MowFord:
image

Happy to clean this up and go for round 2 of getting it merged

MowFord:
So all seems to still work as expected. To recap from above:
- new `mob_valid_targets` flag (value 2048) to be used with the other standard options
  - When combined with `TARGET_SELF`
    - mob will target itself, then include all enemies in range based on `mob_skill_flag`
    - This allows a skill to not rely in the primary target staying in range
  - When combined with `TARGET_ENEMY`
    - Mob will target its battle target, then include all entities in range (mobs and players)
- `mob_skill_flag` determines what types of entities are included in the aoe. Normally this is limited to offensive or defensive targets, based on the primary target. The new logic makes that a fuzzy concept.
  - `mob_skill_flag` 8 includes all entities (mobs and players), new logic handles allegiance filtering based on above choice
  - `mob_skill_flag` 0 only includes party/alliance/anyone with enmity entities _of the mob/source of skill_, so it doesn't quite work for the TARGET_ENEMY option
    - i.e. only use `mob_skill_flag == 0` when using `mob_valid_targets == 2049` (`TARGET_MOB_AND_PLAYER + TARGET_SELF`)

Here it is on a fresh build after rebase using skill id 2439 twice
- once with mowford in range, getting hit and being processed by the `onMobWeaponSkill` function
- and a second time telling it to use but then running out of range

https://imgur.com/LqlJjAS

src/map/entities/battleentity.cpp
```diff
@@ -2096,20 +2096,46 @@ void CBattleEntity::OnMobSkillFinished(CMobSkillState& state, action_t& action)
...
+            actionTarget.messageID  = 0;
+            actionTarget.reaction   = REACTION::MISS | REACTION::HIT | REACTION::GUARDED;
+            actionTarget.speceffect = SPECEFFECT::SELFAOE_MISS;
+        }
+        else
+        {
+            action.actiontype      = ACTION_MOBABILITY_INTERRUPT;
+            action.actionid        = 28787; // Some hardcoded magic for interrupts
+            actionTarget.animation = 0x1FC; // Hardcoded magic sent from the server
+            actionTarget.messageID = 0;
```
Xaver-DaRed: message Id is the same in both code blocks. Could be moved outside with the rest?

src/map/entities/battleentity.h
```diff
@@ -426,6 +427,7 @@ enum TARGETTYPE : uint16
     TARGET_PET                     = 0x0100,
     TARGET_PLAYER_PARTY_ENTRUST    = 0x0200,
     TARGET_IGNORE_BATTLEID         = 0x0400, // Can hit targets that do not have the same battle ID
+    TARGET_MOB_AND_PLAYER          = 0x0800, // Can hit targets from any allegiance simultaneously.
```
Xaver-DaRed: If its allegiance based, maybe could be named TARGET_ANY_ALLEGIANCE or something of the short.
Since mobs can have the same allegiance as players (summons, trusts etc)

src/map/entities/battleentity.h
```diff
@@ -316,6 +316,7 @@ enum class SPECEFFECT : uint8
 {
     NONE         = 0x00,
     BLOOD        = 0x02,
+    SELFAOE_MISS = 0x04,
```
Xaver-DaRed, line 319: I think I've missed something. Where is this used other than to be set as `actionTarget.speceffect`?

MowFord: image

that's it, it's only used when the aoe move targets itself but doesn't process itself when calling `onMobWeaponSkill`

Edit: if you mean where is it used, it's [here in battleentity.cpp](https://github.com/LandSandBoat/server/blob/c13bd17db1655d51ea67aba4870fdb2278f0b18e/src/map/entities/battleentity.cpp#L2129)

src/map/entities/battleentity.h
MowFord: Good point, done.

Xaver-DaRed:
Fix this styling and this should be gtg
imagen

Xaver-DaRed (approved):

MowFord:
additional testing steps for updated touchdown skills (requested by Xaver):
- go to chasm: `!zone 7`
- goto tiamat: `!gotoname Tiamat`
- spawn tiamat: `!spawnmob 16806227`
- trigger touchdown with reasonable range to test with: `!exec target:useMobSkill(1544)`
  - run the above, then run over 15 yalms away
  - using before the updated sql data will have tiamat ready touchdown, then abruptly stop the animation: https://imgur.com/6PJuha5
  - updating the `mob_valid_targets` values by changing from `4` (target enemy) to `2049` (target self + any allegiance)
  - using the ability and see that it triggers even if player is out of range: https://imgur.com/g5xOsUL
  - and added bonus, using mobability 1282 which has a 6-yalm range (so it triggers even if the target is out of range before firing the skill): https://imgur.com/D82wyUT

MowFord:
Rebased to fix merge conflict in sql file

PR #5750, opened 2024-05-16
[LUA] Fix Geocolure spells shouldn't be castable with GEO sub

ampitere:
**_I affirm:_**

Prevents players who are sub GEO from casting Geocolure spells.

From Xaver: 
> went to retail to check. It seems #shitwikisais strikes again:
> 
> Main-job GEO can use Geo-spells with no handbell.
> Sub-job GEO can't because the spell doesn't even appear on the spell list at all.

1. !changejob 1 99
2. !changesjob 21 99
3. Attempt to cast a Geocolure spell either through text or a macro (it won't display them in the magic menu)

image

### DIFF
--- a/scripts/globals/job_utils/geomancer.lua
+++ b/scripts/globals/job_utils/geomancer.lua
@@ -337,6 +337,9 @@ xi.job_utils.geomancer.geoOnMagicCastingCheck = function(caster, target, spell)
         return xi.msg.basic.ALREADY_HAS_A_PET
     elseif not caster:canUseMisc(xi.zoneMisc.PET) then
         return xi.msg.basic.CANT_BE_USED_IN_AREA
+    elseif caster:getMainJob() ~= xi.job.GEO then
+        -- wikis are incorrect, does not require handbell
+        return xi.msg.basic.MAGIC_CANNOT_CAST
     else
         return 0
     end


Xaver-DaRed:
This fix is incorrect.
The requisite is to have a handbell equiped.

EDIT: I lied.

ampitere:
> This fix is incorrect. The requisite is to have a handbell equiped.

Good call, changed to check for handbell instead.

image

WinterSolstice8 (approved):

Xaver-DaRed:
mmm.

I went to retail to check. It seems #shitwikisais strikes again:
- Main-job GEO can use Geo-spells with no handbell.
- Sub-job GEO can't because the spell doesn't even appear on the spell list at all.

So, sorry I missinformed you... However, I guess the job check WILL need a TODO to remove the Geo-spells from the spell list aswell.

So please:
- Change it so it checks for main job again. I guess the message shouldnt happen, but that one is fine aswell.
- Add a comment stating wikis are worng and no handbell is needd.
- Add a TODO to remove the geo-spells from the spell list IF they appear in the list

ampitere:
> mmm.
> 
> I went to retail to check. It seems #shitwikisais strikes again:
> 
> * Main-job GEO can use Geo-spells with no handbell.
> * Sub-job GEO can't because the spell doesn't even appear on the spell list at all.
> 
> So, sorry I missinformed you... However, I guess the job check WILL need a TODO to remove the Geo-spells from the spell list aswell.
> 
> So please:
> 
> * Change it so it checks for main job again. I guess the message shouldnt happen, but that one is fine aswell.
> * Add a comment stating wikis are worng and no handbell is needd.
> * Add a TODO to remove the geo-spells from the spell list IF they appear in the list

Reverted back to check for main GEO. Added a comment, TODO not necessary as it already does not appear in the spell list.

Xaver-DaRed (approved):

PR #5754, opened 2024-05-16
[cpp/sql] Spellreq setting for main-job-only spells

MowFord:
**_I affirm:_**

Rather than a lua restriction for the issue found [here](https://github.com/LandSandBoat/server/pull/5750), this PR adds a new spellreq setting (not a flag) to define spells only available to main job

import sql and compile PR, see that `/ma geo-regen`, etc doesn't work. It doesn't show up in the menu due to the DATs, but was still possible to cast via macro

image


MowFord:
I actually didn't read amp's PR fully. I didn't realize that was the only thing his was doing... Feel free to close if you prefer his solution

Xaver-DaRed (approved):

Issue #5490, opened 2024-04-22
🐛 Zoning into level sync does wacky things


MowFord:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base`

## Steps to reproduce

- Have a party of 3, 2 in zone 1 not
- Sync party
- have 3rd enter zone as a job higher level than the sync party
- 3rd is level restricted to sync level, but has access to abilities/traits of original level
- 3rd player or sync player leveling up fully resolves the situation

I believe the issue is in `onEffectGain` running `target:levelRestriction(effect:getPower())`, because `CLuaBaseEntity::levelRestriction` has a filter for `if (PChar->status != STATUS_TYPE::DISAPPEAR)`

One solution would be to put the call to `levelRestriction` on a timer to allow the player to fully zone in before applying the restriction

Another might be to remove that status filtering (I'm not quite sure what reason there would be to not execute those functions but still apply the buff)

A third (more in-depth) solution might be to rearrange the process of zoning in. I've noticed weird timing issues with various effects on zoning (not all, but some). Perhaps having a flag for effects in the `status_effect_container` that tells them to not resolve until the player has run through one pass of character loading?

PR #5758, opened 2024-05-16
[CPP] Fix costume effect not lost on zone

ampitere:
**_I affirm:_**

image

Currently if you zone you lose the visible costume but the costume status effect remains. This is due to the check for removing the effect coming after the costume has already been removed resulting in it never reaching that code since `m_Costume` is always 0. Instead we'll check if the player has the effect and if they do remove it.

1. !addeffect 127 644 1000
2. Zone or log out and back in

### DIFF
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -1032,10 +1032,9 @@ void CZone::CharZoneIn(CCharEntity* PChar)
         PChar->StatusEffectContainer->DelStatusEffectSilent(EFFECT_MOUNTED);
     }
 
-    if (PChar->m_Costume != 0)
+    if (PChar->StatusEffectContainer->HasStatusEffect(EFFECT_COSTUME))
     {
-        PChar->m_Costume = 0;
-        PChar->StatusEffectContainer->DelStatusEffect(EFFECT_COSTUME);
+        PChar->StatusEffectContainer->DelStatusEffectSilent(EFFECT_COSTUME);
     }
 
     PChar->ReloadPartyInc();


src/map/zone.cpp
```diff
@@ -1032,9 +1032,8 @@ void CZone::CharZoneIn(CCharEntity* PChar)
         PChar->StatusEffectContainer->DelStatusEffectSilent(EFFECT_MOUNTED);
     }
 
-    if (PChar->m_Costume != 0)
+    if (PChar->StatusEffectContainer->HasStatusEffect(EFFECT_COSTUME))
     {
-        PChar->m_Costume = 0;
         PChar->StatusEffectContainer->DelStatusEffect(EFFECT_COSTUME);
```
ampitere: Should this be silent like the mounted effect removal?

Xaver-DaRed: Probably, however, it's never a bad thing to check.
You can check it yourself though. I saw you opened a PR for the quadav costume.
Use the item, zone and check if there is a message in the chatlog, pls.

ampitere: > Probably, however, it's never a bad thing to check. You can check it yourself though. I saw you opened a PR for the quadav costume. Use the item, zone and check if there is a message in the chatlog, pls.

Looks like it is silently falling off:

image

Updated to be silent.

Xaver-DaRed (approved):

src/map/zone.cpp
```diff
@@ -1032,10 +1032,9 @@ void CZone::CharZoneIn(CCharEntity* PChar)
         PChar->StatusEffectContainer->DelStatusEffectSilent(EFFECT_MOUNTED);
     }
 
-    if (PChar->m_Costume != 0)
+    if (PChar->StatusEffectContainer->HasStatusEffect(EFFECT_COSTUME))
     {
-        PChar->m_Costume = 0;
```
claywar, line 1037: If this is being stored on the char, shouldn't it still be reset here?

ampitere: I can leave it in but it seems to always be 0 by this point in time which is why the previous check was always failing and never removed the status effect.

MowFord: Losing the effect sets it to zero in `onEffectLose`

MowFord:
btw the costume is zero after zoning due to an issue similar [to this](https://github.com/LandSandBoat/server/issues/5490).

When a player is zoning in, their status is `DISAPPEAR`. for both `:levelrestriction` and `:setCostume` bindings, the logic behaves differently for that status, so the effects don't apply properly on zone-in

PR #5759, opened 2024-05-17
[LUA/SQL] Add Quadav Belt costume

ampitere:
**_I affirm:_**

Adds the costume effect for Quadav Belt and makes it usable. Verified in retail the model although I could not get the capture addon to actually run a capture.

image
image

1. !additem quadav_belt
2. Equip the item
3. Use it

### DIFF
--- a/scripts/items/quadav_belt.lua
+++ b/scripts/items/quadav_belt.lua
@@ -1,0 +1,20 @@
+-----------------------------------
+-- ID: 10848
+-- Quadav Belt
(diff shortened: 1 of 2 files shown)


WinterSolstice8:
Does this belt only give you one quadav model? some items have multiple to select from

ampitere:
> Does this belt only give you one quadav model? some items have multiple to select from

Yes I used it multiple times in retail and it was always the same model.

Xaver-DaRed (approved):

PR #5736, opened 2024-05-15
[Cleanup] Shift-proof "B" Zones

Frankie-hz:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?
Shift proof
Bubu Pennisula
Bostaunieux Oubliette
Bhaflau Thickets
Behemoths Dominion
Beaucedine Glacier & [S]
Bearclaw
Beadeaux [S]
Beadeaux
Batallia Downs [S]
Batallia Downs
Bastok mines
Bastok Markets [S]

## Steps to test these changes
Lotto NMs/start server see no lookup errors

PR #5760, opened 2024-05-17
[QUEST] For the Birds

jamesbradleym:
**_I affirm:_**

Implements quest For the Birds
Modified from ASB with additional/corrected dialog, proper mob kill check, corrected ki cycle, 

[Based on ASB](https://github.com/AirSkyBoat/AirSkyBoat/commit/d4bdd529983ae7ded29e3f916b47326aa081bceb)
[Capture](https://www.youtube.com/watch?v=i-s-RkHWDBw)


scripts/zones/Beadeaux/IDs.lua
```diff
@@ -56,6 +59,13 @@ zones[xi.zone.BEADEAUX] =
         {
             [17379625] = 17379626, -- 139.642 -2.445 161.557
         },
+        FOR_THE_BIRDS_MOBS =
+        {
+            17379784,
+            17379785,
+            17379786,
+            17379787,
+        }
```
Xaver-DaRed: This PR is going to need a rebase once this PR (#5736) gets merged.

This block of logic will also have to be re-done to not use raw IDs (see that PR to see an example on how to do this)

jamesbradleym: Sounds good. 👍🏼

As far as using GetFirstID/not using raw IDs, it looks like these mobs (Nickel and Magnus Quadav) don't have unique names, do you have an example in mind for how to implement in this case?

Frankie-hz: `FOR_THE_BIRDS_MOBS = GetFirstID('Magnes_Quadav') -- For The Birds Quest`

Then inside the quest lua you can do
```
local mobListTable =
{
    id.mob.FOR_THE_BIRDS_MOBS,
    id.mob.FOR_THE_BIRDS_MOBS + 1,
    id.mob.FOR_THE_BIRDS_MOBS + 2,
    id.mob.FOR_THE_BIRDS_MOBS + 3,
}
```
Then just replace all your `FOR_THE_BIRDS_MOBS` with `mobListTable` or whatever you think the table should be called

jamesbradleym: Hmm, okay so I think I just misunderstood what GetFirstID was doing, this is limited to the zone for which the ID.lua file is for which would prevent conflicts with, for example, Magnes_Quadav in Beadeaux [s]?

claywar: Correct, this only pulls from the same zone, but does not discriminate between mobs and NPCs

scripts/quests/otherAreas/For_the_Birds.lua
```diff
@@ -0,0 +1,211 @@
...
+local beadeauxID = zones[xi.zone.BEADEAUX]
+-----------------------------------
+local quest = Quest:new(xi.questLog.OTHER_AREAS, xi.quest.id.otherAreas.FOR_THE_BIRDS)
+
+quest.reward =
+{
+    item = xi.item.JAGUAR_MANTLE,
+}
+
+local mobListTable =
```
claywar, line 19: You have the offset, and know you're going to go from 0 to 3 every time, no need to explicitly define the table.  Instead, it would be cleaner for the "AllDead" function to iterate `for mobId = beadeauxID.mob.FOR_THE_BIRDS_MOB_OFFSET, beadeauxID.mob.FOR_THE_BIRDS_MOB_OFFSET + 3`

scripts/quests/otherAreas/For_the_Birds.lua
```diff
@@ -0,0 +1,211 @@
...
+
+    return true
+end
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_AVAILABLE and
+            player:hasCompletedQuest(xi.questLog.OTHER_AREAS, xi.quest.id.otherAreas.MISSIONARY_MOBLIN) and
```
claywar: Please indent on multiline returns
```
return something something and
    something else
```

scripts/quests/otherAreas/For_the_Birds.lua
```diff
@@ -0,0 +1,211 @@
...
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_ACCEPTED
+        end,
+
+        [xi.zone.OLDTON_MOVALPOLOS] =
+        {
+            ['Koblakiq'] =
+            {
+                onTrigger = function(player, npc)
+                    local prog = quest:getVar(player, 'Prog')
```
claywar: Can we set a standard here? Most examples use `questProgress`

scripts/quests/otherAreas/For_the_Birds.lua
```diff
@@ -0,0 +1,211 @@
...
+
+local mobListTable =
+{
+    beadeauxID.mob.FOR_THE_BIRDS_MOB_OFFSET,
+    beadeauxID.mob.FOR_THE_BIRDS_MOB_OFFSET + 1,
+    beadeauxID.mob.FOR_THE_BIRDS_MOB_OFFSET + 2,
+    beadeauxID.mob.FOR_THE_BIRDS_MOB_OFFSET + 3,
+}
+
+local function AllDead(mobs)
```
claywar: Please start local functions with a lowercase, only globals should use upper

scripts/zones/Beadeaux/IDs.lua
```diff
@@ -29,18 +29,22 @@ zones[xi.zone.BEADEAUX] =
...
-        GE_DHA_EVILEYE      = GetFirstID('GeDha_Evileye'),
-        MIMIC               = GetFirstID('Mimic'),
-        ZO_KHU_BLACKCLOUD   = GetFirstID('ZoKhu_Blackcloud'),
+        BI_GHO_HEADTAKER          = GetFirstID('BiGho_Headtaker'),
+        DA_DHA_HUNDREDMASK        = GetFirstID('DaDha_Hundredmask'),
+        GA_BHU_UNVANQUISHED       = GetFirstID('GaBhu_Unvanquished'),
+        GE_DHA_EVILEYE            = GetFirstID('GeDha_Evileye'),
+        MIMIC                     = GetFirstID('Mimic'),
+        ZO_KHU_BLACKCLOUD         = GetFirstID('ZoKhu_Blackcloud'),
+        FOR_THE_BIRDS_MOB_OFFSET  = GetFirstID('Magnes_Quadav'),
```
claywar: I've been trying to get away from the `_OFFSET` naming convention, we know its the First ID with the name, so I would suggest just calling it `MAGNES_QUADAV`

scripts/zones/Beadeaux/npcs/GeFhu_Yagudoeye.lua
```diff
@@ -10,7 +10,7 @@ entity.onTrade = function(player, npc, trade)
 end
 
 entity.onTrigger = function(player, npc)
-    player:startEvent(124)
+    player:event(124)
```
claywar: This isn't valid.  There is no `CLuaBaseEntity::event()` function

scripts/zones/Castle_Oztroja/npcs/Daa_Bola_the_Seer.lua
```diff
@@ -10,7 +10,7 @@ entity.onTrade = function(player, npc, trade)
 end
 
 entity.onTrigger = function(player, npc)
-    player:startEvent(86)
+    player:event(86)
```
claywar: Same as above

scripts/zones/Oldton_Movalpolos/npcs/Koblakiq.lua
```diff
@@ -10,7 +10,7 @@ entity.onTrade = function(player, npc, trade)
 end
 
 entity.onTrigger = function(player, npc)
-    player:startEvent(13)
+    player:event(13)
```
claywar: Same as above

claywar (changes requested):

scripts/zones/Oldton_Movalpolos/npcs/Koblakiq.lua
jamesbradleym: Done ✅

scripts/zones/Castle_Oztroja/npcs/Daa_Bola_the_Seer.lua
jamesbradleym: Done ✅

scripts/zones/Beadeaux/npcs/GeFhu_Yagudoeye.lua
jamesbradleym: Done ✅

scripts/zones/Beadeaux/IDs.lua
jamesbradleym: Done ✅

scripts/quests/otherAreas/For_the_Birds.lua
jamesbradleym: Done ✅

scripts/quests/otherAreas/For_the_Birds.lua
jamesbradleym: Done ✅

scripts/quests/otherAreas/For_the_Birds.lua
jamesbradleym: Done ✅

scripts/quests/otherAreas/For_the_Birds.lua
jamesbradleym: Done ✅

Changed the for loop, left the table for the popFromQM call

Xaver-DaRed:
I have some questions:
- What happens if you loose/warp out the quadav fight? You lost the KI presumably. Can you get it again? And if you dont, does talking to the npc after that retrigger the fight anyway?
- Shouldnt the prog variable when killing all 4 quadavs only progress if you dont have the KI? Or else, another player could just join the fight with the KI and it wouldnt be ever taken away.

I did some research and it does seem each party member must individually check the npc to each trigger the fight. So only the one poping the mobs is the one getting the credit.

I think this needs another capture for the edge-cases and fail status.

jamesbradleym:
> I have some questions:
> 
> * What happens if you loose/warp out the quadav fight? You lost the KI presumably. Can you get it again? And if you dont, does talking to the npc after that retrigger the fight anyway?
> * Shouldnt the prog variable when killing all 4 quadavs only progress if you dont have the KI? Or else, another player could just join the fight with the KI and it wouldnt be ever taken away.
> 
> I think this needs another capture

The player can return to Koblakiq to receive another Glittering Fragment if they lose or warp out of the Quadav fight (line 85)

Interesting point, however if it's changed so you progress if you don't have the KI, then another player could join the fight without the KI (for instance if they recently failed a fight) and progress without having it?

I suppose it would be simple enough to do a check if the player has the KI at the end of the fight and if so remove it?

Xaver-DaRed:
It should not progress for the rest of the party members, is my point. Only the one that pops the NMs gets to progress the quest.

So an additional check is needed on mob death. I'll point it out... One sec

jamesbradleym:
> It should not progress for the rest of the party members, is my point. Only the one that pops the NMs gets to progress the quest.
> 
> So an additional check is needed on mob death. I'll point it out... One sec

Ah, I see. So then would it be appropriate to set a var on GeFhu for last pop to indicate who would progress from mob death?

scripts/quests/otherAreas/For_the_Birds.lua
```diff
@@ -0,0 +1,211 @@
...
+
+            ['Nickel_Quadav'] =
+            {
+                onMobDeath = function(mob, player, optParams)
+                    if
+                        quest:getVar(player, 'Prog') == 2 and
+                        allDead()
+                    then
+                        quest:setVar(player, 'Prog', 3)
+                    end
```
Xaver-DaRed:
```lua
                    if
                        quest:getVar(player, 'Prog') == 2 and
                        not player:hasKeyItem(xi.ki.GLITTERING_FRAGMENT) and
                        allDead()
                    then
                        quest:setVar(player, 'Prog', 3)
                    end
```
This can still be cheesed. We may need to set a local var on the mobs on pop with the player ID of the one who poped them so we can check it on mob death aswell, so only that player that poped them can progress the quest.

What we dont want:
- 2 players in party with KIs. Player 1 pops and they fail. Player 1 has no KI.
- Player 2 pops. They both kill them. Since player 1 doesnt have the KI, he would still get credit even though he wasnt the poper this time.

jamesbradleym: Maybe something like this?

```
local function assignPlayer()
    for mobId = beadeauxID.mob.MAGNES_QUADAV, beadeauxID.mob.MAGNES_QUADAV + 3 do
        local mob = GetMobByID(mobId)
        mob:setLocalVar('PLAYERID', player:getID())
    end
end
```

jamesbradleym: And on mob death checking:

```
if
    quest:getVar(player, 'Prog') == 2 and
    allDead() and
    not player:hasKeyItem(xi.ki.GLITTERING_FRAGMENT) and
    mob:getLocalVar('PLAYERID') == player:getID()
then
    quest:setVar(player, 'Prog', 3)
end
```

jamesbradleym:

PR #5784, opened 2024-05-19
[lua/sql] Sugaar NM

MowFord:
**_I affirm:_**

Implements [Sugaar](https://www.bg-wiki.com/ffxi/Sugaar)

NM is very simple, though both of his moves are missing. I will do them [in a separate PR](https://github.com/LandSandBoat/server/pull/5785).

As the PH code is unchanged, no need to test it (but I did and confirmed it works)

goto pashow_s
`!gotoname Sugaar`
`!spawnmob 17145867`
witness fight and drops

Note that since the skills only exist in sql, he won't use them via the lua logic. Simply copy any other skill lua twice and name to the proper mob_skill_names: `grim_glower` and `torpefying_charge`

image


MowFord:
Realized I'll need a family mixin for one of these skills, so just going to do it in this PR. Converting to draft until i get the skills coded

MowFord:
Grim glower:
image

Torpefying Charge:
image

Edit: 
Going to remove draft status as this other PR implements the two skills and should not affect this PR at all

https://github.com/LandSandBoat/server/pull/5785

Xaver-DaRed (approved):

PR #5786, opened 2024-05-19
[core] Always store pet ID when saving pet info

WinterSolstice8:
**_I affirm:_**

Apparently zoning with an automaton caused a fire elemental to appear because it's pet ID was set to zero.

Zone with an automaton and keep the automaton.

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -464,6 +464,7 @@ void CCharEntity::setPetZoningInfo()
     {
         return;
     }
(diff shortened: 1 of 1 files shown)


MowFord:
I think you need more than just the petid though?

I made [this PR](https://github.com/LandSandBoat/server/pull/5531/files) to resolve the fire elemental when zoning

There was a crash condition that is unrelated to the fix in that PR.

claywar (approved):

PR #5787, opened 2024-05-19
[cpp] Add lua binding to set a mob's list of stealable items

MowFord:
**_I affirm:_**

Creates a new mob lua binding to set the stealable item. It's written in such a way that it could be easily extended to support passing a table, but only supports a single itemID for now.

`mob:setStealItem(xi.item.DANCESHROOM)` would
- Reset the mob's flag to be able to be stolen from
- loop through a mob's drop list, remove all stealable items, showing debug prints when it does
- insert the new stealable item

Error handling might be excessive, but exits early if:
- entity is not a mob
- droplist is null (i highly doubt we'd ever need this binding on a mob that has dropid 0)
- binding parameter doesn't map to a valid item

This binding is setting the framework for the steal patterns of WoTG funguar mobs, which have a steal item based on their animation sub

go to a mob:
- `!exec target:setStealItem(xi.item.DANCESHROOM)`
- `!changejob thf 99`
- `/ja steal` and `!reset` until you steal and item
- do steal a few more times to confirm it cannot be stolen from
- do binding again with a different item and see that you can now steal that item

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -16849,6 +16849,79 @@ void CLuaBaseEntity::addTreasure(uint16 itemID, sol::object const& arg1, sol::ob
 }
 
 /************************************************************************
+ *  Function: setStealItem()
+ *  Purpose : Used to reset the mob's steal table
+ *  Example : mob:setStealItem(xi.item.ITEM_NAME)
+ ************************************************************************/
+
+void CLuaBaseEntity::setStealItem(sol::object const& arg1)
+{
+    if (m_PBaseEntity->objtype != TYPE_MOB)
+    {
+        ShowWarning("Attempting to set steal item for invalid entity type (%s).", m_PBaseEntity->getName());
+        return;
+    }
+
+    // TODO support arg1 as a table
+    std::vector<uint16> Items;
+    if (arg1.is<uint16>())
+    {
+        auto item = arg1.as<uint16>();
+        if (itemutils::GetItemPointer(item))
+        {
+            Items.emplace_back(item);
+        }
+    }
+
+    if (Items.size() == 0)
+    {
+        ShowWarning("Attempting to set steal items with no valid itemids (%s).", m_PBaseEntity->getName());
+        return;
+    }
+
+    auto* PMob = dynamic_cast<CMobEntity*>(m_PBaseEntity);
+
+    if (PMob)
+    {
+        DropList_t* PDropList = itemutils::GetDropList(PMob->m_DropID);
+
+        if (!PDropList)
+        {
+            ShowWarning("Attempting to set steal item for mob with no drop table (%s).", m_PBaseEntity->getName());
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


MowFord:
Logs from running `!exec print(mob:getStealItem())`, stealing an item, then printing it again:
image

Xaver-DaRed (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -16848,6 +16848,79 @@ void CLuaBaseEntity::addTreasure(uint16 itemID, sol::object const& arg1, sol::ob
...
+
+void CLuaBaseEntity::setStealItem(sol::object const& arg1)
+{
+    if (m_PBaseEntity->objtype != TYPE_MOB)
+    {
+        ShowWarning("Attempting to set steal item for invalid entity type (%s).", m_PBaseEntity->getName());
+        return;
+    }
+
+    // TODO support arg1 as a table
```
zach2good, line 16865: Can you implement the table handling while you're here please?

src/map/lua/lua_baseentity.cpp
```diff
@@ -16848,6 +16848,79 @@ void CLuaBaseEntity::addTreasure(uint16 itemID, sol::object const& arg1, sol::ob
...
+    {
+        ShowWarning("Attempting to set steal items with no valid itemids (%s).", m_PBaseEntity->getName());
+        return;
+    }
+
+    auto* PMob = dynamic_cast<CMobEntity*>(m_PBaseEntity);
+
+    if (PMob)
+    {
+        DropList_t* PDropList = itemutils::GetDropList(PMob->m_DropID);
```
zach2good, line 16886: Is this modifying the _global droplist_ for this mob, which might be shared among other mobs?

I think that's the concern I have with this PR. I did a droplist modification PR at some point which handles this similarly by keeping track of a modification list per-mob, so the global entry isn't touched.

zach2good (changes requested):

src/map/lua/lua_baseentity.cpp
MowFord: It.... is

Definitely the wrong way to go here, oops. Got a link to that other PR?

zach2good: Looks like my thing was replaced by these:
https://github.com/LandSandBoat/server/pull/3379
https://github.com/LandSandBoat/server/pull/3430

Where there's now a listener for `ITEM_DROPS`, which could go in a mixin, if relevant etc.

zach2good: This was the original, just for funsies:
https://github.com/topaz-next/topaz/pull/2557

MowFord: Amazing. Looks like this listener is called when a mob dies and processes drops alongside the normal dropid? Should simplify the coppercap PR quite a lot.

Thanks. I'll just close this then and refactor the coppercap PR

PR #5788, opened 2024-05-20
[lua] Funguar s mixin

MowFord:
**_I affirm:_**

Creates a mixin for funguar in wotg zones to properly emulate their steal behavior.
- stolen item is based on the currently-remaining bulbs on the head.
  - 3-bulb: WOOZYSHROOM
  - 2-bulb: DANCESHROOM
  - 1-bulb: AGARICUS_MUSHROOM
  - 0-bulb: nothing to steal
- a bulb is lost via a successful steal or the existing use of the respective mobskill
- Also the funguar in wotg zones have similar regen-during-weather behavior as the slug family

All three items above were confirmed in retail

Leaving as a draft until #5787 is merged, as this PR won't function without it

Compile the new binding in #5787 then utilize this PR and 
- `!exec mob:useMobAbility(X)` where X is 310, 311, 312 to get familiar with the animation sub behavior
- `!exec mob:setAnimationSub(0)` then `!mob:setMobMod(xi.mobMod.VAR, 0)` to re-sync (or just kill the mob and respawn it)
- do combinations of steal and the respective mobskill to see that the steal behavior is consistent with retail
  - note that explicitly calling `useMobAbility(` bypasses the skill check, so it will not care what the current bulbs are and the animation will be weird

### DIFF
--- a/scripts/enum/item.lua
+++ b/scripts/enum/item.lua
@@ -2545,6 +2545,7 @@ xi.item =
     DRAGON_FRUIT                        = 5662,
     LYNX_MEAT                           = 5667,
     AILEENS_DELIGHT                     = 5674,
+    AGARICUS_MUSHROOM                   = 5680,
     ANGLERS_CASSOULET                   = 5704,
     HOMEMADE_SALISBURY_STEAK            = 5705,
     DISH_OF_HOMEMADE_CARBONARA          = 5706,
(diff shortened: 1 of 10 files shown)


MowFord:
Well I made a mistake with my other PR's binding. Didn't consider that it was adjusting the global droplist. I will have to pivot into probably a listener that gets called in getStealItem to override the drop table. Simplifies the logic in the family mixin as it would only have to check the animationsub when a steal happens instead of updating it in all the various ways animationsub could change

Might as well make one for getDespoilItem as well

PR #5789, opened 2024-05-20
[lua] Mobskill animationsub changes

MowFord:
**_I affirm:_**

Working on #5788 I noticed that the three funguar mobskills that remove bulbs from the head were not calling `setAnimationSub`. This causes an inconsistency in a player rendering the mob:
- When a funguar animates an `Xshroom` mobskill, the client handles cleanly removing the bulb and sending it to the target
- as long as the player stays in range of the funguar, the model of the funguar is kept consistent, but if you log or go out of range it resets back to animationSub 0 

This PR resolves the situation by creating a generic mixin that can be added to any mob's lua that should get these updates.
- A mixin was chosen as Xaver requested it for consistency
  - Action scripts handle damage, messages, etc
  - Effect scripts handle stats, mods, etc
  - Mixins handle generic mob things: animations, aggro, etc
- This PR is being submitted not as a draft because:
  - while it isn't a complete solution for all mobskills that change animation sub (it is actually a regression for all but the funguar skills, which already didn't update animationSub), the scope of adding a mixin to every mob that uses the 15 mobskills is massive. I would like to create a PR per mob type that relies on this PR as a guide/framework

Note: the PR is split into 4 commits to make reviewing easier. Seeing what changes in mobskills were reflected in the mixin was a bit confusing without the separation.

for funguars
https://imgur.com/a/vr3UtC5

### DIFF
--- a/scripts/actions/mobskills/axe_throw.lua
+++ b/scripts/actions/mobskills/axe_throw.lua
@@ -1,11 +1,12 @@
 -----------------------------------
---  Axe Throw
+-- Axe Throw
 --
---  Description: Ranged attack with the equipped weapon, which is lost.
---  Type: Ranged
---  Utsusemi/Blink absorb: 1 shadow
---  Range: 7.0
---  Notes: Only used by armed BST Mamool Ja
+-- Description: Ranged attack with the equipped weapon, which is lost.
+-- Type: Ranged
+-- Utsusemi/Blink absorb: 1 shadow
+-- Range: 7.0
+-- Notes: Only used by armed BST Mamool Ja
+-- Animation sub is changed by the mixin: mobskill_animation_sub.lua
 -----------------------------------
 local mobskillObject = {}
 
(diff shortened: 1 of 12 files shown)


MowFord:
> ~~I'm also going through and checking the impact to other mobs, but I suspect this will at worst have the same behavior as before, and at best give a nice delay in forcing a mob model update when using `setAnimationSub`~~

> Edit: of course SE isn't consistent with this, so yes it breaks other things. Will remove draft mode when I've got it working properly

Moved this out of the OP as new code simplifies the changes to be explicitly requested by an optional parameter to `setAnimationSub(`

MowFord:
The premise of this PR is accurate, but the solution is fundamentally flawed as hp change, etc all trigger an entity update.

I think since there's only ~13 mobskills so far coded to change animationSub, a helper function to trigger the animationSub change when the mobskill completes would work nicely. Got most of the work done already, will finalize it and add to this PR (mostly weapon break mobskills are left with this behavior)

Edit: Spoke with Xaver and completely pivoted. PR is now marked as Ready again with an updated description

PR #5790, opened 2024-05-20
[LUA/SQL] Add Carpenters' Landing barge

### LINKED ISSUE #97: Carpenters' Landing Barge does not function correctly
- [x] I have searched existing [issues](https://github.com/DerpyProjectGroup/topaz/issues) to see if the issue has already been opened
- [x] I have read the [Contributing Guide](https://github.com/DerpyProjectGroup/topaz/blob/info/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: Any

The  Carpenters' Landing Barge that takes players to the Phanauet Channel zone appears to be missing its instructions, so it's likely lost somewhere despite Chauie's or Felourie's insistence that it's now arriving. In fact, the random dock NPCs may be hollow shells needing work as well.

ampitere:
**_I affirm:_**

Adds Carpenters' Landing barge interactions and transport records. Based on a combination of [ASB's implementation of the barge](https://github.com/AirSkyBoat/AirSkyBoat/blob/staging/scripts/globals/barge.lua), manaclipper.lua, and retail captures:

https://www.youtube.com/watch?v=1Mxu0rIvjC8
https://www.youtube.com/watch?v=JER9lZEIQrA

https://www.bg-wiki.com/ffxi/Phanauet_Channel

Each trip removes 1 use/ticket regardless of landing, if you have no ticket you'll get booted and a unique cutscene will play depending on which landing you're at. The docking order is as follows: Central > South > North > Central > South. Upon arriving at the North or South Landings the boat will drop passengers off and then exit without letting anybody on. A new boat will then appear and allow players to get on coming from the same direction that the previous boat left.

Timings could be a bit better but this is a good start. NPCs on the boat still need work as do the NMs, although one seems to be implemented already.

Resolves https://github.com/LandSandBoat/server/issues/97

North Landing    !pos -302 -2 505 2
Central Landing  !pos -137 -2 60 2
South Landing    !pos 230 -2 -530 2

1. Use one of the landing positions
2. Wait for the boat
3. Get on boat, get kicked off for not having a ticket
4. Buy a ticket
5. Get on boat again, cutscene should play and you should zone into Phanauet Channel
6. Ride the boat, cutscene will play once you zone out and you should be placed at the proper landing


scripts/zones/Phanauet_Channel/DefaultActions.lua
```diff
@@ -0,0 +1,6 @@
+local ID = zones[xi.zone.PHANAUET_CHANNEL]
+
+return {
+    ['Eunirange'] = { messageSpecial = ID.text.ARE_WE_THERE_YET },
```
ampitere: These lines of text should be spoken by the NPC but I couldn't figure out what event to use.

Currently looks like:
image

But should look like:
image

As well as Riche:
image

image

scripts/globals/barge.lua
```diff
@@ -0,0 +1,266 @@
...
+        local gil = player:getGil()
+        if option == 1 and (currentTicket == 1 or currentTicket == 3) then -- If you have BARGE_TICKET
+            -- Event auto plays the correct message
+        elseif option == 1 then
+            if gil >= 50 then
+                player:delGil(50)
+                player:addKeyItem(xi.ki.BARGE_TICKET)
+                player:messageSpecial(ID.text.KEYITEM_OBTAINED, xi.ki.BARGE_TICKET)
+            else
+                player:messageSpecial(7440) -- too poor
```
ampitere: The proper message when you don't have enough gil is this message but I couldn't get the NPC to respond with it in the event so this currently looks like:

image

But the NPC should only respond with the too poor message. It should also come from the NPC similar to the Riche and Eunirange messages.

ampitere:
It seems like some of the timings for leaving the barge/channel zone don't line up exactly with retail (when it triggers the transport event), is there any way to tweak that? This mostly affects arriving at the North and South landings when the boat is out of service and restarts.

Frankie-hz:
> It seems like some of the timings for leaving the barge/channel zone don't line up exactly with retail (when it triggers the transport event), is there any way to tweak that? This mostly affects arriving at the North and South landings when the boat is out of service and restarts.

I did the barge for ASB (over 1.5 years ago at this point) and trying to get anything to line up, even with proper captures is very difficult. Its all in the sqls with time offsets, waiting, anim arrive, anim depart. I got them as close as I could but if you can get them even better more power to ya 💪

scripts/zones/Phanauet_Channel/DefaultActions.lua
Frankie-hz: If its an event it would be in the capture.

ampitere: image

Interesting it doesn't seem to have an event, it's just called NPC chat and provides the text ID.

ampitere: Woops that was the other tonberry but it's the same regardless:

image

MowFord: `npc:showText(7217)` or something similar. Using the variable instead of a magic number of course

MowFord: Sorry based on [this](https://github.com/LandSandBoat/server/blob/0a4fbb6117291db0ff4dfbd4f6183b83dd8ac002/src/map/lua/lua_baseentity.h#L54) you probably want 

`player:showText(npc, 7217, timeLeft)` etc

ampitere: That does work, it turns them towards you and they get stuck in that direction but it's better than nothing I suppose.

scripts/globals/barge.lua
ampitere: Was able to get the NPC to say it using `player:showText()` but you still get the thank you message before the too poor message.

MowFord: I am not at a computer to check, but it could be due to default actions being called because the onTrigger isn't starting an event

PR #5791, opened 2024-05-20
[CPP/LUA] Add delallinventory and delcontaineritems commands

ampitere:
**_I affirm:_**

Adds commands to quickly delete a user's entire inventory or specified containers. Useful when testing and you don't want to throw away every item in your current inventory/container which can be very tedious.

1. Fill your inventory with items
2. Run !delinventory
3. Inventory should be empty

1. Fill a container with items
2. Run !delcontaineritems (you'll need to provide the container's id)
3. Container should be empty

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -3926,7 +3926,7 @@ bool CLuaBaseEntity::addItem(sol::variadic_args va)
  *  Function: delItem()
  *  Purpose : Deletes an item from a player's inventory
  *  Example : player:delItem(4102, 12)
- *  Notes   : Can specify contianer using third variable
+ *  Notes   : Can specify container using third variable
  ************************************************************************/
 
 bool CLuaBaseEntity::delItem(uint16 itemID, int32 quantity, sol::object const& containerID)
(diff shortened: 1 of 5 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -3958,6 +3958,48 @@ bool CLuaBaseEntity::delItem(uint16 itemID, int32 quantity, sol::object const& c
...
+        ShowWarning("Invalid entity type calling function (%s).", m_PBaseEntity->getName());
+        return;
+    }
+
+    uint8 location = containerID.get_type() == sol::type::number ? containerID.as<uint8>() : 0;
+
+    if (location >= CONTAINER_ID::MAX_CONTAINER_ID)
+    {
+        ShowWarning("Lua::delContainerItems: Attempting to delete items from an invalid container. Defaulting to main inventory.");
+        location = CONTAINER_ID::LOC_INVENTORY;
```
CatsEyeXI: What are your thoughts on just bailing out completely if the container is invalid, preventing the end user from emptying the wrong container?

ampitere: Yeah, it makes more sense to just return and not default it for them. Changed.

scripts/commands/delinventory.lua
```diff
@@ -0,0 +1,38 @@
+-----------------------------------
+-- func: delcontaineritems
```
zach2good: PR title, filename, descriptor etc. are all out of line -> `delContainerItems`

zach2good:
I don't know what the pattern is that we have in the codebase, but `delallX` or `clearX` would make more sense to me for nuking your whole inventory. and `delX` would take an argument to nuke a certain slot?

I have no horses in this race, open to hear what everyone else thinks

scripts/commands/delinventory.lua
ampitere: Updated to reflect it now being `delinventory`. Although I could add a second command like the original that takes the container number as a param.

ampitere:
> I don't know what the pattern is that we have in the codebase, but `delallX` or `clearX` would make more sense to me for nuking your whole inventory. and `delX` would take an argument to nuke a certain slot?
> 
> I have no horses in this race, open to hear what everyone else thinks

I mostly based the naming off the existing `delitem` command and method. I think `delallinventory` makes sense, I also added back the `delcontaineritems` command which supports taking a specified container.

zach2good (approved):

PR #5793, opened 2024-05-20
[core] Audit blacklist commands

WinterSolstice8:
**_I affirm:_**

* Rename blacklist packets
* Add support for client requesting blacklist
* Fix issue with blacklists only working on last 12 (page size 12) blacklisted characters

blist < 12 characters, see it work
blist 12 characters, see it work
blist > 12 characters, see it work


WinterSolstice8:
P.S. github thinks I deleted and created a new file, but git tracked it was renamed.
image

Xaver-DaRed (approved):

PR #5794, opened 2024-05-20
[lua] Antlion animations

MowFord:
**_I affirm:_**

This PR solves a few issues:
- Antlion ambush can sometimes leave a mob hidden (if the mobskill doesn't go off)
- Antlion ambush doesn't look like what it [should based on retail](https://youtu.be/6TYCqwkoxqE?t=224).
  - Before this PR, the mob gets out of the ground and immediately has a visible name. It also kinda pops out then blips and digs its way out due to the mobskill setting animationsub(1) before the mobskill animation goes off
- Finally, the Tuchulcha sandpit move allowed players to know his next location
  - the position is not sent to the client if status is set to `INVISIBLE`
  - confirmed via retail videos that he does, indeed disappear without animation after using sandpit, so this behavior is unchanged

Big thing to test is that the ENM "Sheep in antlion's clothing" functions properly:
- `!zone 8`
- `!togglegm` to get the enm option
- `!togglegm` again to be able to aggro the antlions
- search around and see that ambush still functions properly, Tuchulcha still uses sandpit and runs away, etc

Another place to test would be Attohwa Chasm:
- `!zone 7`
- `!gotoname Burrow_Antlion 1`

There's two mobs with the no_aggro version of the mixin, they are spawned during the Rostrum Pumps event:
- `!additem ANTLION_TRAP`
- `!gotoid 16806296`
- trade the trap and go through [the fight](https://www.bg-wiki.com/ffxi/Alastor_Antlion). See that Alastor and Executioner use pit ambush properly

Finally, here's a few videos showing off the various situations
https://imgur.com/a/XicUnSL

### DIFF
--- a/scripts/mixins/families/antlion_ambush.lua
+++ b/scripts/mixins/families/antlion_ambush.lua
@@ -6,6 +6,8 @@ g_mixins = g_mixins or {}
 g_mixins.families = g_mixins.families or {}
 
 g_mixins.families.antlion_ambush = function(antlionAmbushMob)
+    local pitAmbush = 278
+
     antlionAmbushMob:addListener('SPAWN', 'ANTLION_AMBUSH_SPAWN', function(mob)
         mob:hideName(true)
         mob:setUntargetable(true)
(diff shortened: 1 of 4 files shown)


scripts/mixins/families/antlion_ambush.lua
```diff
@@ -2,6 +2,8 @@
 
 require('scripts/globals/mixins')
 
+local pitAmbush = 278
```
Xaver-DaRed: Define this inside the listener its actually used in.
Or at least, inside the actual mixing

MowFord: it's used in two

MowFord: Edit: done

Xaver-DaRed (approved):

PR #5803, opened 2024-05-21
[LUA] Fix mobMagicalMove doesn't return hits landed

ampitere:
**_I affirm:_**

There are several mobskills that use `mobMagicalMove` then attempt to use the hitslanded from the return but `mobMagicalMove` does not use return any hitslanded resulting in it always being nil. This issue currently affects the following mobskills:

- Auroral Uppercut
- Decussate
- PW Decussate
- Knuckle Sandwich

An alternative solution could be to remove the hitlanded from each skill, let me know what you'd prefer.

1. !exec target:useMobAbility(1798)

### DIFF
--- a/scripts/actions/mobskills/decussate.lua
+++ b/scripts/actions/mobskills/decussate.lua
@@ -20,7 +20,7 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local dmgmod = 1.2
     local info = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg() * 3, xi.element.EARTH, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
-    local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.EARTH, math.random(2, 3) * info.hitslanded)
+    local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.EARTH, math.random(2, 3))
 
     target:takeDamage(dmg, mob, xi.attackType.MAGICAL, xi.damageType.EARTH)
     return dmg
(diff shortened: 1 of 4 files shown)


Xaver-DaRed:
I would guess, being magical moves, that it doesnt return hits landed because magic never misses nor has multiple hits.

It sounds like those moves are simply using that function wrong.

I also dont see the need of having this logic exist in the first place... Its wrong, it doesnt floor dmg after each multiplier is applied, it doesnt account for many magical stuff like weather.... This system is horrible all around.

None of this is your issue though.

I would say to edit those skills, seing as those skills are the exception to the rule.

ampitere:
> I would guess, being magical moves, that it doesnt return hits landed because magic never misses nor has multiple hits.
> 
> It sounds like those moves are simply using that function wrong.
> 
> I also dont see the need of having this logic exist in the first place... Its wrong, it doesnt floor dmg after each multiplier is applied, it doesnt account for many magical stuff like weather.... This system is horrible all around.
> 
> None of this is your issue though.
> 
> I would say to edit those skills, seing as those skills are the exception to the rule.

The only reason I went this route first was because of the comment in the `mobMagicalMove` method:

`-- magical mob skills are single hit so provide single Melee hit TP return if primary target`

Changed the method back and removed hitslanded from the mobskills. I could not find shadow interaction for Knuckle Sandwich and Auroral Uppercut so just left it as ignore for now.

scripts/actions/mobskills/decussate.lua
```diff
@@ -20,7 +20,7 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local dmgmod = 1.2
     local info = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg() * 3, xi.element.EARTH, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
-    local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.EARTH, math.random(2, 3) * info.hitslanded)
+    local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.EARTH, math.random(2, 3))
```
Frankie-hz, line 23: The `info.hihtslanded` should be removed anyway but just as a note (does not have to be in this PR) math.random for magical moves seems extremely odd. As far as I am aware this is a huge AOE skill for both NMs that use this. It probably wipes all shadows like all the other magical moves (assuming its magical). We found a lot of skills that were listed wrong the wiki

WinterSolstice8 (approved):

PR #5811, opened 2024-05-22
Add Move Speed latent as Item Mod

### LINKED ISSUE #4766: 🐛 getMaxGearMod() does not account for latent effects
**_I affirm:_**

All/Any

## Branch affected by issue

`base`

## Steps to reproduce
Equip a nation Aketon. With the speed rework, gear uses said function to get the highest value of a specific modifier, comparing each piece of gear one by one, calling `getModifier()` (defined in item_equipment.cpp)

We do not have a similar function for "get latent if latent should apply"

## Expected behavior

If a latent SHOULD apply, `getMaxGearMod()` should return the value of the requested modifier.

### LINKED ISSUE #5810: 🐛 Movement Speed Latent Effect Broken
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

Add Destrier Beret:

```
!additem 11811
```

Equip on job under level 31, note unchanged speed.

## Expected behavior

Latent effects should be checked for speed enhancements.

cocosolos:
**_I affirm:_**

Adds `Mod::MOVE_SPEED_GEAR_BONUS` to list of mods that should apply to the item instead of the player. 

Fixes #4766
Fixes #5810

Under 31:
image
Over 30:
image

```
!additem 11811
!changejob war 30
/equip head "Destrier beret"
!changejob war 31
```

### DIFF
--- a/src/map/latent_effect.cpp
+++ b/src/map/latent_effect.cpp
@@ -114,7 +114,8 @@ bool CLatentEffect::ModOnItemOnly(Mod modID)
         modID == Mod::ITEM_ADDEFFECT_ELEMENT ||
         modID == Mod::ITEM_ADDEFFECT_STATUS ||
         modID == Mod::ITEM_ADDEFFECT_POWER ||
-        modID == Mod::ITEM_ADDEFFECT_DURATION)
+        modID == Mod::ITEM_ADDEFFECT_DURATION ||
+        modID == Mod::MOVE_SPEED_GEAR_BONUS)
     {
         return true;
     }
--- a/src/map/latent_effect_container.cpp
+++ b/src/map/latent_effect_container.cpp
@@ -523,6 +523,8 @@ void CLatentEffectContainer::CheckLatentsJobLevel()
         {
             case LATENT::JOB_MULTIPLE:
             case LATENT::JOB_MULTIPLE_AT_NIGHT:
+            case LATENT::JOB_LEVEL_BELOW:
+            case LATENT::JOB_LEVEL_ABOVE:
                 return ProcessLatentEffect(latentEffect);
                 break;
             default:
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -5934,8 +5934,10 @@ void CLuaBaseEntity::changeJob(uint8 newJob)
 
         PChar->resetPetZoningInfo();
 
+        charutils::RemoveAllEquipMods(PChar);
         PChar->jobs.unlocked |= (1 << newJob);
         PChar->SetMJob(newJob);
+        charutils::ApplyAllEquipMods(PChar);
 
         if (newJob == JOB_BLU)
         {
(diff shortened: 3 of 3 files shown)


cocosolos:
Aketon works as well, and doesn't stack with the beret.

WinterSolstice8 (approved):

cocosolos:
Found an issue with this when wearing and removing multiple pieces of latent speed gear. Looking into it.

cocosolos:
I'm not sure what's causing this issue and won't be able to look into it until tomorrow, but here's how to reproduce:

1. Be in a starting city, level 30 or lower.
2. Equip nation aketon THEN destrier beret.
3. Set level to 31.
4. Remove aketon THEN beret.
5. Re-equip beret, item still has speed mod.

Latents need to be refreshed more often for time and level based latents, so that might end up fixing this issue. Currently you need to remove the item for it to lose its mod (Destrier beret level, ninja kyahan time). But this might also be unrelated to that, since this particular issue only seems to happen when adding and removing items in a certain order.

If anyone has any idea what's going on let me know, otherwise I'll try to investigate more tomorrow.

MowFord:
This is likely the result of using `!changejob` vs using a nomad moogle or gaining xp naturally. You could probably emulate it via `!addexp` to confirm but i'm almost positive this isn't an issue under normal circumstances.

You confirmed in your OP that using `!changejob` doesn't refresh latents so it makes sense there'd be this edge case.

cocosolos:
> This is likely the result of using `!changejob` vs using a nomad moogle or gaining xp naturally. You could probably emulate it via `!addexp` to confirm but i'm almost positive this isn't an issue under normal circumstances.
> 
> You confirmed in your OP that using `!changejob` doesn't refresh latents so it makes sense there'd be this edge case.

I actually did test by manually adding exp and I still had to re-equip to get the latent to remove. I don't think any of the existing latent update functions cover this latent so should be easy to fix. I'm more concerned about the item itself not losing the mod under certain circumstances.

cocosolos:
Okay I added "level above and below" to `CheckLatentsJobLevel` so it gets checked on exp gain and used `charutils::RemoveAllEquipMods` and `charutils::ApplyAllEquipMods` around some Lua bindings instead of directly calling `CheckLatentsJobLevel`. I think either approach works but this is the pattern used in `levelRestriction` so just went with that. Looks like everything works as expected now.

Xaver-DaRed:
Question: What happens when you equp, say, an Destrier beret under lvl 30 with a kupo suit with this change?

cocosolos:
It correctly only uses one of them, and you can mess around with changing your level and adding/removing different combos and it should all work correctly. Same with ninja kyahan at night gives the bigger boost and takes precedence over any of the 12% pieces. Kupo suit should override beret since all the max mod stuff already works correctly, it's just these latents were applying to the player instead of the gear like other movement speed items.

Xaver-DaRed (approved):

PR #5814, opened 2024-05-22
[CPP/SQL] Fix Shield Mastery item mod values

ampitere:
**_I affirm:_**

https://wiki.ffo.jp/html/3316.html
https://www.bg-wiki.com/ffxi/Shield_Mastery

Shield Mastery scales by the value of the trait (n) multiplied by 10 (n*10) so the `SHIELD_MASTERY_TP` mod should also be in multiples of 10. The Reverence Gauntlets +1 were also missing this modifier: https://www.bg-wiki.com/ffxi/Rev._Gauntlets_%2B1

There is also a check for the trait that should look for the mod instead most likely so that it can apply to all jobs rather than just PLD, WAR, and RDM who receive the trait.

1. !changejob 7 99
2. !additem 23180 or 23515 or 27970
3. !getmod SHIELD_MASTERY_TP, 99 PLD have based 40 from traits


sql/item_mods.sql
```diff
@@ -49721,7 +49721,7 @@ INSERT INTO `item_mods` VALUES (23180,31,36);    -- MEVA: 36
 INSERT INTO `item_mods` VALUES (23180,68,32);    -- EVA: 32
 INSERT INTO `item_mods` VALUES (23180,163,-600); -- DMGMAGIC: -600
 INSERT INTO `item_mods` VALUES (23180,384,400);  -- HASTE_GEAR: 400
-INSERT INTO `item_mods` VALUES (23180,485,1);    -- SHIELD_MASTERY_TP: 1
+INSERT INTO `item_mods` VALUES (23180,485,1);    -- SHIELD_MASTERY_TP: 10
```
MowFord, line 49724: You didn't change the actual mod value in all 3 entries

ampitere: That would definitely be helpful, fixed.

src/map/utils/battleutils.cpp
```diff
@@ -2140,7 +2140,7 @@ namespace battleutils
 
                         // Shield Mastery
                         if ((std::max(damage - (PDefender->getMod(Mod::PHALANX) + PDefender->getMod(Mod::STONESKIN)), 0) > 0) &&
-                            charutils::hasTrait((CCharEntity*)PDefender, TRAIT_SHIELD_MASTERY))
+                            PDefender->getMod(Mod::SHIELD_MASTERY_TP))
```
ampitere, line 2143: The other similar checks are looking at the mod not the trait which makes sense since you don't necessarily need to have the trait for the equipment bonus to apply. For example: https://www.bg-wiki.com/ffxi/Adoubeur%27s_Pavise

MowFord: Would make sense as this is similar to `STORE_TP`, which doesn't require the job trait to provide the bonus

sql/item_mods.sql
MowFord: I... can't resolve threads, but cool

src/map/utils/battleutils.cpp
MowFord: Got confirmation from retail:
image

sql/item_mods.sql
```diff
@@ -77499,6 +77499,7 @@ INSERT INTO `item_mods` VALUES (27970,31,26);    -- MEVA: 26
 INSERT INTO `item_mods` VALUES (27970,68,22);    -- EVA: 22
 INSERT INTO `item_mods` VALUES (27970,163,-500); -- DMGMAGIC: -500
 INSERT INTO `item_mods` VALUES (27970,384,400);  -- HASTE_GEAR: 400
+INSERT INTO `item_mods` VALUES (27970,485,10);    -- `SHIELD_MASTE`RY_TP: 10
```
Xaver-DaRed, line 77502: Align this properly and remove that stray ` <--

ampitere: Fixed.

Xaver-DaRed (approved):

sql/item_mods.sql
```diff
@@ -77499,6 +77499,7 @@ INSERT INTO `item_mods` VALUES (27970,31,26);    -- MEVA: 26
 INSERT INTO `item_mods` VALUES (27970,68,22);    -- EVA: 22
 INSERT INTO `item_mods` VALUES (27970,163,-500); -- DMGMAGIC: -500
 INSERT INTO `item_mods` VALUES (27970,384,400);  -- HASTE_GEAR: 400
+INSERT INTO `item_mods` VALUES (27970,485,10);   -- SHIELD_MASTE`RY_TP: 10
```
MowFord, line 77502:
```suggestion
INSERT INTO `item_mods` VALUES (27970,485,10);   -- SHIELD_MASTERY_TP: 10
```

ampitere: FIxed.

zach2good (approved):

PR #5815, opened 2024-05-22
[CPP] Update useMobAbility to check distance

0x05010705:
**_I affirm:_**

This resolves an edge-case where if a mob is issued the command `mob:useMobAbility(####)`, no distance check is performed on the ability being used.  This can cause the mob to ready a skill while the player is already out of range and results in the mob pseudo-stunning itself.  In cases where a mob may be issued multiple calls to `useMobAbility(####)` in quick succession, the mob becomes locked in place preventing it from moving or performing any other actions.

The function itself, in its current iteration, provides no guarantee that the mob will successfully use the ability specified as an argument unless distance checks are performed within that same mob script.  This update does not change this consideration, but rather instead of the mob stunning itself for 3s, it will fail the distance check and just continue without performing the skill.

** Before Rebuilding **
1 - Move a character to any zone with mobs
2 - Find a mob and engage combat with it.
3 - Use `!exec target:useMobAbility(470)`  (or any mob skill)
4 - Update your movement speed `!speed 200`
5 - Run away from the mob and issue `!exec target:useMobAbility(470)` while you are out of range (15y+)
6 - Note that the mob will still "Ready" the skill.
7 - Repeatedly issue the `useMobAbility` command and note that the mob becomes stunlocked.

** After Rebuilding **
 - Repeat the steps as listed above
 - After reaching Step 5, note that the mob no longer attempts to "Ready" the skill if you are out of range
 - After reaching Step 7, note that the mob no longer becomes stun locked -- It instead will continue pathing to the target.
 
 Special Thanks to @TracentEden for helping me develop/update this.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -16587,15 +16587,32 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
     // clang-format off
     m_PBaseEntity->PAI->QueueAction(queueAction_t(0ms, true, [PTarget, skillid, PMobSkill](auto PEntity)
     {
-        if (PTarget)
+        auto mobObj = dynamic_cast<CMobEntity*>(PEntity);
+
+        // has both a valid target (specified by user and mob)
+        if (PTarget && mobObj)
         {
-            PEntity->PAI->MobSkill(PTarget->targid, skillid);
+            float currentDistance = distance(mobObj->loc.p, PTarget->loc.p);
+            if (currentDistance <= PMobSkill->getDistance())
+            {
+                PEntity->PAI->MobSkill(PTarget->targid, skillid);
+            }
         }
-        else if (dynamic_cast<CMobEntity*>(PEntity))
+        // does not have a specified target so default to current battle target
+        else if (mobObj)
         {
             if (PMobSkill->getValidTargets() & TARGET_ENEMY)
             {
-                PEntity->PAI->MobSkill(static_cast<CMobEntity*>(PEntity)->GetBattleTargetID(), skillid);
+                auto defaultTarget = mobObj->GetBattleTarget();
+                if (defaultTarget)
+                {
+                    // check distance from player or mob will use TP move and 'lock' itself
+                    float currentDistance = distance(mobObj->loc.p, defaultTarget->loc.p);
+                    if (currentDistance <= PMobSkill->getDistance())
+                    {
+                        PEntity->PAI->MobSkill(defaultTarget->targid, skillid);
+                    }
+                }
             }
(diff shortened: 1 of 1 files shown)


zach2good:
>  This can cause the mob to ready a skill while the player is already out of range and results in the mob pseudo-stunning itself.

Whats the behaviour on retail? They won't fire the skill if the player is out of range and just keep chasing, and not lose their TP? What relation does this have to kiting, etc.

0x05010705:
So perhaps a more in-depth description

On retail, mobs do not ready a TP Skill on a player unless the player is within range.  A mob can ready a skill while the player is in range, but then fail because the player has moved out of range by the end of the wind-up time.

If a mob has no valid targets for a TP move, but does have access to a self-target skill or buff skill, then it will/should opt for that skill.

This is currently how things work in LSB as well.

However, there are two different cases of useMobAbility based on whether an argument is presented or not.
1 - useMobAbility()  --- No argument
The logic performed when this is called on a mob acts like so:
 - Grab the ability list for this mob from the database.
 - Shuffle the order of the skills that were found.
 - Pick the first skill in the shuffled list and check if targets are in range.
   - If there is a target in range, then ready/use this skill
   - If there is no target in range of this skill, move to the next skill and repeat
     - If the skill is a self-buff (IE: Pollen on Bees), it stops here
 - If the end of the skill list is reached, and no targets are in range, then do not use a skill and continue pathing to player.
 - No further attempts are made to use a skill - The checks have failed on all skills.

*Side Note: This is slightly different than if a mob has TP.  In this case on the next server tick, the mob performs the checks again over and over until it has a skill with a valid target in range (pathing towards the player/target if the ability firing fails).*

2 - useMobAbility(####)  --- Argument Supplied is a skill ID from mob_skills.sql
The logic performed when this is called on a mob acts like so:
 - Immediately ready the skill, ignoring whether or not I have a target in range.

In both cases, neither guarantees that a mob ability will be used successfully.  In the first case, it will just keep pathing to the player.  In the second case, currently, it will end up 'stunning' the mob because it will ready the skill and then fail even when a target was not in range when it went to ready the skill.

This PR attempts to remedy the 2nd case only in that it will match the behavior of the 1st case where no targets are available for any skill (aka: It will just continue pathing to the player and not use the specified skill).

The implementation of the `drawIn` and `getAbilityDistance` functions from PR #5833 are meant to help with remedying this lack of assurance when calling `useMobAbility` with an argument.  When `useMobAbility` is called within a script, it would now be possible for the scripter to check for the distance of the ability they are wanting to use and ensure that there is a target within range before calling it, drawing in the target (if required), or just queue it within the lua script instead of messing with the mobs action queue within core.

WinterSolstice8 (approved):

zach2good (approved):

PR #5816, opened 2024-05-22
[core] new luautil function: getEntitiesInRange()

MowFord:
**_I affirm:_**

Sparked by [this comment](https://github.com/LandSandBoat/server/pull/4868#discussion_r1440575060), I've created a more generic binding for finding entities in range of an entity, using the existing `CTargetFind` class.

I'm putting this in draft mode while I test it, but preliminary testing seems to work (the benefits of using pre-existing systems)

This PR lays some groundwork:
- Creates a missing enum essential to skills/abilities: AOE_TYPE
- Copies all targetfind enums into lua space, with some comments about their caveats

The PR's main change: `getEntitiesInRange()`
- This binding first does basic checks and returns an empty table if anything essential is missing with the parameters
- It then calls a `PAI->Targetfind` flow similar to spells or mobskills
- the return is always a table, empty if no targets were found within the given targetfind parameters

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -10237,6 +10237,100 @@ void CLuaBaseEntity::recalculateAbilitiesTable()
 }
 
 /************************************************************************
+ *  Function: getEntitiesInRange()
+ *  Purpose : Returns a Lua table of entities within range of the base entity
+ *  Example : local players = npc:getEntitiesInRange(target, aoeType, radiusOrigin, distance, findFlags, validTargets)
+ *
+ *  WARNING : APART FROM THIS ONE EXCEPTION, ALL TARGET AND PATH FINDING SHOULD BE DONE EXCLUSIVELY IN CORE
+ *
+ *  Notes   : This is primarily made for Dark Ixion's trample, but there may be valid uses for other mob sequences
+ *               IN GENERAL, this function should not be used in persistent on-tick logic
+ *            if the passed distance is nil or 0 (as well as many other combinations), empty table will be returned
+ *            examples
+ *      mob:getEntitiesInRange(nearbyPlayer, xi.aoeType.ROUND, xi.aoeRadius.ATTACKER, trampleRange, xi.findFlag.HIT_ALL, xi.targetType.PLAYER + xi.targetType.PLAYER_PARTY)
+ *              The parameters should all be enums from xi.aoeType, xi.aoeRadius, xi.findFlag, and xi.targetType
+ ************************************************************************/
+
+auto CLuaBaseEntity::getEntitiesInRange(CLuaBaseEntity* PLuaEntityTarget, sol::variadic_args va) -> sol::table
+{
+    auto* baseEntity = dynamic_cast<CBattleEntity*>(m_PBaseEntity);
+    auto  entities   = lua.create_table();
+    if (!baseEntity || !baseEntity->PAI || !baseEntity->PAI->TargetFind)
+    {
+        ShowWarning("Invalid entity type calling function (%s).", m_PBaseEntity->getName());
+        return entities;
+    }
+
+    auto* PTarget = dynamic_cast<CBattleEntity*>(PLuaEntityTarget->GetBaseEntity());
+    if (!PTarget)
+    {
+        ShowWarning("Invalid target entity in function (%s).", m_PBaseEntity->getName());
+        return entities;
+    }
+
+    // Reset targetfind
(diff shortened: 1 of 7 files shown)


MowFord:
targetfind is funny about conals in that it always adds the primary target my work in #5745 will make it more reliable if one needed to use this binding for raw conal findings, but in general the binding seems to work well.

It's written explicitly to be used for Dark Ixion, so I'm going to start poking at it there

MowFord:
I've done some testing and it seems to do what's on the tin here. Ixion functions with the new binding, but will benefit greatly from the extra `TARGETTYPE` in #5745

src/map/lua/lua_baseentity.cpp
```diff
@@ -10236,6 +10236,98 @@ void CLuaBaseEntity::recalculateAbilitiesTable()
...
+ *               radiusOrigin: xi.aoeRadius.ATTACKER
+ *               findFlags: xi.findFlag.HIT_ALL
+ *               findFlags: xi.findFlag.ALLIANCE | xi.findFlag.PET | xi.findFlag.IGNORE_BATTLEID
+ *               validTargets: xi.targetType.ENEMY | xi.targetType.PLAYER_PARTY
+ ************************************************************************/
+
+sol::table CLuaBaseEntity::getEntitiesInRange(CLuaBaseEntity* PLuaEntityTarget, sol::variadic_args va)
+{
+    auto* baseEntity = dynamic_cast<CBattleEntity*>(m_PBaseEntity);
+    auto  players    = lua.create_table();
```
zach2good: If this is targetfind-based, it can find more than players depending on the flags, no?

src/map/lua/lua_baseentity.h
```diff
@@ -514,8 +514,9 @@ class CLuaBaseEntity
     uint32 canLearnSpell(uint16 spellID);
     void   delSpell(uint16 spellID);
 
-    void recalculateSkillsTable();
-    void recalculateAbilitiesTable();
+    void       recalculateSkillsTable();
+    void       recalculateAbilitiesTable();
+    sol::table getEntitiesInRange(CLuaBaseEntity* PLuaEntityTarget, sol::variadic_args va);
```
zach2good: `auto getEntitiesInRange(CLuaBaseEntity* PLuaEntityTarget, sol::variadic_args va) -> sol::table;`

src/map/lua/lua_baseentity.cpp
```diff
@@ -10236,6 +10236,98 @@ void CLuaBaseEntity::recalculateAbilitiesTable()
     PChar->pushPacket(new CCharAbilitiesPacket(PChar));
 }
 
+/************************************************************************
+ *  Function: getEntitiesInRange()
+ *  Purpose : Returns a Lua table of entities within range of the base entity
+ *  Example : local players = npc:getEntitiesInRange(target, aoeType, radiusOrigin, distance, findFlags, validTargets)
+ *  Notes   : if the passed distance is nil or 0 (as well as many other combinations), empty table will be returned
```
zach2good: I would add a _big fat note_ here, saying something to the tune of: "APART FROM THIS ONE EXCEPTION, ALL TARGET AND PATH FINDING SHOULD BE DONE EXCLUSIVELY IN CORE"

zach2good: and say that this is used for Ixion, and possibly there's another usage for this in assault? I don't remember

src/map/lua/lua_baseentity.cpp
MowFord: Good catch, yes. Started from my old binding. I'll update the variable

src/map/lua/lua_baseentity.cpp
MowFord: Can do

zach2good: And another sub-note: If you have to do expensive operations that are meant for core in Lua, you should be doing them as infrequently as possible: In a mobskill that locks itself off, or anything that isn't on-tick, etc.

src/map/lua/lua_baseentity.cpp
MowFord: Done

src/map/lua/lua_baseentity.cpp
MowFord: Updated the description

zach2good (approved):

zach2good:
For any passers-by: This is logic added using `TargetFind`. Future optimisations using octrees/spatial hashing will happen INSIDE `TargetFind`.

PR #5821, opened 2024-05-22
[Cleanup] Correct mobskill magical redundancies

Xaver-DaRed:
**_I affirm:_**

Mobskill/Summon magical multipliers where stupidly divided among several functions, resulting in several multipliers being repeatedly applied, or not applied at all, depending on the amount of functions used.

This:
- Makes calculations used the magic rewrite new functions.
- Makes 1 function handle all multipliers, which is the function used for all magical skills, another the magic burst, which councidentally is the function used for the only skills that CAN magic burst, and the final one the final operations (phalanx, shadows, one for all, stoneaskin, etc)
- Audits several petskills. Including a magical petskill being treated as physical.

Play, summon, use magical skills and take cover

To review:
- The important changes are located in the global scripts. The individual skill scripts are mostly style and consistency

### DIFF
--- a/scripts/globals/mobskills.lua
+++ b/scripts/globals/mobskills.lua
@@ -7,6 +7,7 @@
 require('scripts/globals/combat/magic_hit_rate')
 require('scripts/globals/magicburst')
 require('scripts/globals/magic')
+require('scripts/globals/spells/damage_spell')
 require('scripts/globals/utils')
 -----------------------------------
 xi = xi or {}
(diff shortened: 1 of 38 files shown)


scripts/globals/mobskills.lua
```diff
@@ -265,59 +265,58 @@ end
...
 
-    finaldmg       = finaldmg * resist * magicDefense
-    returninfo.dmg = finaldmg
+    -- Calculate final damage.
+    finalDamage = math.floor(finalDamage * sdt)
+    finalDamage = math.floor(finalDamage * resist)
+    finalDamage = math.floor(finalDamage * dayAndWeather)
+    finalDamage = math.floor(finalDamage * magicBonusDiff)
+    finalDamage = math.floor(finalDamage * targetMagicDamageAdjustment)
+    finalDamage = math.floor(finalDamage * damageModifier)
```
Frankie-hz, line 309: It looks like (I may have missed one) all these damage modifiers are 1? Do we want to just remove this completely. Not even sure what this is

Xaver-DaRed: I havent checked all 250+ scripts that use this function modifier

Xaver-DaRed: The PR is already big enough as it is already though. If it turns out to always be 1, I'll remove it in a different PR.

For now, the main focus was to fix the pressing issue: MAB and TMDA being applied twice while, at the same time, TMDA and weather multipliers not being used for most magical mobskills

Frankie-hz (approved):

MowFord (commented):
Changes look sane, scanned over all files but didn't pull locally to test.

zach2good (approved):

PR #5822, opened 2024-05-23
[Quest]RNG AF1-3 Series Conversion IF

hooksta4:
**_I affirm:_**

I realize this is a very large PR. My logic behind putting them all into one, besides it being a series, is the good chunk of the quests revolve around Perih Vashai, Koh Lenbalalako, Kapeh Myohrye, and Muhk Johldy. It seemed logical for testing to do them all at once. 

Converts the RNG AF quests 1-3 series to Interaction Framework.
- Scavenge now uses Quest framework var format and enum'd zone and item.
- Added a quest complete section and a check for next quest because the dialog changes on complete quest after each one except for the last. 
- Converts Sin Hunting (AF1) to IF. 
    - Moves default dialog to DefaultActions where appropriate. 
- Converts Fire and Brimstone (AF2) to IF
   - Adds dialogs to NPC's from near by NPC's that was previously not there.
   - Adds zoning requirement before next quest is available.
   - Players were previously able to select any Gravestone and complete that portion of the quest. They now will get correct dialog if they chose the incorrect one. Only allowing the correct Gravestone to produce the progressing event.  
   - Players must have zoned from Batallia Downs as per capture in order to get the cutscene. 
- Converts Unbridled Passion (AF3)
   - Added dialog for different parts of mission based on progress.
   - Based on capture, depending on where player zoned in from, changes the cs but still progresses the same.
   - Allows players to still complete the quest even if they don't gather the 99 ice arrows that are optional.
-  Adjusted The Fanged One to the correct dialog after quest complete but before accepting any other quest as the NPC dialog changes based on progress in the quest series.
- Removed all NPC lua logic invovled with the quest series.

[Capture](https://www.youtube.com/watch?v=qPQwnp1gkMY&feature=youtu.be)
[Capture](https://www.youtube.com/watch?v=zSgfuHEfZ8g)
[Capture](https://www.youtube.com/watch?v=nmwUEdLbqdg)
[Capture](https://www.youtube.com/watch?v=pbowyq3ekHw)
Set job to RNG level 50 or higher and complete the quests. The info has all POS for particular points of interest.

### DIFF
--- a/src/map/latent_effect_container.cpp
+++ b/src/map/latent_effect_container.cpp
@@ -432,13 +432,15 @@ void CLatentEffectContainer::CheckLatentsHours()
 * activates them if the conditions are met.                             *
 *                                                                       *
  ************************************************************************/
-void CLatentEffectContainer::CheckLatentsPartyMembers(size_t members)
+void CLatentEffectContainer::CheckLatentsPartyMembers(size_t members, size_t trustCount)
 {
-    ProcessLatentEffects([this, members](CLatentEffect& latentEffect) {
+    ProcessLatentEffects([this, members, trustCount](CLatentEffect& latentEffect) {
+        size_t totalMembers = members + trustCount;
+
         switch (latentEffect.GetConditionsID())
         {
             case LATENT::PARTY_MEMBERS:
-                if (latentEffect.GetConditionsValue() <= members)
+                if (latentEffect.GetConditionsValue() <= totalMembers)
                 {
                     return latentEffect.Activate();
                 }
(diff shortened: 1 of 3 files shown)


Frankie-hz:
All RNG AF quests have been converted on HXI since May of last year with 0 bug reports/issues that we are aware of.

claywar:
Definite rebase required here, as the commit history doesn't align with the intent of the PR at all.

hookstar:
Yea - I attempted to rebase and it went to shit somehow - will try to fix

PR #5824, opened 2024-05-23
[Quest]A Minstrel in Despair to IF

hooksta4:
**_I affirm:_**

Converts A Minstrel In Despair to IF. Places it in the hiddenQuests.

Complete The Old Monument. Trade the Poetic Parchment to Mertaire.
Quest will now show as completed in log.

### DIFF
--- a/scripts/quests/jeuno/A_Minstrel_In_Despair.lua
+++ b/scripts/quests/jeuno/A_Minstrel_In_Despair.lua
@@ -1,0 +1,47 @@
+-----------------------------------
+-- A Minstrel in Despair
+-----------------------------------
+-- Log ID: 3, Quest ID: 12
+-- Mertaire : !pos -17 0 -61 245
+-----------------------------------
+local quest = Quest:new(xi.questLog.JEUNO, xi.quest.id.jeuno.A_MINSTREL_IN_DESPAIR)
+
+quest.reward =
+{
+    fame     = 30,
+    fameArea = xi.fameArea.JEUNO,
+    gil      = 2100,
+}
+
+quest.sections =
+{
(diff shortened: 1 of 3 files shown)


scripts/quests/hiddenQuests/A_Minstrel_In_Despair.lua
```diff
@@ -0,0 +1,47 @@
+-----------------------------------
+-- A Minstrel in Despair
+-----------------------------------
+-- Log ID: 3, Quest ID: 12
+-- Mertaire : !pos -17 0 -61 245
+-----------------------------------
+local quest = Quest:new(xi.questLog.JEUNO, xi.quest.id.jeuno.A_MINSTREL_IN_DESPAIR)
```
zach2good: This isn't a hidden quest, since it is tracked by a quest log entry. All the hidden quests DONT have a quest entry at any time, and are tracked by char vars (`local quest = HiddenQuest:new('TrustIngrid')` etc.).

This one is only slightly different in that it isn't visible until it's complete. Move it to the jeuno folder.

Presumably because it's only a single step and it doesn't contain `quest:begin` it should act like the wikis say

zach2good (changes requested):

scripts/quests/hiddenQuests/A_Minstrel_In_Despair.lua
hooksta4: That's a great point. It is in the right place on my implementation on hxi and asb - must have saw hidden and then put in hidden quest. Thanks, moved.

zach2good (approved):

PR #5826, opened 2024-05-24
[LUA/SQL] Add Divine Interference quest and instance

ampitere:
**_I affirm:_**

Adds Divine Interference quest and instance. Created based on multiple captures, videos, and wiki information:

https://www.bg-wiki.com/ffxi/Divine_Interference
https://wiki.ffo.jp/html/19593.html
https://www.youtube.com/watch?v=fkIgQxGs9qE (note the first 2 CS in this capture are actually for Unwavering Resolve)
https://www.youtube.com/watch?v=tEDcCLBs9Yk
https://drive.google.com/file/d/1aBexVwCy0inuYlDCfj8Ng_6JpCJWLnNG/view (shows the correct trigger area for starting)
https://www.youtube.com/watch?v=tTrwLPHaRUU

1. !addmission 4 47 (Eternal Merecenary)
2. !addquest 6 74 (Waking the Colossus)
3. !completequest 6 74
4. Naja Salaheem, !pos 22.700 -8.804 -45.591 50
5. Imperial Whitegate, !pos 152.8295 -2.2 0.0613 50
6. !additem slab_of_plumbago 3 (alternatively !addkeyitem lightning_cell but you still need to trigger the CS at the door)
4. Acid-Eaten Door, !pos 271.9045 -32 -88 61
7. Blank Lamp, !pos 209.3953 0.05 20.0207 72
8. Runic Seal, !pos 125.4569 0 19.9914 72
9. Naja Salaheem, !pos 22.700 -8.804 -45.591 50

To test just the instance use !instance 7702

## TODO

- [ ] Get win CS working, doesn't play for some reason (Alzadaal Undersea Ruins - 309 can be seen at this timestamp https://www.youtube.com/watch?v=fkIgQxGs9qE&t=2498s this also affects 307 which is the CS for Waking the Colossus instance win)
- [x] Balance the fight, feels a bit overtuned for 75 right now
- [x] Review mob skills
- [x] One of the params for event 930 (potentially others as well?) is the player's current mercenary rank. Currently says Private Second Class which is from capture data


scripts/quests/ahtUrhgan/Divine_Interference.lua
```diff
@@ -0,0 +1,235 @@
...
+                        return quest:progressEvent(308)
+                    end
+                end,
+            },
+
+            onZoneIn =
+            {
+                function(player, prevZone)
+                    if quest:getVar(player, 'Prog') == 4 then
+                        return 309 -- correct cs from capture but doesn't play for some reason?
```
ampitere: Can't figure out why this cutscene doesn't work, tried passing it more params but doesn't seem to make a difference. Tried !cs and !exec player:startEvent. Event can be seen in a capture here: https://www.youtube.com/watch?v=fkIgQxGs9qE&t=2498s

ampitere: This also seems to affect the same CS for Waking the Colossus which is event 307.

ampitere:
I have Waking the Colossus completed as well but will keep it separate for now. Still can't figure out what the issue is with CS 307 and 309.

scripts/quests/ahtUrhgan/Divine_Interference.lua
claywar: Was thinking about this.  Are there packet caps from the fight? I wonder if there's a text table change or event flag that needs to be set to support displaying the event.

MowFord: I stayed up late with him one night trying to get this right. We emulated every packet exactly as the capture and verified with Captain that they were the same. 

My best guess is some npc doesn't have the right properties and is causing the cutscenes to not proceed, but we tweaked every one in the capture to have exactly the same values and still would black screen. 

Also I highly suspect amp moved on, I haven't heard from him in months

WinterSolstice8: I looked at this the other day coincidentally and I'm more or less thinking the same thing, I'm leaning towards needing the 0x00E entity update packet code fixed before looking more into this

sruon:
Got CS to work after [aligning all the NPC flags and models](https://github.com/sruon/server/commit/59b8f1624100adbfb61e609a978f24d7c5eedd9d#diff-a67beaeee63b54ce2c90fa38e6e4ddd0d7fb29656be7f4e2de230f8ff01711fd) with the capture

Xaver-DaRed:
ampitere hasn't shown signs of life in quite a while. I assume this PR is abandoned, so I will close it.

@sruon Feel free to PR the battlefield if you got the CS working.

PR #5827, opened 2024-05-24
[cpp] Better skillup handling with sch arts active

### LINKED ISSUE #1981: 🐛 Incorrectly Listed Magic Skills During Skillup with Light/Dark Arts

## Branch affected by issue 

`base`

## Steps to reproduce

1. Load a SCH that has 0 skill in enfeebling
2. Activate Light Arts
3. Notice that the enfeebling magic skill was artificially boosted via light arts
4. Begin skilling up
5. On each skillup notice that the artificial skill level increases by 1 rather than staying static until the base skill surpasses the artificial skill level
6. Keep going and you can eventually get higher than 500 skill via this method
7. Zone or re-activate arts and notice that the skill resets (this bug cannot persist, so it's low-impact)

## Expected behavior

The listed skill would stay static until the base skill surpasses the artificial skill level

MowFord:
**_I affirm:_**

Closes #1981 .

at 0 skill enf with dark arts was at 246, skilled up and it stayed the same:
image

at 200 real skill, dark arts put me to 246. Skillup put me to 247:
image

image

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -3029,6 +3029,14 @@ namespace charutils
         }
     }
 
+    bool ArtsBonusActive(CCharEntity* PChar, SKILLTYPE SkillID)
+    {
+        return (SkillID >= SKILL_DIVINE_MAGIC && SkillID <= SKILL_ENFEEBLING_MAGIC &&
+                PChar->StatusEffectContainer->HasStatusEffect({ EFFECT_LIGHT_ARTS, EFFECT_ADDENDUM_WHITE })) ||
+               (SkillID >= SKILL_ENFEEBLING_MAGIC && SkillID <= SKILL_DARK_MAGIC &&
+                PChar->StatusEffectContainer->HasStatusEffect({ EFFECT_DARK_ARTS, EFFECT_ADDENDUM_BLACK }));
+    }
+
     /************************************************************************
      *                                                                       *
      *  Collect the work table of the character skills based on real.        *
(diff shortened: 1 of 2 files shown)


src/map/utils/charutils.cpp
```diff
@@ -3445,7 +3452,15 @@ namespace charutils
 
                 if ((CurSkill / 10) < (CurSkill + SkillAmount) / 10) // if gone up a level
                 {
-                    PChar->WorkingSkills.skill[SkillID] += 1;
+                    // Light/Dark Arts artificially boost certain skills, logic is too complex to duplicate, so just recalculate it all (similar to using the JA)
+                    if (ArtsBonusActive(PChar, SkillID))
```
zach2good, line 3456: I dont know _anything_ about SCH, but what this says to me is that we're going to call `BuildingCharSkillsTable` whenever  `TrySkillUP` is called - presumably on every action or melee swing etc. if `ArtsBonusActive` is on.

There's no DB calls in there, it's just a bunch of math, that seems OK, but logically: when `ArtsBonusActive` falls off, shouldn't we be rebuilding the table _one more time_ to put it back at its original non-inflated state?

MowFord: there's actually more work done when light/dark arts is gained/lost (in `onEffectGain`/`onEffectLose`)

and this is only going to do the extra math if the skillup in question is relevant to the currently-active arts buff. Perhaps that function could be renamed for clarity, but it returns true if SkiiID is one of the skills boosted by a current arts effect

PR #5792, opened 2024-05-20
[Quest]Implement Chips quest

hooksta4:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?
Implements Bastok quest Chips. 

## Steps to test these changes
!addmission 6 638
speak with Cid to get mission status to 1 or !setplayervar Mission[6][638]Status 1
Speak with Ghebi in Tenshodo - will return option. 
If player selects no, they must zone as per [capture](https://www.youtube.com/watch?v=CO5XbPSuY0g)
Return to Cid and trade Carmine Chip, Cyan Chip and Gray Chip to get CCB Polymer. 
NOTE if you drop it for testing, you will need to zone due to recycle bin.
Upon first completion, player will complete quest. Further trades of all 3 chips will reward another CCB Polymer.

PR #5830, opened 2024-05-24
[Quest]Chips Implementation

hooksta4:
**_I affirm:_**

Implements Bastok quest Chips. 

https://github.com/LandSandBoat/server/pull/5792 Cause I suck at rebasing. 

!addmission 6 638
speak with Cid to get mission status to 1 or !setplayervar Mission[6][638]Status 1
Speak with Ghebi in Tenshodo - will return option.
If player selects no, they must zone as per [capture](https://www.youtube.com/watch?v=CO5XbPSuY0g)
Return to Cid and trade Carmine Chip, Cyan Chip and Gray Chip to get CCB Polymer.
NOTE if you drop it for testing, you will need to zone due to recycle bin.
Upon first completion, player will complete quest. Further trades of all 3 chips will reward another CCB Polymer.

### DIFF
--- a/scripts/quests/bastok/Chips.lua
+++ b/scripts/quests/bastok/Chips.lua
@@ -1,0 +1,123 @@
+-----------------------------------
+-- Chips
+-----------------------------------
+-- Log ID: 1, Quest ID: 82
+-- NPC: Ghebi Damomohe !pos 17.05 -0.11 -5.53 245
+-- NPC: Cid !pos -12.53 -10.98 1.09 237
+-----------------------------------
+local metalworksID = zones[xi.zone.METALWORKS]
+-----------------------------------
+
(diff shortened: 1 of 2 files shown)


scripts/quests/bastok/Chips.lua
```diff
@@ -0,0 +1,123 @@
...
+                        player:confirmTrade()
+                    end
+                end,
+            },
+        },
+    },
+
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_COMPLETED -- You must zone if dropped as it will be in recycle bin and not allow you get get another.
```
zach2good: > -- You must zone if dropped as it will be in recycle bin and not allow you get get another.

Does that mean this section needs some `mustZone()`, or is that entirely handled by section 0?

hooksta4: No - just a note for if testing it. I can remove it as it is not needed.

PR #5831, opened 2024-05-24
[Module] Create module for quest workarounds

hooksta4:
**_I affirm:_**

Allows players to complete the quest Wish Upon a Star. Currently, Bastok does not have a "NONE" weather and therefore will never take the fallen star item. Adding sunshine (clear) remediates this.

Adds an or player:getWeather() == xi.weather.SUNSHINE
Run through the quest players will now be able to complete the quest.

### DIFF
--- a/modules/custom/lua/quest_workarounds.lua
+++ b/modules/custom/lua/quest_workarounds.lua
@@ -1,0 +1,40 @@
+-----------------------------------
+-- This module is a compilation of overrides to temporarily fix
+-- parts of quests that are not currently fully functional due to something
+-- larger (like a system) not being implemented/operable.
+-----------------------------------
+-- Include the following:
+-- - Explanation of what quest(s) needs temporary fix
+-- - Link to issue reported on LSB
+-- - Explanation (if known) of how to fix
+-- - Include a TODO so it can be searched.
+-----------------------------------
+require('modules/module_utils')
+-----------------------------------
+local m = Module:new('quest_workarounds')
+-----------------------------------
+
+m:addOverride('xi.server.onServerStart', function()
+    super()
+
+    -- Bastok zones don't currently produce NONE weather, so we are adding SUNSHINE as an accepted weather so the quest becomes completable.
+    -- See: https://github.com/LandSandBoat/server/pull/5831
+    xi.module.modifyInteractionEntry('scripts/quests/bastok/Wish_Upon_a_Star', function(quest)
+        quest.sections[2][xi.zone.BASTOK_MARKETS]['Enu'].onTrade = function(player, npc, trade)
+        local isNight = VanadielTOTD() == xi.time.NIGHT or VanadielTOTD() == xi.time.MIDNIGHT
+            if npcUtil.tradeHasExactly(trade, xi.item.FALLEN_STAR) then
+                if
+                    player:getWeather() == xi.weather.NONE or
+                    player:getWeather() == xi.weather.SUNSHINE and
+                    isNight
+                then
+                    return quest:progressEvent(334)
+                else
+                    return quest:event(337)
+                end
+            end
+        end
+    end)
+end)
+
+return m


zach2good (approved):

zach2good:
Would the right thing to do here be to give Bastok `NONE` weather?

zach2good (changes requested):
This is a workaround, not a fix

zach2good:
Looking at my weather tracker on retail:
image

hooksta4:
> Would the right thing to do here be to give Bastok `NONE` weather?

Yes, that would be the correct fix.

zach2good:
image

zach2good:
[db.zip](https://github.com/LandSandBoat/server/files/15447488/db.zip)

Here is my current sqlite db for the weather reporter, there should be enough information in there to get you started, plus the query in the SS above. 

zone.cpp explains in great detail how to set up weather information for zones:
```cpp

/*************************************************************************
 *                                                                        *
 *  Loads weather for the zone from zone_bweather SQL Table               *
 *                                                                        *
 *  Weather is a rotating pattern of 2160 vanadiel days for each zone.    *
 *  It's stored as a blob of 2160 16-bit values, each representing 1 day  *
 *  starting from day 0 and storing 3 5-bit weather values each.          *
 *                                                                        *
 *              0        00000       00000        00000                   *
 *              ^        ^^^^^       ^^^^^        ^^^^^                   *
 *          padding      normal      common       rare                    *
 *                                                                        *
 *************************************************************************/

void CZone::LoadZoneWeather()
{
    TracyZoneScoped;
    static const char* Query = "SELECT weather FROM zone_weather WHERE zone = %u";

```

zach2good:
Reason I won't accept this ([as I explained once in a stream](https://www.twitch.tv/videos/1904611177)) is that if this goes in, it'll never be properly fixed because it's so small an innocuous. All the information to do it properly is available.

hooksta4:
That's fair.

zach2good:
If you're not able to do the weather data breakdown, you can change this PR to provide a 'workarounds' Lua module that changes this condition using [xi.module.modifyInteractionEntry](https://github.com/LandSandBoat/server/blob/base/modules/module_utils.lua#L24), which will be off by default.

zach2good:
In that module, you can provide the steps needed to remedy and get rid of the workaround

hooksta4:
@zach2good  - tested this out good to go. Reverted the original file back to what it is currently.

modules/era/lua/wish_upon_a_star.lua
```diff
@@ -0,0 +1,34 @@
+-----------------------------------
+-- Module to make Wish Upon a Star completable.
+-- TODO include NONE in weather for bastok markets.
+-- https://github.com/LandSandBoat/server/pull/5831 comments for fix action
+-- Currently, Bastok does not have a "NONE" weather
+-- Therefore, the quest cannot be completed.
+-- Using this module, will allow the quest to be completed
+-- Using "Sunny" weather instead.
```
zach2good: Use the _original_ "fix" you proposed by using both NONE and SUNSHINE.

modules/era/lua/wish_upon_a_star.lua
```diff
@@ -0,0 +1,34 @@
+-----------------------------------
```
zach2good: Move this whole thing to a `quest_workarounds` module, where `xi.module.modifyInteractionEntry('scripts/quests/bastok/Wish_Upon_a_Star', function(quest)` is just a single part of it, with the comments above it.

```
m:addOverride('xi.server.onServerStart', function()
    super()

-- Module to make Wish Upon a Star completable.
-- TODO include NONE in weather for bastok markets.
-- https://github.com/LandSandBoat/server/pull/5831 comments for fix action
-- Currently, Bastok does not have a "NONE" weather
-- Therefore, the quest cannot be completed.
-- Using this module, will allow the quest to be completed
-- Using "Sunny" weather instead.
    xi.module.modifyInteractionEntry('scripts/quests/bastok/Wish_Upon_a_Star', function(quest)
        quest.sections[2][xi.zone.BASTOK_MARKETS]['Enu'].onTrade = function(player, npc, trade)
        local isNight = VanadielTOTD() == xi.time.NIGHT or VanadielTOTD() == xi.time.MIDNIGHT
            if npcUtil.tradeHasExactly(trade, xi.item.FALLEN_STAR) then
                if
                    player:getWeather() == xi.weather.SUNSHINE and
                    isNight
                then
                    return quest:progressEvent(334)
                else
                    return quest:event(337)
                end
            end
        end
    end)
end)
```

Something like this, but formatted correctly, etc.

We don't want an explosion of modules for no reason

modules/era/lua/wish_upon_a_star.lua
```diff
@@ -0,0 +1,34 @@
...
+require('modules/module_utils')
+-----------------------------------
+local m = Module:new('wish_upon_a_star')
+
+m:addOverride('xi.server.onServerStart', function()
+    super()
+
+    xi.module.modifyInteractionEntry('scripts/quests/bastok/Wish_Upon_a_Star', function(quest)
+        quest.sections[2][xi.zone.BASTOK_MARKETS]['Enu'].onTrade = function(player, npc, trade)
+        local isNight = VanadielTOTD() == xi.time.NIGHT or VanadielTOTD() == xi.time.MIDNIGHT
```
zach2good: Missing an indent level here

zach2good (commented):
Almost there

modules/era/lua/quest_workarounds.lua
```diff
@@ -0,0 +1,70 @@
...
+                then
+                    return quest:progressEvent(334)
+                else
+                    return quest:event(337)
+                end
+            end
+        end
+    end) -- end of Wish Upon a Star
+-- Begin Next Override
+end) -- End of all modules
```
zach2good: These comments aren't really needed, if people want additional clarity they can just pull out a function.

modules/era/lua/quest_workarounds.lua
```diff
@@ -0,0 +1,70 @@
...
+-- end) -- end of Wish Upon a Star
+-----------------------------------
+-- Module to make Wish Upon a Star completable.
+-- TODO include NONE in weather for bastok markets.
+-- https://github.com/LandSandBoat/server/pull/5831 comments for fix action
+-- Currently, Bastok does not have a "NONE" weather
+-- Therefore, the quest cannot be completed.
+-- Using this module, will allow the quest to be completed
+-- Using "Sunny" weather instead.
+-----------------------------------
```
zach2good: This whole section is overkill.

You only need to say at the lines below:
```lua
-- Bastok zones don't currently produce NONE weather, so we are adding SUNSHINE as an accepted weather so the quest becomes completable.
-- See: https://github.com/LandSandBoat/server/pull/5831
xi.module.modifyInteractionEntry('scripts/quests/bastok/Wish_Upon_a_Star', function(quest)
```
```

File: modules/era/lua/quest_workarounds.lua [hunk 6]
```diff
@@ -0,0 +1,70 @@
...
+-- This module is intended to be a complilation
```
zach2good: complilation -> compilation

modules/era/lua/quest_workarounds.lua
```diff
@@ -0,0 +1,70 @@
...
+-- of overrides that temporarily fix
+-- parts of quests that currently, for whatever
+-- reason, are not fully functional due to something
+-- larger (like a system) that has not been impliemented
```
zach2good: impliemented -> implemented

modules/era/lua/quest_workarounds.lua
hooksta4: Done.

zach2good:
```lua
-- Bastok zones don't currently produce NONE weather, so we are adding SUNSHINE as an accepted weather so the quest becomes completable.
-- See: https://github.com/LandSandBoat/server/pull/5831
xi.module.modifyInteractionEntry('scripts/quests/bastok/Wish_Upon_a_Star', function(quest)
```

^ As shown, I meant for you to put it above the code

zach2good:
Please don't resolve conversation threads on your own, you've checked this box:
```
 I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
```

It's there to make reviewers lives easier.

zach2good (approved):

zach2good:
`modules/era/lua/quest_workarounds.lua`

This doesn't have anything to do with a specific era, I'm going to pull this down into a different branch and `git mv` it, and submit it in a new PR (attribution preserved, obviously)

PR #5832, opened 2024-05-24
[core] Fix hp and mp visible gear latents

TracentEden2:
**_I affirm:_**

This PR adds the functionality for two rare types of latent gear. Specifically MP_UNDER_VISIBLE_GEAR and HP_OVER_VISIBLE_GEAR which are used by Uggalepih Pendant and Parade Gorget respectively. Retail testing can be see for example [here](https://github.com/AirSkyBoat/AirSkyBoat/pull/2099) and [JP Wiki](https://wiki.ffo.jp/html/2495.html).

Still in draft as I figure out how this will interact with another PR I am making to fix other latent and HP issues.

Calculate your HP or MP given only the +HP or +MP in the defined slots (SLOT_MAIN, SLOT_SUB, SLOT_RANGED, SLOT_AMMO, SLOT_BODY, SLOT_HANDS, SLOT_LEGS, SLOT_FEET) and then, with the pendant, for example, check your MATT at different MPs near 51%

### DIFF
--- a/src/map/latent_effect_container.cpp
+++ b/src/map/latent_effect_container.cpp
@@ -1011,57 +1011,36 @@ bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect)
             expression = m_POwner->health.hp < latentEffect.GetConditionsValue() && m_POwner->animation == ANIMATION_ATTACK;
             break;
         case LATENT::MP_UNDER_VISIBLE_GEAR:
-            // TODO: figure out if this is actually right
-            // CItemEquipment* head = (CItemEquipment*)(m_POwner->getStorage(LOC_INVENTORY)->GetItem(m_POwner->equip[SLOT_HEAD]));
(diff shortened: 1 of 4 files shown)


src/map/latent_effect_container.cpp
```diff
@@ -1011,57 +1011,36 @@ bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect)
...
+            std::vector<CItemEquipment*> visibleEquip = m_POwner->getVisibleEquip();
+
+            int32 visibleMP = m_POwner->health.maxmp;
+
+            for (auto equip: visibleEquip)
+            {
+                visibleMP += equip->getModifier(Mod::MP);
+            }
+            
+            // there is evidence that 0/0 (e.g., 0/20 with uggalepih pendant) does not trigger this latent
```
zach2good, line 1025: Leave a link to the evidence

zach2good (approved):
Seems reasonable to me

PR #5833, opened 2024-05-26
[core] add lua bindings for drawIn and getAbilityDistance

0x05010705:
**_I affirm:_**

Adds two new LUA Bindings for the following functions:

1 - `getAbilityDistance(####)`
Currently, there is no way to query the distance that is used on a mob skill from the `mob_skills` database.  This allows for mob scripts to check the distance on abilities before performing calls such as `useMobAbility(####)` to provide some assurance that the skill being used will actually succeed.

2 - `drawIn([playerObject])`
While draw-in mob mods and mechanics do exist, there is no way to forcibly draw in a player.  This function allows a mob to draw-in outside of the specified fence/distance in the mob mods and at any point.  With no argument, the mob draws in the current target, or a player object can be used to draw in specific targets.  This was created to allow for use in conjunction with getAbilityDistance.

The combination of the two functions provides a reasonable level of assurance that an ability called with useMobAbility will be able to be performed successfully.  Its a more modular approach that prevents modifying the existing useMobAbility beyond what is done in #5815 

Edit any mob script to include the function calls on any entity

Example - Bumblebee.lua in /scripts/zones/East_Sarutabaruta/mobs/Bumblebee.lua
```lua
entity.onMobFight = function(mob)
  if (mob:checkDistance(mob:getTarget()) > mob:getAbilityDistance(334)) then
    mob:drawIn(mob:getTarget())
  end
  mob:useMobAbility(334)
end
```

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -16632,6 +16632,25 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
 }
 
 /************************************************************************
+ *  Function: getAbilityDistance()
+ *  Purpose : Returns the distance for a specified ability from mob_skills
+ *  Example : mob:getAbilityDistance(740)
+ *  Notes   : Uses Ability ID Only
+ ************************************************************************/
+float CLuaBaseEntity::getAbilityDistance(uint16 skillID)
+{
+    auto* PMobSkill = battleutils::GetMobSkill(skillID);
+
+    if (PMobSkill == nullptr)
+    {
+        ShowWarning("Invalid SkillID (%d) returned null, returning 0 for distance.", skillID);
+        return 0.0f;
+    }
+
+    return PMobSkill->getDistance();
+}
+
+/************************************************************************
  *  Function: hasTPMoves()
  *  Purpose : Returns true if a Mob has TP moves in its skill list
  *  Example : if mob:hasTPMoves() then
@@ -16658,6 +16677,58 @@ bool CLuaBaseEntity::hasTPMoves()
     const std::vector<uint16>& MobSkills = battleutils::GetMobSkillList(familyID);
 
     return !MobSkills.empty();
+}
+
+/************************************************************************
+ *  Function: drawIn()
+ *  Purpose : Draws in the target, or current target if not specified
+ *  Example : mob:drawIn()     mob:drawIn(player)
+ *  Notes   : Draws in a player even if within the draw-in leash
+ ************************************************************************/
+void CLuaBaseEntity::drawIn(sol::variadic_args va)
+{
+    if (m_PBaseEntity->objtype != TYPE_MOB)
+    {
+        ShowError("Attempting to draw-in from non-mob entity: %s", m_PBaseEntity->getName());
+        return;
+    }
+
+    auto mobObj = dynamic_cast<CMobEntity*>(m_PBaseEntity);
+
+    if (va.size() == 0)
+    {
+        auto defaultTarget = mobObj->GetBattleTarget();
+
+        if (defaultTarget == nullptr)
+        {
+            return;
+        }
+        battleutils::DrawIn(defaultTarget, mobObj, mobObj->GetMeleeRange() - 0.2f);
+        return;
+    }
+
+    CLuaBaseEntity* PLuaBaseEntity = va.get<CLuaBaseEntity*>(0);
+
+    if (!PLuaBaseEntity)
+    {
+        ShowError("Attempt to draw-in non-valid target.");
+        return;
+    }
+
+    CBaseEntity*   PBaseEntity = PLuaBaseEntity->m_PBaseEntity;
+    CBattleEntity* PTarget     = nullptr;
+
+    if (PBaseEntity)
+    {
+        PTarget = dynamic_cast<CBattleEntity*>(PBaseEntity);
+    }
+
+    if (PTarget)
+    {
+        battleutils::DrawIn(PTarget, mobObj, mobObj->GetMeleeRange() - 0.2f);
+    }
+
+    return;
 }
 
 /************************************************************************
@@ -18240,7 +18311,9 @@ void CLuaBaseEntity::Register()
     SOL_REGISTER("castSpell", CLuaBaseEntity::castSpell);
     SOL_REGISTER("useJobAbility", CLuaBaseEntity::useJobAbility);
     SOL_REGISTER("useMobAbility", CLuaBaseEntity::useMobAbility);
+    SOL_REGISTER("getAbilityDistance", CLuaBaseEntity::getAbilityDistance);
     SOL_REGISTER("hasTPMoves", CLuaBaseEntity::hasTPMoves);
+    SOL_REGISTER("drawIn", CLuaBaseEntity::drawIn);
 
     SOL_REGISTER("weaknessTrigger", CLuaBaseEntity::weaknessTrigger);
     SOL_REGISTER("restoreFromChest", CLuaBaseEntity::restoreFromChest);
--- a/src/map/lua/lua_baseentity.h
+++ b/src/map/lua/lua_baseentity.h
@@ -854,7 +854,9 @@ public:
     void castSpell(sol::object const& spell, sol::object const& entity); // forces a mob to cast a spell (parameter = spell ID, otherwise picks a spell from its list)
     void useJobAbility(uint16 skillID, sol::object const& pet);          // forces a job ability use (players/pets only)
     void useMobAbility(sol::variadic_args va);                           // forces a mob to use a mobability (parameter = skill ID)
+    auto getAbilityDistance(uint16 skillID) -> float;                    // Returns the specified distance for mob skill
     bool hasTPMoves();
+    void drawIn(sol::variadic_args va); // Forces a mob to draw-in the specified target, or its current target with no args
 
     void weaknessTrigger(uint8 level);
     void restoreFromChest(CLuaBaseEntity* PLuaBaseEntity, uint32 restoreType);


src/map/lua/lua_baseentity.cpp
```diff
@@ -16631,6 +16631,19 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
...
+ *  Function: getAbilityDistance()
+ *  Purpose : Returns the distance for a specified ability from mob_skills
+ *  Example : mob:getAbilityDistance(740)
+ *  Notes   : Uses Ability ID Only
+ ************************************************************************/
+float CLuaBaseEntity::getAbilityDistance(uint16 skillID)
+{
+    auto* PMobSkill{ battleutils::GetMobSkill(skillID) };
+
+    return PMobSkill->getDistance();
```
WinterSolstice8, line 16650: Need to check null `PMobSkill` here

src/map/lua/lua_baseentity.cpp
```diff
@@ -16660,6 +16673,36 @@ bool CLuaBaseEntity::hasTPMoves()
...
+ ************************************************************************/
+void CLuaBaseEntity::drawIn(sol::variadic_args va)
+{
+    auto mobObj = dynamic_cast<CMobEntity*>(m_PBaseEntity);
+
+    if (va.size() == 0)
+    {
+        auto defaultTarget = mobObj->GetBattleTarget();
+
+        battleutils::DrawIn(defaultTarget, mobObj, mobObj->GetMeleeRange() - 0.2f);
```
WinterSolstice8, line 16706: What happens if `defaultTarget` is null?

src/map/lua/lua_baseentity.cpp
0x05010705: Good catch.  Added a check for null and just returns silently
```cpp
        auto defaultTarget = mobObj->GetBattleTarget();

        if (defaultTarget == nullptr)
        {
            return;
        }
        battleutils::DrawIn(defaultTarget, mobObj, mobObj->GetMeleeRange() - 0.2f);
        return;
```

src/map/lua/lua_baseentity.cpp
0x05010705: Good catch.  Added a check for null and just returns silently
```cpp
        auto defaultTarget = mobObj->GetBattleTarget();

        if (defaultTarget == nullptr)
        {
            return;
        }
        battleutils::DrawIn(defaultTarget, mobObj, mobObj->GetMeleeRange() - 0.2f);
        return;
```

WinterSolstice8 (approved):

src/map/lua/lua_baseentity.cpp
0x05010705: Added a check for null here that shows a warning about invalid skillID being specified and return `0.0f`
```cpp
    if (PMobSkill == nullptr)
    {
        ShowWarning("Invalid SkillID (%d) returned null, returning 0 for distance.", skillID);
        return 0.0f;
    }
```

src/map/lua/lua_baseentity.cpp
```diff
@@ -16660,6 +16679,40 @@ bool CLuaBaseEntity::hasTPMoves()
...
 
+/************************************************************************
+ *  Function: drawIn()
+ *  Purpose : Draws in the target, or current target if not specified
+ *  Example : mob:drawIn()     mob:drawIn(player)
+ *  Notes   : Draws in a player even if within the draw-in leash
+ ************************************************************************/
+void CLuaBaseEntity::drawIn(sol::variadic_args va)
+{
+    auto mobObj = dynamic_cast<CMobEntity*>(m_PBaseEntity);
```
zach2good, line 16696: This could be called accidentally on a player or an NPC, then mobObj would be `nullptr`, and then there's no check or logging to say what's going on. Then if va.size() == 0 it'll get dereferenced and there'll be a crash

src/map/lua/lua_baseentity.h
```diff
@@ -851,10 +851,12 @@ class CLuaBaseEntity
...
     bool actionQueueEmpty();
 
-    void castSpell(sol::object const& spell, sol::object const& entity); // forces a mob to cast a spell (parameter = spell ID, otherwise picks a spell from its list)
-    void useJobAbility(uint16 skillID, sol::object const& pet);          // forces a job ability use (players/pets only)
-    void useMobAbility(sol::variadic_args va);                           // forces a mob to use a mobability (parameter = skill ID)
-    bool hasTPMoves();
+    void  castSpell(sol::object const& spell, sol::object const& entity); // forces a mob to cast a spell (parameter = spell ID, otherwise picks a spell from its list)
+    void  useJobAbility(uint16 skillID, sol::object const& pet);          // forces a job ability use (players/pets only)
+    void  useMobAbility(sol::variadic_args va);                           // forces a mob to use a mobability (parameter = skill ID)
+    float getAbilityDistance(uint16 skillID);                             // Returns the specified distance for mob skill
```
zach2good: auto methodName() -> float;, so you don't have to realign everything

src/map/lua/lua_baseentity.cpp
```diff
@@ -16631,6 +16631,25 @@ void CLuaBaseEntity::useMobAbility(sol::variadic_args va)
...
 
+/************************************************************************
+ *  Function: getAbilityDistance()
+ *  Purpose : Returns the distance for a specified ability from mob_skills
+ *  Example : mob:getAbilityDistance(740)
+ *  Notes   : Uses Ability ID Only
+ ************************************************************************/
+float CLuaBaseEntity::getAbilityDistance(uint16 skillID)
+{
+    auto* PMobSkill{ battleutils::GetMobSkill(skillID) };
```
zach2good: While this is correct, I think we tend to use assignment for this sort of initialisation around the codebase, same inside of drawIn()

src/map/lua/lua_baseentity.cpp
```diff
@@ -16660,6 +16679,40 @@ bool CLuaBaseEntity::hasTPMoves()
...
+        return;
+    }
+
+    CBattleEntity*  PTarget{ nullptr };
+    CLuaBaseEntity* PLuaBaseEntity = va.get<CLuaBaseEntity*>(0);
+    PTarget                        = PLuaBaseEntity ? (CBattleEntity*)PLuaBaseEntity->m_PBaseEntity : nullptr;
+
+    if (PTarget)
+    {
+        battleutils::DrawIn(PTarget, mobObj, mobObj->GetMeleeRange() - 0.2f);
```
zach2good, line 16728: Where does this 0.2f come from? What happens if GetMeleeRange returns zero? (I haven't looked to see if that's a possibility)

src/map/lua/lua_baseentity.cpp
```diff
@@ -16660,6 +16679,40 @@ bool CLuaBaseEntity::hasTPMoves()
...
+        if (defaultTarget == nullptr)
+        {
+            return;
+        }
+        battleutils::DrawIn(defaultTarget, mobObj, mobObj->GetMeleeRange() - 0.2f);
+        return;
+    }
+
+    CBattleEntity*  PTarget{ nullptr };
+    CLuaBaseEntity* PLuaBaseEntity = va.get<CLuaBaseEntity*>(0);
```
zach2good, line 16710: Presumably this will return a nullptr if you pass in a number to the function, right?

zach2good (commented):
Looking like it's heading in the right direction 👍

src/map/lua/lua_baseentity.h
0x05010705: Updated to
```cpp
auto getAbilityDistance(uint16 skillID);
```

zach2good:
We still care about seeing the return type in the header, for quick
reference

src/map/lua/lua_baseentity.cpp
0x05010705: This check at the start of the function should correct the wrong entity type calling this, and/or the de-referencing of the nullptr

```cpp
    if (m_PBaseEntity->objtype != TYPE_MOB)
    {
        ShowError("Attempting to draw-in from non-mob entity: %s", m_PBaseEntity->getName());
        return;
    }
```

src/map/lua/lua_baseentity.cpp
0x05010705: The `GetMeleeRange() - 0.2f` is from looking at how Draw-In was done from the `mob_controller.cpp` file
Line 694 - 701 in `mob_controller.cpp`:
```cpp
            if (PMob->getMobMod(MOBMOD_DRAW_IN) > 0)
            {
                if (currentDistance >= PMob->GetMeleeRange() * 2 && battleutils::DrawIn(PTarget, PMob, PMob->GetMeleeRange() - 0.2f))
                {
                    FaceTarget();
                    return;
                }
            }
```

src/map/lua/lua_baseentity.cpp
0x05010705: The following line should fix this no?
```cpp
PTarget                        = PLuaBaseEntity ? (CBattleEntity*)PLuaBaseEntity->m_PBaseEntity : nullptr;
```
If `PLuaBaseEntity != nullptr`, then `PTarget` is assigned the BaseEntity
Otherwise if `PLuaBaseEntity` was `nullptr`, then `PTarget` is assigned `nullptr`.

Then immediately after that:
```cpp
if (PTarget)   // If PTarget is NULL, we skip this segment
{ [ ..snip..] }
return;        // and end up returning
```

Or are you saying that `va.get<CLuaBaseEntity*>(0)` will crash on the invalid argument? (IE: int instead of Player/Target object)

WinterSolstice8: avoid usage of C style casts, 

`(CBattleEntity*)PLuaBaseEntity->m_PBaseEntity` -> `dynamic_cast<CBattleEntity*>(PLuaBaseEntity->m_PBaseEntity)`

src/map/lua/lua_baseentity.h
0x05010705: Saw your reply (not sure why its disconnected from the feedback on the line)

Re-Updated to:
```cpp
    auto getAbilityDistance(uint16 skillID) -> float;                    // Returns the specified distance for mob skill
```

src/map/lua/lua_baseentity.cpp
0x05010705: Updated to use assignment for the initialization:
```cpp
auto* PMobSkill = battleutils::GetMobSkill(skillID);
```

src/map/lua/lua_baseentity.cpp
0x05010705: CLuaBaseEntity is not polymorphic 

Implementing that suggestion generates the following error:
`the operand of a runtime dynamic_cast must have a polymorphic class type`

WinterSolstice8: try `dynamic_cast<CBattleEntity*>(PLuaBaseEntity->GetBaseEntity())`

0x05010705: Re-Addressed both of these with making it a bit more explicit in the source
```cpp
    CLuaBaseEntity* PLuaBaseEntity = va.get<CLuaBaseEntity*>(0);

    if (!PLuaBaseEntity)
    {
        ShowError("Attempt to draw-in non-valid target.");
        return;
    }

    CBaseEntity* PBaseEntity = PLuaBaseEntity->m_PBaseEntity;
    CBattleEntity* PTarget = nullptr;
    
    if (PBaseEntity)
    {
        PTarget = dynamic_cast<CBattleEntity*>(PBaseEntity);
    }

    if (PTarget)
    {
        battleutils::DrawIn(PTarget, mobObj, mobObj->GetMeleeRange() - 0.2f);
    }

    return;
}
 ```
 
 RE: Zach's initial question -- If a Number is passed into this function, a LUA error in generated in the xi_map.exe process.  No crashes occur in the actual map process though.
image

zach2good (approved):

PR #5840, opened 2024-05-28
[LUA] Fix Temenos North avatars should not start spawned, error in Astral Flow

ampitere:
**_I affirm:_**

Currently in Temenos North there's SMN avatars just hanging out doing nothing. After reviewing a capture video it appears they should only spawn when a SMN uses Astral Flow. To resolve this we will set them to not be spawned initially and add the avatar mixin similar to other Temenos battlefields. Additionally there is an error when Astral Flow is used currently due to `updateEnmity` not being passed anything so we will update the logic to check for a master and master's target to use in this call to resolve the error currently occuring.

https://www.youtube.com/watch?v=k7yx2lykDjM

Here is a capture where you can see the avatars only appear after a SMN uses Astral Flow. Elementals do appear already spawned in.

1. !pos 580.000 -2.375 104.000 37
2. !addkeyitem white_card
3. !addkeyitem cosmo_cleanse
4. Enter and go down a couple floors until you're either at Kindreds or Cryptonberrys
5. Whack on a SMN until they use Astral Flow (!togglegm to test that they still target you even with GM on)

### DIFF
--- a/scripts/mixins/families/avatar.lua
+++ b/scripts/mixins/families/avatar.lua
@@ -27,7 +27,13 @@ g_mixins.families.avatar = function(avatarMob)
 
         -- When GM is enabled, mobs will not automatically engage.  Update Enmity one more
         -- time to ensure that the listener will actually be triggered.
-        mob:updateEnmity()
+        local master = mob:getMaster()
+        if master ~= nil then
+            local target = master:getTarget()
+            if target ~= nil then
+                mob:updateEnmity(target)
+            end
+        end
 
         -- If something goes wrong, the avatar will clean itself up in 5s
         mob:timer(5000, function(mobArg)
--- a/scripts/battlefields/Temenos/temenos_northern_tower.lua
+++ b/scripts/battlefields/Temenos/temenos_northern_tower.lua
@@ -224,11 +224,13 @@ content.groups =
     },
 
     {
-        mobs =
-        {
-            'Kindreds_Elemental',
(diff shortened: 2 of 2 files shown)


scripts/mixins/families/avatar.lua
```diff
@@ -27,7 +27,13 @@ g_mixins.families.avatar = function(avatarMob)
 
         -- When GM is enabled, mobs will not automatically engage.  Update Enmity one more
         -- time to ensure that the listener will actually be triggered.
-        mob:updateEnmity()
+        local master = mob:getMaster()
```
claywar, line 30: Just wanted to make sure this was tested here.  The avatar may not be the mob's "pet", especially in cases where an elemental is also valid.

ampitere: > Just wanted to make sure this was tested here. The avatar may not be the mob's "pet", especially in cases where an elemental is also valid.

Can you think of any spots in game I can test this? I tested in Temenos North and Jungle Boogeyman since that seems to be where this originated from. Both were able to hit me with GM toggled on.

claywar: If Jungle worked and still had the avatar AF with elemental up, its a 👍 in my book

ampitere: image

Yeah should be good.

Xaver-DaRed (approved):

PR #5847, opened 2024-05-29
[CPP] Fix map crash when counting trusts

### LINKED ISSUE #5844: 🐛 Map.exe crash when party member zones.
**_I affirm:_**

Windows 10

## Branch affected by issue 

`base` (Unmodified)

## Steps to reproduce

1. Make 2 characters on 2 accounts (Or have a friend help test)
2. Have 2 characters make a party. Have one of them zone in or out of the zone the other is in.
3. See that the server crashes.

image
image

[xi_map.exe_29-5_6-0-35.dmp](https://github.com/LandSandBoat/server/files/15484689/xi_map.exe_29-5_6-0-35.dmp)

## Expected behavior

Server should not crash.

ampitere:
**_I affirm:_**

Fixes a map crash due to party leader being null when they are initially zoning. Resolves https://github.com/LandSandBoat/server/issues/5844

1. Make a group with 2 or more
2. Be in the same zone together and have the party leader zone out and in

### DIFF
--- a/src/map/party.cpp
+++ b/src/map/party.cpp
@@ -335,7 +335,12 @@ void CParty::RemoveMember(CBattleEntity* PEntity)
                     }
                 }
 
-                auto trustCount = static_cast<CCharEntity*>(m_PLeader)->PTrusts.size();
+                size_t trustCount = 0;
+                if (m_PLeader != nullptr)
+                {
+                    trustCount = static_cast<CCharEntity*>(m_PLeader)->PTrusts.size();
+                }
+
                 PChar->PLatentEffectContainer->CheckLatentsPartyMembers(members.size(), trustCount);
 
                 PChar->pushPacket(new CPartyDefinePacket(nullptr));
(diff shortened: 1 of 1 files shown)


src/map/party.cpp
```diff
@@ -335,7 +335,17 @@ void CParty::RemoveMember(CBattleEntity* PEntity)
                     }
                 }
 
-                auto trustCount = static_cast<CCharEntity*>(m_PLeader)->PTrusts.size();
+                size_t trustCount = 0;
+                if (m_PLeader != nullptr)
+                {
+                    std::vector<CTrustEntity*> PTrusts = static_cast<CCharEntity*>(m_PLeader)->PTrusts;
```
zach2good: No need to make a copy of the trusts vector here, same for below.

trustCount is init'd as zero:
```
size_t trustCount = 0;
```

The size of a vector is zero when it's empty.

So you can simplify this block (and below) with something like:

```cpp
trustCount = static_cast<CCharEntity*>(m_PLeader)->PTrusts.size();
```

ampitere: Simplified both this line and 867.

zach2good (approved):

PR #5849, opened 2024-05-29
[Quest]WHM AF1 Messenger From Beyond

hooksta4:
**_I affirm:_**

Converts WHM AF1 to IF. 
- Removes logic from NPC files.
- Adds a dialog when spawning NM.
- [Capture](https://www.youtube.com/watch?v=PTgMbnjW1yg)

Be on whm 40+ Speak with Narchera
Pop and Kill NM 
Return Tavnazia Pass to Narchera

### DIFF
--- a/scripts/quests/sandoria/WHM_AF1_Messenger_From_Beyond.lua
+++ b/scripts/quests/sandoria/WHM_AF1_Messenger_From_Beyond.lua
@@ -1,0 +1,86 @@
+-----------------------------------
+-- Messenger from Beyond
+-- !addquest: 0 87
+-- Narcheral: !pos 128 -11 128 231
+-- qm2      : !pos -714 -9 68 103
+-----------------------------------
+local valkurmID = zones[xi.zone.VALKURM_DUNES]
+-----------------------------------
+
+local quest = Quest:new(xi.questLog.SANDORIA, xi.quest.id.sandoria.MESSENGER_FROM_BEYOND)
+
+quest.reward =
+{
+    item     = xi.item.BLESSED_HAMMER,
+    fame     = 20,
+    fameArea = xi.fameArea.SANDORIA,
+}
+
+quest.sections =
+{
+    -- Section: Quest available.
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_AVAILABLE and
+                player:getMainJob() == xi.job.WHM and
+                player:getMainLvl() >= xi.settings.main.AF1_QUEST_LEVEL
+        end,
+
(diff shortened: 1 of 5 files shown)


scripts/quests/sandoria/WHM_AF1_Messenger_From_Beyond.lua
```diff
@@ -0,0 +1,92 @@
+-----------------------------------
+-- Messenger from Beyond
+-- !addquest: 0 87
+-- Narcheral: !pos 128 -11 128 231
+-- qm2      : !pos -714 -9 68 103
+-----------------------------------
+local valkID = zones[xi.zone.VALKURM_DUNES]
```
claywar: Lets not abbreviate more than necessary, `valkurmID`

scripts/quests/sandoria/WHM_AF1_Messenger_From_Beyond.lua
```diff
@@ -0,0 +1,92 @@
...
+            ['qm2'] =
+            {
+                onTrigger = function(player, npc)
+                    if
+                        not player:hasItem(xi.item.TAVNAZIA_PASS) and
+                        npcUtil.popFromQM(player, npc, valkID.mob.MARCHELUTE, { claim = true, hide = 0 })
+                    then
+                        return quest:messageSpecial(valkID.text.FOUL_PRESENSE)
+                    else
+                        return quest:messageSpecial(valkID.text.NOTHING_OUT_OF_ORDINARY)
```
claywar: This should be in DefaultActions

scripts/quests/sandoria/WHM_AF1_Messenger_From_Beyond.lua
```diff
@@ -0,0 +1,92 @@
...
+            ['Narcheral'] =
+            {
+                onTrade = function(player, npc, trade)
+                    if npcUtil.tradeHasExactly(trade, xi.item.TAVNAZIA_PASS) then
+                        return quest:progressEvent(690) -- Finish quest.
+                    end
+                end,
+
+                onTrigger = function(player, npc)
+                    return quest:event(688)
```
claywar: This is not relevant to this quest, and is either a default action or should remain in his script

claywar (changes requested):
Overall fairly reasonable, but removing the default and attaching it to a quest is not acceptable.  This must be changed.

scripts/quests/sandoria/WHM_AF1_Messenger_From_Beyond.lua
hooksta4: made changes to all.. Thanks!

PR #5856, opened 2024-05-30
Revert "[lua] Recalculate stats after player calcs gear set mods"

zach2good:
Reverts LandSandBoat/server#5744

Watch CI and labels get angry, I'm doing this from my phone

**_I affirm:_**


zach2good:
Pending a solution that doesn't send out a load of redundant packets on gear swap

claywar (approved):

PR #5857, opened 2024-05-30
Fix Lua onItemCheck

The-Aerec:
onItemCheck was incorrectly returning the user in two parameters instead of the item in the second param slot

**_I affirm:_**

This PR fixes onItemCheck to actually allow the item being used during the check to be referenced instead of the caster. The function is defined as: 

`OnItemCheck(CBaseEntity* PTarget, CItem* PItem, ITEMCHECK param, CBaseEntity* PCaster)`

and was returning the following:

`Target, Caster, param`

it now returns:

`result = onItemCheck(CLuaBaseEntity(PTarget), CLuaItem(PItem), static_cast<uint32>(param), caster)`

Reference item anywhere in onItemCheck

print(item) to see that you are indeed referencing an CLuaItem instead of CLuaBaseEntity


zach2good:
So caster is now unused?

The-Aerec:
> So caster is now unused?

Yes, I can extend or modify the function to include the caster again. Or just change the place of the item return. Up to yall. Item was actually never being returned so you could never check anything about the item in onItemCheck

The-Aerec:
I have added caster to its proper place: 
`OnItemCheck(CBaseEntity* PTarget, CItem* PItem, ITEMCHECK param, CBaseEntity* PCaster)`

`auto result = onItemCheck(CLuaBaseEntity(PTarget), CLuaItem(PItem), static_cast<uint32>(param), caster)`

zach2good:
There's code that expects the original layout (note the comma):
image

I agree with you about matching the C++ signature: `std::tuple<int32, int32, int32> OnItemCheck(CBaseEntity* PTarget, CItem* PItem, ITEMCHECK param, CBaseEntity* PCaster)`

So you'll need to change the existing onItemCheck usages, plus replace the original simplest version with the new extended version:

```
itemObject.onItemCheck = function(target, player)
and
itemObject.onItemCheck = function(target)

to

itemObject.onItemCheck = function(target, item, param, caster)
```

and make sure everything matches up in the `itemObject.onItemCheck = function(target, player)` case etc.

The-Aerec:
Zachs request has been compeleted:
Please note I can't test every item in the game!

zach2good (approved):
I checked ccb_polymer_pump, all looks fine :+1:

In the future, typically, you'd make the smaller logical changes or more sensitive changes in their own commit, and then the huge 2000+ file find/replace changes in another commit so things are more easily reviewable.

PR #5858, opened 2024-05-30
[core] Clear zone wide treasure pool when last player leaves zone

TracentEden2:
**_I affirm:_**

This PR clears the zone wide treasure pool when the last player leaves the zone. This prevents situations where, for example, a group exits dyna (or is kicked) with treasure still in pool and the next group enters and can lot that treasure.

Set a zone to have zone wide treasure pool and then get items in pool and do !goto playerName and see that the treasure is gone after zoning back in

### DIFF
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -1174,6 +1174,14 @@ void CZone::CharZoneOut(CCharEntity* PChar)
         PChar->PTreasurePool->DelMember(PChar);
     }
 
+    // If zone-wide treasure pool but no players in zone then destroy current pool and create new pool
+    // this prevents loot from staying in zone pool after the last player leaves the zone
+    if (m_TreasurePool && m_TreasurePool->GetPoolType() == TREASUREPOOL_ZONE && m_zoneEntities->CharListEmpty())
+    {
+        destroy(m_TreasurePool);
+        m_TreasurePool = new CTreasurePool(TREASUREPOOL_ZONE);
+    }
+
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

src/map/zone.cpp
```diff
@@ -1174,6 +1174,15 @@ void CZone::CharZoneOut(CCharEntity* PChar)
         PChar->PTreasurePool->DelMember(PChar);
     }
 
+    // If zone-wide treasure pool but no players in zone then destroy current pool and create new pool
+    // this prevents loot from staying in zone pool after the last player leaves the zone
+    if (m_TreasurePool && m_TreasurePool->GetPoolType() == TREASUREPOOL_ZONE && m_zoneEntities->CharListEmpty())
+    {
+        destroy(m_TreasurePool);
+        delete m_TreasurePool;
```
zach2good: This is now a double delete, which is UB

```cpp
template <typename T>
inline void destroy(T*& ptr)
{
    delete ptr; // cpp.sh allow
    ptr = nullptr;
}
```

TracentEden: > This is now a double delete, which is UB
> 
> ```c++
> template <typename T>
> inline void destroy(T*& ptr)
> {
>     delete ptr; // cpp.sh allow
>     ptr = nullptr;
> }
> ```

Ah, true, thanks, I have updated the PR.

zach2good (approved):

PR #5859, opened 2024-05-31
[SQL] Fix Bastok Markets NPC data

ampitere:
**_I affirm:_**

Fixes incorrect data for an NPC in Bastok Markets.
- Removes duplicated speed value, which shifted data.
- Adds missing entity flags

image

1. !pos -261.242 -12.36 80.643 235
2. NPC should be hidden


sql/npc_list.sql
```diff
@@ -27716,7 +27716,7 @@ INSERT INTO `npc_list` VALUES (17736072,'Tales_Beginning','Tales\' Beginning',11
 -- Bastok Markets (Zone 235)
 -- ------------------------------------------------------------
 
-INSERT INTO `npc_list` VALUES (17739777,'NPC[1]','NPC',171,-261.242,-12.136,80.643,0,50,50,50,0,0,0,6,0x0000620000000000000000000000000000000000,32,NULL,0);
+INSERT INTO `npc_list` VALUES (17739777,'NPC[1]','NPC',171,-261.242,-12.136,80.643,0,50,50,50,0,0,3,6,0x0000620000000000000000000000000000000000,32,NULL,0);
```
Xaver-DaRed, line 27719: This fix is incorrect.
Just because the npc should not be visible, doesnt automatically mean the NPC status should be "Invisible"

For this kind of stuff, a capture is needed. One can't just make up this values.

Once you look at the sql entry, the issue is obvious. It has an extra speed value.

Xaver-DaRed (changes requested):
```sql
INSERT INTO `npc_list` VALUES (17739777,'NPC[1]','NPC',171,-261.242,-12.136,80.643,0,50,50,0,0,0,6,27,0x0000620000000000000000000000000000000000,32,NULL,0);
```

This is th correct data that should be put in there.

sql/npc_list.sql
ampitere: Fixed.

Xaver-DaRed (approved):

PR #5860, opened 2024-05-31
[core, lua] Add zone uptime command

TracentEden2:
**_I affirm:_**

This PR adds a command to display the uptime of the current zone to a GM (several private servers also expose such a command to normal players). This command is something specifically requested by Horizon GMs and very useful for investigating, for example, player reports of zone crashes and validating the zone actually crashed (rather than just player DC) when refunding pop items (like BCNM orbs).

Boot up a zone and use !uptime

### DIFF
--- a/scripts/commands/uptime.lua
+++ b/scripts/commands/uptime.lua
@@ -1,0 +1,50 @@
+-----------------------------------
+-- func: uptime
+-- desc: prints zone uptime
+-----------------------------------
+local commandObj = {}
+
+commandObj.cmdprops =
+{
+    permission = 1,
+    parameters = ''
+}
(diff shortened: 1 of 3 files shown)


WinterSolstice8 (approved):
why not?

zach2good (approved):
Sure, but can you use `fmt` or `string.format` for building strings in Lua instead of lots of `.. `

TracentEden2:
> Sure, but can you use `fmt` or `string.format` for building strings in Lua instead of lots of `.. `

Yep, I updated the PR, thanks for feedback!

zach2good (approved):

PR #5863, opened 2024-05-31
[Quest]An Undying Pledge Conv

hooksta4:
**_I affirm:_**

- Converts Quest An Undying Pledge to IF.
- Adds Light Buckler to item.lua
- Removes npc lua logic.
- Adds defaultAction for qm5 and Stray Cloud
- Adds dialog for after quest completion
- Removes trigger area from zone.lua and into IF file
- Removes logic from NM
-[ Capture](https://www.youtube.com/watch?v=xRe8UgeA-Ko)


scripts/quests/outlands/An_Undying_Pledge.lua
```diff
@@ -0,0 +1,132 @@
...
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_ACCEPTED
+        end,
+
+        [xi.zone.NORG] =
+        {
+            ['Stray_Cloud'] =
+            {
+                onTrigger = function(player, npc)
+                    if quest:getVar(player, 'Prog') == 0 then
```
claywar: This charvar should be read into a variable and then used in the condition.

scripts/zones/Sea_Serpent_Grotto/DefaultActions.lua
```diff
@@ -2,6 +2,7 @@ local ID = zones[xi.zone.SEA_SERPENT_GROTTO]
 
 return {
     ['qm1'] = { messageSpecial = ID.text.NOTHING_OUT_OF_ORDINARY },
+    ['qm5'] = { messageSpecial = ID.text.NOTHING_OUT_OF_ORDINARY },
```
claywar: Can you please align all of those ='s in each line here? Not sure how this got out of control

scripts/quests/outlands/An_Undying_Pledge.lua
```diff
@@ -0,0 +1,132 @@
+-----------------------------------
+-- An Undying Pledge
+-----------------------------------
+-- Log ID: 5, Quest ID: 149
+-- Stray Cloud : !pos -18 1 -20 252
+-- qm5         : !pos 133 -9 221 176
+-----------------------------------
+local ssgID = zones[xi.zone.SEA_SERPENT_GROTTO]
```
claywar, line 8: Just a small nitpick here, but can you style that similar to other formats? (with --- block under the local)

scripts/quests/outlands/An_Undying_Pledge.lua
hooksta4: What do you mean by this?

scripts/quests/outlands/An_Undying_Pledge.lua
```diff
@@ -0,0 +1,132 @@
...
+                    if
+                        quest:getVar(player, 'Prog') == 1 and
+                        npcUtil.popFromQM(player, npc, ssgID.mob.GLYRYVILU, { claim = true, hide = 0 })
+                    then
+                        return quest:messageSpecial(ssgID.text.BODY_NUMB_DREAD)
+                    elseif
+                        quest:getVar(player, 'Prog') == 2
+                    then
+                        return quest:progressEvent(18)
+                    end
```
Xaver-DaRed, line 99:
```lua
                    local questProgress = quest:getVar(player, 'Prog')
                    if
                        questProgress == 1 and
                        npcUtil.popFromQM(player, npc, ssgID.mob.GLYRYVILU, { claim = true, hide = 0 })
                    then
                        return quest:messageSpecial(ssgID.text.BODY_NUMB_DREAD)
                    elseif questProgress == 2 then
                        return quest:progressEvent(18)
                    end
```
To expand:
- As claywar said, whenever you have to check the same thing more than once, store it on a variable first so we dont repeat operations.
- When there is only one condition, theres no need to align the `if/elseif` separate from the condition or the `then`

scripts/quests/outlands/An_Undying_Pledge.lua
hooksta4: you mean like local prog = quest:getVar(Blah blah_

if prog == 0 then?

claywar: Regardless of what goes on behind the scenes, I'll present the following analogy:

Let's say you're doing a research project on whales, and you need to look up the average length of a newborn humpbacked whale.  You have an encyclopedia to look it up, but every time you do, you have to flip through and find the right page.

Instead, what is being suggested is do the research once and write it down.  Xaver beat me to the punch, but I wanted to explain _why_ things like these are important, because its less overhead and a better practice.

hooksta4: Ok typically in the past if its been more than 2 id do the do that but now will make it for 2 or more. thanks.

scripts/quests/outlands/An_Undying_Pledge.lua
```diff
@@ -0,0 +1,134 @@
...
+local ssgID = zones[xi.zone.SEA_SERPENT_GROTTO]
+-----------------------------------
+local quest = Quest:new(xi.questLog.OUTLANDS, xi.quest.id.outlands.AN_UNDYING_PLEDGE)
+
+quest.reward =
+{
+    item = xi.item.LIGHT_BUCKLER,
+    fameArea = xi.fameArea.NORG,
+    fame = 50,
+}
```
Xaver-DaRed, line 17: As a general rule, please have this aligned like so (applicable to all PRs)
```lua
quest.reward =
{
    item     = xi.item.LIGHT_BUCKLER,
    fameArea = xi.fameArea.NORG,
    fame     = 50,
}
```

PR #5867, opened 2024-06-01
[Core] Change musicId and blockid types to uint16

xkoredev:
**_I affirm:_**

* Changes musicId and blockid types to uint16

This matches the client types and allows for custom music to be added if desired.

Perform actions that change the music, such as:

* Mount / Dismount a chocobo
* Use items that can change the music, such as serene serinette
* Enter jeuno during holidays (maybe fake holiday schedule)


sql/zone_settings.sql
```diff
@@ -22,10 +22,10 @@ CREATE TABLE `zone_settings` (
   `zoneip` tinytext NOT NULL,
   `zoneport` smallint(5) unsigned NOT NULL DEFAULT '0',
   `name` tinytext NOT NULL,
-  `music_day` tinyint(3) unsigned NOT NULL DEFAULT '0',
-  `music_night` tinyint(3) unsigned NOT NULL DEFAULT '0',
-  `battlesolo` tinyint(3) unsigned NOT NULL DEFAULT '0',
-  `battlemulti` tinyint(3) unsigned NOT NULL DEFAULT '0',
+  `music_day` smallint(3) unsigned NOT NULL DEFAULT '0',
```
xkoredev, line 25: How can we migrate existing DB types? Is this assumed to fall on the server maintainers? Maybe a migration script is necessary?

src/map/packets/change_music.cpp
```diff
@@ -35,6 +35,6 @@ CChangeMusicPacket::CChangeMusicPacket(uint8 BlockID, uint8 MusicTrackID)
     this->setType(0x5F);
     this->setSize(0x08);
 
-    ref<uint8>(0x04) = BlockID;      // block
-    ref<uint8>(0x06) = MusicTrackID; // music
+    ref<uint16>(0x04)  = BlockID;      // block
```
UmeboshiXI: Just a minor thing, extra space got added here.

sql/zone_settings.sql
claywar: See migration 030 (home zone size) should be what you're looking for

xkoredev: Done. Added migration and tested on a fresh db

src/map/packets/change_music.cpp
xkoredev: Done. Thanks for the catch!

sql/zone_settings.sql
cocosolos: Zone_settings isn't protected and shouldn't need a migration. Any changes servers make to this table should be done through modules, as it's expected to be dropped whenever there are changes.

zach2good: +1 to what coco said. It's a shame that zone_settings sits at the intersection of live/static data - but it is what it is

zach2good: Plus, if you have module'd inserts for this information, it's all forwards compatible anyway because the size of the field is going up

src/map/packets/change_music.cpp
```diff
@@ -35,6 +35,6 @@ CChangeMusicPacket::CChangeMusicPacket(uint8 BlockID, uint8 MusicTrackID)
     this->setType(0x5F);
     this->setSize(0x08);
 
-    ref<uint8>(0x04) = BlockID;      // block
-    ref<uint8>(0x06) = MusicTrackID; // music
+    ref<uint16>(0x04)  = BlockID;     // block
```
Xaver-DaRed: CI fails because of styling, like this one. Please correct those alignments

Xaver-DaRed: imagen

xkoredev: Fixed

sql/zone_settings.sql
xkoredev: Removed migration. Will add alter statement in modules on our end

WinterSolstice8 (approved):

zach2good (approved):

PR #5873, opened 2024-06-01
Add the ability to parry, guard, or block a player WS (adds checks to lua)

### LINKED ISSUE #3321: Dynamis - Mob Parry Combat Skill
### Checklist

- [X] I have searched for existing issues for issues like this one. The issue has not been posted. (Duplicate reports slow down development.)
- [X] I have provided reproducable steps. (No "as the title says" posts please. Provide reproducable steps even if it seems like a waste of time.)
- [X] I have provided my client version in the details. (type /ver into your game window)

### Details

Orc, Yagudo, Quadav, & Goblin mobs/NMs in Dynamis zones should be able to parry if they are jobs that have parrying combat skill (WAR, RDM, THF, PLD, DRK, BST, BRD, SAM, NIN, DRG). Currently these mobs do not parry at all.

Unsure of how much parrying skill they should have but they might be capped out at their jobs respective ranking for parrying for their level. Testing on retail has also shown that Hydra & Kindred do not parry. 

image
image
image
image
image
image
image
image
image
image

version 30181205_0

TracentEden2:
**_I affirm:_**

This PR makes adds the potential for parrying, guarding, or blocking a player weaponskill. This is done by adding lua logic for these cases to physical_utilities.lua and weaponskills.lua. The lua logic is based on the analogous core logic. 

The idea is that the logic is generic enough to be used also in the future for mobskills and eventually normal auto-attack logic once that is in lua. Though one question is weather is makes sense to try to already call up to these from within core (for auto-attacks)? I can try to do that if desired.

I made the PR draft for now to get comments on the above and from other people working on the battle system migration to lua.

Also a related bug fix, on LSB currently all dynamis beastmen can parry, however captures show that it should be only beastmen with jobs with native parry skill. Specifically I checked log of dyna farming capture by Siknoz and only saw monster parries from BRD, NIN, BST, DRG, DRK, WAR, SAM, and PLD, which aligns with most jobs that have natural parry skill. Furthermore, Eden also limited parry in the same way (see this fixed [bug report](https://github.com/EdenServer/community/issues/3321)). 

Finally, another bug fix is related to SHIELD_DEF_BONUS which should be a flat damage reduction that occurs before the absorb calculation rather than a percent reduction (see [here](https://wiki.ffo.jp/html/20048.html) and [here](https://www.bg-wiki.com/ffxi/Shield_Def._Bonus)), this was changed in both the lua and core code. The original incorrect implementation (from 6 years ago) has no source or info.

Perform WS on mobs/players that can parry, guard, and block

### DIFF
--- a/scripts/globals/combat/physical_utilities.lua
+++ b/scripts/globals/combat/physical_utilities.lua
@@ -70,6 +70,16 @@ xi.combat.physical.pDifWeaponCapTable =
     [xi.skill.THROWING        ] = { 3.25 },
 }
 
+local shieldSizeToBlockRateTable =
+{
+    [1] = 55, -- Buckler
+    [2] = 40, -- Round
+    [3] = 45, -- Kite
+    [4] = 30, -- Tower
+    [5] = 50, -- Aegis and Srivatsa
+    [6] = 100, -- Ochain  https://www.bg-wiki.com/ffxi/Category:Shields
+}
+
 local elementalGorget = -- Ordered by element.
 {
     xi.item.FLAME_GORGET,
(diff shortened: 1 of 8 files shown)


scripts/globals/combat/physical_utilities.lua
```diff
@@ -684,3 +684,291 @@ end
...
+end
+
+xi.combat.physical.isParried = function(defender, attacker)
+    local parried = false
+    if
+        xi.combat.physical.canParry(defender, attacker) and
+        xi.combat.physical.calculateParryRate(defender, attacker) > math.random(100)
+    then
+        parried = true
+        if defender:isPC() then
```
WinterSolstice8, line 948: can you add a TODO to implement this Turms mod
image

TracentEden2: > can you add a TODO to implement this Turms mod image

Yep, I added a TODO, thanks!

scripts/globals/combat/physical_utilities.lua
```diff
@@ -684,3 +684,298 @@ end
...
+            else
+                blockSkill = defender:getMaxSkillLevel(defender:getMainLvl(), defender:getMainJob(), xi.skill.SHIELD)
+            end
+        else -- No block mobmod so zero rate
+            return 0
+        end
+    end
+
+    if defender:isPC() then
+        if shieldSize == 1 then -- Buckler
```
claywar: I would probably table the blockrates and do a lookup based on shield size instead of a long condition

TracentEden2: > I would probably table the blockrates and do a lookup based on shield size instead of a long condition

Thanks for the suggestion, I updated the PR!

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5874, opened 2024-06-01
[lua][chore] Equipment.lua clean up

UmeboshiXI:
**_I affirm:_**

Replaces item IDs with item enum names.
Adds missing items to item enum lua.
Fixes BRAVURA_119_III enum using incorrect item ID.

NOTE: I noticed when swapping out the Item IDs that Homam legs are included in the artifact range? I have not looked into this further at this time on how this table is used exactly. Left a note in place.

image

Item Enum should match Item ID. No changes should be noted from existing enum usage aside from BRAVURA_119_III which was incorrect.


scripts/globals/equipment.lua
```diff
@@ -124,24 +279,24 @@ xi.equipment.relicIDs =
...
-    { 14521, 14523 }, -- ToAU Body
-    { 14928, 14930 }, -- ToAU Hand
-    { 15684, 15686 }, -- ToAU Feet
-    { 15600, 15602 }, -- ToAU Legs
-    { 16138, 16140 }, -- WotG Head
-    { 14578, 14580 }, -- WotG Body
-    { 15002, 15004 }, -- WotG Hand
-    { 15746, 15748 }, -- WotG Feet
-    { 15659, 15661 }, -- WotG Legs
+    { xi.item.FIGHTERS_MASK, xi.item.EVOKERS_HORN },           -- Original Head
```
claywar: I get these are only ranges, but could you stylize this a bit to align beginning and end for the rows?

UmeboshiXI: Adjusted

Xaver-DaRed (approved):

PR #5878, opened 2024-06-02
[lua][sql] mob_spawn_mods.sql Part 1

UmeboshiXI:
**_I affirm:_**

Migrates mods/mobmods out of mob_spawn_mods into corresponding mob scripts.

Part 01 / Part ?? to keep PR size down.

Mobs should have same modifiers as they did before, but located in their lua files. See no changes.


Xaver-DaRed (approved):

Frankie-hz (approved):
I sent ume the patch I did for some of this already, feel free to finish the entire thing <3

WinterSolstice8 (approved):

PR #5890, opened 2024-06-05
[Quest]Blackmail

hooksta4:
**_I affirm:_**

Converts quest Blackmail to Interaction Framework.
- Adds defaultAction for Dauperiat.
- Adds checks for rank 3 (any nation) in order to accept the quest. 
    - Because it is in the Chateau, it is assumed that rank 3 (or on the mission to get inside the Chateau), there is no evidence supporting from[ captures.](https://www.youtube.com/watch?v=vW7BOYG7qVE) Putting it at Rank 3 as one wiki mentions it specifically while the other does not. 

Talk to NPC to begin. Then repeat as you get floor plans.

### DIFF
--- a/scripts/quests/sandoria/Blackmail.lua
+++ b/scripts/quests/sandoria/Blackmail.lua
@@ -1,0 +1,125 @@
+-----------------------------------
+-- Blackmail
+-----------------------------------
+-- Log ID: 0, Quest ID: 71
+-- Dauperiat !pos -19 -.1 -24 231
+-- Halver !pos 2 0 -.6 233
+-----------------------------------
+local quest = Quest:new(xi.questLog.SANDORIA, xi.quest.id.sandoria.BLACKMAIL)
+
+quest.reward =
+{
+    fame     = 30,
+    fameArea = xi.fameArea.SANDORIA,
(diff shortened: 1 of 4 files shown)


scripts/quests/sandoria/Blackmail.lua
```diff
@@ -0,0 +1,125 @@
...
+            {
+                [646] = function(player, csid, option, npc)
+                    if option == 1 then
+                        quest:setVar(player, 'Prog', 2)
+                    end
+                end,
+
+                [648] = function(player, csid, option, npc)
+                    if player:hasCompletedQuest(quest.areaId, quest.questId) then
+                        quest.fame = 5
```
claywar: This is incorrect.  `fame` attached to the quest object does nothing.

claywar: Also, lua tables are passed by reference in most places.  If your intent is to modify quest.rewards, it will need to be tested with a separate character as well following the modification to ensure it is not permanently changed (my bet is it will be)

hooksta4: Good point. Ill just add the addFame() instead.

PR #5893, opened 2024-06-07
[Mission] COP 8-1 Conversion

hooksta4:
**_I affirm:_**

- Converts COP Mission  8-1 to IF.
- Removed AlTaieu/Globals (now inside mission file)
- Aligned RuAern lua to match with GetFirstID (it changed)
- Removed all logic from Towers and Portal
- Added a PromethiaStatus 1 to the door until the next mission is converted. Added TODO next to it.
- 

!addmission 6 800
Zone into AlTaieu to begin. 
Visit each Crystal to spawn 3x Ruaern
After defeating each, zone into Grand Palace
Touch Center portal for Tavnazian Ring
Examine east door to get CS with Prishe and begin next mission.
[Capture](https://www.youtube.com/watch?v=EQ9Dgmh2nqM) shows next mission does not start until touching the east gate.


scripts/missions/cop/8_1_Garden_of_Antiquity.lua
```diff
@@ -0,0 +1,244 @@
...
+-----------------------------------
+local altaieuID = zones[xi.zone.ALTAIEU]
+local palaceID  = zones[xi.zone.GRAND_PALACE_OF_HUXZOI]
+-----------------------------------
+
+local mission = Mission:new(xi.mission.log_id.COP, xi.mission.id.cop.GARDEN_OF_ANTIQUITY)
+
+local antiquityVars =
+{
+    -- [crystalOffset] = { thisFightDone, thisCsAcquired, otherCrystal, otherCrystal2, firstCsParam, secondCsParam1, secondCsParam2 }
```
claywar, line 16: This should be replaced with missionStatus, as there are bits specific for this purpose:
>index 7 = 8-1 Rubious Crystals Bits
`player:setMissionStatus(COP, 6, x)`
```
0x2 - second rubious crystal
0x3 - third rubious crystal
```

cocosolos: Just wanna note there's some minor issues with that documentation. There's a typo in that command (6 should be 7 since it's index 7) and the function signature has changed (7 and x swapped places). This mission does set index 6 to 0xF when you do the first crystal, then moves to index 7 for the last 2.

claywar:
Superseded by #5924

Issue #5891, opened 2024-06-06
🐛 San d'Orian Grapes generating SQL errors?


CatsEyeXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Windows10/MariaDB 10.6

## Branch affected by issue 

`base`

## Steps to reproduce
I unfortunately can't gather much context surrounding this error but looks like the formatting contained within this item name is causing SQL errors.

```
Zones_6.txt:[06/06/24 13:24:00:446][map][error] mysql_real_query: SQL_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'or._grape' OR sortname LIKE 'san_d'or._grape'' at line 1 (1064) (SqlConnection::QueryStr:328)
```

## Expected behavior
SQL statements should be formatted to sanely handle this item?

cocosolos:
Looks like `sortname` is only used in a couple areas and this looks like GetItemIDByName called by !additem command.

RAIST5150:
Herm... could it be the single quotes monkeying up the query?

```sql
INSERT INTO `item_basic` VALUES (4431,0,'bunch_of_san_dorian_grapes','san_dor._grape',12,1548,@INGREDIENTS,0,19);;
```

Only question it because I once saw an issue with addresses in a tax records file conversion run that was using single quotes during an extraction.  When people put in something like ST. JOHN'S  in an address it threw the formatting out of whack.

Just weird to see it cut off right where you have that weirdness we have to watch out for with San d'oria, when it isn't formatted like that in items_basic.

'san_d'or._grape'' in the query

san_dor._grape in the data

Xaver-DaRed:
This is what we have
![imagen](https://github.com/LandSandBoat/server/assets/60053999/830b5423-750c-40f1-819e-229992e675d6)

This is the error
`near 'or._grape' OR sortname LIKE 'san_d'or._grape''`

[...]

PR #5895, opened 2024-06-07
Core: Refactor GetItemIDByName function to use prepared statement

zach2good:
**_I affirm:_**

As title, to use a prepared statement

`!additem san_dor._grape`

`!additem san_d\'or._grape` Note the apostrophe

Noticed during: https://github.com/LandSandBoat/server/issues/5891


Xaver-DaRed (approved):

zach2good:
image

No errors or warnings in the log - this is all as expected. I don't think it's worth stripping apostrophes from the input, since we don't have any in the name or sortname of items

claywar (approved):

PR #5905, opened 2024-06-10
Experimental Battlefield: Ouryu Cometh

CatsEyeXI:
**_I affirm:_**

Adds an experimental battlefield: Ouryu Cometh
Mainly this PR adds a battlefield script that will make use of the missing ID's which are available to all other battlefield zones.
Without these ID's present, any battlefields added to this zone will error out in the battlefield global.

!pos 732.550 -32.5 -506.544 30
!additem cloud_evoker
Trade it to the Spatial Displacement


cocosolos:
This isn't how this battlefield works. `16900340` is the ID of the entry NPC, the Unstable Displacement at I-9. Similar to Bahamut V2 there's only 1 battlefield area in this zone and the exit NPC is `16900359`.

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
```diff
@@ -0,0 +1,45 @@
...
+    requiredItems =
+    {
+        xi.item.CLOUD_EVOKER,
+        wearMessage = riverneID.text.TIME_LIMIT_FOR_THIS_BATTLE_IS + 2,
+        wornMessage = riverneID.text.TIME_LIMIT_FOR_THIS_BATTLE_IS + 1,
+    },
+    experimental = true,
+})
+
+content:addEssentialMobs({ 'Ouryu' })
```
claywar: addEssentialMobs should not be necessary since you're explicitly defining groups and death handlers

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
```diff
@@ -0,0 +1,45 @@
...
+
+content.groups =
+{
+    {
+        mobs = { 'Ouryu' },
```
claywar: Can you please align assignments here and below?

scripts/zones/Riverne-Site_A01/IDs.lua
```diff
@@ -18,8 +18,11 @@ zones[xi.zone.RIVERNE_SITE_A01] =
         MEMBERS_LEVELS_ARE_RESTRICTED = 7023, -- Your party is unable to participate because certain members' levels are restricted.
         CONQUEST_BASE                 = 7064, -- Tallying conquest results...
         A_GLOWING_MIST                = 7259, -- A glowing mist of ever-changing proportions floats before you...
+        MEMBERS_OF_YOUR_PARTY         = 7535, -- Currently, # members of your alliance (including yourself) have clearance to enter the battlefield.
```
claywar: You're going to want to check for additional missing Text IDs that should be defined for battlefields.  Easiest way to know what is absolutely required is by searching `.text.` in battlefield.lua

CatsEyeXI: image

CatsEyeXI:
thank you both for your guidance.

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
```diff
@@ -0,0 +1,43 @@
...
+    },
+    experimental = true,
+})
+
+content.groups =
+{
+    {
+        mobs      = { 'Ouryu' },
+        superlink = true,
+        spawned   = true,
```
claywar: Spawned is the default case, and doesn't need to be declared on either of these groups.

Updated: Only one is win condition, so separate groups

CatsEyeXI: im assuming i missed your previous message before you edited and this table should remain broken out into two groups.. I've removed the default spawned condition.

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
```diff
@@ -0,0 +1,41 @@
...
+        wornMessage = riverneID.text.TIME_LIMIT_FOR_THIS_BATTLE_IS + 1,
+    },
+    experimental = true,
+})
+
+content.groups =
+{
+    {
+        mobs      = { 'Ouryu' },
+        superlink = true,
```
claywar: So just one last thing as I look with a fresh mind: Groups superlink to their own, so instead for here and below, you're going to need to use `superlinkGroup = 1` instead (can have multiple groups defined, but if they need to link together, need to be the same)

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
```diff
@@ -0,0 +1,41 @@
...
+        death     = utils.bind(content.handleAllMonstersDefeated, content),
```
claywar: This death handler will most likely need to be defined locally for when this BCNM has loot support since there is no armoury crate, but instead a direct drop to loot pool.  For now I would suggest just a comment and a function to set battlefield status to `WON`

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
```diff
@@ -0,0 +1,41 @@
...
+local content = Battlefield:new({
+    zoneId        = xi.zone.RIVERNE_SITE_A01,
+    battlefieldId = xi.battlefield.id.OURYU_COMETH,
+    maxPlayers    = 18,
+    levelCap      = 75,
+    timeLimit     = utils.minutes(30),
+    index         = 0,
+    area          = 1,
+    entryNpc      = 'Unstable_Displacement',
+    exitNpc       = 'Spatial_Displacement',
```
cocosolos: I'm not well versed on the new battlefield system. How exactly are these strings used? There are several NPC with both of these names, so will any of them trigger the battlefield functions? Do they need to be renamed to be distinguished from the other ones?

claywar: I'm more of a fan of id definitions since it's more clear, but in certain cases it will add all

claywar: For exit, it indeed applies to all so if there are some that this does not apply to, it needs a unique name

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
CatsEyeXI: retested and everything still works as it should.

image

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
CatsEyeXI: should be all set, mister. I added the missing elementals as well.

scripts/battlefields/Riverne_Site_A01/ouryu_cometh.lua
CatsEyeXI: > For exit, it indeed applies to all so if there are some that this does not apply to, it needs a unique name

Just so we're on the same page, is that out of scope for this PR? I'm happy to follow up with another to rename the exit NPC to something unique like Spatial_Displacement_BF (for battlefield)

claywar: That is absolutely in scope for this PR.  You do _not_ want to assign the exit functions to NPCs where it does not apply.

cocosolos: For what it's worth I tested this in a party so I could trigger the outside NPCs while having BCNM status and all the outside spatial displacements behaved normally and I only got the exit prompt after entering the battlefield. Still should distinguish them but good to know there seems to be some safety.

CatsEyeXI: > That is absolutely in scope for this PR. You do _not_ want to assign the exit functions to NPCs where it does not apply.

No problem, I've added a commit for that change as well.

TracentEden2:
I am wondering about the status of this PR as there is quite some logic missing from the Ouryu mob script. I can add the logic in a separate PR but not sure if I need to add battlefield code as well if this PR is dormant.

zach2good:
The author is repo-banned for reasons you can look up on your own, so I wouldn't expect this PR to go anywhere. I'll close it.

PR #5908, opened 2024-06-11
[Core] Add setMeleeRange LUA binding

0x05010705:
**_I affirm:_**

- Creates a LUA binding `mob:setMeleeRange(float)` that allows for a mobs 'hitbox' or melee range to be dynamically modified.
- This allows for mobs with larger/smaller than normal hitboxes to be changed to values that deviate from the standard range that is listed in `mob_family_pools.sql` without changing every instance of that mob.

*NOTE: This does not modify/change the behavior of mob pathing in any way, only the melee range that both the player and the mob can hit each other from.*

This helps resolve issues where a mobs flags might trigger model scaling (IE: Flag Values 0x02, 0x04) which increase the mobs size, and then players have to be right up next to the mob/inside the mob to maintain melee range.

*After Rebuilding*
1 - Engage in combat with a mob in any zone
2 - Increase the melee range with `!exec target:setMeleeRange(10)`
3 - Move your character far away and note that both the player and the mob can melee each other from this new range.

This is more impactful on things with larger models such as on Wyrms.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -15706,6 +15706,47 @@ uint8 CLuaBaseEntity::getModelSize()
 }
 
 /************************************************************************
+ *  Function: setMeleeRange()
+ *  Purpose : Sets the maximum melee range for a mob
+ *  Example : mob:setMeleeRange(12.0)
+ *  Notes   : This affects the distance players can hit the mob from
+ ************************************************************************/
+void CLuaBaseEntity::setMeleeRange(float range)
+{
+    // Only valid for mobs
+    if (m_PBaseEntity->objtype != TYPE_MOB)
+    {
+        ShowWarning("Attempt to set melee range for non-mob entity (%s).", m_PBaseEntity->getName());
+        return;
+    }
+
+    auto* PMob = static_cast<CMobEntity*>(m_PBaseEntity);
+
+    // Ensure that the cast to MobEntity worked properly and we dont have a NULL PTR
+    if (!PMob)
+    {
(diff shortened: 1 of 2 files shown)


Xaver-DaRed (approved):

zach2good:
You've added a new binding, but not used it anywhere. Can you add it to some relevant mobs to show that you've tested this properly?

0x05010705:
> You've added a new binding, but not used it anywhere. Can you add it to some relevant mobs to show that you've tested this properly?

Are you asking to update mob luas in the PR?  Or just to add it to some mobs and show that it works through screenshots?

zach2good:
Presumably, the plan is to apply this to mobs that have a large model size and, therefore a larger melee range, so do that. I'm assuming this wasn't written exclusively to use as `!exec target:setMeleeRange(10)`.

@WinterSolstice8 said something about this data being sent through `0x00E - Entity Update`. I think the long term goal would be to have 0x00E fully implemented against @atom0s's RE packet definitions (https://github.com/atom0s/XiPackets/tree/main/world/server/0x000E#graphsize) and then have this as data that's assigned at the pool level, not the script level - but this PR will be fine for now if there are big Wryms etc. that really need to have their melee ranges pushed out.

0x05010705:
Sounds good - I should have some more time to get to this in another day or two.

WinterSolstice8 (approved):
personally I don't really have a problem with this until we get around to fixing 0x00E and adding in the range to mobs

It may be a while before that happens - I think the data we want hasn't been captured before which means every mob type would have to be captured and made generic (I assume most mobs of the same type/size would have the same extra range)

but re-capping the entire game is a pretty tall order, especially because the tooling hasn't been made yet

tl;dr: 0x00E update will probably happen before the range is implemented anyway

PR #5919, opened 2024-06-14
[Job Utils] Bluemage job abilities

andeluvian:
**_I affirm:_**

This PR contains the jobutils for Blue Mage, jobabilities have been moved not to jobutils

Create a new player with blue mage job either 75 or 99 and each ability should work as they did before the change.

### DIFF
--- a/scripts/actions/abilities/azure_lore.lua
+++ b/scripts/actions/abilities/azure_lore.lua
@@ -8,13 +8,11 @@
 local abilityObject = {}
 
 abilityObject.onAbilityCheck = function(player, target, ability)
-    ability:setRecast(math.max(0, ability:getRecast() - player:getMod(xi.mod.ONE_HOUR_RECAST) * 60))
-
-    return 0, 0
+    xi.job_utils.blue_mage.checkAzureLore(player, target, ability)
 end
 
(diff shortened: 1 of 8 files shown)


zach2good:
As a general rule: your commit messages should say what your piece of work is doing in relation to the codebase.

If you introduce some code in an early commit, and then change the formatting (etc.) in a later commit, that's got no relevance to the codebase, it's specific to your code - so squash it please.

Commits as atomic pieces of work, etc.

zach2good:

zach2good:
While you're doing these PRs, the ability use binding should _always_ have the action param, even if it isn't used:
```
abilityObject.onUseAbility = function(player, target, ability, action)
    return xi.job_utils.rune_fencer.useBattuta(player, target, ability, action)
end
```

etc.

It can safely be omitted, because Lua, but Core is piping it in, so it should be shown in case people want/need it later

zach2good:
https://github.com/LandSandBoat/server/blob/base/src/map/lua/luautils.cpp#L4201

zach2good (approved):
Seems fine to me, just needs work so CI doesn't complain

zach2good:
Replaced by https://github.com/LandSandBoat/server/pull/6071

PR #5955, opened 2024-06-21
Battlefield: Partial A.M.A.N. Trove impl (default off)

zach2good:
**_I affirm:_**

This PR has been sat for ages and it's very self-contained, so I've slapped it behind some settings so others can tinker on it if they want to.

- [x] You can start the RoE for AMAN Trove
- [x] The setting will stop Greyson giving you the intro or main menu if you trigger/trade
- [x] The setting will stop you trading a mars orb to the BC entrance in Balgas and getting the BCNM menu

=== Original Description

I'm about to travel for work, so I'm putting this up in draft in case anyone wants to iterate on it while I'm away.

Main points I'm trying to hit:

It should be non-abusable (debouncing, value checks, checks against injection/inspection etc.)
It should be as accurate as we can figure out based on retail
It should be highly customisable (document everything well, expose everything globally and let people mess with it at their own peril)
Should also include:

Everything to do with Greyson, buying silver vouchers from him
ONLY Mars Orb Balgas Dias, playable Venus and other zones will come AFTER this one
We need to figure out the loot pools (Are we going to do something similar to what we have for gobbie box? - global bindings?)
The chest rolling logic should be standalone so it can be tested. This is very important because the peanut gallery in their infinite wisdom will use this as an excuse to complain about RNG. So make sure it's extractable and testable so they have nothing to complain about.
Gil reward for LT
Make sure model IDs are correct for all chests/mimics - also between runs they should be properly reset!
All messaging correct
Tenshodo teleports should also work (if they aren't already)
The logic to replace R/E items and abjurations with whatever items if you already have them
Personal/shared loot pools? I don't remember how this works
Steps to test these changes
TODO

Make sure all TODOs are handled
Make sure Greyson works as intended
Make sure relevant parts are broken out into settings
Make sure Monthly voucher reset is handled somehow or well documented
Make sure loot tables are as good as they can be, and also easily customisable
Make sure the Mimic is as correct as we can get it, with the correct moves etc.


scripts/globals/aman_trove.lua
```diff
@@ -0,0 +1,210 @@
...
+xi.amanTrove.prepareMimic = function(player, npc)
+    print("The chest is a mimic!")
+
+    npc:setStatus(xi.status.UPDATE)
+    npc:setModelId(258)
+    npc:setAnimation(1) -- Combat
+    npc:setAnimationSub(1) -- Mouth open
+
+    -- Make things disappear:
+    -- npc:entityAnimationPacket('kesu', npc)
```
Xaver-DaRed: We have this packet names enum'd.
Instead of `'kesu'` use `xi.animationString.STATUS_DISAPPEAR`

(I know it's commented out, but for future use)

cocosolos:
Just some notes:
- Terminal Coffer also has loot (no message).
- If you already have an R/Ex item I don't think you get any substitute, just less loot.
- Gobbie box items are marked in client, these aren't. Similar to ANV Key.
- Loot is normal BCNM rules (shared pool).

zach2good:
Updated this to now be my partial implementation, hidden completely behind a default-off setting which comes with the same warning Monstrosity has

PR #5959, opened 2024-06-22
resolve error: call to non-'constexpr' function

zer0blivion:
**_I affirm:_**

This resolves an error message encountered when building:
```
/server/src/login/auth_session.cpp: In function 'constexpr bool {anonymous}::isBcryptHash(const string&)':
/server/src/login/auth_session.cpp:54:32: error: call to non-'constexpr' function 'std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::length() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type = long unsigned int]'
   54 |         return (passHash.length() == 60 &&
      |                 ~~~~~~~~~~~~~~~^~
In file included from /usr/include/c++/11/string:55,
                 from /server/docker_build/_deps/asio-src/asio/include/asio/ssl/context.hpp:20,
                 from /server/docker_build/_deps/asio-src/asio/include/asio/ssl.hpp:18,
                 from /server/src/login/auth_session.h:24,
                 from /server/src/login/auth_session.cpp:22:
/usr/include/c++/11/bits/basic_string.h:926:7: note: 'std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::length() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::size_type = long unsigned int]' declared here
  926 |       length() const _GLIBCXX_NOEXCEPT
      |       ^~~~~~
```
By using the constexpr function ``std::size()`` instead of ``std::string.length()``

Build successfully

### DIFF
--- a/src/login/auth_session.cpp
+++ b/src/login/auth_session.cpp
@@ -51,7 +51,7 @@ namespace
 
     constexpr bool isBcryptHash(const std::string& passHash)
     {
-        return passHash.length() == 60 &&
+        return std::size(passHash) == 60 &&
                passHash[0] == '$' &&
                passHash[1] == '2' &&
                (passHash[2] == 'a' || passHash[2] == 'b' || passHash[2] == 'y' || passHash[2] == 'x') && // bcrypt hash versions


zach2good:
Interesting, I didn't know std::size was constexpr.

What compiler and version are you using? We don't see this issue on any of our mainline combinations (but we're due to upgrade everything to whatever Ubuntu 24.04 provides soon)

zach2good:
I'm away, so I can't look at the docs: does std size return the equivalent length, or is it retuning the non-filled capacity?

zer0blivion:
I'm building on CachyOS, using podman and the Dockerfile from the wiki: https://github.com/LandSandBoat/server/wiki/Docker with cmake
-- The C compiler identification is GNU 11.4.0

https://en.cppreference.com/w/cpp/iterator/size
std::size(passHash) will return a constexpr auto of passHash.size()  

As I understand it, there's no functional difference between string.size() and string.length(), so this should work as intended.

zach2good:
I wonder why this catches for you, we build with gcc13 et al. https://github.com/LandSandBoat/server/blob/257f2b383ae6b48b9ea0aab932520cd7a834721b/.github/workflows/build.yml#L279

With warnings cranked up, with a set of exceptions 🤷🏻‍♂️

zach2good (approved):

PR #5967, opened 2024-06-26
Monster Follow Pathing

jmcmorris:
**_I affirm:_**

This is old code I wrote for Eden about 5 years ago. It adds a simple method for monsters to follow some target.

```lua
mob:follow(target, xi.follow.ROAM)
```

This is useful for many monsters such as:
 * Slave Globes for Mother Globe
 * Hpemde monster family
 * Lamia No.19 pets (almost any pet really)
 * Apkallu (they runaway from the player)

* Travel to Mother Globe with `!gotoid 17506396` and observe pathing
* You can kill the slaves out of order with `!hp 0` and they will properly adjust which mob they are following
* You can engage with Mother Globe and the slaves properly path the the character and after the character dies they resume their follow pathing

Additional steps taken from #3629 
1. Go to Al'taieu.
2. Kill a baby Ul'xzomit and respawn it. Observe it automatically re-attaches.
3. Kill a mother Ul'xzomit. Observe the baby Ul'xzomits start roaming by themselves.
4. Kill a baby Ul'xzomit and respawn it. Observe it starts roaming by itself.
5. Respawn the mother Ul'xzomit. Observe the babies start following it again.
6. Fight a mother Ul'xzomit. Kill a baby Ul'xzomit and respawn it. Observe the baby only links when it gets close.
7. Kill an Ul'xzomit and let it respawn normally. Observe its respawn time is now 5.5 minutes.


src/map/ai/controllers/mob_controller.cpp
```diff
@@ -568,40 +568,60 @@ void CMobController::DoCombatTick(time_point tick)
...
-        {
-            Disengage();
-            return;
-        }
+        Disengage();
+        return;
+    }
 
-        TryLink();
+    TryLink();
```
Xaver-DaRed, line 577: shouldnt this remain inside the PTarget check?

jmcmorris: At the top of TryLink it does a PTarget nullptr check so this is safe

jmcmorris: Turns out there is already some logic for following inside https://github.com/LandSandBoat/server/blob/base/scripts/globals/follow.lua

I think it might make the most sense to combine these two since there is some nice logic for handling multiple followers (like we need to with Mother Globe) but it is also missing some functionality such as the runaway. I'm not positive the existing logic works with mobs following players/npcs either but this new follow logic does.

jmcmorris:
Turns out there is already some logic for following inside https://github.com/LandSandBoat/server/blob/base/scripts/globals/follow.lua

I think it might make the most sense to combine these two since there is some nice logic for handling multiple followers (like we need to with Mother Globe) but it is also missing some functionality such as the runaway. I'm not positive the existing logic works with mobs following players/npcs either but this new follow logic does.

scripts/zones/The_Shrine_of_RuAvitau/mobs/Slave_Globe.lua
```diff
@@ -6,6 +6,21 @@ local ID = zones[xi.zone.THE_SHRINE_OF_RUAVITAU]
 -----------------------------------
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:addListener("DEATH", "SLAVE_DEATH", function(mobArg, killer)
```
Xaver-DaRed: `mob:addListener('DEATH', 'SLAVE_DEATH', function(mobArg, killer)`

scripts/enum/follow.lua
```diff
@@ -0,0 +1,10 @@
+-----------------------------------
+-- Follow Behavior
+-----------------------------------
+
+xi.follow =
```
Frankie-hz: missing `xi = xi or { }`

jmcmorris: fixed

jmcmorris:
I have made an additional commit to the initial two that reworks the original follow logic: https://github.com/LandSandBoat/server/pull/5967/commits/95886fbd6ddad0c8081d0d3690ba5847c1fda0b2

sql/mob_spawn_mods.sql
```diff
@@ -47,48 +47,224 @@ INSERT INTO `mob_spawn_mods` VALUES (16867642,55,180,1); -- IDLE_DESPAWN: 180
...
+INSERT INTO `mob_spawn_mods` VALUES (16912796,71,1,1); -- LEADER: 1
+INSERT INTO `mob_spawn_mods` VALUES (16912797,71,-1,1); -- LEADER: -1
+
+-- Om'Xzomit
+INSERT INTO `mob_spawn_mods` VALUES (16912800,71,1,1); -- LEADER: 1
+INSERT INTO `mob_spawn_mods` VALUES (16912801,71,-1,1); -- LEADER: -1
+
+-- Om'Xzomit
+INSERT INTO `mob_spawn_mods` VALUES (16912805,71,1,1); -- LEADER: 1
+INSERT INTO `mob_spawn_mods` VALUES (16912806,71,-1,1); -- LEADER: -1
```
Xaver-DaRed: Couldnt this be placed in mob scripts instead of the database?

There is work going on to retire this table.

jmcmorris: The only way to do this would be if the mother and children have different scripts. Currently they have the same script and so there is no way to distinguish them from inside the script. This database allows us to individualize them.

Frankie-hz: If we add this into the `ID.lua`:
`ULXZOMIT_OFFSET = GetTableOfIDs('Ulxzomit'),`

Inside the Ulxzomit script we can put something like:

```lua
local leaderTable =
{
    ID.mob.ULXZOMIT_OFFSET[1],
    ID.mob.ULXZOMIT_OFFSET[4],
    ID.mob.ULXZOMIT_OFFSET[7],
    ID.mob.ULXZOMIT_OFFSET[10],
    ID.mob.ULXZOMIT_OFFSET[13],
    ID.mob.ULXZOMIT_OFFSET[16],
    ID.mob.ULXZOMIT_OFFSET[19],
    ID.mob.ULXZOMIT_OFFSET[22],
    ID.mob.ULXZOMIT_OFFSET[25],
    ID.mob.ULXZOMIT_OFFSET[28],
    ID.mob.ULXZOMIT_OFFSET[31],
    ID.mob.ULXZOMIT_OFFSET[34],
    ID.mob.ULXZOMIT_OFFSET[37],
    ID.mob.ULXZOMIT_OFFSET[40],
    ID.mob.ULXZOMIT_OFFSET[43],
}

    local mobId    = mob:getID()
    local leaderId = leaderTable[mobId]

    if leaderId ~= nil then
        mob:setMobMod(xi.mobMod.LEADER, 1)

        local childOne = mobId  + 1
        local childTwo = mobId  + 2
        childOne:setMobMod(xi.mobMod.LEADER, -1)
        childTwo:setMobMod(xi.mobMod.LEADER, -2)
    end
end
```
Note: this is untested of course, throwing out ideas since that SQL is being removed

claywar: Just a brief glance, but even that could be simplified with some modulo action on the tabled result

jmcmorris: That's a good idea @Frankie-hz! I can change things to work like that. A modulo might work for the Ul'Xzomit since I think they all have 2 babies each but the Om'Xzomit have 1 or 2 babies each.

Frankie-hz: I don't really see a pattern for Omxzomit so would have to think of something else for those.

jmcmorris: Frank's suggestion mostly worked. I have every mob calculate its own LEADER value and put the logic in `xi.follow.assignLeaderMod()` for easy reuse.

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -568,40 +568,59 @@ void CMobController::DoCombatTick(time_point tick)
...
-            return;
+            if (!PMob->PAI->PathFind->IsFollowingPath())
+                PMob->PAI->PathFind->PathTo(PFollow->loc.p);
+            PMob->PAI->PathFind->FollowPath(m_Tick);
         }
-        else if (IsSpecialSkillReady(currentDistance) && TrySpecialSkill())
+        else
         {
-            return;
+            PMob->PAI->EventHandler.triggerListener("RUN_AWAY", CLuaBaseEntity(PMob), CLuaBaseEntity(PFollow));
```
cocosolos: Can you document this in AI_Events.txt please. Not sure why CI is passing since it did pick this up...

jmcmorris: Good catch! Done!

Xaver-DaRed:
Needs rebase

Xaver-DaRed (approved):
Lua side looks fine.

Core side looks fine aswell, but needs another set of eyes.

src/map/lua/lua_baseentity.cpp
```diff
@@ -1999,6 +1999,44 @@ void CLuaBaseEntity::wait(sol::object const& milliseconds)
     PBattle->PAI->Inactive(std::chrono::milliseconds(waitTime), true);
 }
 
+/************************************************************************
+ *  Function: follow(target)
+ *  Purpose : Makes a Mob follow the provided target
+ *  Example : mob:follow(target, dsp.follow.ROAM)
```
zach2good: dsp -> xi

src/map/ai/controllers/mob_controller.h
```diff
@@ -88,6 +101,7 @@ class CMobController : public CController
     time_point m_NeutralTime;
     time_point m_WaitTime;
     time_point m_mobHealTime;
+    FOLLOWTYPE m_follow;
```
zach2good: This would be better named as `m_followType`, otherwise it comes off as "I am following this entity", rather than the type of following

src/map/ai/controllers/mob_controller.h
```diff
@@ -25,6 +25,13 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "controller.h"
 #include "entities/mobentity.h"
 
+enum FOLLOWTYPE
```
zach2good: Can we get this as an enum class, with the FollowType::None style naming?

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -1264,6 +1302,30 @@ bool CMobController::Cast(uint16 targid, SpellID spellid)
     return CController::Cast(targid, spellid);
 }
 
+void CMobController::SetFollowTarget(CBaseEntity* PTarget, FOLLOWTYPE followType)
```
zach2good: Would it make sense to default one or both of these to nullptr and None? Setting the target to nullptr AND the follow type to None in the same function seems a little smelly, any other ideas?

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -1122,7 +1147,9 @@ void CMobController::Reset()
     PMob->m_neutral = true;
     m_NeutralTime   = m_Tick;
 
-    PTarget = nullptr;
+    PTarget  = nullptr;
+    PFollow  = nullptr;
```
zach2good: PFollow -> PFollowTarget, I reckon makes more sense

scripts/zones/AlTaieu/mobs/Ulxzomit.lua
```diff
@@ -2,43 +2,41 @@
...
 -----------------------------------
+mixins = { require('scripts/mixins/follow') }
+local ID = zones[xi.zone.ALTAIEU]
+-----------------------------------
 local entity = {}
 
-local followOptions = { forceRepathInterval = 1 }
+local leaders =
+{
+    [ID.mob.ULXZOMIT_OFFSET[1]] = 2,
```
zach2good: If these are all max distances, and then we set a group max distance of 2 in the call to assignLeaderMod, isn't all of this redundant?

scripts/mixins/follow.lua
```diff
@@ -0,0 +1,41 @@
+require('scripts/globals/mixins')
```
zach2good, line 14: Not sure if we need these requires anymore (maybe check with clay)

But we 100% need a header in here saying what this mixin is, and how to use it, and maybe why you'd use this and not the raw :follow/:unfollow bindings

scripts/enum/mob_mod.lua
```diff
@@ -78,7 +78,7 @@ xi.mobMod =
     ALLI_HATE              = 68, -- Range around target to add alliance member to enmity list.
     NO_LINK                = 69, -- If set, mob cannot link until unset.
     NO_REST                = 70, -- Mob cannot regain hp (e.g. re-burrowing antlions during ENM).
-    LEADER                 = 71, -- Used for mobs that follow a defined "leader", such as Ul'xzomit mobs.
+    LEADER                 = 71, -- Used for mobs that follow a defined "leader". Positive designates a leader and number of follows. Negative is the leader ID relative to the mob.
```
zach2good: Is it ALWAYS the case that the IDs of leaders will be lower than the followers, am I understanding that right?

scripts/enum/follow.lua
```diff
@@ -0,0 +1,11 @@
+-----------------------------------
+-- Follow Behavior
+-----------------------------------
+xi = xi or {}
+
+xi.follow =
```
zach2good: followType or followBehaviour

scripts/globals/follow.lua
```diff
@@ -6,144 +6,19 @@ require('scripts/globals/utils')
...
-            -- Go to leader while leaving room for other followers.
-            targetAngle = getAverageWorldAngle(leaderPos, filteredFollowersPos)
-            targetAngle = targetAngle + (separationAngle * (followerIndex - 1)) - (separationAngle * (followerCount - 1) / 2)
-        end
-
-        local targetPos = utils.getNearPosition(leaderPos, followDistance, targetAngle)
-        follower:pathTo(targetPos.x, targetPos.y, targetPos.z, options.pathFlags)
-    end
-end
+xi.follower = xi.follower or {}
```
zach2good: This now no longer lines up with the file that contains is follow / follower etc.

src/map/ai/controllers/mob_controller.h
jmcmorris: That's how I like to do it but I was trying to follow examples (pun intended)

src/map/ai/controllers/mob_controller.cpp
jmcmorris: I opted for adding `ClearFollowTarget()` which ended up actually being desired in several places

scripts/zones/AlTaieu/mobs/Ulxzomit.lua
jmcmorris: The `leaders` table is read from checking for which mobs are the leaders. I was able to trim this down to the more compact code.
```lua
local leaders = {}
for i = 0, 14 do
    leaders[ID.mob.ULXZOMIT_OFFSET[1 + i * 3]] = 2
end
```

scripts/enum/mob_mod.lua
jmcmorris: Yes, the ID of leaders is always lower than the followers. I reworded this comment to hopefully clarify things a bit better.

`Used for mob following. Positive number is how many followers a leader has. A negative number is the distance from this mob ID to the leader ID.`

src/map/ai/controllers/mob_controller.h
```diff
@@ -25,6 +25,13 @@ along with this program.  If not, see http://www.gnu.org/licenses/
 #include "controller.h"
 #include "entities/mobentity.h"
 
+enum class FollowType : uint8
+{
+    None,
+    Roam,
+    RunAway,
+};
```
Xaver-DaRed, line 33: We usually have enums in all caps, i think

jmcmorris: That's what I've primarily (solely?) seen but I made this change after being requested by @zach2good above. Maybe I misunderstood the casing request though?

Xaver-DaRed: I dunnow, tbh. Waitingon Zach then.

zach2good: This is correct for enum class, old style enums are in all caps

zach2good (approved):

PR #5971, opened 2024-06-28
[Core] Adds missing quantity count to charutils getItemCount

loxleygames:
**_I affirm:_**

`getItemCount` isn't factoring in quantity, eg. `player:getItemCount(1455)` returns 1 if a player has 1 Byne Bill x99.
This change fixes the issue.

Tested with stacking and non-stacking items and if player has no matching items.


Xaver-DaRed (approved):

Xaver-DaRed:
Nothing like having 5 functions with the same name

PR #5974, opened 2024-06-30
Marid Tusk Breaking

jmcmorris:
**_I affirm:_**

This implements the family logic for Marids to has their tusks broken and drop as loot. Marid's can have their tusks broken at a 20% chance per critical hit (similar to Imps and their horns). Whenever this happens the Marid animation changes to reflect this. Additionally, whenever the Marid dies it will drop Marid Tusks equal to the number broken.

This is logic I originally implemented for Eden years ago. Here is the capture I used thanks to @Wiggo32: https://www.youtube.com/watch?v=UEwacJNiyoU

1. Go fight a Marid `!gotoid 16986144`
2. Hit it with critical attacks to see its tusks break
3. Kill it with varying number of broken tusks to see that the number of tusks dropped are the same as the broken tusks

### DIFF
--- a/scripts/mixins/families/marid.lua
+++ b/scripts/mixins/families/marid.lua
@@ -1,0 +1,29 @@
+require('scripts/globals/mixins')
+
+g_mixins = g_mixins or {}
+g_mixins.families = g_mixins.families or {}
+
+g_mixins.families.marid = function(maridMob)
+    -- 20% chance to break tusk on critical hit
+    maridMob:addListener('CRITICAL_TAKE', 'MARID_CRITICAL_TAKE', function(mob)
+        if not utils.chance(20) then
+            return
+        end
+
+        local broken = mob:getAnimationSub()
+        if broken >= 2 then
+            return
+        end
+
+        mob:setAnimationSub(broken + 1)
+    end)
+
+    maridMob:addListener('ITEM_DROPS', 'MARID_ITEM_DROPS', function(mob, loot)
+        local broken = mob:getAnimationSub()
+        for _ = 1, broken do
+            loot:addItem(xi.item.MARID_TUSK, xi.drop_rate.GUARANTEED)
+        end
+    end)
+end
+
+return g_mixins.families.marid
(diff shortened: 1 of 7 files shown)


scripts/mixins/families/marid.lua
```diff
@@ -0,0 +1,29 @@
...
+        end
+    end)
+
+    maridMob:addListener('ITEM_DROPS', 'MARID_ITEM_DROPS', function(mob, loot)
+        local broken = mob:getLocalVar('TuskBroken')
+        for _ = 1, broken do
+            loot:addItem(xi.item.MARID_TUSK, xi.drop_rate.GUARANTEED)
+        end
+    end)
+end
```
Xaver-DaRed, line 27: If animationSub happens to be the same as what the local var will be, couldnt we just check for the animationSub?

Xaver-DaRed: Like so...

```lua
g_mixins.families.marid = function(maridMob)
    -- 20% chance to break tusk on critical hit
    maridMob:addListener('CRITICAL_TAKE', 'MARID_CRITICAL_TAKE', function(mob)
        local brokenTusks = mob:getAnimationSub()

        if
            brokenTusks < 2 and
            math.random(1, 100) <= 20
        then
            mob:setAnimationSub(brokenTusks + 1)
        end
    end)

    maridMob:addListener('ITEM_DROPS', 'MARID_ITEM_DROPS', function(mob, loot)
        local brokenTusks = mob:getAnimationSub()

        for _ = 1, brokenTusks do
            loot:addItem(xi.item.MARID_TUSK, xi.drop_rate.GUARANTEED)
        end
    end)
end
```

Xaver-DaRed:
Also, there are marids in Nyzul Island aswell.

jmcmorris:
I spent an hour running around killing Marid on retail. It took awhile to find one in Nyzul Isle but I confirmed that they do have their tusks break so I fixed that. I also tested the ones in Abyssea Vunkerl and they do not have their tusks break. This video is good evidence that Marid NMs do not have this mechanic either. https://www.youtube.com/watch?v=3RazpCA2y3g&feature=youtu.be

Xaver-DaRed (approved):

PR #5975, opened 2024-06-30
Fix cpp ci only reporting stderr

cocosolos:
**_I affirm:_**

Writes stdout to file instead of just stderr. Fixes issue with CI passing when it souldn't.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,7 +78,7 @@ jobs:
         for changed_file in ${{ steps.changed-files.outputs.all }}; do
           if [[ -f $changed_file ]]; then
             if [[ $changed_file == *.cpp ]]; then
-              bash tools/ci/cpp.sh ${changed_file} 2>> cpp_checks.txt || true
+              bash tools/ci/cpp.sh ${changed_file} 2>> cpp_checks.txt >> cpp_checks.txt || true
             fi
           fi
         done


zach2good (approved):

Xaver-DaRed (approved):

.github/workflows/build.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         for changed_file in ${{ steps.changed-files.outputs.all }}; do
           if [[ -f $changed_file ]]; then
             if [[ $changed_file == *.cpp ]]; then
-              bash tools/ci/cpp.sh ${changed_file} 2>> cpp_checks.txt || true
+              bash tools/ci/cpp.sh ${changed_file} >> cpp_checks.txt || true
```
WinterSolstice8: I can't get this to redirect both stderr and stdout without `2>&1` (redirecting stderr to stdout)

file contents of bash script:
```
printf "stderr\n" >&2
printf "stdout\n" >&1
```

```
./test.sh 2>&1 > temp ; echo "temp contents:" ; cat temp
stderr
temp contents:
stdout
```

and...

```
(./test.sh 2>&1) > temp ; echo "temp contents:" ; cat temp
temp contents:
stderr
stdout
```

but also
```
./test.sh >> temp 2>>temp ; cat temp
stderr
stdout
```
works, so perhaps...

```suggestion
              bash tools/ci/cpp.sh ${changed_file} 2>> cpp_checks.txt >> cpp_checks.txt || true
```
?

WinterSolstice8: Note: I am not sure of the order of the output if you redirect stderr/stdout in that fashion

cocosolos: None of the other checks use stderr so wasn't sure if this was intentional or not. I don't think the order matters so I can change to that if you want.

WinterSolstice8: Yeah I don't think order matters either. It will output "problems to solve", I think it would be worth having both just in case

WinterSolstice8 (approved):

PR #5979, opened 2024-07-02
Qutrub skill changes for retail accuracy

jmcmorris:
**_I affirm:_**

Many changes made to Qutrub mob skills to make them more retail accurate. This is primarily based off of all the work Jimmayus did found in the following document:
https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=57955395#gid=57955395

One of the larger changes here is the addition of a new function, `xi.mobskills.calculateDuration`, which simply calculates the duration giving the minimum/maximum and the TP which evenly lerps between the two as expected.

Go fight a Qutrub (`!gotoid 16998466`) and test the various skills.

### DIFF
--- a/scripts/actions/mobskills/leaping_cleave.lua
+++ b/scripts/actions/mobskills/leaping_cleave.lua
@@ -20,9 +20,13 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
     local accmod = 3
-    local dmgmod = 4
+    local dmgmod = 2.25
     local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
     local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.PHYSICAL, xi.damageType.SLASHING, info.hitslanded)
+
+    local duration = xi.mobskills.calculateDuration(mob:getTP(), 15, 30)
+    xi.mobskills.mobPhysicalStatusEffectMove(mob, target, skill, xi.effect.STUN, 1, 0, duration)
+
     target:takeDamage(dmg, mob, xi.attackType.PHYSICAL, xi.damageType.SLASHING)
     return dmg
 end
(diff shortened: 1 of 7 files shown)


Xaver-DaRed (approved):

claywar (approved):

scripts/actions/mobskills/leaping_cleave.lua
```diff
@@ -20,9 +20,13 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
     local accmod = 3
-    local dmgmod = 4
+    local dmgmod = 2.25
     local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
     local dmg = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.PHYSICAL, xi.damageType.SLASHING, info.hitslanded)
+
+    local duration = xi.mobskills.calculateDuration(mob:getTP(), 15, 30)
```
Frankie-hz, line 27: This doesnt have to be in this PR, maybe Xaver knows better than me but I believe the check inside `xi.mobskills.mobStatusEffectMove` should take care of the difference in duration. I always thought it should be the duration is (in this case) 30 and then the resist check will modify that duration based on the resist. Jimmayus sheet says " Stun duration varies with TP(?)" so its not 100% confirmed. My guess would be its just based off the resist state like everything else.

Xaver-DaRed: Probably the case for enfeeblement mobskills, but for the time being, with the info we have, this is the best course of action.
At least it seems sane that first the TP would affect the base duration and then multipliers, such as resist states, would take effect after.

jmcmorris: I went and tested on retail. My understanding is that resists can be 1/2 or 1/4 on retail for status effects. I was testing with a 99 job so I suspect I kept resisting but I was getting either 3~ or 7~ second stuns. Since the mob was below 25% they'd use the mobskill at 1000TP. With 15s stun a 1/4 resist would be about 3s and 1/2 would be about 7~. I think the only way to get more accurate with this would be to get some 75 characters that wouldn't resist much and test it a bunch.

PR #5989, opened 2024-07-09
[Fix] Correct trust recast time from 24 to 240 seconds

Xaver-DaRed:
**_I affirm:_**

Adds a missing 0 to all trust spell recast timers.

Summon a trust. See the recast be 240 seconds. (Thats 4 minutes, btw.)


claywar (approved):

zach2good (approved):
We desperately need to mark database columns and members in core relating to time with `_s` and `_ms`, and maybe another term for when things are in `1/10'000th of second` - this sort of thing is way to easy to miss for a long time (in another PR, obvs).

PR #5990, opened 2024-07-09
[Core] Implement Fish Ranking Contest

paladindamarus:
**_I affirm:_**

- Implement Fish Ranking contest in Selbina
- Adds Pelican Ring item and effect
- Adds database tables for contest and player fishing history (history to be implemented separately)
- Includes `!addfish` command to create fish items with size data in exdata
- Modifies packet guard to allow 0x04b packet with player in CS substate
- Updates packet handler for 0x04b (fish ranking data shares packet type with server message request)
- Adds entry to `xi.settings.main` to auto-run fishing contest

- Once added, server instance that loads Selbina will auto-initiate the contest
- If no contest exists, it will create one
- Speak to Chenon in Selbina to interact
- Use `!addfish` to create appropriate fish type with size/length data
- Trade fish to Chenon during appropriate stage of contest

## To-Do

- Pelican ring effect should stack to maximum 2, but issues with effect overwrite prevents this.  Require update to core to adjust?
- Create commands for manual control (note: there are menus built in to the client for GM/Staff control of contest, but it is unknown if this is used or put on auto-pilot by retail servers)


scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
...
+    { id = 4474, name = 'gigant_squid'        , realname = 'Gigant Squid'    },
+    { id = 4475, name = 'sea_zombie'          , realname = 'Sea Zombie'      },
+    { id = 4476, name = 'titanictus'          , realname = 'Titanictus'      },
+    { id = 4477, name = 'gavial_fish'         , realname = 'Gavial Fish'     },
+    { id = 4478, name = 'three-eyed_fish'     , realname = 'Three-eyed Fish' },
+    { id = 4479, name = 'bhefhel_marlin'      , realname = 'Bhefhel Marlin'  },
+    { id = 4480, name = 'gugru_tuna'          , realname = 'Gugru Tuna'      },
+    { id = 5120, name = 'titanic_sawfish'     , realname = 'Titanic Sawfish' },
+    { id = 5127, name = 'gugrusaurus'         , realname = 'Gugrusaurus'     },
+    { id = 5129, name = 'lik'                 , realname = 'Lik'             },
```
Xaver-DaRed: This need xi.item enums

scripts/items/pelican_ring.lua
```diff
@@ -0,0 +1,49 @@
...
+    -- Allow for duplicate enchantment effects, max 2
+    if target:getMod(effect, subtype) < 2 then
+        target:addStatusEffectEx(effect, effect, power, tick, duration, subtype, subpower, tier, flag)
+    end
+end
+
+itemObject.onEffectGain = function(target, effect)
+    if
+        target:getMod(xi.mod.PELICAN_RING_EFFECT) < 2
+    then
```
Xaver-DaRed: `if target:getMod(xi.mod.PELICAN_RING_EFFECT) < 2 then`

Xaver-DaRed:
I would recomend:
- Make the changes requested by CI
- Forget Pelican ring. Can be done in a different PR. The smaller the PR, the better for everyone involved.
- This will need testing on base LSB

scripts/items/pelican_ring.lua
paladindamarus: I'm okay to change that if you want, but the style guide on the wiki says to avoid single-line conditions:
image

Good either way, just want to confirm!

paladindamarus:
> I would recomend:
> 
> * Make the changes requested by CI
> * Forget Pelican ring. Can be done in a different PR. The smaller the PR, the better for everyone involved.
> * This will need testing on base LSB

Will do, thanks!  I'll remove the pelican ring from the items, but leave it as the reward (it will just be a useless item until fixed).  
As for testing:  I didn't state it, but the branch I built this from (despite coming from an ASB repo) is a pure LSB branch pulled from lsb/base.  Only done that way because I can't create another fork from both LSB and ASB at the same time.  However, this was done on my pure LSB installation.

zach2good:
Leaving a comment so I get email updates. I'll be doing the C++ reviews once CI passes

scripts/items/pelican_ring.lua
zach2good:
```lua
    if target:getMod(xi.mod.PELICAN_RING_EFFECT) < 2 then -- This is a single condition
        target:addMod(xi.mod.PELICAN_RING_EFFECT, 1)
    end
```

Is what Xaver was getting at.

```lua
if singleCondition then
   ....
end

if
    conditionA and
    condition
then
    ...
end
```

scripts/zones/Selbina/Zone.lua
```diff
@@ -7,12 +7,19 @@ local zoneObject = {}
...
 zoneObject.onInitialize = function(zone)
     xi.server.setExplorerMoogles(ID.npc.EXPLORER_MOOGLE)
+    InitializeFishingContestSystem()
 end
 
 zoneObject.onGameHour = function(zone)
     SetServerVariable('Selbina_Destination', math.random(1, 100))
 end
 
+zoneObject.onZoneTick = function(zone)
```
zach2good, line 17: If there are no players in the zone, the contest won't tick - is this intentional?

sql/char_fishing.sql
```diff
@@ -0,0 +1,22 @@
+-- --------------------------------------------------------
+-- Add history for player fishing data
+-- --------------------------------------------------------
+
+CREATE TABLE IF NOT EXISTS `char_fishing` (
```
zach2good: If this is just the history for a given player, then this should be `char_fishing_history`

Does this also need to be added to `triggers.sql` so it isn't left behind on char deletion? Or is it more like AH history?

sql/char_fishing.sql
```diff
@@ -0,0 +1,22 @@
+-- --------------------------------------------------------
+-- Add history for player fishing data
```
zach2good: This comment isn't like the other ones you've added, make it more like those - or remove it. It's pretty obvious these days that all of these files are defining tables.

sql/char_fishing.sql
zach2good: All the other definitions are DROP then CREATE

sql/char_fishing.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE TABLE IF NOT EXISTS `char_fishing` (
+    -- Fishing Catch History
+    `charid`           int(10) unsigned NOT NULL,
+    `fish_list`        blob,
+    `fish_linescast`   int(6)  unsigned NOT NULL DEFAULT 0,
+    `fish_reeled`      int(5)  unsigned NOT NULL DEFAULT 0,
+    `fish_longest`     int(4)  unsigned NOT NULL DEFAULT 0,
+    `fish_longest_id`  int(4)  unsigned NOT NULL DEFAULT 0,
+    `fish_heaviest`    int(4)  unsigned NOT NULL DEFAULT 0,
+    `fish_heaviest_id` int(4)  unsigned NOT NULL DEFAULT 0,
```
zach2good: Out of interest, why are the display widths for these all a bit different?

sql/char_fishing.sql
```diff
@@ -0,0 +1,22 @@
...
+
+    -- Fishing Contest History
+    `contest_rank_1`   smallint(3)  unsigned NOT NULL DEFAULT 0,
+    `contest_rank_2`   smallint(3)  unsigned NOT NULL DEFAULT 0,
+    `contest_rank_3`   smallint(3)  unsigned NOT NULL DEFAULT 0,
+    `contest_rank_4`   smallint(3)  unsigned NOT NULL DEFAULT 0,
```
zach2good: Why are catch and contest history split here, but are part of the same table? Is this hinting that they should be separate tables?

sql/fishing_contest.sql
```diff
@@ -0,0 +1,13 @@
+--
+-- Table structure for table `fishing_contest`
+-- This table holds the parameters for the current fish ranking contest
+--
+
+CREATE TABLE IF NOT EXISTS `fishing_contest` (
```
zach2good, line 7: Is this going to be a single entry at all times, or does the entire history forever need to be retained (I have no idea how fishing or contests work)

sql/fishing_contest_entries.sql
```diff
@@ -0,0 +1,21 @@
...
+    `race`        tinyint(2)  unsigned    NOT NULL,
+    `allegiance`  tinyint(2)  unsigned    NOT NULL,
+    `fishrank`    tinyint(2)  unsigned    NOT NULL,
+    `score`       int(4)      unsigned    NOT NULL DEFAULT 0,
+    `submittime`  int(4)      unsigned    NOT NULL DEFAULT 0,
+    `contestrank` tinyint(2)  unsigned    NOT NULL DEFAULT 0,
+    `share`       tinyint(2)  unsigned    NOT NULL DEFAULT 0,
+    `claimed`     tinyint(1)  unsigned    NOT NULL DEFAULT 0,
+    PRIMARY KEY (`name`) -- Only one entry per character
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
```
zach2good: Is there a reason why a lot of this data isn't pulled from the submitting char with a JOIN?

src/map/lua/lua_baseentity.cpp
```diff
@@ -17709,6 +17710,98 @@ uint8 CLuaBaseEntity::getMannequinPose(uint16 itemID)
     return 0;
 }
 
+// Fishing Contest Utilities
+
+void CLuaBaseEntity::submitContestFish(uint32 score)
+{
+    if (auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity))
```
zach2good, line 17925: +1 for correct use of dynamic_cast, probably also want a warning for the cast where this fails (in all bindings too)

src/map/lua/lua_baseentity.cpp
```diff
@@ -17709,6 +17710,98 @@ uint8 CLuaBaseEntity::getMannequinPose(uint16 itemID)
...
+
+    if (m_PBaseEntity->objtype == TYPE_PC)
+    {
+        auto*       PChar = static_cast<CCharEntity*>(m_PBaseEntity);
+        const char* Query = "SELECT contestrank, share "
+                            "FROM   fishing_contest_entries "
+                            "WHERE  name = '%s' "
+                            "AND    claimed != 1; ";
+
+        int32 ret = _sql->Query(Query, PChar->getName());
```
zach2good: Please use the new `db::` system with `preparedStatement()`, there's lots of examples in `xi_world` and `xi_search`

src/map/lua/lua_baseentity.cpp
```diff
@@ -17709,6 +17710,98 @@ uint8 CLuaBaseEntity::getMannequinPose(uint16 itemID)
...
+
+        if (ret != SQL_ERROR && _sql->NumRows() > 0 && _sql->NextRow() == SQL_SUCCESS)
+        {
+            reward["rank"]  = _sql->GetUIntData(0);
+            reward["share"] = _sql->GetUIntData(1);
+        }
+    }
+
+    return reward;
```
zach2good, line 17990: Is an empty table being returned herer OK?

src/map/lua/lua_baseentity.cpp
```diff
@@ -17709,6 +17710,98 @@ uint8 CLuaBaseEntity::getMannequinPose(uint16 itemID)
...
+}
+
+auto CLuaBaseEntity::getContestRankHistory() -> sol::table
+{
+    auto table = lua.create_table();
+
+    if (m_PBaseEntity->objtype == TYPE_PC)
+    {
+        auto history = fishingcontest::GetFishingContestHistory(static_cast<CCharEntity*>(m_PBaseEntity));
```
zach2good: Use dynamic cast inside the if here and then pass to GetFishingContestHistory

src/map/lua/lua_baseentity.cpp
```diff
@@ -17709,6 +17710,98 @@ uint8 CLuaBaseEntity::getMannequinPose(uint16 itemID)
...
+        table[1]     = get<0>(history);
+        table[2]     = get<1>(history);
+        table[3]     = get<2>(history);
+        table[4]     = get<3>(history);
```
zach2good: Could these table entries be named and then not extracted by index?

src/map/lua/lua_baseentity.cpp
```diff
@@ -17709,6 +17710,98 @@ uint8 CLuaBaseEntity::getMannequinPose(uint16 itemID)
...
+    }
+
+    return table;
```
zach2good, line 18012: Is an empty table here OK?

src/map/lua/lua_baseentity.cpp
```diff
@@ -17709,6 +17710,98 @@ uint8 CLuaBaseEntity::getMannequinPose(uint16 itemID)
...
+}
+
+void CLuaBaseEntity::claimContestReward()
+{
+    if (auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity))
+    {
+        return fishingcontest::ClaimContestReward(PChar);
```
zach2good: function is marked as returning void, but there is a return here - which is intended?

src/map/lua/lua_item.cpp
```diff
@@ -370,6 +370,31 @@ void CLuaItem::setExData(sol::table const& newData)
...
 
+auto CLuaItem::getFishData() -> sol::table
+{
+    if (!fishingutils::IsFish(m_PLuaItem))
+    {
+        ShowError("ERROR: Attempted to retrieve fish data from non-fish item.");
+        return sol::lua_nil;
+    }
+
+    if (auto PFish = static_cast<CItemFish*>(m_PLuaItem))
```
zach2good: Using static_cast here means that ANY non nullptr pointer will pass this check and garbage will go inside the if statement

src/map/lua/lua_item.cpp
```diff
@@ -370,6 +370,31 @@ void CLuaItem::setExData(sol::table const& newData)
...
+    {
+        sol::table table = lua.create_table();
+
+        table["length"] = PFish->GetLength();
+        table["weight"] = PFish->GetWeight();
+        table["ranked"] = PFish->IsRanked();
+
+        return table;
+    }
+    else
```
zach2good: This else is redundant, you'll get here if the if fails because of the return (we still want the logging and the lua_nil return)

src/map/lua/luautils.cpp
```diff
@@ -5724,4 +5732,47 @@ namespace luautils
...
+        // This function will destroy the current contest and start a new one
+        fishingcontest::InitNewContest();
+    }
+
+    void SetContestParameters(uint16 fishId, uint8 measure, uint8 criteria)
+    {
+        // Note: Validation of fish item id should be done in lua, since the fish table is also in lua
+        fishingcontest::SetContestFish(fishId);
+        fishingcontest::SetContestMeasure((FISHING_CONTEST_MEASURE)std::min((uint8)measure, (uint8)CONTEST_MEASURE_SMALLEST));
+        fishingcontest::SetContestCriteria((FISHING_CONTEST_CRITERIA)std::min((uint8)criteria, (uint8)CONTEST_CRITERIA_BOTH));
```
zach2good: Don't use c-style casting, especially not if you have enums. Make your function take the enums as their args

src/map/lua/luautils.cpp
```diff
@@ -5724,4 +5732,47 @@ namespace luautils
...
+    }
+
+    void ProgressFishingContest()
+    {
+        // This function will get called every tick when automatic progression is enabled
```
zach2good, line 5743: Even if fishing is turned off?

src/map/lua/luautils.cpp
```diff
@@ -5724,4 +5732,47 @@ namespace luautils
...
+        fishingcontest::ProgressContest();
+    }
+
+    void InitializeFishingContestSystem()
+    {
+        // IMPORTANT: This should only be called on the Zone Init in Selbina
```
zach2good, line 5749: Could you add safety checks for this (and logging) to make sure this is enforced?

src/map/packet_guard.cpp
```diff
@@ -35,6 +35,7 @@ namespace PacketGuard
         allowList[SUBSTATE_IN_CS][0x016] = true; // Entity Information Request
         allowList[SUBSTATE_IN_CS][0x01A] = true; // Player Action
         allowList[SUBSTATE_IN_CS][0x03A] = true; // Sort Inventory
+        allowList[SUBSTATE_IN_CS][0x04B] = true; // Fish Ranking Packet
```
zach2good, line 38: 👍

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+            }
+        }
+    }
+    else if (msg_type == 2) // Fish Ranking Packet
+    {
+        // The Message Chunk acts as a "sub-type" for the request
+        // 1 = First packet of ranking table
+        // 2 = Subsequent packet of ranking table
+        // 10 = ???
+        // 11 = ??? Prepare to withdraw?
```
zach2good, line 2399: Are these unknown, or the actual messages (I don't know fishing at all)

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+        // 12 = Response to a fish submission (No ranking or score - both 0) - Before ranking
+        // 13 = Fish Rank Self, including the score and rank (???) following fish submission (How is it ranked??)
+        std::vector<FishingContestEntry*> entries;
+
+        uint8                maxFakes    = 15;
+        uint8                maxEntries  = 100;
```
zach2good: Can/should these be broken out into settings?

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+        uint8                realEntries = fishingcontest::FishingRankEntryCount();
+        uint8                fakeEntries = realEntries >= maxFakes ? 0 : (uint8)std::min((int)maxFakes, maxFakes - realEntries); // Max 15 Fakes
```
zach2good: No c-style casts

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+        uint8                msg_entries = std::min(maxEntries, (uint8)(realEntries + fakeEntries));                             // Max 100 entries
```
zach2good: No c-style casts

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+        FishingContestEntry* entry       = nullptr;
+        uint8                entryVal    = 0;
+        uint8                blockSize   = sizeof(FishingContestEntry); // Should be 36
```
zach2good: Can you enforce this with a static_assert somewhere?

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+            {
+                // For type 1, include the player's Fish Rank Self packet from the contest.  Otherwise, just use some current info.
+                if (msg_chunk == 1)
+                {
+                    entry->resultCount = msg_entries;
+                }
+                else
+                {
+                    FishingContestEntry* original = entry;
+                    entry                         = new FishingContestEntry;
```
zach2good: I'm _very_ surprised CI hasn't gotten angry here: no use of raw new/delete. Use shared/unique_ptr if you need heap allocation

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+
+                    strcpy(entry->name, PChar->name.c_str());
```
zach2good: Can this be strcpy_n, or whatever the bounded version is? Also, std::

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+                    entry->contestRank = original->contestRank;
+                    entry->resultCount = msg_entries;
+                    entry->share       = original->share;
+                    entry->dataset_b   = original->dataset_b;
+                }
+                entries.push_back(entry);
+            }
+            else // Get the base data if there's no player data
+            {
+                entry = new FishingContestEntry;
```
zach2good: As above, no raw new

src/map/packet_system.cpp
```diff
@@ -8424,7 +8529,7 @@ void PacketParserInitialize()
     PacketSize[0x03D] = 0x00; PacketParser[0x03D] = &SmallPacket0x03D; // Blacklist Command
     PacketSize[0x041] = 0x00; PacketParser[0x041] = &SmallPacket0x041;
     PacketSize[0x042] = 0x00; PacketParser[0x042] = &SmallPacket0x042;
-    PacketSize[0x04B] = 0x0C; PacketParser[0x04B] = &SmallPacket0x04B;
+    PacketSize[0x04B] = 0x00; PacketParser[0x04B] = &SmallPacket0x04B;
```
zach2good, line 8511: Does this HAVE to be 0x00 (unchecked) because the size is variable, or can you look up what the size is meant to be?

src/map/packets/fishing_rank.cpp
```diff
@@ -0,0 +1,84 @@
+﻿/*
+===========================================================================
+
+  Copyright (c) 2010-2015 Darkstar Dev Teams
```
zach2good: Wrong year/project name

src/map/packets/fishing_rank.h
```diff
@@ -0,0 +1,38 @@
+﻿/*
+===========================================================================
+
+  Copyright (c) 2010-2015 Darkstar Dev Teams
```
zach2good: Same here

src/map/packets/fishing_rank.h
```diff
@@ -0,0 +1,38 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see http://www.gnu.org/licenses/
+
+===========================================================================
+*/
+
+#ifndef _CFishRankingPacket_H
```
zach2good: I dont remember what newer files use, please check to see if we use pragma once or header guards - if we use header guards they're all caps though

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,716 @@
...
+
+    // Contest Control
+    void ProgressContest()
+    {
+        /***********************************************
+         *  Update the phase of the fishing contest
+         *  See enum FISHING_CONTEST_STATUS in fishingutils.h
+         *  CALLED BY ZONE TICK IN SELBINA - DO NOT ADD TO FUNCTION
+         ***********************************************/
+        uint32 currentTime = (uint32)time(NULL);
```
zach2good: use C++ time functions

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,716 @@
...
+                return;
+            }
+        }
+    }
+
+    void UpdateRankingHistory()
+    {
+        // This should only be run once per contest, once the contest progresses from "Accepting" to "Releasing"
+        // Traverse the entries and update the database for each player
+        for (FishingContestEntry* it : FishingContestEntries)
```
zach2good: rename it to entry, it isn't an iterator in this case

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,716 @@
...
+            if (ret != SQL_ERROR && _sql->NumRows() != 0 && _sql->NextRow() == SQL_SUCCESS)
+            {
+                charID = _sql->GetUIntData(0);
+            }
+
+            if (charID > 0)
+            {
+                // Get the rank for column header
+                uint8 rankGroup = 0;
+                if (it->contestRank > 0 and it->contestRank <= 4)
```
zach2good: we don't use verbal `and` or `or` (I don't know how CI hasn't caught all of this...), use the ampersand and pipe versions

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,716 @@
...
+            }
+        }
+    }
+
+    void ScoreContest()
+    {
+        TracyZoneScoped;
+
+        // Sort the list first
+        // clang-format off
```
zach2good, line 222: Good try! you've wrapped the lambda correctly, but the formatting inside is wrong 😆 

It's truly a shame that clang-format hasn't made this possible yet

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,716 @@
...
+                            "WHERE name = '%s';";
+
+        int32 ret = _sql->Query(Query, PChar->getName());
+        if (ret == SQL_ERROR)
+        {
+            ShowDebug("Unable to remove award on fishing contest for player %s.", PChar->getName());
+        }
+    }
+
+    auto GetFishingContestHistory(CCharEntity* PChar) -> std::tuple<uint16, uint16, uint16, uint16>
```
zach2good: Can this tuple be a named struct? This is very anonymous

src/map/utils/fishingcontest.h
```diff
@@ -0,0 +1,150 @@
...
+#include <vector>
+
+class CBasicPacket;
+class CCharEntity;
+
+enum FISHING_CONTEST_CRITERIA : uint8
+{
+    CONTEST_CRITERIA_SIZE   = 0,
+    CONTEST_CRITERIA_WEIGHT = 1,
+    CONTEST_CRITERIA_BOTH   = 2
```
zach2good: Missing ending commas on some of these enums

src/map/utils/fishingcontest.h
```diff
@@ -0,0 +1,150 @@
...
+
+#include <cstring>
+#include <list>
+#include <map>
+#include <vector>
+
+class CBasicPacket;
+class CCharEntity;
+
+enum FISHING_CONTEST_CRITERIA : uint8
```
zach2good: Can these be enum classes?

src/map/utils/fishingcontest.h
```diff
@@ -0,0 +1,150 @@
...
+    CONTEST_STATUS_CLOSED     = 6,
+};
+
+struct FishingContest
+{
+    FISHING_CONTEST_STATUS   status     = CONTEST_STATUS_CONTESTING;
+    FISHING_CONTEST_CRITERIA criteria   = CONTEST_CRITERIA_SIZE;
+    FISHING_CONTEST_MEASURE  measure    = CONTEST_MEASURE_GREATEST;
+    uint16                   fishId     = 0;
+    uint32                   startTime  = 0xEFFFFFFF; // to allow for changetimes to not overflow
```
zach2good: Can you explain this? It isn't immediately obvious how this works

zach2good (changes requested):
Only glanced though the C++ changes, it looks like the C++ CI hasn't run properly... (or we're missing rules)

scripts/zones/Selbina/Zone.lua
paladindamarus: Yes -- The tick is only checking to see if the stage progression should trigger (and then progressing if it should).  Once someone is in the zone, it will progress.

sql/char_fishing.sql
paladindamarus: Good point on triggers -- I did not add it there, but will.  

This is an on-going history of the player's fishing, as there is an NPC they can talk to (to be added in separate PR) that will tell them how many fish they've ever caught, biggest fish, etc.

paladindamarus: Would that not force the table to be dropped anytime someone ran the file even accidentally?  I thought that was only used for static data?  Does it just get added to the "protected" list in dbtool then?

sql/char_fishing.sql
paladindamarus: I feel like I may have typoed some of those.  I intended for them to be roughly in line with what would make sense (would someone actually cast 1 million lines?  Maybe...) but then the IDs should be 5-wide and not 4.  I can adjust these to be standardized, but not sure if there's a preferred way of having them done.

zach2good:
See this section in dbtool.py

# These are the 'protected' files
player_data = [
    "accounts.sql",
    "accounts_banned.sql",
    "auction_house_items.sql",
    "auction_house.sql",
    "char_blacklist.sql",
    "char_chocobos.sql",
    "char_effects.sql",
    "char_equip.sql",
...

scripts/items/pelican_ring.lua
paladindamarus: Just realized I totally misunderstood what you were getting at here -- sorry about that!  I'll adjust this one.

src/map/utils/fishingcontest.cpp
paladindamarus: absolutely -- I only did it this way because it's only used once to return the rank 1 / rank 2 / rank 3 / rank 4-10 values.  I'll build a struct!

src/map/utils/fishingcontest.h
paladindamarus: I can probably get rid of this and just use 0 as the start time -- When I initially coded this on Horizon, I was storing completed contests in the database and starting new ones in cache before updating the DB, so I needed a start time trigger of '0' to flag the contest as beginning, and I used a guaranteed future time just to show that it's in a creation state.  I didn't use 0xFFFFFFFF because if it somehow got through and calculated stage timings, it would overflow.

sql/char_fishing.sql
cocosolos: Add them to the protected table in dbtool and it'll handle making sure there's no overwrites. I'm actually curious _why_ we were dropping the tables in the first place and not just using `IF NOT EXISTS` like here. It doesn't even seem to be consistent across all protected tables. Either way, adding it to dbtool helps with the "lite backup" option.

src/map/packet_system.cpp
cocosolos: Looks like it checks for `delete` but not `new`.

sql/fishing_contest_entries.sql
cocosolos: Also `name` isn't necessarily unique. IDK if that could lead to any issues, but probably want to use charid instead and a join like Zach said.

sql/char_fishing.sql
paladindamarus: No reason they can't be.  I only grouped them because they're both history of the character's fishing activities.  They can be char_fishing_history and char_fishing_contest_history if that's preferred.

sql/fishing_contest.sql
paladindamarus: For now, it's a single entry.  I built the original of this to capture all history, but took that out since it's not really something that retail does.  With this method, there should only be a single entry in the table, ever.  I could do this as a server var as well, but this made it easier to have the data not buried by everything else.

sql/fishing_contest_entries.sql
paladindamarus: Only done this way for consistency with the transmitted packet.  Can very easily be a join.  However, if a character gets deleted, their contest entry should remain intact on the leaderboard, by name.

src/map/lua/lua_baseentity.cpp
paladindamarus: Ahhh whoops.  It should be void, which is why it worked and I didn't pick it up.  Will adjust.

sql/fishing_contest_entries.sql
cocosolos: The issue I'm seeing is that if a character is deleted it'll still exist in the chars table, and a new char could be created with the same name. Would that prevent the new character from also entering the contest?

src/map/lua/luautils.cpp
paladindamarus: Not sure I understand this one -- the `std::min` function wont take the enums as they are, but spits out an `int`, hence the casting to the enum to pass into the function.

src/map/lua/luautils.cpp
paladindamarus: Not if the fishing contest is turned off... but, I could easily modify the lua calls to only reach into here if both fishing contest is automatic AND fishing is enabled, if that works?

src/map/packet_system.cpp
paladindamarus: These are unknown.  I saw them in the packets, but there was no discernable difference with those msg_chunk values.

src/map/packet_system.cpp
paladindamarus: Yes and no -- The 100 max limit is a limit on the client (it wouldn't handle more than 100).  The 15 fake entries is what shows on retail, but it will work with or without them.  That can be an easy option.

src/map/packet_system.cpp
paladindamarus: TIL.  I'll work that in!

src/map/packet_system.cpp
paladindamarus: It's because the size is variable depending on the type.  Packet 0x04B is size 0x0C for a regular server message packet, but different size for the ranking packet (I don't remember the actual size ... 0x18 I think?)  I can re-investigate this.

scripts/commands/addfish.lua
```diff
@@ -0,0 +1,88 @@
...
+                                        {
+                                            [0] = bit.band(length, 0x00FF),
+                                            [1] = bit.rshift(bit.band(length, 0xFF00), 8),
+                                            [2] = bit.band(weight, 0x00FF),
+                                            [3] = bit.rshift(bit.band(weight, 0xFF00), 8),
+                                            [4] = ranked,
+                                        } })
+    if obtained then
+        player:messageSpecial(ID.text.ITEM_OBTAINED, itemToGet)
+    end
```
zach2good: Can/should this be done with npcUtils?

scripts/commands/addfish.lua
```diff
@@ -0,0 +1,88 @@
...
+        length == 0 or
+        length == nil or
+        weight == 0 or
+        weight == nil
+    then
+        error(player, 'Size/Weight not provided.')
+        return
+    end
+
+    -- TODO: check qty and stack size + remaining inventory space instead of hardcoded == 0 check
```
zach2good: Address TODO

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
...
+-- Data for Fishing
+-----------------------------------
+require('scripts/globals/utils')
+-----------------------------------
+xi = xi or {}
+xi.fishingcontest = xi.fishingcontest or {}
+
+xi.fishingcontest.fish =
+{
+    -- NOTE: If you don't want certain fish included in the contest (era restrictions, or whatever),
```
zach2good, line 18: Good note 👍

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
+-----------------------------------
+-- Data for Fishing
+-----------------------------------
+require('scripts/globals/utils')
+-----------------------------------
+xi = xi or {}
+xi.fishingcontest = xi.fishingcontest or {}
```
zach2good: xi.fishingContest, like others: `xi.additionalEffect = xi.additionalEffect or {}`

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
+-----------------------------------
+-- Data for Fishing
```
zach2good: If this is _just_ the data, and there's a driver somewhere else the whole file can/should be `fishing_constest_data.lua`

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
...
+    for i = 0, trade:getSlotCount() - 1 do
+        if trade:getItemId(i) == fish then
+            return i
+        end
+    end
+
+    return 0
+end
+
+-- The npcUtil.giveItem function currently does not support custom exdata
```
zach2good, line 124: Would be useful to add exdata to npcUtil.giveItem?

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
...
+-- Since we need to include the measurements of the fish, we need a modified function
+local function giveFish(player, params)
+    params = params or {}
+    local ID = zones[player:getZoneID()]
+    local fishid = params.id
+    -- does player have enough inventory space?
+    if player:getFreeSlotsCount() < 1 then
+        local messageId = (ID.text.ITEM_CANNOT_BE_OBTAINED + 4) or ID.text.ITEM_CANNOT_BE_OBTAINED
```
zach2good: I don't understand what's happening here? In Lua the only things that are falsy are nil and false, not 0. So if any part of ID.text.ITEM_CANNOT_BE_OBTAINED is nil, then so will the other or side

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
...
+
+    return nil
+end
+
+-----------------------------------
+-- HOOKED FUNCTIONS
+-----------------------------------
+xi.fishingcontest.selectContestFish = function()
+    local fishList = xi.fishingcontest.fish
+    return xi.fishingcontest.fish[math.random(#fishList)]['id']
```
zach2good: We have utils to wrap this so you can do utils.randomEntry(fishList)['id'], I think

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
...
+        stage < xi.fishingcontest.status.CLOSED
+    then
+        return stageStartTime + xi.fishingcontest.interval[stage]
+    end
+
+    return xi.fishingcontest.interval[xi.fishingcontest.status.CLOSED]
+end
+
+xi.fishingcontest.onStageProgress = function(newStage)
+    -- Place holder -- I believe there is an announcement made by the NPC at certain stage progressions
```
zach2good: TODO this

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
...
+end
+
+xi.fishingcontest.onTrigger = function(player, npc)
+    local contest = GetFishingContest()
+
+    player:startEvent(10006, {
```
zach2good, line 311: Sneaky sneaky, I totally forgot startEvent() could take a table in it's sol::variadic_args

scripts/globals/fishing_contest.lua
```diff
@@ -0,0 +1,417 @@
...
+    end
+end
+
+xi.fishingcontest.onEventFinish = function(player, csid, option, npc)
+    local contest = GetFishingContest()
+
+    if csid == 10007 then
+        -- Give the player back a fish with the same stats, but tagged for Ranking
+        if option == 145 then
+            local length = player:getLocalVar('[FishContest]Length')
```
zach2good, line 453: Is this needed on either side of the trade? Or is this a thing that can be lost if the player zones?

sql/char_fishing.sql
zach2good: As I'm looking at this more (and I still don't know how any part of fishing works), this is denormalised right? We'll only ever have one row per char which we read, update, and then write back to the db with their best results etc.? 

We absolutely cannot have a forever history of  every single entry of every fish of every char that has been submitted to every contest etc. - because we'll then have to search over it and having that historical data isn't worth the cost (like it is for AH, which we have utils to truncate)

sql/fishing_contest.sql
zach2good: I think it's fine to have the table with a single row, we do something very similar with conquest I think, and will probably do the same in the future with besieged - no need to stuff everything into server vars

src/map/lua/luautils.cpp
```diff
@@ -230,6 +231,13 @@ namespace luautils
         lua.set_function("RoeParseRecords", &roeutils::ParseRecords);
         lua.set_function("RoeParseTimed", &roeutils::ParseTimedSchedule);
 
+        // Fishing Contest Functions
+        lua.set_function("GetFishingContest", &luautils::GetFishingContest);
+        lua.set_function("InitNewFishingContest", &luautils::InitNewFishingContest);
+        lua.set_function("SetContestParameters", &luautils::SetContestParameters);
+        lua.set_function("ProgressFishingContest", &luautils::ProgressFishingContest);
+        lua.set_function("InitializeFishingContestSystem", &luautils::InitializeFishingContestSystem);
```
zach2good, line 246: I _think_ you'll have to add these to the Lua ci scripts if you haven't already, so it doesn't complain about unknown global funcs

src/map/lua/luautils.cpp
zach2good: Don't worry about it for now, this'll need extra plumbing in core for the safety I eventually want. Just make sure you replace your c-style casts with static_casts etc.

(the plumbing I'm looking at: https://github.com/Neargye/magic_enum)

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,133 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
+    uint8  msg_language    = data.ref<uint8>(0x07);  // Language request id (2 = English, 4 = French)
+    uint32 msg_timestamp   = data.ref<uint32>(0x08); // The message timestamp being requested
+    uint32 msg_offset      = data.ref<uint32>(0x10); // The offset to start obtaining the server message
+    uint32 msg_request_len = data.ref<uint32>(0x14); // The total requested size of send to the client
 
-    // TODO: kill player til theyre dead and bsod
-    const char* fmtQuery = "SELECT version_mismatch FROM accounts_sessions WHERE charid = %u";
-    int32       ret      = _sql->Query(fmtQuery, PChar->id);
-    if (ret != SQL_ERROR && _sql->NextRow() == SQL_SUCCESS)
+    // uint8  msg_unknown1   = data.ref<uint8>(0x05);  // Unknown always 0
```
zach2good: Could you check with atom0s and see if he knows what the other unknown parts of this packet are? He's usually great for this sort of thing

src/map/packet_system.cpp
zach2good: Ah, I remember now, we still use a lot of raw `new` in the packet system and I haven't had time to fix it yet. The point is that if you `new` up memory, you're going to leak it if you don't eventually and consistently `delete` it later on. Smart pointers handle this for you when they fall out of scope (or their ref count reaches zero). 

It's important here because it's in a packet handler, so a sufficiently motivated bad actor could send in many bad packets, make the server allocate a bunch of memory that'll never be free'd, and then starve the process/machine of resources

zach2good: and as a general rule about C++, initialisation is scary and you should be careful! I'll try and explain as best I can:

```cpp
// default init
// default constructor will be called, but member vars won't be default initialised (despite the name), and so will be left uninitialised - typically you'd never use this. Uninitialised memory is garbage, remember.
// You've manually assigned all the members of FishingContestEntry a default value, so those will be used, but this isn't a typical use-case.
auto a = new Object;

// value init
// default constructor will be called, and member vars will be value initialised depending on their types - which is much safer.
auto a = new Object();
```

zach2good: Behold, nightmares: https://en.cppreference.com/w/cpp/language/initialization

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,716 @@
...
+        int32       ret   = _sql->Query(Query);
+
+        if (ret == SQL_ERROR)
+        {
+            ShowDebug("Error removing contest data.");
+            return;
+        }
+
+        // Clear the fishing contest entries from cache and database
+        FishingContestEntries.clear();
```
zach2good, line 622: In the current form, this is a memory leak of all the fishing contest entries - the vector of pointers is cleared, but the memory those pointers point to is not.

You're going to refactor things to be smart pointers, so this code won't need to change - it'll work as-is. 

If you were to keep the current manual memory management this would need to look like this:
```cpp
for (auto entry : FishingContestEntries)
{
    delete entry;
}
FishingContestEntries.clear()
```

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,716 @@
...
+#include "charutils.h"
+#include "common/vana_time.h"
+#include "itemutils.h"
+#include "lua/luautils.h"
+#include "zoneutils.h"
+
+namespace fishingcontest
+{
+    FishingContest                    CurrentFishingContest; // Currently active contest
+    std::vector<FishingContestEntry*> FishingContestEntries; // Currently active contest entries
```
zach2good: If these are only used in here, put them into an anonymous namespace (I think you can see this in zone_entities)

zach2good (commented):
Another round of reviews before I start work

zach2good:
And as a general question: did this work start off somewhere else and does it need git attribution/co-author data, or is it entirely new? I say that because of the DSP headers on the C++ files, and some familiar-looking patterns in the C++ code

scripts/commands/addfish.lua
paladindamarus: I would like it to be but I'd need to add code to npc_util to accept exdata.  I can certainly do that if you want me to build it into this PR.

scripts/globals/fishing_contest.lua
paladindamarus: I'm happy to do that if you want it in this PR, but yes -- it would definitely be useful (probably for others too, I assume?)

sql/char_fishing.sql
paladindamarus: That is correct.  At the start of every contest, the entry table is wiped clean, and the only thing that gets saved is the number of times a player has achieved either rank 1, rank 2, rank 3, or rank 4-10.  One row per char only.,

paladindamarus:
> And as a general question: did this work start off somewhere else and does it need git attribution/co-author data, or is it entirely new? I say that because of the DSP headers on the C++ files, and some familiar-looking patterns in the C++ code

This is entirely my work alone, except for some data from captures (some mine, some from the capture discord), and the packet header data that I got from the PV tool / atom0s.  I wrote it initially specifically on Horizon, but it had a bunch of extra features that I'm ripping out (hence why are there a few of the oddities in the code -- i refactored and missed some things).  The DSP headers were just me copying a new file header to start, and I never went back to change the header.  I probably copied some patterns from other spots in the code, mostly looking for the proper structure of things.

sql/char_fishing.sql
paladindamarus: For info, I've removed these entries in the upcoming push as they will be separate when I add the fishing history

sql/fishing_contest_entries.sql
paladindamarus: > The issue I'm seeing is that if a character is deleted it'll still exist in the chars table, and a new char could be created with the same name. Would that prevent the new character from also entering the contest?

In theory yes, but that's a case I'm not sure how retail handles.  I could simply remove the entry if the character is deleted.

paladindamarus: As I just played around with this, I think it's far more straightforward to just use the name in the table rather than the charid.  If I use the ID, I need to do another table lookup any time it gets references as the player is not guaranteed to be online if another player pulls up the list, for example.  I'd be more worried about lookup errors using the ID.

cocosolos: Not sure about retail either, and it's definitely a fringe issue. Plus I was made aware that we don't actually even open a name back up for reuse on deletion yet, but should in the future. I would use charid as primary key since that for sure will never be changed or reused under normal circumstances and just pull all of the char data that is currently here with a join instead, including name.

cocosolos: > If I use the ID, I need to do another table lookup any time it gets references as the player is not guaranteed to be online if another player pulls up the list, for example. I'd be more worried about lookup errors using the ID.

Not sure what you mean. Player being online or not doesn't change anything about char lookup by charid.

src/map/lua/lua_item.cpp
paladindamarus: This may disappear when I push the updates, but I've removed this function completely since I've realized that I can't cast to a fish item.  Instead, I'm just using the getExData function and working with the data in lua directly instead.

WinterSolstice8: > I can't cast to a fish item. 

Would this work?

https://github.com/LandSandBoat/server/pull/5704/files

Edit: this has a `new` in it so be careful about the life of the object

scripts/globals/fishing_contest.lua
paladindamarus: I copied this from the `npcUtil.giveItem` function to be consistent as an interim (as mentioned before, the npcUtil functions don't currently handle exdata so I had to make my own for now):
image

I can adjust this, but I'm not really sure what to adjust it to if the goal is to guarantee that the ITEM_CANNOT_BE_OBTAINED is not null (or more specifically, equals a particular value)

scripts/globals/fishing_contest.lua
paladindamarus: There shouldn't be any zoning since this is during the interaction with the NPC.  The local var is just a temporary holder for the fish stats so that they can receive the same fish back but marked for ranking (i.e. so they cannot use it a second time).

paladindamarus:
Latest batch of updates applied, addressing comments above.  Note that the one check that failed seems to be a socket binding error that I cannot fix on my end.  Everything else seems good.

src/map/lua/lua_item.cpp
paladindamarus: Winter:  I just saw that now.  It probably _would_ work, but I've removed any casting to a `CFishItem` right out of the code and just use the ExData functions.  When I wrote this on Horizon, we didn't have the `getExData` and `setExData`, but those work perfectly for this.

src/map/utils/fishingutils.cpp
```diff
@@ -2891,8 +2891,11 @@ namespace fishingutils
 
         if (PItem && FishList[itemid])
         {
-            // CItemFish constructor uses `const CItem&` input so this is ok
-            return new CItemFish(*PItem);
+            if (CItemUsable* PItemUsable = dynamic_cast<CItemUsable*>(PItem))
+            {
+                // CItemFish constructor uses `const CItem&` input so this is ok
+                return new CItemFish(*PItemUsable);
```
zach2good: Even if this was here before you got here, it's still a memory leak. You can fix it like so: https://godbolt.org/z/dKoez1hvc

src/map/utils/fishingcontest.h
```diff
@@ -0,0 +1,164 @@
...
+{
+    // Contest Data: Sets
+    void SetContestStatus(FISHING_CONTEST_STATUS newStatus);
+    void SetContestCriteria(FISHING_CONTEST_CRITERIA newCriteria);
+    void SetContestMeasure(FISHING_CONTEST_MEASURE newMeasure);
+    void SetContestFish(uint16 newFish);
+    void SetContestStartTime(uint32 startTime);
+
+    // Contest Data: Gets
+    FISHING_CONTEST_STATUS   GetContestStatus();
```
zach2good: With all of these long types, you can use auto and trailing return syntax to make it less eye-bleedy

src/map/utils/fishingcontest.h
```diff
@@ -0,0 +1,164 @@
...
+    CONTEST_CRITERIA_BOTH   = 2,
+};
+
+enum class FISHING_CONTEST_MEASURE : uint8
+{
+    CONTEST_MEASURE_GREATEST = 0,
+    CONTEST_MEASURE_SMALLEST = 1,
+};
+
+enum class FISHING_CONTEST_STATUS : uint8
```
zach2good, line 49: Notice that the vast majority of all our other enum class declarations don't repeat the name of the enum:
```cpp
enum class Emote : uint8
{
    POINT      = 0,
    BOW        = 1,
    SALUTE     = 2,
    KNEEL      = 3,
    LAUGH      = 4,
    CRY        = 5,
    NO         = 6,
    YES        = 7,
    ...
```

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,790 @@
...
+
+        // Clear old data and store new data in the DB
+        WriteInitialContestData();
+    }
+
+    void LoadContestParameters()
+    {
+        // load the overhead contest parameters
+        std::string Query = "SELECT "
+                            "status, "       // 0
```
zach2good: No need to annotate with the column number, we look up by name as you've already done below

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,790 @@
...
+        ret   = db::query(Query);
+
+        if (!ret)
+        {
+            ShowDebug("Error removing contest entry data.");
+            return;
+        }
+
+        // Clamp vars, just in case
+        criteria = (FISHING_CONTEST_CRITERIA)std::min((uint8)criteria, (uint8)FISHING_CONTEST_CRITERIA::CONTEST_CRITERIA_BOTH);
```
zach2good: c-style casts -> static_casts

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,790 @@
...
+
+namespace
+{
+    // Cache containers for Contest data
+    FishingContest                   CurrentFishingContest; // Currently active contest
+    std::vector<FishingContestEntry> FishingContestEntries; // Currently active contest entries
+    std::vector<FishingContestEntry> FakeContestEntries;    // Pre-established vector of placeholder entries
+
+    // Define the "next stage" for each stage for automatic progression
+    std::unordered_map<FISHING_CONTEST_STATUS, FISHING_CONTEST_STATUS> nextStage = {
```
zach2good: This is just a ring, so just define as a vector and access with currentStage  = currentStage % vec.size() so it wraps around, instead of you having to define this order. And then have separate logic for bailing out early if closed.

src/map/utils/fishingcontest.cpp
```diff
@@ -0,0 +1,790 @@
+﻿/*
+===========================================================================
+
+Copyright (c) 2024 LandSandBoat Dev Teams
```
zach2good: This is not formatted like our other headers

src/map/packets/fish_ranking.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "fish_ranking.h"
+
+#include "common/socket.h"
+#include "common/vana_time.h"
+
+#include "utils/fishingcontest.h"
+
+#include <cstring>
+
+CFishRankingPacket::CFishRankingPacket(std::vector<FishingContestEntry>* entries, int8 language, int32 timestamp, int32 message_offset, uint32 totalEntries, uint8 msg_chunk)
```
zach2good: Don't pass in a pointer, take vec<Entries> by reference.

zach2good: message_offset -> messageOffset 
msg_chunk -> msgChunk

src/map/packet_system.cpp
```diff
@@ -2378,29 +2379,122 @@ void SmallPacket0x042(map_session_data_t* const PSession, CCharEntity* const PCh
...
     {
-        if ((bool)_sql->GetUIntData(0))
+        std::string login_message = luautils::GetServerMessage(msg_language);
+
+        PChar->pushPacket(new CServerMessagePacket(login_message, msg_language, msg_timestamp, msg_offset));
+        PChar->pushPacket(new CCharSyncPacket(PChar));
+
+        // TODO: kill player til theyre dead and bsod
+        const char* fmtQuery = "SELECT version_mismatch FROM accounts_sessions WHERE charid = %u";
+        int32       ret      = _sql->Query(fmtQuery, PChar->id);
```
zach2good: Use db:: system

src/map/utils/fishingcontest.cpp
paladindamarus: Odd... Can you point me to one that is correct?  I tried to copy one from something recent like the campaign_system.cpp (2023) file and just change the date to 2024.  I'm not sure which one should be the latest.

src/map/utils/fishingutils.cpp
paladindamarus: This is way out of my scope here.  I don't mind making this fix, but this touches the entire fishing system.  I've reset this back to the exact way it is on LSB now (will be updated in next push).  

In looking at this though ... I've realized that `GetFish` is overloaded in fishingutils.cpp:

`fish_t*                   GetFish(uint32 fishId);`
`CItemFish*       GetFish(uint16 itemid); // creates a 'new' CItemFish if possible`

WinterSolstice8: Doesn't leak if any CItem or derivative makes it into a container:
```
CItemContainer::~CItemContainer()
{
    for (uint8 SlotID = 0; SlotID <= m_size; ++SlotID)
    {
        destroy(m_ItemList[SlotID]);
    }
}
```

src/map/utils/fishingutils.cpp
```diff
@@ -2891,8 +2891,11 @@ namespace fishingutils
 
         if (PItem && FishList[itemid])
         {
-            // CItemFish constructor uses `const CItem&` input so this is ok
-            return new CItemFish(*PItem);
+            if (CItemUsable* PItemUsable = dynamic_cast<CItemUsable*>(PItem))
```
WinterSolstice8: What is the purpose of this cast?
 `class CItemEquipment : public CItemUsable` 
Neither of these classes are parents or children of CItemFish
What problem were you trying to solve here?

This is weird because I know (some) fish can be eaten by mithras or with some hidden effect items. Maybe CItemFish should be a child of CItemUsable..?

paladindamarus: Yeah I was trying something with casting and forgot to remove this piece.  I've taken this out and am resetting it back to what is on LSB.  I'm just not even using the GetFish function and instead checking the exdata directly.

WinterSolstice8: Oh ok, makes sense. I still wonder what's going on with usable fish then... hmm...

src/map/utils/fishingcontest.cpp
zach2good: Use regex and count.

```
^This program is free software: you can redistribute it and/or modify
^  This program is free software: you can redistribute it and/or modify
^    This program is free software: you can redistribute it and/or modify
```

src/map/utils/fishingutils.cpp
zach2good: I'll handle this is some ongoing work I have: https://github.com/LandSandBoat/server/pull/5185

src/map/utils/fishingcontest.cpp
zach2good: Clarified in this PR: https://github.com/LandSandBoat/server/pull/6172

src/map/utils/fishingutils.cpp
zach2good: I suppose it's out of scope for this PR, but here is the usage:
```cpp
    int32 CatchFish(CCharEntity* PChar, uint16 FishID, bool BigFish, uint16 length, uint16 weight, uint8 Count = 1)
    {
        uint16 MessageOffset = GetMessageOffset(PChar->getZone());
        PChar->animation     = ANIMATION_FISHING_CAUGHT;

        if (PChar->getStorage(LOC_INVENTORY)->GetFreeSlotsCount() != 0)
        {
            CItemFish* Fish = GetFish(FishID);

            if (Fish == nullptr)
            {
                ShowError("Invalid ItemID %i for fished item\n", FishID);
                PChar->animation = ANIMATION_FISHING_STOP;
                PChar->pushPacket(new CMessageTextPacket(PChar, MessageOffset + FISHMESSAGEOFFSET_LOST));
                return 0;
            }

            if (BigFish && length > 1 && weight > 1)
            {
                Fish->SetLength(length);
                Fish->SetWeight(weight);
            }

            Fish->setQuantity(Count);
            charutils::AddItem(PChar, LOC_INVENTORY, Fish);

            if (Count > 1)
            {
                PChar->loc.zone->PushPacket(PChar, CHAR_INRANGE_SELF, new CCaughtFishPacket(PChar, FishID, MessageOffset + FISHMESSAGEOFFSET_CATCH_MULTI, Count));
            }
            else
            {
                PChar->loc.zone->PushPacket(PChar, CHAR_INRANGE_SELF, new CCaughtFishPacket(PChar, FishID, MessageOffset + FISHMESSAGEOFFSET_CATCH, Count));
            }

            return 1;
        }
        else
        {
            PChar->loc.zone->PushPacket(PChar, CHAR_INRANGE_SELF, new CCaughtFishPacket(PChar, FishID, MessageOffset + FISHMESSAGEOFFSET_CATCH_INV_FULL, Count));
        }

        return 0;
    }
```

Quite right, if it ends up in a container it'll be cleaned up eventually, but there's a path here that contains a leak. I'll handle this in my other PR when I get around to it.

src/map/utils/zoneutils.h
```diff
@@ -43,17 +43,19 @@ namespace zoneutils
...
-    bool         IsResidentialArea(CCharEntity*);                // returns whether or not the area is a residential zone
-    bool         IsAlwaysOutOfNationControl(REGION_TYPE region); // returns true if a region should never trigger "in areas outside own nation's control" latent effect; false otherwise.
+    CZone*              GetZone(uint16 ZoneID);
+    CNpcEntity*         GetTrigger(uint16 TargID, uint16 ZoneID);
+    CBaseEntity*        GetEntity(uint32 ID, uint8 filter = -1);
+    CCharEntity*        GetCharByName(std::string const& name);
+    CCharEntity*        GetCharFromWorld(uint32 charid, uint16 targid);  // returns pointer to character by id and target id
+    CCharEntity*        GetChar(uint32 id);                              // returns pointer to character by id
+    CCharEntity*        GetCharToUpdate(uint32 primary, uint32 ternary); // returns pointer to preferred char to update for party changes
+    std::vector<uint16> GetZonesOnThisProcess();
```
zach2good: This can be auto and trailing return so you don't have to move all the other funcitons

tools/dbtool.py
```diff
@@ -190,6 +190,7 @@ def load_into_dict(filename, settings):
     "char_equip.sql",
     "char_equip_saved.sql",
     "char_exp.sql",
+    "char_fishing_contest_history",
```
zach2good: Is this likely to get very large over time, and need truncating?

paladindamarus: it shouldn't -- it's one row per character at most.  This is like `char_history`, but only holds the number of times each character has come in either first, second, third, or fourth-tenth place.  It is _not_ an ongoing log of individual contest results.  It will look something like this:

image

src/map/utils/fishingcontest.cpp
paladindamarus: Ah, got it!  Thanks!

src/map/lua/luautils.cpp
paladindamarus: Is that `global_objects` in lua.sh?

zach2good (approved):
Since fishing is off by default, and this has already gone through one round of reviews (by me at least), we may as well merge it before it rots (I have some heavy packet work that will _ruin_ this PR coming soon, so it's easier for everyone if it just goes in)

Xaver-DaRed (approved):

PR #5991, opened 2024-07-09
Add nullptr check for PLeader in Latent Effect Container (2)

CatsEyeXI:
**_I affirm:_**

Adds a nullptr check to PLeader in latent effect container.
image

If PLeader is in another map server instance the value can be null.

### DIFF
--- a/src/map/latent_effect_container.cpp
+++ b/src/map/latent_effect_container.cpp
@@ -788,9 +788,9 @@ bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect)
         {
             size_t partyCount = 0;
             size_t trustCount = 0;
-            if (m_POwner->PParty != nullptr)
+            auto* PLeader = m_POwner->PParty != nullptr ? dynamic_cast<CCharEntity*>(m_POwner->PParty->GetLeader()) : nullptr;
+            if (PLeader)
             {
-                auto PLeader = (CCharEntity*)m_POwner->PParty->GetLeader();
                 trustCount = PLeader->PTrusts.size();
                 partyCount = m_POwner->PParty->members.size();
             }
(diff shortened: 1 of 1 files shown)


zach2good (approved):

src/map/latent_effect_container.cpp
```diff
@@ -788,9 +788,9 @@ bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect)
         {
             size_t partyCount = 0;
             size_t trustCount = 0;
-            if (m_POwner->PParty != nullptr)
+            auto PLeader = (CCharEntity*)m_POwner->PParty->GetLeader();
+            if (PLeader != nullptr && m_POwner->PParty != nullptr)
```
WinterSolstice8: `m_POwner->PParty` must be checked for null before you can call a function on it

I don't particularly like this but

```
            if (m_POwner->PParty && dynamic_cast<CCharEntity*>(m_POwner->PParty->GetLeader()))
            {
                auto PLeader = static_cast<CCharEntity*>(m_POwner->PParty->GetLeader());
                trustCount = PLeader->PTrusts.size();
                partyCount = m_POwner->PParty->members.size();
            }
```

zach2good: You probably also want to use this setup in the block after this: `case LATENT::PARTY_MEMBERS_IN_ZONE:`

(We're just desperately patching things to do with the Party/Alliance system until someone becomes spiteful and motivated enough to nuke and rewrite it)

WinterSolstice8 (approved):

Xaver-DaRed (approved):

src/map/latent_effect_container.cpp
```diff
@@ -788,9 +788,9 @@ bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect)
         {
             size_t partyCount = 0;
             size_t trustCount = 0;
-            if (m_POwner->PParty != nullptr)
+            if (m_POwner->PParty && dynamic_cast<CCharEntity*>(m_POwner->PParty->GetLeader()))
```
zach2good: For both `case LATENT::PARTY_MEMBERS:` and `case LATENT::PARTY_MEMBERS_IN_ZONE:`; doing this `dynamic_cast` and throwing away the result to then do a `static_cast` to get that same result seems very off. Since you only really want to continue if there is a leader, extract out the leader first.

```cpp
            size_t partyCount = 0;
            size_t trustCount = 0;
            auto* PLeader = m_POwner->PParty != nullptr ?  dynamic_cast<CCharEntity*>(m_POwner->PParty->GetLeader()) : nullptr;
            if (PLeader) ...
 ....

PR #5992, opened 2024-07-09
Mamool Ja Family Improvements

jmcmorris:
**_I affirm:_**

This includes several changes related to the Mamool Ja family.
- Makes it possible to break their weapons with critical hits
- Improves weapon throwing (previously they'd play the weapon break animation at the same time as the weapon throw skill animation)
- Fixed when some skills could be used based on weapon state
- Assigned correct mob skill list to mage Mamool Ja (blue ones)
- Changed many default animation subs from 4 to 0
- Adjusted mob skills according to Jimmayus' document:
https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=57955395#gid=57955395

Travel to any location with some Mamool Ja. Their weapons can be broken from critical hits (10% chance). Use `!tp 3000` repeatedly to test their weapons skills and see that they use them appropriately based on the weapon state.


Xaver-DaRed:
I usually prefer to limit animation and subanimation changes to family mixings, instead of the actual mobskill scripts.

Because that helps with compability and with some very weird edge-cases where subanimation shouldnt change
(Not saying they shouldnt change in this cases, nor that this happens with this specific family)

Having said that, good stuff. Really.

scripts/actions/mobskills/forceful_blow.lua
```diff
@@ -10,14 +10,18 @@
...
     return 0
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
     local accmod = 1
-    local dmgmod = 3
-    local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT)
+    local dmgmod = 1
+    local info = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.physicalTpBonus.DMG_VARIES, 1, 1.5, 2)
```
Frankie-hz, line 24: Funny enough `xi.mobskills.physicalTpBonus.DMG_VARIES` doesnt actually do anything. It enters this:
```lua
    if tpEffect == xi.mobskills.physicalTpBonus.DMG_VARIES then
        hitdamage = hitdamage * MobTPMod(skill:getTP() / 10)
    end
```
and then even at 3000 tp the function `MobTPMod` just returns 1.
I want to say the "Damage varies with TP" is just the FTP values you provided at the end.

jmcmorris:
> I usually prefer to limit animation and subanimation changes to family mixings, instead of the actual mobskill scripts.
> 
> Because that helps with compability and with some very weird edge-cases where subanimation shouldnt change (Not saying they shouldnt change in this cases, nor that this happens with this specific family)
> 
> Having said that, good stuff. Really.

While playing around with it I discovered that changing the subanimation in the mobskill script is actually bad since it messes with animations. Like for this if you change it too early for the weapon throwing skills then the weapon disappears during the skill animation which is obviously wrong.

scripts/actions/mobskills/forceful_blow.lua
jmcmorris: I looked into some of these functions and there are many issues with them. I've started to wrap my head around how this is all suppose to work but there is still much I'm piecing together. Correcting this is a whole project in itself.

Xaver-DaRed:
You could use a listener in the family mixing for mobskill state exit.

I kind of remember this being an issue with mushroom mobs. I dont remember 100% if it was fixed, but I think it was. I would look there to see how those moskills are handled.

jmcmorris:
> You could use a listener in the family mixing for mobskill state exit.
> 
> I kind of remember this being an issue with mushroom mobs. I dont remember 100% if it was fixed, but I think it was. I would look there to see how those moskills are handled.

That's what I did
https://github.com/LandSandBoat/server/pull/5992/files#diff-1108cf4c7e319ba9fa79859eb0a2135431589ed18f783668c9656411b6f49de4

Xaver-DaRed (approved):

MowFord:
> You could use a listener in the family mixing for mobskill state exit.
> 
> I kind of remember this being an issue with mushroom mobs. I dont remember 100% if it was fixed, but I think it was. I would look there to see how those moskills are handled.

[it wasn't](https://github.com/LandSandBoat/server/pull/5789)

PR #5993, opened 2024-07-11
Wivre Family Improvements

jmcmorris:
**_I affirm:_**

This focuses on improving the Wivre family to be more retail accurate. This is primarily focused on Wivre skills. All changes here are either directly taken from the [Monster Stats sheet](https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=57955395#gid=57955395) or from retail captures I did myself.

The biggest change revolves around the skill Granite Skin which provides immunity to physical attacks while the attacker is directly in front of the target. With Larceny I was able to verify the status effect is Defense Boost. I opted to use the unused subpower on this effect to act as an angle from the front that determines if the attack is negated or not. This may be unnecessary and the angle could possibly be hardcoded instead with the subpower simply indicating that this effect has this special case.

The usual - go to a Wivre, `!gotoid 17137983` and spam `!tp 3000` to see the various mob skills.

### DIFF
--- a/scripts/actions/mobskills/boiling_blood.lua
+++ b/scripts/actions/mobskills/boiling_blood.lua
@@ -2,8 +2,6 @@
 -- Boiling Blood
 -- Description: Boiling Blood
 -- Foe gains Haste +25% and Berserk +50%
--- TODO: Verify ability duration
--- https://www.bg-wiki.com/ffxi/Locus_Wivre
 -----------------------------------
 local mobskillObject = {}
 
@@ -12,8 +10,8 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    xi.mobskills.mobBuffMove(mob, xi.effect.HASTE, 2500, 0, 60)
-    xi.mobskills.mobBuffMove(mob, xi.effect.BERSERK, 50, 0, 60)
+    xi.mobskills.mobBuffMove(mob, xi.effect.HASTE, 2500, 0, 180)
+    xi.mobskills.mobBuffMove(mob, xi.effect.BERSERK, 50, 0, 180)
     skill:setMsg(xi.msg.basic.NONE)
     return 0
 end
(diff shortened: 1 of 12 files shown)


scripts/actions/mobskills/batterhorn.lua
```diff
@@ -4,17 +4,19 @@
...
+    if target:isInfront(mob) then
+        return 0
+    end
+
+    return 1
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    local numhits = 3
+    local numhits = math.random(2, 3)
```
Frankie-hz: I don't think we should have any randoms inside numhits. This means the skill is multi-hit.
 The 2-3 shadows is from a miss

jmcmorris: I believe that and will happily change this but this is a thing in LOTS of mob skills

src/map/entities/battleentity.h
```diff
@@ -291,6 +291,7 @@ enum class REACTION : uint8
     HIT     = 0x08, // Hit
     EVADE   = 0x09, // Evasion (MISS + HIT)
     ABILITY = 0x10, // Observed on JA and WS
+    DEFLECT = 0x0A, // Attack hit but no damage dealt because of status effect such as Granite Skin
```
Xaver-DaRed: Wouldnt this (0x0A) come before  0x10?

scripts/actions/mobskills/batterhorn.lua
Frankie-hz: Yeah most are just copy/paste from the wiki. I would just fix the ones you run into per family if there are any

jmcmorris: Got it! 👍

src/map/entities/battleentity.h
jmcmorris: Yes, I don't know what I was thinking there. I completely removed this though since it is really just hit + parry.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

jmcmorris:
Made a last minute change where I just discovered that for back conal mobskills you need aoe set to `8` instead of `4` and Clobber wasn't set correctly

PR #5994, opened 2024-07-11
Steal All Status Effect Data

jmcmorris:
**_I affirm:_**

This change copies over all of the data associated with a stolen status effect. This came about directly from trying to test the Wivre PR I have up (#5993). Originally it was only copying the subpower but after further inspection I believe it makes sense to copy over all of this data. See discussion below.

Possibly the easiest way to test this is to apply this change on top of the Wivre PR I have up (#5993). With Granite Skin from that or another status that has a subpower, get the status effect on the mob and then use Larceny (THF99 has it). You should have the subpower of the stolen status effect.

### DIFF
--- a/scripts/globals/job_utils/thief.lua
+++ b/scripts/globals/job_utils/thief.lua
@@ -301,8 +301,12 @@ xi.job_utils.thief.useLarceny = function(player, target, ability, action)
         local newPower    = effectStolen:getPower()
         local newTick     = effectStolen:getTick()
         local newDuration = effectStolen:getDuration() + jpValue
+        local newSubType  = effectStolen:getSubType()
+        local newSubPower = effectStolen:getSubPower()
+        local newTier     = effectStolen:getTier()
+        local newFlags    = effectStolen:getEffectFlags()
 
-        player:addStatusEffectEx(newID, newIcon, newPower, newTick, newDuration)
+        player:addStatusEffectEx(newID, newIcon, newPower, newTick, newDuration, newSubType, newSubPower, newTier, newFlags)
         target:delStatusEffect(newID)
 
         effectID = newID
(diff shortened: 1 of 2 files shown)


scripts/globals/job_utils/thief.lua
```diff
@@ -299,12 +299,18 @@ xi.job_utils.thief.useLarceny = function(player, target, ability, action)
...
         local newDuration = effectStolen:getDuration() + jpValue
 
         player:addStatusEffectEx(newID, newIcon, newPower, newTick, newDuration)
         target:delStatusEffect(newID)
 
+        if newSubPower ~= 0 and player:hasStatusEffect(newID) then
+            local newEffect = player:getStatusEffect(newID)
+            newEffect:setSubPower(newSubPower)
+        end
+
```
Xaver-DaRed, line 311: Why not add the subpower directly to the `player:addStatusEffectEx()` above?
Like so
`player:addStatusEffectEx(newID, newIcon, newPower, newTick, newDuration, 0, newSubPower)`

jmcmorris: Great suggestion! I didn't realize that was an optional parameter.

jmcmorris:
Is there a reason we don't copy these other aspects of effects? After this change we are still ignoring the subID, tier and flags.

Xaver-DaRed:
We probably should be if we can, I would think.

Then again, if there's a reason we are not, that I don't know.

jmcmorris:
I dug into this some. This StealStatusEffect was introduced about 10 years ago so the motivation for not including these other aspects are lost to time unless the Village Elder (@TeoTwawki) knows.

From what I can tell, there are actually some issues with leaving this information out. Such as the subID is used to determine if a status effect is the same as another. The tier is used for determining if a status effect should overwrite another. I believe this means that if a high level haste is stolen it can be overwritten with a slow effect that is technically lower level than what should be possible. Flags are technically set with the flags defined in the db when the effect is added to a container so I'm not entirely sure that is necessary. There are a handful of places that we change the status effect flag with `AddEffectFlag` and `DelEffectFlag`. I don't see it hurting us copying this flag over.

jmcmorris:
I went ahead and made this change since I do believe that is what we ultimately want.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5995, opened 2024-07-12
[core] Camouflage Enhancements

9001-Sols:
**_I affirm:_**

Updates the logic for Camouflage to determine when it should or should not fall off, based on recent captures.

## Capture Summary

image

https://docs.google.com/spreadsheets/d/1EWNu3Z-qTJmVgYTmE4bHYavVPq7uGE6Vk30XHzl7Ebg

```
Crawlers were my initial target. They have a meleehit box of 3'6. When I tested back shots at 4'5 i had no breaks in camo.
When tried Gigas, I was losing camo on all shots at 4'5. Gigas had a melee hit box of 4'9. So I added 9, and tested at 5'8, and had no loss in camo. The minimum distance on a gigas I lost camo was 5'4, with many in the 4'9~5'3 range. Based on this, for back shots at least.. it might be that it's minimum melee hit box + '6 is the safe spot, with anything between melee and safe spot being having a chance to lose camo. It doesn't appear to be random in the sense that 1 out of 2 shots at 5'3 loses camo... more like one mob of the same size will let you pew-pew away at 5'3, but the next one you'll lose it - possibly mob-to-player level dependent.

Side shots also appear to have different distances dependent on mob size too. I just ended up adding .9 yalms to account for the size difference between the crawler and the gigas and I was able to maintain camo at 9.9 yalms. When testing minimum, the closest I lost it was 8.4 and farthest was 9.8.
```

**Speculated Formula by Capturer**

```
Back

Max melee distance + .6 = safe

Max melee distance + (.1~.5) = chance of deactivation

Under Max melee distance = certain deactivation
```

```
Side

Max melee distance + 5 yalms = safe 

(Max melee distance + 3.3) + (0.0~1.6) = chance of deactivation

Under Max melee distance + 3.3 = certain deactivation
```

```
Front

At melee distance + 8.1 yalms = safe 

(Max melee distance + 7.1) + (0.0~.99) = chance of deactivation

Under Max melee distance + 7.1 = certain deactivation
```

## What might be Missing

@Xaver-DaRed shared some helpful data below that implied mob levels might also have an impact on Camouflage retainability. The capturer also mentioned that mob level might play into some of the seemingly random cases where Camouflage fell off at specific distances. It might be prudent to go attempt some more tests with that in mind.

Additionally, the "chance of deactivation" percentage is currently estimated. I kept the existing estimation from work completed in ASB way before my time. Due to the nature of Camouflage being a 5 minute cooldown, it might take a long time to capture enough data if we think this is important. I would appreciate direction in how large of a sample size we'd need to be comfortable with a failure rate estimation.

~~I've started doing retail testing and can confirm barrage breaks the camouflage effect after fired. Once I gather enough data to be sure it is guaranteed and not a chance, I'll update my PR.~~

I've updated the PR to drop Camouflage if Barrage is active during the shot, to match retail behavior.

image

## PR Testing Video (done on ASB, can re-record for LSB if needed):
https://beta.ffxivclock.com/camouflage_testing.mp4

```
!changejob rng 75
!additem loxley_bow
!additem kabura_arrow 99
Equip the Bow and Arrow
Use Camouflage
Shoot from various angles and distances
Use !reset to reset the Camouflage cooldown when needed
```

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -2063,8 +2063,93 @@ void CCharEntity::OnRangedAttack(CRangeState& state, action_t& action)
     }
     battleutils::ClaimMob(PTarget, this);
     battleutils::RemoveAmmo(this, ammoConsumed);
-    // only remove detectables
-    StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_DETECTABLE);
+
+    // Handle Camouflage effects
+    if (this->StatusEffectContainer->HasStatusEffect(EFFECT_CAMOUFLAGE, 0))
+    {
+        int16 retainChance     = 40; // Estimate base ~40% chance to keep Camouflage on a ranged attack
+        uint8 rotAllowance     = 25; // Allow for some slight variance in direction faced to be "behind" or "beside" the mob
+        float distanceToTarget = distance(this->loc.p, PTarget->loc.p);
+        float meleeRange       = PTarget->GetMeleeRange();
+
+        if (isBarrage)
+        {
+            // Camouflage is never retained if Barrage is fired
+            retainChance = 0;
+        }
+        else if (behind(this->loc.p, PTarget->loc.p, rotAllowance))
+        {
+            // Max melee distance + .6 = safe
+            // Max melee distance + (.1~.5) = chance of deactivation
+            // Under max melee distance = certain deactivation
+            if (distanceToTarget > meleeRange + .6)
+            {
+                retainChance = 100;
+            }
+            else if (distanceToTarget > meleeRange + .1)
+            {
+                retainChance += 1.6 * distanceToTarget;
+            }
+            else
+            {
+                retainChance = 0;
+            }
+        }
+        else if (beside(this->loc.p, PTarget->loc.p, rotAllowance))
+        {
+            // Max melee distance + 5 yalms = safe
+            // (Max melee distance + 3.3) + (0.0~1.6) = chance of deactivation
+            // Under Max melee distance + 3.3 = certain deactivation
+            if (distanceToTarget > meleeRange + 5)
+            {
+                retainChance = 100;
+            }
+            else if (distanceToTarget > meleeRange + 3.3)
+            {
+                retainChance += 1.6 * distanceToTarget;
+            }
+            else
+            {
+                retainChance = 0;
+            }
+        }
+        else
+        {
+            // Max melee distance + 8.1 yalms = safe
+            // (Max melee distance + 7.1) + (0.0~.99) = chance of deactivation
+            // Under Max melee distance + 7.1 = certain deactivation
+            if (distanceToTarget > meleeRange + 8.1)
+            {
+                retainChance = 100;
+            }
+            else if (distanceToTarget > meleeRange + 7.1)
+            {
+                retainChance += 1.6 * distanceToTarget;
+            }
+            else
+            {
+                retainChance = 0;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
Ok, first thing, please, do not use ASB's contribution schema here. Use ours.

I have edited your PR comment. Please, click the checkboxes acepting LSBs conditions.

Xaver-DaRed:
How was camouflage tested in retail, if I may ask? Against what kind of mobs and at what level ranges + relative positions?

Just in case this helps, heres some data against VT and DC mobs
imagen

9001-Sols:
> How was camouflage tested in retail, if I may ask? Against what kind of mobs and at what level ranges + relative positions?
> 
> Just in case this helps, heres some data against VT and DC mobs imagen

I pulled data from the #capture-requests channel in ASB Discord

Conversation starts here and continues for a while:
https://discord.com/channels/966004751244329011/979421308678987776/1243308038186602578

claywar:
Ok, so.  Let's start with some groundwork here.  We're upstream of ASB, and some of us are not privy to their communications, captures, or information shared.

Now that that has been said, there has been no issue raised in this tracker to address these things.  This is our trigger to start investigating.  If we aren't aware there is a problem, then why would we try to solve it?

I'll be the first person to tell you that I will not hunt down a discord chat to verify data.  Provide your caps, show your proof and variances, and then things will be evaluated.

9001-Sols:
> Ok, so. Let's start with some groundwork here. We're upstream of ASB, and some of us are not privy to their communications, captures, or information shared.

Sorry, I thought the communities were a bit more tight-knit. Their discord is public as far as I know; but I've noted it for the future.

> Now that that has been said, there has been no issue raised in this tracker to address these things. This is our trigger to start investigating. If we aren't aware there is a problem, then why would we try to solve it?

I wasn't aware an issue needed to be raised and discussed and wasn't asking y'all to solve it. I was just doing my best to help out and implement a completely missing feature in LSB that already has partial implementation in ASB. When I submitted the improvements to ASB, I was asked to migrate them upstream.

> I'll be the first person to tell you that I will not hunt down a discord chat to verify data. Provide your caps, show your proof and variances, and then things will be evaluated.

Sure, I'll try to summarize all the capture information and discussions and provide a write-up for this. I didn't do the original captures and was operating off of conversations and summarized data provided by others and just wanted to help out. I've flipped this back to draft.

9001-Sols:
I've updated the Summary to provide a summary of the capture data I had access to. I think we might be missing data for different Mob Level <> Player Level deltas, so that might be something we need to look into if we want this to go in perfectly.

src/map/entities/charentity.cpp
```diff
@@ -2063,8 +2063,88 @@ void CCharEntity::OnRangedAttack(CRangeState& state, action_t& action)
...
+            else if (distanceToTarget > meleeRange + 7.1)
+            {
+                retainChance += 1.6 * distanceToTarget;
+            }
+            else
+            {
+                retainChance = 0;
+            }
+        }
+
```
9001-Sols, line 2134: I'm still stabilizing my LSB install (juggling it with ASB/Horizon), but as soon as it is stable I will implement behavior to make Camouflage drop off if a barrage auto occurs.

Once that is coded I'll redo my testing.

9001-Sols: This has been updated above.

Testing:
image

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #5997, opened 2024-07-12
Z's Trident Latent

9001-Sols:
**_I affirm:_**
## Please enter a player-facing description

Latent effect was added to `Z's Trident`

## Source:

https://wiki-ffo-jp.translate.goog/html/3911.html?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en

image

## What does this pull request do? (Please be technical)

This adds support for a `during WS` latent trigger. The existing one also applied when TP was less than 1000, which isn't what we wanted for `Z's Trident`.

Additionally, it adds the `STR+12` latent to Z's Trident to be active during Weapon Skills
image

You can add a `printf` in the `weaponskills.lua` to print player STR during WS, here is a screenshot of me doing that:

image

## Special Deployment Considerations

* No new considerations for fresh installs
* The script will need to likely be run manually for pre-existing installs, unless there is a migration process or something I can update

### DIFF
--- a/sql/item_latents.sql
+++ b/sql/item_latents.sql
@@ -2042,6 +2042,9 @@ INSERT INTO `item_latents` VALUES (18099,23,24,6,1000);  -- Attack+24 while TP <
 INSERT INTO `item_latents` VALUES (18099,25,5,6,1000);   -- Accuracy+5 while TP <100%
 INSERT INTO `item_latents` VALUES (18099,287,5,6,1000);  -- DMG+5 while TP <100%
 
+-- Z's Trident
+INSERT INTO `item_latents` VALUES (18101,8,12,41,0);     -- STR+12 during WS
+
 -- Leviathan's Couse
 -- TODO: INSERT INTO `item_latents` VALUES (18109,431,1,21,12); -- Additional effect: Water damage while you or a party member has Leviathan summoned
 
--- a/src/map/latent_effect_container.cpp
+++ b/src/map/latent_effect_container.cpp
@@ -168,6 +168,25 @@ void CLatentEffectContainer::CheckLatentsTP()
 
 /************************************************************************
 *                                                                       *
+* Checks all latents that are occur during WS and activates them        *
+*                                                                       *
+ ************************************************************************/
+void CLatentEffectContainer::CheckLatentsWS(bool isDuringWs)
+{
+    ProcessLatentEffects([this, isDuringWs](CLatentEffect& latentEffect) {
+        switch (latentEffect.GetConditionsID())
+        {
+            case LATENT::DURING_WS:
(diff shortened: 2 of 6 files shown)


src/map/latent_effect_container.cpp
```diff
@@ -686,7 +705,7 @@ void CLatentEffectContainer::ProcessLatentEffects(const std::function<bool(CLate
 
 // Processes a single CLatentEffect* and finds the expression to evaluate for
 // activation/deactivation and attempts to apply
-bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect)
+bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& , bool isDuringWs)
```
WinterSolstice8:
```suggestion
bool CLatentEffectContainer::ProcessLatentEffect(CLatentEffect& latentEffect, bool isDuringWs)
```

src/map/latent_effect_container.h
```diff
@@ -78,6 +79,7 @@ class CLatentEffectContainer
     std::vector<CLatentEffect> m_LatentEffectList;
 
     void ProcessLatentEffects(const std::function<bool(CLatentEffect&)>& logic);
+    bool ProcessLatentEffect(CLatentEffect& latentEffect, bool isDuringWs = false);
     bool ProcessLatentEffect(CLatentEffect& latentEffect);
```
WinterSolstice8: Looks like you forgot to delete this second line

9001-Sols: Thanks! Was midway through porting this PR over from ASB and got pulled into Dynamis 😆

9001-Sols:
I've re-tested this with a fresh LSB install after porting it from ASB and confirmed it is still working as intended:

First WS was with Gungnir
Second WS was with Z's Trident

image

image

sql/item_latents.sql
```diff
@@ -2058,6 +2058,9 @@ INSERT INTO `item_latents` VALUES (18133,26,5,53,1);     -- RACC +5 in areas out
 INSERT INTO `item_latents` VALUES (18134,8,2,53,1);      -- STR +2 in areas outside own nation's control
 INSERT INTO `item_latents` VALUES (18135,8,3,53,1);      -- STR +3 in areas outside own nation's control
 
+-- Z's Trident
+INSERT INTO `item_latents` VALUES (18101,8,12,41,0);     -- STR+12 during WS
+
```
cocosolos: This is out of order, can you move it up above Leviathan's Couse?

cocosolos (changes requested):

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #6000, opened 2024-07-12
[cpp] Adds standback range mod for mobs

Frankie-hz:
**_I affirm:_**

Our standback range is hardcoded to 20 which is inaccurate. We need to gather more retail research on this but a lot of mobs have different standback ranges. This PR adds standback range mod for mobs.  This will allow us to apply the range needed for proper standback.

Example of Dynamis capture that needs this: https://www.youtube.com/watch?v=JriGL3j062c

You can use Vanguard Hitman as an example or any ranger/ninja mob you would like.
Apply the mod: `mob:setMobMod(xi.mobMod.STANDBACK_RANGE, 16)`
See the mob standback at 16
[Imgur](https://imgur.com/7je7YaN) (dont ask I am bringing it back a few years bc the gif was too large)


Xaver-DaRed (approved):
lgtm

PR #6005, opened 2024-07-13
[core] Adds damage cap mod

Frankie-hz:
**_I affirm:_**

Adds damage cap mod for mobs that have set values for damage they take.

There are several pieces of content that have "damage caps" in the game. Some examples are: Crustacean Conundrum bcnm, Suttung (dyna-qufim), Antaeus (dyna-qufim) and colonization reives.

Unfortunately we do not know or understand EXACTLY how these damage caps works and the amount of research required to fully understand the system is out of reach for a long long time. Nonetheless here is my proposal for the damage mod cap.

Lets use Dynamis-Qufim Suttung as an example. From Siknoz retail capture it has been viewed that the mob takes 99 capped damage but anything over 99 seems to have a "random" factor between a certain variant (observed to be roughly 10).  We do not know or understand the exact mechanic behind this as it can have 100 different factors.

### The proposal:
Implement two mods. `RECEIVED_DAMAGE_CAP` and `RECEIVED_DAMAGE_VARIANT`
By having these two mods we can closely replicate the observed damage seen. By setting the DAMAGE_CAP to 99 and the DAMAGE_VARIANT to 9 the damage observed will be between 90 and 99 is the damage is above 99.

For cases like Crustacean Conundrum bcnm where the damage cap appears to be 2 or for Reives if you don't meet the requirements then the damage is 0 or 1 this will also work. By setting the DAMAGE_CAP to 1 and the DAMAGE_VARIANT to 1 the output will generate either a 0 or a 1.

Things I tested:
damage spells, abilities (jump, chi blast, EES, etc.), bluemagic, bio/dia, summons, pets, physical damage, ranged damage, magic damage, breathe damage

Go to any mob you want.
!setmod RECEIVED_DAMAGE_CAP 100
!setmod RECEIVED_DAMAGE_VARIANT 10
Hit the mob for over 100 damage and see the damage random between 90-100

!setmod RECEIVED_DAMAGE_CAP 1
!setmod RECEIVED_DAMAGE_VARIANT 1
Hit the mob for over 1 damage and see the damage random between 0-1


src/map/utils/battleutils.cpp
```diff
@@ -5023,6 +5032,37 @@ namespace battleutils
         PChar->PClaimedMob = nullptr;
     }
 
+    // Checks to see if the mob has a damage cap value
+    // This is used for instances like Suttung, Antaeus, Crustacean Conundrum bcnm, Colonization reives
+    int32 DamageCapDistribution(int32 damage, CBattleEntity* PDefender)
```
Frankie-hz: I should think of a better name for this function. Not sure what I was thinking here. `CheckDamageCap` maybe?

zach2good: I like verboseness: `CheckAndApplyDamageCap`

scripts/globals/bluemagic.lua
```diff
@@ -349,6 +349,9 @@ xi.spells.blue.useDrainSpell = function(caster, target, spell, params, softCap,
     dmg = addBonuses(caster, spell, target, dmg)
     dmg = adjustForTarget(target, dmg, spell:getElement())
 
+    -- Check if the mob has a damage cap
+    target:checkDamageCap(dmg)
```
zach2good: This does nothing?

src/map/utils/battleutils.cpp
```diff
@@ -5023,6 +5032,37 @@ namespace battleutils
...
+    {
+        int32 damageCap     = PDefender->getMod(Mod::RECEIVED_DAMAGE_CAP);     // The max damage cap
+        int32 damageVariant = PDefender->getMod(Mod::RECEIVED_DAMAGE_VARIANT); // The value you want the damage to have a variance by
+
+        // If the target has no mod or the damage is less than the cap return normal damage
+        if ((damageCap == 0) || damage < damageCap)
```
zach2good: No need for those inner parens

src/map/utils/battleutils.cpp
```diff
@@ -5023,6 +5032,37 @@ namespace battleutils
...
+        {
+            return damage;
+        }
+        else
```
zach2good: No need for this else, we'll continue if we don't return anyway

src/map/utils/battleutils.cpp
```diff
@@ -5023,6 +5032,37 @@ namespace battleutils
...
+        {
+            // Clamp the damage since the damage is higher than the cap
+            damage = std::clamp(damage, 0, damageCap);
+        }
+
+        // If for whatever reason your damage variant is set too high set the variant to 0 as a fail safe
+        if (damageVariant > damageCap)
```
zach2good, line 5048: Do you think this would need some warning logging, as this would be user setup error?

src/map/utils/battleutils.cpp
```diff
@@ -5023,6 +5032,37 @@ namespace battleutils
...
+        {
+            damageVariant = 0;
+        }
+
+        // see https://bugs.llvm.org/show_bug.cgi?id=18767#c1 ; essentially, [min, max] range on this RNG call excludes the max
```
zach2good: `[min, max)` <- semi-open interval

`[min, max]` <- closed interval

zach2good (commented):
Code in `DamageCapDistribution` seems OK, I don't know much about combat so I'll leave the application of it to others

src/map/modifier.h
```diff
@@ -1020,7 +1023,7 @@ enum class Mod
     // 221 to 222
     // 274 to 275
     //
-    // SPARE = 1083 and onward
+    // SPARE = 1084 and onward
```
Xaver-DaRed: I know this is a bitch to ask, but can we use IDs 221 and 222 and be done with that blank range?

src/map/utils/battleutils.cpp
Frankie-hz: Changed

scripts/globals/bluemagic.lua
Frankie-hz: I accidently deleted the dmg = when I changed the name over and move it below adjustments. Fixed.

Edit: turns out that is not what I did and this isn't needed. It hits the applySpellDamage below it and runs the cap then which is why it worked.

src/map/utils/battleutils.cpp
Frankie-hz: I added a warning to let them know the variant is too high

src/map/modifier.h
Frankie-hz: Sure!

src/map/utils/battleutils.cpp
```diff
@@ -5023,6 +5032,35 @@ namespace battleutils
...
+        // If for whatever reason your damage variant is set too high set the variant to 0 as a fail safe
+        if (damageVariant > damageCap)
+        {
+            ShowWarning("battleutils::CheckAndApplyDamageCap - RECEIVED_DAMAGE_VARIANT is > than RECEIVED_DAMAGE_CAP");
+            damageVariant = 0;
+        }
+
+        // see https://bugs.llvm.org/show_bug.cgi?id=18767#c1 ; essentially, [min, max) range on this RNG call excludes the max
+        // so we must add +1 to our max to achieve the range we want
+        damage -= xirand::GetRandomNumber(0, damageVariant + 1);
```
zach2good: Even though it's implied through the types, I'd probably specify the template argument here too: `GetRandomNumber<int32>`

src/map/utils/battleutils.cpp
```diff
@@ -2400,6 +2403,9 @@ namespace battleutils
             damage = 0;
         }
 
+        // Handle damage cap mod
```
zach2good: I reckon you don't need the comments everywhere saying this, now that the function is called `CheckAndApplyDamageCap`

scripts/globals/bluemagic.lua
```diff
@@ -437,7 +437,12 @@ xi.spells.blue.applySpellDamage = function(caster, target, spell, dmg, params, t
             return dmg
         end
 
+        -- Check if the mob has a damage cap
+        dmg = target:checkDamageCap(dmg)
```
zach2good: Does this _have_ to be before oneforall? Otherwise you could combine the two cap calls and move just outside of the next end block

Xaver-DaRed: Realistically, the target wont ever have both one for all AND a damage cap.

However, I think damage cap should come/be applied AFTER all damage calculations, including phalanx, one for all AND stoneskin

Frankie-hz: I can move it to after, I wasn't exactly sure *where* to put it but thinking about it again I think that makes the most sense.

WinterSolstice8: I don't think it matters, but afaik is untestable. The difference between before and after is one makes stoneskin and magic stoneskin godlike and the other makes it trivial 

Unless contagion transfer can move buffs into mobs that is...

Frankie-hz: As far as I know the handful of things that actually have a damage cap doesn't have stoneskin so we can't really test it

WinterSolstice8: We may want to defer to "what's worse for the player"

which would be damage cap first, because that means even a weak stoneskin could be impenetrable without dispel, or require different strats to eat through (such as DW + enspells) if they are dispel immune

Frankie-hz: After talking to winter about this I think moving this *BEFORE* stoneskin is the correct call. After looking at the crab bcnm that can potentially use stoneskin (I have asked for a retail capture but this may take a while as its not confirmed it can even cast it) it would get knocked off in one hit if this check was after.

zach2good (approved):
Seems fine on my end 👍

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #6007, opened 2024-07-14
[core, lua] Improve draw-in specification to fix some issues

TracentEden2:
**_I affirm:_**

This PR changes the specification of draw-in functions in several places to support flexibility and the additional options needed for the draw-in logic of different mobs. 

Current LSB draw-in problems:
- The `DRAW_IN` mob mod is currently representing two different options that do not always occur together. Specifically, a `DRAW_IN > 1` indicates alliance draw-in and a `DRAW_IN > 1` also sets the draw-in trigger distance to the mob mod. Thus all mobs with a non-standard trigger distance are also forced to have alliance draw-in. This creates several draw-in bugs.
- No ability to draw-in to the front of the mob (as opposed to mob center)
- No ability to set a max draw-in distance (beyond which a player or alliance member will not be draw-in)
- The lua draw-in function (used for ad-hoc draw-ins in special case mobs) does not allow any options and does not allow overriding the mob or hard coded draw-in defaults.

Changes by this PR:
- Changes the `DRAW_IN` mob mod to be a bitmask mod that represents several different draw-in options including perform normal draw-in logic, include alliance members in draw-in, and draw-in to front of mob. Also adds mob mods MOBMOD_DRAW_IN_TRIGGER_DIST and MOBMOD_DRAW_IN_MAX_RANGE to allow specifying the draw-in trigger distance and the max draw-in range. The bitmask approach allows flexibility and easy future expansion to new draw-in types that LSB is still missing.
- Changes the lua draw-in function to allow options as parameters. The order of precedence is thus: specified function option -> mob mod(s) -> hardcoded default option
- Changes mobs with draw-in to use the new and updated draw-in mob mods

Fight mobs with draw-in and vary the draw-in mob mods to set different options, also test the lua drawIn function call with different parameters

### DIFF
--- a/scripts/enum/mob_mod.lua
+++ b/scripts/enum/mob_mod.lua
@@ -20,7 +20,7 @@ xi.mobMod =
     HP_HEAL_CHANCE         = 9,  -- can cast cures below this HP %
     SUBLINK                = 10, -- sub link group
     LINK_RADIUS            = 11, -- link radius
-    DRAW_IN                = 12, -- 1 - player draw in, 2 - alliance draw in -- only add as a spawn mod!
+    DRAW_IN_BITMASK        = 12, -- bitmask with different binary options for draw-in (see DRAWIN enum for options)
     SEVERE_SPELL_CHANCE    = 13, -- % chance to use a severe spell like death or impact
     SKILL_LIST             = 14, -- uses given mob skill list
     MUG_GIL                = 15, -- amount gil carried for mugging
@@ -90,4 +90,6 @@ xi.mobMod =
     CANNOT_GUARD           = 79, -- Check if the mob does not guard (despite being a MNK or PUP mob)
     SKIP_ALLEGIANCE_CHECK  = 80, -- Skip the allegiance check for valid target (allows for example a mob to cast a TARGET_ENEMY spell on itself)
     ABILITY_RESPONSE       = 81, -- Mob can respond to player ability use with onPlayerAbilityUse()
+    DRAW_IN_TRIGGER_DIST   = 82, -- Distance to trigger a draw-in (overrides a default of 2 * melee dist)
+    DRAW_IN_MAX_RANGE      = 83, -- Max range to draw-in an entity (overrides a default of entire zone)
 }
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -708,11 +708,11 @@ void CMobController::Move()
         if (((currentDistance > closeDistance) || move) && PMob->PAI->CanFollowPath())
         {
             // TODO: can this be moved to scripts entirely?
-            if (PMob->getMobMod(MOBMOD_DRAW_IN) > 0)
+            if ((PMob->getMobMod(MOBMOD_DRAW_IN_BITMASK) & DRAWIN::DRAWIN_NORMAL))
             {
-                if (currentDistance >= PMob->GetMeleeRange() * 2 && battleutils::DrawIn(PTarget, PMob, PMob->GetMeleeRange() - 0.2f))
+                bool didDrawin = HandleDrawin(currentDistance);
+                if (didDrawin)
                 {
-                    FaceTarget();
                     return;
                 }
             }
@@ -1414,4 +1414,27 @@ bool CMobController::IsSpellReady(float currentDistance)
     }
 
     return m_Tick >= m_nextMagicTime;
+}
+
+bool CMobController::HandleDrawin(float currentDistance)
+{
+    // get specific draw-in settings that can be set on any given mob otherwise use default values
+    const uint32 drawInTriggerDistance = PMob->getMobMod(MOBMOD_DRAW_IN_TRIGGER_DIST) ? PMob->getMobMod(MOBMOD_DRAW_IN_TRIGGER_DIST) : PMob->GetMeleeRange() * 2;
+    const uint32 drawInMaxDistance     = PMob->getMobMod(MOBMOD_DRAW_IN_MAX_RANGE) ? PMob->getMobMod(MOBMOD_DRAW_IN_MAX_RANGE) : std::numeric_limits<int16>::max();
+    const bool   includeAlliance       = (PMob->getMobMod(MOBMOD_DRAW_IN_BITMASK) & DRAWIN::DRAWIN_INCLUDE_ALLIANCE);
+    const bool   toFrontOfMob          = (PMob->getMobMod(MOBMOD_DRAW_IN_BITMASK) & DRAWIN::DRAWIN_TO_FRONT_OF_MOB);
+
+    bool didDrawin = false;
+
+    if (currentDistance >= drawInTriggerDistance)
+    {
+        bool didDrawin = battleutils::DrawIn(PTarget, PMob, PMob->GetMeleeRange() - 0.2f, drawInMaxDistance, includeAlliance, toFrontOfMob);
+
+        if (didDrawin)
+        {
+            FaceTarget();
+        }
+    }
+
+    return didDrawin;
 }
--- a/src/map/entities/mobentity.h
+++ b/src/map/entities/mobentity.h
@@ -102,6 +102,14 @@ enum BEHAVIOUR : uint16
     BEHAVIOUR_NO_TURN      = 0x400  // mob does not turn to face target
 };
 
+enum DRAWIN : uint16
+{
+    DRAWIN_NONE             = 0x000, // no draw-in (mob can only draw-in by explictly calling the lua drawin function)
+    DRAWIN_NORMAL           = 0x001, // basic draw-in where mob will draw-in battle target when target some distance from mob
+    DRAWIN_TO_FRONT_OF_MOB  = 0x002, // determines if draw-in to the center of mob (bit not set) or front of mob (bit set)
+    DRAWIN_INCLUDE_ALLIANCE = 0x004, // determines if draw-in includes just the battle target (bit not set) or entire alliance (bit set)
+};
+
 class CMobSkillState;
 
 class CMobEntity : public CBattleEntity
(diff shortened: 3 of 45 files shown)


Xaver-DaRed:
Why not  change draw in to mean 1 -> target player + pets, and 2 for party/alliance, (so no numbers over 2) and add another mobmod for draw in trigger distance, without the need of having 4 local variables?

And actually, the original draw in could be changed to set draw in type (player or party / to in front of mob OR to specified locations, like Khaimaira)

TracentEden2:
> Why not change draw in to mean 1 -> target player + pets, and 2 for party/alliance, (so no numbers over 2) and add another mobmod for draw in trigger distance, without the need of having 4 local variables?
> 
> And actually, the original draw in could be changed to set draw in type (player or party / to in front of mob OR to specified locations, like Khaimaira)

The idea is there are a lot of potential differences between different draw-ins so would need to have several more mob mods as the dimensions are at least: alliance, in front, trigger dist, max dist, type (leash vs. no leash), include dead/mounted, etc. For brevity not all of these are included in this PR, leash and others will come later on. Also the local variables make understanding mob draw-in logic easier to understand from just the mob file (without needing to look through other logic).

Xaver-DaRed:
Since you are already reducing the original mobmod to a bit, one could use the original mobmod as a bitmask.
original mobmod (type)
bit 0 -> has draw in
bit 1 -> include alliance (vs target only)
bit 2 -> specified pos (vs in front)
bit 3 -> include dead
bit 4 -> include mounted
bit 5 -> leash (i don even know what this is)
..and so on

New mobmod
draw in trigger distance

This are just ideas. I'm personally not a fan of abusing the use of local variables for global systems

TracentEden2:
> Since you are already reducing the original mobmod to a bit, one could use the original mobmod as a bitmask. original mobmod (type) bit 0 -> has draw in bit 1 -> include alliance (vs target only) bit 2 -> specified pos (vs in front) bit 3 -> include dead bit 4 -> include mounted bit 5 -> leash (i don even know what this is) ..and so on
> 
> New mobmod draw in trigger distance
> 
> This are just ideas. I'm personally not a fan of abusing the use of local variables for global systems

Yeah, could maybe do it with a bitmask and only three new mob mods: trigger distance, max distance, and leash distance, however we lose the readability and explicitness and have to rely on potential code comments. In the end I do not mind either method, but I guess depends on what should be prioritized.

Leash draw-in is where the mob only allows the players to move a specified distance away from a fixed center point of a shape before drawing in potentially to a center point or other point in that shape, thus the mobs never leaves that shape (like dog on leash), this is like Aspid for example (note aspid also has normal non-leash draw-in it can use inside the circle)

TracentEden2:
> Since you are already reducing the original mobmod to a bit, one could use the original mobmod as a bitmask. original mobmod (type) bit 0 -> has draw in bit 1 -> include alliance (vs target only) bit 2 -> specified pos (vs in front) bit 3 -> include dead bit 4 -> include mounted bit 5 -> leash (i don even know what this is) ..and so on
> 
> New mobmod draw in trigger distance
> 
> This are just ideas. I'm personally not a fan of abusing the use of local variables for global systems

Ok, I updated the PR to use a bitmask mob mod for DRAW_IN and two new mob mods for setting the trigger distance and max range :)

WinterSolstice8:
CI has a real complaint saying PTarget is not used in a lambda

TracentEden:
> CI has a real complaint saying PTarget is not used in a lambda

Thanks for pointing this out, I have updated the PR now!

scripts/enum/mob_mod.lua
```diff
@@ -19,7 +19,7 @@ xi.mobMod =
     HP_HEAL_CHANCE         = 9,  -- can cast cures below this HP %
     SUBLINK                = 10, -- sub link group
     LINK_RADIUS            = 11, -- link radius
-    DRAW_IN                = 12, -- 1 - player draw in, 2 - alliance draw in -- only add as a spawn mod!
+    DRAW_IN                = 12, -- bitmask with different binary options for draw-in (see DRAWIN enum for options)
```
zach2good: To make this _even more obvious_ (since this is a departure from the existing logic) I would probably name this `DRAW_IN_BITMASK` or `DRAW_IN_MASK` or `DRAW_IN_SETTINGS` etc.

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -710,11 +710,11 @@ void CMobController::Move()
         if (((currentDistance > closeDistance) || move) && PMob->PAI->CanFollowPath())
         {
             // TODO: can this be moved to scripts entirely?
-            if (PMob->getMobMod(MOBMOD_DRAW_IN) > 0)
+            if ((PMob->getMobMod(MOBMOD_DRAW_IN) & DRAWIN::DRAWIN_NORMAL) > 0)
```
zach2good: The `>0` is redundant here

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -1412,3 +1412,26 @@ bool CMobController::IsSpellReady(float currentDistance)
...
     return (m_Tick >= m_LastMagicTime + std::chrono::milliseconds(PMob->getBigMobMod(MOBMOD_MAGIC_COOL) - bonusTime));
 }
+
+bool CMobController::HandleDrawin(float currentDistance)
+{
+    // get specific draw-in settings that can be set on any given mob otherwise use default values
+    uint32 drawInTriggerDistance = PMob->getMobMod(MOBMOD_DRAW_IN_TRIGGER_DIST) > 0 ? PMob->getMobMod(MOBMOD_DRAW_IN_TRIGGER_DIST) : PMob->GetMeleeRange() * 2;
+    uint32 drawInMaxDistance     = PMob->getMobMod(MOBMOD_DRAW_IN_MAX_RANGE) > 0 ? PMob->getMobMod(MOBMOD_DRAW_IN_MAX_RANGE) : std::numeric_limits<int16>::max();
+    bool   includeAlliance       = (PMob->getMobMod(MOBMOD_DRAW_IN) & DRAWIN::DRAWIN_INCLUDE_ALLIANCE) > 0 ? true : false;
+    bool   toFrontOfMob          = (PMob->getMobMod(MOBMOD_DRAW_IN) & DRAWIN::DRAWIN_TO_FRONT_OF_MOB) > 0 ? true : false;
```
zach2good: Same here, you don't need to be doing all of these extra steps...

```cpp
const bool toFrontOfMob = PMob->getMobMod(MOBMOD_DRAW_IN) & DRAWIN::DRAWIN_TO_FRONT_OF_MOB;
```

zach2good: Also, if you're not going to change the value of the variable later, make it const.

src/map/entities/mobentity.h
```diff
@@ -102,6 +102,14 @@ enum BEHAVIOUR : uint16
     BEHAVIOUR_NO_TURN      = 0x400  // mob does not turn to face target
 };
 
+enum DRAWIN : uint16
+{
+    DRAWIN_NONE             = 0x000, // no draw-in (mob can only draw-in by explictly calling the lua drawin function)
+    DRAWIN_NORMAL           = 0x001, // basic draw-in where mob will draw-in battle target when target some distance from mob
+    DRAWIN_TO_FRONT_OF_MOB  = 0x002, // determines if draw-in to the center of mob (bit not set) or front of mob (bit set)
+    DRAWIN_INCLUDE_ALLIANCE = 0x004  // determines if draw-in includes just the battle target (bit not set) or entire alliance (bit set)
```
zach2good: Needs trailing comma (if that's relevant in C++, I don't remember?)

src/map/utils/battleutils.cpp
```diff
@@ -5820,10 +5821,18 @@ namespace battleutils
                 }
                 else
                 {
-                    // draw in!
-                    PMember->loc.p.x = nearEntity.x;
-                    PMember->loc.p.y = nearEntity.y;
-                    PMember->loc.p.z = nearEntity.z;
+                    if (toFrontOfMob)
+                    {
+                        PMember->loc.p.x = nearEntity.x;
```
zach2good, line 5637: Would it make sense to calculate these positions at the very start of the draw in, validate them on the navmesh, and then assign them here?

TracentEden: > Would it make sense to calculate these positions at the very start of the draw in, validate them on the navmesh, and then assign them here?

I think the function is already doing what you describe? It calculates and validates near the start of the draw-in function. Or do you mean even earlier in the draw-in process like in a different function?

scripts/enum/mob_mod.lua
TracentEden2: > To make this _even more obvious_ (since this is a departure from the existing logic) I would probably name this `DRAW_IN_BITMASK` or `DRAW_IN_MASK` or `DRAW_IN_SETTINGS` etc.

Yep, I have updated to DRAW_IN_BITMASK, thanks for comment.

src/map/ai/controllers/mob_controller.cpp
TracentEden2: > The `>0` is redundant here

Yep, I have fixed this, thanks for comment.

src/map/ai/controllers/mob_controller.cpp
TracentEden2: > Also, if you're not going to change the value of the variable later, make it const.

Yep, I have fixed both of these, thanks for comment.

src/map/entities/mobentity.h
TracentEden2: > Needs trailing comma (if that's relevant in C++, I don't remember?)

Yep, I have fixed this, thanks for comment.

zach2good:
I don't think this PR is relevant anymore, since draw-in got moved into Lua? https://github.com/LandSandBoat/server/pull/6566

TracentEden2:
> I don't think this PR is relevant anymore, since draw-in got moved into Lua? #6566

Yep, I think better to add any remaining fixes from this PR in separate Lua PR.

PR #6009, opened 2024-07-15
Implement missing Poroggo mobskills

jmcmorris:
**_I affirm:_**

This goes through and implements the missing Poroggo mobskills. They are fortunately fairly straightforward with the mechanics and just needed to actually be implemented.

https://ffxiclopedia.fandom.com/wiki/Category:Poroggos

Go fight a Poroggo such as in Mamook and use `!tp 3000` to test out the various mobskills.

### DIFF
--- a/scripts/mixins/families/poroggo.lua
+++ b/scripts/mixins/families/poroggo.lua
@@ -1,0 +1,17 @@
+require('scripts/globals/mixins')
+
+g_mixins = g_mixins or {}
+g_mixins.families = g_mixins.families or {}
+
+g_mixins.families.poroggo = function(poroggoMob)
+    -- Change Poroggo spell list back to standard BLM after casting a providence spell
+    poroggoMob:addListener('MAGIC_START', 'POROGGO_MAGIC_START', function(mob, spell, action)
+        if mob:getLocalVar('providence') == 1 then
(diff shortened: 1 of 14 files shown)


scripts/mixins/families/poroggo.lua
```diff
@@ -0,0 +1,17 @@
...
+g_mixins.families = g_mixins.families or {}
+
+g_mixins.families.poroggo = function(poroggoMob)
+    -- Change Poroggo spell list back to standard BLM after casting a providence spell
+    poroggoMob:addListener('MAGIC_START', 'POROGGO_MAGIC_START', function(mob, spell, action)
+        if mob:getLocalVar('providence') == 1 then
+            mob:setSpellList(2)
+            mob:setLocalVar('providence', 0)
+            mob:setMobMod(xi.mobMod.MAGIC_COOL, 35)
+        end
```
Xaver-DaRed, line 13: Do we have a `:getSpellList()` method?

If so, we could check spell list and be done with the local variable

EDIT: We dont. We have a `:hasSpellList()` method, but that wont work.

Xaver-DaRed (approved):

zach2good (approved):

PR #6012, opened 2024-07-15
Add nullsong mobskill and made Dahaks able to use it

jmcmorris:
**_I affirm:_**

This implements the missing mobskill, Nullsong, which is used by Dahaks. This implements a new function `countEffectWithFlag`. This is needed since Dahak will only use Nullsong if the there are 3 or more effects to dispel. This work is based off of the data collected by Jimmayus in the [Monster Stats Spreadsheet](https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=57955395#gid=57955395) as well as the [capture done by Melo](https://discord.com/channels/443544205206355968/443894311881408522/1262474594128891957).

Go fight a dahak and spam `!tp 3000`. It'll never use nullsong. Put some buffs up and then try again and it should eventually use nullsong. Dragons have a lot of mobskills so it can take a while.


scripts/actions/mobskills/nullsong.lua
```diff
@@ -0,0 +1,37 @@
...
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    if mob:hasStatusEffect(xi.effect.SILENCE) then
+        return 1
+    end
+
+    if target:countEffectWithFlag(xi.effectFlag.DISPELABLE) < 3 then
+        return 1
+    end
```
Xaver-DaRed, line 17:
```lua
    if 
        mob:hasStatusEffect(xi.effect.SILENCE) or
        target:countEffectWithFlag(xi.effectFlag.DISPELABLE) < 3
    then
        return 1
    end
```

scripts/actions/mobskills/nullsong.lua
```diff
@@ -0,0 +1,37 @@
...
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local count = target:dispelAllStatusEffect(xi.effectFlag.DISPELABLE)
+
+    if count == 0 then
+        skill:setMsg(xi.msg.basic.SKILL_NO_EFFECT)
+        return count
+    end
```
Frankie-hz, line 28: If the skill cant fire off without having at least 3 effects to dispel is this check ever going to even get hit?

Xaver-DaRed: Its an AoE. Other targets (non-main target) may not have any buffs

Frankie-hz: Ah you are right, missed the AOE part

Xaver-DaRed (approved):

zach2good (approved):

PR #6014, opened 2024-07-16
[Weaponskills] [Lua] Only add fixed enmity if Weaponskill lands

9001-Sols:
**_I affirm:_**

Based on Kanican's Blog, fixed enmity Weaponskills such as Coronach and Namas Arrow should not grant enmity on a miss.

Kanican Blog [Enmity Testing Part IX](https://kanican.livejournal.com/31788.html).

image

```
!tp 3000 [your name]
Use Namas Arrow or Coronach until it misses
Make sure the fixed enmity isn't added (easiest to test this with print statements in LUA, but !getenmity could also be used)
```

## Testing Results

| Before | After |
|-|-|
| image | image |

Proof that a missed WS after the changes still claims the mob as expected:

https://github.com/user-attachments/assets/86893a08-36a8-44e1-b4ca-5db6f1cfe1ec

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -1052,7 +1052,11 @@ xi.weaponskills.takeWeaponskillDamage = function(defender, attacker, wsParams, p
 
     local enmityEntity = wsResults.taChar or attacker
 
-    if wsParams.overrideCE and wsParams.overrideVE then
+    if
+        wsParams.overrideCE and
+        wsParams.overrideVE and
+        wsResults.tpHitsLanded + wsResults.extraHitsLanded > 0
+    then
         defender:addEnmity(enmityEntity, wsParams.overrideCE, wsParams.overrideVE)
     else
         local enmityMult = wsParams.enmityMult or 1


Xaver-DaRed:
question: Wll this still tag/make the mobs engage, even on a miss?

9001-Sols:
> question: Wll this still tag/make the mobs engage, even on a miss?

I'll test this right now, I believe yes since it still runs through the standard `updateEnmityFromDamage` routine but will report back.

Edit: Confirmed it claims as expected and added a video to the summary

scripts/globals/weaponskills.lua
```diff
@@ -1052,7 +1052,7 @@ xi.weaponskills.takeWeaponskillDamage = function(defender, attacker, wsParams, p
 
     local enmityEntity = wsResults.taChar or attacker
 
-    if wsParams.overrideCE and wsParams.overrideVE then
+    if wsParams.overrideCE and wsParams.overrideVE and wsResults.tpHitsLanded > 0 then
```
Xaver-DaRed:
```lua
if
    wsParams.overrideCE and
    wsParams.overrideVE and
    wsResults.tpHitsLanded > 0
then
```

9001-Sols: Thanks!

scripts/globals/weaponskills.lua
```diff
@@ -1052,7 +1052,11 @@ xi.weaponskills.takeWeaponskillDamage = function(defender, attacker, wsParams, p
 
     local enmityEntity = wsResults.taChar or attacker
 
-    if wsParams.overrideCE and wsParams.overrideVE then
+    if
+        wsParams.overrideCE and
+        wsParams.overrideVE and
+        wsResults.tpHitsLanded > 0
```
Xaver-DaRed: I think this last line should be
`wsResults.tpHitsLanded + wsResults.extraHitsLanded > 0`

9001-Sols: Ooh good catch, I'm not aware of any multi hittable (non-ranged) fixed enmity weaponskills but this is definitely more correct and future-proof. I'll update it!

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #6020, opened 2024-07-18
[core, lua] Give mobs guard skill and prevent some mobs from guarding

TracentEden2:
**_I affirm:_**

This PR fixes several related issues:
1. On LSB currently MNK and PUP mobs do not have any guard skill. This results in the current guard formula using zero for the skill and thus guarding at a lower rate than intended. This is causing an inconsistency because the lua WS guarding code has a check for this lack of skill, while the normal c++ auto-attack code does not.

2. Currently on LSB all MNK and PUP mobs can guard, however captures show that there are a few MNK mobs that should not be able to guard.

3. On LSB Vanguard Grappler can currently erroneously parry and guard

This PR does the following to fix these issues:
1. Adds guard skill to all MNK and PUP mobs that can guard (similar to how LSB adds parry skills to mobs that can parry). The guard skill level is set to C (the same known level as mob parry skill). Also changes the parry skill code to add skill rather than PARRY mod as this is how all other skills are set in the code. Note that Siknoz and others are currently working to validate and check the guard formula itself through captures, however this requires further captures and thus changes will come in a later PR for brevity sake.

3. Retail [captures](https://discord.com/channels/443544205206355968/674750598939279400/1260007481741148240) by Siknoz of many different MNK and PUP mob types illustrate that most MNK and PUP mobs should guard (see listing below). Thus given the small number of MNK or PUP mobs that do not guard, a `CANNOT_GUARD` mob mod was added and applied to Kindred Monk, King Arthro, Proto-Omega, and Seiryu.

      The following were tested and shown to guard:
      Orcish Champion (Orc)
      Yagudo Initiate (Yag)
      Mandragora (Mandy)
      Fomor Monk (Fomor)
      Korrigan (Mandy)
      Arrapago Apkallu (Apkallu)
      Sabotender Corrido (Sabotender)
      Molech (Taurus)
      Hilltroll Puppetmaster (Troll)
      Bugbear Deathman (Bugbear)
      Eo'aern (mnk) (Aern)
      
      The following were also tested and did not guard:
      Sea Monks (probably wars with incorrect job on LSB)
      Ancient Goobbue (probably also war)
      Kindred Monk (dyna xarc, maybe using axe weapon?)
      King Arthro
      Proto-Omega
      Seiryu

4. Remove the CAN_PARRY mod from Vanguard Grappler so it can only guard

Fight MNK or PUP mobs and see that they guard WS and normal attacks at the same rate. Also see that for example Seiryu no longer guards, and that Vanguard Grappler does not parry but only guards.

### DIFF
--- a/scripts/enum/mob_mod.lua
+++ b/scripts/enum/mob_mod.lua
@@ -87,4 +87,5 @@ xi.mobMod =
     NO_WIDESCAN            = 76, -- Disables widescan for a specific mob
     TRUST_DISTANCE         = 77, -- TRUSTS ONLY: Set movement type/distance. See trust.lua for details.
     STANDBACK_RANGE        = 78, -- Applies a specific standback range for the mob
+    CANNOT_GUARD           = 79, -- Check if the mob does not guard (despite being a MNK or PUP mob)
 }
--- a/scripts/globals/combat/physical_utilities.lua
+++ b/scripts/globals/combat/physical_utilities.lua
@@ -774,7 +774,7 @@ xi.combat.physical.canGuard = function(defender, attacker)
             defender:isPet() or
             defender:isTrust()
         then
-            canGuard = defender:getMainJob() == xi.job.MNK or defender:getMainJob() == xi.job.PUP
+            canGuard = (defender:getMainJob() == xi.job.MNK or defender:getMainJob() == xi.job.PUP) and defender:getMobMod(xi.mobMod.CANNOT_GUARD) == 0
         end
     end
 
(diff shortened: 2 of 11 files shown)


WinterSolstice8 (approved):

Xaver-DaRed (approved):

Xaver-DaRed:
Rebase, please, so the last check can be done.

TracentEden2:
> Rebase, please, so the last check can be done.

Yep, done now.

PR #6025, opened 2024-07-20
Fix a logical error in !yell command

CatsEyeXI:
**_I affirm:_**

Corrects a logical error in the !yell command.
Properly passed `msg` to `error` function.

!yell ban Player 365


Xaver-DaRed (commented):
This changes look ok, however, the `else` at line 51 is dead code, since this check already invalidates the possibility of the value being anything but `ban` and `unban`

Xaver-DaRed (approved):

PR #6029, opened 2024-07-21
[Core] Prevent players casting lots on items they can't obtain

loxleygames:
**_I affirm:_**

We shouldn't be trusting the client. Currently, you can cast lots on Rare items you already possess by sending a packet, for example by using certain Ashita plugins that can be configured to lot everything that drops. This change solves the issue.

Tested in game with a few characters using `!exec player:addTreasure(13014)`. Was not able to cast lot once already obtained, even by forcing a packet.

### DIFF
--- a/src/map/treasure_pool.cpp
+++ b/src/map/treasure_pool.cpp
@@ -266,6 +266,22 @@ void CTreasurePool::LotItem(CCharEntity* PChar, uint8 SlotID, uint16 Lot)
         return;
     }
 
+    CItem* PItem = itemutils::GetItem(m_PoolItems[SlotID].ID);
+
+    // Cannot lot if player's inventory is full
+    if (PChar->getStorage(LOC_INVENTORY)->GetFreeSlotsCount() == 0)
+    {
+        ShowError(fmt::format("Player {} is trying to lot on item {} while full inventory (Packet injection)!", PChar->getName(), m_PoolItems[SlotID].ID));
+        return;
+    }
+
+    // Cannot lot if item is RARE and player already has it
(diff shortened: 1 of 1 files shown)


zach2good:
> We shouldn't be trusting the client. 

Agreed, of course.

> Currently, you can cast lots on Rare items you already possess by sending a packet

I can't think if I've seen this scenario on retail, but this 99.9% sounds like something that's allowed though? Unless someone else weighs in I'm really going to want to see a capture of this.

zach2good:
(probably easily done many ways, but the one that comes to mind is having your Dynamis drops and going to farm more, and then seeing if you can lot on them. Will be easy with mass-pulls on retail)

zach2good:
I'm having possibly false memories of helping people grind for Joyeuse and still being able to lot on it, despite already having one

LoxleyX:
To clarify the current state is: The client stops you from lotting the items, the server doesn't.

Xaver-DaRed (commented):
lgtm. Veryfied in retail that the client does stop you.

Since I assume this is for packet injection, (couse the client already stops you), should this be pushing the messages aswell?

zach2good (approved):
This should either push a message, or log because to get here you must have been injecting packets

Xaver-DaRed:
Full inventory message
imagen

Lotting Rare item in possesion
imagen

EDIT: oops, sorry. Didnt mean to close it

zach2good (approved):
I don't remember what getName returns, is it a string or a const char*? I don't know if printf style %s takes strings. Regardless, can you change these to the Fmt variant loggers, and use the exploit logger instead of warning?

Otherwise looks good

Xaver-DaRed (approved):

PR #6033, opened 2024-07-21
[Proposal] Create mobskill enum file

Xaver-DaRed:
**_I affirm:_**

Proposes the creation of an enum file with all mobskill IDs

This PR contains all entries of "foot kick", "sheep song", etc... among others

I could have added more, however, I rather know everyones opinion before commiting to add 4000 entries by hand.
This gives a clear example of what the file would look like and allows for certain patterns to be recognized.

Its use would be to be able to have tables with mobskill info, like we do for spells, for instance. Or at least to not use raw IDs in mob family scripts and such.

None

### DIFF
--- a/scripts/enum/mob_skills.lua
+++ b/scripts/enum/mob_skills.lua
@@ -1,0 +1,87 @@
+xi = xi or {}
+
+xi.mobSkill =
+{
+    FOOT_KICK_1       =  257,
+    DUST_CLOUD_1      =  258,
+    WHIRL_CLAWS_1     =  259,
+    LAMB_CHOP_1       =  260,
+    RAGE_1            =  261,
+    SHEEP_CHARGE_1    =  262,
+    SHEEP_BLEAT_1     =  263,
+    SHEEP_SONG_1      =  264,
+    RAGE_2            =  265,
+    RAM_CHARGE        =  266,
+    RUMBLE            =  267,
+    GREAT_BLEAT       =  268,
+    PETRIBREATH       =  269,
+    ROAR_1            =  270,
+    RAZOR_FANG_1      =  271,
+    RANGED_ATTACK_1   =  272,
+    CLAW_CYCLONE_1    =  273,
+    SHEEP_CHARGE_2    =  274,
+    SANDBLAST_1       =  275,
+    SANDPIT_1         =  276,
+    VENOM_SPRAY_1     =  277,
+    PIT_AMBUSH_1      =  278,
+    MANDIBULAR_BITE_1 =  279,
+
+    RANGED_ATTACK_2   =  412,
+
+    RANGED_ATTACK_3   = 1154,
+
+    RANGED_ATTACK_4   = 1202,
+    RANGED_ATTACK_5   = 1203,
+    RANGED_ATTACK_6   = 1204,
+    RANGED_ATTACK_7   = 1205,
+    RANGED_ATTACK_8   = 1206,
+
+    RANGED_ATTACK_9   = 1209,
+    RANGED_ATTACK_10  = 1210,
+    RANGED_ATTACK_11  = 1211,
+    RANGED_ATTACK_12  = 1212,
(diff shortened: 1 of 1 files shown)


zach2good:
I'm not against it, but maybe we should name things based on where they're used. RANGED_ATTACK_GOBLIN instead of _1 etc.?

zach2good:
(only for skill names that are repeated, obviously)

jmcmorris:
I like this as I see it one step moving towards being able to arbitrarily add/remove mobskills from a mob entity - either during initialization or during combat for a phase change or whatever.

zach2good:
(out of scope of this PR, but worth starting this conversation)

Historically, an issue we have with dynamic lists (think mob skill lists, and lists of mob loot) - is how we'd track them and changes to them over time.

- If a mob starts with skill list A, it uses it, it dies, it respawns and continues to use skill list A
- If a mob starts with skill list A, changes to B at runtime, then it respawns and is set back to using skill list A
- If a mob starts with skill list A, has "RANGED_ATTACK_GOBLIN" added, then respawns and is set back to using skill list A

I think in the past one of the worries is about duplication of data - if there are 1024 mobs per zone (not always, but lets use it for the maths), and each mob has 3 skills in their list, and there are 300 zones, that's 921'600 entries across all mobs - which is cut down significantly by doing list lookups.

We had a drop list + modification list thing for treasure pools a while ago, but it was replaced with something else.

This sort of thing will need some planning is all.

jmcmorris:
I agree this needs planning and is completely out of scope for this PR. This is the sort of thing I have had to worry about a lot in my personal gamedev. Here are my thoughts:

- First, it is possible to store this mob list on each mob as a vector<uint16>. The uint16 can function as a lookup to the mobskill since the mobskills are stateless they can be shared between all mobs. This would keep the data down to likely only a few MB in memory for all mobs.
- If we are still worried about memory then another trick would be to have a uint16 that tracks which mobskill list to use (which we already do as a mobmod). If this is set then we use that mobskill list otherwise we set it to either 0 to indicate that the mob uses its own skill list. This would keep the custom lists down and also mitigate the code churn required for this change.
- I propose that this custom mobskill list would also be a weighted list - perhaps a vector<pair<uint8, uint16>>. There is a reasonable amount of evidence that mobskills are not always weighted evenly. This would greatly simplify achieving this.

jmcmorris:
I forgot to add this:

I think it would be safe to always reset the mob state at spawn time which is what we do for localvars. If we really wanted to we could use some special mobskill list IDs to indicate this. Such as 0 could be reset during spawn which is likely what we typically want. Then we could have 1 (or UINT16_MAX) be never reset. Anything other than that is considered a mobskill list ID. This would allow us to easily setup the custom list one time during mob initialization or initialize it every time during the mob spawn.

zach2good:
My main concern here is by changing the underlying storage method, we'd be giving _every mob in the game_ a customisable skill list (a vector, or a weighted lookup etc.) when the vast majority won't need them.

In the grand scheme of things the amount of memory used by this would be negligible, but there's no need to do it.

The pattern I was thinking of (and used for customisable loot pools before you did the rewrite with Lua bindings):
- By default, mobs have their skill list id
- If you call any customisation function on them (and they don't have their local list ready), it'll copy the contents of their global skill list to their local one, and then apply changes to it.
- ^ This way all mobs only have an empty vector, or an optional<vector<pair<id, weight>>> until a customisation is requested. 
- It can/will be rebuilt every time the mob spawns/resets

EDIT: I don't like the idea of a magic number skill list to tell the mob to use custom things, we should enforce using actual types/mechanisms like optional

zach2good:
TIL: Skill lists are mobMods and not just an innate part of the mob object... Why...

zach2good (approved):
Back on topic; I reckon this PR is fine. Especially with https://github.com/LandSandBoat/server/pull/6039 on the horizon - which will allow us to make sure enum renames don't get missed in areas where they're being used.

PR #6036, opened 2024-07-22
[cpp] Adds ability to modify the aoe radius of mobskills

Frankie-hz:
**_I affirm:_**

This will add the ability to specify the AOE radius of the mobskill instead of defaulting it to the distance check.

Currently the radius is just defaulted to `m_Distance` for whatever reason, it just looks like it was never implemented. I have added a default of 0 with default checks inside the `getRadius` function for mobskills we do not have the info for and defaulted it to what we had in the codebase already.

This will simply give us the ability to modify the aoe radius as each mobskill is audited/implemented.

In order to test this will you need two characters and to test a skill that has a `mob_skill_aoe` of 2 and set its `mob_skill_aoe_radius` to the desired value.

I used `trebuchet` as my test skill as I was implementing it for Dynamis mobskill audit.
Run this query:
```
UPDATE `mob_skills` SET mob_skill_aoe = 2, mob_skill_aoe_radius = 12, mob_skill_distance = 16 WHERE mob_skill_id = 1636;
```
1) Make a party with your characters
2) Go to any ranger/ninja mob (I used a dynamis ninja)  `!gotoid 16945247`
3) Have char 1 aggro the mob and stand 15 away from the target
4) Have the Chare 2 stand less than 10 yards away from Char 1
5) `!exec target:useMobAbility(1636)` on the Hitman
You should see the AOE hit you from a distance of less than 10
6) Now step farther than 10 away from Char 1
7) `!exec target:useMobAbility(1636)` on the Hitman
You should see the AOE only hit Char 1 since you are too far away


zach2good:
Adding aoe radius to all mob skills doesn't really make sense, since not all mob skills are aoe - so all the single-target skills will just have this 0.0 bit of data for no reason.

Is there actually fine-grained AoE distance? Or could we get away with extending the flags to be more than SINGLE/CONAL/CONAL_BACK/AOE and add AOE_SMALL, AOE_MEDIUM, AOE_LARGE, etc.

Frankie-hz:
> Adding aoe radius to all mob skills doesn't really make sense, since not all mob skills are aoe - so all the single-target skills will just have this 0.0 bit of data for no reason.
> 
> Is there actually fine-grained AoE distance? Or could we get away with extending the flags to be more than SINGLE/CONAL/CONAL_BACK/AOE and add AOE_SMALL, AOE_MEDIUM, AOE_LARGE, etc.

I don't think we can add flags like that because the aoe distances just from dynamis captures alone are all different. Example: Nightmare Efts:
3 skills are 10' aoe

Scorpions:
earthbreakr 40' aoe

Crabs:
Bubble shower is 11' aoe

Bunny:
Whirl Claws is 14' aoe

Ataeus:
Ground slam is 20' aoe
Treb is 12' aoe around target

Suttung:
Crystal Rain is 12' aoe

Diremites:
Spinning Top is 14' aoe
Grapple 12.5'

Rocs:
Stormwind 20' aoe
There are also several conal AOEs that have 13, 17, 16.5, 12. A lot are variance in the distances these can hit at. I couldn't think of any other way to really handle it.

WinterSolstice8 (approved):

zach2good:
> Currently the radius is just defaulted to m_Distance for whatever reason, it just looks like it was never implemented. I have added a default of 0 with default checks inside the getRadius function for mobskills we do not have the info for and defaulted it to what we had in the codebase already.

Why not just use m_Distance and the associated field in SQL for the aoe_distance?

```sql
SELECT * FROM mob_skills WHERE mob_skill_distance > 0 and mob_skill_aoe > 0
```

Seems like there are already some AoE things that are using it, or am I missing the point here? (Pre-coffee ☕ )

zach2good:
ie. is there a difference between the aoe_distance (when the skill is fired, anyone in this radius will be targetted) and the skill_distance (the distance at which the main target has to be in range for the skill to not be interrupted)?

zach2good:
(I'm rambling)

If that's the case, then this isn't strictly AoE: we'd have a targeting distance and an execution distance for all skills, regardless of AoE-ness?

Frankie-hz:
> > Currently the radius is just defaulted to m_Distance for whatever reason, it just looks like it was never implemented. I have added a default of 0 with default checks inside the getRadius function for mobskills we do not have the info for and defaulted it to what we had in the codebase already.
> 
> Why not just use m_Distance and the associated field in SQL for the aoe_distance?
> 
> ```sql
> SELECT * FROM mob_skills WHERE mob_skill_distance > 0 and mob_skill_aoe > 0
> ```
> 
> Seems like there are already some AoE things that are using it, or am I missing the point here? (Pre-coffee ☕ )

They are not the same thing. The distance (m_Distance) is the execution range as to when a skill can fire off at. This is not always and almost never the same as the aoe distance (the radius from either the target or the mob depending on what type of aoe skill it is).

For instance trebuchet is used at a maximum of 16.4 yalms but the aoe radius of the skill is 12. The function getRadius is already implemented but it simply is defaulting to the execution range. That means it fires off at 16.4 and is a 16.4 yard radius around the target, which is incorrect.

zach2good:
In which case, I'd take all the aoe-ness out of this and rename
these columns as something like targetting_distance and execution_distance,
and put an explanation of all of this somewhere. Then things aren't coded
as aoe-only etc.

Frankie-hz:
I think the execution distance would be a different column. That isn’t the same as the radius of the aoe for what I am trying to implement. Most aoe skills are actually self targeting and should only have a distance check for the player to be in range, i guess that would be your “execution distance”? They are supposed to target themselves and is a radius around themselves in which we would need  another column to modify the aoe radius anyway. The self targeted mobskills is being implemented in a different PR (not be me).

i know some targeted skills can be interrupted (like throat stab) and some cannot because the prep time is supposed to be 0 so we most likely need an execution _distance column anyway but I would assume thats out of scope for this PR, or maybe its not but I was only trying to make it so we can modify the radius of the aoe skill

I think we have a misunderstanding on what is what because i do agree things are poorly named and were never finished being implemented.

zach2good:
I don't think we have a misunderstanding, just a misalignment of what's currently available. 

If an AOE skill is self-targetting, we'll probably end up marking it as such with its AoE flags, and then the targeting distance (even if set) will be irrelevant. Or, if the targetting distance is 0.0 - then it's self-targetting (though that makes things seem unclear and it just "you have to know this" information rather than being something named). 

At that point, the execution distance is used to pick who is going to get hit with the skill.

To me (now over-caffienated), it seems like we can eventually cover all usages across all skills (aoe and single target) with the combinations of AoE-flags, execution distance, and targetting distance.

Frankie-hz:
Ah I see what you mean. I don’t think the targeting distance would be 0 regardless because it needs a player close enough to actually use the self-targeting skill anyway (I think).

Is there a better name than execution distance you can think of? I can’t really think of any that would match all cases but I can write something for documentation. I know the person doing the self targeting is writing one up as well.

When someone says “execution distance” to me that sounds like the distance as to which the skill should fire off. Targeting distance sounds very similar.

Frankie-hz:
So after a long conversation with Zach the main concern of implementing this SQL column is that there will be a lot of "dead data" sitting here being unused. All the single-target skills will never use this column. We tried to come up with a naming solution that would fit all circumstances and make the column useable for other reasons as well but there are too many variables are needed to be implemented.

`activation_distance` would be the readying distance
`execution_distance` would be the distance at which the skill can still fire (and can be greater than Activation)
However, a single-target skill has no AoE Radius.  This is not the case for an AoE skill - In this case, there could potentially be 4 Values Needed:

1 - Activation Range  (Currently Set is mob_skill_distance)
2 - Execution Range  (Currently Set in mob_skill_distance as well -- This should be isolated)
3 - Radius of the Executed AoE (Currently Set as mob_skill_distance as well -- This should also be isolated because the ready/execution range of the skill != radius of the AoE)
4 - Type of AoE (Currently Set as mob_skill_aoe)

Seeing we need all of this we would still need an sql column for the radius of the executed AOE regardless of how you look at it.

There is another way to bake this into flags but it takes the ease of implementing them via an sql column to fully understand bits with some little magic to implement any of them correctly. See below explanation by Beasty.

*Note: THIS IS A LOT MORE WORK*

**What if you use some bitwise magic to encode values**

Use `mob_skill_aoe` to set a bit-level flag on the most significant bit, then use the lesser bits to specify value
This would require that mob_skill_aoe / m_Aoe checks also be updated (more work... ofc)

However, this would let you encode a flag + distance in that same value

Example / Details
```
8 Bits of Data -- Each Bits Value:
0 0 0 0  0 0 0 0
| | | 16 | | | 1 
| | 32   | | 2
| 64     | 4
128      8

Just an example of how the bits would work:
0010 1000 = 0 + 0 + 32 + 0 + 8 + 0 + 0 + 0 = 40
```
**How do you make this work for skills?**
- The MSB (Most Significant Bit) value is 128
- The 7 lesser bits max value is 127 
- Since mob_skill_distance is the range at which the skill can be prepared/readied you leave that value as it is

You specify an AOE distance by setting the MSB to 1 + use the 7 lesser bits to specify the actual distance from the center of the AoE

```cpp
// In the AOE_RADIUS Enum:
enum class AOE_RADIUS : uint8
{
    ATTACKER   = 1,
    TARGET     = 2
    AOE_RADIUS = 128,
};
```

```cpp
// In the value checks:
if (PSkill->m_Aoe & AOE_RADIUS)
{
  PSkill->m_AoeRadius = PSkill->m_Aoe ~ AOE_RADIUS;
}
```

To break this down a bit:
Lets say you have a skill that you want an AOE Radius of 20:
You would set the following bits:
`1 0 0 1   0 1 0 0`  (148)

```cpp
// The bit magic flag
(m_Aoe & AOE_RADIUS)

1 0 0 1   0 1 0 0
1 0 0 0   0 0 0 0  BITWISE AND
------------------------------
1 0 0 0   0 0 0 0   (True)
```
This is a value greater than 1, so this essentially returns TRUE and you are stating then `if (BITWISE AND (&) OPERATION RETURNS TRUE)`
Then it drops into the code block below
```cpp
// Second Bitwise Magic (NOT)
PSkill->m_Aoe ~ AOE_RADIUS;

1 0 0 1   0 1 0 0
1 0 0 0   0 0 0 0  BITWISE NOT
------------------------------
0 0 0 1   0 1 0 0   (20 -- 0 + 0 + 0 + 16 + 0 + 4 + 0 + 0 = 20)
```

So now you have populated the AoE Radius with the value of 20 by removing that MSB Flag.

I am inclined to just go with the sql column even though there will be a lot of dead data inside but any other way we can think of sounds and feels like a massive amount of work for little gain.

Any other recommendations is greatly appreciated :D

WinterSolstice8:
> So after a long conversation with Zach the main concern of implementing this SQL column is that there will be a lot of "dead data" sitting here being unused. All the single-target skills will never use this column. We tried to come up with a naming solution that would fit all circumstances and make the column useable for other reasons as well but there are too many variables are needed to be implemented.

skillchains fields for mobskills are largely dead, so i don't really have a problem with this. The other option is to set AoE stats in the mobskill itself with lua calls

Frankie-hz:
Fixed merge conflicts and left the addition of the one column in here.

Following the last explanation These are the list of columns we need to fully implement mobskills properly
1 - Activation Range (Currently Set is mob_skill_distance)
2 - Execution Range (Currently Set in mob_skill_distance as well -- This should be isolated)
3 - Radius of the Executed AoE (Currently Set as mob_skill_distance as well -- This should also be isolated because the ready/execution range of the skill != radius of the AoE)
4 - Type of AoE (Currently Set as mob_skill_aoe)

This PR implements only 3. (4 is already implemented) Will do a separate PR implementing 1 and 2 since they are more closely related.

PR #6037, opened 2024-07-23
[core] PARTY_MEMBERS_IN_ZONE: Ensure that PMember is not nullptr in Latent Effect Container (3)

CatsEyeXI:
**_I affirm:_**

Handles an edge case where calling `getZone()` may result in retrieving a nullptr.
image

Not entirely sure how to reproduce this reliably.

### DIFF
--- a/src/map/latent_effect_container.cpp
+++ b/src/map/latent_effect_container.cpp
@@ -454,15 +454,15 @@ void CLatentEffectContainer::CheckLatentsPartyMembers(size_t members, size_t tru
                     auto inZone = 0;
                     for (size_t m = 0; m < members; ++m)
                     {
-                        auto* PMember = (CCharEntity*)m_POwner->PParty->members.at(m);
-                        if (PMember->getZone() == m_POwner->getZone())
+                        auto* PMember = dynamic_cast<CCharEntity*>(m_POwner->PParty->members.at(m));
+                        if (PMember != nullptr && PMember->getZone() == m_POwner->getZone())
                         {
                             inZone++;
                         }
                     }
 
-                    auto* PLeader = (CCharEntity*)m_POwner->PParty->GetLeader();
-                    if (m_POwner->getZone() == PLeader->getZone())
+                    auto* PLeader = dynamic_cast<CCharEntity*>(m_POwner->PParty->GetLeader());
+                    if (PLeader != nullptr && m_POwner->getZone() == PLeader->getZone())
                     {
                         inZone = inZone + static_cast<int>(trustCount);
                     }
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
There's a very brief window of time where loc/loc.zone is null when someone zones in. It realistically should never be null but I believe you when it's hard to produce

zach2good (approved):

zach2good:
`loc` isn't a pointer type, so you can't compare it with `nullptr`:
https://github.com/LandSandBoat/server/blob/base/src/map/entities/baseentity.h#L309

If you're looking at a crash and it says `this` is nullptr, it means the current class, not where the crash is.

https://en.cppreference.com/w/cpp/language/this

You'll want to look one level up from here, where latents are being processed, where `GetZone` is being called on a `nullptr` PChar, or something like that

zach2good:
This doesn't even compile because the code isn't valid?

https://godbolt.org/z/joeK5o38s

zach2good (changes requested):
See inline comments

zach2good:
Or even better, it's involved with `ReloadParty()`, RIP

CatsEyeXI:
> See inline comments

I don't see any inline comments, but I think I've moved the nullptr checks to the right place in the latent effect container.

src/map/latent_effect_container.cpp
```diff
@@ -455,14 +455,14 @@ void CLatentEffectContainer::CheckLatentsPartyMembers(size_t members, size_t tru
                     for (size_t m = 0; m < members; ++m)
                     {
                         auto* PMember = (CCharEntity*)m_POwner->PParty->members.at(m);
-                        if (PMember->getZone() == m_POwner->getZone())
+                        if (PMember != nullptr && PMember->getZone() == m_POwner->getZone())
```
zach2good, line 458: All of these are the same concept as your other PR: a `static_cast` will just stuff whatever type of pointer it finds into the shape you request, `dynamic_cast` will actually check that it's valid. If a non-player is here (and non-player parties exist, I think) this will force a Mob into the Char pointer.

zach2good: Inline comments: These, as part of the review.

Not inline comments: just comments on the PR

zach2good (changes requested):

src/map/latent_effect_container.cpp
CatsEyeXI: My apologies, can you please break it down a bit further for me, and tell me what exactly needs to change here?
Much appreciated.

zach2good: `auto* PMember = dynamic_cast<CCharEntity*>(m_POwner->PParty->members.at(m));`

CatsEyeXI: Thank you!

zach2good (approved):

PR #6039, opened 2024-07-23
Tools: Add LuaLS checks

zach2good:
**_I affirm:_**

Adds a script that uses a command line version of the Lua language server to generate an output we can use for CI. This will also pick up and use type annotations once we have/generate them, but for now is good for picking up general mistakes.

```
file:///Users/zrtoo/dev/server/scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua:1772:77 | undefined-field | Undefined field `itemId`.
file:///Users/zrtoo/dev/server/scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua:1767:37 | undefined-field | Undefined field `itemId`.
file:///Users/zrtoo/dev/server/scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua:1729:40 | undefined-field | Undefined field `augments`.
file:///Users/zrtoo/dev/server/scripts/zones/RuLude_Gardens/npcs/Delivery_Crate.lua:20:63 | redundant-parameter | This function expects a maximum of 3 argument(s) but instead it is receiving 4.
file:///Users/zrtoo/dev/server/scripts/zones/Sauromugue_Champaign/mobs/Old_Sabertooth.lua:29:31 | missing-parameter | This function requires 2 argument(s) but instead it is receiving 1.
```

Not yet tested on Linux and Windows

```sh
python3 ./tools/ci/lua_lang_server.py 
```


zach2good:
OK, this runs and uploads the warnings/errors file, and doesn't provide a fail - we can merge this now and I can tinker on it as I get time

claywar (approved):

PR #6040, opened 2024-07-23
[Lua] Change `xi.mobskills.mobMagicalMove()` function to NOT return a table

Xaver-DaRed:
**_I affirm:_**

Adresses a datatype issue were we used a variable to store int and table data willy-nilly. ThisPR will:
- Adress that by changing the offending function to NOT return a table **with a single entry** and instead return a regular value.
- Audit all scripts which used that function so:
  - They all have the same style
  - Improve readability
  - Status effects are always applied AFTER the skill damage is done. This may or may not solve issues with instant-break sleep, bind etc. Order is now as follows: Dispeling, damage, buff and defuff

Let mobs do their thing without making your server vomit error messages non-stop?

### DIFF
--- a/scripts/actions/mobskills/aerial_blast.lua
+++ b/scripts/actions/mobskills/aerial_blast.lua
@@ -9,13 +9,14 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    local dmgmod = 3
-    local info   = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg() * 9, xi.element.WIND, dmgmod, xi.mobskills.magicalTpBonus.NO_EFFECT, 1)
-    local dmg    = xi.mobskills.mobFinalAdjustments(info.dmg, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.WIND, xi.mobskills.shadowBehavior.WIPE_SHADOWS)
+    local damage = mob:getWeaponDmg() * 9
 
-    target:takeDamage(dmg, mob, xi.attackType.MAGICAL, xi.damageType.WIND)
+    damage = xi.mobskills.mobMagicalMove(mob, target, skill, damage, xi.element.WIND, 3, xi.mobskills.magicalTpBonus.NO_EFFECT, 1)
+    damage = xi.mobskills.mobFinalAdjustments(damage, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.WIND, xi.mobskills.shadowBehavior.WIPE_SHADOWS)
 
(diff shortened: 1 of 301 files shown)


scripts/actions/mobskills/aerial_blast.lua
```diff
@@ -9,13 +9,14 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    local dmgmod = 3
```
Frankie-hz, line 12: Is there a reason we are removing the local definition and just putting the number inside the damage function?

Xaver-DaRed: Yes. I'm instead storing the base damage separately, which extremely oftenly is already being multiplied. Whereas, that separate multiplier is `1` half of the time

Consider it a cleanup for future revision. I dont see the need of a separate multiplier when `base damage` is already being multiplied 99% of the time by random values.

Frankie-hz: Since you were going through all of them already wouldn't it of been easier to just remove it and do that revision now? I guess its kind of late for that since you're already 210 files in ha. The dmgmod I am guessing was the "ftp" and whoever started to add weapondmg * (random number) just didnt understand it as there was no real documentation from when it was originally written.

Talk for another day lol

Xaver-DaRed: The purpose of this PR was to fix a very simple issue. The whole "While im here, im gonna reformat all scripts that used the original function" is just my OCD not letting me ignore the other issues.

Reviewing this PR is already gonna be a nightmare without a lot of faith, so I rahter leave that for a different PR, which will be much easier to review, because all scripts will only have a single line change

scripts/actions/mobskills/ion_shower.lua
```diff
@@ -12,13 +12,15 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local damage = mob:getWeaponDmg() * 3.5
+
+    damage = xi.mobskills.mobMagicalMove(mob, target, skill, mob:getWeaponDmg() * 3.5, xi.element.THUNDER, 1, xi.mobskills.magicalTpBonus.NO_EFFECT)
```
Frankie-hz: `mob:getWeaponDmg() * 3.5` to `damage`

Frankie-hz (changes requested):

Frankie-hz:
Ran a bunch of regex commands for everything I could think of, lgtm

Xaver-DaRed:
TY man.

zach2good (approved):

Issue #4675, opened 2023-11-03
🐛 Off hand weapon or strap is unequipped on zone


Alsbuser:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Mac

## Branch affected by issue 

`base`

## Steps to reproduce

Equip a weapon or strap in the sub weapon slot. Zone. You should now have that strap or weapon unequipped (see below additional info). It may not happen if the unordered map provides the right order back on zone, so you may have to try different combinations. Lower level gear, or perhaps container slot may play a factor in how the hash is created in the unordered map

## Expected behavior

Equipment should stay as equipped in other zones when zoning.

Additional info: The issue seems to be with the LoadEquip method in charutils. Specifically the logical strategy to keep the sql return inside an unordered map called "equipSlotData". Since the unordered map isn't going to reliably output the main weapon first in the loop, you end up in situations where sub weapon slots that can't be equipped without a main weapon (anything not a shield I believe) never get re-equipped on character load.

RAIST5150:
Would note there is a discussion thread where we found a way to duplicate it on demand... have not had a chance to test different gear sets yet but it was easy to force it in our PJ's

#4611

Xaver-DaRed:
This was fixed in PR 6044

PR #6044, opened 2024-07-24
Fix sub slot being unequipped on zone

DotesMiqotes:
**_I affirm:_**

Fixes the issue of off hand weapons, shields, and grips being unequipped after zoning on Ubuntu 22.04 servers. #4675 

I compiled the server on Ubuntu 22.04 and tried zoning multiple times on both my gm and non gm character with different shields and offhand weapons equipped.


zach2good (approved):

src/map/utils/charutils.cpp
```diff
@@ -1023,7 +1023,7 @@ namespace charutils
         if (rset)
         {
             // equipSlotData[equipSlotId] = { slotId, containerId }
-            std::unordered_map<uint8, std::pair<uint8, uint8>> equipSlotData;
+            std::map<uint8, std::pair<uint8, uint8>> equipSlotData;
```
zach2good, line 1026: As part of this, do we need a `SORT ASC/DESC` on the query? Or is there some other mechanism enforcing that?

DotesMiqotes: We shouldn't need to, elements of a map are sorted by their key. From what I can gather from the code, and what I've seen in my own testing, the sorting here seems to be based off of SLOTTYPE in battleentity.h.

Xaver-DaRed (approved):

PR #6051, opened 2024-07-26
[Lua] Add player parrying, guarding, and blocking of physical (non-ranged) mobskills

### LINKED ISSUE #836: Monster physical TP moves should be able to be guarded, blocked, or parried.
- [x] I have specified what branch this happens on branch: EDITME

You should be able to block/parry/guard all hits of a physical mob TP move. 
Example: Mob uses goblin rush. You get really lucky, and block 3/3 hits.

TracentEden2:
**_I affirm:_**

This PR adds player parrying, guarding, and blocking of physical (non-ranged) mobskills. This is done by leveraging the functions from the previous [PR](https://github.com/LandSandBoat/server/pull/5873) that added monster parrying, guarding, and blocking of player weaponskills. When a player guards or parries a physical mobskill hit the hit simply misses (though in contrast to a normal miss there is the possibility of a guard or parry skillup), see this [source](https://genomeffxi.livejournal.com/18269.html) for evidence for guard for example.

Closes https://github.com/LandSandBoat/server/issues/836

Fight mobs and notice that they miss physical mobskills more (due to guarding and parrying) and that you can get parrying and guarding skillups during these. Also similar for blocking.

### DIFF
--- a/scripts/globals/mobskills.lua
+++ b/scripts/globals/mobskills.lua
@@ -110,6 +110,25 @@ xi.mobskills.mobRangedMove = function(mob, target, skill, numberofhits, accmod, 
     return xi.mobskills.mobPhysicalMove(mob, target, skill, numberofhits, accmod, dmgmod, xi.mobskills.magicalTpBonus.RANGED)
 end
 
+-- helper function to handle a single hit and check for parrying, guarding, and blocking
+local function handleSinglePhysicalHit(mob, target, hitdamage, hitslanded, finaldmg, tpEffect, minRatio, maxRatio)
+    -- if a non-ranged physical mobskill then can parry or guard
+    if
+        tpEffect == xi.mobskills.magicalTpBonus.RANGED or
+        (not xi.combat.physical.isParried(target, mob) and
+        not xi.combat.physical.isGuarded(target, mob))
+    then
+        local pdif = math.random((minRatio * 1000), (maxRatio * 1000)) --generate random PDIF
+        pdif = pdif / 1000 --multiplier set.
+        finaldmg = finaldmg + hitdamage * pdif
+        -- also handle blocking
+        finaldmg = xi.combat.physical.handleBlock(target, mob, finaldmg)
+        hitslanded = hitslanded + 1
+    end
+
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):
I don't think guarding actually nullifies a hit, the messaging is probably just wrong to the player
that said, this is better than what we had so approved

TracentEden:
> I don't think guarding actually nullifies a hit, the messaging is probably just wrong to the player that said, this is better than what we had so approved

You mean that probably incorrect code by SE and it should be an evasion skillup in those screenshots?

WinterSolstice8:
> > I don't think guarding actually nullifies a hit, the messaging is probably just wrong to the player that said, this is better than what we had so approved
> 
> You mean that probably incorrect code by SE and it should be an evasion skillup in those screenshots?

It probably lowers the PDIF by 1.0 and still does damage, but SE messages a "miss" (if all hits were guarded) because a standard guard has the "evaded" bit and their logic is probably confused. wouldn't be the first time they message junk (see wyvern breath MBs recording that the wyvern recovers HP instead of dealing damage to the target)

scripts/globals/mobskills.lua
```diff
@@ -116,6 +116,9 @@ end
 -- dmgMod   : linear multiplier for damage (1 default)
 -- tpEffect : Defined in xi.mobskills.physicalTpBonus
 -----------------------------------
+-- TODO: Reduce complexity
+-- Disable cyclomatic complexity check for this function:
+-- luacheck: ignore 561
```
zach2good: You've expanded two blocks that look exactly the same at a glance. Break them out into a helper function with multiple args and multiple returns and you'll be able to get rid of this exception

TracentEden: > You've expanded two blocks that look exactly the same at a glance. Break them out into a helper function with multiple args and multiple returns and you'll be able to get rid of this exception

Thanks for the suggestion, I broke out the logic to a helper function as suggested and updated the PR.

zach2good (approved):

PR #6055, opened 2024-07-27
Trust: Fablinix

DotesMiqotes:
**_I affirm:_**

Implements Trust: Fablinix based on the information from [https://www.bg-wiki.com/ffxi/Cipher:_Fablinix](url) and [https://ffxiclopedia.fandom.com/wiki/Trust:_Fablinix](url). The changes include implementing all known mechanics, as well as Fablinix's and the Moogle's teamwork messages.

I slowly added each skill and ability and play tested them and their interactions, only making commits to my personal branch when the results reflected the information given about the trust. Unfortunately I don't have this trust on retail so I'm limited to the information that's publicly available. As such, I slightly under tuned Fablinix's reported ranged attack rate to be on the safe side, and left a TODO that the rate needs to be confirmed.


src/map/spell.cpp
```diff
@@ -698,6 +698,10 @@ namespace spell
                 // ensure trust level is appropriate+
                 if (PCaster->objtype == TYPE_TRUST && PCaster->GetMLevel() < JobMLVL && PCaster->GetSLevel() < JobSLVL)
                 {
+                    if (spell->getID() == SpellID::Stun && !PCaster->name.compare("fablinix") && PCaster->GetMLevel() >= spell->getJob(JOB_BLM))
+                    {
+                        return true;
+                    }
```
zach2good: What's happening here? We typically don't make special exceptions in the C++ code for certain spells, or entities.

If spell is stun, and anyone but fablinix casts stun, and the caster's main job level is higher than their BLM level?

zach2good (changes requested):
Seems fine, Fablinix isn't typically an S or A-tier trust, so if they're a little off for delays and things that's probably fine. If this was a main healer or buffing trust it might be a different story.

I think I have this Trust, what additional information do you want/need?

src/map/spell.cpp
DotesMiqotes: This is to emulate Fablinix's "second main" job of BLM since THF can't cast stun. I debated using CanUseSpellWith() here, however it seemed less efficient, and would still require the name check, or trusts like Adelheid would be able to cast Stun under level.

The second part is if Fablinix casts stun since std::string::compare() returns 0 if both strings are equal.

As far as additional information goes, the number of auto attacks between ranged attacks should be the only thing I need to know to get the rate down on Fablinix's ranged attacks. His TP per hit for his dagger and ranged attacks would help too, but it seems like auto attack delay is a trust wide issue right now, and would be a bit beyond the scope of just implementing this trust. Getting the delay and TP per hit issues for trusts sorted out is on my todo list.

zach2good:
From memory, it ultimately doesn't matter the order in which you specify
the jobs for mobs (Trusts are children of CMobEntity, after all) - since
they get equal levels of their main and sub-jobs, and all the traits of
both. So if you need behaviour that comes from BLM main, just set BLM main
and leave a note in the Trust's script saying that's what you've done.

DotesMiqotes:
Is there any way to change an entity's main job via script aside from
changeJob()? The issue I'm having with changeJob(xi.job.BLM) is that it
recalculates all of Fablinix's stats, and ends up giving him around 2500 mp
at level 75.

zach2good:
Even after taking off the MPP 2.5x modifier from the mob pool mods?

DotesMiqotes:
According to FFXI Wiki, his main job shouldn't give him any MP. Fablinix's
high MP pool and access to Stun <https://www.bg-wiki.com/ffxi/Stun> at the
BLM level (lv.42) may give the impression that he is a BLM or has multiple
main jobs. However, if you use him in area with subjob restrictions he will
have 0 MP. If you give him at least +8MP through AoE Food
<https://www.bg-wiki.com/ffxi/AoE_Food> (before applying his +350%) or by
raising your item level, he will still be able to cast Stun but none of the
other spells.

DotesMiqotes:
I had a hunch, so I went through all of the trusts on BGWiki and put
together a list of the special cases where trusts get spells they
shouldn't. They'll likely have the same issue as Fablinix with spell.cpp
line 699. None of them are fully implemented yet, but would need special
cases to be able to cast some of their spells. The level check is based on
the level the trust's job/sub job get the spell. I'd imagine that if I
changed it to check the min/max level found in mob_spell_lists.sql, it
would negate the need for any special cases.
King of Hearts - RDM/WHM - Firaga IV @ level 92 (mob only)
Noillurie - SAM/PLD - Cure IV (PLD 55 - Would need SAM 110)
Iroha - SAM/WHM - Protectra V, Shellra V (WHM 75)
Iroha II - SAM/WHM - P5, S5, Flare II (BLM 75 reported as near insta cast)

zach2good:
If the case is that there are exceptions where an entity will get access to spells outside of the regular level they might be getting them - the answer will be to add a new Lua binding, or a mobMod, or some other mechanism to allow them to be added through scripting - rather than adding special code including names or IDs inside of core.

I'm pretty sure (it's been a long time) that Trusts lean heavily on `mob_spell_container`.

In _this_ case, I think it would be more relevant for you to mark down these edge cases you've found in the trust scripts that have them, and then also make a GH issue to help track it. 

We have a few things (skill lists, spell lists, loot lists) that allow you to set different lists through scripting, but not to really modify the content of the list on a per-entity basis. We talk about it here (https://github.com/LandSandBoat/server/pull/6033), but we haven't come up with a solution yet.

DotesMiqotes:
Understood, I'll go ahead and revert the changes to spell.cpp. The edge
cases I listed still need gambits, it's just an issue I foresee after
implementing Fablinix. I think I may have figured out a solution for the
level check on trust spells, I just need some time to work on it when
things get less hectic for me. It'll make the level check for trusts
casting spells based on the min_level value in their spell list versus
their main job level. From what I can see, everything needed from the
database is already there. It should fix the problem without any special
overrides being needed anywhere, script or core.

sql/mob_skills.sql
```diff
@@ -3272,8 +3272,8 @@ INSERT INTO `mob_skills` VALUES (3257,1299,'shibaraku',1,7.0,2000,1500,4,0,0,1,1
 INSERT INTO `mob_skills` VALUES (3258,363,'shiko_no_mitate',0,7.0,2000,1500,1,0,0,0,0,0,0);
 INSERT INTO `mob_skills` VALUES (3259,364,'happobarai',1,7.0,2000,1500,4,0,0,0,5,8,0);
 INSERT INTO `mob_skills` VALUES (3260,354,'rinpyotosha',1,7.0,2000,1500,1,0,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (3261,335,'bomb_toss',0,7.0,2000,1500,4,0,0,0,0,0,0);
-INSERT INTO `mob_skills` VALUES (3262,334,'goblin_rush',0,7.0,2000,1500,4,0,0,0,0,0,0);
+INSERT INTO `mob_skills` VALUES (3261,335,'bomb_toss',2,13.5,2000,1500,4,0,0,0,3,6,0);
```
zach2good, line 3275: Good job picking these as the skills - I was going to question how you picked this specific `bomb_toss`, but this is where all the trust skills live 👍

zach2good (approved):

PR #6056, opened 2024-07-28
[lua] [food] Remove fake effect of Movalpolos Water

WinterSolstice8:
**_I affirm:_**

Removes the fake effect of movalpolos water that only ffxiclopedia claimed was "true".
Applies the real effect (apparently 30 minutes of food effect?...), JP wiki also claims this.

I never once got refresh on lightsday with 0 MP
image

Trying to use movalpolos water twice:
image

yes I spent hours yesterday killing mobs with signet up on my ilvl blm to claim movalpolos just for this test...

Use movalpolos water, get food "effect".


Frankie-hz (approved):
😉

Xaver-DaRed (approved):

PR #6058, opened 2024-07-29
Flan Family Mobskills

jmcmorris:
**_I affirm:_**

This focuses on the Flan mob family - mostly the mobskills for it. There were several missing mobskills that had to be added in and some that needed a few changes based on retail data.

One of the larger changes related to this is that I made it so that `CMobSkill` will store its targets and expose it to scripts. This is needed because we need to use logic based on all of the targets inside the mobskill, Synergism.

Another change that may seem unusual is that Flan do not link but when they use Synergism they will steal HP from nearby Flan. The best way to do this was to put them into a party by saying that they link but then setting `MOBMOB_NO_LINK` so that they do not actually link.

## Retail Proof

[Jimmayus' Monster Stats Sheet](https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=57955395#gid=57955395)
[Melo's Flan captures](https://discord.com/channels/443544205206355968/443894311881408522/1262474594128891957)
[Flan Wiki page](https://ffxiclopedia.fandom.com/wiki/Category:Flans)

1. `!gotoid 17027318`
2. Fight a Flan and use `!tp 3000` to get it to use various tp moves
3. `!exec target:useMobAbility(1823)` -- xenoglossia
4. `!exec target:useMobAbility(1825)` -- amorphic_scythe
5. `!exec target:useMobAbility(1826)` -- synergism


scripts/actions/mobskills/synergism.lua
```diff
@@ -0,0 +1,35 @@
+-----------------------------------
+-- Synergism
+-- Description: Steals HP from any nearby flans.
```
Frankie-hz, line 3: It might be a good idea to put a small explanation on *how* we got these because someone will come in and eventually ask.

scripts/actions/mobskills/synergism.lua
```diff
@@ -0,0 +1,35 @@
...
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    if target:getID() == mob:getID() then
+        local absorbed = 0
+        local count = skill:getTotalTargets() - 1  -- Do not include self
+        if count > 0 then
+            local percentage = math.max(0.3, (100 - skill:getMobHPP()) / 200)
```
Frankie-hz: Is there a reason for the math.max? Wouldnt this mean anything under 30% gets defaulted to 30%

scripts/actions/mobskills/synergism.lua
jmcmorris: Added a comment in the code

scripts/actions/mobskills/synergism.lua
jmcmorris: You're right, this doesn't belong. It was something I was experimenting with while trying to figure out the algorithm. I removed it.

src/map/mobskill.h
```diff
@@ -61,27 +65,28 @@ class CMobSkill
...
+    float                              getDistance() const;
+    uint8                              getFlag() const;
+    uint16                             getAnimationTime() const;
+    uint16                             getActivationTime() const;
+    uint16                             getMsg() const;
+    uint16                             getAoEMsg() const;
+    uint16                             getValidTargets() const;
+    int16                              getTP() const;
+    uint8                              getHPP() const;
+    const std::vector<CBattleEntity*>& getTargets() const;
```
zach2good: You can make this auto and trailing return (->) to avoid changing all these lines with clang-format

zach2good (approved):

Xaver-DaRed (approved):

Issue #6063, opened 2024-07-31
🐛 Trusts are unable to cast spells that players don't normally get


DotesMiqotes:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

Ubuntu 22.04

## Branch affected by issue 

`base`

## Steps to reproduce

Add a gambit to a trust for a spell that they should be able to cast, but a player can't with the trust's job and sub job level.

## Expected behavior

Trusts should be able to cast spells based on their level and the min level listed in their spell list.

## Additional Notes
The issue is with spell.cpp line 699 and how it checks the trust main job and sub job levels against what level a player can cast the spell.
https://github.com/LandSandBoat/server/blob/17b4bb81602bbb8b653bd4b7f1e51fa53f10636b/src/map/spell.cpp#L685-L686
https://github.com/LandSandBoat/server/blob/17b4bb81602bbb8b653bd4b7f1e51fa53f10636b/src/map/spell.cpp#L699-L702

It's keeping Stun from working on my implementation of Fablinix #6055 .
https://github.com/LandSandBoat/server/blob/17b4bb81602bbb8b653bd4b7f1e51fa53f10636b/sql/mob_spell_lists.sql#L3504

It will most likely be a problem for the trusts and spells listed below. I'm already planning on implementing these trusts, and am not opposed to including their spell gambits as part of the fix I'm working on.

King of Hearts - RDM/WHM - Firaga I-III (BLM) Firaga IV @ level 92 (mob only)
https://github.com/LandSandBoat/server/blob/17b4bb81602bbb8b653bd4b7f1e51fa53f10636b/sql/mob_spell_lists.sql#L4132-L4135
https://github.com/LandSandBoat/server/blob/17b4bb81602bbb8b653bd4b7f1e51fa53f10636b/scripts/actions/spells/trust/king_of_hearts.lua#L14-L16
[...]

PR #6064, opened 2024-08-01
Implements GetSpellMinLevel() to fix issue #6063

DotesMiqotes:
**_I affirm:_**

Fixes issue #6063 by doing the following:
Adds `uint16 GetSpellMinLevel(SpellID spellId)` to `mob_spell_list` that returns the `min_level` of a spell from a mob's spell list.
Changes the trust level check in spell.cpp to use `GetSpellMinLevel()`.

Recompiled the server and tested multiple trusts at different levels to make sure they were casting the right spells based on their `min_level` in `mob_spell_lists.sql`.

### DIFF
--- a/src/map/utils/mobutils.cpp
+++ b/src/map/utils/mobutils.cpp
@@ -287,23 +287,6 @@ namespace mobutils
 
     /************************************************************************
      *                                                                       *
-     *  Returns the min_level of a spell from a mob's spell list             *
-     *                                                                       *
-     ************************************************************************/
-    uint16 GetSpellMinLevel(CMobEntity* PMob, SpellID spellId)
-    {
-        for (std::vector<MobSpell_t>::iterator it = PMob->m_SpellListContainer->m_spellList.begin(); it != PMob->m_SpellListContainer->m_spellList.end(); ++it)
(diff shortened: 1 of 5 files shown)


src/map/utils/mobutils.cpp
```diff
@@ -285,6 +285,23 @@ namespace mobutils
         return sJobStat;
     }
 
+    /************************************************************************
+     *                                                                       *
+     *  Returns the min_level of a spell from a mob's spell list             *
+     *                                                                       *
+     ************************************************************************/
+    uint16 GetSpellMinLevel(CMobEntity* PMob, SpellID spellId)
```
zach2good: https://godbolt.org/z/q74cW6c8Y

DotesMiqotes: Thank you, I was missing the ` const` in `auto const&` that was keeping the Ranged-based for loop from working. I was also able to take it a step further based on the code in your godbolt.org link, and moved the function to `mob_spell_list`.

It cleans up everything nicely. I'll also update the details in the PR to reflect the move.

zach2good (approved):

PR #6072, opened 2024-08-04
Core: Gardening updates from Eden

zach2good:
**_I affirm:_**
- [not yet] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

I was poking around open PRs and I saw we have some stuff about gardening, so I went and did some comparisons with Eden. I clarified some comments and took what changes I noticed by eye

### DIFF
--- a/src/map/utils/gardenutils.cpp
+++ b/src/map/utils/gardenutils.cpp
@@ -90,7 +90,7 @@ namespace gardenutils
         for (auto containerID : { LOC_MOGSAFE, LOC_MOGSAFE2 })
         {
             CItemContainer* PContainer = PChar->getStorage(containerID);
-            for (int slotID = 0; slotID < PContainer->GetSize(); ++slotID)
+            for (int slotID = 1; slotID <= PContainer->GetSize(); ++slotID)
             {
                 CItem* PItem = PContainer->GetItem(slotID);
                 if (PItem != nullptr && PItem->isType(ITEM_FURNISHING))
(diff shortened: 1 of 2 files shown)


src/map/utils/gardenutils.cpp
```diff
@@ -275,7 +278,14 @@ namespace gardenutils
             strength += dominantAura / 10;
         }
 
-        strength += (int16)((100 - strength) * (PItem->getStrength() / 32.0f));
+        strength += (int16)((100 - strength) * (PItem->getStrength() / 31.0f));
```
lapislosh, line 281: I have a previous PR that fixes this a different way: https://github.com/LandSandBoat/server/pull/4522
I chose to increase the RNG to 0-32 and keep the divisor as 32, I guess they chose to drop the divisor to 31 but leave the range from 0-31.  Either is probably fine tbh but we can't have both changes together or it'll be wrong.

lapislosh:
Also for completeness sake, I had removed double ore harvests from my previous PR with the comment:

> If we did want to keep the maximum as 2 but make it super rare, we would need to adjust the gardening_results.sql to have multiple entries for each ore, and only the last entry would have a yield of 2.

This PR makes this change, and so double-ore harvests will now be possible again (and it's done the correct way here with 2 separate entries to make it rare, instead of having one entry with a range of 1-2).

src/map/utils/gardenutils.cpp
zach2good: Thanks for weighing in, I was just pulling things over willy-nilly 👍

zach2good:
I've abandoned this PR. Downstream has apparently been testing and trying to fix some of the small bugs in gardening, so hopefully they'll eventually find their way back upstream one day

PR #6074, opened 2024-08-04
[cpp/sql/lua] Add WEATHER_CONDITION for latents + sunlight pole latent effect

jamesbradleym:
**_I affirm:_**

Implements a weather condition check during latent effect checks for NONE, SUNSHINE, CLOUDS, FOG
Updates Sunshine Pole latent effect with [1/tick regen during sunny weather](https://ffxiclopedia.fandom.com/wiki/Sunlight_Pole?oldid=1115539). 

Build map, update item_latents.sql with dbtool

!additem 17529 -- Sunlight Pole

!setweather 1 -- Sunshine

### DIFF
--- a/scripts/enum/latent.lua
+++ b/scripts/enum/latent.lua
@@ -43,7 +43,7 @@ xi.latent =
     JOB_MULTIPLE_AT_NIGHT    = 39, -- PARAM: 0: ODD, 2: EVEN, 3-99: DIVISOR
     EQUIPPED_IN_SLOT         = 40, -- When item is equipped in the specified slot (e.g. Dweomer Knife, Erlking's Sword, etc.) PARAM: slotID
     DURING_WS                = 41, -- During Weaponskill
-    -- 42 free to use
+    WEATHER_CONDITION        = 42, -- See weather.lua for xi.weather enum
     WEAPON_DRAWN_HP_UNDER    = 43, -- PARAM: HP PERCENT
     NATION_CITIZEN           = 44, -- Triggered by player being citizen of nation matching param: 0 San d'Oria, 1 Bastok, 2 Windurst
     MP_UNDER_VISIBLE_GEAR    = 45, -- mp less than or equal to %, calculated using MP bonuses from visible gear only
--- a/src/map/latent_effect.h
+++ b/src/map/latent_effect.h
@@ -71,27 +71,27 @@ enum class LATENT : uint16
     JOB_MULTIPLE_AT_NIGHT  = 39, // PARAM: 0: ODD, 2: EVEN, 3-X: DIVISOR
     EQUIPPED_IN_SLOT       = 40, // When item is equipped in the specified slot (e.g. Dweomer Knife, Erlking's Sword, etc.) PARAM: slotID
     DURING_WS              = 41, // During WS
-    // 42 free to use
-    WEAPON_DRAWN_HP_UNDER = 43, // PARAM: HP PERCENT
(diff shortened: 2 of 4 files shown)


src/map/latent_effect.h
```diff
@@ -71,27 +71,27 @@ enum class LATENT : uint16
...
-    ZONE_HOME_NATION      = 54, // in zone and citizen of nation (aketons)
-    MP_OVER               = 55, // mp greater than # - PARAM: MP #
-    WEAPON_DRAWN_MP_OVER  = 56, // while weapon is drawn and mp greater than # - PARAM: MP #
-    ELEVEN_ROLL_ACTIVE    = 57, // corsair roll of 11 active
-    IN_ASSAULT            = 58, // is in an Instance battle in a TOAU zone
-    VS_ECOSYSTEM          = 59, // Vs. Specific Ecosystem ID (e.g. Vs. Plantoid: Accuracy+3)
-    VS_FAMILY             = 60, // Vs. Specific Family ID (e.g. Vs. Korrigan: Accuracy+3)
-    VS_SUPERFAMILY        = 61, // Vs. Specific SuperFamily ID (e.g. Vs. Mandragora: Accuracy+3)
-    MAINJOB               = 62, // mainjob - PARAM: JOBTYPE
+    WEATHER_CONDITION      = 42, // PARAM: 0: FAIR, 1: SUNSHINE, 2: CLOUDS, 3: FOG
```
WinterSolstice8: Comment should be "See Weather.h for WEATHER enum" since it's not just for those 4 params
also, not sure why there's an extra space for alignment here when it seems to be aligning to nothing?

scripts/enum/latent.lua
```diff
@@ -43,7 +43,7 @@ xi.latent =
     JOB_MULTIPLE_AT_NIGHT    = 39, -- PARAM: 0: ODD, 2: EVEN, 3-99: DIVISOR
     EQUIPPED_IN_SLOT         = 40, -- When item is equipped in the specified slot (e.g. Dweomer Knife, Erlking's Sword, etc.) PARAM: slotID
     DURING_WS                = 41, -- During Weaponskill
-    -- 42 free to use
+    WEATHER_CONDITION        = 42, -- PARAM: 0: FAIR, 1: SUNSHINE, 2: CLOUDS, 3: FOG
```
WinterSolstice8: Comment should be "see xi.weather enum"

src/map/latent_effect.h
jamesbradleym: Updates comment 😊

Not sure I understand your note, "not sure why there's an extra space for alignment here when it seems to be aligning to nothing", prior to adding that space the enum was not aligned? I thought it was required to have enums `=` aligned to one another?

Screenshot 2024-08-04 at 11 19 50 PM

scripts/enum/latent.lua
jamesbradleym: Updated comments, thanks! 😊

src/map/latent_effect.h
WinterSolstice8: ah, I see what you did now. That wasn't visible on the diff but it is when I look at the code after the commit. nevermind there

WinterSolstice8 (approved):

src/map/latent_effect.h
jamesbradleym: Ahh, gotcha. Thanks for the review!

zach2good (approved):

Issue #5318, opened 2024-03-28
🐛 Player pets do not respect mob_spell_list.sql or castSpell()


UmeboshiXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base` LandSandBoat (Non modified)

## Steps to reproduce

Regular mobs and dynamic entity mobs only limitations for spell casting are currently MP. They can cast any spell (Sans GEO spells) if forced to with castSpell(). They will respect spell levels ranges which are defined in mob_spell_list.sql.

However for player summoned pets (Spirits, Avatars, Jug Pets, etc), they respect the level/job requirements listed in spell_data.sql. Their spell lists can be pulled from mob_spell_lists.sql but if they are not the correct job/level, they won't be able to cast them even if within the level range defined in mob_spell_lists.

I tested the spell requirements by attempting to have a BLM Player pet(A Spirit) cast poison_ii using "!exec target:castSpell(221)". It could cast it. I then tried making it cast Poison_III and it would not cast.

By copying the Job/Level blob in spell_data for Poison_II and copying it over to Poison_III, the spirit was then able to cast it. Poison_III normally has no level/job retirement since its only casted by mobs.

Another test for job requirement it just trying to get a SMN spirit to cast a self buff like an En-Spell or a spell BLM does not have access to.. If the spirit is a BLM job, it will not cast even if you try to force it to with castSpell().

## Expected behavior

* Mobs regardless if they are a player pet should not be restricted and should respect the parameters set in mob_spell_lists.

I'm not sure if there would be any outliers where this might affect retail emulation accuracy but it is an inconsistency I ran into doing custom work.

DotesMiqotes:
My recent PR #6064 should be able to fix this issue. It should just be a matter of changing what the entity's main level is being checked against.

DotesMiqotes:
[...]

PR #6075, opened 2024-08-04
Fixed level check for pet spells

DotesMiqotes:
**_I affirm:_**

Fixes issue #5318 by changing the level check for pets casting spells to check the `min_level` value from `mob_spell_lists.sql` instead of checking if a player of the pet's job and level can cast the spell.
Fully optimizes `CanUseSpell()` by converting `PCaster->objtype` if statements to `switch(PCaster->objtype)`, removing no longer needed variables, and moving all function calls to their relevant cases.

Summoned Light Spirit at various levels in Cape Teriggan and let it go through it's spell list multiple times while fighting things, before and after recompiling the server with the changes. Before it would stop casting Banish after level 90, and it was able to cast Banish IV (non-player spell on it's spell list) after.
Tested every possible case I could think of covered by the optimization to ensure everything is working properly.

### DIFF
--- a/src/map/spell.cpp
+++ b/src/map/spell.cpp
@@ -679,36 +679,26 @@ namespace spell
     // Check If user can cast spell
     bool CanUseSpell(CBattleEntity* PCaster, CSpell* spell)
     {
-        bool usable = false;
-
-        if (spell != nullptr)
+        if (spell == nullptr)
         {
-            uint8 JobMLVL      = spell->getJob(PCaster->GetMJob());
-            uint8 JobSLVL      = spell->getJob(PCaster->GetSJob());
-            uint8 requirements = spell->getRequirements();
+            return false;
+        }
 
-            if (PCaster->objtype == TYPE_MOB || (PCaster->objtype == TYPE_PET && static_cast<CPetEntity*>(PCaster)->getPetType() != PET_TYPE::ADVENTURING_FELLOW) ||
-                PCaster->objtype == TYPE_TRUST)
-            {
-                // cant cast cause im hidden or untargetable
+        bool  usable = false;
+        uint8 requirements;
+
+        switch (PCaster->objtype)
+        {
+            case TYPE_MOB:
+                // Unable to cast because caster is hidden or untargetable
                 if (PCaster->IsNameHidden() || static_cast<CMobEntity*>(PCaster)->GetUntargetable())
                 {
                     return false;
                 }
-
-                // Ensure pet or trust is level appropriate
-                if (((PCaster->objtype == TYPE_PET && static_cast<CPetEntity*>(PCaster)->getPetType() != PET_TYPE::AUTOMATON) || PCaster->objtype == TYPE_TRUST) &&
-                    PCaster->GetMLevel() < static_cast<CMobEntity*>(PCaster)->m_SpellListContainer->GetSpellMinLevel(spell->getID()))
-                {
-                    return false;
-                }
-
                 // Mobs can cast any non-given char spell
                 return true;
-            }
 
-            if (PCaster->objtype == TYPE_PC)
-            {
+            case TYPE_PC:
                 if (spell->getSpellGroup() == SPELLGROUP_TRUST)
                 {
                     return true; // every PC can use trusts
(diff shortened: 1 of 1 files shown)


src/map/spell.cpp
```diff
@@ -687,7 +687,7 @@ namespace spell
             uint8 JobSLVL      = spell->getJob(PCaster->GetSJob());
             uint8 requirements = spell->getRequirements();
 
-            if (PCaster->objtype == TYPE_MOB || (PCaster->objtype == TYPE_PET && static_cast<CPetEntity*>(PCaster)->getPetType() == PET_TYPE::AUTOMATON) ||
+            if (PCaster->objtype == TYPE_MOB || (PCaster->objtype == TYPE_PET && static_cast<CPetEntity*>(PCaster)->getPetType() != PET_TYPE::ADVENTURING_FELLOW) ||
```
zach2good: Original logic:
- If the caster if a mob, or a pet automaton, or a trust

New logic:
- If the caster is a mob, or a pet that isn't an adventuring fellow, or a trust

Does this drag in and apply changes to the other pet types?
```cpp
enum class PET_TYPE : uint8
{
    AVATAR             = 0,
    WYVERN             = 1,
    JUG_PET            = 2,
    CHARMED_MOB        = 3,
    AUTOMATON          = 4,
    ADVENTURING_FELLOW = 5,
    CHOCOBO            = 6,
    LUOPAN             = 7,
};
```

src/map/spell.cpp
```diff
@@ -696,8 +696,9 @@ namespace spell
                     return false;
                 }
 
-                // ensure trust level is appropriate+
-                if (PCaster->objtype == TYPE_TRUST && PCaster->GetMLevel() < static_cast<CMobEntity*>(PCaster)->m_SpellListContainer->GetSpellMinLevel(spell->getID()))
+                // Ensure pet or trust is level appropriate
+                if (((PCaster->objtype == TYPE_PET && static_cast<CPetEntity*>(PCaster)->getPetType() != PET_TYPE::AUTOMATON) || PCaster->objtype == TYPE_TRUST) &&
```
zach2good: Same down here, the logic is getting very crowded and hard to reason about. I'd recommend breaking a bunch of these out into more human readable flags at the top of this block (near `uint8 requirements = spell->getRequirements();`:
```cpp
const bool isPet = PCaster->objtype == TYPE_PET;
const bool isMob = PCaster->objtype == TYPE_MOB;
const bool isTrust =  PCaster->objtype == TYPE_TRUST;
const bool isPetButNotFellow = isPet  && static_cast<CPetEntity*>(PCaster)->getPetType() != PET_TYPE::ADVENTURING_FELLOW;
```

and then use those, stacked on different lines:
```cpp
if (isMob ||
    isPetButNotFellow ||
    isTrust)
{
    ....
```

zach2good:
Please look into how to use `git commit --amend` and `git push --force`, or `git rebase` to squash commits. 

`Removed 2 spaces in spell.cpp` isn't a useful commit and should be rolled into your original commit.

src/map/spell.cpp
DotesMiqotes: It does, it's also why I excluded adventuring fellows in particular. From what I could see they were the only pet type that can cast spells but didn't have a spell list in mob_spell_lists.sql.

src/map/spell.cpp
DotesMiqotes: Sounds good, I'll add these.

DotesMiqotes:
Understood, I'll go ahead and squash those into one commit, and use `git commit --amend` in the future for those minor formatting issues.

src/map/spell.cpp
DotesMiqotes: I have a copy of `spell.cpp` with those changes added and working. While I was looking over the commit, I realized that I could handle the different `objtype` cases with a switch. I got a second `spell.cpp` working with the switch statement, the change is drastic so I wanted to clear it with you before making the commit.
```c++
// Check If user can cast spell
bool CanUseSpell(CBattleEntity* PCaster, CSpell* spell)
{
    if (spell == nullptr)
    {
        return false;
    }

    bool usable = false;
    uint8 requirements = spell->getRequirements();

    switch(PCaster->objtype)
    {
        case TYPE_MOB:
            // Unable to cast because caster is hidden or untargetable
            if (PCaster->IsNameHidden() || static_cast<CMobEntity*>(PCaster)->GetUntargetable())
            {
                return false;
            }
            // Mobs can cast any non-given char spell
            return true;

        case TYPE_PC:
            if (spell->getSpellGroup() == SPELLGROUP_TRUST)
            {
                return true; // every PC can use trusts
            }
            else if (luautils::OnCanUseSpell(PCaster, spell))
            {
                return true;
            }
            [[fallthrough]];
        case TYPE_FELLOW:
        case TYPE_NPC:
        case TYPE_SHIP: // This can be removed it does technically follow current logic though
            // Make sure caster has the right main job and level
            if (PCaster->GetMLevel() >= spell->getJob(PCaster->GetMJob()))
```
Removed what goes between here because it's the current code.
```c++
        case TYPE_PET:
        case TYPE_TRUST:
            // Unable to cast because caster is hidden or untargetable
            if (PCaster->IsNameHidden() || static_cast<CMobEntity*>(PCaster)->GetUntargetable())
            {
                return false;
            }

            if (static_cast<CPetEntity*>(PCaster)->getPetType() == PET_TYPE::AUTOMATON)
            {
                return true;
            }

            // Ensure pet or trust is level appropriate
            if (PCaster->GetMLevel() < static_cast<CMobEntity*>(PCaster)->m_SpellListContainer->GetSpellMinLevel(spell->getID()))
            {
                return false;
            }
            return true;
```

zach2good (approved):

PR #6080, opened 2024-08-06
[Lua] Fix more lua errors from log

Xaver-DaRed:
**_I affirm:_**

Fixes multiple cases of wrong or non-existing enums

Lowers the error number from 122 to 66 ^^

None


scripts/globals/expansion_areas.lua
```diff
@@ -2,14 +2,14 @@ xi = xi or {}
 xi.expansionAreas = xi.expansionAreas or {}
 
 xi.expansionAreas.abyssea = set{
-    xi.region.ABYSSEA_KONSCHTAT,
```
claywar, line 5: I thought we had a separate region enum somewhere?

Xaver-DaRed: I had to guess those were zones, not regions. We have 1 Abyssea region in xi.region, called `ABYSSEA`
Meanwhile, those zones did exist

Xaver-DaRed: To expand:
This table is called by the file roe_records.lua, by certain table entries
`zoneNot = xi.expansionAreas.abyssea`

And then that table entry is checked in roe.lua
```lua
    zoneNot = function(self, player, params)  -- Player not in Zone
        return (not self.reqs.zoneNot[player:getZoneID()]) and true or false
    end,
```

So those are definately meant to be zones. With some minor tweaking, in a different PR, we could dispose of that file entirely

scripts/zones/RuLude_Gardens/npcs/Delivery_Crate.lua
```diff
@@ -7,18 +7,18 @@
...
 end
 
 entity.onEventUpdate = function(player, csid, option, npc)
-    xi.magian.deliveryCrateOnEventUpdate(player, csid, option, npc)
+    xi.magian.deliveryCrateOnEventUpdate(player, csid, option)
 end
 
 entity.onEventFinish = function(player, csid, option, npc)
-    xi.magian.deliveryCrateOnEventFinish(player, csid, option, npc)
+    xi.magian.deliveryCrateOnEventFinish(player, csid, option)
```
zach2good: I don't know what the others think, but I prefer the pattern where we just forward _all_ the args into the global function, instead of picking and choosing which ones we want.

So the change here would be going into xi.magian functions and making them take the npc as an argument, even if it isn't going to use it

zach2good (approved):

scripts/zones/RuLude_Gardens/npcs/Delivery_Crate.lua
Xaver-DaRed: I have no strong feelings one way or another. So whatever you guys decide works for me

claywar: I'm going to agree with Zach on this one.  Even though its unused, its one of those functions that I like seeing them all present

Xaver-DaRed: kk. Addressed

claywar (approved):

PR #6083, opened 2024-08-07
Wamouracampa Family

jmcmorris:
**_I affirm:_**

These changes focus on the Wamouracampa family introduced in the highly acclaimed expansion Treasures of Aht Urhgan released in 2006.

The first set of changes revolve around improving Wamouracampa mobskills.

The other changes implement the logic for Wamouracampa to evolve into Wamoura via eclosion. How this works is that we get all of the Wamoura in each zone that they are in and then in the wamouracampa family mixin the code looks to see if the mob is followed by a Wamoura. If it is then that means that the Wamouracampa can evolve into the Wamoura after being left alone for some amount of time. This works by the mob using the mobskill 'eclosion' which plays an animation and despawns the Wamouracampa and then spawns the Wamoura in its place.

## Retail Proof

[Jimmayus' Mob Sheet](https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=57955395#gid=57955395)
[Amber Scutum potency and duration capture](bd1f79e63e433162fcfd2a07538eee8290773656)
[Eclosion capture](https://discord.com/channels/443544205206355968/485290259404292096/1130203312516509727)

**NOTE:** Changes with Ranch Wamouracampa and Ranch Wamoura in Wamoura Farm Raid are untested since that assault is not fully implemented from what I can tell.

### Testing mobskills
`!gotoid 17027197`
`!exec target:useMobAbility(1815)`
`!exec target:useMobAbility(1816)`
`!exec target:useMobAbility(1817)`
`!exec target:useMobAbility(1818)`
`!exec target:useMobAbility(1819)`
`!exec target:useMobAbility(1820)`

### Testing Eclosion
`!gotoid 17027197`
Do not engage.
`!exec target:useMobAbility(1970)`
In order to test it naturally evolving you need to wait for 40-50 minutes or change the code to have a shorter eclosion time.
After the evolved Wamoura dies (`!hp 0`) the original Wamouracampa will respawn after about 6 minutes.

### DIFF
--- a/scripts/actions/mobskills/heat_barrier.lua
+++ b/scripts/actions/mobskills/heat_barrier.lua
@@ -13,9 +13,9 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    -- TODO: Enfire power, Blaze Spikes reduced power in Salvage zones
-    skill:setMsg(xi.mobskills.mobBuffMove(mob, xi.effect.BLAZE_SPIKES, math.random(50, 67), 0, 180))
-    -- xi.mobskills.mobBuffMove(mob, xi.effect.ENFIRE, ???, 0, 180)
+    -- TODO: Blaze Spikes reduced power in Salvage zones
+    skill:setMsg(xi.mobskills.mobBuffMove(mob, xi.effect.BLAZE_SPIKES, mob:getMainLvl() * 0.8, 0, 180))
+    xi.mobskills.mobBuffMove(mob, xi.effect.ENFIRE, mob:getMainLvl() * 0.4, 0, 300)
 
     return xi.effect.BLAZE_SPIKES
 end
--- a/scripts/enum/mob_skills.lua
+++ b/scripts/enum/mob_skills.lua
@@ -63,6 +63,8 @@ xi.mobSkill =
 
     RANGED_ATTACK_15  = 1949,
 
+    ECLOSION          = 1970,
+
     ROAR_3            = 2406,
 
     SHEEP_SONG_3      = 3433,
--- a/scripts/actions/mobskills/amber_scutum.lua
+++ b/scripts/actions/mobskills/amber_scutum.lua
@@ -13,15 +13,9 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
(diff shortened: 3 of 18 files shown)


zach2good (approved):

scripts/actions/mobskills/heat_barrier.lua
```diff
@@ -13,9 +13,9 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    -- TODO: Enfire power, Blaze Spikes reduced power in Salvage zones
-    skill:setMsg(xi.mobskills.mobBuffMove(mob, xi.effect.BLAZE_SPIKES, math.random(50, 67), 0, 180))
-    -- xi.mobskills.mobBuffMove(mob, xi.effect.ENFIRE, ???, 0, 180)
+    -- TODO: Blaze Spikes reduced power in Salvage zones
+    skill:setMsg(xi.mobskills.mobBuffMove(mob, xi.effect.BLAZE_SPIKES, mob:getMainLvl() * 0.8, 0, 180))
+    xi.mobskills.mobBuffMove(mob, xi.effect.ENFIRE, mob:getMainLvl() * 0.4, 0, 300)
```
Xaver-DaRed, line 18: Shouldnt this operations be floored? I dont think we want floats in power

scripts/enum/mob_skills.lua
```diff
@@ -63,6 +63,8 @@ xi.mobSkill =
 
     RANGED_ATTACK_15  = 1949,
 
+    ECLOSION          = 1970,
```
Xaver-DaRed, line 66: :heart:

scripts/actions/mobskills/heat_barrier.lua
jmcmorris: I thought that might have been the case as well but we actually read it as a double and then cast it to uint16 which essentially floors it. So this is technically okay but I'm happy to floor it here as well.

https://github.com/LandSandBoat/server/blob/base/src/map/lua/lua_baseentity.cpp#L12781

Xaver-DaRed (approved):

scripts/actions/mobskills/heat_barrier.lua
Xaver-DaRed: Nah, its ok. If it gets floored down the line, I'm happy

PR #6087, opened 2024-08-07
Trusts: Iroha and Iroha II Spell Implementation

DotesMiqotes:
**_I affirm:_**

Adds spell gambits for Iroha and Iroha II.
Adds `MPP` mods for both, and gives Iroha II `BLACK_MAGIC_CAST` -85 for her "near instacast" Flare II.
Removes lower tier versions of Protectra and Shellra from both spell lists.
Adds TODO messages for everything that still needs to be implemented.

I play tested the changes made to Iroha and Iroha II in a party with Fablinix and Moogle.

### DIFF
--- a/scripts/actions/spells/trust/iroha.lua
+++ b/scripts/actions/spells/trust/iroha.lua
@@ -13,6 +13,23 @@ end
 
 spellObject.onMobSpawn = function(mob)
     xi.trust.message(mob, xi.trust.messageOffset.SPAWN)
+
+    -- TODO: Weaponskills
+    --       - Uncomment entries in mob_pool_mods.sql
+    --       - Needs Self Skillchain logic - if Meditate and Hagakure are available and 2000 TP
+    --       - Self Skillchain = Hanadoki > Choun > Fuga > Gachirin > Gachirin
+    --       - Needs entries in mob_skill_lists.sql
+    --       - Needs skillchain properties in mob_skills.sql
+    --       - Needs Weaponskill LUAs
+    --       - Holds 2500 TP To close Skillchains
+    -- TODO: Abilities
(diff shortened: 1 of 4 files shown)


zach2good:
I'd want some big fat TODO comments in these scripts; saying that this is not the full implementation for these Trusts. List the things that are missing, please.

Things that are only partially implemented tend to have a much lower chance of ever being finished than things that are completely missing. The "build it and they will come" initiative for giving people the tools to implement Trusts has failed - we may as well start accepting things piecemeal (not applicable to anything apart from Trusts, I'm not opening a free-for-all for people to be lazy).

DotesMiqotes:
I can add those, and I understand. Implementing everything trust related is actually fairly high on my TODO list (currently working on Aldo). I have a little bit more done for these two trusts than what's here, nothing that I feel is 100% ready for a PR though.

It would actually be easier for me to implement trust features all at once, instead of trust by trust. The order would be something along the lines of teamwork messages > spells > mods (would include getting store tp working properly for trusts) > abilities > weaponskills > teamwork interactions.

zach2good:
That may be easier for you, but we've witnessed these points many times over over the course of time:
- Things that are partially complete will be misconstrued as complete, and then never be revisited and actually completed.
- Contributors will start large bodies of work with the best of intentions, and then disappear for whatever reason - despite their most steadfast promises that they wouldn't do so. This leads to the first point.

So these things being said: for example, instead of submitting all missing Trust friendship texts at once etc., please submit each Trust as completely as you can. It'll also mean they get better and closer reviews.

DotesMiqotes:
In that case, I'll keep the partially completed trusts [here](https://github.com/DotesMiqotes/server/tree/trusts). My main concern is saving work for anyone that can finish implementing a trust before me.

This spell implementation for Iroha I and II is mostly an extension of the work I was doing with trust spells in general, as well as fixing the minor issue of her potentially having access to lower tier protectra and shellra when she shouldn't.

DotesMiqotes:
#6122

PR #6088, opened 2024-08-07
[Digging] Audit data tables

Xaver-DaRed:
**_I affirm:_**

It audits the data tables with the info from https://www.bg-wiki.com/ffxi/Category:Chocobo_Digging for available items and rates.
Still takes the info from xiclopedia for BORE and BURROW exclusive items, which arent implemented.

Translates all item IDs to enums 

This is supposed to be the preamble for a rewrite. I decided to do it in 2 parts so to not clutter the next one (the more important one)

NOTE: I took the data from there as it was. This resulted in an overall lower rate for most items. Some very few items got lucky.
NOTE 2: For BURROW and BORE information, when in doubt, it was done against players.
NOTE 3: This will temporarily disable the ability to dig eggs when wearing the egg helm. It will be reintroduced in the next PR.

Logic is mostly the same


claywar (approved):
Nice work, and fine with the changes.  The only thing that I'd like others to chime in with is the use of ints for the `NONE` type, to not be confused with `BORE` when perusing the list.

zach2good:
I'm not sure I like `0` being the representative for no reqs. It's not as short, but `NONE` could be `NO_REQS` and then it's visually distinct from `BORE`

Also, comments like `-- Fire Crystal` can now be removed

Otherwise looks good 👍

Xaver-DaRed:
I've done some more digging, no pun intended, and it turns out the BURROW and BORE abilities open up new independent "strata" layers. What this means is that, if you have bore, for instance, you can get 1 item from the regular pool AND an additional item from the BORE pool/strata.

So those requirements will dissapear in my next PR and instead, each zone will have 3 different pools defined (regular, burrow and bore pools) so this can happen:
imagen

Xaver-DaRed:
Changed `NONE` to `NO_REQS` and replaced all 0s with it. Also deleted unnecesary comments

zach2good (approved):

PR #6090, opened 2024-08-08
[Digging] Part 2. Clean core and implement fatigue properly

Xaver-DaRed:
**_I affirm:_**

- Updates core logic so we dont have the need to hit the global file TWICE per digging attempt
- Implements the fatigue mechanic properly.
  - It now respects the value in stting instead of it being hardcoded
  - Uses an expirable variable to track daily digs
  - It now doesnt prevent you from digging when fatigued, as retail does.
- Implements positional check to prevent repeatedy digging without moving.

Dig normaly

To review:
- Review by commit

Commit 1:
- Ignore the zone files. That way its simpler.
- What was done is removing an extra parameter. The way it was coded, this was the workflow:
  - Call global lua function first time, with arg being true.
  - Call global lua function a second time, with arg being false, IF it returned `true` the first time
- This was... I dunnow, not very good. So the logic was changed so the global lua function is only called once per attempt.

Commit 2:
- Handles the main setting for fatigue. changing it from being used as a boolean to being a proper int defining max daily digs.
- Reorders code to hit as less code as necesary
- Compensates deleted code in core, adding some messages.

Commit 3:
- In retail you must move 2+ yalms to be able to dig successfully again. So this implements that.

Commit 4:
- Once again, ignore all zone files.
- We make a change in core to call the global function, that was being called in lua in all touched zone scripts, directly.
- We add a table of allowed zones to dig in and check first thing, to guarantee we wont dig in undefined zones.

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -5316,20 +5316,22 @@ namespace luautils
         }
     }
 
-    bool OnChocoboDig(CCharEntity* PChar, bool pre)
+    bool OnChocoboDig(CCharEntity* PChar)
     {
         TracyZoneScoped;
 
-        auto name = PChar->loc.zone->getName();
-
+        // Check if zone can be digged in.
+        auto name         = PChar->loc.zone->getName();
         auto onChocoboDig = lua["xi"]["zones"][name]["Zone"]["onChocoboDig"];
+
         if (!onChocoboDig.valid())
         {
             ShowWarning("luautils::onChocoboDig");
             return false;
         }
 
-        auto result = onChocoboDig(CLuaBaseEntity(PChar), pre);
+        // Run lua-side chocobo digging.
+        auto result = onChocoboDig(CLuaBaseEntity(PChar));
         if (!result.valid())
         {
             sol::error err = result;
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -5320,10 +5320,9 @@ namespace luautils
     {
         TracyZoneScoped;
 
-        // Check if zone can be digged in.
-        auto name         = PChar->loc.zone->getName();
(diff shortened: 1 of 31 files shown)


Xaver-DaRed:
Tested it for a few hours without errors. Ready for review.

src/map/lua/luautils.cpp
```diff
@@ -5316,20 +5316,22 @@ namespace luautils
...
-    bool OnChocoboDig(CCharEntity* PChar, bool pre)
+    bool OnChocoboDig(CCharEntity* PChar)
     {
         TracyZoneScoped;
 
-        auto name = PChar->loc.zone->getName();
-
+        // Check if zone can be digged in.
+        auto name         = PChar->loc.zone->getName();
         auto onChocoboDig = lua["xi"]["zones"][name]["Zone"]["onChocoboDig"];
```
zach2good: Instead of modifying every zone where digging could happen, would it be better/easier to just have a table in the chocobo digging global that lists all the zones where it's possible? Then in `onChocoboDig` you can check to see if the player is in that zone - then you're done, no extra changes required apart from in the global - also easier to modify if people want that

zach2good (approved):
Can't say I know anything about chocobo digging, this is in your hands 🙏

src/map/lua/luautils.cpp
Xaver-DaRed: I thought about doing that, however, I didnt want to clutter the PR radically changing how everything works and instead decided to focus on first cleaning up obviously wrong logic.

If you think I should just call the global directly and add a set table to the global to check allowed zones, I can do that.
That would involve some more core edits, couse that would make the zone function useless.

zach2good: I would say get rid of the onChocoboDig() functions in all the zone files, we're gearing up to do this sort of transformation with zone mixins too

zach2good (approved):

PR #6093, opened 2024-08-09
Apkallu family

jmcmorris:
**_I affirm:_**

These changes are focused on making the Apkallu mob family more retail accurate. This includes the following:

1. Introducing Apkallu communal hate system
2. Adjusting mobskills to deal appropriate damage and status effects
3. Mobskills are only available at certain communal hate thresholds
4. Apkallu NPCs that exist when communal hate is extremely low. You can trade fish to them and they will follow you.
5. Additional changes to make the previously mentioned changes possible

Apkallu communal hate works per zone. As Apkallu are killed it increases the hate slightly. Each game day lowers the communal hate some. As communal hate rises it unlocks new mobskills for the Apkallu and changes their behavior to be more aggressive. These changes happen at set thresholds.

### Mobskills
0-10: Yawn mobskill only
10+: Beak Lunge unlocked
20+: Wing slap unlocked
30+: Frigid Shuffle unlocked
40+: Wing Whirl unlocked

### Behavior
0-5: NPC Apkallu
25+: Aggro by sight and damage reduction based on hate level
45+: Links and will run away at critical health

## Retail Proof

Most of the research on the communal hate was done years ago by Meloetta and myself for Eden. Unfortunately I think this was done purely with notes and not a capture. It's been a long time...

[Apkallu Wiki](https://ffxiclopedia.fandom.com/wiki/Category:Apkallu)
[Apkallu NPC Interactions Capture](https://discord.com/channels/443544205206355968/443894311881408522/1262474594128891957)
[Jimmayus' Mobs Sheet](https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=57955395#gid=57955395)

`!gotoid 16998410`

Play around with the NPC Apkallu such as giving it a fish for it to follow you (`!additem 5463`)
Kill 5 Apkallu to see the NPC turn into a normal mob.
Look at the various mobskill hate thresholds. Mobs should only use that mobskill once the hate is greater than or equal to the threshold. You can directly modify the `ApkalluHate_{zoneid}` server variable to speed up testing.

### DIFF
--- a/scripts/actions/mobskills/frigid_shuffle.lua
+++ b/scripts/actions/mobskills/frigid_shuffle.lua
@@ -8,11 +8,17 @@
 local mobskillObject = {}
 
 mobskillObject.onMobSkillCheck = function(target, mob, skill)
-    return 0
+    return xi.apkallu.canUseAbility(mob, 30)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    skill:setMsg(xi.mobskills.mobStatusEffectMove(mob, target, xi.effect.PARALYSIS, 50, 0, 60))
+    if not target:isFacing(mob) then
+        skill:setMsg(xi.msg.basic.SKILL_MISS)
+        return
+    end
+
+    local duration = math.random(60, 120)
+    skill:setMsg(xi.mobskills.mobStatusEffectMove(mob, target, xi.effect.PARALYSIS, 50, 0, duration))
 
     return xi.effect.PARALYSIS
 end
--- a/scripts/globals/apkallu.lua
+++ b/scripts/globals/apkallu.lua
@@ -1,0 +1,147 @@
+xi = xi or {}
+xi.apkallu = xi.apkallu or {}
+xi.apkallu.zones = xi.apkallu.zones or {}
+
+xi.apkallu.fish =
+{
+    xi.item.AHTAPOT,
+    xi.item.VEYDAL_WRASSE,
+    xi.item.YAYINBALIGI,
+}
+
+xi.apkallu.track = function(mob)
+    local zoneID = mob:getZoneID()
+    xi.apkallu.zones[zoneID] = xi.apkallu.zones[zoneID] or {}
+    table.insert(xi.apkallu.zones[zoneID], mob)
+end
+
+xi.apkallu.initialize = function(mob)
+    local zoneID = mob:getZoneID()
+    local ID     = zones[zoneID]
+    local hate   = GetServerVariable('ApkalluHate_'..zoneID)
+    local tier   = xi.apkallu.getHateTier(hate)
+
+    if tier == 0 and mob:getID() == ID.mob.APKALLU_NPC then
+        -- Convert this Apkallu into an NPC
+        mob:setStatus(xi.status.NORMAL)
+    elseif tier == 2 then
+        -- Apkallu now aggro by sight
+        mob:setAggressive(true)
+    end
+
+    if tier ~= 3 then
+        -- Apkallu should not link
+        mob:setLink(0)
+    end
+
+    if hate > 25 then
+        local reduction = hate - 25
+        mob:setMod(xi.mod.DMGPHYS, -reduction)
+        mob:setMod(xi.mod.DMGRANGE, -reduction)
+        mob:setMod(xi.mod.DMGMAGIC, -reduction)
+    end
+end
+
+xi.apkallu.updateHate = function(zoneID, amount)
+    local ID = zones[zoneID]
+    local previousHate = GetServerVariable('ApkalluHate_'..zoneID)
+    local hate = previousHate + amount
+    local shouldCheckTiers = true
+    if hate > 100 then
+        hate = 100
+        shouldCheckTiers = false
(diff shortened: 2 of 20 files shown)


scripts/globals/apkallu.lua
```diff
@@ -0,0 +1,143 @@
...
+end
+
+xi.apkallu.canUseAbility = function(mob, threshold)
+    if mob:isNM() then
+        return 0
+    end
+
+    local zoneID = mob:getZoneID()
+    if
+        (zoneID == 54 or zoneID == 61) and
```
Xaver-DaRed: `(zoneID == xi.zone.ARRAPAGO_REEF or zoneID == xi.zone.MOUNT_ZHAYOLM) and`

scripts/globals/apkallu.lua
```diff
@@ -0,0 +1,143 @@
+xi = xi or {}
+xi.apkallu = xi.apkallu or {}
+xi.apkallu.zones = xi.apkallu.zones or {}
```
Xaver-DaRed, line 3: this last one goes unused, no?

scripts/mixins/families/apkallu.lua
```diff
@@ -0,0 +1,51 @@
...
+        end
+    end)
+
+    apkalluMob:addListener('RUN_AWAY', 'APKALLU_RUN_AWAY', function(mob, followed)
+        followed:updateEnmity(mob:getTarget())
+    end)
+
+    apkalluMob:addListener('DEATH', 'APKALLU_DEATH', function(mob)
+        xi.apkallu.updateHate(mob:getZoneID(), 1)
+    end)
```
Xaver-DaRed, line 50: Wont this be called once per player fighting it? That function is large and we only want it to run once.

Maybe what we want to do is add a step onMobDeath to ensure its only run once:
```lua
if optPrarams.isKiller ot optParams.noKiller then
    xi.apkallu.updateHate(mob:getZoneID(), 1)
end
```

jmcmorris: Fortunately no, the "onMobDeath" function is called per character and its corresponding callback is "DEFEATED_MOB". The "DEATH" callback is called from CAIContainer::PostTick() and so it is only called once.

scripts/globals/apkallu.lua
jmcmorris: It does get used. I just made a change to the code some but it is used to track the Apkallu mobs per zone.

scripts/actions/mobskills/frigid_shuffle.lua
```diff
@@ -8,11 +8,17 @@
...
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    skill:setMsg(xi.mobskills.mobStatusEffectMove(mob, target, xi.effect.PARALYSIS, 50, 0, 60))
+    if not target:isFacing(mob) then
+        skill:setMsg(xi.msg.basic.SKILL_MISS)
+        return
+    end
+
+    local duration = xi.mobskills.calculateDuration(skill:getTP(), 60, 120)
```
Frankie-hz: Is there a reason we are assuming duration always changes with varying TP?

jmcmorris: I think there are two other options:

1. Purely random
2. It can be resisted

Having duration scale with TP makes more sense design wise than randomness. None of the wikis mention this nor does Jimmayus' mob spreadsheet. I think it'd just require lots more retail testing which is a little hard because you can't know the mobs actual TP (unless you can charm them). Maybe monstrosity would be able to tell us?

Frankie-hz: I don't think any durations are purely random other than the bind research tracent did. Would be the first time I am hearing about that. I thought it was just a resist check. I just have a feeling we are going to end up removing all of these at some point.

You would have to get it to TP above 50% (at 3000 tp) and see durations. Then get it below 25% for 1k tp and see if the durations are longer/shorter/same.

If the duration varies with TP it looks like it was noted during his research for monstrosity (see example) 
image

jmcmorris: I see. Perhaps I need to talk to Jimmayus about this then to better understand the duration range. I don't understand why he would provide a range if the lower bound is simply a resist value. There must be a reason though.
image

Frankie-hz: If its a half resist check it still lands but the duration is halved. I would follow up with him before you spend all this time doing the calc duration stuff and it has to be re-done. He might have an answer for you immediately

jmcmorris: After talking to Jimmayus I believe the proper solution here is to use `math.random`. I want to do some thorough testing of at least some mobskill debuff durations and revise all places I've used the TP to calculate the duration but that is outside the scope of this PR.

Frankie-hz: lgtm!

Xaver-DaRed (approved):

PR #6095, opened 2024-08-10
Orobon family

jmcmorris:
**_I affirm:_**

These changes focus on the Orobon family. This does the following:

- Adds all of the missing Orobon mobskills except for Mayhem Lantern which is from Abyssea era
- Implements mob family logic for breaking the eyestalks and getting Orobon Lure as loot

## Retail Proof

[Orobon Wiki](https://ffxiclopedia.fandom.com/wiki/Category:Orobon)
[Jimmayus' Mob Spreadsheet](https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=57955395#gid=57955395)
[Giant Orobon Capture 1](https://www.youtube.com/watch?v=3AiUOr3cRWE)
[Giant Orobon Capture 2](https://www.youtube.com/watch?v=YvJNyHEbzqo)
[Angler Orobon (ISNM) Capture](https://www.youtube.com/watch?v=D-aiHPGJRYc)

`!zone Arrapago Reef`
`!spawnmob 16998406`
`!mobhere 16998406`

Engage the Orobon and use the following:
`!exec target:useMobAbility(1693)`
`!exec target:useMobAbility(1694)`
`!exec target:useMobAbility(1695)`
`!exec target:useMobAbility(1696)`
`!exec target:useMobAbility(1697)`
`!exec target:useMobAbility(1698)`
`!exec target:useMobAbility(1977)`
`!exec target:useMobAbility(1978)`

With `!tp 3000` you can test to make sure it only uses Hypnic Lamp while it has its eyestalks. Destroy the eyestalks by doing critical hits. Orobon Lure should only drop as loot when the eyestalks are destroyed.


scripts/actions/mobskills/leeching_current.lua
```diff
@@ -0,0 +1,27 @@
...
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    -- We use a var here so that mobs can use different values
+    local base = mob:getLocalVar('leechingCurrent')
+    if base == 0 then
+        base = 1000
+    end
+
+    local damage = base / skill:getTotalTargets()
```
Xaver-DaRed, line 19: Chances of this happening are probably minimal or non-existing, but...

Just in case, we should clamp this to ensure we never divide by 0.
Something like...

`local damage = base / utils.clamp(skill:getTotalTargets(), 1, 18)`

Xaver-DaRed (approved):

scripts/actions/mobskills/leeching_current.lua
jmcmorris: This is the same as all of the other skills like 1000 Needles. How could the total targets be 0?

If the target dies or goes out of range then the action would be cancelled. Also if players are fighting in an alliance and then leave the alliance but still have hate then I believe they'd still be able to be targeted by this. In that case limiting the total targets to 18 would be incorrect, right?

zach2good (approved):

PR #6102, opened 2024-08-12
[Quest] Convert (Beastmaster AF2) Scattered into Shadow to IF

jamesbradleym:
**_I affirm:_**

Converts Beast AF2 Quest (Scattered into Shadow) to IF

[Capture](https://www.youtube.com/watch?v=4De8MhSzMjg)

Complete Beast AF2 Quest


scripts/zones/Castle_Oztroja/IDs.lua
```diff
@@ -32,6 +32,7 @@ zones[xi.zone.CASTLE_OZTROJA] =
         MEMBERS_LEVELS_ARE_RESTRICTED      = 7208, -- Your party is unable to participate because certain members' levels are restricted.
         FISHING_MESSAGE_OFFSET             = 7268, -- You can't fish here.
         CHEST_UNLOCKED                     = 7439, -- You unlock the chest!
+        TEBHI_ACCEPTS                      = 7450, -- Tebhi happily accepts the
```
zach2good: This will break our update tooling and require manual intervention. We match on the comment, so the comment has to be exactly in line with what our update tooling spits out.

These client extracts are old (ignore the numbers), but they show you what the text _should_ be:

https://github.com/LandSandBoat/UpdateExtractor/blob/main/out/scripts/zones/Castle_Oztroja/Text.lua.raw.txt#L7362

We use `<>` for wildcard while doing these checks, so the comment should be: `Tebhi happily accepts the <item> and leaves.`

scripts/quests/jeuno/Scattered_into_Shadow.lua
```diff
@@ -0,0 +1,204 @@
...
+            {
+                onTrade = function(player, npc, trade)
+                    if
+                        npc:getStatus() == xi.status.NORMAL and
+                        npcUtil.tradeHasExactly(trade, xi.item.BEAST_COLLAR)
+                    then
+                        player:tradeComplete()
+                        quest:setVar(player, 'Prog', 5)
+                        npc:setStatus(xi.status.DISAPPEAR)
+                        npc:updateNPCHideTime(900) -- Tebhi disappears for 15min
```
zach2good, line 150: 👍

scripts/quests/jeuno/Scattered_into_Shadow.lua
```diff
@@ -0,0 +1,204 @@
...
+
+            onEventFinish =
+            {
+                [18] = function(player, csid, option, npc)
+                    player:delKeyItem(xi.ki.AQUAFLORA3)
+                    quest:setVar(player, 'Prog', 3)
+                end,
+
+                [20] = function(player, csid, option, npc)
+                    player:delKeyItem(xi.ki.AQUAFLORA2)
```
zach2good, line 129: For these various add/del key items, is this messaged in the CS that's just finished, or should these handle their own messaging? What does the cap show?

scripts/zones/Castle_Oztroja/IDs.lua
jamesbradleym: Didn't know about this!! Thanks 🙏🏼

scripts/quests/jeuno/Scattered_into_Shadow.lua
jamesbradleym: The CS/event for each of the flora removals delivers the message to the player, however I missed the message when giving the player the aquaflora from Brutus. Added!

jamesbradleym:

PR #6109, opened 2024-08-15
Core: Check player meets skill requirement to use weaponskills

loxleygames:
**_I affirm:_**

Currently if a player can use a weaponskill from the menu, the server just allows it without checking their skill level. In certain cases where the menu isn't refreshed, this allows players to use weaponskills without meeting the requirements (And presumably could be completely bypassed with packet injection).

```
!changejob war 76
!additem fortitude_torque
```
Without Fortitude Torque, you will not have access to Fell Cleave (Without merits). Once Fortitude Torque is equipped, you will be able to use Fell Cleave. Currently, you can remove the Torque and continue to use Fell Cleave. With this fix, you can no longer continue to use it once the requirements are no longer met.

### DIFF
--- a/src/map/ai/controllers/player_controller.cpp
+++ b/src/map/ai/controllers/player_controller.cpp
@@ -197,7 +197,7 @@ bool CPlayerController::WeaponSkill(uint16 targid, uint16 wsid)
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


WinterSolstice8 (approved):
Thanks

zach2good (approved):

PR #6111, opened 2024-08-16
Core: Adds region check for Ionis and fixes conquest latent error

loxleygames:
**_I affirm:_**

Currently, logs are spammed with the error `Invalid conquest region passed to function` under the following conditions:
* Any mob killed in Adoulin regions, regardless of status effects
* When "conquest" latent effects activate outside the regular Conquest areas eg. Dynamis

Items with the "In areas outside own nation's control" latent  effect should trigger (And currently do) in certain regions even though they are not subject to conquest, but it still produces an error.

This PR fixes the issues by changing the following:
* Region ID must be outside of valid latent areas before `getRegionOwner` produces the error message (Extra permitted areas will simply return 0 - No owner)
* Adds a check for Conquest regions before simply passing off any mob kill to Signet (Which will produce errors in unhandled regions)
* Implemented Ionis effect requirement to receive crystals in Adoulin regions
* Added some extra padding and reuse variable to reduce extra lookups

Tested in-game using multiple different region types and conditions. Conquest area with/without control, Lumoria, Adoulin areas with/without Ionis, etc. Re-tested with Winter's rewrite of `ConquestData::getRegionOwner`

### DIFF
--- a/src/map/conquest_data.cpp
+++ b/src/map/conquest_data.cpp
@@ -90,7 +90,12 @@ uint8 ConquestData::getRegionOwner(REGION_TYPE region) const
         return regionControls[regionNum].current;
     }
 
-    ShowError(fmt::format("Invalid conquest region passed to function ({})", regionNum));
+    // This can be called for Latent effects that work outside of conquest areas eg. Dynamis/Limbus
+    if (regionNum > (uint8)REGION_TYPE::LIMBUS)
+    {
+        ShowError(fmt::format("Invalid conquest region passed to function ({})", regionNum));
+    }
+
     return 0;
 }
 
--- a/src/map/conquest_data.cpp
+++ b/src/map/conquest_data.cpp
@@ -85,17 +85,28 @@ uint8 ConquestData::getRegionOwner(REGION_TYPE region) const
 {
     uint8 regionNum = static_cast<uint8>(region);
 
-    if (regionNum < regionControls.size())
+    // Handle some conquest regions that don't have conquest info as non-error
+    // TODO: Do Sandoria/Bastok/Windurst count as "owned by" themselves, no one, or some other state where latents don't work?
+    switch (region)
     {
-        return regionControls[regionNum].current;
+        case REGION_TYPE::SANDORIA:
+        case REGION_TYPE::BASTOK:
+        case REGION_TYPE::WINDURST:
+        case REGION_TYPE::JEUNO:
+        case REGION_TYPE::DYNAMIS:
+        case REGION_TYPE::TAVNAZIAN_MARQ:
+        case REGION_TYPE::PROMYVION:
+        case REGION_TYPE::LUMORIA:
+        case REGION_TYPE::LIMBUS:
+            return NATION_TYPE::NATION_BEASTMEN;
+        default:
+            if (regionNum < regionControls.size())
+            {
+                return regionControls[regionNum].current;
+            }
     }
 
-    // This can be called for Latent effects that work outside of conquest areas eg. Dynamis/Limbus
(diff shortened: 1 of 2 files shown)


zach2good (approved):

src/map/conquest_data.cpp
```diff
@@ -90,7 +90,12 @@ uint8 ConquestData::getRegionOwner(REGION_TYPE region) const
         return regionControls[regionNum].current;
     }
 
-    ShowError(fmt::format("Invalid conquest region passed to function ({})", regionNum));
+    // This can be called for Latent effects that work outside of conquest areas eg. Dynamis/Limbus
+    if (regionNum > (uint8)REGION_TYPE::LIMBUS)
```
claywar: This appears to just be suppressing a warning but still returning 0.  Can you guarantee latents will work in this case?

loxleygames: Yes, the conditions are different, regionControls.size() is 18 (Tavnazia is the last one) but the Latent for "Areas outside of nation's control" (Expects a 0) will continue to work up to 27 (Limbus). The error is meaningless for 19-27, as it is the expected use case.

loxleygames: @claywar Are you looking for this check to be reverted before this can be merged? Just let me know, so I understand how to proceed.

loxleygames: Please let me know what you want me to do with this PR.

WinterSolstice8: this is what I thought up of. There's some "valid" conquest regions that are outside of the conquest data but can still activate latents

```c++
uint8 ConquestData::getRegionOwner(REGION_TYPE region) const
{
    uint8 regionNum = static_cast<uint8>(region);

    // Handle some conquest regions that don't have conquest info as non-error
    // TODO: Do Sandoria/Bastok/Windurst count as "owned by" themselves, no one, or some other state where latents don't work?
    switch (region)
    {
        case REGION_TYPE::SANDORIA:
        case REGION_TYPE::BASTOK:
        case REGION_TYPE::WINDURST:
        case REGION_TYPE::JEUNO:
        case REGION_TYPE::DYNAMIS:
        case REGION_TYPE::TAVNAZIAN_MARQ:
        case REGION_TYPE::PROMYVION:
        case REGION_TYPE::LUMORIA:
        case REGION_TYPE::LIMBUS:
            return NATION_TYPE::NATION_BEASTMEN;
        default:
            if (regionNum < regionControls.size())
            {
                return regionControls[regionNum].current;
            }
    }

    ShowError(fmt::format("Invalid conquest region passed to function ({})", regionNum));
    return 0;
}
```

WinterSolstice8: Updated my comment, I double checked why tavnazia is listed twice and one was a legitimate region with conquest data

loxleygames: Thanks for your suggestions, I appreciate you spending time looking into this. Sorry, I've wanted to finish it but also want to make sure I take the time to properly test it and I've been very busy the last week. Will update soon.

LoxleyX:
Any chance of getting this re-reviewed please? It has been 2 months.

zach2good:
Last interaction I saw was this:
```
[LoxleyXI](https://github.com/LoxleyXI) [2 weeks ago](https://github.com/LandSandBoat/server/pull/6111#discussion_r1784742098)
Thanks for your suggestions, I appreciate you spending time looking into this. Sorry, I've wanted to finish it but also want to make sure I take the time to properly test it and I've been very busy the last week. Will update soon.
```

Any update on that testing?

LoxleyX:
It has been retested along with the last commit (I added this to the PR itself)

WinterSolstice8 (approved):

PR #6115, opened 2024-08-18
[core] Attempt to improve session cleanup logic

### LINKED ISSUE #6106: 🐛 Erroneous clearing of an account session
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

Got a fun one here.

Exact steps to reproduce:
- two map servers with zones split across them
- have an account with two chars, one in a zone on each map server
- log into char 1.
  - The map server it's on we'll call map server 1
- log out
- repeat previous two steps until you get a message like so: `[error] map_encipher_packet: bad packet from <127.0.0.1> (map_decipher_packet:640)`
  - A pretty reliable way to trigger this condition is to be a GM (so you instant log out) and use an item just before you `/logout`
  - `!additem airborne 12`
  - `/item airborne` then immediately `/logout`
- This causes a session on map server 1 not tied to a PChar. It will be cleaned on the next `map_cleanup`
- log into char 2 on map server 2 before the cleanup happens
- when map server 1 does a cleanup, it runs: `DELETE FROM accounts_sessions WHERE client_addr = %u AND client_port = %u`
- this deletes the accounts_session entry for char 2, since the ffxi client uses the same source port as char 1 when it logs into char 2

screenshot with the two chars replicating the scenario:
image

## Expected behavior

`accounts_sessions` entries should not be removed for unrelated, active characters

WinterSolstice8:
**_I affirm:_**

Attempts to fix #6106 
Shifts some logic around and stores charID in session instead of on PChar to make most logic similar.
Removes some weird `zoning == 0` check for session cleanup. Doesn't make sense especially if PChar is null.

See #6106 
Logout, login, force d/c, /terminate, etc etc and see sessions still get cleaned up

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -189,7 +189,10 @@ int32 do_init(int32 argc, char** argv)
 #endif // TRACY_ENABLE
 
     ShowInfo("do_init: begin server initialization");
+
+    // These aren't set unless --ip or --port is set, respectively.
     map_ip.s_addr = 0;
+    map_port      = 0;
 
     for (int i = 1; i < argc; i++)
     {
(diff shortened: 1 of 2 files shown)


WinterSolstice8:
@MowFord I couldn't quite reproduce, but I saw some logic flaws/improvements I could make. Can you try to repro with this branch?

MowFord:
image

Had a helluva a time getting the decipher packet failure to trigger but once i did it did in fact not clear my other char session like it was doing in the issue. Looks like it's fixed!

zach2good (approved):

WinterSolstice8:
> image
> 
> Had a helluva a time getting the decipher packet failure to trigger but once i did it did in fact not clear my other char session like it was doing in the issue. Looks like it's fixed!

Sweet! Thanks for confirming. PR is undrafted

PR #6117, opened 2024-08-18
Core: Long-distance/Chunked pathing support

### LINKED ISSUE #3522: 🔨 Long distance pathing support
**_I affirm:_**

## Describe the feature

I was chit-chatting with the Horizon folks about how long-distance pathing could work for future content (Besieged and Campaign both have mobs running clear across zones before battles start). 

I like the pattern we use for TriggerAreas, and I think that could be a good way to handle paths.

```
I'd probably still be happier if we registered long distance paths in zones like we do with trigger areas, then we assign mobs to that path, so it'll handle rejoining it etc. in the core logic.
```
```
onZoneInit:
    zone:registerLongDistancePath(id, x1, y1, z1, x2, y2, z2, opts...)

onMobSpawn:
   mob:assignLongDistancePath(id)


-- Internally
    If mob isn't doing anything else, work through the path nodes for assigned long path
    If mob is distracted, find the next point on the long path after they were distracted and start pathing from there
    If mob dies or reaches the end of the path (with radius), disconnect from path.
```
```
Those IDs can be enum'd in the campaign/besieged helpers
```
```
I think this system is sounding promising, and isn't much more than some if-checks pushed into the current code
options for jitter and randomness and spread between points, etc.
Hmmm, could get expensive if you want to make it safe
because
generate base path
for each point in base path: find random point nearby, add to new list
for each point in new list: stitch them together with findPath
any non-minor adjustment to base path could result in wonk
```
**NOTE:** I haven't measured to see if long-distance pathing is super expensive in about 2 years, so we might not need to do any of this.

Another fun thing to look at (though, almost certainly overkill) is DetourCrowd.

zach2good:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/3522

Allows mobs to navigate a zone's entire length from a single, non-specialised call to our existing pathing functions - without tanking performance and without them resorting to wallhacking. For use in Besieged, Campaign, etc.

### Work

- [x] Fafnir can long-distance path the whole length of Bhaflau Thickets without resorting to wallhacking or getting trapped in the tunnel local minima.
- [x] Improve Tracy plumbing for more data: https://github.com/LandSandBoat/server/pull/6118
- [ ] Lots of testing on all types of mobs and path, and NPC pathing
- [ ] Abuse testing
- [ ] Enhance the debug logging in PathFind and NavMesh
- [ ] Performance profiling of different poly buffer sizes and overall load of chunking (if chunking even kicks in during regular use)
- [ ] Before/after performance profiling of regular use (The Kuftal Tunnel chase, etc.)
- [ ] New/before/after performance profiling of spawning 32 mobs, and sending them on long-distance pathing
- [ ] Audit the logic in PathFind
- [ ] Explain about the problem with local minima
- [ ] Explain about the problems with the logic in PathFind (when "wallhack" kicks in)
- [ ] Explain about "chunking" in Detour
- [ ] Explain about the cost of the poly buffer in NavMesh
- [ ] Explain about the final "bogus" point in an incomplete path
- [ ] Explain about incomplete paths
- [ ] Make the internals of pathfind and navmesh as const as possible
- [ ] Play with reducing poly search extents to as small as possible
- [x] Remove all allocations possible inside pathing routines
- [ ] Address all TODOs in code and PR
- [ ] Keep a list of waypoints, so you can manually define waypoints for a mob to navigate around/between
- [ ] Lots of mobs are trying to path from their valid start position to (0,0,0) and this causes a failure, what's going on here?
- [ ] The Debug logging for navmesh takes up a lot of time, how can I ask Tracy to ignore that specific line/symbol?
- [x] Move cpp members in navmesh into CNavmesh class, so they're not shared between instances (protecting against future threading issues)
- [ ] The logic inside FindPath should be: If I've requested a path and it's incomplete or invalid and the wallhack flag is on, that's when I should wallhack to my destination - to go after people who aren't on-mesh. If I have a complete or a partial path, then I should just follow that. This will need careful testing!

Notes:
```
My long-distance pathing work has ballooned out into changing the pathing->wallhacking logic

The current logic (if the wallhack flag is on):
I will try ask the navmesh a single time for for directions to my destination. I will put that destination as the last point of my directions so I will always wallhack there. If I got back no points, I will wallhack straight away.

The new logic:
I will ask the navmesh for directions. If I get back points I will follow them. If I reach the end of my points I will ask the navmesh again and follow them. Repeat. If I'm at my destination: success! If I asked the navmesh for more points, I got none back, and I'm not yet at my destination: I will wallhack the rest of the way to my destination.
 
It's also possible to look up why a mob might be wallhacking to go off-mesh
We can make the mob produce logs saying "I'm going off-mesh to chase after player X. Navmesh says that player X is n distance away from the nearest navmesh poly (x, y, z, zoneId)" 

That then produces logs that can be used to find broken/abuse points
```
image

```
1. Start Point
2. Local Minima Point
3. End Point
```

```
Info: Bhaflau Thickets
Start Point (Zone In): 383.7, -10.1, 10.0
End Point (North of the map): 70.5, -32.2, 643.5
```

- Truncate `mob_spawn_points` so the server is empty of anything that might try to path
- Enable `DEBUG_NAVMESH` in Logging
- `!zone {Bha` -> `{Bhaflau Thickets}`
- `!togglegm` (So you don't get aggro'd)
- `!fafnir`
- `!exec target:pathTo(70.5, -32.2, 643.5)`
- `/follow` Fafnir
- Observe Fafnir's path
- Observe the debug logging about the navmesh/pathing

### DIFF
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -124,22 +124,11 @@ bool CPathFind::PathTo(const position_t& point, uint8 pathFlags, bool clear)
 
     if (isNavMeshEnabled())
     {
-        bool result = false;
-
-        if (m_pathFlags & PATHFLAG_WALLHACK)
-        {
-            result = FindClosestPath(m_POwner->loc.p, point);
-        }
-        else
-        {
-            result = FindPath(m_POwner->loc.p, point);
-        }
-
+        bool result = FindPath(m_POwner->loc.p, point, m_pathFlags & PATHFLAG_WALLHACK);
         if (!result)
         {
             Clear();
         }
-
         return result;
     }
     else
@@ -448,7 +437,7 @@ void CPathFind::StepTo(const position_t& pos, bool run)
     m_POwner->updatemask |= UPDATE_POS;
 }
 
-bool CPathFind::FindPath(const position_t& start, const position_t& end)
+bool CPathFind::FindPath(const position_t& start, const position_t& end, bool wallhack)
 {
     TracyZoneScoped;
 
@@ -466,6 +455,11 @@ bool CPathFind::FindPath(const position_t& start, const position_t& end)
 
     m_points       = std::move(points);
     m_currentPoint = 0;
+
+    if (result != CNavMesh::PathResult::Complete && wallhack)
+    {
+        m_points.emplace_back(pathpoint_t{ end, 0, false }); // this prevents exploits with navmesh / impassible terrain
+    }
 
     if (m_points.empty())
     {
@@ -525,39 +519,6 @@ bool CPathFind::FindRandomPath(const position_t& start, float maxRadius, uint8 m
     }
 
     return !m_points.empty();
-}
-
-bool CPathFind::FindClosestPath(const position_t& start, const position_t& end)
-{
-    TracyZoneScoped;
-
-    if (arePositionsClose(start, end))
-    {
-        return false;
-    }
-
-    if (!isNavMeshEnabled())
-    {
-        return false;
-    }
-
-    auto [result, points] = m_POwner->loc.zone->m_navMesh->findPath(start, end);
-
-    m_points       = std::move(points);
-    m_currentPoint = 0;
-
-    // TODO: This is just for testing, this wallhack behaviour is very important!
-    // m_points.emplace_back(pathpoint_t{ end, 0, false }); // this prevents exploits with navmesh / impassible terrain
-
-    /* this check requirement is never met as intended since m_points are never empty when mob has a path
-    if (m_points.empty())
-    {
-        // this is a trick to make mobs go up / down impassible terrain
-        m_points.emplace_back(end);
-    }
-*/
-
-    return true;
 }
 
 void CPathFind::LookAt(const position_t& point)
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -462,7 +462,9 @@ bool CPathFind::FindPath(const position_t& start, const position_t& end)
         return false;
     }
 
-    m_points       = m_POwner->loc.zone->m_navMesh->findPath(start, end);
+    auto [result, points] = m_POwner->loc.zone->m_navMesh->findPath(start, end);
+
+    m_points       = std::move(points);
     m_currentPoint = 0;
 
     if (m_points.empty())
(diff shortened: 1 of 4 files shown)


zach2good:
Testing (after changes, `-- Build type: RelWithDebInfo`):

- Change `!fafnir` command to spawn 32 (maybe 33?) Fafnirs, and then set them to do the path from the bottom to the top of Thickets.

- Whole operation (GM call, entity generation, and setting up of pathing for 32 or 33 mobs is 50ms or so)

image

image

zach2good:
Testing (after changes):

About 3k samples of kiting Robber Crabs around in Kuftal:

image

zach2good:
Before:

image

Changes are significantly worse? I'll have to dig in a bit

zach2good:
I've had another thought about this: we could do 2-phase pathfinding, as well as or instead of chunked pathing.

The outcome we want is to be able to tell a mob: "Hey, go here on the other side of the map" - and they do it. They do it without a huge blocking operation while it works out the path, and that path ends up being the whole distance, doesn't trap itself in local minima, etc. 

The local minima thing is a function of medium pathing running out of node buffer space.

We also want to avoid the situation of having all pathing happening on a background thread and the mob waiting still for the results to be delivered to it.

We could have it do a tiny bit of blocking pathing, while we also work out the full path on a background thread, and then reconcile the two when the big results come back. I don't know if this is better than chunking or not.

src/map/ai/helpers/pathfind.cpp
```diff
@@ -456,9 +449,16 @@ bool CPathFind::FindPath(const position_t& start, const position_t& end)
...
 
-    m_points       = m_POwner->loc.zone->m_navMesh->findPath(start, end);
+    auto [result, points] = m_POwner->loc.zone->m_navMesh->findPath(start, end);
+
+    m_points       = std::move(points);
     m_currentPoint = 0;
 
+    if (result != CNavMesh::PathResult::Complete && wallhack)
+    {
+        m_points.emplace_back(pathpoint_t{ end, 0, false }); // this prevents exploits with navmesh / impassible terrain
```
zach2good, line 461: TODO: Add logging here: "I'm going off-mesh after a certain target!", makes for a good indicator of where we have navmesh problems or abuse

zach2good:
Im going to keep this around and put a hold on it, even after the scheduler rework this would still be useful to eventually finish

MowFord:
I don't see anything explicitly mentioned here and I read through (and understood some of) it. Currently when a mob has careful pathing enabling, they can get stuck in places that (i assume) have breaks in the navmesh where pathfind will return a valid list of points, but stepping towards the next point doesn't get you far enough to not snap back to the previous point.

I had an admittedly ugly solution that I made as patch for cexi and [posted here](https://github.com/LandSandBoat/server/discussions/4694), but I'm not sure if it's something you know about/are considering here.

Just confirmed on current lsb it's still a thing:
- go to dyna sandy, then `!pos 113.8 2 9.9`
- turn on wireframe
- go NE a bit, pull the first mob you find and then `!pos 101 -4 1` and watch the mob path. it eventually gets to an equillibrium spot and just stands there

zach2good:
This was handled in: https://github.com/LandSandBoat/server/pull/10334

PR #6118, opened 2024-08-18
Ext: Update to Tracy v0.11.0

zach2good:
**_I affirm:_**

- Update binary download to new names with new version
- Audit and move around the define/settings
- Add admin/root checks to make sure you get all the information possible when using Tracy.


zach2good:
Ready to review, I'm not going to add more to this

WinterSolstice8 (approved):

PR #6119, opened 2024-08-18
Tem. N fix to equalize chameleon yarn and scarlet odoshi drops

lapislosh:
**_I affirm:_**

Limbus AF+1 drops are completely made-up on LSB and each item is intended to have one Normal drop chance and one Low drop chance per zone. The image below shows the current drop rates, and note that Chameleon Yarn has 3 entries and Scarlet Odoshi has 1 entry. This PR switches the Low drop chance for Chameleon Yarn on floor 2 to Scarlet Odoshi instead.

temn

Kill the three center Gigas in Tem N and keep opening chests until odoshi drops


zach2good (approved):
Thanks for the good explanation and the colour coded table!

PR #6123, opened 2024-08-19
Core: Adds nullptr check to treasure pool lookup

loxleygames:
**_I affirm:_**

Follow up on https://github.com/LandSandBoat/server/pull/6029
We had a few (rare) occurrences of PItem being nullptr and causing a crash, so I've added an additional check to the condition.

Cast a lot on a rare item you already have using packets

### DIFF
--- a/src/map/treasure_pool.cpp
+++ b/src/map/treasure_pool.cpp
@@ -276,7 +276,7 @@ void CTreasurePool::LotItem(CCharEntity* PChar, uint8 SlotID, uint16 Lot)
     }
 
     // Cannot lot if item is RARE and player already has it
-    if ((PItem->getFlag() & ITEM_FLAG_RARE) && charutils::HasItem(PChar, m_PoolItems[SlotID].ID))
+    if ((PItem != nullptr) && (PItem->getFlag() & ITEM_FLAG_RARE) && charutils::HasItem(PChar, m_PoolItems[SlotID].ID))
     {
         ShowError(fmt::format("Player {} is trying to lot on item {} (Rare) while already holding one (Packet injection)! ", PChar->getName(), m_PoolItems[SlotID].ID));
         return;


src/map/treasure_pool.cpp
```diff
@@ -276,7 +276,7 @@ void CTreasurePool::LotItem(CCharEntity* PChar, uint8 SlotID, uint16 Lot)
     }
 
     // Cannot lot if item is RARE and player already has it
-    if ((PItem->getFlag() & ITEM_FLAG_RARE) && charutils::HasItem(PChar, m_PoolItems[SlotID].ID))
+    if ((PItem != nullptr) && (PItem->getFlag() & ITEM_FLAG_RARE) && charutils::HasItem(PChar, m_PoolItems[SlotID].ID))
```
zach2good, line 279: I'd like this to be laid out similar to my last PR, with the additional logging too (adjust what the logging says as required):
```cpp
    CItem* PItem = itemutils::GetItem(m_PoolItems[SlotID].ID);
    if (PItem == nullptr)
    {
        ShowWarning(fmt::format("Player {} is trying to lot on an item that doesn't exist (PItem was nullptr) (Packet injection?)!", PChar->getName()).c_str());
        return;
    }
```

zach2good: A pattern I'm wanting to move towards (in both cpp and Lua):
```cpp
auto* mightBeNull = ThingThatCanReturnPointersThatCanBeNullptr();
if (mightBeNull == nullptr)
{
    // handle and bail out
    // log
    // return
}

// Go on with your life, using mightBeNull with confidence
```

loxleygames: I can see this is already merged into base https://github.com/LandSandBoat/server/pull/6120/files but it doesn't appear in the diff for this PR? In any case, this one is no longer required. Thanks for catching the issue. Will use this error handling pattern going forwards.

PR #6124, opened 2024-08-19
[core] Add mob mod to skip allegiance check (for healing with damage spells)

TracentEden2:
**_I affirm:_**

This PR adds a new mod mob (`SKIP_ALLEGIANCE_CHECK`) that allows skipping the allegiance check in the `CBattleEntity::ValidTarget` method. This is needed because some mobs use damaging spells (with `TARGET_ENEMY` as the validTargets flag) to heal themselves. Specifically, for example, the elementals in Waking the Beast heal both themselves and the avatars by casting T4 spells on themselves that they absorb. Thus this PR is helping to lay the ground work for WTB implementation (that will come in another PR). 

Unfortunately, this cannot be done by changing the flag on the spell object validTargets because `CBattleEntity::ValidTarget` is called before any available lua hook (such as the mobs `onSpellPrecast`) and thus would require a new hook or reordering of existing hooks. Therefore, I think this PR is the cleanest solution at the moment.

Find a casting mob and target it
Use !exec target:castSpell(144,target) and see no casting of Fire on itself
Use !exec target:addMobMod(79,1) to add the mob mod
Use !exec target:castSpell(144,target) and see the mob cast Fire on itself

### DIFF
--- a/scripts/enum/mob_mod.lua
+++ b/scripts/enum/mob_mod.lua
@@ -88,4 +88,5 @@ xi.mobMod =
     TRUST_DISTANCE         = 77, -- TRUSTS ONLY: Set movement type/distance. See trust.lua for details.
     STANDBACK_RANGE        = 78, -- Applies a specific standback range for the mob
     CANNOT_GUARD           = 79, -- Check if the mob does not guard (despite being a MNK or PUP mob)
+    SKIP_ALLEGIANCE_CHECK  = 80, -- Skip the allegiance check for valid target (allows for example a mob to cast a TARGET_ENEMY spell on itself)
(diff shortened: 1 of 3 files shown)


WinterSolstice8 (approved):
probably need a binding to specifically allow a certain spell to override this in a script but this is a decent workaround

Xaver-DaRed (approved):

Xaver-DaRed:
Needs rebase to resolve the conflicts

TracentEden2:
> Needs rebase to resolve the conflicts

Yep, should be fixed now, thanks!

PR #6133, opened 2024-08-20
Fix or Ignore remaining Lua Language Server warnings and errors

claywar:
**_I affirm:_**

Fixes or adds exceptions for all remaining LLS errors so that CI can be enforced
* Separated Cruor Prospector tables with different data structures
* Remove unnecessary bit.bor call with missing parameter
* Nil check manaclipper nextEvent variable
* Nil check Lower Jeuno Treasure Coffer 'prize' variable
* Add LLS exception for __eq and new for missions and hidden quests
* Nil check dynamis 'te' variable
* Change default table data type for caskets timeTable
* Nil check logInfo in setMissionStatus GM Command
* Define variables on need, remove debug print in avatar code
* Add TODOs and exceptions for remaining LLS errors and warnings

No impact to gameplay or function should be observed.


Xaver-DaRed (approved):
Great job

PR #6136, opened 2024-08-20
Use @enum annotation for tables in enum directory

claywar:
**_I affirm:_**

Uses `@enum` annotation for tables existing in scripts/enum.  See: https://github.com/LuaLS/lua-language-server/wiki/Annotations#enum

No impact to server, CI should pass

Negative test scenario:
```
local function getTrialProgress(player, trialId)
    local zoneType     = xi.zoneType.DOES_NOT_EXIST
    local playerTrials = getPlayerTrialData(player)
    local trialSlot    = playerTrials.slotLookup[trialId]

    if trialSlot then
        return playerTrials.trialData[trialSlot].progress
    end

    return nil
end
```

Result:
```
Code: undefined-field
Severity: Warning
Message: Undefined field `DOES_NOT_EXIST`.
```

### DIFF
--- a/scripts/enum/action.lua
+++ b/scripts/enum/action.lua
@@ -3,7 +3,7 @@
 -----------------------------------
 xi = xi or {}
 
----@class action
+---@enum action
 xi.action =
 {
     NONE                  = 0,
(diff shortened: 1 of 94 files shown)


Xaver-DaRed (approved):

claywar:
One thing of note for this, it appears that the annotations are not case sensitive (also in the docs as such), but I don't like inconsistencies

zach2good (approved):

PR #6141, opened 2024-08-23
[Core] [Lua] Rework TH effect on drop rates

Xaver-DaRed:
**_I affirm:_**

Applies the TH chart to treasure hunter. This means, no multiple rolls and instead following the chart provided by retail.
- Removes multiple rolls. That's not how TH works.
- Define the chart in lua and calculate new drop rate in there while we are at it.
- Enforce the brackets no matter what the drop rate in the database is.
- Ensure fixed drop rates for items and groups are both unaffected by TH and by the pre-defined brackets.

Notes:
- Drop rates in core STILL need a major cleanup.
- TH triggering still needs to be coded.
- Drop rates had to be multiplied by 10 here. drop/1000 just didnt cut it, considering the chart 0.01 precision.

imagen

Kill mobs and get drops without the server creating a singularity in the space/time continuum.

Review by commit. It really helps to see the changes made, specially because of the removal of loops for multiple rolls.

### DIFF
--- a/scripts/globals/combat/treasure_hunter.lua
+++ b/scripts/globals/combat/treasure_hunter.lua
@@ -3,7 +3,7 @@ xi.combat = xi.combat or {}
 xi.combat.treasureHunter = xi.combat.treasureHunter or {}
 
 -- https://forum.square-enix.com/ffxi/threads/56550
-local treasureHunterTable =
+xi.combat.treasureHunter.treasureHunterTable =
 {
 -- TH lvl    VC    C     UC    R     VR    SR   UR
     [ 0] = { 2400, 1500, 1000,  500,  100,  50,  10 },
@@ -23,7 +23,7 @@ local treasureHunterTable =
     [14] = { 8000, 7000, 3250, 2000, 1000, 500, 150 },
 }
 
-local thBracketTable =
+xi.combat.treasureHunter.dropBracketTable =
 {
     [1] = { 2400 },
     [2] = { 1500 },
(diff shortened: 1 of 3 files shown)


Frankie-hz:
I love you

src/map/entities/mobentity.cpp
```diff
@@ -715,49 +715,50 @@ void CMobEntity::DropItems(CCharEntity* PChar)
...
-                        previousRateValue += item.DropRate;
+                        break;
                     }
-                    break;
+                    previousRateValue += item.DropRate;
                 }
             }
         });
 
+        // Ungropued drops. This are affected by TH UNLESS they have an specified fixed rate.
```
EpicTaru: Spelling nit-pickiness: Ungropued -> Ungrouped

EpicTaru (changes requested):

Xaver-DaRed:
Tested and ready to review.
imagen

scripts/globals/combat/treasure_hunter.lua
```diff
@@ -0,0 +1,67 @@
+xi = xi or {}
+xi.combat = xi.combat or {}
+xi.combat.treasureHunter = xi.combat.treasureHunter or {}
+
+-- https://forum.square-enix.com/ffxi/threads/56550
+local treasureHunterTable =
```
zach2good: Make these local tables attached to the global so they can be more easily modified by modules 👍

Xaver-DaRed: Done

claywar (approved):

EpicTaru (approved):

PR #6146, opened 2024-08-23
[lua] Refactor Additional Effects to resolve cyclomatic complexity

WinterSolstice8:
**_I affirm:_**

Moves additional effects into global functions that can be called from elsewhere
Remove cyclomatic complexity call
Also fix Onyos knife

Test every single add effect type and see no error
This PR does NOT fix broken things that have no visuals!


TeoTwawki (approved):
❤️

scripts/globals/additional_effects.lua
```diff
@@ -110,209 +111,326 @@ xi.additionalEffect.calcDamage = function(attacker, element, defender, damage)
...
+    then
+        return 0, 0, 0 -- NMs immune or roll failed so return out
+    else
+        msgID = xi.msg.basic.ADD_EFFECT_STATUS
+        msgParam = xi.effect.KO
+        defender:setHP(0)
+    end
+
     return subEffect, msgID, msgParam
 end
```
Frankie-hz: There are two of these, not sure if this is intended?

WinterSolstice8: Whoops, sure isn't

WinterSolstice8: fixed

PR #6151, opened 2024-08-24
[Quest] Convert Making Amends to IF

jamesbradleym:
**_I affirm:_**

Converts `Making Amends` to IF

Pre-Reqs:
Windy fame 2+
Horutoto Ruins Experiment Mission not active

Start NPC:
Hakkuru-Rinkuru : !pos -111 -4 101 240

Trade Hakkuru-Rinkuru Animal Glue (!additem 937)

### DIFF
--- a/scripts/quests/windurst/Making_Amends.lua
+++ b/scripts/quests/windurst/Making_Amends.lua
@@ -1,0 +1,94 @@
+-----------------------------------
+-- Making Amends
+-----------------------------------
+-- Log ID: 2, Quest ID: 3
+-- Hakkuru-Rinkuru : !pos -111 -4 101 240
+-- Kuroido-Moido   : !pos -112 -4 103 240
+-----------------------------------
+
+local quest = Quest:new(xi.questLog.WINDURST, xi.quest.id.windurst.MAKING_AMENDS)
+
+quest.reward =
+{
+    fame     = 75,
+    fameArea = xi.fameArea.WINDURST,
+    title    = xi.title.QUICK_FIXER,
+}
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            local horutotoRuinsExperimentNotActive = player:getCurrentMission(xi.mission.log_id.WINDURST) ~= xi.mission.id.windurst.THE_HORUTOTO_RUINS_EXPERIMENT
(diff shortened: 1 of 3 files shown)


scripts/quests/windurst/Making_Amends.lua
```diff
@@ -0,0 +1,92 @@
...
+    fame     = 75,
+    fameArea = xi.fameArea.WINDURST,
+    title    = xi.title.QUICK_FIXER,
+}
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            local horutotoRuinsExperimentStatus = player:getMissionStatus(xi.mission.log_id.WINDURST, xi.mission.id.windurst.THE_HORUTOTO_RUINS_EXPERIMENT)
```
claywar: Just noting this now, but missionStatus will be returned regardless of nation.  You're going to need to be more specific

jamesbradleym: I might be misunderstanding the mechanics at play here, so pardon my ignorance.

As you note - this checks regardless of nation, but should still return `0` for `player:getMissionStatus` regardless of nation if you aren't actually on `THE_HORUTOTO_RUINS_EXPERIMENT`, right? In that case this check does the right thing. I suppose I could add a nation check and make this an or for not Windurst or not on `THE_HORUTOTO_RUINS_EXPERIMENT`?

cocosolos: Generally, `getMissionStatus` only takes one argument, which is the log you're checking (there are exceptions e.g. CoP). `player:getMissionStatus(xi.mission.log_id.WINDURST)` would return whatever the mission status variable is for whatever your current mission is for Windurst, and 0 if you're any other nation or no current mission.

cocosolos: For this check, you don't want to be checking the mission status, but `player:getCurrentMission(xi.mission.log_id.WINDURST) ~= xi.mission.id.windurst.THE_HORUTOTO_RUINS_EXPERIMENT` instead.

cocosolos: You also want to check `if player:getCurrentMission(xi.mission.log_id.WINDURST) == xi.mission.id.windurst.SAINTLY_INVITATION and player:getMissionStatus(xi.mission.log_id.WINDURST) > 1` to verify they're not blocked by the Saintly Invitation BCNM.

jamesbradleym: Thanks @cocosolos 🙏🏼

jamesbradleym:

PR #6153, opened 2024-08-24
[Quest] Convert Making Amens! to IF

jamesbradleym:
**_I affirm:_**

Converts quest `Making Amens!` to IF + adds some missing dialog

-- Log ID: 2, Quest ID: 8
-- Kuroido-Moido   : !pos -111 -4 101 240
-- Hakkuru-Rinkuru : !pos -111 -4 101 240
-- Mashira         : !pos 141 -6 138 200
-- Crematory Hatch : !pos 139 -6 127 200

Begin quest with Kuroido-Moido after completing `Making Amends`, zoning, and with Windurst fame >= 4
Go to Crematory Hatch, enter with Garliage Citadel Key (!additem 502)
Speak to Mashira, ask about wand
Complete quest by talking to Kuroido-Moido

### DIFF
--- a/scripts/quests/windurst/Making_Amens!.lua
+++ b/scripts/quests/windurst/Making_Amens!.lua
@@ -1,0 +1,122 @@
+-----------------------------------
+-- Making Amens!
+-----------------------------------
+-- Log ID: 2, Quest ID: 8
+-- Kuroido-Moido   : !pos -111 -4 101 240
+-- Hakkuru-Rinkuru : !pos -111 -4 101 240
+-- Mashira         : !pos 141 -6 138 200
+-- Crematory Hatch : !pos 139 -6 127 200
+-----------------------------------
+
+local quest = Quest:new(xi.questLog.WINDURST, xi.quest.id.windurst.MAKING_AMENS)
+
+quest.reward =
+{
+    fame     = 150,
+    fameArea = xi.fameArea.WINDURST,
+    gil      = 6000,
+    title    = xi.title.HAKKURU_RINKURUS_BENEFACTOR,
+}
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_AVAILABLE and
+                player:hasCompletedQuest(xi.questLog.WINDURST, xi.quest.id.windurst.MAKING_AMENDS) and
+                player:getFameLevel(xi.fameArea.WINDURST) >= 4 and
+                not quest:getMustZone(player)
+        end,
+
+        [xi.zone.PORT_WINDURST] =
+        {
+            ['Kuroido-Moido'] = quest:progressEvent(280),
+
+            onEventFinish =
+            {
+                [280] = function(player, csid, option, npc)
+                    quest:begin(player)
+                end,
+            },
+        },
+    },
+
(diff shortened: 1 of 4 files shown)


scripts/quests/windurst/Making_Amens!.lua
```diff
@@ -0,0 +1,124 @@
...
+-- Crematory Hatch : !pos 139 -6 127 200
+-----------------------------------
+
+local quest = Quest:new(xi.questLog.WINDURST, xi.quest.id.windurst.MAKING_AMENS)
+
+quest.reward =
+{
+    fame     = 150,
+    fameArea = xi.fameArea.WINDURST,
+    gil      = 6000 * xi.settings.main.GIL_RATE,
```
claywar: You're double-dipping here. Settings multiplier needs to be removed

scripts/quests/windurst/Making_Amens!.lua
```diff
@@ -0,0 +1,124 @@
...
+}
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_AVAILABLE and
+                player:hasCompletedQuest(xi.questLog.WINDURST, xi.quest.id.windurst.MAKING_AMENDS) and
+                player:getFameLevel(xi.fameArea.WINDURST) >= 4 and
+                not player:needToZone()
```
claywar: Need to move this to set/getMustZone, see the ADL

scripts/quests/windurst/Making_Amens!.lua
```diff
@@ -0,0 +1,124 @@
...
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_ACCEPTED and
+                not player:hasKeyItem(xi.ki.BROKEN_WAND)
+        end,
+
+        [xi.zone.GARLAIGE_CITADEL] =
+        {
+            ['Mashira'] =
+            {
+                onTrigger = function(player, npc)
```
claywar, line 54: Useless definition

scripts/quests/windurst/Making_Amens!.lua
jamesbradleym: Ahhh whoops!

scripts/quests/windurst/Making_Amens!.lua
jamesbradleym: Could you clarify this comment?

This is so the player can't just be tractored in or otherwise avoid opening the Hatchery Door to talk to Mashira.

claywar: A nil return will fall through to any other script that may use this door, so if it _is_ a door, it'll still open if defined elsewhere.  If you want a hard stop with no messaging, then you'll need to return `quest:noAction()`.  My suggestion though with the way you currently have it implemented, check if Prog > 0, then send the event in that case.  The broken out condition just makes it confusing

jamesbradleym: Gotcha! Makes sense! Intention is for Mashira to respond with his default so I'll simply check if Prog ~= 0 to validate.

jamesbradleym:

PR #6154, opened 2024-08-24
[Quest] Convert Mihgo's Amigo to IF

jamesbradleym:
**_I affirm:_**

Converts `Mihgo's Amigo` to IF + adds missing dialog

Pre-Reqs:
Quests `The Tenshodo Showdown` and `As Thick as Thieves` can not be active.
Mission `Lost For Words` can not be active.
Item `Marauder's Knife` can not be in inventory

Quest:
Speak to `Nanaa Mihgo` with at least 1 Norg fame to flag quest.
Optional -- Speak to other NPCs in area for quest dialog
Trade 4 `Yagudo Bead Necklace`s to Nanaa Mihgo

### DIFF
--- a/scripts/quests/windurst/Mihgo_s_Amigo.lua
+++ b/scripts/quests/windurst/Mihgo_s_Amigo.lua
@@ -1,0 +1,152 @@
+-----------------------------------
+-- Mihgo's Amigo
+-----------------------------------
+-- Log ID: 2, Quest ID: 25
+-- Nanaa Mihgo       : !pos 62 -4 240 241
+-- Cha Lebagta       : !pos 58 -6 216 241
+-- Bopa Greso        : !pos 59 -6 216 241
+-- Mheca Khetashipah : !pos 66 -6 185 241
+-- Gioh Ajihri       : !pos 108 -5 176 241
+-- Wani Casdohry     : !pos 104 -5 176 241
+-----------------------------------
+
+local quest = Quest:new(xi.questLog.WINDURST, xi.quest.id.windurst.MIHGO_S_AMIGO)
+
+local gilPerTrade = 200
+
+quest.reward =
+{
+    fame     = 60, -- 16 era based on 212 Yagudo Necklaces for 4 Norg fame
+    fameArea = xi.fameArea.NORG,
+    title    = xi.title.CAT_BURGLAR_GROUPIE,
+}
+
+quest.sections =
+{
+    {
+        -- QUEST AVAILABLE
+        check = function(player, status, vars)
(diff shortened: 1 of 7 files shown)


scripts/quests/windurst/Mihgo_s_Amigo.lua
```diff
@@ -0,0 +1,151 @@
...
+                        return quest:progressEvent(81) -- come to apologize
+                    else
+                        return quest:progressEvent(80)
+                    end
+                end,
+            },
+
+            onEventFinish =
+            {
+                [80] = function(player, csid, option, npc)
```
claywar, line 53: Given there's two events, I'd bet there's an option value that needs to be checked

scripts/quests/windurst/Mihgo_s_Amigo.lua
```diff
@@ -0,0 +1,151 @@
...
+            },
+
+            ['Wani_Casdohry'] = quest:event(86, 0, xi.item.YAGUDO_BEAD_NECKLACE), -- Migho's Amigo hint dialog
+
+            onEventFinish =
+            {
+                [88] = function(player, csid, option, npc)
+                    if quest:complete(player) then
+                        player:confirmTrade()
+                        player:addGil(xi.settings.main.GIL_RATE * gilPerTrade)
```
claywar, line 105: Is there a reason quest.reward value isn't being set and used here? That handles the gil rate config

scripts/quests/windurst/Mihgo_s_Amigo.lua
jamesbradleym: There's two events that result in the same outcome, it's just different dialog depending on whether or not you're currently on the `Crying for Onions` quest. Though it probably does make sense to set a quest var to remember this dialog was used for the following section (so if the player gets the `comes to apologize` message then they'll also get the `apology accepted` message even if they've completed the `Crying for Onions` quest in the meanwhile.

scripts/quests/windurst/Mihgo_s_Amigo.lua
jamesbradleym: Yep! The CS itself includes the gil received message so by setting it in the rewards table I was getting double messages.

hooksta4: There are a few others that have the gil reward baked into the CS dialog itself; see
https://github.com/LandSandBoat/server/blob/base/scripts/quests/sandoria/The_Sweetest_Things.lua#L91-L95

jamesbradleym:

Issue #6125, opened 2024-08-19
🐛 Monomi: Ichi duration is incorrect


mogulbgod:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)
All

## Branch affected by issue 

`base`

## Steps to reproduce

Cast Monomi: Ichi, duration is 2 minutes.

## Expected behavior

Monomi: Ichi duration is 7 minutes (420 seconds) after the January 21, 2014 FFXI update. [https://forum.square-enix.com/ffxi/threads/39564-Jan-21-2014-%28JST%29-Version-Update](https://forum.square-enix.com/ffxi/threads/39564-Jan-21-2014-%28JST%29-Version-Update)

almuth150:
Both Tonko spells also have incorrect durations.

clanofartisans:
This was fixed here: https://github.com/LandSandBoat/server/pull/6155

Xaver-DaRed:
ty @clanofartisans

PR #6155, opened 2024-08-24
Magic: Fixes the duration of Monomi and Tonko

loxleygames:
**_I affirm:_**

The duration values of Monomi and Tonko are based on 75 era values and are not retail accurate:
https://www.bg-wiki.com/ffxi/Monomi:_Ichi
https://www.bg-wiki.com/ffxi/Tonko:_Ichi
https://www.bg-wiki.com/ffxi/Tonko:_Ni

This duration was updated in 2014:
https://www.bg-wiki.com/ffxi/The_History_of_Final_Fantasy_XI/2014
> [Tonko: Ichi](https://www.bg-wiki.com/ffxi/Tonko:_Ichi)'s duration increased from 180 seconds to 420 seconds.
    [Tonko: Ni](https://www.bg-wiki.com/ffxi/Tonko:_Ni)'s duration increased from 300 seconds to 600 seconds.
    [Monomi: Ichi](https://www.bg-wiki.com/ffxi/Monomi:_Ichi)'s duration increased from 180 seconds to 420 seconds.

Cast Monomi and Tonko in-game. Ichi should be 7 minutes and Ni should be 10 minutes.


LoxleyX:
Should this be configurable? For example, we could expose the `pTable` table as a global or add a setting to use a different table. If this is wanted, let me know and the best way to proceed, and I will follow up.

TracentEden2:
> Should this be configurable? For example, we could expose the `pTable` table as a global or add a setting to use a different table. If this is wanted, let me know and the best way to proceed, and I will follow up.

From an ASB/Horizon rebase perspective it would be helpful it such tables were global so could be overridden, hopefully ok with LSB devs as well.

Xaver-DaRed (approved):

clanofartisans:
Resolves: https://github.com/LandSandBoat/server/issues/6125

PR #6156, opened 2024-08-24
[LLS] Add Lua Language Server definitions and annotations

claywar:
**_I affirm:_**

One of the great benefits of Lua Language Server is its ability to type-check classes and functions with their appropriate inputs and outputs.  In order to do this though, LLS needs to know what expected results are from anything it does check (on both the Core and Lua side of the house).  This PR manually implements and annotates all core Lua classes with the expected inputs and outputs for their various bindings.  Ideally, definitions from core should come from an automated process, and also add `@source` annotations as well.  At this time, there are some functions that would suffer from this, as we could not strictly check expected types, or know some of the multiple behaviors of inputs in functions.

Rules applied for implementing core types:
* `@class` definition matches `SOL_USERTYPE()` declaration
* Filename is named after `SOL_USERTYPE()` declaration
* `sol::variadic_args` and `sol::object` variables are limited to expected casts
* All getters are annotated with `@nodiscard` attribute
* Definition files must _still_ follow all lua stylecheck rules

NOTE: Some bindings have the ability to handle multiple inputs, and the annotations that are currently implement restrict even further to the expected parameter list for each method of entry.  Some such as `AddStatusEffectEx` have a floating boolean value as the last parameter, and this has been restricted to the currently-used parameter lists (two in all).

Implements the following typed objects for Lua Language Server.  Since these declarations are _not_ case sensitive, a `T` has been prepended to the definition in order to not mistakenly pull from another source or definition:
* `TAbilityAutomaton`
* `TAbilityPet`
* `TAbility`
* `TAttachment`
* `TCommand`
* `TEffect`
* `TItem`
* `TItemFood`
* `TItemFurniture`
* `TMobSkill`
* `TSpell`
* `TSpellTrust`
* `TWeaponSkill`

NOTE: When adding type definitions, the type will _only_ allow functions declared.  In the case of our style of defining an empty table, there is no lookahead with LLS to see if they're defined later in the script, so all function definitions must be flagged as nullable at this time.

In addition to the above implementations, all reported issues from LLS have been either corrected or temporarily disabled pending a workaround.  Number->Integer conversion is one of the issues disabled at the moment for parameter type mismatch.

No impact to gameplay outside of corrections made, CI should pass upon completion


claywar:
Ready for review, will address further annotations in a separate PR since this is getting a bit large

WinterSolstice8 (approved):
Skimmed through each commit, looked at things that edited > 1 line
LGTM

Xaver-DaRed (approved):

PR #6173, opened 2024-08-26
[LLS] Add core global definitions (luautils), clarify enums, and correct issues

claywar:
**_I affirm:_**

* Adds luautils defintions for global functions accessible on the Lua side
* Renames enums defined to use the same explicit name as position in xi table (zone enum == xi.zone)
* Corrects issues identified by LLS from the above changes

No impact to gameplay should be observed, CI should pass


Xaver-DaRed (approved):

claywar:
Not sure whats going on with the Full Startup Windows check.  Not throwing any errors up when testing locally on Windows, and seems is getting a nil return from a luautils binding (which shouldn't be possible)

claywar:
Aha!  so when the `--load_all` parameter is used with xi_map, everything is cached, including spec files.  Latest commit adds a chunk to ignore items in that directory in this CI case

PR #6187, opened 2024-08-29
[sql] Add enlight anim to Excalibur additional effect

jamesbradleym:
**_I affirm:_**

Updates the additional effect of Excalibur to light

[As shown here](https://www.youtube.com/watch?v=mZPa7ZLmCKc#t=3m33s)

Trigger excalibur additional effect, see the light


WinterSolstice8 (approved):

Xaver-DaRed:
mmm isnt this PR setting excalibur element to light?

Not gonna comment on the animation, that I'm sure its fine, but...

From my understanding...
The issue with excalibur's additional effect, is that it isnt elemental damage, in nature.
One of the reasons Teo wanted "damage type" to be able to be fed into magic functions, was because this additional effect deals "slashing damage"

Xaver-DaRed:
I'll wait on winter to chime in, but, I think this lines:
```
INSERT INTO `item_mods` VALUES (18276,950,7); -- ITEM_ADDEFFECT_ELEMENT: ELEMENT_LIGHT
```
Should either:
- Be deleted and a comment added stating the additional effect needs to be set to "Slashing damage" somehow.
- Be kept IF they are needed, with a TODO comment stating the element is incorrect and it should be "slashing damage" instead.

WinterSolstice8:
> I'll wait on winter to chime in, but, I think this lines:
> 
> ```
> INSERT INTO `item_mods` VALUES (18276,950,7); -- ITEM_ADDEFFECT_ELEMENT: ELEMENT_LIGHT
> ```
> 
> Should either:
> 
>     * Be deleted and a comment added stating the additional effect needs to be set to "Slashing damage" somehow.
> 
>     * Be kept IF they are needed, with a TODO comment stating the element is incorrect and it should be "slashing damage" instead.

Really, excalibur needs its own add effect OR we need to add a ton of parameters to make it possible

excal is an hp% based slashing damage attack but I know that acid bolts are also "piercing damage" but are subject to magic resist checks in certain ways

but yes, probably need to add a TODO here, the PR does solve the problem stated, though it's still not quite accurate. it looks like add effect damage assumes elemental, which is true in most cases

PR #6188, opened 2024-08-29
Add Synergy menu research

zach2good:
**_I affirm:_**
- [x] understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.

I'm over-caffeinated today and remember hearing that synergy looked very complicated - so I went and exhaustively threw numbers at the menus until I got it somewhat under control, and stuffed it into an inert global. No idea how the recipe lookup works, but everything else is somewhat mapped.

I used my `Synergy - Galley Kitchen` cap for reference.

### DIFF
--- a/scripts/globals/synergy.lua
+++ b/scripts/globals/synergy.lua
@@ -1,0 +1,283 @@
+-----------------------------------
+-- Synergy
+-----------------------------------
+require('scripts/globals/utils')
+-----------------------------------
+xi = xi or {}
+xi.synergy = xi.synergy or {}
+
+-----------------------------------
+-- Synergy Furnace
+-----------------------------------
+
+-- NOTE: Change this to play with the synergy furnace
+local testSynergy = false
+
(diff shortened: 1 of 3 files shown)


scripts/globals/synergy.lua
```diff
@@ -0,0 +1,267 @@
...
+
+xi.synergy.synergyFurnaceOnTrade = function(player, npc, trade)
+    if not TEST_SYNERGY then
+        return
+    end
+
+    local arg0 = 8
+    local arg1 = 83886085 -- Galley Kitchen?
+    local arg2 = 0
+    local arg3 = 69121
```
cocosolos: Item ID looks like it's the first 16 bits of arg3.

69121 = `0001 0000 1110 0000 0001`
Galley Kitchen 3585 = `1110 0000 0001`

zach2good: Quite right, thanks!

zach2good:
Done battling with CI, ready to go

claywar (approved):

PR #6190, opened 2024-08-30
[core] Force crit fail on synth if player zones for any reason

WinterSolstice8:
**_I affirm:_**

Force crit fail on synth if player zones for any reason

!zone after starting synth and crit fail

### DIFF
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -485,6 +485,12 @@ void CZoneEntities::DecreaseZoneCounter(CCharEntity* PChar)
         }
     }
 
+    // Duplicated from charUtils, it is theoretically possible through d/c magic to hit this block and not sendToZone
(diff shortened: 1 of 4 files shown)


Xaver-DaRed (approved):

src/map/zone_entities.cpp
```diff
@@ -485,6 +485,22 @@ void CZoneEntities::DecreaseZoneCounter(CCharEntity* PChar)
         }
     }
 
+    // Duplicated from charUtils, it is theoretically possible through d/c magic to hit this block and not sendToZone
```
zach2good, line 488: Just to be anal, can this be extracted out into something in charutils or synthutils? `failSynthOnZone` or something

WinterSolstice8: updated

zach2good (approved):

PR #6192, opened 2024-08-30
Map out Synergy messaging and animations

zach2good:
**_I affirm:_**

Some work with my morning coffee:
- Extract out the messages and animations I saw in my `Galley Kitchen` capture
- Add some state tracking so you can claim, trade, relinquish, and do distance checks with the furnace

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -325,6 +325,9 @@
         "SendToJailOffline",
         "GetPlayerIDByName",
         "PlayerHasValidSession",
+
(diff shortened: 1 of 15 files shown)


Xaver-DaRed:
Adding a note here:
The .sql table will need a new column to contain the skill ID used n the recipe.

Not all skills use synergy skill.
https://www.bg-wiki.com/ffxi/Synergy/Recipes/Amateur

Xaver-DaRed (approved):

PR #6196, opened 2024-08-31
[Quest] Move The Tenshodo Showdown mob steal code to IF file

jamesbradleym:
**_I affirm:_**

Adds `onSteal` to IF
Moves npc code handling `onSteal` for the THF AF1 quest into the quest lua

Steal from Quadavs during THF AF1 quest, see unchanged results.

## Notes
Some players indicated they could steal the stew prior to flagging the quest:
[1](https://ffxi.allakhazam.com/db/quests.html?fquest=225#:~:text=1/1%20on%20a%20Silver%20Quadav%20pulled%20from%20the%20top%20%3A%20didn%27t%20have%20the%20quest%20started.%20Good%20luck%20to%20the%20rest%20of%20you!)
[2](https://ffxi.allakhazam.com/db/quests.html?fquest=225#:~:text=I%20would%20like,rolenberry%20for%20me.)

### DIFF
--- a/scripts/quests/windurst/THF_AF1_The_Tenshodo_Showdown.lua
+++ b/scripts/quests/windurst/THF_AF1_The_Tenshodo_Showdown.lua
@@ -1,6 +1,7 @@
 -----------------------------------
 -- The Tenshodo Showdown
 -----------------------------------
+-- Log ID: 2, Quest ID: 69
 -- !addquest 2 69
 -- Nanaa Mihgo : !pos 62 -4 240 241
 -- Harnek      : !pos 44 0 -19 245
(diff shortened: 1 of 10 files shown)


claywar:
onSteal is not currently observed by Interaction.

jamesbradleym:
> onSteal is not currently observed by Interaction.

I swear when I tested I got the stew item but now I'm not getting it..

jamesbradleym:
@claywar Looks like it worked because the stew is already in the mob_droplist steal section, this makes the whole `onSteal` escapade unnecessary?

claywar:
The way to verify this is to see if quadav stew can be stolen by people not on the quest (I'm betting this is no).

jamesbradleym:
Ah, and in that case it would need to be removed from the drop list + onSteal implemented for IF. 👍🏻

jamesbradleym:
As a note: [Yes you can steal the stew prior to quest start](https://discord.com/channels/443544205206355968/493243620258873368/1279887185113383014), but stew seems to be prioritized for those currently on the quest per player reports on wikis

jamesbradleym:
@claywar I've made the necessary updates for onSteal to be recognized in IF, would you mind taking a look to make sure this is following the right pattern? Took a look at how `onTrigger` and `onTrade` operate in IF to put together `onSteal` so hopefully it's clear cut.

Tested with the quest and on successful steal the `onSteal` in the quest lua is called and returns the `Quadav Stew` to be given to the player

jamesbradleym:
Hey @claywar Just wanted to check in on this to see if things were looking good?

WinterSolstice8 (approved):
cant comment on the pattern (not my expertise) but it looks fine to me

jamesbradleym:
Thanks @WinterSolstice8 🙏🏼
@claywar checking in on this to see if we're good to go

scripts/quests/windurst/THF_AF1_The_Tenshodo_Showdown.lua
```diff
@@ -1,7 +1,7 @@
 -----------------------------------
 -- The Tenshodo Showdown
 -----------------------------------
--- !addquest 2 69
+-- Log ID: 2, Quest ID: 69
```
zach2good, line 4: `!addquest 2 69` is here so you can quickly copy/paste it for testing, same as with the commands below it

zach2good (approved):
Seems fine, just needs that comment ammended

scripts/quests/windurst/THF_AF1_The_Tenshodo_Showdown.lua
jamesbradleym: Gotcha - plopped it back in

PR #6114, opened 2024-08-17
[BCNM] Make Ifrit TBF and TSTBF more retail accurate

TracentEden2:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?
Adjusts Ifrit Prime for Trial by Fire aka TBF and Trial Size Trial by Fire aka TSTBF to be more retail accurate.

This is done by first splitting the TBF Ifrit and TSTBF Ifrit into two separate luas (as they have different logic and having separate helps with readability and understandability). 

TBF is given several retail accurate mods including status immunities, fire absorb, enfire effect, udmgphy, udmgrange, and sight and magic aggro ranges. The TBF skill list was corrected by removing Inferno (which it was using as normal skill) and adding Burning Strike and Fire VI. Also the Ifrit mystic avatars (from temenos) that were also using the TBF skill list were changed to use a different Ifrit skill list that better aligns with the fact that they can use any Ifrit skills.

TSTBF is given the same immunities, udmgphy, udmgrange, and sight and magic aggro ranges as TBF. However the enfire damage level was changed to be accurate to retail videos (as no captures exist), 3000 TP is given on engage as retail videos show mobskill use on engage. And a LIGHT_MEVA of -35 is given so that searing light resist rate roughly matches that seen in retail videos (24/27 unresisted searing lights from my checking).

TBF, TSTBF, and ASA were given the correct job of drk/blm (like player avatars).

The changes were based on captures including [here](https://youtu.be/EG-sSXfX908) and [here](https://youtu.be/GFF3YSEmBS4), a personal capture I will upload, trial by fire videos like [here](https://www.youtube.com/watch?v=MGN7jSZjAJU) and [here](https://www.youtube.com/watch?v=xbqTk5xoEyw), and trial size trial by videos (of all different avatars) for searing light resist rate and MEVA change. Also mystic avatar change based on capture [here](https://youtu.be/U34QYlfzvgs).

[...]

PR #6208, opened 2024-09-03
[BCNM] Make remaining TB and TSTB BCNMs more retail accurate

TracentEden2:
**_I affirm:_**

This is a follow up PR of this [PR](https://github.com/LandSandBoat/server/pull/6114) (which only focused on Ifirit as an initial test case). This PR makes analogous changes to the other Trial By (TB) and Trial Size Trial By (TSTB) BCNMs. The changes are based on the captures by Siknoz and Aether (Levi [here](https://youtu.be/Jfq54n-7IT8) and [here](https://www.dropbox.com/s/8oa8x9trkhgkl6v/leviathan.rar?dl=0), Garuda [here](https://youtu.be/3bhl92mOD4w) and [here](https://youtu.be/u8knF9Q1514), Ramuh [here](https://youtu.be/JBMSjJNq4L0) and [here](https://youtu.be/RSLFWXQ1iwE), Shiva [here](https://youtu.be/Kdxf-WzsIAI) and [here](https://youtu.be/I7PPtqe_WwU), and Titan [here](https://youtu.be/Ap4T4XOqz-M) and [here](https://youtu.be/F0d09R79Wng)). These captures give info like status immunities, TP moves (though not all), HPs, and so forth. Note the HPs of the TB avatars have been standardized to 8000, as that is the most common HP seen in the HP TRACK ranges across all 10 TB captures. This is similar to the already existing standardization of 700 HP for TSTB avatars.

The PR also fixes an issue with TSTBE where the `onEventFinishWin` function was missing (probably due to copy paste error from the normal TBE which does not have such a function.

Fight the remaining (non-Ifrit) trial by and trial size trial by fights


Xaver-DaRed (approved):
Looks good to me. I'm not the biggest fan of the acronyms, but they do make sense

PR #6140, opened 2024-08-22
[BCNM] Add Waking the Beast Ifrit BCNM

TracentEden2:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?
This PR adds the waking the beast Ifrit BCNM fight. This is the first in a series of PRs for adding WTB. Note that this does not add the actual quest so players cannot yet enter the BCNM fight, as my plan is to add all of the BCNM fights before a final PR that adds in the quest. Also note that the commented out `-- mob:setMobMod(xi.mobMod.SKIP_ALLEGIANCE_CHECK, 1)` will be enabled once this [related PR](https://github.com/LandSandBoat/server/pull/6124) that adds the specific mob mod is merged.

I am not sure if this should be marked as experimental or not?

The info for the PR has been taken from a variety of captures and retail videos including:
- General behavior, mobskills, HP, immunities, res ranks, and MDT ([here](https://www.youtube.com/watch?v=i9dTJ-A0MCg), [here](https://youtu.be/poednY37W1Q) see [capture logs](https://www.dropbox.com/sh/cge8c766pmam3ha/AAAb2ROna7OPu-e6eCPJtLcIa?dl=0) as chat log cutoff)
- Elementals refocusing hate on avatar target periodically ([here](https://youtu.be/A2SKt45n578?t=27))
- Elemental En-Plague Power ([here](https://www.youtube.com/watch?v=8EbAhDq5SFA&t=11s))

## Steps to test these changes
As a GM go to the BCNM and enter the fight and fight the mobs

PR #6214, opened 2024-09-04
[BCNM] Add remaining WTB avatar BCNMs

TracentEden2:
**_I affirm:_**

This is a follow up PR of this [PR](https://github.com/LandSandBoat/server/pull/6140) (which only focused on Ifirit as an initial test case). This PR continues by implementing the remaining WTB avatar BCNMs (final fight and quest itself will come in future PRs). 

The implementations are based on the captures by Siknoz and Aether (Levi [here](https://youtu.be/gVKKc0m01JQ), [here](https://youtu.be/pxIb73jqyVA), [here](https://youtu.be/kootHYrwq5s) and [here](https://youtu.be/_danfhoGgZI), Garuda [here](https://youtu.be/WaeV8JC7s4A), [here](https://youtu.be/tK6Gcj13f00), [here](https://youtu.be/mjUtFUa6NSE), [here](https://youtu.be/NiHUcHnBdF0), [here](https://youtu.be/pV7bgMdSk64) and [here](https://youtu.be/_-yCA5YPlpA), Ramuh [here](https://youtu.be/oI0BD__2d50) and [here](https://youtu.be/SEaBingm5Gw), Shiva [here](https://youtu.be/KHM4yG4PAog) and [here](https://youtu.be/JCqSKOdogtI), and Titan [here](https://youtu.be/grAamkCzYbw) and [here](https://youtu.be/yVAOP2VSAtM)). These captures give info like status immunities, TP moves, HPs, and so forth. 

The main anomalies between the avatars are that Garuda does not use Pred Claws (in 100+ TP moves over 4 captures), while all other avatars do use their lvl 70 skills. Also fixes some small issues with the Ifrit fight that were missed previously, specifically that Ifrit does use Flaming Crush.

Also cloister of gales is missing the `ENTERING_THE_BATTLEFIELD_FOR` text ID and this PR adds this so the correct text is shown when entering any cloister of gales BCNM.

Fight the (non-Ifrit) WTB avatar fights (use GM char to enter as the quest is not yet available)


scripts/battlefields/Cloister_of_Frost/waking_the_beast.lua
```diff
@@ -0,0 +1,79 @@
...
+
+            {
+                cloisterOfFrostID.mob.SHIVA_PRIME_WTB + 11,
+                cloisterOfFrostID.mob.SHIVA_PRIME_WTB + 12,
+                cloisterOfFrostID.mob.SHIVA_PRIME_WTB + 13,
+                cloisterOfFrostID.mob.SHIVA_PRIME_WTB + 14,
+            },
+        },
+    },
+}
```
jmcmorris: You should be able to use names rather than ids here. It should automatically figure out the stride for the mobs and which mobs belong to each battlefield area. I'd try it with one and test it. It should work. If it does work then you can simplify them all!
```lua
{
    mobs = { "Shiva_Prime_WTB" }
    allDeath = ...
},
{
    mobs = { "Ice_Elemental" }
}
```

TracentEden: > You should be able to use names rather than ids here. It should automatically figure out the stride for the mobs and which mobs belong to each battlefield area. I'd try it with one and test it. It should work. If it does work then you can simplify them all!
> 
> ```lua
> {
>     mobs = { "Shiva_Prime_WTB" }
>     allDeath = ...
> },
> {
>     mobs = { "Ice_Elemental" }
> }
> ```

Thanks, yep, this should now be fixed.

zach2good (approved):

PR #6220, opened 2024-09-06
[BCNM] Implement BCNM20 Shooting Fish

The-Aerec:
**_I affirm:_**

Implements the Aqua Ball Mechanic for BCNM20 shooting fish.

Basis and evidence for following changes:
https://youtu.be/V0Qp99O1AkA
https://drive.google.com/open?id=1ZpZ9FEQ3TYeCc9lPig-y2p6pCRLNkdyF
https://ffxiclopedia.fandom.com/wiki/Shooting_Fish

image

Ability 313 it is - renamed from counterspore for clarity

Enter BCNM20 shooting fish and see the mobs properly only use their knockback aquaball as a auto-attack.

### DIFF
--- a/scripts/actions/mobskills/aqua_ball_knockback.lua
+++ b/scripts/actions/mobskills/aqua_ball_knockback.lua
@@ -1,0 +1,31 @@
+-----------------------------------
+-- Aqua Ball
+-- Deals ranged water damage that causes knockback.
+-- Used by pugils in BCNM: Shooting Fish
+-----------------------------------
+require('scripts/globals/mobskills')
+-----------------------------------
+---@type TMobSkill
+local mobskillObject = {}
+
+mobskillObject.onMobSkillCheck = function(target, mob, skill)
+    return 0
+end
+
+mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local damage   = mob:getWeaponDmg() * 3
+    local power    = 20
+    local tick     = 3
+    local duration = power * tick
+
+    damage = xi.mobskills.mobMagicalMove(mob, target, skill, damage, xi.element.WATER, 1, xi.mobskills.magicalTpBonus.MAB_BONUS, 1.5)
+    damage = xi.mobskills.mobFinalAdjustments(damage, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.WATER, xi.mobskills.shadowBehavior.IGNORE_SHADOWS)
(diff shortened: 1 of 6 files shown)


sql/mob_skill_lists.sql
```diff
@@ -3911,7 +3911,11 @@ INSERT INTO `mob_skill_lists` VALUES ('Avatar_Ifrit_WTB',2013,844); -- Waking th
 INSERT INTO `mob_skill_lists` VALUES ('Avatar_Ifrit_WTB',2013,845); -- Waking the Beast, fire iv
 INSERT INTO `mob_skill_lists` VALUES ('Avatar_Ifrit_WTB',2013,847); -- Waking the Beast, meteor_strike
 
--- Next available ID: 2014
+INSERT INTO `mob_skill_lists` VALUES ('Shooting_Fish',2014,313); -- knockback aquaball - BCNM20 Shooting Fish
+
+-- Next available ID: 2008
```
Frankie-hz: Can remove this, 2008 is used

scripts/actions/mobskills/aqua_ball_knockback.lua
```diff
@@ -0,0 +1,30 @@
+-----------------------------------
+-- Aqua Ball
+-- Deals ranged water damage that causes knockback.
+-- Used by pugils in BCNM: Shooting Fish
+-----------------------------------
+require('scripts/globals/mobskills')
+-----------------------------------
+local mobskillObject = {}
```
jmcmorris, line 9: ```---@type TMobSkill```

jmcmorris (approved):
Looks good! Just missing the new type info for the skill.

zach2good:
I can't comment on the exact line, but is everything in place for this to not be considered experimental? Mob levels, mob hp, resistances, the mechanics, the loot, the entry, etc.

LacunaXI:
The loot in the .lua is set to the exact rates of Clopedia with limited samples. Given the limited drop variance, Platoon weapons for example are more likely each an equal weight with the others in the slot. Would it make sense for the weights to be more equalized for some of these drops?

jmcmorris:
Yes, ideally we use loot with the weights defined in `xi.loot.weight`. https://github.com/LandSandBoat/server/blob/base/scripts/globals/battlefield.lua#L64-L72

zach2good (approved):

PR #6222, opened 2024-09-07
Clear the mob's party when removing a party member

jmcmorris:
**_I affirm:_**

When removing a party member from a mob party the party member's PParty would not be set to nullptr. This would cause it to not be allowed to be added to other parties.

I was testing this with #6214 with the local change of the Shiva WTB battlefield having this:
```lua
content.groups =
{
    {
        mobs = { 'Shiva_Prime_WTB' },
        death = function(battlefield, mob)
            -- when avatar defeated then all elementals should also die
            for i = 1, 4 do
                local elemental = GetMobByID(mob:getID() + i)
                if elemental and elemental:isAlive() then
                    elemental:setHP(0)
                end
            end

            battlefield:setStatus(xi.battlefield.status.WON)
        end,
    },

    {
        mobs =
        {
            'Shiva_Prime_WTB',
            'Ice_Elemental',
        },
        isParty   = true,
        superlink = true,
    },
}
```

With these changes Shiva Prime and the Ice Elementals should link properly due to them being in a party together. Without the change in party.cpp they do not get added to a party together.

### DIFF
--- a/src/map/party.cpp
+++ b/src/map/party.cpp
@@ -346,7 +346,6 @@ void CParty::RemoveMember(CBattleEntity* PEntity)
                 PChar->pushPacket(new CPartyDefinePacket(nullptr));
                 PChar->pushPacket(new CPartyMemberUpdatePacket(PChar, 0, 0, PChar->getZone()));
                 PChar->pushPacket(new CCharUpdatePacket(PChar));
-                PChar->PParty = nullptr;
 
                 _sql->Query("DELETE FROM accounts_parties WHERE charid = %u", PChar->id);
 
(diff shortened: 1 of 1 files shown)


WinterSolstice8 (approved):

claywar (approved):
The other option here, since there's not a way to define some mobs twice in all scenarios (say, only the prime is required to kill, but elementals not), would be to set the same superlinkGroup for each defined content group.  Same group across requirements will superlink in that scenario (I want to say it just takes the upper half of superlink numbers and assigns a unique value on BCNM start)

PR #6223, opened 2024-09-07
Rework Battlefield Group Death Counter

jmcmorris:
**_I affirm:_**

Before this change it would track the number of mobs that have died in a battlefield group. Once all the mobs have died it would call the `allDeath` callback. This works fine for mobs that are spawned at the start of the battlefield and have to die. This does not work if not all mobs are spawned at the start and/or can be spawned multiple times and consequently the death count can reach the target amount without having all mobs dead.

This is fixed by tracking how many mobs are dead for the group when a mob dies.

I'm not sure if we have any battlefields that currently work like this but I have one in the works (Call To Arms ISNM) that does. You can cheat this though by going to a battlefield that does have an `allDeath` and respawn a mob after killing it to see that it doesn't work without this change and will work with this change.

Divine Punishers BCNM has allDeath
`!additem 1130`
`!zone Balgas Dias`
Enter Divine Punishers and go kill a mob with `!hp 0` after they despawn then respawn it. Kill the other mobs one by one and the bcnm should end with one mob remaining.

### DIFF
--- a/src/map/battlefield.cpp
+++ b/src/map/battlefield.cpp
@@ -902,72 +902,61 @@ void CBattlefield::handleDeath(CBaseEntity* PEntity)
         return;
     }
 
-    for (auto& group : m_groups)
-    {
-        for (uint32 mobId : group.mobIds)
-        {
-            if (mobId == PEntity->id)
-            {
-                ++group.deathCount;
-
-                break;
-            }
-        }
-    }
-
+    // Create a copy of groups since m_groups may change during the callbacks
     auto groups(m_groups);
 
     for (auto& group : groups)
     {
+        // Calculate the total mobs that are dead for this group
+        uint8 deathCount = 0;
+
+        for (uint32 mobID : group.mobIds)
+        {
+            CMobEntity* PMob = dynamic_cast<CMobEntity*>(zoneutils::GetEntity(mobID, TYPE_MOB | TYPE_PET));
+            if (PMob == nullptr || PMob->isDead())
+            {
+                ++deathCount;
+            }
+        }
+
         for (uint32 mobId : group.mobIds)
         {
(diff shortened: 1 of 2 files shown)


src/map/battlefield.cpp
```diff
@@ -902,72 +902,61 @@ void CBattlefield::handleDeath(CBaseEntity* PEntity)
...
+    for (auto& group : groups)
     {
-        for (uint32 mobId : group.mobIds)
+
+        // Calculate the total mobs that are dead for this group
+        uint8 deathCount = 0;
+        for (uint32 mobID : group.mobIds)
         {
-            if (mobId == PEntity->id)
+            CMobEntity* PMob = (CMobEntity*)zoneutils::GetEntity(mobID, TYPE_MOB | TYPE_PET);
```
WinterSolstice8:
```suggestion
            CMobEntity* PMob = dynamic_cast<CMobEntity*>(zoneutils::GetEntity(mobID, TYPE_MOB | TYPE_PET));
```

WinterSolstice8:
I think CI wants you to remove the whitespace below that brace, it has difficulty displaying removing whitespace in actions

WinterSolstice8 (approved):
LGTM but want Clay's input

claywar (approved):
Seems reasonable to me assuming all else works.  Was scratching my head thinking about broken mechanics, since I knew that I implemented some fights that had reraise mechanics.  Looking at it, there's nothing that quite encountered this scenario yet.

Undying Promise has a reraise mechanic, but only one mob per group in the final set which does reraise.  In this case, a local is tracking it, so _should_ be no impact.

The other cases are more around pet respawn mechanics (such as Divine Might, AA battles), but those aren't in the required group.

PR #6230, opened 2024-09-09
[Quest] Implement Fisherman's Heart Quest and supporting tools

paladindamarus:
**_I affirm:_**

- Implements FIsherman's Heart Quest
- Adds data structures to log fishing history as reported by NPC in Mhaura (Katsunaga)

- Do any amount of fishing and report to Katsunaga in Mhaura
- Katsunaga will not provide stats until the quest is complete.  Requires Fishing level 20, min Recruit, and a Gugru Tuna (ID: 4480)
- As you catch fish types, it should be marked with a STAR in the list provided by Katsunaga.
- You can fake catching a fish by self-targeting and using `!exec target:setFishCaught(1234, true)` (replacing 1234 with the fish ID)

## Sources

- Capture discord (captures from Siknoz)
- Custom captures
- Trial / Error on bitmapping using local client


scripts/globals/quests.lua
```diff
@@ -482,7 +482,7 @@ xi.quest.id =
         THE_SAND_CHARM                  = 8,  -- +
         A_POTTER_S_PREFERENCE           = 9,  -- +
         THE_OLD_LADY                    = 10, -- +
-        FISHERMAN_S_HEART               = 11,
+        FISHERMAN_S_HEART               = 11, -- + Converted
```
claywar, line 485: Just for my own sake, can you please rename this enum to `FISHERMANS_HEART`?

scripts/quests/otherAreas/Fishermans_Heart.lua
```diff
@@ -0,0 +1,147 @@
...
+    fame = 30,
+}
+
+quest.sections =
+{
+    -- Quest available, but fishing rank is 1 and skill < 20
+    -- (NOTE: Fishing Rank 0 is handled as default action)
+    {
+        check = function(player, status, vars)
+            return
```
claywar, line 20: Formatting-wise, return <first cond> <and/or> newline, then following ones with 4-space indent

scripts/quests/otherAreas/Fishermans_Heart.lua
```diff
@@ -0,0 +1,147 @@
...
+                status == xi.questStatus.QUEST_AVAILABLE and
+                player:getFameLevel(xi.fameArea.WINDURST) >= 2 and
+                player:getCharSkillLevel(xi.skill.FISHING) >= 200
+        end,
+
+        [xi.zone.MHAURA] =
+        {
+            ['Katsunaga'] =
+            {
+                onTrigger = function(player, npc)
```
claywar, line 45: simplified: `onTrigger = quest:progressEvent(192, xi.item.GUGRU_TUNA),`

scripts/quests/otherAreas/Fishermans_Heart.lua
```diff
@@ -0,0 +1,147 @@
...
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_COMPLETED
+        end,
+
+        [xi.zone.MHAURA] =
+        {
+            ['Katsunaga'] =
+            {
+                onTrigger = function(player, npc)
```
claywar, line 108: See above

scripts/quests/otherAreas/Fishermans_Heart.lua
```diff
@@ -0,0 +1,147 @@
...
+                    return quest:progressEvent(197, player:getFishingCatchCount() or 0) -- Number of fish types caught
+                end,
+            },
+
+            onEventUpdate =
+            {
+                [197] = function(player, csid, option, npc)
+                    if option == 33554432 then
+                        -- 'Catching Fish'
+                        local fishingStats = player:getFishingStats() or {}
```
claywar, line 118: Haven't gotten there yet, but I'd rather break out if fishing stats are non-existent.  updateEvent can take 0 params, so no need to pass an empty table.  Also, is this data returned always the same, and could we consolidate it to always return what we need in the format needed?

scripts/quests/otherAreas/Fishermans_Heart.lua
```diff
@@ -0,0 +1,147 @@
...
+                            [1] = fishingStats['fishLinesCast'] or 0,
+                            [2] = fishingStats['fishReeled'] or 0,
+                            [3] = fishingStats['fishLongestLength'] or 0,
+                            [4] = fishingStats['fishLongestId'] or 0,
+                            [5] = fishingStats['fishHeaviestWeight'] or 0,
+                            [6] = fishingStats['fishHeaviestId'] or 0,
+                        })
+                    elseif option == 0 then
+                        -- 'Types of Fish Caught'
+                        local fishingCatches = player:getFishingCatches() or {}
```
claywar, line 130: Need to check this datatype to see if it can be simplified, similar to above

scripts/quests/otherAreas/Fishermans_Heart.lua
paladindamarus: Good to go on the first part, but I'm not sure I understand the second question?  the code for `getFishingStats()` checks the contents of PChar->m_fishHistory, which is a fixed struct.  If the C++ function fails, it returns a table of 0s instead with the same labels.  

```lua
 local fishingStats = player:getFishingStats()
 if fishingStats ~= nil then
     player:updateEvent({
         [0] = player:getFishingCatchCount() or 0,
         [1] = fishingStats['fishLinesCast'] or 0,
         [2] = fishingStats['fishReeled'] or 0,
         [3] = fishingStats['fishLongestLength'] or 0,
         [4] = fishingStats['fishLongestId'] or 0,
         [5] = fishingStats['fishHeaviestWeight'] or 0,
         [6] = fishingStats['fishHeaviestId'] or 0,
     })
 else
     player:updateEvent(0, 0, 0, 0, 0, 0, 0)
 end

File: scripts/quests/otherAreas/Fishermans_Heart.lua (hunk 6 above)
Comment (paladindamarus) [reply scripts/quests/otherAreas/Fishermans_Heart.lua:130]: The 24-byte bitset (for 196 bits) uses 163 of those bits to build a set of 6 32-bit integers as flags for which fish have been caught.  I'm not sure if that's actually _the_ easiest method, but it is how the client receives the data to show "which fish should I put a star next to?"

File: src/map/utils/charutils.cpp [hunk 7]
```diff
@@ -7000,6 +7000,73 @@ namespace charutils
...
+        if (PChar == nullptr)
+        {
+            return;
+        }
+
+        auto Query = "SELECT fish_catches, fish_linescast, fish_reeled, fish_longest, fish_longest_id, fish_heaviest, fish_heaviest_id "
+                     "FROM char_fishing_history "
+                     "WHERE charid = (?);";
+
+        auto ret = db::preparedStmt(Query, PChar->id);
```
zach2good, line 7016: The new db system returns `rset` (result set), the old system returns `ret` (return code)

scripts/quests/otherAreas/Fishermans_Heart.lua
paladindamarus: I completely forgot about this PR because for some reason I thought it got merged -- I see several other comments which I'll have to address, but this one in particular I want to make sure I know what you mean.

What did you have in mind for this?

PR #6236, opened 2024-09-11
[SQL/LUA] Improve Gration Spawn and Correct Drop Rate

LacunaXI:
**_I affirm:_**

- Improves the trade function for Gration and sets the drop rate in the lua depending on the item traded. Previously no matter which was traded to spawn the drop rate in mob_droplist.sql was set to 50%. This then was also being reduced to 24% from the recent Treasure Hunter merge (#6141).
- Removes the droplist from the mob_group.
- Adds the missing gil drop.
- Adds the Tatami Shield drop to scripts\enum.

- Spawn Gration with the Hickory Shield, and get no drop. Spawn it again with a Hickory Shield and get a drop.
- Spawn Gration with the Picaroon's Shield and get a drop.
- Success

### DIFF
--- a/scripts/zones/Misareaux_Coast/npcs/qm_gration.lua
+++ b/scripts/zones/Misareaux_Coast/npcs/qm_gration.lua
@@ -9,11 +9,19 @@ local ID = zones[xi.zone.MISAREAUX_COAST]
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
+    local shieldChance = 0
+    if npcUtil.tradeHasExactly(trade, xi.item.HICKORY_SHIELD) then
+        shieldChance = 500
+    elseif npcUtil.tradeHasExactly(trade, xi.item.PICAROONS_SHIELD) then
+        shieldChance = 1000
+    end
+
     if
-        (npcUtil.tradeHas(trade, xi.item.PICAROONS_SHIELD) or npcUtil.tradeHas(trade, xi.item.HICKORY_SHIELD)) and
+        shieldChance > 0 and
         npcUtil.popFromQM(player, npc, ID.mob.GRATION)
     then
         player:confirmTrade()
+        GetMobByID(ID.mob.GRATION):setLocalVar('DropRate', shieldChance)
     end
 end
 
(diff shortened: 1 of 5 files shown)


sql/mob_droplist.sql
```diff
@@ -10066,8 +10066,7 @@ INSERT INTO `mob_droplist` VALUES (1217,0,0,1000,4374,500); -- Sleepshroom (50.0
 INSERT INTO `mob_droplist` VALUES (1217,0,0,1000,4374,250); -- Sleepshroom (25.0%)
 INSERT INTO `mob_droplist` VALUES (1217,2,0,1000,4374,0);   -- Sleepshroom (Steal)
 
--- ZoneID:  25 - Gration
-INSERT INTO `mob_droplist` VALUES (1218,0,0,1000,12360,500); -- Tatami Shield (50.0%)
+-- 1218 Available
```
WinterSolstice8, line 10069: This may get lost in the spam, usually available slots like this are noted at the end of file

LacunaXI: > This may get lost in the spam, usually available slots like this are noted at the end of file

I was keeping consistent with how this is in the rest of the file. Where the spot is labeled available at where it sits sequentially. If you would like this placed at the end of the file though then that is no problem.

scripts/zones/Misareaux_Coast/npcs/qm_gration.lua
```diff
@@ -9,11 +9,18 @@ local ID = zones[xi.zone.MISAREAUX_COAST]
 local entity = {}
 
 entity.onTrade = function(player, npc, trade)
+    local shieldChance = 0
+    if npcUtil.tradeHasExactly(trade, xi.item.HICKORY_SHIELD) then
+        shieldChance = 500
+    elseif npcUtil.tradeHasExactly(trade, xi.item.PICAROONS_SHIELD) then
+        shieldChance = 1000
+    end
     if
```
Xaver-DaRed, line 19: CI check. You need a newline in between the `end` and the next `if`

sql/mob_droplist.sql
WinterSolstice8: Oh? I don't usually make mobs so you can ignore me then

scripts/zones/Misareaux_Coast/npcs/qm_gration.lua
LacunaXI: Pushed a correction and CI passed :)

Xaver-DaRed (approved):

PR #6239, opened 2024-09-12
[Synthesis] Fix possible underflow at high difficulties

Xaver-DaRed:
**_I affirm:_**

At high difficulties, success rate would underflow, making the synth impossible to break.

Review by commit. Craft way under the recipe level. Better tested with desynths.


zach2good (approved):

zach2good:
Code is fine, but for these commits:
image

There's no context if you were to look through the commit log to know what loop you're simplifying, what underflow, etc.

Just an additional qualifier like `crafting:` would help a lot, or specifying `difficulty roll loop`, etc.

Xaver-DaRed:
Renamed commits. Hopefully they are better now.

PR #6246, opened 2024-09-14
[LUA]Removes defaultAction for shattered teleporters

hooksta4:
**_I affirm:_**

Currently, when a player interacts with the three Shattered Telepoints, it cycles between the default action and the proper cutscene to enter the Hall of Transference. 

Removes the two instances of default Actions in the zone and puts the "default message" into the npc script. 

Adds message special to each of the shattered_telepoint.lua 

If players have not completed The Mothercrystals, they will get the default message.
If players are on the mission or past it, they will get the cs option to investigate further and enter the Hall of Transference.


claywar:
I would prefer to see this handled as another section in CoP 1-3 mission script.  Something along the following lines:
```
    {
        check = function(player, currentMission, missionStatus, vars)
            return player:hasCompletedMission(mission.areaId, mission.missionId)
        end,

        [xi.zone.TAHRONGI_CANYON] =
        {
            ['Shattered_Telepoint'] = mission:progressEvent(913),
        },
    }
```
And so on

hooksta4:
Ok.

PR #6249, opened 2024-09-15
[login] [core] Improve robustness of zoning

WinterSolstice8:
**_I affirm:_**

This commit attempts to improve retail accuracy where retail recovers from packet loss of you never recieving 0x00B As well as the server not caring if 0x00D is sent or not. We predictably know when the key should be incremented, and while not an ideal solution, seems to work

uhhh zone a lot and log in a lot I guess, there's a ton of scenarios like warping, teleporting !zone, /shutdown, etc
create char after logging in normally then logging out
create char right away after logging in

It is very important to test /logout and /shutdown without a GM character

The main one is this:
[1725996787757.webm](https://github.com/user-attachments/assets/e397720d-0f18-4019-a453-39ba673e21ff)

### DIFF
--- a/src/common/blowfish.h
+++ b/src/common/blowfish.h
@@ -29,6 +29,7 @@ enum BLOWFISH
     BLOWFISH_WAITING,
     BLOWFISH_SENT,
     BLOWFISH_ACCEPTED,
+    BLOWFISH_PENDING_ZONE,
 };
 
 struct blowfish_t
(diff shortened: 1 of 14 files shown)


WinterSolstice8:
Fixed the last minute bug I found, I can't seem to find or think of any other issues atm so this is ready for review.

Xaver-DaRed (approved):

PR #6255, opened 2024-09-16
login: Add ACCOUNT_CHARACTER_LIMIT setting

zach2good:
**_I affirm:_**
- [NOT YET LOL] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

Adds this setting:
```lua
    -- Maximum number of characters per account (0 for no limit)
    -- This is enforced through xi_connect and xiloader, you can bypass this if you create characters through other means
    ACCOUNT_CHARACTER_LIMIT = 0,
```

And enforces it during attempted character creation

### DIFF
--- a/settings/default/login.lua
+++ b/settings/default/login.lua
@@ -40,6 +40,10 @@ xi.settings.login =
     -- Number of simultaneous game sessions per IP (0 for no limit)
     LOGIN_LIMIT = 0,
 
+    -- Maximum number of characters per account (0 for no limit)
+    -- This is enforced through xi_connect and xiloader, you can bypass this if you create characters through other means
+    ACCOUNT_CHARACTER_LIMIT = 0,
+
     -- Expansion display on the client's login screen. This does NOT effect in game content whatsoever!
     RISE_OF_ZILART          = true,
     CHAINS_OF_PROMATHIA     = true,
--- a/src/login/view_session.cpp
+++ b/src/login/view_session.cpp
@@ -176,127 +176,163 @@ void view_session::read_func()
         case 0x22: // 34: Checking name and Gold World Pass
         {
             // block creation of character if in maintenance mode or generally disabled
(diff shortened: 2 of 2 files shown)


zach2good:
Xaver just informed me that we dictate how many characters can be made per account with SQL: `accounts.content_ids`. However, we default this to be 16, and we don't offer any way outside of modifying the table definition to change this as characters are being inserted. Should we?

zach2good:
When we create an account, we don't specify content_ids, so it goes to the default (16):
```cpp
                if (_sql->Query("INSERT INTO accounts(id,login,password,timecreate,timelastmodify,status,priv) \
                                VALUES(%d,'%s','%s','%s',NULL,%d,%d)",
                                accid, username, BCrypt::generateHash(escaped_pass), strtimecreate, ACCOUNT_STATUS_CODE::NORMAL, ACCOUNT_PRIVILEGE_CODE::USER) == SQL_ERROR)
                {
                    ref<uint8>(data_, 0) = LOGIN_ERROR_CREATE;
                    do_write(1);
                    return;
                }
```

zach2good:
Do we think it's reasonable to have this setting and default it to 16, and then pass that value in during the `INSERT INTO accounts`? That way it's more obvious, more easily configurable, and it acts as documentation for the account_ids thing - which I didn't know about until I looked at this

zach2good:
As a general point, I don't think we should default any SQL column to a value we provide in the SQL table definition. We should _always_ provide _every_ column value on insert.

WinterSolstice8:
You should also clamp the input here so the players client cant even hit the character create button. This way it's more obvious that they _cant_ create a character because of no free slots

https://github.com/LandSandBoat/server/blob/bc306a88edd75a7191fa7de830807a51ac650f87/src/login/data_session.cpp#L188

zach2good:
This isn't needed, when we have `content_ids` defined in `accounts`

PR #6257, opened 2024-09-18
[BCNM] Add final Waking the Beast BCNM fight (Carbuncle Prime)

TracentEden2:
**_I affirm:_**

This PR is part of a series to add the full Waking the Beast quest. This PR adds the final Carbuncle Prime BCNM in Full Moon Fountain. This is a complex fight consists of 7 different stages. The logic is mostly concentrated in the battlefield script in `onBattlefieldTick` (and private helper functions) as coordination is needed to, for example, spawn and despawn multiple mob groups (during phase changes) and facilitate mobs using simultaneous 2hrs. I think this centralized logic is better than having the logic distributed across many mob scripts which would add complexity and reduce understandability.

The fight logic and flow is based on many captures including by Siknoz ([here](https://youtu.be/IcwxObsxrnE) and [here](https://youtu.be/0g58waSkmxs)), Aether ([here](https://youtu.be/ig4K_C4i4PU)), and KnowOne ([here](https://youtu.be/hxxnCfiG2BY)).

Also the PR includes one small core change: adding a `setSpawnAnimation` function so that the elemental avatars can be spawned with their special spawn animations (as on retail).

Note that this does not actually yet enable the battlefield for normal players as the final PR in the series will be the actual quest.

Fight the BCNM and compare to retail captures


scripts/zones/Full_Moon_Fountain/mobs/Titan_Prime.lua
```diff
@@ -0,0 +1,45 @@
...
+
+entity.onMobDeath = function(mob, player, optParams)
+    if player then
+        player:showText(mob, ID.text.CORRUPTED_POWER)
+    end
+
+    if mob:getLocalVar('onDeathCalled') == 0 then
+        mob:setLocalVar('onDeathCalled', 1)
+        -- set some local variables useful in determining the phases of the fight
+        mob:setLocalVar('wasDefeated', 1)
```
Xaver-DaRed: Whats this local var for? `wasDefeated`

scripts/zones/Full_Moon_Fountain/mobs/Titan_Prime.lua
```diff
@@ -0,0 +1,45 @@
...
+        local battlefield = mob:getBattlefield()
+        if battlefield then
+            battlefield:setLocalVar('avatarsDefeated', battlefield:getLocalVar('avatarsDefeated') + 1)
+        end
+    end
+end
```
Xaver-DaRed, line 33:
```lua
    if player then
        player:showText(mob, ID.text.CORRUPTED_POWER)
    end

    if optParams.isKiller or optParams.noKiller then
        local battlefield = mob:getBattlefield()
        if battlefield then
            battlefield:setLocalVar('avatarsDefeated', battlefield:getLocalVar('avatarsDefeated') + 1)
        end
    end
end
```

Xaver-DaRed: This extends to all mobs, obviously.

TracentEden2: > This extends to all mobs, obviously.

Thanks, as per suggestion from claywar I moved the setting of these local vars from onMobDeath (within the mob files) to a death listener in the battlefield file. Thus the battlefield file has all the logic. I will remember your suggestion for this in future PRs though.

scripts/zones/Full_Moon_Fountain/mobs/Titan_Prime.lua
TracentEden2: > Whats this local var for? `wasDefeated`

That var is to make sure the mob is actually defeated by a player rather than being `isDead()` due to some other condition like the mob is not spawned yet. I added it to handle some specific race conditions with running the fight multiple times. I am not sure if it is 100% needed given the `phase` var as well but I think better to keep it because it makes the logic more explicit.

zach2good (approved):
I'm fine with this, as long as @Xaver-DaRed is too

sql/mob_resistances.sql
```diff
@@ -61,7 +61,7 @@ INSERT INTO `mob_resistances` VALUES (30,'Automaton-Stormwaker',0,0,0,0,0,0,0,0,
 INSERT INTO `mob_resistances` VALUES (31,'Automaton-Valoredge',0,0,0,0,0,0,0,0,0,0,0,0,0,-1,-1,-1,-1,-1,-1,-1,-1);
 INSERT INTO `mob_resistances` VALUES (32,'Avatar-Atomos_Prime',0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0);
 INSERT INTO `mob_resistances` VALUES (33,'Avatar-Alexander_ToAU',0,0,0,0,0,0,0,0,0,0,0,0,0,5,5,5,5,5,5,11,5);
-INSERT INTO `mob_resistances` VALUES (34,'Avatar-Carbuncle_Prime',0,0,0,0,0,0,0,0,0,0,0,0,0,6,6,6,6,6,6,11,0);
+INSERT INTO `mob_resistances` VALUES (34,'Avatar-Carbuncle_Prime',0,0,0,0,0,0,0,0,0,0,0,0,0,4,4,4,4,4,4,11,0);
```
Xaver-DaRed: where were this resistances gotten from?

TracentEden2: > where were this resistances gotten from?

Thanks for the review. I had some logic for the value change, but I cannot find it anymore. In any case the values are close enough that I am fine sticking with the 6 value as I think that is what Jimmayus sheet says. I have updated the PR to stick with the 6 value.

Xaver-DaRed (approved):

PR #6266, opened 2024-09-24
[core] Small counter audit

WinterSolstice8:
**_I affirm:_**

Disable counter with PD up
Disable counter if the defender is stunned/slept etc

Use PD vs a mob with 100% counter and 9999 acc, see it not counter when you attack it
Stun a mob and see it not counter with the above conditions minus PD

### DIFF
--- a/src/map/attack.cpp
+++ b/src/map/attack.cpp
@@ -445,7 +445,8 @@ bool CAttack::CheckCounter()
         return false;
     }
 
-    if (!m_victim->PAI->IsEngaged())
+    // Don't counter if not engaged or stunned, slept, etc.
+    if (!m_victim->PAI->IsEngaged() || m_victim->StatusEffectContainer->HasPreventActionEffect(true))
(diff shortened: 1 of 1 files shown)


zach2good (approved):

TiberonKalkaz:
A+ on getting this fixed
Any need to consider ALL_MISS - the old TOO_HIGH?

My up to 75 cap knowledge says no - anything that is ALL_MISS is not meleeing back a character - rather shifted to Wyrm like magical breath shots - but no idea what post 75 cap used it for.

WinterSolstice8:
yeah, ALL_MISS makes sense too, if you can't hit it period you shouldn't be able to counter it

go ahead and make a PR if you're so inclined

TiberonKalkaz:
Will do - I'll make sure no where else in the chain is accounting for it first.
At a quick glance, getHitRateEx doesnt seem to

PR #6269, opened 2024-09-25
[core] Fix ammo weapon damage (shurikens)

WinterSolstice8:
**_I affirm:_**

FSTR2 and weapon rank were not being used correctly on ammo-only weapons such as shurikens.

/ra a shuriken as NIN99/whm 1

Per Spensa:

``` Fuma damage: 72, weapon rank: 8, fSTR2 cap = 32, pDIF cap = 3.25 + 0.1 = 3.35. Damage = (72 + 32) * 3.35 = 348. Crit damage = 348 * 1.25 = 435.```

Ranged attacks
image
And Daken:
image

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -562,10 +562,14 @@ uint16 CBattleEntity::GetSubWeaponRank()
 
 uint16 CBattleEntity::GetRangedWeaponRank()
 {
-    if (auto* weapon = dynamic_cast<CItemWeapon*>(m_Weapons[SLOT_RANGED]))
+    // Check ranged slot first, otherwise use ammo if it's null
+    CItemEquipment* item = m_Weapons[SLOT_RANGED] ? m_Weapons[SLOT_RANGED] : m_Weapons[SLOT_AMMO];
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/map/entities/battleentity.cpp [hunk 1]
```diff
@@ -566,6 +566,12 @@ uint16 CBattleEntity::GetRangedWeaponRank()
     {
         return (weapon->getDamage() + getMod(Mod::RANGED_DMG_RANK)) / 9;
     }
+
+    if (auto* weapon = dynamic_cast<CItemWeapon*>(m_Weapons[SLOT_AMMO]))
+    {
+        return (weapon->getDamage() + getMod(Mod::RANGED_DMG_RANK)) / 9;
+    }
```
Frankie-hz: Since it returns the same damage values is it cleaner if you use `(m_Weapons[SLOT_RANGED] ? m_Weapons[SLOT_RANGED] : m_Weapons[SLOT_AMMO])` inside the if before this?

WinterSolstice8: suppose it doesn't matter as long as there's a dynamic cast

WinterSolstice8: alright, changed it

zach2good (approved):

PR #6272, opened 2024-09-27
[core] Session cleanup and improvements

WinterSolstice8:
**_I affirm:_**

When you login, the server pulls a charid from your login packet, but due to the robust login packet changes, sometimes there's junk data that looks decryptable but isn't
This should severely limit any junk making it across as charid

Cleanup session immediately instead of deffering cleanup to a secondary cleanup if session timed out.
Remove accounts_session immediately on timeout

Remove a bunch of excess/copy pasted code from session cleanup

Login as normal when zoning,

Disconnect with the following methods
* standard force d/c
* /logout
* /shutdown
* block outgoing C2S packets after casting warp/teleports/etc

### DIFF
--- a/src/common/mmo.h
+++ b/src/common/mmo.h
@@ -665,4 +665,26 @@ public:
     ~char_mini(){};
 };
 
+// https://github.com/atom0s/XiPackets/tree/main/world/client/0x000A
+// Defined here for use in both map.cpp and packet_system.cpp
+struct GP_CLI_LOGIN
+{
+    uint16_t id : 9;
+    uint16_t size : 7;
+    uint16_t sync;
+    uint8_t  LoginPacketCheck; // PS2: LoginPacketCheck
+    uint8_t  padding00;        // PS2: dam__
+    uint16_t unknown00;        // PS2: MyPort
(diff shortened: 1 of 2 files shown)


claywar (dismissed):

WinterSolstice8:
Chasing some bug here and finding other stuff that needs cleanup, currently drafted

WinterSolstice8:
Probably done with PR because it's getting long in the tooth but will take some more time to test

For reviewing the changes to `map_cleanup`, better to view the file as-is with this PR and compare to the old one. I more or less just squished it down into a much more readable less else/if copypaste.

WinterSolstice8:
Seems to be fine

WinterSolstice8:
I lied. I pushed one more commit because I realized I could actually validate 0x00As in some meaningful way.

Xaver-DaRed (approved):
Shameless disclaimer: My approval is out of respect for you and your work, not because I actually understand what's going on here.

zach2good (approved):

WinterSolstice8:
Squashed some commits, but code unchanged

zach2good (approved):

zach2good (approved):

PR #6274, opened 2024-09-27
Movement speed adjustments

cocosolos:
**_I affirm:_**

Fixes some issues with various speed related things:

- Fix `MOVE_SPEED_STACKABLE` being cast as unsigned. Need this for gear reductions.
- Use signed int for intermediate calculations. Also bumped the size up so we can work with the full unsigned 8-bit range before clamping.
- Quickening effect should only apply if speed > 0 based on retail research.
- Adjust Flee mod for more precision. Hermes Quencher was found to be +87.5% flee effect through retail testing.
- Adjusts various speed related mods and latents based on retail testing. Councilor's Garb and Ninja Kyahan were found to be +24% movement speed, so I extrapolated that to all +25% gear.
- Add new latent for Councilor's Garb. This will likely need some kind of adjustment when Alliance Shirt +1 is implemented as that one doesn't work in Mog Garden.

You need an addon to either show your speed or capture and display incoming packet 0x037.

```
!changejob cor 99
!additem 5497
/item "Bolter's Die" <me>
/ja "Bolter's Roll" <me>
```

Note the increase in speed.

```
!changejob war 99
!changesjob thf 49
!additem 18519
!additem 11817
!additem 12573
!additem 12701
!additem 16344
!additem 10372
!additem 10980
```

Equip all of that gear. That should bring you to -4 speed, plus whatever that noted increase from Bolter's Roll was.

```
/ja Flee <me>
```

Note your speed should have decreased.

```
!additem 27923
!zone 256
```

Equip Councilor's Garb, note speed increase,

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -243,7 +243,7 @@ int32 CBattleEntity::GetMaxMP() const
 uint8 CBattleEntity::GetSpeed()
 {
     uint8 baseSpeed   = speed;
-    uint8 outputSpeed = 0;
+    int16 outputSpeed = 0;
 
     // Mount speed. Independent from regular speed and unaffected by most things.
     // Note: retail treats mounted speed as double what it actually is! 40 is in fact retail accurate!
@@ -255,8 +255,8 @@ uint8 CBattleEntity::GetSpeed()
         return std::clamp<uint8>(outputSpeed, std::numeric_limits<uint8>::min(), std::numeric_limits<uint8>::max());
     }
 
-    // Gear penalties, Bolters Roll.
-    uint8 additiveMods = static_cast<uint8>(getMod(Mod::MOVE_SPEED_STACKABLE));
+    // Gear penalties.
+    int8 additiveMods = static_cast<int8>(getMod(Mod::MOVE_SPEED_STACKABLE));
 
     // Gravity and Curse. They seem additive to each other and the sum seems to be multiplicative.
     float weightFactor = std::clamp<float>(1.0f - static_cast<float>(getMod(Mod::MOVE_SPEED_WEIGHT_PENALTY)) / 100.0f, 0.1f, 1.0f);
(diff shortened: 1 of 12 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -282,17 +302,20 @@ uint8 CBattleEntity::GetSpeed()
     uint8 mazurkaQuickeningEffect = std::clamp<uint8>(getMod(Mod::MOVE_SPEED_QUICKENING) + getMod(Mod::MOVE_SPEED_MAZURKA), 0, 10);
 
     // We have all the modifiers needed. Calculate final speed.
-    // This MUST BE DONE IN THIS ORDER. Using uint8 data type, we use that to floor.
+    // This MUST BE DONE IN THIS ORDER. Using int8 data type, we use that to floor.
     outputSpeed = baseSpeed + additiveMods;
```
Xaver-DaRed, line 306: This will need a clamp. This should not be going bellow 0, ever.

Actually, all of them will need clamps to ensure value never goes below 0.

cocosolos: It does. Retail testing:
```
+14 from bolters roll turns into +10 at -54 movement speed
-54 gear + 15 bolters + flee = 7 speed (-4 * 2 + 15)
```

cocosolos: Gravity is also effected by this. With negative speed + bolter's roll, gravity will actually increase your speed.

Xaver-DaRed: And that is the exact point I will just ignore retail accuracy in favor of sanity. Thats obviously not intended behavior, I would think

cocosolos: It gets clamped before leaving the function. I can't see how this would break anything. I'd really prefer retail accuracy.

cocosolos: The most you can get from gear is -65 reduction, so the max negative you can get would be like -39 with cheer, gear and flee. Which will ultimately be 0. I don't think there's any danger of abuse. We're talking about a 1-2 speed difference from gravity in this case.

cocosolos: Let's say we have -65 from gear, -99% gravity, +3% cheer, +24% gear, +100% flee, +36 max bolter's roll (+10 from XI roll, +10 crooked cards, +16 Rostam), +10 quickening. You're looking at a final 54 speed. But consider the setup for that. Taking any attack will remove bolter's and quickening, bringing you back down to 0. You would have to get a stronger gravity than anything actually does and then roll a perfect roll all while wearing no defensive gear to see any effect, and at most you're still running at basically normal speed. I'm more interested in an accurate emulator than preventing some infinitesimally small chance of impractical abuse.

cocosolos: And just as an additional note: using an unsigned int for the above situation would result in a 59 final speed, higher than the "exploit" speed.

Xaver-DaRed (approved):

src/map/entities/battleentity.cpp
Xaver-DaRed: Not exactly my point, just poorly worded on my end.

My point is, I dont believe SE actually intended for gravity to ever increase your speed. I believe its an edge case that results in ununtended behavior on their end.

Such cases we usually go with the "better than retail" approach. In this case it may not be something big or concerning, however, my OCD gets triggered by it.

cocosolos: It really only happens because bolter's roll is unconditionally applied. We _could_  add a condition similar to quickening where it's only applied when speed >= 0, but I still feel like it's unnecessary. I'm a big believer in emulation accuracy, faults and all.

PR #6275, opened 2024-09-27
CI: Update GH Ubuntu runner to 24.04 + Compilers

### LINKED ISSUE #4670: 🔨 Ubuntu 24.04 LTS (Noble Numbat) Upgrade (eta. April 25, 2024)
**_I affirm:_**

## Describe the feature

Following the same pattern we've followed since like, 2020: When the latest ubuntu LTS comes out AND BECOMES AVAILABLE AS A GITHUB ACTIONS RUNNER: we will upgrade and pick whatever compilers and toolchain things that come packaged with that runner - and then update our Windows requirements as necessary.

This will include, but not be limited to:
- C++ compiler
- Use of std::format, std::expected, and other goodies that will come from having C++20 across the board
- All the additional errors and warnings that will become available with new compiler tooling
- New Python, etc.
- Not having to build LuaJIT on certain runners in CI
- Updating the version of spdlog we use to use the version that uses std::format and not fmtlib
- Check versions of MariaDB, etc.
- OpenSSL, etc.

zach2good:
**_I affirm:_**

Just by chance I thought about "is that runner out of beta yet?", and it is. So, updating the runners, and maxing out the available C++ compilers for CI. I suspect this will be the CI dance for a bit with versions of Python, other libraries, and possibly me trying to uncomment the "cache build folder" step.

Closes https://github.com/LandSandBoat/server/issues/4670

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

`Clang-18`
`GCC-14`
`Python 3.12`
`mariadb-client-10.6` -> `mariadb-client`
`mariadb-server-10.6` -> `mariadb-server`
Noble apt apparently doesn't have these mariadb packages anymore?

Started opening issues on the mariadb-cpp connector, but will likely maintain my own fork because I can't grok how quickly they work. It looks like it just has a single maintainer?
- https://jira.mariadb.org/browse/CONCPP-134


zach2good:
I'm expanding the scope of this PR to also be: Make all CI tooling easier and obvious to use from your local machine.

clanofartisans:
As noted in the original message, when I run `sudo ./mariadb_repo_setup --mariadb-server-version="10.6"` from a 24.04 machine, I get told that the version number isn't valid. Looks like they're not targeting that particular combination anymore?

zach2good:
Replaced by: https://github.com/LandSandBoat/server/pull/6604

PR #6277, opened 2024-09-27
[Lua] Fix Proto-Omega movement speed and gravity immunity

TracentEden2:
**_I affirm:_**

This PR fixes a few small issues with Proto-Omega:

1. Proto-Omega should be immune to Gravity (see this [capture](https://youtu.be/qk-f6a1kBBg))
2. Proto-Omega should have a base speed of 60 during all phases (with boosting up to 150 speed, which is a normal 2.5x multiplier, see this capture [video](https://youtu.be/0g_vzzSTrE8?si=Hws2gpAxoUm01uXn) and [packets](https://drive.google.com/file/d/1rIniG1aN6V6qVAthP1Q4jGseK6up1Jqv/view?usp=sharing). Also note that LSB has built in speed boosting though it is not fully working at the moment. 
3. I also add a TODO on gunpod as it has a normal base speed of 40 but a non-standard speed boost multiplier of 5x, which can be added after custom speed boost multiplier functionality is added to LSB.

Fight Proto-Omega and cast gravity on it and then kite it around to see speed


sql/mob_pools.sql
```diff
@@ -3261,7 +3261,7 @@ INSERT INTO `mob_pools` VALUES (3204,'Progenerator','Progenerator',138,0x00006D0
 INSERT INTO `mob_pools` VALUES (3205,'Promathia','Promathia',134,0x0000300400000000000000000000000000000000,7,7,3,240,100,0,0,0,0,18,6172,32,3979,155,0,0,127,0,0,134,134);
 INSERT INTO `mob_pools` VALUES (3206,'Propagator','Propagator',138,0x00006D0400000000000000000000000000000000,1,1,3,240,100,0,1,1,0,2,0,0,2629,135,13,0,0,0,0,138,138);
 INSERT INTO `mob_pools` VALUES (3207,'Proteus','Proteus',218,0x0000610100000000000000000000000000000000,2,2,7,360,100,0,1,0,0,2,0,32,236,157,0,0,0,0,0,218,218);
-INSERT INTO `mob_pools` VALUES (3208,'Proto-Omega','Proto-Omega',54,0x0000370400000000000000000000000000000000,2,1,6,240,100,1024,1,1,0,18,6228,32,2773,157,1,0,0,0,0,727,518);
+INSERT INTO `mob_pools` VALUES (3208,'Proto-Omega','Proto-Omega',54,0x0000370400000000000000000000000000000000,2,1,6,240,100,1024,1,1,0,18,6230,32,2773,157,1,0,0,0,0,727,518);
```
MowFord, line 3264: I believe the preference going forward was to use `onMobInitialize` to set immunities and define them as zero in sql for NMs

TracentEden2: > I believe the preference going forward was to use `onMobInitialize` to set immunities and define them as zero in sql for NMs

Yeah, that is mainly what we do on ASB/Horizon but I was not sure about on LSB. I will modify the PR to use that notation.

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #6282, opened 2024-09-29
[core] Move some character cleanup from zone into char destructor

WinterSolstice8:
**_I affirm:_**

Move some character cleanup that was in CZone for some reason into CCharEntity's destructor
This also fixes a rare corner case crash wherein a char was not inserted into a zone but had linkshells/unity chat etc

Zone in/out, with and without a party, with and without a pet, as PUP or not PUP, with and without level sync, etc and see no new issues

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -71,6 +71,8 @@
 #include "items/item_weapon.h"
 #include "job_points.h"
 #include "latent_effect_container.h"
+#include "linkshell.h"
+#include "message.h"
 #include "mobskill.h"
 #include "modifier.h"
 #include "packets/char_job_extra.h"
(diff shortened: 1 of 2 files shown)


WinterSolstice8:
Most of things should've already been in the char destructor but weren't for some reason. I've done some light testing with Valgrind so far and saw no leaks, but still need to test a bit more. It's difficult to keep valgrind running for long enough to actually do testing with multiple clients

WinterSolstice8:
This doesn't seem to have any memory leaks in valgrind and no other obvious problems. I figured out how to make valgrind run a LOT longer if you simply load only one zone instead of all of them. Bonus points for it taking a LOT less time to start too.

Xaver-DaRed (approved):

zach2good (approved):

PR #6283, opened 2024-09-29
Salvage II plans currency storage handling

clanofartisans:
**_I affirm:_**

Implements the very basics of handling the storage of "plans" from Salvage II, specifically the database storage and reporting in the currency packet.

Give yourself some stored plans (I just manually added some in the database). Log in and check your Currency menu to see if it matches the assigned values.
image
image

### DIFF
--- a/src/map/packets/currency1.cpp
+++ b/src/map/packets/currency1.cpp
@@ -19,154 +19,162 @@
 ===========================================================================
 */
 
-#include "common/socket.h"
-
 #include "currency1.h"
 
+#include "common/cbasetypes.h"
 #include "entities/charentity.h"
 #include "utils/charutils.h"
+
+// TODO: implement the packet wholesale
+// from https://github.com/atom0s/XiPackets/tree/main/world/server/0x0113
+// clang-format off
+struct bitPackedCurrency1
+{
+    uint64_t bloodshed_plans_stored   : 9;
+    uint64_t umbrage_plans_stored     : 9;
+    uint64_t ritualistic_plans_stored : 9;
+    uint64_t tutelary_plans_stored    : 9;
+    uint64_t primacy_plans_stored     : 9;
+    uint64_t unused                   : 19;
+};
+// clang-format on
 
 CCurrencyPacket1::CCurrencyPacket1(CCharEntity* PChar)
 {
     this->setType(0x113);
     this->setSize(252);
 
-    const char* query = "SELECT sandoria_cp, bastok_cp, windurst_cp, beastman_seal, kindred_seal, kindred_crest, \
-                        high_kindred_crest, sacred_kindred_crest, ancient_beastcoin, valor_point, scyld, \
-                        guild_fishing, guild_woodworking, guild_smithing, guild_goldsmithing, guild_weaving, \
-                        guild_leathercraft, guild_bonecraft, guild_alchemy, guild_cooking, cinder, fire_fewell, \
-                        ice_fewell, wind_fewell, earth_fewell, lightning_fewell, water_fewell, light_fewell, \
-                        dark_fewell, ballista_point, fellow_point, chocobuck_sandoria, chocobuck_bastok, \
-                        chocobuck_windurst, daily_tally, research_mark, tunnel_worm, morion_worm, phantom_worm, \
-                        moblin_marble, infamy, prestige, legion_point, spark_of_eminence, shining_star, \
-                        imperial_standing, leujaoam_assault_point, mamool_assault_point, lebros_assault_point, \
-                        periqia_assault_point, ilrusi_assault_point, nyzul_isle_assault_point, zeni_point, jetton, \
-                        therion_ichor, allied_notes, aman_vouchers, login_points, cruor, resistance_credit, \
-                        dominion_note, fifth_echelon_trophy, fourth_echelon_trophy, third_echelon_trophy, \
-                        second_echelon_trophy, first_echelon_trophy, cave_points, id_tags, op_credits, \
-                        voidstones, kupofried_corundums, pheromone_sacks, rems_ch1, rems_ch2, \
-                        rems_ch3, rems_ch4, rems_ch5, rems_ch6, rems_ch7, rems_ch8, rems_ch9, rems_ch10, \
-                        reclamation_marks, unity_accolades, fire_crystals, ice_crystals, wind_crystals, \
-                        earth_crystals, lightning_crystals, water_crystals, light_crystals, dark_crystals, deeds \
-                        FROM char_points WHERE charid = %d";
+    auto rset = db::preparedStmt("SELECT * FROM char_points WHERE charid = ?", PChar->id);
 
-    int ret = _sql->Query(query, PChar->id);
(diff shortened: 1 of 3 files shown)


WinterSolstice8:
char_points table is protected so this needs a migration to add fields (see tools/migrations)

src/map/packets/currency1.cpp
```diff
@@ -153,6 +154,14 @@ CCurrencyPacket1::CCurrencyPacket1(CCharEntity* PChar)
         ref<uint8>(0xD6) = _sql->GetUIntData(80); // rems_ch9
         ref<uint8>(0xD7) = _sql->GetUIntData(81); // rems_ch10
 
+        // Salvage II Stored Plans
+        ref<uint64>(0xD8) = PackSalvageIIPlans(
```
WinterSolstice8: you can simplify this by defining an enum class defined in this cpp file:

```
// TODO: implement the packet wholesale
// from https://github.com/atom0s/XiPackets/tree/main/world/server/0x0113
struct bitPackedCurrency1
{
    uint64_t bloodshed_plans_stored   : 9;
    uint64_t umbrage_plans_stored     : 9;
    uint64_t ritualistic_plans_stored : 9;
    uint64_t tutelary_plans_stored    : 9;
    uint64_t primacy_plans_stored     : 9;
    uint64_t unused                   : 19;
};
```

Make sure to co-author atom0s by putting this in your commit text:
`Co-authored-by: atom0s <atom0s@live.com>`

src/map/packets/currency1.cpp
```diff
@@ -153,6 +154,14 @@ CCurrencyPacket1::CCurrencyPacket1(CCharEntity* PChar)
...
+            _sql->GetUIntData(93),
+            _sql->GetUIntData(94),
+            _sql->GetUIntData(95),
+            _sql->GetUIntData(96),
+            _sql->GetUIntData(97));
```
zach2good: While you're in here, you may as well convert all of this to use the new `db::` system, so you don't have to track the indexes of each column, you can just pull them out by name

src/map/packets/currency1.cpp
zach2good: +1 for this

zach2good (commented):
+1 for this should also have a migration, I think

zach2good:
Honestly, I thought we had all the currencies hooked up 🤷

clanofartisans:
> +1 for this should also have a migration, I think

Oops, you and Winter are right. It totally slipped my mind, thank you!

clanofartisans:
> char_points table is protected so this needs a migration to add fields (see tools/migrations)

I've created the missing migration and tested it.

src/map/packets/currency1.cpp
clanofartisans: > you can simplify this by defining an enum class defined in this cpp file:
> 
> ```
> // TODO: implement the packet wholesale
> // from https://github.com/atom0s/XiPackets/tree/main/world/server/0x0113
> struct bitPackedCurrency1
> {
>     uint64_t bloodshed_plans_stored   : 9;
>     uint64_t umbrage_plans_stored     : 9;
>     uint64_t ritualistic_plans_stored : 9;
>     uint64_t tutelary_plans_stored    : 9;
>     uint64_t primacy_plans_stored     : 9;
>     uint64_t unused                   : 19;
> };
> ```
> 
> Make sure to co-author atom0s by putting this in your commit text: `Co-authored-by: atom0s <atom0s@live.com>`

I've added the struct and updated the code to use it. I also made sure to credit @atom0s <3

src/map/packets/currency1.cpp
clanofartisans: > While you're in here, you may as well convert all of this to use the new `db::` system, so you don't have to track the indexes of each column, you can just pull them out by name

I've updated the cpp file to use the new `db::` system and things are still working as expected when testing!

Unrelated but is there a reason that the `plaudits` column is signed and nullable? Shouldn't really affect anything but seemed like it was probably an oversight.

src/map/packets/currency1.cpp
```diff
@@ -26,12 +26,26 @@
 #include "entities/charentity.h"
 #include "utils/charutils.h"
 
+// TODO: implement the packet wholesale
+// from https://github.com/atom0s/XiPackets/tree/main/world/server/0x0113
+// clang-format off
+struct bitPackedCurrency1
```
WinterSolstice8, line 31: I dont think you need to turn off clang format for this

iirc from other PRs I did (purely from memory) it wants the bits not aligned for some reason

``` 
    uint64_t bloodshed_plans_stored : 9;
    uint64_t umbrage_plans_stored : 9;
    uint64_t ritualistic_plans_stored : 9;
    uint64_t tutelary_plans_stored : 9;
    uint64_t primacy_plans_stored : 9;
    uint64_t unused : 19;
```

but that said i dont really care either way since alignment is fine

WinterSolstice8 (approved):

src/map/packets/currency1.cpp
```diff
@@ -47,127 +61,153 @@ CCurrencyPacket1::CCurrencyPacket1(CCharEntity* PChar)
...
+        ref<uint16>(0xF4) = static_cast<uint16>(rset->getUInt("light_crystals"));
+        ref<uint16>(0xF6) = static_cast<uint16>(rset->getUInt("dark_crystals"));
+
+        ref<uint16>(0xF8) = static_cast<uint16>(rset->getUInt("deeds"));
     }
 
     ref<uint32>(0xC0) = charutils::getAvailableTraverserStones(PChar); // traverser_stones
 }
+
+uint64 CCurrencyPacket1::PackSalvageIIPlans(uint32 bloodshed, uint32 umbrage, uint32 ritualistic, uint32 tutelary, uint32 primacy)
```
zach2good: You can get rid of the packing function here with named struct initialiser syntax, I reckon

Xaver-DaRed (approved):

src/map/packets/currency1.cpp
clanofartisans: > You can get rid of the packing function here with named struct initialiser syntax, I reckon

I think I've got that re-done correctly. The CIs weren't happy with my use of `reinterpret_cast` previously so, if it passes CIs this time, I think I'm good to squash and move forward if you like the changes.

clanofartisans: Nope, it griped about strict aliasing again. Would my previous commit work? I can kill the second one if so. Am open to suggestions otherwise!

zach2good:
I've updated your PR to use prepared statements, simplified the query, and used the bit-packed currency section with 
designated initializer syntax

zach2good:
(I haven't tested it though, so someone needs to smoke test it)

clanofartisans:
> (I haven't tested it though, so someone needs to smoke test it)

Looks much better! I was curious why the query wasn't just selecting `*` so I'm glad you cleaned that up. Appreciate your help finishing this off! You actually missed passing the `PChar->id` to the prepared statement but that was easy to fix. Thanks again!

zach2good (approved):

PR #6287, opened 2024-09-29
[core] Update STR to ATT values by weapon type

Xaver-DaRed:
**_I affirm:_**

https://www.bg-wiki.com/ffxi/Strength

We were, seemingly, using outdated valies for the STR effect on attack, according to BG wiki.
We wee also, seemingly, not making a distinction between main and off-hand?

I'm rquesting WinterS to review this, since he is the most knowledgable on this stuff. Sorry, Winter ><

Note: The else/if chain could be optimized, but I figured having clear cases would be easier to review whats going on.

Attack without imploding

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -779,17 +779,27 @@ uint16 CBattleEntity::ATT(SLOTTYPE slot)
     int32 ATT    = 8 + m_modStat[Mod::ATT];
     auto  ATTP   = m_modStat[Mod::ATTP];
     auto* weapon = dynamic_cast<CItemWeapon*>(m_Weapons[slot]);
-    if (weapon && weapon->isTwoHanded())
+
+    // https://www.bg-wiki.com/ffxi/Strength
+    if (weapon && weapon->isTwoHanded()) // 2-handed weapon
     {
-        ATT += (STR() * 3) / 4;
+        ATT += STR();
     }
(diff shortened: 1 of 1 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -779,17 +779,27 @@ uint16 CBattleEntity::ATT(SLOTTYPE slot)
...
+        ATT += STR();
     }
-    else if (weapon && weapon->isHandToHand())
+    else if (weapon && weapon->isHandToHand()) // H2H Weapon
     {
-        ATT += (STR() * 5) / 8;
+        ATT += STR() * 3 / 4;
     }
-    else
+    else if (slot == SLOT_RANGED) // Ranged weapon.
```
WinterSolstice8: Don't forget ammo

Xaver-DaRed: added ammo aswell. I assume its the same as regular ranged

WinterSolstice8 (approved):

WinterSolstice8:
Hmm. BG wiki and jp wiki disagree on h2h. JP wiki says .75 like here but bg wiki says it's 1:1 just like two hand and main handed 1h weapon

WinterSolstice8:
Lets see here
image

8 + 16 H2H skill + 127 = 151 + gifts of 50 = 201


8 + 16 H2H skill + 127*.75 = 119.25 + gifts of 50 = 169.25

jp wiki is correct

Issue #4577, opened 2023-10-02
🐛 Performance report: SQL Query took x ms:


CatsEyeXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)
Windows10

## Branch affected by issue 

`base`

## Steps to reproduce
Scraped map server logs for "SQL query took". Wanted to share here in case there was anything that could be improved on, or shed some light on a potential issue.

Map servers and MariaDB are on the same box communicating over loopback address.

```
SQL query took 2881ms: INSERT INTO char_equip SET charid = 12063, equipslotid = 10 , slotid  = 19, containerid = 8 ON DUPLICATE KEY UPDATE slotid  = 19, containerid = 8;
SQL query took 107ms: INSERT INTO char_equip SET charid = 12063, equipslotid = 11 , slotid  = 60, containerid = 10 ON DUPLICATE KEY UPDATE slotid  = 60, containerid = 10;
[map][warn] SQL query took 133ms: UPDATE char_look SET head = 20, body = 129, hands = 21, legs = 88, feet = 6, main = 265, sub = 36, ranged = 0 WHERE charid = 14338; (SqlConnection::QueryStr:352)
[map][warn] SQL query took 243ms: UPDATE chars SET isstylelocked = 0 WHERE charid = 14338; (SqlConnection::QueryStr:352)
[map][warn] SQL query took 148ms: INSERT INTO char_equip SET charid = 14338, equipslotid = 8 , slotid  = 22, containerid = 8 ON DUPLICATE KEY UPDATE slotid  = 22, containerid = 8; (SqlConnection::QueryStr:352)
[map][warn] SQL query took 123ms: UPDATE chars SET pos_zone = 103,pos_prevzone = 245,pos_rot = 0,pos_x = 128.163,pos_y = -7.320,pos_z = 95.083,moghouse = 0,boundary = 0 WHERE charid = 2161; (SqlConnection::QueryStr:352)
[...]

PR #6300, opened 2024-10-02
db: Add ResultSetWrapper type

zach2good:
**_I affirm:_**

Changes syntax from:
```cpp
    auto zoneId = rset->getUInt("pos_zone");
```

to

```cpp
    auto zoneId = rset->get<uint16>("pos_zone");
```

Related issues: 
https://github.com/LandSandBoat/server/issues/4577
https://github.com/LandSandBoat/server/issues/6355

Testing
- [x] Login: Connect, Select Character, Enter game
- [x] Login: Create new character
- [x] Login: Delete character
- [x] Search: /sea all
- [x] Search: List and buy AH item
- [x] World: Send self a /tell
- [x] World: Send others a /tell
- [x] World: Party mechanics
- [x] Map: First time Login and CS
- [x] Map: Zoning
- [x] Map: GM commands, givegil, zoning, ah, etc.
- [x] Map: Stats are as they should be
- [x] HTTP Server works
- [x] Message Server works
- [x] Daily Tally works
- [x] Conquest System "works" as before
- [x] Synergy recipes are loaded
- [x] PUP attachments are loaded
- [x] Status effects work
- [x] Currency menus work
- [x] Augments work
- [x] Leave running overnight, see if there are ping things, etc.
- [x] You can enter and leave monstrosity correctly as your monster
- [x] Database metadata like names and versions are available to db::
- [x] Validation, reconnect, pinging, etc.
- [x] Charset validation
- [x] Keepalive settings
- [x] Char encoding
- [x] Autocommit func helpers
- [x] Transactions
- [x] Sprintf in stmt and query
- [x] Profiling


zach2good:
While I'm here I'm also going to convert the SQL queries in charutils to be preparedstmts: https://github.com/LandSandBoat/server/issues/4577

zach2good:
For my own peace of mind, I'm probably also going to add a test executable that communicates with a temporary test database, and writes/reads things into the db and confirms they've made it back OK.

I can also do comparative testing with the current db stuff. Once everything is confirmed, I can push harder to replace the existing system.

TODO:
- String escaping
- Validation/reconnection/pinging
- Charset stuff
- Keepalive settings
- Encoding
```
    std::string GetDatabaseName();
    std::string GetClientVersion();
    std::string GetServerVersion();
```
- Query helpers with fmt and sprintf built in
- autocommit
- transaction stuff
- profiling tools

zach2good:
Even if I'm considering moving to a different SQL connector lib in the future, this is still useful as a layer of abstraction on top so the main code only changes once and I can futz with the underlying driver code

zach2good:
Ready for review, I've gone through my testing list and everything is working as expected. Leaving it overnight tonight to see how it handles things

claywar (approved):

PR #6302, opened 2024-10-02
[Combat] Create status effect file and magic cleanup

Xaver-DaRed:
**_I affirm:_**

An effort to make actions as independent as possible.
- The purpose of this new file is to have a centralized place to handle checking for all the variables and modifiers associated with each status effect.
- Created a new, more flexible magic hit rate helper function, able to deal with both nukes and enfeebles with minimal action data.
- Put both new things to practice in all existing systems that used the originals.

Next PR will deal with extending the new magic hit rate function to all actions that use other macc calculations. I didnt want to bloat this PR too much, since its already a lot to take in.

Cast spells. No changes should be observed.

### DIFF
--- a/scripts/globals/combat/status_effect_tables.lua
+++ b/scripts/globals/combat/status_effect_tables.lua
@@ -81,7 +81,7 @@ end
 -- Helper functions to check target effect nullification.
 -----------------------------------
 xi.combat.statusEffect.isTargetImmune = function(target, effectId, actionElement)
-    if target:isMob() then
+    if not target:isMob() then
         return false
     end
 
--- a/scripts/globals/combat/status_effect_tables.lua
+++ b/scripts/globals/combat/status_effect_tables.lua
@@ -76,3 +76,47 @@ xi.combat.statusEffect.getAssociatedImmunobreakModifier = function(effectId)
 
     return associatedImmunobreakModifier
 end
+
+-----------------------------------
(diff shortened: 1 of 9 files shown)


scripts/globals/combat/status_effect_tables.lua
```diff
@@ -0,0 +1,122 @@
...
+    [xi.effect.SLOW         ] = { xi.immunity.SLOW,       xi.mod.SLOWRES,     xi.mod.SLOW_MEVA,     xi.mod.SLOW_IMMUNOBREAK     },
+    [xi.effect.STUN         ] = { xi.immunity.STUN,       xi.mod.STUNRES,     xi.mod.STUN_MEVA,     0                           },
+    [xi.effect.WEIGHT       ] = { xi.immunity.GRAVITY,    xi.mod.GRAVITYRES,  xi.mod.GRAVITY_MEVA,  xi.mod.GRAVITY_IMMUNOBREAK  },
+}
+
+-----------------------------------
+-- Helper functions to easily fetch table data.
+-----------------------------------
+xi.combat.statusEffect.getAssociatedImmunity = function(effectId, actionElement)
+    -- Sanitize fed values
```
zach2good, line 32: Liking this `Sanitize fed values` pattern 👍

zach2good (approved):

PR #6303, opened 2024-10-02
[Core] Add retail mob movement speed boost mechanism

TracentEden2:
**_I affirm:_**

This PR adds a retail accurate version of the mob movement speed boost. Essentially, on retail a mob boosts it's speed (almost always by 2.5x in retail) when it's target is out of melee range. This boost can be seen in packet captures of mobs when kiting (as speed is included in mob update packets). This speed boost is applied after other speed modifications (such as gravity), however if the mob has a negative speed modification then the boost is only about half as powerful (instead of 2.5x it is 1.2x). This smaller boost means that spells such as gravity are still useful when kiting (otherwise they would be worthless).

The implementation includes settings in the map settings file to turn the system on/off and to set the default multiplier (to between 1 to 5, the default is the retail value of 2.5). Also a mob mod is added for special cases where the boost multiplier of a mob is not 2.5x. The only mobs verified to have non standard multipliers are Gunpod and maybe Cactuar Rapido.

The system is based on captures including kiting proto-omega ([video](https://youtu.be/0g_vzzSTrE8?si=Hws2gpAxoUm01uXn) and [packets](https://drive.google.com/file/d/1rIniG1aN6V6qVAthP1Q4jGseK6up1Jqv/view?usp=sharing)) and several other captures I am still uploading.

Also the PR adds a check to only apply the speed cap of 80 to players (as no evidence is applies to others and certain very fast mobs can have high base speeds).

Kite mobs with and without gravity

### DIFF
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -26,6 +26,8 @@
 #include "entities/baseentity.h"
 #include "entities/mobentity.h"
 #include "lua/luautils.h"
+#include "mob_modifier.h"
+#include "status_effect_container.h"
 #include "zone.h"
 
 #include <cfloat>
@@ -578,15 +580,54 @@ float CPathFind::GetRealSpeed()
         {
             realSpeed = 20;
         }
-        else if (m_POwner->animation == ANIMATION_ATTACK)
-        {
-            auto speedMod = settings::get<int8>("map.MOB_SPEED_MOD");
-            if (speedMod < -90)
-            {
-                speedMod = -90;
-            }
 
-            realSpeed *= 1.0f + speedMod / 100.0f;
+        if (settings::get<bool>("map.USE_MOB_SPEED_BOOST_MULTIPLIER"))
+        {
+            if (auto* mobEntity = dynamic_cast<CMobEntity*>(m_POwner))
+            {
+                bool mobIsPlayerPet = mobEntity->PMaster && mobEntity->PMaster->objtype == TYPE_PC;
+
+                auto* battleTarget         = mobEntity->GetBattleTarget();
+                bool  targetWouldBeDrawnin = battleTarget && (distance(mobEntity->loc.p, battleTarget->loc.p) - battleTarget->m_ModelRadius) > mobEntity->GetMeleeRange();
+                bool  isEngaged            = mobEntity->animation == ANIMATION_ATTACK;
+
+                if (isEngaged && !mobIsPlayerPet && targetWouldBeDrawnin && realSpeed > 0)
+                {
+                    // use default multiplier for most mobs
+                    uint16 intMultiplier = settings::get<uint16>("map.DEFAULT_MOB_SPEED_BOOST_MULTIPLIER");
+
+                    // unless the mob has a custom multiplier
+                    if (mobEntity->getMobMod(MOBMOD_SPEED_BOOST_MULT) > 0)
+                    {
+                        intMultiplier = static_cast<uint16>(mobEntity->getMobMod(MOBMOD_SPEED_BOOST_MULT));
+                    }
+
+                    // add warning for when speed boost multipler is lower than 1x as likely misconfiguration
+                    if (intMultiplier < 100 || intMultiplier > 25500)
+                    {
+                        ShowWarning("%s speed boost multiplier is lower than 100 (1x) or higher than 25550 (255x) and thus is misconfigured.", mobEntity->getName());
+                    }
+
+                    // convert the integer based multiplier (based on 100) to float multiplier
+                    float floatMultiplier = intMultiplier / 100.0f;
+
+                    // if some weight penalty (like gravity) then cut the boost multiplier according to retail
+                    // (for mobs with default retail boost of 2.5 then boost becomes 1.20)
+                    if (mobEntity->getMod(Mod::MOVE_SPEED_WEIGHT_PENALTY) > 0)
+                    {
+                        floatMultiplier *= 0.48f;
+                    }
+
+                    // Ensure the float multiplier is at least 1.0 so that multiplier never decreases speed
+                    // For example if the server default or mob mod is 207 then with gravity the
+                    // multiplier would be (207/100) * 0.48 = 0.99 without this check
+                    floatMultiplier = std::max<float>(floatMultiplier, 1.0f);
+
+                    // apply the multiplier and also clamp to make sure of a valid speed (in case multiplier is large)
+                    uint16 newSpeed = std::clamp<uint16>(realSpeed * floatMultiplier, 0, 255);
+                    realSpeed       = static_cast<uint8>(newSpeed);
+                }
+            }
         }
     }
 
--- a/scripts/enum/mob_mod.lua
+++ b/scripts/enum/mob_mod.lua
@@ -90,4 +90,5 @@ xi.mobMod =
     CANNOT_GUARD           = 79, -- Check if the mob does not guard (despite being a MNK or PUP mob)
     SKIP_ALLEGIANCE_CHECK  = 80, -- Skip the allegiance check for valid target (allows for example a mob to cast a TARGET_ENEMY spell on itself)
     ABILITY_RESPONSE       = 81, -- Mob can respond to player ability use with onPlayerAbilityUse()
+    SPEED_BOOST_MULT       = 82, -- Multiplier for the base speed of a mob when the mob's target is out of range (range between 100 and 500, 250 means 2.5x, mechanism is from retail)
 }
--- a/settings/default/map.lua
+++ b/settings/default/map.lua
@@ -112,8 +112,13 @@ xi.settings.map =
     -- Note retail treats the mounted speed as double what it actually is.
     MOUNT_SPEED_MOD = 0,
 
-    -- This is an integer percentage. Modifier to apply to agro'd monster speed (i.e. while engaged in combat). 0 is the retail accurate default. Negative numbers will reduce ALL mobs's speed.
-    MOB_SPEED_MOD = 0,
+    -- Enable or disable a mechanism for boosting the speed of engaged mobs (by a multiplier) when their target is out of range.
+    -- This is a retail mechanism that makes kiting more difficult
+    USE_MOB_SPEED_BOOST_MULTIPLIER = true,
+    -- The default boost multiplier of almost all mobs on retail is 250 (so 2.5x or 2.5 times their normal speed)
+    -- The minimum value is 100 (1x) which means that mobs get no such speed boost (though better to use the boolean toggle above)
+    -- The maximum value is 25500 (255x) which would put the speed of any mob at the system cap
+    DEFAULT_MOB_SPEED_BOOST_MULTIPLIER = 250,
 
     -- Allows you to manipulate the constant multiplier in the skill-up rate formulas, having a potent effect on skill-up rates.
     SKILLUP_CHANCE_MULTIPLIER = 1.0,
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -314,8 +314,11 @@ uint8 CBattleEntity::GetSpeed()
         outputSpeed = outputSpeed + mazurkaQuickeningEffect;
     }
 
-    // Set cap (Default 80).
-    outputSpeed = std::clamp<int16>(outputSpeed, 0, 80 + settings::get<int8>("map.SPEED_MOD"));
+    // Set cap if a PC (Default 80).
+    if (objtype == TYPE_PC)
+    {
+        outputSpeed = std::clamp<int16>(outputSpeed, 0, 80 + settings::get<int8>("map.SPEED_MOD"));
+    }
 
     // Speed cap can be bypassed. Ex. Feast of swords. GM speed.
     // TODO: Find exceptions. Add them here.
--- a/src/map/mob_modifier.h
+++ b/src/map/mob_modifier.h
@@ -110,6 +110,7 @@ enum MOBMODIFIER : int
     MOBMOD_CANNOT_GUARD           = 79, // Check if the mob does not guard(despite being a MNK or PUP mob)
     MOBMOD_SKIP_ALLEGIANCE_CHECK  = 80, // Skip the allegiance check for valid target (allows for example a mob to cast a TARGET_ENEMY spell on itself)
     MOBMOD_ABILITY_RESPONSE       = 81, // Mob can respond to player ability use with onPlayerAbilityUse()
+    MOBMOD_SPEED_BOOST_MULT       = 82, // Multiplier for the base speed of a mob when the mob's target is out of range (range between 100 and 25500, 250 means 2.5x, mechanism is from retail)
 };
 
 #endif


src/map/ai/helpers/pathfind.cpp
```diff
@@ -588,6 +590,37 @@ float CPathFind::GetRealSpeed()
...
+                bool  targetWouldBeDrawnin = battleTarget && (distance(mobEntity->loc.p, battleTarget->loc.p) - battleTarget->m_ModelRadius) > mobEntity->GetMeleeRange();
+
+                if (!mobIsPlayerPet && targetWouldBeDrawnin)
+                {
+                    // base multiplier for most mobs
+                    float multiplier = 2.5f;
+
+                    if (mobEntity->getMobMod(MOBMOD_SPEED_BOOST_MULT) > 0)
+                    {
+                        multiplier = mobEntity->getMobMod(MOBMOD_SPEED_BOOST_MULT);
```
cocosolos: Think mods are an int so may want to divide by 100.

src/map/ai/helpers/pathfind.cpp
```diff
@@ -588,6 +590,42 @@ float CPathFind::GetRealSpeed()
...
+                    float multiplier = settings::get<float>("map.DEFAULT_MOB_SPEED_BOOST_MULTIPLIER");
+
+                    // unless the mob has a custom multiplier
+                    if (mobEntity->getMobMod(MOBMOD_SPEED_BOOST_MULT) > 0)
+                    {
+                        multiplier = mobEntity->getMobMod(MOBMOD_SPEED_BOOST_MULT);
+                    }
+
+                    // clamp the multiplier between 1 and 5 to prevent underflows and overflows
+                    multiplier = std::clamp<float>(multiplier, 1.0f, 5.0f);
```
cocosolos: Probably clamp this after the weight penalty adjustment below, since if it's lower than 2.0 it'll actually decrease speed in chase mode.

src/map/ai/helpers/pathfind.cpp
```diff
@@ -588,6 +590,42 @@ float CPathFind::GetRealSpeed()
...
+
+                    // if some weight penalty (like gravity) then cut the boost multiplier
+                    // (for mobs with default boost of 2.5 then boost becomes 1.2)
+                    if (mobEntity->getMod(Mod::MOVE_SPEED_WEIGHT_PENALTY) > 0)
+                    {
+                        multiplier *= 0.48;
+                    }
+
+                    realSpeed *= multiplier;
```
cocosolos: What if we did the multiplier as a whole number int modifier as a percentage? Default setting as 250, then in the weight penalty block you can just divide by 2 to 125, then at the end here you can do:
```cpp
multiplier /= 10;
realSpeed *= multiplier / 10.f;
```
This would handle both of the issues I mentioned above and the 2.5 to 1.2 adjustment relatively cleanly (personally not a fan of the 0.48 adjustment).

src/map/ai/helpers/pathfind.cpp
TracentEden2: > Think mods are an int so may want to divide by 100.

This should now be fixed.

src/map/ai/helpers/pathfind.cpp
TracentEden2: > Probably clamp this after the weight penalty adjustment below, since if it's lower than 2.0 it'll actually decrease speed in chase mode.

This should now be fixed.

src/map/ai/helpers/pathfind.cpp
TracentEden2: This should now be fixed. Thanks for the comments!

cocosolos: I think you want to do this last part as 2 divisions by 10 instead of 1 by 100. That way the .5 will get dropped. As it is now I think you'll get 125% instead of the intended 120%.

Otherwise everything looks good! 👍

cocosolos:
Is there any reason anyone would want a mob to move slower than its normal speed while in combat? If not, I say delete the old setting and the new boolean setting and just adjust the function like this:
```cpp
                if (isEngaged && !mobIsPlayerPet && targetWouldBeDrawnin)
                {
                    int16 intMultiplier = 100;

                    // apply combat speed boost
                    if (mobEntity->getMobMod(MOBMOD_SPEED_BOOST_MULT) > 0)
                    {
                        intMultiplier += mobEntity->getMobMod(MOBMOD_SPEED_BOOST_MULT);
                    }
                    else
                    {
                        intMultiplier += settings::get<int16>("map.DEFAULT_MOB_SPEED_BOOST_MULTIPLIER");
                    }

                    // if some weight penalty (like gravity) then cut the boost multiplier
                    // (for mobs with default boost of +150% then boost becomes +20%)
                    if (mobEntity->getMod(Mod::MOVE_SPEED_WEIGHT_PENALTY) > 0)
                    {
                        intMultiplier /= 2;
                    }

                    // clamp the multiplier between +0% and +400% to prevent underflows and overflows
                    intMultiplier = std::clamp<int16>(intMultiplier, 100, 500);

                    // convert the integer based multiplier (based on 100) to float multiplier
                    intMultiplier /= 10;
                    realSpeed *= intMultiplier / 10.0f;
                }
```
This way, if someone wants to disable the speed boost they can just set the setting to 0 and they'll move at their normal speed all the time (except the ones with a mobmod set, which is just the gunpods? idk can leave the boolean setting too). This would also make the comments for the setting more clear, as it can just be described as a percentage boost to the mobs speed while in combat (+150% default).

src/map/ai/helpers/pathfind.cpp
TracentEden2: > I think you want to do this last part as 2 divisions by 10 instead of 1 by 100. That way the .5 will get dropped. As it is now I think you'll get 125% instead of the intended 120%.
> 
> Otherwise everything looks good! 👍

My concern is that then you have unclear logic for the reader. I would rather just do the 0.48 multiplication if accuracy is more important, as that is clearer for the reader.

Xaver-DaRed: If I may give my opinion, no matter * 0.48 or 2 steps of flooring /10, a comment is needed stating whats going on and why

TracentEden2:
> This way, if someone wants to disable the speed boost they can just set the setting to 0 and they'll move at their normal speed all the time (except the ones with a mobmod set, which is just the gunpods? idk can leave the boolean setting too). This would also make the comments for the setting more clear, as it can just be described as a percentage boost to the mobs speed while in combat (+150% default).

Yeah my plan was to keep the current system primarily because I was unsure how existing servers are using it, but yeah if LSB people agree then can remove it for sure.

As to the boolean setting, part of my reasoning for keeping the setting is that unclear how many mobs with the custom mod there will eventually be as I think it is not something checked on many mobs.  Also yeah I am not sure if a multiplier or a percentage increase is better for understandability as I think that there are both in LSB. I like the multiplier because I think it is a bit easier for a new person to reason about like a 2x or 3x increase because it is so commonly used in real life, but I am open to other arguments.

src/map/ai/helpers/pathfind.cpp
TracentEden2: > If I may give my opinion, no matter * 0.48 or 2 steps of flooring /10, a comment is needed stating whats going on and why

Yeah for sure

cocosolos:
> > This way, if someone wants to disable the speed boost they can just set the setting to 0 and they'll move at their normal speed all the time (except the ones with a mobmod set, which is just the gunpods? idk can leave the boolean setting too). This would also make the comments for the setting more clear, as it can just be described as a percentage boost to the mobs speed while in combat (+150% default).
> 
> Yeah my plan was to keep the current system primarily because I was unsure how existing servers are using it, but yeah if LSB people agree then can remove it for sure.
> 
> As to the boolean setting, part of my reasoning for keeping the setting is that unclear how many mobs with the custom mod there will eventually be as I think it is not something checked on many mobs. Also yeah I am not sure if a multiplier or a percentage increase is better for understandability as I think that there are both in LSB. I like the multiplier because I think it is a bit easier for a new person to reason about like a 2x or 3x increase because it is so commonly used in real life, but I am open to other arguments.

The global (player) speed mod setting is just a straight int increase to the base movement speed, so there's no consideration for consistency when considering % or multiplier base for this setting. I like % personally because most addons in game will show "+200%" (though this isn't technically accurate as these addons mostly use a base speed of 50 I assume to show player relative speed even though players aren't necessarily moving at 50 speed). If you're gonna stick with a 0.48 modifier then maybe it's best to stick with 2.5 setting, as mixing them in the function I think is more confusing. Idk how well that 0.48 will work out when more speeds are tested, but it's easy enough to change so whatever works. The double division by 10 is just to round the number down to the nearest multiple of 10 instead of 5. Personally I think divide in half then divide by 10 for rounding is easier to understand than x0.48. Another thing is that with either 2.5 or 250 as the setting, values under 1.0 or 100 are meaningless and just treated as 1.0 or 100 regardless. My suggestion above was to set the setting to 150 (as in +150%) then add the base 100% in the function calculations, but that doesn't make sense if you're using a float multiplier (1.5x isn't the same as +150%). A key difference IMO when it comes to clarity is the wording used. Are they moving at 2.5x or 250% speed, or are they being boosted by +150% speed. Gear, for example, is +18% speed but you're moving at 118% speed, if that makes sense.

To be clear, I'm not really requesting any changes. Either way is fine, just clarifying why I would make those decisions.

TracentEden2:
Yep, thanks for your comments :)

> If you're gonna stick with a 0.48 modifier then maybe it's best to stick with 2.5 setting, as mixing them in the function I think is more confusing.

Yeah I think I will stick with 250 instead of 2.5 setting because the mob mod is an integer and I do not want to have the mod an int and the setting a float.

> The double division by 10 is just to round the number down to the nearest multiple of 10 instead of 5. Personally I think divide in half then divide by 10 for rounding is easier to understand than x0.48.

I think I will stick with 0.48 because I do not think we have evidence that SE is doing a divide by half and then two divisions by 10, so more straightforward to just use a simple single number and refer to retail and easy to change later if we need. Similarly, I think I will stick with the multiple for now as I think that is easier to reason about and I put in detailed comments.

src/map/ai/helpers/pathfind.cpp
```diff
@@ -578,15 +580,45 @@ float CPathFind::GetRealSpeed()
...
+                    if (mobEntity->getMod(Mod::MOVE_SPEED_WEIGHT_PENALTY) > 0)
+                    {
+                        floatMultiplier *= 0.48f;
+                    }
+
+                    // clamp the multiplier between 1x and 5x to prevent underflows and overflows
+                    floatMultiplier = std::clamp<float>(floatMultiplier, 1.0f, 5.0f);
+
+                    // apply the multiplier
+                    realSpeed *= floatMultiplier;
```
Xaver-DaRed: Instead of clamping the multiplier, I would use a new int16 variable to clamp the output speed, to really ensure that custom speeds or weird future edge-cases dont overflow.
```c++
int16 newSpeed = std::clamp<int16>(realSpeed * floatMultiplier, 0, 255);

realSpeed = static_cast<uint8>(newSpeed);
```

Xaver-DaRed: I'm sure theres a better way to do this though

TracentEden: > Instead of clamping the multiplier, I would use a new int16 variable to clamp the output speed, to really ensure that custom speeds or weird future edge-cases dont overflow.
> 
> ```c++
> int16 newSpeed = std::clamp<int16>(realSpeed * floatMultiplier, 0, 255);
> 
> realSpeed = static_cast<uint8>(newSpeed);
> ```

I added a `clamp` to the `realSpeed` but I also need to keep at least `floatMultiplier = std::max<float>(floatMultiplier, 1.0f);` to make sure that the multiplier never decreases the speed. I also added a lot of comments to make sure the logic is clearer, thanks for the suggestion.

zach2good:
I know there's been A LOT of chatter here, and on various discords about this - what's going on in this PR (I havent' read it because speed scares me)? Is it ready, is it still up for discussions, is it waiting on more examples/captures/testing, etc.? What's the goal/acceptance criteria, kiting a boss around as a RDM, etc.?

TracentEden2:
> I know there's been A LOT of chatter here, and on various discords about this - what's going on in this PR (I havent' read it because speed scares me)? Is it ready, is it still up for discussions, is it waiting on more examples/captures/testing, etc.? What's the goal/acceptance criteria, kiting a boss around as a RDM, etc.?

At least from my understanding it is ready as we have checked quite a few retail mobs with varying speeds and gravity levels. The acceptance criteria is that it replicates the speeds values seen in actual retail capture packets when kiting (with and without movement speed modifiers like gravity). Thus kiting mobs should result in similar behavior to retail (in terms of the mob catching up and hitting player and such).

The default case is a mob with speed of 40 that boosts to 100 (2.5x) when kiting (if player is out of melee range), and if the mob has normal gravity it should have speed of 30 that boosts to 34 (1.2x). Though this gravity part will be slightly different as LSB gravity is a bit off and I think cocosolos is working on that.

cocosolos:
This is good as-is. Most of the discussion was basically formatting stuff. Mechanically this pretty much matches retail behavior, there's just some weird stuff with the gravity formula in relation to enfeebling effect+ gear and saboteur, but shouldn't have an effect on this stuff, and if it does it's easy enough to address once that's figured out.

zach2good (approved):

ShiyoKozuki:
Is this for every mob?
If I dia a crab on retail(with no +move speed), then autorun, it never hits me once before it deaggros.

cocosolos:
> Is this for every mob? If I dia a crab on retail(with no +move speed), then autorun, it never hits me once before it deaggros.

Retail seems to change the speed for every mob we've observed, but it's possible there are other factors at play. Retail could be doing pathfinding without using the full speed available (if you toggle walk mode your speed value is still 50 or whatever, but you're obviously moving a lot slower). Anecdotally I've been able to catch up with Sandworm Guivre as it was running away at 150 speed by using flee (80 speed). The speed value is frequently changing as well, so it's also possible that the speed goes back to normal as the chasing mob gets close to you just as you go back out of range, making it constantly increase and decrease speed while staying just out of attack range. For example, a crab normally moves 20% slower (40) than a player (50), and when you're out of range it'll be 20% faster (60) than you, but just as it gets into range it'll go back to 20% slower and at the same time you're moving back out of range.

In short, more research is needed but the speed value that is sent in the packets does change. In the current state of the code, have you tested if it's producing your expected results?

TracentEden2:
Yeah from my retail testing the speed drops back down when the mob nears or hits the player and then increases again when the player moves back out of range. Also from my testing a lot of the reason that a mob cannot catchup in retail is due to navmesh and how the mob paths. Thus given both of these the mobs effective speed is not really 2.5x especially compared to the player.

ShiyoKozuki:
> > Is this for every mob? If I dia a crab on retail(with no +move speed), then autorun, it never hits me once before it deaggros.
> 
> Retail seems to change the speed for every mob we've observed, but it's possible there are other factors at play. Retail could be doing pathfinding without using the full speed available (if you toggle walk mode your speed value is still 50 or whatever, but you're obviously moving a lot slower). Anecdotally I've been able to catch up with Sandworm Guivre as it was running away at 150 speed by using flee (80 speed). The speed value is frequently changing as well, so it's also possible that the speed goes back to normal as the chasing mob gets close to you just as you go back out of range, making it constantly increase and decrease speed while staying just out of attack range. For example, a crab normally moves 20% slower (40) than a player (50), and when you're out of range it'll be 20% faster (60) than you, but just as it gets into range it'll go back to 20% slower and at the same time you're moving back out of range.
> 
> In short, more research is needed but the speed value that is sent in the packets does change. In the current state of the code, have you tested if it's producing your expected results?

Can't reproduce what you're saying on current retail
I can nuke  at 20.5 - > run away until mob deaggros - > nuke at 20.5 and never get hit once (While running in a straight line with no objects in my path)
I've been able to do this since 75 era as well

cocosolos:
> > > Is this for every mob? If I dia a crab on retail(with no +move speed), then autorun, it never hits me once before it deaggros.
> > 
> > 
> > Retail seems to change the speed for every mob we've observed, but it's possible there are other factors at play. Retail could be doing pathfinding without using the full speed available (if you toggle walk mode your speed value is still 50 or whatever, but you're obviously moving a lot slower). Anecdotally I've been able to catch up with Sandworm Guivre as it was running away at 150 speed by using flee (80 speed). The speed value is frequently changing as well, so it's also possible that the speed goes back to normal as the chasing mob gets close to you just as you go back out of range, making it constantly increase and decrease speed while staying just out of attack range. For example, a crab normally moves 20% slower (40) than a player (50), and when you're out of range it'll be 20% faster (60) than you, but just as it gets into range it'll go back to 20% slower and at the same time you're moving back out of range.
> > In short, more research is needed but the speed value that is sent in the packets does change. In the current state of the code, have you tested if it's producing your expected results?
> 
> Can't reproduce what you're saying on current retail I can nuke at 20.5 - > run away until mob deaggros - > nuke at 20.5 and never get hit once (While running in a straight line with no objects in my path) I've been able to do this since 75 era as well

I never said you would get hit, I said the speed increases but there may be reasons you still don't get hit. Use something to look at your targets speed, I use Windower so TargetInfo or PacketViewer to view incoming packet 0x00E, and you will see the speed value change from 40 to 100 when you go out of range (I misspoke the increased value in my previous comment).

Are you getting hit on LSB in situations that you wouldn't on retail? If so, then pathfinding or something may need some kind of adjustment.

ShiyoKozuki:
I'm saying this PR is not retail accurate.
The packet you're looking at is not an accurate representation of a monsters movement speed.

TracentEden2:
> I'm saying this PR is not retail accurate. The packet you're looking at is not an accurate representation of a monsters movement speed.

Effective movement speed of a monster relies on a combination of factors include the speed value seen in the packets but also the  navmesh pathing and other factors. We know with certainty that the packet speed values **correlate** to movement speed from testing on many mobs (both slow and fast) and from testing spells like gravity, if you have evidence (videos and packet captures) otherwise feel free to show it (make sure the monster is in camera view so that navmesh pathing of the mob can be seen).

ShiyoKozuki:
Please explain why I have been able to kite things in a straight line without being hit with +0% movement speed since 75 era.
If what you're saying is true, I should not be able to, correct?

When pulling for AOE pulls in abyssea/escha, it's very common for mobs to de-aggro me(with +0% movement speed).
If what you're saying is true, shouldn't this be impossible?

Xaver-DaRed:
This PR is merged. If further adjustments need to be made, a new PR will follow up.
And if thats the case, I'm sure we will have actual evidence on both sides, not just on one. TY Coco and Tracent

PR #3164, opened 2022-11-09
Jug pet and other pet fixes

xkoredev:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

- Fixes jug pet and automaton persistence through zone
- Despawns jug pet after the time specified in pet_list
- Closes https://github.com/HorizonFFXI/HorizonXI-Issues/issues/368

## Steps to test these changes

Set a pet "time" column in pet_list to a low number like 30.
Summon that pet by using the proper jug. Observe it despawn after 30 seconds.
Try zoning while pet is up. Pet should reappear in the other zone with the same HP it had. Pet should still despawn 30 secs after call beast was used.

PR #6318, opened 2024-10-04
[Core] Fix issue where charmed mobs never uncharm due to time

TracentEden2:
**_I affirm:_**

This PR fixes an issue whereby monsters charmed by players never uncharm due to the charm time ending. I think the issue was introduced in this [PR](https://github.com/LandSandBoat/server/pull/3164) where a check was added with `objtype == TYPE_PET` which only applies to jug pets and not charmed mobs. Thus `charmTime` is never checked.

Also the linked PR added a comment that the check in `shouldDespawn` was not longer needed, however this is not the case. I am unsure of the reasoning for that comment, thus this PR removes it.

Charm a mob as a BST and watch the mob eventually uncharm after some time (rather than staying charmed forever).

### DIFF
--- a/src/map/ai/controllers/pet_controller.cpp
+++ b/src/map/ai/controllers/pet_controller.cpp
@@ -40,10 +40,30 @@ void CPetController::Tick(time_point tick)
     TracyZoneScoped;
     TracyZoneString(PPet->getName());
 
-    if (PPet->objtype == TYPE_PET && static_cast<CPetEntity*>(PPet)->shouldDespawn(tick))
+    bool isPlayerPet = PPet->objtype == TYPE_PET || (PPet->objtype == TYPE_MOB && PPet->PMaster && PPet->PMaster->objtype == TYPE_PC);
+
+    // if a player pet then check if a charmed mob or jug pet and if it should despawn
+    if (isPlayerPet)
     {
-        petutils::DespawnPet(PPet->PMaster);
-        return;
+        // if a charmed mob and charm time is up then despawn
+        if (PPet->isCharmed && tick > PPet->charmTime)
+        {
+            petutils::DespawnPet(PPet->PMaster);
+            return;
+        }
+
+        // if a jug pet and the current time > jug spawn time + jug duration then despawn
+        auto* PPetEntity = dynamic_cast<CPetEntity*>(PPet);
+        if (PPetEntity && PPetEntity->isAlive() && PPetEntity->getPetType() == PET_TYPE::JUG_PET)
+        {
+            // need to covert tick to unix time (in seconds) because getJugSpawnTime and getJugDuration give unix time
+            auto tickAsUnixTime = static_cast<uint32>(std::chrono::duration_cast<std::chrono::seconds>(tick.time_since_epoch()).count());
+            if (tickAsUnixTime > PPetEntity->getJugSpawnTime() + PPetEntity->getJugDuration())
+            {
+                petutils::DespawnPet(PPetEntity->PMaster);
+                return;
+            }
+        }
     }
     CMobController::Tick(tick);
 }
--- a/src/map/entities/charentity.h
+++ b/src/map/entities/charentity.h
@@ -186,8 +186,8 @@ struct teleport_t
 struct PetInfo_t
 {
     bool     respawnPet;   // Used for spawning pet on zone
-    int32    jugSpawnTime; // Keeps track of original spawn time in seconds since epoch
-    int32    jugDuration;  // Number of seconds a jug pet should last after its original spawn time
+    uint32   jugSpawnTime; // Keeps track of original spawn time in seconds since epoch
+    uint32   jugDuration;  // Number of seconds a jug pet should last after its original spawn time
     uint8    petID;        // ID as in wyvern(48) , carbuncle(8) ect..
     PET_TYPE petType;      // Type of pet being transferred
     uint8    petLevel;     // Level the pet was spawned with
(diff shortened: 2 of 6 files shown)


zach2good (approved):

src/map/ai/controllers/pet_controller.cpp
```diff
@@ -40,7 +40,9 @@ void CPetController::Tick(time_point tick)
     TracyZoneScoped;
     TracyZoneString(PPet->getName());
 
-    if (PPet->objtype == TYPE_PET && static_cast<CPetEntity*>(PPet)->shouldDespawn(tick))
+    bool isPlayerPet = PPet->objtype == TYPE_PET || (PPet->objtype == TYPE_MOB && PPet->PMaster && PPet->PMaster->objtype == TYPE_PC);
+
+    if (isPlayerPet && static_cast<CPetEntity*>(PPet)->shouldDespawn(tick))
```
zach2good: This doesn't look particularly safe.

For example:
```cpp
if (PPet->objtype == TYPE_PET)
{
    static_cast<CPetEntity*>(PPet)->shouldDespawn(tick));
```
^ This is fine, because you've confirmed that PPet is a pet type, and then you're static casting and using it as a pet type.

By adding in this condition:
```cpp
(PPet->objtype == TYPE_MOB && PPet->PMaster && PPet->PMaster->objtype == TYPE_PC);
```

If the type is a mob, it has a master, and that master is a player - you're confirming that its a Mob AND THEN forcing it to be a Pet.

Pet is derived from Mob, sure, but `shouldDespawn` is a member function of `CPetEntity`. This is UB.

zach2good: Example:
```cpp
#include <unordered_map>
#include <vector>
#include <cmath>
#include <iostream>

enum Type
{
    Mob,
    Pet,
};

class CMobEntity
{
public:
    Type type = Type::Mob;

    virtual void sharedFunction()
    {
        std::cout << "Hello from CMobEntity\n";
    }
};

class CPetEntity final : public CMobEntity
{
public:
    Type type = Type::Pet;

    void sharedFunction() override
    {
        std::cout << "Hello from CPetEntity\n";
    }

    void specificFunction()
    {
        std::cout << "A function only CPetEntity has\n";
    }
};

int main() {
    auto* PMob = new CMobEntity();
    auto* PPet = new CPetEntity();

    if (PMob->type == Type::Mob)
    {
        std::cout << "PMob is a mob\n";
    }
    else
    {
        std::cout << "PMob is NOT a mob\n";
    }

    if (PPet->type == Type::Pet)
    {
        std::cout << "PPet is a pet\n";
    }
    else
    {
        std::cout << "PPet is NOT a pet\n";
    }

    // Try the bad casting
    if (PMob->type == Type::Mob)
    {
        auto* NotAPet = static_cast<CPetEntity*>(PMob); // This cast is UB!
        NotAPet->specificFunction(); // This might work, it might explode, we don't know!
    }

    delete PPet;
    delete PMob;

    return 0;
}
```

zach2good (changes requested):

src/map/ai/controllers/pet_controller.cpp
zach2good: As a general point, if you're dealing with entity pointers, you should always be using `dynamic_cast`, which will return `nullptr` if you try to do an invalid cast. You can also use `objtype` + `static_cast`, but only in the most simple combination.

TracentEden2: Yep, thanks for the comments! In the end I think the most straightforward fix is to remove the `shouldDespawn` function (as it was only called in that one spot in any case) and migrate the logic back to the pet controller. Otherwise you would need to put the function in mobentity and it is confusing to have it there given the very specific scope of the logic.

Also I changed `isSpawned` to `isAlive` because otherwise the logic ends up calling `DespawnPet` like 10 times (with a lot of warnings) when jug is dead (`DespawnPet` kills jugs) but the pet has not yet despawned.

zach2good (approved):

src/map/ai/controllers/pet_controller.cpp
```diff
@@ -40,10 +40,30 @@ void CPetController::Tick(time_point tick)
...
+            petutils::DespawnPet(PPet->PMaster);
+            return;
+        }
+
+        // if a jug pet and the current time > jug spawn time + jug duration then despawn
+        auto* PPetEntity = dynamic_cast<CPetEntity*>(PPet);
+        if (PPetEntity && PPetEntity->isAlive() && PPetEntity->getPetType() == PET_TYPE::JUG_PET)
+        {
+            // need to covert tick to unix time (in seconds) because getJugSpawnTime and getJugDuration give unix time
+            auto tickAsUnixTime = static_cast<int32>(std::chrono::duration_cast<std::chrono::seconds>(tick.time_since_epoch()).count());
```
zach2good: Should this be uint32? A time can't be negative. If getJugSpawnTime and getJugDuration are returning int32 they need to be changed too. Also, if these are ints in seconds, are there any situations where the duration can be fractions of a second?

TracentEden2: > Should this be uint32? A time can't be negative. If getJugSpawnTime and getJugDuration are returning int32 they need to be changed too. Also, if these are ints in seconds, are there any situations where the duration can be fractions of a second?

I updated so that jugSpawnTime and jugDuration are uint32 everywhere they can be. The only place they cannot be is when they are saved as charVars (for persisting jug pets when zoning) as charVars are int32. The durations cannot be fractions of a second because they are read from the database via uint reader and then converted to a duration with a `seconds` construction. I am not sure if all the castings I do are fully needed but I guess better to be explicit rather than implicit. Thanks for the comments!

Xaver-DaRed (approved):

PR #6320, opened 2024-10-04
Fix issue with Indi-type spell targeting

cocosolos:
**_I affirm:_**

Sets Indi-type spells to have TARGET_SELF flag and gets rid of some now unnecessary checks (CBattleEntity::ValidTarget handles self targeting). This could maybe be handled in the same check as pianissimo, but that also has some pet checks that I'm not sure about.

```
!addspell 981
!addspell 768
!changejob war 99
```
Cast Sylvie trust (no log errors), engage a target. She should Fury herself then entrust Frailty to you.
```
!changejob bst 99
!additem 17860
```
Equip the jug, call beast, use entrust, target pet, try casting indi spell on pet.


cocosolos:
Hmm idk how that build fail happened, must've changed the condition after testing. Will fix and retest.

zach2good (approved):

PR #6322, opened 2024-10-05
Fix item enhancement/effect: multiple effects, wear on zone. Add enhancement/effect source

jamesbradleym:
**_I affirm:_**

Migrates a fix from ASB to LSB for Enhancements:

Originally adapted from PR by Tracent https://github.com/AirSkyBoat/AirSkyBoat/pull/2935:

- Allows multiple enchantment effects to be active on a player at the same time (was validated on retail by siknoz)
- Introduces the ability to add an effect by `EffectSourceType` and `EffectSourceTypeParam` which:
- Allows multiple enchantment effects to be active on a player at the same time
- Enchantment status effects now wear off when zoning
- Fixes an exploit where players could keep an enchantment even after unequipping the gear that gives the enchantment
- Adds status effects for items (listed below)
- Corrects status effects for items (listed below)
- Adds items to item.lua

- Use an item that grants an effect on use and check effect
- Use two effect items such as high breath mantle (15487) and vision ring (15559) to see two status effects and then remove gear one by one to see only correct enchantment removed
- Use aero mufflers then remove and see that enaero status effect remains
- Zone with an enchantment to see it wear off

First 60 seconds of this video is pre fix showing only 1 enhancement being applied at a time, not being removed when equipment is removed, and not removed on zone. 60s onwards is post fix from this pr.

https://www.dropbox.com/s/14sxrexj1vbfkdh/Post-Fix%20Enchantment.mp4?dl=0

Waiting on confirmations for a couple items but wanted to get an initial review/test the waters.

While this touches a lot of files, nearly all status effects are unchanged in terms of values - implementation is updated to use the `EffectSourceType` and `EffectSourceTypeParam` for tracking where an effect came from - this also sets up effects of other `EffectSourceType` such as `MOB`, `TEMPORARY_ITEM`, etc.

## Additions

| **Item**               |
|------------------------|
| Melt Baselard           |
| Aiming Gloves           |
| Death Chakram           |
| Hallowed Sword          |
| Hydra Tiara             |
| Hydra Tights            |
| Koen                   |
| Manashell Ring          |
| Memento Muffler         |
| Palmer's Bangles        |
| Prominence Axe          |
| Prominence Sword        |
| Purgatory Collar        |
| Sacred Degen            |
| Sacred Lance            |
| Sacred Mace             |
| Sacred Maul             |
| Sacred Sword            |
| Sacred Wand             |
| Talisman Obi            |
| Vulcan Blade            |
| Vulcan Claymore         |
| Vulcan Degen            |
| Vulcan Sword            |

---

## Changes

| **Item**           | **Change**                                                          |
|--------------------|---------------------------------------------------------------------|
| Pacifist Ring       | Corrected to use `ENMITY_DOWN`         |
| Tough Belt          | Corrected from `600s` to `60s`.                                    |

## Notes

| **Item**           | **Change**                                                          |
|--------------------|---------------------------------------------------------------------|
| Mighty Ring       | Should stack - need to track this by some instance based ID         |


zach2good:
Looking at what the PR itself does, it's doing a few fixes, adding a new binding, and enabling this behaviour:

```lua
    if target:hasEquipped(xi.item.ACROBATS_BELT) then
        target:addStatusEffect(xi.effect.ENCHANTMENT, 0, 0, 60, 0, 0, 0, xi.item.ACROBATS_BELT)
    end
end
```

I wonder if this is really the right way to handle this. This feels like a thing we should be marking items with at the data input level (SQL) as: "I'm usable, but only if you have me equipped" and then all of the followup is in Core, rather than checking in Lua. But I don't really care that much, I'll defer to the others about this.

I think that item source stuff is a good idea, but it probably wants to be generalised as a sourceType (equipped item, temporary item, mob, etc.) and then the sourceTypeParam (the itemid, etc.). If you add a binding for addStatusEffectFromSourceItem, you could add the equipped check in there?

If things are going to stay in Lua, this can probably follow the pattern we have of using item utils that wrap this repeated behaviour.

I'd probably split out any little fixes etc. that are coming along with this into their own commits, but I appreciate that this is implementing a whole feature, so the commit itself is going to be large.

zach2good:
> Migrate enhancement fixes to LSB 

Is also doesn't fairly describe whats going on in this PR and commit, we don't care that it was migrated from somewhere - can you please name these in a way that's describing the actual work?

```
This PR fixes several issues:

Fixes an exploit where players could keep an enchantment even after unequipping the gear that gives the enchantment
Also allows multiple enchantment effects to be active on a player at the same time (was validated on retail by siknoz)
Fixes breath mantle to actually give enchantment
Fixes several pieces of gear that give the status effects that should not be wear off upon unequipping gear (based on JP wiki)
Enchantment status effects now wear off when zoning
```

^ Each of these could be their own commits. I know you're porting this from downstream, but that doesn't mean we want to have 4-5 distinct chunks of work smushed into one.

jamesbradleym:
> Looking at what the PR itself does, it's doing a few fixes, adding a new binding, and enabling this behaviour:
> 
> ```lua
> itemObject.onItemUse = function(target)
>     if target:hasEquipped(xi.item.ACROBATS_BELT) then
>         target:addStatusEffect(xi.effect.ENCHANTMENT, 0, 0, 60, 0, 0, 0, xi.item.ACROBATS_BELT)
>     end
> end
> ```
> 
> I wonder if this is really the right way to handle this. This feels like a thing we should be marking items with at the data input level (SQL) as: "I'm usable, but only if you have me equipped" and then all of the followup is in Core, rather than checking in Lua. But I don't really care that much, I'll defer to the others about this.
> 
> I think that item source stuff is a good idea, but it probably wants to be generalised as a sourceType (equipped item, temporary item, mob, etc.) and then the sourceTypeParam (the itemid, etc.). If you add a binding for addStatusEffectFromSourceItem, you could add the equipped check in there?
> 
> If things are going to stay in Lua, this can probably follow the pattern we have of using item utils that wrap this repeated behaviour.
> 
> I'd probably split out any little fixes etc. that are coming along with this into their own commits, but I appreciate that this is implementing a whole feature, so the commit itself is going to be large.

Appreciate the direction here - went ahead and implemented this as `SourceType` and `SourceTypeParam`, happy to switch out the lua implementation for sql pending further feedback 👍🏻

jamesbradleym:
> > Migrate enhancement fixes to LSB
> 
> Is also doesn't fairly describe whats going on in this PR and commit, we don't care that it was migrated from somewhere - can you please name these in a way that's describing the actual work?
> 
> ```
> This PR fixes several issues:
> 
> Fixes an exploit where players could keep an enchantment even after unequipping the gear that gives the enchantment
> Also allows multiple enchantment effects to be active on a player at the same time (was validated on retail by siknoz)
> Fixes breath mantle to actually give enchantment
> Fixes several pieces of gear that give the status effects that should not be wear off upon unequipping gear (based on JP wiki)
> Enchantment status effects now wear off when zoning
> ```
> 
> ^ Each of these could be their own commits. I know you're porting this from downstream, but that doesn't mean we want to have 4-5 distinct chunks of work smushed into one.

The pr notes were (admittedly, lazily) taken directly from Tracent's original PR. The notes about fixes aren't necessarily fixes in LSB, for instance some of the fixed item luas do not currently exist in LSB, thus they are additions rather than fixes.. Once everything looks good I can go ahead and break apart the commits for new items, fixes for existing items, and adding items to the item enum if preferable.

jamesbradleym:
@zach2good I'm misunderstanding something about how lua methods are being called..

In CBaseEntity.lua i have:

```
---@nodiscard
---@param StatusID integer
---@param SourceType integer
---@param SourceTypeParam integer
---@return CStatusEffect?
function CBaseEntity:getStatusEffectBySource(StatusID, SourceType, SourceTypeParam)
end
```

and `target` in `onItemCheck` within the item luas is a `CBaseEntity` by virtue of `TItem`, 
`target:getStatusEffectBySource(xi.effect.ENFIRE, xi.effectSourceType.EQUIPPED_ITEM, xi.item.VULCAN_SWORD)`

and yet, the sanity check says 
`Could not find function match for getStatusEffectBySource (./scripts/items/vulcan_sword.lua:11:15)`...

claywar:
This is not a LLS failure, but you'll need to add the function to the binding checks in tools

jamesbradleym:
> This is not a LLS failure, but you'll need to add the function to the binding checks in tools

Thanks @claywar - just to confirm does that mean adding it to the `check_lua_binding_usage.py` `function_names` list?

Or perhaps detangling this from the existing method and SOL registering `getStatusEffectBySource` and setting it up the rest in `lua_baseentity`?

src/map/lua/lua_baseentity.cpp
```diff
@@ -3742,6 +3743,28 @@ std::optional<CLuaItem> CLuaBaseEntity::getEquippedItem(uint8 slot)
...
+ *  Notes   :
+ ************************************************************************/
+
+bool CLuaBaseEntity::hasEquipped(uint16 equipmentID)
+{
+    if (m_PBaseEntity->objtype == TYPE_PC)
+    {
+        for (uint8 equipmentSlot = 0; equipmentSlot <= 15; equipmentSlot++)
+        {
+            if (getEquipID((SLOTTYPE)equipmentSlot) == equipmentID)
```
WinterSolstice8:
```suggestion
            if (getEquipID(static_cast<SLOTTYPE>(equipmentSlot)) == equipmentID)
```
?

WinterSolstice8 (approved):
Not sure about the LLS stuff, otherwise seems ok (did mostly a cursory glance on the lua stuff)

jamesbradleym:
> Not sure about the LLS stuff, otherwise seems ok (did mostly a cursory glance on the lua stuff)

Had thought a rebase might resolve but it looks like other unrelated PRs are also hitting this dealer_moogle.lua error

jamesbradleym:
@zach2good @claywar How are we feeling about this?

PR #6323, opened 2024-10-06
Better level sync effect handling

jamesbradleym:
**_I affirm:_**

Prevent zoning in with pre-existing level sync but no party: [Source](https://github.com/AirSkyBoat/AirSkyBoat/commit/2461a8a5119aa9d1155e66f3b30cdedbce2ba45c)

Remove buffs in level capped zone: [Source](https://github.com/AirSkyBoat/AirSkyBoat/commit/497093728a28856d7537ae76b47b9e6a1369c545)

Moves Level sync magic numbers to message_standard.h

Zone with party level sync where party disbands prior to zoning in.

In level capped zone, trigger new level sync and see status effects of flag wear: EFFECTFLAG_DISPELABLE, EFFECTFLAG_ERASABLE, EFFECTFLAG_ATTACK, EFFECTFLAG_ON_ZONE, EFFECTFLAG_SONG, EFFECTFLAG_ROLL, EFFECTFLAG_SYNTH_SUPPORT, EFFECTFLAG_BLOODPACT

### DIFF
--- a/src/map/packets/message_basic.cpp
+++ b/src/map/packets/message_basic.cpp
@@ -22,6 +22,7 @@
 #include "common/socket.h"
 
 #include "message_basic.h"
+#include "message_standard.h"
 
 #include "entities/baseentity.h"
 
@@ -39,6 +40,11 @@ CMessageBasicPacket::CMessageBasicPacket(CBaseEntity* PSender, CBaseEntity* PTar
     ref<uint32>(0x0C) = param;
     ref<uint32>(0x10) = value;
     ref<uint16>(0x18) = messageID;
+}
+
+CMessageBasicPacket::CMessageBasicPacket(CBaseEntity* PSender, CBaseEntity* PTarget, int32 param, int32 value, MsgStd messageID)
+: CMessageBasicPacket(PSender, PTarget, param, value, static_cast<uint16>(messageID))
+{
 }
 
 uint16 CMessageBasicPacket::getMessageID()
--- a/src/map/packets/message_basic.h
+++ b/src/map/packets/message_basic.h
@@ -24,6 +24,8 @@
 
 #include "common/cbasetypes.h"
 
+#include "message_standard.h"
+
 #include "basic.h"
 
 /*
@@ -221,7 +223,9 @@ class CBaseEntity;
 class CMessageBasicPacket : public CBasicPacket
 {
 public:
+    // TODO: Replace uint16 with MsgStd version
     CMessageBasicPacket(CBaseEntity* PSender, CBaseEntity* PTarget, int32 param, int32 value, uint16 messageID);
+    CMessageBasicPacket(CBaseEntity* PSender, CBaseEntity* PTarget, int32 param, int32 value, MsgStd messageID);
     uint16 getMessageID();
 };
 
--- a/src/map/packets/message_standard.h
+++ b/src/map/packets/message_standard.h
@@ -110,6 +110,7 @@ enum class MsgStd
     LevelSyncWarning             = 235, // Warning! This is a Level Sync party ...
     CannotInviteLevelSync        = 236, // You cannot invite that person at this time. This player is either undergoing Level Sync...
     CannotJoinLevelSync          = 237, // You cannot join this party.  You are either undergoing Level Sync...
+    LevelSyncSet                 = 238, // The party's level has been restricted to <level>.
     Compass                      = 239, // The compass reads: ...
     CannotHere                   = 256, // You cannot use that command in this area.
     HeadgearShow                 = 260,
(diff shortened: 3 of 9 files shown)


src/map/zone.cpp
```diff
@@ -1126,6 +1144,34 @@ void CZone::CharZoneOut(CCharEntity* PChar)
...
+                for (uint32 i = 0; i < PChar->PParty->members.size(); ++i)
+                {
+                    if (PChar->PParty->members.at(i) != PChar && PChar->PParty->members.at(i)->getZone() == PChar->PParty->GetSyncTarget()->getZone())
+                    {
+                        count++;
+                    }
+                }
+                if (count < 2) // 3, because one is zoning out - thus at least 2 will be left
+                {
+                    PChar->PParty->SetSyncTarget("", 552);
```
zach2good: Inside of `SetSyncTarget`, this magic 552 is being cast to `static_cast<MsgStd>(message)));`. While you're around this bit of code, can you add new entries to `MsgStd` and then use them in all the places `SetSyncTarget` is used, and then remove that cast?

It'll end up looking like:
```cpp
// Callsite
PChar->PParty->SetSyncTarget("", MsgStd::TheNewEnumWhatever);

// Function sig
void CParty::SetSyncTarget(const std::string& MemberName, MsgStd message)

// In the function (old)
member->pushPacket(new CMessageStandardPacket(PChar->GetMLevel(), 0, 0, 0, static_cast<MsgStd>(message)));

// new
member->pushPacket(new CMessageStandardPacket(PChar->GetMLevel(), 0, 0, 0, message));
```

zach2good (approved):
I don't know much about level sync, but this is probably fine? 🤷

src/map/zone.cpp
jamesbradleym: Thanks for the guidance here! 

Went ahead and added the existing magic numbers to MsgStd, also found a few more that I figured would be good to add that aren't currently implemented (but figured it would be better to have them at the ready while I was there, let me know if those unused should be removed)

The enum names are semi-generic, happy to go through them and give them more specific naming if desired.

WinterSolstice8 (approved):

zach2good:
You've not marked the "I've tested this" checkbox, is this tested and working?

jamesbradleym:
Updated to address #6346 by reverting the level sync check changes from #6282 back to `zone.cpp` -> `CharZoneOut`

Tested and confirmed no crashing from level sync zoning

jamesbradleym:
@zach2good Tested these changes (level sync designee/leader leaving party, "" zoning,  "" deleveling, "" leveling, "" unsyncing, "" logging out, trying to level sync when pt member has effect that blocks it, trying to level sync when designee is in another zone), everything is looking good as far as I can see including the fix for crashing when zoning while level synced.

src/map/packets/message_standard.h
```diff
@@ -110,6 +110,7 @@ enum class MsgStd
     LevelSyncWarning             = 235, // Warning! This is a Level Sync party ...
     CannotInviteLevelSync        = 236, // You cannot invite that person at this time. This player is either undergoing Level Sync...
     CannotJoinLevelSync          = 237, // You cannot join this party.  You are either undergoing Level Sync...
+    LevelSyncSet                 = 238, // The party's level has been restricted to <level>.
```
zach2good, line 113: This one is named correctly

src/map/packets/message_standard.h
```diff
@@ -133,6 +134,21 @@ enum class MsgStd
     TrustEnmity                  = 300, // You cannot use Trust magic while having gained enmity.
     TrustSoloOrLeader            = 301, // You cannot use Trust magic unless you are solo or the party leader.
     AnErrorHasOccured            = 308, // An error has occurred.
+    LevelSyncMsg540              = 540, // Level Sync activated. Your level has been restricted to <Level>. Equipment effected by the level restriction will be adjusted accordingly. Experience...
```
zach2good: Embedding the number of the message into the human readable enum key isn't helping anything. 

If the message is: `Level Sync activated. Your level has been restricted to <Level>. Equipment effected by the level restriction will be adjusted accordingly. Experience...`

Then the key should be something like `LevelSyncActivated`, or `LevelSyncStart`.

In the eventually use case, what does this tell the reader?
```cpp
SetSyncTarget("", MsgStd::LevelSyncMsg553);
```

Nothing more than it was before. The point is the make these human readable.

src/map/packets/message_basic.h
```diff
@@ -222,6 +224,7 @@ class CMessageBasicPacket : public CBasicPacket
 {
 public:
     CMessageBasicPacket(CBaseEntity* PSender, CBaseEntity* PTarget, int32 param, int32 value, uint16 messageID);
+    CMessageBasicPacket(CBaseEntity* PSender, CBaseEntity* PTarget, int32 param, int32 value, MsgStd messageID);
```
zach2good, line 228: Can you leave a TODO next to the original function signature saying that this needs to be replaced with the MsgStd version. It's out of scope for you to convert everything right now in this PR

src/map/packets/message_basic.cpp
```diff
@@ -41,6 +42,22 @@ CMessageBasicPacket::CMessageBasicPacket(CBaseEntity* PSender, CBaseEntity* PTar
     ref<uint16>(0x18) = messageID;
 }
 
+CMessageBasicPacket::CMessageBasicPacket(CBaseEntity* PSender, CBaseEntity* PTarget, int32 param, int32 value, MsgStd messageID)
```
zach2good, line 45: There's no need to copy the full implementation of this constructor again.

```cpp
#include <iostream>

enum SomeEnum : int
{
    SomeThing  = 1,
    OtherThing = 2,
};

class SomePacket
{
public:
    SomePacket(int key)
    {
        // Do something with int key
        std::cout << "Constructing SomePacket with: " << key << "\n";
    }

    SomePacket(SomeEnum asEnum)
    : SomePacket(static_cast<int>(asEnum)) // 
    {
    }

};

int main()
{
    auto packet = SomePacket(SomeEnum::SomeThing);
    return 0;
}
```

See https://en.cppreference.com/w/cpp/language/constructor#Delegating_constructor

zach2good (changes requested):

zach2good:
If you're including rolling back another set of changes as part of your PR, they should be their own "Reverts X" commit. You're essentially combining two different pieces of work into one and then making it hard to tell them apart.

When @WinterSolstice8 has the time/will, they'll need to look at this. I think they're currently thinking about a fix for the level sync zoning thing

WinterSolstice8:
> Updated to address #6346 by reverting the level sync check changes from #6282 back to `zone.cpp` -> `CharZoneOut`
> 
> Tested and confirmed no crashing from level sync zoning

Just add those back to CharZoneOut, I plan to eventually delete PChar earlier and there's no reason to remove it from the destructor of CCharEntity when it's stateful malloced memory that shouldn't be assumed to always be handled by some other function

I just dont have time to work on this garbage because I work a 10 hr 4 day weekend shift now and I'm too tired every day that's not my days off

and sometimes I want my days off to be fun and not 72 hours of fighting bad assumptions from DSP to try to improve zoning and kill black screens

jamesbradleym:
> > Updated to address #6346 by reverting the level sync check changes from #6282 back to `zone.cpp` -> `CharZoneOut`
> > Tested and confirmed no crashing from level sync zoning
> 
> Just add those back to CharZoneOut, I plan to eventually delete PChar earlier and there's no reason to remove it from the destructor of CCharEntity when it's stateful malloced memory that shouldn't be assumed to always be handled by some other function
> 
> I just dont have time to work on this garbage because I work a 10 hr 4 day weekend shift now and I'm too tired every day that's not my days off
> 
> and sometimes I want my days off to be fun and not 72 hours of fighting bad assumptions from DSP to try to improve zoning and kill black screens

Totally, you should be enjoying your days off 🙏🏼

src/map/packets/message_basic.cpp
jamesbradleym: 🙏🏼 Just what I needed, thanks!

src/map/packets/message_basic.h
jamesbradleym: 🫡 Sure thing

src/map/packets/message_standard.h
jamesbradleym: No problem, updated!

zach2good (approved):

PR #6324, opened 2024-10-06
[Lua] Fix several issue with Shen and Shen Filtrates

TracentEden2:
**_I affirm:_**

This PR fixes several issues with Shen and Shen Filtrates:
1. Change the Shen in-shell skill list from Cop_Wardens (which are thunder based skills) to the correct Uragnite skill list
2. Change the in-shell and out-shell time-based switching logic to remove a race condition and be closer to retail capture observations
3. Add the correct absorb mod to Shen so it absorbs water spells and so Filtrates periodically cure Shen with water spells
4. Fix the insta-cast Flood logic so the player no longer sees a chainspell wears off message on Shen
5. Move the Shen and Filtrate status immunities to mob scripts (as standard) and validate those immunities against the captures
6. Fix the message displayed when spawning Shen and when clicking on the Shen ???

The changes are based the captures by Siknoz [here](https://youtu.be/2Lfh4R8DOfM) and zach2good [here](https://drive.google.com/open?id=1r5_5VGBoDc0jFKPLRZfafyodPhFNIPee).

Fight Shen and check compare logic against the retail captures.


zach2good (approved):

scripts/zones/Bibiki_Bay/mobs/Shen.lua
```diff
@@ -29,44 +30,71 @@ local function exitShell(mob)
...
+        mobAction == xi.act.MOBABILITY_INTERRUPT or
+        mobAction == xi.act.MOBABILITY_FINISH or
+        mobAction == xi.act.MAGIC_START or
+        mobAction == xi.act.MAGIC_CASTING or
+        mobAction == xi.act.MAGIC_INTERRUPT or
+        mobAction == xi.act.MAGIC_FINISH or
+        not mob:actionQueueEmpty()
+    then
+        mobIsBusy = true
     end
```
Xaver-DaRed, line 99: You know, I'm pretty sure we have something similar somehwere. And if not, we should definately have this globalized.
Was it with those flying snakes, like tiamat and Jorm? idr

I mean, I'm sure we WANT to easily check if a mob is busy for many mobs, without having to define this logic every time, so, if you could add a TODO comment at the very least, that would help

TracentEden2: > You know, I'm pretty sure we have something similar somehwere. And if not, we should definately have this globalized. Was it with those flying snakes, like tiamat and Jorm? idr
> 
> I mean, I'm sure we WANT to easily check if a mob is busy for many mobs, without having to define this logic every time, so, if you could add a TODO comment at the very least, that would help

Thanks, I have updated the PR with a TODO comment. I will look into a global function at some point and put in a separate PR.

Xaver-DaRed (approved):

PR #6328, opened 2024-10-07
[Lua] Fix several Dalham (CoP 7-4 mob) issues

TracentEden2:
**_I affirm:_**

This PR fixes several issues with Dalham:
1. Add 100% enwater additional effect
2. Add the correct immunities, resistances, and water absorb
3. Add the correct increase in attacks per round and the 2hr animation at different HP levels
4. Change to the correct war job, correct HP, and correct delay

The changes are based on the captures by Siknoz [here](https://youtu.be/u9APAn_F5TU), zach2good [here](https://drive.google.com/open?id=157JfVoVzba4f_4DV1-q9GYV0pewuFkpH), and Jimmayus [here](https://youtu.be/0WmmSZvB77g).

Fight Dalham and compare to retail capture videos and information


zach2good (approved):

scripts/zones/Bibiki_Bay/mobs/Dalham.lua
```diff
@@ -7,13 +7,44 @@ local entity = {}
...
-    -- "Increases swings per attack round as its HP lowers. At 15-20% swings like Charybdis giving the illusion of Hundred Fists."
-    local swings = 1 + math.floor((100 - mob:getHPP()) / 18)
-    mob:setMobMod(xi.mobMod.MULTI_HIT, swings)
+    if mob:getHPP() < 75 and mob:getLocalVar('multihitControl') == 0 then
+        mob:setLocalVar('multihitControl', 1)
+        -- set DOUBLE_ATTACK to 100
+        -- cannot use MULTI_HIT because that is just upto X hits
+        mob:setMod(xi.mod.DOUBLE_ATTACK, 100)
+        -- dummy mob ability for 2hr animation (does not use tp)
+        mob:useMobAbility(626)
```
WinterSolstice8: Based on the cap there's no way this is actually a real ability and you should just use injectActionPacket

`mob:injectActionPacket(mob:getID(), 11, 437, 0, 0x18, 0, 0, 626)`
most 2hr dustclouds like this should probably just be injections

(note the back to back WS/finish packets)
image

WinterSolstice8 (changes requested):

scripts/zones/Bibiki_Bay/mobs/Dalham.lua
TracentEden2: > Based on the cap there's no way this is actually a real ability and you should just use injectActionPacket
> 
> `mob:injectActionPacket(mob:getID(), 11, 437, 0, 0x18, 0, 0, 626)` most 2hr dustclouds like this should probably just be injections

Thanks for the info, I have updated the PR to use `injectActionPacket` instead!

WinterSolstice8 (approved):

scripts/zones/Bibiki_Bay/mobs/Dalham.lua
```diff
@@ -7,13 +7,44 @@ local entity = {}
 
 entity.onMobInitialize = function(mob)
     mob:setMobMod(xi.mobMod.IDLE_DESPAWN, 180)
+end
+
+entity.onMobSpawn = function(mob)
+    mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
+    mob:setMod(xi.mod.SLEEPRES, 90)
```
Xaver-DaRed: This modifier is exclusively used for job traits and similar, which make it so it returns an special resist message, allowing us to know it was the job trait that triggered

There should be a MEVA mod for sleep

scripts/zones/Bibiki_Bay/mobs/Dalham.lua
```diff
@@ -7,13 +7,44 @@ local entity = {}
...
+        -- inject packet for 2hr animation
+        mob:injectActionPacket(mob:getID(), 11, 437, 0, 0x18, 0, 0, 626)
+    elseif mob:getHPP() < 25 and mob:getLocalVar('multihitControl') == 2 then
+        mob:setLocalVar('multihitControl', 3)
+        -- set QUAD_ATTACK to 100
+        mob:setMod(xi.mod.TRIPLE_ATTACK, 0)
+        mob:setMod(xi.mod.QUAD_ATTACK, 100)
+        -- inject packet for 2hr animation
+        mob:injectActionPacket(mob:getID(), 11, 437, 0, 0x18, 0, 0, 626)
+    end
```
Xaver-DaRed, line 50: We could store both the mob_local_variable and the mob_HPP into local variables before the if/elseif chain, so we dont call them potentially 3 times per tick.

scripts/zones/Bibiki_Bay/mobs/Dalham.lua
TracentEden2: > This modifier is exclusively used for job traits and similar, which make it so it returns an special resist message, allowing us to know it was the job trait that triggered
> 
> There should be a MEVA mod for sleep

Thanks, I have updated the PR!

scripts/zones/Bibiki_Bay/mobs/Dalham.lua
TracentEden2: > We could store both the mob_local_variable and the mob_HPP into local variables before the if/elseif chain, so we dont call them potentially 3 times per tick.

Thanks, I have updated the PR!

Xaver-DaRed (approved):

PR #6340, opened 2024-10-09
[Proposal] Create module for era composure effect

Xaver-DaRed:
**_I affirm:_**

As title sais

Enable module. Use composure. Have it give 15 acc at lvl 75 and lower at lower levels

### DIFF
--- a/modules/era/lua/era_effect_composure.lua
+++ b/modules/era/lua/era_effect_composure.lua
@@ -1,0 +1,16 @@
+-----------------------------------
+-- Era Composure effect
+-- Pre 8 Feb 2019
+-- https://forum-square--enix-com.translate.goog/ffxi/threads/55024?_x_tr_sl=ja&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=sc
+-----------------------------------
+require('modules/module_utils')
+-----------------------------------
+local m = Module:new('era_effect_composure')
+
+m:addOverride('xi.effects.composure.onEffectGain', function(target, effect)
+    local power = math.floor(target:getMainLvl() / 5)
+
+    effect:addMod(xi.mod.ACC, power)
+end)
+
+return m


claywar (approved):

modules/era/lua/era_status_effects.lua
```diff
@@ -0,0 +1,20 @@
...
+-----------------------------------
+-- Module definition
+-----------------------------------
+-- This module attempts to have a centralized file to override status effects with era changes.
+local eraStatusEffects = Module:new('era_status_effects')
+
+-----------------------------------
+-- Module enable/disable
+-----------------------------------
+eraStatusEffects:setEnabled(false)
```
zach2good: I'm probably going to get rid of `setEnabled`, if you load the module file - it'll be enabled. So this little section can be removed

zach2good (approved):
Exactly the sort of thing we want for era modules, thanks!

modules/era/lua/era_status_effects.lua
zach2good: Maybe a comment in the header saying when it changed - if you can find that info!

Xaver-DaRed: - Removed section
- Added date change
- Included link

modules/era/lua/era_status_effects.lua
```diff
@@ -0,0 +1,18 @@
+-----------------------------------
+-- Common Requires
+-----------------------------------
+require('modules/module_utils')
+-----------------------------------
+-- Module definition
+-----------------------------------
+-- This module attempts to have a centralized file to override status effects with era changes.
+local eraStatusEffects = Module:new('era_status_effects')
```
zach2good: So, the problem here is that there's no way to selectively choose overrides within a module. If all the era status changes go in here, it'll mix up all the different eras. I guess for now it would make sense to change this to `era_composure.lua` and `era_composure` module name, and leave it in this folder. At some point later when we have more era modules someone else can start to organise them into folders based on period etc.

modules/era/lua/era_status_effects.lua
```diff
@@ -0,0 +1,18 @@
+-----------------------------------
+-- Common Requires
```
zach2good: No need for all the extra headers, this module is a good example:
```lua
-- Pre-RMT Countermeasure Drops
require('modules/module_utils')
require('scripts/globals/treasure')
local m = Module:new('pre_rmt_drops')

m:addOverride('xi.zones.Castle_Oztroja.Zone.onInitialize', function(zone)
    xi.treasure.treasureInfo[xi.treasure.type.COFFER].zone[xi.zone.CASTLE_OZTROJA].item = { 0.150, xi.item.ASTRAL_RING }
end)

return m
```

and then information about the date of the changes could go in the title block at the top

modules/era/lua/era_status_effects.lua
Xaver-DaRed: I went with `era_effect_composure.lua`

We may end up being the ones organizing the folder structure, so in case theres more effect modules down the line, I figure having them all together will make things slighty easier

zach2good (approved):

PR #6350, opened 2024-10-13
[Lua] Fix Feeler Antlion ??? despawning and respawning

TracentEden2:
**_I affirm:_**

The PR fixes an issue where the ??? for Feeler Antlion never despawns when popping Feeler Antlion.

The PR adds code to despawn the ??? and also adds a custom function  `canStartFeelerQMTimer` to control the respawning of the ???. A custom function is needed as `npcUtil.popFromQM` does not support ??? timers when mobs spawn other mobs (such as Feeler Antlion spawning Alastor Antlion and Executioner Antlion). The custom function is called in `onMobDespawn` of Feeler, Alastor, and Executioner and checks if any these mobs are still spawned and if not then start the ??? respawn timer.

Spawn Feeler Antlion from ??? and test different scenarios of killing Feeler, Alastor, Executioner and seeing when the ??? respawns.

### DIFF
--- a/scripts/zones/Attohwa_Chasm/globals.lua
+++ b/scripts/zones/Attohwa_Chasm/globals.lua
@@ -1,0 +1,37 @@
+-----------------------------------
+-- Zone: Attohwa_Chasm
+-- Desc: this file contains functions that are shared by multiple luas in this zone's directory
+-----------------------------------
+local ID = zones[xi.zone.ATTOHWA_CHASM]
+-----------------------------------
+
+local attohwaChasmGlobal =
+{
+    -- Check to see if any mobs associated with Feeler Antlion are still spawned
+    -- if not then can start the timer to respawn the ???; a custom function is needed
+    -- because npcUtil.popFromQM does not support ??? timer when mobs pop other mobs
+    canStartFeelerQMTimer = function(despawningMobID)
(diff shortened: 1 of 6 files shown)


scripts/zones/Attohwa_Chasm/globals.lua
```diff
@@ -0,0 +1,39 @@
+-----------------------------------
+-- Zone: Attohwa_Chasm
+-- Desc: this file contains functions that are shared by multiple luas in this zone's directory
+-----------------------------------
+local ID = zones[xi.zone.ATTOHWA_CHASM]
+-----------------------------------
+
+local attohwaChasmGlobal =
+{
+    --[[..............................................................................................
```
zach2good: Lets not use this style of comments, these are horrifying 🤢

zach2good (approved):
I don't know these mobs, but seems fine to me 👍

scripts/zones/Attohwa_Chasm/globals.lua
TracentEden2: > Lets not use this style of comments, these are horrifying 🤢

Yep, I was unsure if that was used in many globals for some reason :) , have changed the style, thanks!

zach2good (approved):

PR #6354, opened 2024-10-14
Update Ilrusi Atoll to Assault Framework

dallano:
**_I affirm:_**

- Updates Arrapago Reef's Runic Seal to use the existing assault framework global
- Adds messaging offset for assaults in Arrapago Reef

Enter into any pre-existing Ilrusi Atoll assault using the Runic Seal in Arrapago Reef

### DIFF
--- a/scripts/zones/Arrapago_Reef/npcs/_jic.lua
+++ b/scripts/zones/Arrapago_Reef/npcs/_jic.lua
@@ -12,94 +12,18 @@ entity.onTrade = function(player, npc, trade)
 end
 
 entity.onTrigger = function(player, npc)
-    -- TODO: Fix, implement & balance Assault
-    --[[
-    if player:hasKeyItem(xi.ki.ILRUSI_ASSAULT_ORDERS) then
-        local assaultid = player:getCurrentAssault()
-        local recommendedLevel = getRecommendedAssaultLevel(assaultid)
-        local armband = 0
-        if player:hasKeyItem(xi.ki.ASSAULT_ARMBAND) then
-            armband = 1
-        end
-        player:startEvent(219, assaultid, -4, 0, recommendedLevel, 4, armband)
-    else
+    if not xi.instance.onTrigger(player, npc, xi.zone.ILRUSI_ATOLL) then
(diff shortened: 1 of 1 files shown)


zach2good (approved):
I didn't even know we had an assault framework 🤷

scripts/zones/Arrapago_Reef/IDs.lua
```diff
@@ -23,6 +23,7 @@ zones[xi.zone.ARRAPAGO_REEF] =
         STAGING_GATE_ILRUSI           = 7327, -- Ilrusi Atoll Staging Point.
         CANNOT_LEAVE                  = 7332, -- You cannot leave this area while in the possession of <keyitem>.
         RESPONSE                      = 7341, -- There is no response...
+        ASSAULT_START_OFFSET          = 7454, -- Max MP Down removed for <name>.
```
zach2good: This is a strange message to have as the offset for assaults. Is this because the actual assault messages have all of those huge ugly lists of areas/points etc. that are hard/ugly to capture in the comments for UpdateExtractor?

dallano:
> I didn't even know we had an assault framework 🤷

There's at least a global that handles all shared calls required for runic portal interactions - as well as shared and common functions of assault mobs. Things like progressing instance, adding temp items, adding bonus points to assault, etc.

scripts/zones/Arrapago_Reef/IDs.lua
dallano: I'm happy you pointed this out. 
A. It belongs in Ilrusi Atoll IDs.lua
B. All messaging for assaults in the assault instance zones begin from this line, correct. Checking quickly and there are around 250 messages for Ilrusi Atoll alone

PR'ing the adjustment to Ilrusi Atoll now

dallano: adjustments made

zach2good: I didn't even catch the wrong zone :D

zach2good: (Makes sense to have this message as the starting offset message)

PR #6359, opened 2024-10-15
Core: Add generate_ipc_stubs workflow

zach2good:
**_I affirm:_**

Currently inert, shouldn't do anything noticeable.

Adds the Python3 dependency to the CMake build that I have in another branch, and the automatic calling of a script. That script generates all the stubs for IPC. If you then don't fill in the stubs you get a very clear error message telling you exactly what you need to do:
```
/Users/zrtoo/dev/alpaca_py_test/build/generated/ipc_stubs.h:41:19: error: static assertion failed due to requirement 'is_struct_complete<ipc::SomeOtherData, void>::value': You must fully define this struct in common/ipc_structs.h.
   41 |     static_assert(is_struct_complete<SomeOtherData>::value, "You must fully define this struct in common/ipc_structs.h.");
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


zach2good:
I made a little test program that works perfectly:

```cpp

int main(int argc, char* argv[])
{
    const auto bytes_to_zmq_message = [&](const auto& bytes) -> zmq::message_t {
        zmq::message_t msg = zmq::message_t(bytes.size());
        std::memcpy(msg.data(), bytes.data(), bytes.size());
        return msg;
    };

    // Dealer sends to router
    auto dealerThread = std::thread([&]() {
        // ZMQ Setup
        spdlog::info(fmt::format("Dealer: Setting up ZMQ"));
        zmq::context_t ctx;
        zmq::socket_t sock(ctx, zmq::socket_type::dealer);
        sock.connect("tcp://127.0.0.1:6666");

        while (true)
        {
            std::vector<zmq::message_t> send_msgs;

            // Build message
            send_msgs.emplace_back(bytes_to_zmq_message(ipc::to_bytes(ipc::SomeData{
                .messages = { "Hello", "World" }
            })));

            // Send
            if (zmq::send_multipart(sock, send_msgs))
            {
                spdlog::info(fmt::format("Dealer: Sent message"));
            }
        }
    });

    // Router gets from dealer
    auto routerThread = std::thread([&]() {
        // ZMQ Setup
        spdlog::info(fmt::format("Router: Setting up ZMQ"));
        zmq::context_t ctx;
        zmq::socket_t sock(ctx, zmq::socket_type::router);
        sock.bind("tcp://127.0.0.1:6666");

        // Prepare handler
        ipc::IPCMessageHandler handler;

        while (true)
        {
            std::vector<zmq::message_t> recv_msgs;
            const auto ret = zmq::recv_multipart(sock, std::back_inserter(recv_msgs), zmq::recv_flags::dontwait);
            if (ret.has_value())
            {
                // Discard first message, this is the identity of the sender if we want to reply
                const auto identity = std::move(recv_msgs.front());
                recv_msgs.erase(recv_msgs.begin());

                spdlog::info(fmt::format("Router: Received message of size {}", recv_msgs.size()));

                for (const auto& zmq_msg : recv_msgs)
                {
                    // TODO: Make handleMessage accept a zmq::message_t
                    handler.handleMessage(std::vector<uint8_t>(static_cast<const uint8_t*>(zmq_msg.data()), static_cast<const uint8_t*>(zmq_msg.data()) + zmq_msg.size()));
                }
            }
        }
    });
```

Xaver-DaRed (approved):

claywar (approved):

WinterSolstice8 (approved):

PR #6360, opened 2024-10-16
[Lua] Fix distance of mob astral flow skills and related issues

TracentEden2:
**_I affirm:_**

This fixes several small issues related to mob astral flow including:

1 All mob astral-flow-based skills should have 30 yalm range. The different types are:

A) SMN mobs that spawn an untargetable avatar instantly when using Astral Flow and the avatar then immediately uses a move (like Diamond Dust). Several examples include Kirin, Crimson-toothed Pawberry, Viscount Morax, etc. Range capture by Carnap from Crimson-toothed Pawberry [here](https://www.youtube.com/watch?v=YFvxUhx4moE).

B) SMN mobs that spawn a (targetable) avatar on engage and later use Astral Flow causing that avatar to use a move (like Diamond Dust). Examples include any SMN beastmen in Dynamis. Captures of Ranges by Carnap from Dynamis-Valk [here](https://www.youtube.com/watch?v=bBxfi0b2rDE) and  [here](https://www.youtube.com/watch?v=KFY7puh0d5I).

C) Mobs that actually are avatars that use a move (like Diamond Dust) as their normal 2hr. Examples include the Trial By avatars and Mystic Avatars in Temenos. Captures of Range by Carnap from Trial By fights [here](https://youtu.be/P90Eo7TpuNQ), [here](https://youtu.be/FdnL50soDxM), and [here](https://youtu.be/SLRj4CAHi_o).

2 The mob skills for Type B should use the mobskill ID of Type C (rather than Type A). Capture of mobskill ID by Siknoz from Dynamis Bastok [here](https://youtu.be/_DgjxpcvDpE?t=1572)

3 The actual astral flow animation for Type B should be 438 instead of 724. Capture of animation ID by Siknoz from Dynamis Bastok [here](https://youtu.be/_DgjxpcvDpE?t=1572)

Fight the example mobs described above to test range and other changes.

### DIFF
--- a/scripts/actions/mobskills/astral_flow_pet.lua
+++ b/scripts/actions/mobskills/astral_flow_pet.lua
@@ -33,14 +33,14 @@ end
 -- [mobskillId] = { petFamily1, petFamily2, ... }
 local petAstralFlowAbility =
 {
-    [839] = { 36, 381 }, -- Fenrir (Howling Moon)
-    [913] = { 38, 383 }, -- Ifrit (Inferno)
-    [914] = { 45, 388 }, -- Titan (Earthen Fury)
-    [915] = { 40, 384 }, -- Leviathan (Tidal Wave)
-    [916] = { 37, 382 }, -- Garuda (Aerial Blast)
-    [917] = { 44, 387 }, -- Shiva (Diamond Dust)
-    [918] = { 43, 386 }, -- Ramuh (Judgment Bolt)
-    [919] = { 34, 379 }, -- Carbuncle (Searing Light)
+    [xi.mobSkill.HOWLING_MOON_2]  = { 36, 381 }, -- Fenrir (Howling Moon)
+    [xi.mobSkill.INFERNO_1]       = { 38, 383 }, -- Ifrit (Inferno)
+    [xi.mobSkill.EARTHEN_FURY_1]  = { 45, 388 }, -- Titan (Earthen Fury)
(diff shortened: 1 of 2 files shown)


Xaver-DaRed:
While everything looks fine, I have some questions.

For instance, if we are absolutely sure of those skill IDs

Xaver-DaRed (approved):

TracentEden2:
> While everything looks fine, I have some questions.
> 
> For instance, if we are absolutely sure of those skill IDs

At least the captures support the skill IDs found in this PR. Also the use of the same skill IDs for types B and C makes sense because they are both pre-existing targetable avatar mobs that use such moves, while A is much quicker readying time and cannot be stunned (because the avatar is untargetable).

Xaver-DaRed:
After a quick glance, all I can say is we are gonna need to revisit this. Hard. I'll open a PR soon that will make clear why.
However, for the time being, I'm prefectly fine with this changes.

zach2good:
Now that this is in: https://github.com/LandSandBoat/server/pull/6369

Can those magic numbers be enum'd?

Xaver-DaRed:
This will also need a rebase after Frank's AoE PR (which added a new column to the table) went in.

TracentEden2:
> Now that this is in: #6369
> 
> Can those magic numbers be enum'd?

Yep, I have updated the PR to do so, thanks!

TracentEden2:
> This will also need a rebase after Frank's AoE PR (which added a new column to the table) went in.

Yep, updated the PR with a rebase, thanks!

WinterSolstice8 (approved):

PR #6361, opened 2024-10-17
[core] Add nullness checking from static analysis

WinterSolstice8:
**_I affirm:_**

Adds some nullness checking I found in a static code analysis

Test things effected, see no change

### DIFF
--- a/src/map/enmity_container.cpp
+++ b/src/map/enmity_container.cpp
@@ -381,17 +381,21 @@ void CEnmityContainer::SetVE(CBattleEntity* PEntity, const int32 amount)
 void CEnmityContainer::UpdateEnmityFromDamage(CBattleEntity* PEntity, int32 Damage)
 {
     TracyZoneScoped;
-    Damage          = (Damage < 1 ? 1 : Damage);
-    int16 damageMod = battleutils::GetEnmityModDamage(m_EnmityHolder->GetMLevel());
 
-    int32 CE = (int32)(80.f / damageMod * Damage);
-    int32 VE = (int32)(240.f / damageMod * Damage);
-
-    UpdateEnmity(PEntity, CE, VE);
-
-    if (m_EnmityHolder && m_EnmityHolder->m_HiPCLvl < PEntity->GetMLevel())
+    if (m_EnmityHolder)
     {
-        m_EnmityHolder->m_HiPCLvl = PEntity->GetMLevel();
+        Damage          = (Damage < 1 ? 1 : Damage);
+        int16 damageMod = battleutils::GetEnmityModDamage(m_EnmityHolder->GetMLevel());
+
+        int32 CE = (int32)(80.f / damageMod * Damage);
+        int32 VE = (int32)(240.f / damageMod * Damage);
+
+        UpdateEnmity(PEntity, CE, VE);
+
+        if (m_EnmityHolder->m_HiPCLvl < PEntity->GetMLevel())
+        {
+            m_EnmityHolder->m_HiPCLvl = PEntity->GetMLevel();
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


Xaver-DaRed (approved):

src/map/enmity_container.cpp
```diff
@@ -381,17 +381,21 @@ void CEnmityContainer::SetVE(CBattleEntity* PEntity, const int32 amount)
 void CEnmityContainer::UpdateEnmityFromDamage(CBattleEntity* PEntity, int32 Damage)
 {
     TracyZoneScoped;
-    Damage          = (Damage < 1 ? 1 : Damage);
-    int16 damageMod = battleutils::GetEnmityModDamage(m_EnmityHolder->GetMLevel());
 
-    int32 CE = (int32)(80.f / damageMod * Damage);
-    int32 VE = (int32)(240.f / damageMod * Damage);
+    if (m_EnmityHolder)
```
zach2good, line 385: Can we invert this so the whole function isn't indented?

`if m_EnmityHolder == nullptr, return`

src/map/merit.cpp
```diff
@@ -340,7 +340,7 @@ void CMeritPoints::RaiseMerit(MERIT_TYPE merit)
 {
     Merit_t* PMerit = GetMeritPointer(merit);
 
-    if (m_MeritPoints >= PMerit->next && PMerit->count < PMerit->upgrade && GetMeritCountInSameCategory(merit) < meritCatInfo[GetMeritCategory(merit)].MaxPoints)
+    if (PMerit && m_MeritPoints >= PMerit->next && PMerit->count < PMerit->upgrade && GetMeritCountInSameCategory(merit) < meritCatInfo[GetMeritCategory(merit)].MaxPoints)
```
zach2good, line 343: Similarly here, should we be doing `if PMerit is nullptr, message, and bail out early` instead of rolling it into all of these other checks?

src/map/merit.cpp
```diff
@@ -372,30 +372,32 @@ void CMeritPoints::RaiseMerit(MERIT_TYPE merit)
 void CMeritPoints::LowerMerit(MERIT_TYPE merit)
 {
     Merit_t* PMerit = GetMeritPointer(merit);
-
-    if (PMerit->count > 0)
+    if (PMerit)
```
zach2good, line 375: Same here, check and bail out please

src/map/packet_system.cpp
```diff
@@ -4855,7 +4855,7 @@ void SmallPacket0x077(map_session_data_t* const PSession, CCharEntity* const PCh
     {
         case 0: // party
         {
-            if (PChar->PParty != nullptr && PChar->PParty->GetLeader() == PChar)
+            if (PChar && PChar->PParty != nullptr && PChar->PParty->GetLeader() == PChar)
```
zach2good, line 4858: Same here, we can check PChar once before the switch and then continue afterwards with impunity

src/map/zone_entities.cpp
```diff
@@ -1300,7 +1300,7 @@ void CZoneEntities::PushPacket(CBaseEntity* PEntity, GLOBAL_MESSAGE_TYPE message
             case CHAR_INRANGE_SELF: // NOTE!!!: This falls through to CHAR_INRANGE so both self and the local area get the packet
             {
                 TracyZoneCString("CHAR_INRANGE_SELF");
-                if (PEntity->objtype == TYPE_PC)
+                if (PEntity && PEntity->objtype == TYPE_PC)
```
zach2good, line 1303: Theres part of a PEntity check in the GM section, but the PEntity check can go with or next to the packet check at the very start

PR #6365, opened 2024-10-17
Chore: Replace magic numbers with enum in Yurim lua

loxleygames:
**_I affirm:_**

Replaced all item IDs with enums and reformatted table based on current style standards.
Not looking to refactor into IF at this time.

Re-tested all quest steps and dialog with updated IDs. Everything behaves as expected.


Xaver-DaRed (approved):

zach2good (approved):
Well, that original script is gross, thanks for this!

PR #6369, opened 2024-10-18
[Chore] Add avatar astral flows and other missing mobskill copies to mobskill enum file

Xaver-DaRed:
**_I affirm:_**

A little extra work I made while reviewing PR #6360 

That PR doesnt need any extra work, right now. But may need revisiting.

None.


TracentEden2:
Looking good, I do know that 839 (Howling Moon) is used by Fenrir in Moonlit Path (see capture [here](https://youtu.be/7UOb-WlGgv0?t=138)). I will make a PR to make Fenrir more retail accurate at some point.

zach2good (approved):

PR #6373, opened 2024-10-19
[Lua] Fix several Den of Rancor mob issues

TracentEden2:
**_I affirm:_**

1. Add to Hakutaku the correct gil, status immunities, silence meva, double attack, and triple attack based on retail [capture](https://youtu.be/ZQMUnCc5Zko) by Siknoz. Also add similar amount of MP as verified HP (as no reports of Hakutaku running out of MP). 

2. Double the main damage of Ogama when under 50% HP (based on [retail video](https://www.youtube.com/watch?v=qHVbD0MhCxo) showing damage as HP changes).

Fight Hakutaku and Ogama and compare to retail capture and retail video.

### DIFF
--- a/scripts/zones/Den_of_Rancor/mobs/Ogama.lua
+++ b/scripts/zones/Den_of_Rancor/mobs/Ogama.lua
@@ -5,6 +5,25 @@
 ---@type TMobEntity
 local entity = {}
 
+entity.onMobFight = function(mob, target)
+    local mobHPP = mob:getHPP()
+    local appliedDmgBoost = mob:getLocalVar('appliedDmgBoost')
+
+    if
+        mobHPP <= 50 and
+        appliedDmgBoost == 0
+    then
+        mob:addMod(xi.mod.MAIN_DMG_RATING, mob:getMainLvl() + 2)
+        mob:setLocalVar('appliedDmgBoost', 1)
+    elseif
+        mobHPP > 50 and
+        appliedDmgBoost == 1
+    then
+        mob:delMod(xi.mod.MAIN_DMG_RATING, mob:getMainLvl() + 2)
+        mob:setLocalVar('appliedDmgBoost', 0)
+    end
+end
+
 entity.onMobDeath = function(mob, player, optParams)
     xi.hunts.checkHunt(mob, player, 398)
 end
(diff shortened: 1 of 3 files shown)


scripts/zones/Den_of_Rancor/mobs/Ogama.lua
```diff
@@ -5,6 +5,25 @@
...
+        mob:addMod(xi.mod.MAIN_DMG_RATING, mob:getMainLvl() + 2)
+        mob:setLocalVar('appliedDmgBoost', 1)
+    elseif
+        mobHPP > 50 and
+        appliedDmgBoost == 1
+    then
+        mob:delMod(xi.mod.MAIN_DMG_RATING, mob:getMainLvl() + 2)
+        mob:setLocalVar('appliedDmgBoost', 0)
+    end
+end
```
zach2good, line 25: Cool two-way latch 👍

zach2good (approved):

scripts/zones/Den_of_Rancor/mobs/Ogama.lua
Xaver-DaRed: Why not simply set (not add, set) new dmg rating when under 50% hp, set original dmg rating in onMobSpawn and be done with the local var, simplifying onMobFight logic?

onMobFight is run per tick, so we want it as simple as possible

TracentEden2: > Why not simply set (not add, set) new dmg rating when under 50% hp, set original dmg rating in onMobSpawn and be done with the local var, simplifying onMobFight logic?
> 
> onMobFight is run per tick, so we want it as simple as possible

If the mob heals back above 50% while fighting or roaming then I think we want the boost to be removed, thus the need for the check.

Xaver-DaRed: The mob has no regen nor any healing mechanics whatsoever though.
As easy as placing it onMobRoam aswell

TracentEden2: > The mob has no regen nor any healing mechanics whatsoever though. As easy as placing it onMobRoam aswell

Yeah but if we later find out that some mobskill it has does add regen on retail then the logic becomes incorrect. You have a good point about simplicity but I guess just my preference to make it more robust at the cost of more checks.

Xaver-DaRed (approved):

PR #6374, opened 2024-10-20
[core] Make UDP sockets nonblocking to print error on EWOULDBLOCK

WinterSolstice8:
**_I affirm:_**

Make UDP sockets nonblocking to print error on EWOULDBLOCK

Horizon has had a new event where they have >350 people in a single zone and there's a ton of lag. This PR is an attempt to prove/disprove that too many people causes sockets to block from not having enough buffer to send.

Run server, see no changes under normal load.

### DIFF
--- a/src/common/socket.cpp
+++ b/src/common/socket.cpp
@@ -1241,6 +1241,8 @@ int32 makeBind_udp(uint32 ip, uint16 port)
     server_address.sin_addr.s_addr = htonl(ip);
     server_address.sin_port        = htons(port);
 
+    set_nonblocking(fd, true);
+
     result = sBind(fd, (struct sockaddr*)&server_address, sizeof(server_address));
     if (result == SOCKET_ERROR)
     {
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
Finished testing on Linux and Windows. Can't test OSX but not sure why it wouldn't work.

src/common/socket.cpp
```diff
@@ -1292,7 +1294,43 @@ int32 recvudp(int32 fd, void* buff, size_t nbytes, int32 flags, struct sockaddr*
...
-    return sSendto(fd, (const char*)buff, (int)nbytes, flags, from, addrlen);
+    int retval = 0;
+
+    // Block while we can't send data to emulate old behavior...
+    while ((retval = sSendto(fd, (const char*)buff, (int)nbytes, flags, from, addrlen)) < 0)
+    {
+        if (retval < 0)
+        {
+            int err = 0;
+#ifdef WIN32_
```
zach2good, line 1305: I believe you don't need to ifdef this, we have win/unix things like this defined:
```cpp
#define sErrno         WSAGetLastError()
#define S_ENOTSOCK     WSAENOTSOCK
#define S_EWOULDBLOCK  WSAEWOULDBLOCK
#define S_EINTR        WSAEINTR
#define S_ECONNABORTED WSAECONNABORTED

and

#define sErrno         errno
#define S_ENOTSOCK     EBADF
#define S_EWOULDBLOCK  EAGAIN
#define S_EINTR        EINTR
#define S_ECONNABORTED ECONNABORTED
```

So you'd check against `S_ENOTSOCK || S_EWOULDBLOCK`, or somesuch

zach2good: And for OSX in general: if it compiles, assume it's working. It's a secondary platform for us that exists so I can make builds on my Mac :D

PR #6375, opened 2024-10-20
[Lua] Handle chocobo ticket and chocopass trade

jamesbradleym:
**_I affirm:_**

Adds the ability to trade Chocopass and Chocobo Tickets to Chocobo Renter NPCs for free rides.

On trade with a Chocobo Rental NPC, Chocobo rental price will be 0.

Chocopass:
Does not require `Chocobo License` Key Item or have level restrictions.
Only works in 3 start nations, [nothing happens if traded outside of these nations](https://ffxi.allakhazam.com/db/item.html?fitem=9293#:~:text=Ummm...Hello%2C%20Well%20Today%20I%20Tried%20using%20a%20Chocopass%20in%20Jeuno%2C%20and%20I%20traded%20it%20to%20the%20Guy%20near%20the%20Choco%27s%20and%20nothing%20happend.%20Tried%20it%20again%2C%20and%20again...%20Nothing...Why%20does%20it%20do%20that%3F).
Does not wear on zone (Confirmed by Siknoz)
3 Minute wear
[Capture](https://youtu.be/J-Nq85y7oW4)

Chocobo Ticket:
[Capture](https://youtu.be/C-r_iGZQGmY)

Adapted from: https://github.com/AirSkyBoat/AirSkyBoat/pull/241/commits/c2ff2e3195c8c7a4217a000c76c3049999bd67f0

Trade a `Free Chocopass` or `Chocobo Ticket` to a Chocobo Renter NPC

!additem 1789 (Free Chocopass)
!additem 1514 (Chocobo Ticket)

### DIFF
--- a/scripts/globals/chocobo.lua
+++ b/scripts/globals/chocobo.lua
@@ -81,6 +81,50 @@ xi.chocobo.initZone = function(zone)
     end
 end
 
+xi.chocobo.renterOnTrade = function(player, npc, trade, eventSucceed, eventFail)
+    local zoneId = player:getZoneID()
+    local info   = chocoboInfo[zoneId]
+
+    if not info then
+        printf('[warning] player %s passed bad zoneId %i in xi.chocobo.renterOntrade', player:getName(), zoneId)
+        return
+    end
+
+    local validChocopassZones = set({
+        xi.zone.WINDURST_WOODS,
+        xi.zone.BASTOK_MINES,
+        xi.zone.SOUTHERN_SAN_DORIA,
+    })
+
+    if
+        npcUtil.tradeHasExactly(trade, xi.item.FREE_CHOCOPASS) and
+        validChocopassZones[zoneId] -- Chocopass does nothing for Non-Starter Cities
+    then
+        local currency = player:getGil()
+        local price = 0
+        player:setLocalVar('Chocopass', 1)
+        player:setLocalVar('ChocopassDuration', 180)
+        player:startEvent(eventSucceed, price, currency)
+    elseif npcUtil.tradeHasExactly(trade, xi.item.CHOCOBO_TICKET) then
+        if not info.past then -- Does nothing in past zones
+            if
+                player:getMainLvl() >= 20 and
+                player:hasKeyItem(xi.ki.CHOCOBO_LICENSE)
+            then
+                local currency = player:getGil()
+                local price    = 0
+                local duration = 1800 + (player:getMod(xi.mod.CHOCOBO_RIDING_TIME) * 60)
+
+                player:setLocalVar('Chocopass', 1)
+                player:setLocalVar('ChocopassDuration', duration)
+                player:startEvent(eventSucceed, price, currency)
+            else
+                player:startEvent(eventFail) -- Fail Closed
+            end
+        end
+    end
+end
+
 xi.chocobo.renterOnTrigger = function(player, eventSucceed, eventFail)
     local mLvl   = player:getMainLvl()
     local zoneId = player:getZoneID()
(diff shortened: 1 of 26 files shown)


scripts/globals/chocobo.lua
```diff
@@ -118,8 +150,21 @@ xi.chocobo.renterOnEventFinish = function(player, csid, option, eventSucceed)
...
+            printf('[warning] player %s passed bad zoneId %i in xi.chocobo.renterOnEventFinish', player:getName(), zoneId)
+            return
+        end
+
+        if trade == 1 then-- If the player used a Chocopass/Chocobo Ticket
+            player:setLocalVar('Chocopass', 0)
+            player:setLocalVar('ChocopassDuration', 0)
+            player:tradeComplete()
+
+            duration = player:getLocalVar('ChocopassDuration')
```
lapislosh: I think this needs to happen before line 163

scripts/globals/chocobo.lua
```diff
@@ -118,8 +150,21 @@ xi.chocobo.renterOnEventFinish = function(player, csid, option, eventSucceed)
...
-        if info then
+        if not info then
+            printf('[warning] player %s passed bad zoneId %i in xi.chocobo.renterOnEventFinish', player:getName(), zoneId)
+            return
+        end
+
+        if trade == 1 then-- If the player used a Chocopass/Chocobo Ticket
+            player:setLocalVar('Chocopass', 0)
+            player:setLocalVar('ChocopassDuration', 0)
+            player:tradeComplete()
```
cocosolos, line 175: If you can do a trade with more than 1 quantity (e.g. 2 chocobo tickets) and it still succeeds then `trade:hasItemQty` isn't the right approach. If you can have various other items in the trade window and it succeeds then you shouldn't be using `player:tradeComplete()` unless all of those other items are also consumed. You should probably be using `npcUtil.tradeHas` and `player:confirmTrade()`

scripts/globals/chocobo.lua
```diff
@@ -81,6 +81,38 @@ xi.chocobo.initZone = function(zone)
     end
 end
 
+xi.chocobo.renterOnTrade = function(player, npc, trade, eventSucceed, eventFail)
+    local zoneId = player:getZoneID()
+    local info   = chocoboInfo[zoneId]
+    if
+        trade:hasItemQty(xi.item.FREE_CHOCOPASS, 1) and
+        zoneId == (xi.zone.WINDURST_WOODS or xi.zone.BASTOK_MINES or xi.zone.SOUTHERN_SAN_DORIA) -- Chocopass does nothing for Non-Starter Cities
```
cocosolos: This doesn't work. This is just `zoneId == xi.zone.WINDURST_WOODS`. Each zone needs to be checked individually. If you're pulling a commit from a different fork, you need to check if there were any changes/bug fixes applied to the code after the commit.

cocosolos (commented):
> I have tested my code and the things my code has changed since the last commit in the PR and will test after any later commits.

More thorough testing should be done. You shouldn't just be testing if something works, but also if you can break it or get unexpected results by doing unexpected things.

scripts/globals/chocobo.lua
jamesbradleym: Appreciate the flag here - pulled the valid zones out and made a key check instead.

scripts/globals/chocobo.lua
jamesbradleym: updated to `npcUtil.tradeHasExactly` 🙏🏼

scripts/globals/chocobo.lua
jamesbradleym: Bah, thanks for spotting that

scripts/globals/chocobo.lua
```diff
@@ -81,6 +81,45 @@ xi.chocobo.initZone = function(zone)
     end
 end
 
+xi.chocobo.renterOnTrade = function(player, npc, trade, eventSucceed, eventFail)
+    local zoneId = player:getZoneID()
+    local info   = chocoboInfo[zoneId]
+
+    local validChocopassZones = {
```
zach2good: We have `set` for this

scripts/globals/chocobo.lua
```diff
@@ -81,6 +81,45 @@ xi.chocobo.initZone = function(zone)
     end
 end
 
+xi.chocobo.renterOnTrade = function(player, npc, trade, eventSucceed, eventFail)
+    local zoneId = player:getZoneID()
+    local info   = chocoboInfo[zoneId]
```
zach2good, line 86: You check for `info` being nil in `renterOnEventFinish`, but not here

scripts/globals/chocobo.lua
jamesbradleym: Done.

scripts/globals/chocobo.lua
jamesbradleym: Done.

WinterSolstice8 (approved):

jamesbradleym:
Any other comments/feedback for this?

PR #6382, opened 2024-10-22
[Lua] Fix several Eastern Altepa Desert and Fei'yin NM issues

TracentEden2:
**_I affirm:_**

The PR fixes several Eastern Altepa Desert and Fei'yin NM issues:
1. Add to Decurio I-III Invincible use and correct HP (retail video [here](https://youtu.be/FqRK75tqIag?t=210), JP wiki [here](https://wiki.ffo.jp/html/20863.html))
2. Add to Miser Murphy a large endrain effect, status immunities, and correct HP (KnowOne capture [here](https://youtu.be/u_p6Vv_pDH0))
3. Add to each shadow NM an Always Aggro mod (retail video for an example shadow NM [here](https://youtu.be/Gay5qFbvZtU?t=8))
4. Add to Mind Hoarder an endrain effect, status immunities, magic taken mod, correct attack speed, correct mobskill, and correct HP (Siknoz capture [here](https://youtu.be/zAMHaPcdRVs))
5. Add to Altedour I Tavnazia endrain effect, status immunities, HP, regain, and missing dialog upon death (for quest Pieuje's Decision) (Aether capture [here](https://youtu.be/PTgMbnjW1yg?t=2345))

The Centurio X-III bodyguard behavior was moved (along with other bodyguard NMs) to a separate PR for ease of reviewing and better separated PRs.

Fight these mobs and compared to the captures or retail videos


scripts/zones/FeiYin/mobs/Miser_Murphey.lua
almuth150, line 1: Minor spelling mistake with the name of this file - should be Miser_Murphy.lua, not Miser_Murphey.lua

almuth150:
Might be a silly question, but what happens to the bodyguards during a conquest tally on retail? Particularly if the region becomes no longer controlled by beastmen.

scripts/zones/FeiYin/mobs/Miser_Murphey.lua
TracentEden2: > Minor spelling mistake with the name of this file - should be Miser_Murphy.lua, not Miser_Murphey.lua

Thanks for comment, I updated the PR!

TracentEden2:
> Might be a silly question, but what happens to the bodyguards during a conquest tally on retail? Particularly if the region becomes no longer controlled by beastmen.

That is unknown (as no capture for that situation on retail). The current logic of this PR is that they do not despawn in such a situation. Siknoz is currently checking their behavior if Centurio is killed (while they are still alive) and some variations of that situation but maybe too much of a pain to verify the change in conquest for now. Though I can ask him.

scripts/mixins/rotz_bodyguarded_nm.lua
```diff
@@ -0,0 +1,77 @@
...
+
+g_mixins = g_mixins or {}
+
+g_mixins.bodyguard = function(bodyguardedNM)
+    bodyguardedNM:addListener('SPAWN', 'ROTZ_BODYGUARDED_NM_SPAWN', function(mob)
+        -- retail captures show these mods are not dependent on region control
+        mob:setMod(xi.mod.UDMGPHYS, -5000)
+        mob:setMod(xi.mod.UDMGRANGE, -5000)
+        mob:setMod(xi.mod.UDMGBREATH, -5000)
+        mob:setMod(xi.mod.UDMGMAGIC, -5000)
```
Xaver-DaRed: 2 things:
1. This are literally all damage global types. Just use the 1 modifier for them all: `xi.mod.DMG`
2. Even if all of this mobs happen to share the -50% damage taken, this mixing should stick to the bodyguard logic. What this means is this modifier should be placed in themob script

scripts/zones/Eastern_Altepa_Desert/mobs/Centurio_XII-I.lua
```diff
@@ -2,11 +2,12 @@
...
 --   NM: Centurio XII-I
 -----------------------------------
+mixins = { require('scripts/mixins/job_special'), require('scripts/mixins/rotz_bodyguarded_nm') }
+-----------------------------------
 ---@type TMobEntity
 local entity = {}
 
-entity.onMobDeath = function(mob, player, optParams)
-end
+-- all damage taken mods and body guard functionality in the rotz_bodyguarded_nm mixin
```
Xaver-DaRed: As I said in the comment above, do place the modifiers in the mob script and leave the mixing to focus solely on the bodyguard logic.

IF the modifiers were dependeant on said logic, that would be a different thing and the mods would actually belong there, but since they are independent from one another, lets keep them separate.

scripts/zones/FeiYin/mobs/Mind_Hoarder.lua
```diff
@@ -5,6 +5,31 @@
...
 
+entity.onMobInitialize = function(mob)
+    mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
+end
+
+entity.onMobWeaponSkillPrepare = function(mob, target)
+    -- mind hoarder only uses mind drain
+    -- therefore override the normal list
+    return xi.mobSkill.MIND_DRAIN
+end
```
Xaver-DaRed: I am missing something. If the mob only uses 1 skill no matter what, why not have their list only have the one skill they can use?

scripts/mixins/rotz_bodyguarded_nm.lua
TracentEden2: > 2 things:
> 
> 1. This are literally all damage global types. Just use the 1 modifier for them all: `xi.mod.DMG`
> 2. Even if all of this mobs happen to share the -50% damage taken, this mixing should stick to the bodyguard logic. What this means is this modifier should be placed in themob script

1. Yeah I thought about using DMG but that is capped at -50% so if the mob also uses some DMG or DMGPHY modifying buff then I think it would have no impact as DMG would already be at the cap.
2. Yep, thanks for feedback I have updated the PR!

scripts/zones/FeiYin/mobs/Mind_Hoarder.lua
TracentEden2: > I am missing something. If the mob only uses 1 skill no matter what, why not have their list only have the one skill they can use?

They do not have their own skill list currently as they share a list with generic pots, thus this is alternative to creating a new list just for mind hoarder. I can do that instead if it would be better.

scripts/mixins/rotz_bodyguarded_nm.lua
TracentEden2: Just to note here as well the Centurio X-III changes and bodyguard behavior was moved to a separate PR for ease of review and better separation of PRs. I will link to this comment in that PR.

scripts/zones/FeiYin/mobs/Mind_Hoarder.lua
Xaver-DaRed: Yes, I would definately give it its own list, instead of coding an exception.

Maybe in a future, we will be able to define skill lists in the mob-script directly and be done with that sql, but for now...

Xaver-DaRed:
Overall looks very good. Just one last thing and its good to go.

scripts/zones/FeiYin/mobs/Mind_Hoarder.lua
TracentEden2: > Yes, I would definately give it its own list, instead of coding an exception.
> 
> Maybe in a future, we will be able to define skill lists in the mob-script directly and be done with that sql, but for now...

Yep, I updated the PR to add its own list, thanks!

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #6388, opened 2024-10-25
Core: Prevent delGil accepting negative values

loxleygames:
**_I affirm:_**

Passing a negative value to `delGil` gives the player gil which can have severe consequences if bad values make their way to the function. As we have `addGil` we don't need this behavior.

Also added checks to make sure the player has enough gil before deleting it.

```cpp
!exec player:delGil(-100)
```
Should produce an error and not give the player 100 gil

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -8813,18 +8813,24 @@ bool CLuaBaseEntity::delGil(int32 gil)
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

Xaver-DaRed:
I wonder, while we are here, if we should add more safety to not allow deleting more gil than the player currently has

zach2good (approved):

PR #6392, opened 2024-10-28
Core: Use EntityID_t for PWideScanTarget

zach2good:
**_I affirm:_**

We love scope creep. I think this is going to become my long awaited introduction of some sort of `EntityToken` structure. We already have `EntityID_t` which does a lot of what I want. This will refactor it into the token, add the `.resolve()` function, and then replace all existing usaged with the new one.

Obviously will include proper testing of everything, renaming, refactoring, comments, all the good stuff. I'll also keep an eye out for anywhere I might be able to add this for stability that won't be HUGE (so, PPet, PMaster etc. is probably too big for this round, but it's the obvious candidate)

### DIFF
--- a/src/map/entities/baseentity.cpp
+++ b/src/map/entities/baseentity.cpp
@@ -31,8 +31,12 @@
 
 #include <cstring>
 
+// TODO: Don't use static here, use a proper unique identifier generator.
+static uint64 sUniqueIdentifier = 0;
+
 CBaseEntity::CBaseEntity()
-: id(0)
+: hash(++sUniqueIdentifier)
+, id(0)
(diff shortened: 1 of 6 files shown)


zach2good:
I'm warming up to the interface looking like:

```cpp
const bool wasResolved = token.resolve([&](auto* PEntity)
{
    // Use PEntity here safely from token
});
if (!wasResolved)
{
    // Log error
}
```

It really locks things up so you can't possibly shoot yourself in the foot, at the cost of being more verbose

zach2good:
Replaced by https://github.com/LandSandBoat/server/pull/6449

PR #6297, opened 2024-10-01
[Core] Add setting and logic for Bazaar logging

clanofartisans:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

Adds logging in several places involving bazaars.
* New setting in `logging.lua`: `DEBUG_BAZAARS` ~ Defaults `false`
* Log: When setting bazaar message
* Log: When leaving a bazaar
* Log: When viewing wares
* Log: When attempting to purchase an invalid quantity (`Warning`)
* Log: When purchasing fails because it's your own bazaar
* Log: When purchasing fails because inventory is full
* Log: When purchase packet is received but gil is insufficient (`Warning`)
* Log: When purchase is successful
* Log: When a bazaar has sold its last item
* Log: When setting a price on an item
* Log: When a bazaar is closed
* Log: When opening the "Set Prices" menu

## Steps to test these changes

Set `DEBUG_BAZAARS` to `true` in `logging.lua`. Attempt all the above actions.

## Example Logs
```
[10/01/24 16:23:02:163][map][debug] Bazaar Interaction [Set Message] - Character: Hugin, Message: 'Test Message                            asdf                                    Line three!
          ' (SmallPacket0x0DE:6492)
[10/01/24 16:23:05:463][map][debug] Bazaar Interaction [Setting Prices] - Character: Hugin (SmallPacket0x10B:8049)
[10/01/24 16:23:13:585][map][debug] Bazaar Interaction [Price Set] - Character: Hugin, Item: onion_sword, Price: 5 (SmallPacket0x10A:8014)
[10/01/24 16:23:20:633][map][debug] Bazaar Interaction [View Wares] - Buyer: Hugin, Seller: Hugin (SmallPacket0x105:7787)
[10/01/24 16:23:22:703][map][debug] Bazaar Interaction [Purchase Failed / Self Bazaar] - Character: Hugin, Item: onion_sword (SmallPacket0x106:7840)
[...]

PR #6395, opened 2024-10-30
Core: Enhancements to logging Bazaar and Vendor transactions

loxleygames:
**_I affirm:_**

Improved the logging of some transactions to provide better metadata

* Added logging of gil amount to bazaar purchases
* Added logging of gil amount to vendor sales

Tested in game using bazaar and vendor

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -5052,9 +5052,11 @@ void SmallPacket0x085(map_session_data_t* const PSession, CCharEntity* const PCh
             return;
         }
 
-        charutils::UpdateItem(PChar, LOC_INVENTORY, 0, quantity * PItem->getBasePrice());
+        const auto cost = quantity * PItem->getBasePrice();
+
+        charutils::UpdateItem(PChar, LOC_INVENTORY, 0, cost);
         charutils::UpdateItem(PChar, LOC_INVENTORY, slotID, -(int32)quantity);
-        ShowInfo("SmallPacket0x085: Player '%s' sold %u of itemID %u [to VENDOR] ", PChar->getName(), quantity, itemID);
+        ShowInfo("SmallPacket0x085: Player '%s' sold %u of itemID %u (Total: %u gil) [to VENDOR] ", PChar->getName(), quantity, itemID, cost);
         PChar->pushPacket(new CMessageStandardPacket(nullptr, itemID, quantity, MsgStd::Sell));
         PChar->pushPacket(new CInventoryFinishPacket());
         PChar->Container->setItem(PChar->Container->getSize() - 1, 0, -1, 0);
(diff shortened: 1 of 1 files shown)


src/map/packet_system.cpp
```diff
@@ -7982,7 +7984,7 @@ void SmallPacket0x106(map_session_data_t* const PSession, CCharEntity* const PCh
         PTarget->pushPacket(new CInventoryItemPacket(PBazaar->GetItem(SlotID), LOC_INVENTORY, SlotID));
         PTarget->pushPacket(new CInventoryFinishPacket());
 
-        DebugBazaarsFmt("Bazaar Interaction [Purchase Successful] - Buyer: {}, Seller: {}, Item: {}, Qty: {}", PChar->name, PTarget->name, PItem->getName(), Quantity);
+        ShowInfo("SmallPacket0x106: Player '%s' purchased %u of item '%s' (%u) from player '%s' (Bazaar) (Total: %u gil)", PChar->name, Quantity, PItem->getName(), PItem->getID(), PTarget->name, PriceWithTax);
```
zach2good: Would it be more useful to bin all of this, and just introduce an audit_bazaar SQL table, and then async write into it like we do with chat messages?

That aside, why the rollback away from the safer and less error-prone Fmt logger?

Also also, we don't have it written as a rule anywhere, but it makes sense that the default amount of logging isn't spamming regular player actions. All of that sort of _additional_ logging should be opt in. I like to think of the default user as someone playing with their friends, and then the enhanced case is servers with lots of players who want to flip additional switches to turn more server-y things on.

src/map/packet_system.cpp
```diff
@@ -5052,9 +5052,11 @@ void SmallPacket0x085(map_session_data_t* const PSession, CCharEntity* const PCh
             return;
         }
 
-        charutils::UpdateItem(PChar, LOC_INVENTORY, 0, quantity * PItem->getBasePrice());
+        auto cost = quantity * PItem->getBasePrice();
```
zach2good: if this isn't ever changing, this should be `const auto`

src/map/packet_system.cpp
loxleygames: I'd like to create more tables but this will need some discussion and I thought improving the existing logging first was a good approach. More than happy to look at this though. But I can replace with the fmt version for now.

The logging for vendor sales and trades are not opt-in, so it made sense to me that bazaar purchase shouldn't be either. The debug Bazaars setting creates a *massive* amount of extra logging which isn't necessary for most use cases.

zach2good: For now, apply your new information to `DebugBazaarsFmt`, since that was only added a month ago: https://github.com/LandSandBoat/server/pull/6297

and I'll hook up the audit tables probably on Sunday - with truncation jobs in `dbtool` etc.

zach2good (approved):

TracentEden2:
Yeah, I should be able to contribute some code from Horizon that works like Zach's idea, I will make a new PR with the code hopefully in a few weeks.

zach2good:
I'm not sure why the `Linux_LuaLanguageServer` is crying, can you rebase onto `base` and force push back up?

zach2good:
We have a temp fix in `base` if you want to try rebasing to shut the LLS job up 😆

PR #6396, opened 2024-10-30
[Lua+Core] Add interruption of player spells by physical mobskills

TracentEden2:
**_I affirm:_**

This PR adds the potential for interruption of players casting spells when hit for positive damage by a physical mobskill. Retail capture of such an interruption can be seen in Siknoz capture of Kirin [here](https://youtu.be/gZi8NwnXu4M?t=937). The PR assumes that each hit of the mobskill gives an interruption chance (as per discussion in PR comments). The PR adds a `CLuaBaseEntity::tryHitInterrupt` to accomplish this interruption logic.

The PR also refactors the `handleBlock` function and splits it into two functions `isBlocked` and `getDamageReductionForBlock` so that it is more consistent with other functions (like `isParried`) and removes the need for two return parameters. Also the refactor fixes a small bug that was in `handleBlock` related to applying the damage reduction to the hit damage.

Finally the PR renames the `finaldmg` variable to `hitdamage` in the `getSingleHitDamage` function in weaponskills.lua to improve readability and avoid confusion (as a variable named `finaldmg` is used differently in `handleSinglePhysicalHit` mobskill.lua).

Fight a mob and turn off the mobs auto-attacks and start casting a long spell while spamming the mob with !tp 3000 until the mob uses many physical mobskills which hopefully interrupt the spell.

### DIFF
--- a/scripts/globals/combat/physical_utilities.lua
+++ b/scripts/globals/combat/physical_utilities.lua
@@ -910,29 +910,40 @@ xi.combat.physical.calculateBlockRate = function(defender, attacker)
     return blockRate
 end
 
-xi.combat.physical.handleBlock = function(defender, attacker, damage)
-    if
-        xi.combat.physical.canBlock(defender, attacker) and
-        xi.combat.physical.calculateBlockRate(defender, attacker) > math.random(100)
-    then
+xi.combat.physical.getDamageReductionForBlock = function(defender, attacker, damage)
+    -- save original damage for comparison
+    local originalDamage = damage
+
+    -- do not reduce if damage is negative
+    if damage > 0 then
         -- shield def bonus is a flat raw damage reduction that occurs before absorb
-        -- however do not reduce below 0 or if damage is negative
-        if damage > 0 then
-            damage = math.max(0, damage - defender:getMod(xi.mod.SHIELD_DEF_BONUS))
-        end
+        damage = math.max(0, damage - defender:getMod(xi.mod.SHIELD_DEF_BONUS))
 
         if defender:isPC() then
             local shield = defender:getEquippedItem(xi.slot.SUB)
-            local absorb = 100
-            absorb = utils.clamp(absorb - shield:getShieldAbsorptionRate(), 0, 100)
+            local absorb = utils.clamp(100 - shield:getShieldAbsorptionRate(), 0, 100)
             damage = math.floor(damage * (absorb / 100))
-            defender:trySkillUp(xi.skill.SHIELD, attacker:getMainLvl())
         else
             damage = math.floor(damage * 0.5)
         end
     end
 
-    return damage
+    -- return the difference between original and new damage
+    -- in other words the damage reduction (as a flat value)
+    return originalDamage - damage
+end
+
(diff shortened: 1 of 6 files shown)


WinterSolstice8:
don't think there's any reason not to have it try to interrupt each hit, it's well known you can skill up guard/parry more than once on a multihit TP move

if weaponskills are only doing it once that could be wrong too

TracentEden2:
> don't think there's any reason not to have it try to interrupt each hit, it's well known you can skill up guard/parry more than once on a multihit TP move
> 
> if weaponskills are only doing it once that could be wrong too

Yeah, I think you are right and I have updated the PR to reflect this. I think I will wait for now until WS logic is in Lua space to try to fix that. Thanks for the feedback!

WinterSolstice8 (approved):

scripts/globals/combat/physical_utilities.lua
```diff
@@ -932,7 +934,7 @@ xi.combat.physical.handleBlock = function(defender, attacker, damage)
         end
     end
 
-    return damage
+    return damage, didBlock
```
Xaver-DaRed: I'm not a big fan of creating functions that return 2 parameters

Then again, I'm also not a big fan of functions that return a converted value instead of a factor to convert the value.

I would think that, if the value fed is different than the value returned, the block did happen though

zach2good: ^ That's a good point. Is it always the case that a block will result in less damage? Because you can use the value changing as your "did a block happen" signal. It does seem like it's cleaner to just have a function that works out "did I block"? and then apply any changes that need to happen because of the block after - rather than combining them and having this double return

scripts/globals/weaponskills.lua
```diff
@@ -370,7 +370,7 @@ local function getSingleHitDamage(attacker, target, dmg, ftp, wsParams, calcPara
     finaldmg = (dmg + consumeManaBonus(attacker)) * ftp * calcParams.pdif
 
     -- handle blocking
-    finaldmg = xi.combat.physical.handleBlock(target, attacker, finaldmg)
+    finaldmg, _ = xi.combat.physical.handleBlock(target, attacker, finaldmg)
```
zach2good: Yeah, this is signalling that the double return is probably not the best layout for this

scripts/globals/combat/physical_utilities.lua
TracentEden2: Thanks for the suggestions! The updated PR now refactors the `handleBlock` function and splits it into two functions `isBlocked` and `getDamageReductionForBlock` so that it is more consistent with other functions (like `isParried`) and removes the need for two return parameters. The damage reduction value is a flat reduction that is then subtracted from the hit damage.

scripts/globals/weaponskills.lua
TracentEden2: > Yeah, this is signalling that the double return is probably not the best layout for this

Yep, I agree, I have updated the PR as per my reply to the comment above and the updated summary of what the PR does. Thanks for your suggestion

WinterSolstice8 (approved):

zach2good (approved):

PR #6405, opened 2024-11-03
[Lua+Core] Fix 24 hour respawn RoTZ NMs (including bodyguarding)

TracentEden2:
**_I affirm:_**

This PR unifies and fixes behavior for a related group of RoTZ NMs that all have bodyguard mobs and long respawn windows (21-24 hours). These NMs include Centurio XII-I, Meteormauler Zhagtegg, Coo Keja the Unseen, Meww the Turtlerider, Bo'Who Warmonger, and Bright-handed Kunberry. 

All the NMs have two bodyguard mobs that spawn with and follow the NM if the region the NM spawns in is beastmen controlled. This bodyguard behavior is centralized in a new mixin `rotz_bodyguarded_nm.lua` that all of the six NM now include. The bodyguards link normally with the NM (but do not superlink and thus are not in a party with the NM). The bodyguards despawn if they disengage (and roam) at any time while the bodyguarded NM is dead, but they will not despawn if they are still engaged. Also bodyguards do not despawn even if dragged with the NM far away from home (they path back home just following the NM).

Also all of the NMs have a 50% damage taken reduction mod (which is implemented as the 4 `UDMG` mods because of the 50% cap on the `DMG` mod. Also the incorrect mob pool mod (`DMG: -50` ) of these NMs has been removed.

Also related issues with the NMs has been fixed such as their bodyguards spawning as normal respawn mobs or not having spawn coordinates at all. And Centurio_XII has had it's HP corrected.

The behavior of these NMs and bodyguards was validated in several captures and additional retail testing by Siknoz and myself including captures found [here](https://youtu.be/xzbLHSVSYqQ), [here](https://youtu.be/LXFCCpyamEM), [here](https://youtu.be/rZFtVY4G5A8), and an additional capture still to be uploaded.

Spawn any of these NMs with beastmen in control of the region and a player nation in control of the region. Also can test the bodyguard despawn behavior after the NM is killed and then doing !goto playerName.

### DIFF
--- a/scripts/zones/Jugner_Forest/Zone.lua
+++ b/scripts/zones/Jugner_Forest/Zone.lua
@@ -14,6 +14,9 @@ zoneObject.onInitialize = function(zone)
     UpdateNMSpawnPoint(ID.mob.FRAELISSA)
     GetMobByID(ID.mob.FRAELISSA):setRespawnTime(math.random(900, 10800))
 
+    UpdateNMSpawnPoint(ID.mob.METEORMAULER)
+    GetMobByID(ID.mob.METEORMAULER):setRespawnTime(math.random(900, 10800))
+
     xi.conq.setRegionalConquestOverseers(zone:getRegionID())
 
     xi.helm.initZone(zone, xi.helmType.LOGGING)
--- a/scripts/mixins/rotz_bodyguarded_nm.lua
+++ b/scripts/mixins/rotz_bodyguarded_nm.lua
@@ -1,0 +1,60 @@
+-- Mixin for behavior that is common to six RoTZ NMs that have 21-24 hour respawns
+-- These mobs are Centurio XII-I, Meteormauler Zhagtegg, Coo Keja the Unseen, Meww the Turtlerider,
+-- Bo'Who Warmonger, and Bright-handed Kunberry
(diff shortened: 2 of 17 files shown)


zach2good (approved):
All seems reasonable enough to me, but I don't know gameplay things so well. I'm sure the others can weigh in on this better than I can

scripts/zones/Jugner_Forest/Zone.lua
```diff
@@ -14,6 +14,9 @@ zoneObject.onInitialize = function(zone)
     UpdateNMSpawnPoint(ID.mob.FRAELISSA)
     GetMobByID(ID.mob.FRAELISSA):setRespawnTime(math.random(900, 10800))
 
+    UpdateNMSpawnPoint(ID.mob.METEORMAULER)
+    GetMobByID(ID.mob.METEORMAULER):setRespawnTime(math.random(900, 10800))
```
WinterSolstice8, line 18: Interesting.  I would expect CI to trip on this because theoretically GetMobByID can return nil, but it can't for static mobs like this. Just leaving a comment here for Clay to see (you don't need to do anything until CI complains about it)

WinterSolstice8 (approved):

PR #6406, opened 2024-11-04
[lua] Fix Casket Drop Rate Prowess

UmeboshiXI:
**_I affirm:_**

Fixes treasure casket drop rate bonus from GoV Prowess effects. The bonus being based on effect power was being fed into
`rand < utils.clamp(xi.settings.main.CASKET_DROP_RATE + kupowersMMBPower + prowessCasketsPower, 0, 1)`
without any adjustments to convert it into a float. This caused the casket drop rate to be capped out because the value would always be more than 1(The clamp for the float).

This PR divides the effect power by 100 so that the final bonus value reflects the proper percent bonus to be added to the base casket drop rate. An effect power of 4 will now be 0.04 which would equate to a 4% increase.

1. Go to a GoV book zone, complete regime until you get casket rate increase or use `!exec player:addStatusEffect(xi.effect.PROWESS_CASKET_RATE, 4, 0, 0)` to give yourself 1 rank of Prowess Casket Rate bonus(4%).
2. Kill mobs and see that the drop rate is not 100%.
2b. Use a print in the above function to verify the actual drop rate is correct.

### DIFF
--- a/scripts/globals/caskets.lua
+++ b/scripts/globals/caskets.lua
@@ -185,7 +185,7 @@ local function dropChance(player)
     --end
 
     if player:hasStatusEffect(xi.effect.PROWESS_CASKET_RATE) then
-        prowessCasketsPower = casketProwessEffect:getPower()
+        prowessCasketsPower = casketProwessEffect:getPower() / 100
     end
 
     local rand = math.random()


WinterSolstice8 (approved):

scripts/globals/caskets.lua
```diff
@@ -185,7 +185,7 @@ local function dropChance(player)
     --end
 
     if player:hasStatusEffect(xi.effect.PROWESS_CASKET_RATE) then
-        prowessCasketsPower = casketProwessEffect:getPower()
+        prowessCasketsPower = casketProwessEffect:getPower() / 100
```
zach2good, line 188: While you're here and you understand the code, can you change the `local rand = math.random()` below this to be `math.random(1, 100)` (like we're trying to use in the rest of the code?

Would that remove the need for this /100 divisor?

UmeboshiXI: I can change/update it. May need to make some adjustments. I could remove /100 divisor but something similar will still be needed depending on which option you prefer.

xi.settings.main.CASKET_DROP_RATE = 0.10 Default Setting.

kupowersMMBPower and prowessCasketsPower are integers (effect:getPower())

Options:

1. I could change the casket droprate in settings to use 1-100 instead of a 0.0-1.0 float.

2. I could multiply the settings drop rate by 100 to convert it into an integer. (This is the code block right below the original commit that adds/clamps everything)

    `if math.random(1, 100) < utils.clamp((xi.settings.main.CASKET_DROP_RATE * 100) + kupowersMMBPower + prowessCasketsPower, 0, 100) then
        return true
    end`
    
Whichever style/option you prefer.

zach2good: If it's a much wider change, then leave it as it is 👍

zach2good (approved):

PR #6408, opened 2024-11-04
Lua: Fixes condition for qm1 in the Stars of Ifrit quest

loxleygames:
**_I affirm:_**

This NPC should appear between 18 and 6 on full moon. However, the condition is wrong, so that it never appears. This PR corrects the issue. https://www.bg-wiki.com/ffxi/The_Stars_of_Ifrit

During full moon, the NPC will appear between 18 and 6.

### DIFF
--- a/scripts/zones/San_dOria-Jeuno_Airship/Zone.lua
+++ b/scripts/zones/San_dOria-Jeuno_Airship/Zone.lua
@@ -27,8 +27,8 @@ zoneObject.onGameHour = function(zone)
 
     if
         IsMoonFull() and
-        vanadielHour >= 18 and
-        vanadielHour < 6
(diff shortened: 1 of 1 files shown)


scripts/zones/San_dOria-Jeuno_Airship/Zone.lua
```diff
@@ -27,8 +27,8 @@ zoneObject.onGameHour = function(zone)
 
     if
         IsMoonFull() and
-        vanadielHour >= 18 and
-        vanadielHour < 6
+        (vanadielHour >= 18 or
```
zach2good, line 30: Sneaky sneaky conditions!

zach2good (approved):
Good catch 👍

PR #6419, opened 2024-11-08
[Lua] Purge of 'exp' to 'xp', simplified npc_util.lua logic

Flibe-XI:
**_I affirm:_**

Implements a very elegant solution to ambiguous naming convention for IF xp parameter.

If there is truly a need for the handling of both, would recommend 
```
if params['xp'] == nil then
    params['xp'] = params['exp']
end
```
written before each of the three implementations to prioritize "xp" over "exp" as it seems to be the preferred nomenclature in this codebase.

Complete a quest and check amount of exp gained.

### DIFF
--- a/scripts/globals/npc_util.lua
+++ b/scripts/globals/npc_util.lua
@@ -650,11 +650,7 @@ function npcUtil.completeQuest(player, area, quest, params)
         player:messageSpecial(ID.text.BAYLD_OBTAINED, params['bayld'] * xi.settings.main.BAYLD_RATE)
     end
 
-    -- TODO: Find a more elegant way to handle this, but allow for xp vs exp keys.  This should
-    -- be one or the other, not both.
-    if params['exp'] ~= nil and type(params['exp']) == 'number' then
-        player:addExp(params['exp'] * xi.settings.main.EXP_RATE)
-    elseif params['xp'] ~= nil and type(params['xp']) == 'number' then
+    if params['xp'] ~= nil and type(params['xp']) == 'number' then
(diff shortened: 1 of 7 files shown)


Xaver-DaRed:
I would go the oposite way, tbh.

Flibe-XI:
Additionally, are both `params['ki']` and `params['keyItem']`  required? Seems like `keyItem` is more prevalent and can be merged to keyItem as well.

Flibe-XI:
> I would go the oposite way, tbh.

[xp_to_exp](https://github.com/LandSandBoat/server/compare/base...Flibe-XI:LSB:xp_to_exp)

Want to hear feedback about the ki issue and double check for issues before submitting this alternative.

claywar:
> I would go the oposite way, tbh.

Absolutely agree with Xaver here.  There's no harm in being verbose

Flibe-XI:
> > I would go the oposite way, tbh.
> 
> Absolutely agree with Xaver here. There's no harm in being verbose

Can I get clarity on the path forward? Is it to use "exp" instead of "xp"?

Or to maintain having a check for both in all functions, similar to how it is originally?

claywar:
> Can I get clarity on the path forward? Is it to use "exp" instead of "xp"?
> 
> Or to maintain having a check for both in all functions, similar to how it is originally?

My preference is `keyItem` and `exp` for the reward table overall.  My opinion is that it increases readability and they're not overly long as is.  That's why I went that route when converting quests/missions

Flibe-XI:
Excellent. Thanks.  I'll submit the other PR to replace this one.

PR #6420, opened 2024-11-08
[Lua] Make Ix'Aern (Monk) and adds (Qn'aern) more retail accurate

TracentEden2:
**_I affirm:_**

This PR makes Ix'Aern (Monk) and adds (Qn'aern RDM and WHM) more retail accurate via the following changes:
1. Adjusts HPs, status immunities, and levels of Ix'Aern and adds
2. Adjust bracer mode logic to start when any of Ix'Aern or adds goes below 60% HP
3. Remove dependence between adds using 2hrs and bracer mode
4. Delay adding braclets animation to Ix'Aern (Monk) until after hundred fists
5. Change bracer mode delay reduction to apply to both Ix'Aern and adds and use the `DELAY` mod (as `HASTE_ABILITY` is capped and not really for use by mobs), also note the delay is a true delay reduction (that impacts TP gain and TP imparted) as validated by retail testing and thus `DELAY` is appropriate.

These changes are based on retail captures of Ix'Aern by Siknoz [here](https://youtu.be/jj3slbaXClc) and [here](https://youtu.be/6BaCwpw1LGk).

Fight Ix'Aern (Monk) and adds and compare to retail captures


WinterSolstice8:
[haste_ability is still applied](https://github.com/LandSandBoat/server/blob/ce2f580ac4f6a24a6a3e6424425782763c4ea950/src/map/entities/battleentity.cpp#L390) there is a separate mod only for 2H that's additive a few lines down
reducing delay also reduces TP gain/TP given being hit, though I can't really tell from the cap (too many mobs at once) if the delay is actually being reduced, but I haven't seen an instance where that is the case (to memory, at least)

WinterSolstice8:
Though that said, applying haste would reduce recast timers for spells, though in the real world you're going to kill the adds before you actually fight the NM. Maybe we need a separate mod to just reduce auto attack delay only that doesn't modify TP gain or recasts in any way

TracentEden2:
> [haste_ability is still applied](https://github.com/LandSandBoat/server/blob/ce2f580ac4f6a24a6a3e6424425782763c4ea950/src/map/entities/battleentity.cpp#L390) there is a separate mod only for 2H that's additive a few lines down
> reducing delay also reduces TP gain/TP given being hit, though I can't really tell from the cap (too many mobs at once) if the delay is actually being reduced, but I haven't seen an instance where that is the case (to memory, at least)

Ah yeah true, I am not sure then.

TracentEden2:
> [haste_ability is still applied](https://github.com/LandSandBoat/server/blob/ce2f580ac4f6a24a6a3e6424425782763c4ea950/src/map/entities/battleentity.cpp#L390) there is a separate mod only for 2H that's additive a few lines down reducing delay also reduces TP gain/TP given being hit, though I can't really tell from the cap (too many mobs at once) if the delay is actually being reduced, but I haven't seen an instance where that is the case (to memory, at least)

You can find caps of just the NM (1 organ popping) [here](https://youtu.be/7YtDok1lhvo) and [here](https://youtu.be/6BaCwpw1LGk), maybe possible to check for delay reduction from those?

TracentEden2:
> > [haste_ability is still applied](https://github.com/LandSandBoat/server/blob/ce2f580ac4f6a24a6a3e6424425782763c4ea950/src/map/entities/battleentity.cpp#L390) there is a separate mod only for 2H that's additive a few lines down reducing delay also reduces TP gain/TP given being hit, though I can't really tell from the cap (too many mobs at once) if the delay is actually being reduced, but I haven't seen an instance where that is the case (to memory, at least)
> 
> You can find caps of just the NM (1 organ popping) [here](https://youtu.be/7YtDok1lhvo) and [here](https://youtu.be/6BaCwpw1LGk), maybe possible to check for delay reduction from those?

For completeness I did an additional cap to check this and without bracers I was getting 12 TP per hit and with bracers I was getting 10 TP per hit, so seems it is an actual delay reduction. I also verified with Jimmayus and he said that he knows of cases where it is a delay reduction, so pretty interesting.

WinterSolstice8 (approved):

scripts/zones/Grand_Palace_of_HuXzoi/mobs/Ixaern_MNK.lua
```diff
@@ -7,6 +7,38 @@ local ID = zones[xi.zone.GRAND_PALACE_OF_HUXZOI]
...
+    -- slightly delay adding local var to avoid adding bracers to Ix'Mnk
+    -- while Hundred Fists is still active.
+    mob:timer(3000, function(mobArg)
+        mobArg:setLocalVar('enableBracerAnimation', 1)
+    end)
+end
+
 entity.onMobInitialize = function(mob)
     mob:addListener('ITEM_DROPS', 'ITEM_DROPS_IXAERN_MNK', function(mobArg, loot)
         local rate = mob:getLocalVar('[SEA]IxAern_DropRate')
```
LacunaXI, line 44: Unsure if you want to make the change on the drop rate here or if I am overstepping in suggesting it. From what I can gleam in the available data between FFXIDB and the user reports on Clopedia. I want to say the capes from the various Ix'Aerns are a 15% drop rate and not 25%.

This would then also correspond to the official drop rates of Common, Uncommon, and Rare when trading 3, 2 or 1 organs.

TracentEden2: > Unsure if you want to make the change on the drop rate here or if I am overstepping in suggesting it. From what I can gleam in the available data between FFXIDB and the user reports on Clopedia. I want to say the capes from the various Ix'Aerns are a 15% drop rate and not 25%.
> 
> This would then also correspond to the official drop rates of Common, Uncommon, and Rare when trading 3, 2 or 1 organs.

Yeah you are probably correct but I think better if someone puts such rate changes in a separate PR so they are all changed and harmonized at once.

Xaver-DaRed (approved):

PR #6431, opened 2024-11-14
Port Sengikori from Cactuar & implement gear effects

WinterSolstice8:
**_I affirm:_**

As title says, implements Sengikori from @UmeboshiXI's base code.

Note the implementation is a bit strange due to how Sengikori works, it seemingly attaches a "debuff" to the skillchain, but due to how skillchains work (they delete/re add a new effect) it can't be attached to the skillchain

see https://www.bg-wiki.com/ffxi/Sengikori (use it for SC & MB) and see bonuses. I would recommend using breakpoints in SC damage and this patch for magic bursts
Also test with/without gear mods
```diff
diff --git a/scripts/globals/spells/damage_spell.lua b/scripts/globals/spells/damage_spell.lua
index 57e464994b..ab7c65d8bb 100644
--- a/scripts/globals/spells/damage_spell.lua
+++ b/scripts/globals/spells/damage_spell.lua
@@ -817,6 +817,7 @@ xi.spells.damage.calculateIfMagicBurst = function(target, spellElement, skillcha
         magicBurst = magicBurst + target:getMod(xi.mod.SENGIKORI_MB_DMG_DEBUFF) / 100
         target:setMod(xi.mod.SENGIKORI_MB_DMG_DEBUFF, 0) -- Consume the "Effect" upon magic burst.
     end
+    print(magicBurst)
 
     return magicBurst
 end
```

### DIFF
--- a/scripts/effects/sengikori.lua
+++ b/scripts/effects/sengikori.lua
@@ -5,16 +5,16 @@
 local effectObject = {}
 
 effectObject.onEffectGain = function(target, effect)
-    target:addMod(xi.mod.SKILLCHAINDMG, 2500)
-    target:addMod(xi.mod.UDMGMAGIC, 2500)
+    -- target:addMod(xi.mod.SKILLCHAINDMG, 2500)
+    -- target:addMod(xi.mod.UDMGMAGIC, 2500)
 end
 
 effectObject.onEffectTick = function(target, effect)
 end
 
 effectObject.onEffectLose = function(target, effect)
-    target:delMod(xi.mod.SKILLCHAINDMG, 2500)
-    target:delMod(xi.mod.UDMGMAGIC, 2500)
+    -- target:delMod(xi.mod.SKILLCHAINDMG, 2500)
+    -- target:delMod(xi.mod.UDMGMAGIC, 2500)
 end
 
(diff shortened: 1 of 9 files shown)


WinterSolstice8:
Hmm. github doesn't seem to be properly authoring the first commit. Will try to fix
edit: seems that email isn't linked to an account

WinterSolstice8:
There we go. Attribution fixed

Xaver-DaRed (approved):

UmeboshiXI:
Think I had my email set to private. I added a new public email. umeboshiffxi@gmail.com

Thanks for pointing out bug. Was an oversight on my part. Was having some trouble thinking how to implement this one properly when I did it.

I might commit some time to sending stuff up stream this month, been busy and have a back log. If there is ever anything you're looking for just let me know. I could also make an issue ticket for things missing that I might have but may not up to LSB standards/might be customized for our server.

WinterSolstice8:
You did a great job tbh, only that one little bit with MBs didn't quite work and that's not really your fault -- the skillchain system was NOT designed to work with Sengikori. The logic made sense just looking at it, at least. 

I'll change the attribution to that email

PR #6432, opened 2024-11-15
[LUA] Dealer Moogle: always return table for item selection

jamesbradleym:
**_I affirm:_**

- Returns a table from `getItemSelection` in `dealer_moogle.lua`
- Modifies `debugInfo` to accept items and uses the first value for `messageSpecial` which is looking for a single value, only entering this method if #items > 0
- uses the first value for `messageBasic` which is looking for a single value

finish event with dealer moogle

### DIFF
--- a/scripts/globals/dealer_moogle.lua
+++ b/scripts/globals/dealer_moogle.lua
@@ -3348,15 +3348,15 @@ local getItemSelection = function(player, list, idx, idxAlt1, idxAlt2)
         if debug.ENABLED and not debug.SHOWITEM then
             item = 0
         else
-            --- TODO: Refactor function to always return table, and find better way to determine behavior based on list type.
+            --- TODO: Find better way to determine behavior based on list type.
             ---@diagnostic disable-next-line: cast-local-type
             item = itemList[list][idxAlt1][idxAlt2]
         end
 
         if list == 12 then  -- Item, Quantity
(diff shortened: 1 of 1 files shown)


scripts/globals/dealer_moogle.lua
```diff
@@ -3348,15 +3348,15 @@ local getItemSelection = function(player, list, idx, idxAlt1, idxAlt2)
...
             ---@diagnostic disable-next-line: cast-local-type
             item = itemList[list][idxAlt1][idxAlt2]
         end
 
         if list == 12 then  -- Item, Quantity
-            --- TODO: Refactor function to always return table, and find better way to determine behavior based on list type.
+            --- TODO: Find better way to determine behavior based on list type.
             ---@diagnostic disable-next-line: cast-local-type
-            item = { item } -- Tabling here to save 100 pairs of { }
+            item = item -- Tabling here to save 100 pairs of { }
```
WinterSolstice8: Is that comment here obsolete?

jamesbradleym: Ah, you beat me - yeah I just removed it

WinterSolstice8 (approved):
seems ok to me but will defer to Clay

Xaver-DaRed:
This has yet to be tested

jamesbradleym:
@Xaver-DaRed Tested with Dealer Moogle, all seems well

PR #6440, opened 2024-11-18
[Lua] Fix fSTR calculation for physical mobskills and avatar pacts

TracentEden2:
**_I affirm:_**

This PR is the start of a series of PRs that will help improve mobskill logic and calculations. 

The PR fixes the fSTR calculation for physical mobskills and avatar bloodpacts to be retail accurate (as per Jimmayus [spreadsheet](https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=224123492#gid=224123492&range=C50)). LSB has already changed the mob and pet auto-attack fSTR calculation in this [PR](https://github.com/LandSandBoat/server/pull/3371) but the mobskill and avatar blood pact calculations were not changed at that time.

The change is implemented by modifying the fSTR functions `calculateMeleeStatFactor` and `calculateRangedStatFactor` in `physical_utilities.lua` to also work for mobs and pets (which share the same fSTR calculation). These are the new global functions added by Xaver in this [PR](https://github.com/LandSandBoat/server/pull/3624) that will eventually be used by all scripts.

Also remove the WSC TODO comment as according to Jimmayus calculations that does not exist for mobskills.

Fight mobs and notice that mobskills and avatar bloodpacts do somewhat different damage based depending on the dSTR (which is converted to fSTR by the function).

### DIFF
--- a/scripts/globals/combat/physical_utilities.lua
+++ b/scripts/globals/combat/physical_utilities.lua
@@ -107,45 +107,61 @@ local elementalBelt = -- Ordered by element.
 -- 'fSTR' in English Wikis. 'SV function' in JP wiki and Studio Gobli.
 -- BG wiki: https://www.bg-wiki.com/ffxi/FSTR
 -- Gobli Wiki: https://w-atwiki-jp.translate.goog/studiogobli/pages/14.html?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp
+-- Mob calculation: https://docs.google.com/spreadsheets/d/1YBoveP-weMdidrirY-vPDzHyxbEI2ryECINlfCnFkLI/edit?gid=224123492#gid=224123492&range=C50
 xi.combat.physical.calculateMeleeStatFactor = function(actor, target)
     local fSTR = 0 -- The variable we want to calculate.
 
     -- Calculate statDiff.
     local statDiff     = actor:getStat(xi.mod.STR) - target:getStat(xi.mod.VIT)
-    local weaponRank   = actor:getWeaponDmgRank()
-    local statLowerCap = (7 + weaponRank * 2) * -2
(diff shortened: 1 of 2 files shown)


scripts/globals/combat/physical_utilities.lua
```diff
@@ -107,45 +107,61 @@ local elementalBelt = -- Ordered by element.
...
-        fSTR = statDiff + 12
+
+    -- Pets and mobs have own calculation
+    if actor:isMob() or actor:isPet() then
+        local fSTRupperCap = 24
+        local fSTRlowerCap = -20
+        fSTR = math.floor((statDiff + 4) / 4)
+        fSTR = utils.clamp(fSTR, fSTRlowerCap, fSTRupperCap)
+        -- level 1 mobs always have fSTR of 1
+        if actor:isMob() and actor:getMainLvl() == 1 then
```
WinterSolstice8: I think this should be `<= 1`, at some point we may actually support level -1 in some meaningful way, but level 0 is technically "valid" too. I remember helping to confirm this fSTR of 1 though, that was a trip

Xaver-DaRed: Now would be a good time to ask what calculation trusts use

TracentEden2: > I think this should be `<= 1`, at some point we may actually support level -1 in some meaningful way, but level 0 is technically "valid" too. I remember helping to confirm this fSTR of 1 though, that was a trip

Yep, thanks for info, I have updated the PR!

WinterSolstice8 (approved):

scripts/globals/combat/physical_utilities.lua
TracentEden2: > Now would be a good time to ask what calculation trusts use

Yeah, I asked jimmayus and he said that from his knowledge they use the player calculation.

PR #6445, opened 2024-11-19
Core: Cache augment data on startup

### LINKED ISSUE #6441: 🔨 Cache Augments at startup
**_I affirm:_**

## Describe the feature

Augments should be cached on startup, see this item_equipment.cpp function that hits the DB every time an item is loaded with an augment:

```c++
void CItemEquipment::SetAugmentMod(uint16 type, uint8 value)
{
    if (type != 0)
    {
        setSubType(ITEM_AUGMENTED);
        ref<uint8>(m_extra, 0x00) |= 0x02;
        ref<uint8>(m_extra, 0x01) |= 0x03;
    }

    // obtain augment info by querying the db
    const char* fmtQuery = "SELECT augmentId, multiplier, modId, `value`, `isPet`, `petType` FROM augments WHERE augmentId = %u";

    int32 ret = _sql->Query(fmtQuery, type);

    while (ret != SQL_ERROR && _sql->NumRows() != 0 && _sql->NextRow() == SQL_SUCCESS)
    {
...
```

zach2good:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/6441

We absolutely shouldn't be querying for static data at runtime... ever... for any reason... lol

Also while looking at this, we seem to store the data in the db sanely, but we're doing some sort of sillybuggers about negative values once we load it? Will need some investigation. Did this quickly on my lunch break, need to fire this up and test on my Windows machine before merging.

### DIFF
--- a/src/map/items/item_equipment.cpp
+++ b/src/map/items/item_equipment.cpp
@@ -24,6 +24,52 @@
 #include "map.h"
 #include <cstring>
 
+namespace
+{
+    /*
+    From augments.sql:
+
+    `augmentId` smallint(5) unsigned NOT NULL,
+    `multiplier` smallint(2) NOT NULL DEFAULT 0,
+    `modId` smallint(5) unsigned NOT NULL DEFAULT 0,
+    `value` smallint(5) NOT NULL DEFAULT 0,
+    `isPet` tinyint(1) NOT NULL DEFAULT 0,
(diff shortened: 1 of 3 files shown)


WinterSolstice8:
I cherry picked this in, seems to work fine
image

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #6449, opened 2024-11-21
Core: Replace entity ptr for WideScanTarget with EntityID_t

zach2good:
**_I affirm:_**

Replaces: https://github.com/LandSandBoat/server/pull/6392

The previous PR became a staging ground for me testing entity lookup. This is the minimum amount of work for this ptr replacement.


claywar (approved):

zach2good:
Tested, working with setting a widescan target, seeing them on the minimap, zoning and then your set target isn't set anymore, etc.

PR #6450, opened 2024-11-21
Chore: Update clang-format rules (clang-format-19)

zach2good:
**_I affirm:_**

Of note/TODO:
- Death to all `clang-format on/off` blocks
- We now live in a west const world
- I'm still fighting with macros
- I'm still fighting with things like settings::overloaded
- I'm still fighting with conditional reflowing, etc.
- I'm still fighting with type and assignment alignment
- I decided to keep namespace nesting
- Spaces around simple {} etc.
- Make it easy to setup the tooling for Windows/Linux/OSX, have it work easily/automatically in VSCode
- Make it hard to use the wrong versions of clang-format, etc.
- Update CI

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,33 +1,275 @@
-﻿---
+﻿
 # https://clang.llvm.org/docs/ClangFormatStyleOptions.html
-# We base things of WebKit + Allman braces
-# BasedOnStyle: WebKit
-AccessModifierOffset: -4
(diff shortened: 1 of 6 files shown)


zach2good:
https://github.com/LandSandBoat/server/discussions/6423

zach2good:
This is replaced by: https://github.com/LandSandBoat/server/pull/8623

PR #6454, opened 2024-11-23
Chakra, Dodge, Focus retail functionality audit

WinterSolstice8:
**_I affirm:_**

Does what it says on the tin, updates Chakra, Dodge and Focus to match retail stuff in the following manners

- basic functionality
- recasts
- duration (where applicable)
- merit timer reductions

Use Focus/Dodge/Chakra and have them approximately match the following wiki pages:
https://www.bg-wiki.com/ffxi/Focus
https://www.bg-wiki.com/ffxi/Dodge
https://www.bg-wiki.com/ffxi/Chakra


scripts/globals/job_utils/monk.lua
```diff
@@ -51,9 +52,20 @@ xi.job_utils.monk.useChakra = function(player, target, ability)
         end
     end
 
+    local monkLevel = 0
+
+    if target:getMainJob() == xi.job.MNK then
+        monkLevel = target:getMainLvl()
+    else
+        monkLevel = target:getSubLvl()
+    end
```
Xaver-DaRed: We are using this logic a lot, I think we also use it in other places.
Maybe we should just turn it to an util in the globals/utils.lua
Like...

```lua
function utils.getCurrentLevel(actor, job)
    local level = 0
    
    if actor:getMainJob() == job then
        level = actor:getMainLvl()
    elseif actor:getSubJob() == job then
        level = actor:getSubLvl()
    end

    return level
end
```

Probably better for a different PR, but putting it there

Xaver-DaRed (approved):

zach2good (approved):

scripts/globals/job_utils/monk.lua
WinterSolstice8: Done

WinterSolstice8: Oh, while I was looking at my diff I noticed that function already exists as `utils.GetActiveJobLevel`

zach2good (approved):

PR #6467, opened 2024-11-27
Core: Add ENV64BIT definition for OSX

### LINKED ISSUE #6235: 🐛 ENV64BIT is not set on 64-bit OSX build
**_I affirm:_**

OSX (possibly all clang builds?)

## Branch affected by issue 

`base`

## Steps to reproduce

```cpp
// Check GCC
#if __GNUC__
#if __x86_64__ || __ppc64__
#define ENV64BIT
#else
#define ENV32BIT
#endif 
```

This flag seems like it won't get set on any clang build? Therefore all clang builds are forced to be 32-bit?

zach2good:
**_I affirm:_**

Closes: https://github.com/LandSandBoat/server/issues/6235


zach2good:
Cmake definitions are broken by the cache for some reason, hmm

zach2good:
Scaled this back to just the ENV definitions while I figure out the cache definition stuff

WinterSolstice8 (approved):

PR #6468, opened 2024-11-27
HPP/MPP adjustments, use divisors for healing breath trigger points

### LINKED ISSUE #5093: 🐛 HPP and MPP are calculated incorrectly
**_I affirm:_**

Windows10

## Branch affected by issue

`base`

## Steps to reproduce

1. Calculate your max yellow HP threshold, which is the largest HP that is less than 75% of your max HP. (For example, with 1000 max HP, you should be yellow at 749 HP or less).
2. Set your HP to this max yellow HP threshold (749 HP, in the previous example).
3. Observe your HP is white. (`/echo <hpp>` will output 75%).
4. Keep reducing your HP until it finally turns yellow (this will be at 74% HP or less, rather than <75% HP).

Alternatively:

1. Set your current HP to your max HP.
2. Do `/echo <hpp>` (Observe 100%)
3. Set your HP to 1 less than max HP.
4. Do `/echo <hpp>` (Observe 100%)

## Expected behavior

3. Your HP turns yellow. `/echo <hpp>` will output 74%.

Alternatively:

4. `/echo <hpp>` outputs 99%.

## Analysis

The problem is in [`CBattleEntity::GetHPP()`](https://github.com/LandSandBoat/server/blob/base/src/map/entities/battleentity.cpp#L193) and [`CBattleEntity::GetMPP()`](https://github.com/LandSandBoat/server/blob/base/src/map/entities/battleentity.cpp#L209). These functions directly feed into the `CCharHealthPacket`/`0x0DF` packet which informs the client of their HPP (and hence red/orange/yellow/white HP and the output of `/echo <hpp>`).

```C++
uint8 CBattleEntity::GetHPP() const
{
    return (uint8)ceil(((float)health.hp / (float)GetMaxHP()) * 100);
}
```

It is doing `ceil()` whereas it should be doing `floor()`. To account for retail only yielding 0 HPP when your HP is 0 (that is, you're dead), it should look something like this:

```C++
uint8 CBattleEntity::GetHPP() const
{
    if (health.hp == 0)
    {
        return 0;
    }
    
    return std::max<uint8>(1, (uint8)floor((float)health.hp / (float)GetMaxHP()) * 100);
}
```

There could potentially be a variety of knock-on effects from other systems having previously compensated for this inaccuracy, so it probably needs a bit of discussion before fixing.

Possibly affected systems:

* Undead aggro threshold
* HPP latent effects
* Sublimation deactivation
* Mob TP behaviors
* Breath damage
* Spirits Within damage
* Wyvern healing breath threshold (should be 1/2, 1/3, 1/4 rather than HPP from my testing, but that's a separate issue).
* ???

WinterSolstice8:
**_I affirm:_**

See #5093 
Closes #5093  (mostly)

Noteworthy: HPP/MPP values adjusted down by 1% and a lot of code may have anticipated the error. This PR does not intend to fix everything, as there's too many known unknowns. I did a brief examination of mobskills and some have unverified hp% trigger values so I left them alone. Same with all mob behavior scripts. That said, I would assume most behavior changes due to this will not be significant.

See #5093 for details -- but in essence just see HPP/MPP act like retail does it. Test healing breath with hybrid sub to use floor(MaxHP/3) as trigger point

### DIFF
--- a/scripts/actions/mobskills/calamitous_wind.lua
+++ b/scripts/actions/mobskills/calamitous_wind.lua
@@ -12,7 +12,7 @@
 local mobskillObject = {}
 
 mobskillObject.onMobSkillCheck = function(target, mob, skill)
-    if mob:getHPP() > 50 then
+    if mob:getHPP() >= 50 then
         return 1
     else
         return 0
--- a/scripts/actions/mobskills/colossal_blow.lua
+++ b/scripts/actions/mobskills/colossal_blow.lua
@@ -13,7 +13,7 @@ mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local currentHP = target:getHP()
     local damage    = currentHP
 
-    -- if have more hp then 30%, then reduce to 5%
+    -- if we have more than 30% hp, reduce to 5%
     if target:getHPP() > 30 then
         damage = currentHP * .95
     end
--- a/scripts/actions/mobskills/feral_peck.lua
+++ b/scripts/actions/mobskills/feral_peck.lua
@@ -17,7 +17,7 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local damage    = target:getHP()
 
-    -- If have more hp then 30%, then reduce a 10%
+    -- If we have more than 30% HP, then reduce by 10%
     if target:getHPP() > 30 then
         damage = damage * 0.9
     end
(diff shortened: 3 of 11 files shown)


Xaver-DaRed (approved):

DiscipleOfEris:
I think if you have 1 HP this will report 0 HPP? It should have a `std::max(1, std::floor(HP / 100.f))` I think.

WinterSolstice8:
> I think if you have 1 HP this will report 0 HPP? It should have a `std::max(1, std::floor(HP / 100.f))` I think.

Whoops, was too fixated on the flooring and healing breath, #6472

PR #6488, opened 2024-12-04
[core] CodeQL warning cleanup for zone_entities.cpp

WinterSolstice8:
**_I affirm:_**

Stop reuse of `it` within appropriate scope causing messy code (and possibly undefined behavior?)
Add nullptr check to a raw pointer

Load up server, see nothing obviously different


claywar (approved):

zach2good (approved):
Good for now, we'll eventually want to replace all of these with range-for loops and collect entities to remove, then remove them later rather than in line

PR #6489, opened 2024-12-05
[Core/SQL] Adds Support for Multiple Mob Pets

0x05010705:
**_I affirm:_**

Adds Support for Mobs with Multiple Pets through mob_pets and core
- This allows for multi-pet AI to be handled within core instead of within each lua file where multiple pet behavior was not always properly handled and removes redundant loops from lua files.
- Updates `mob_pets.sql` to use a composite key of the Master Mob ID with the Pet Offset and uses an index on the Master Mob ID instead of just a Primary Key on the Master Mob ID.
- Created linked lists within the pets entities.  The pets have links with each other (PPetNext), but always contain a link back to the parent (PMaster).
  - The new pointer is always initialized to a nullptr.
  - Loops are used within the mob_controller that iterates over these linked pets until it reaches this nullptr to control all the associated pets at the same time.

Some of the mobs that summon multiple pets:
  - Jailer of Love
  - Ix'aern DRG
  - Absolute Virtue
  - Pandemonium Warden
  - Odin V2
  - Xzomits
  - Some Escha NMs
  - Some VW NMs
  - etc...

For sake of simplicity, and to show a basic implementation of using multiple pets - I have updated Ix'Aern DRG with some basic logic to summon all 3 of its wynavs with 'Call Wyvern' and, since it uses the ability multiple times, how tracking can be done to re-trigger the ability and show that they are all resummoned and follow the proper pet logic.

NOTE: This is only a slight modification to replace the looping structure that was in place for Ix'aern DRG from before.  This is not an accurate re-implementation of the encounter.  Someone else is currently working on that with retail capture data and will PR that soon.  This is strictly to provide an example usage of this multi-pet support PR.

1 - `!zone The Garden of Ru'Hmet`
2 - `!spawnmob 16921022`
3 - `!gotoid 16921022`
4 - Enage Ix'Aern DRG
Note that all 3 pets are summoned through the Call Wyvern JA

5 - Kite the mob away from its spawn point and then `!goto <me>` to drop aggro
Note that the pets now follow their master back

6 - Attempt to use the SMN Pull Technique to only aggro Ix'Aern DRG
  - `!changejob smn 75`
  - Summon Carbuncle and then Assault on Ix'Aern DRG
  - After the mobs are engaged to the pet, Retreat Carbuncle
  - After carbuncle is near you, Release carbuncle
Note that the pets now detect that the master is still engaged and continue to fight the player.
_Previously, this would allow players to engage Ix'Aern DRG while the pets remained in an idle/roaming state._

7 - Use `!hp 0` on 1~2 of the pets
8 - Wait till Ix'Aern DRG attempts to re-use Call Wyvern
Note that the missing pets are re-summoned and engage with whatever the masters current target is.

### DIFF
--- a/scripts/zones/The_Garden_of_RuHmet/mobs/Ixaern_DRG.lua
+++ b/scripts/zones/The_Garden_of_RuHmet/mobs/Ixaern_DRG.lua
@@ -7,24 +7,22 @@ local ID = zones[xi.zone.THE_GARDEN_OF_RUHMET]
 ---@type TMobEntity
 local entity = {}
 
-entity.onMobFight = function(mob, target)
-    -- Spawn the pets if they are despawned
-    -- TODO: summon animations?
-    local mobId = mob:getID()
-    local x = mob:getXPos()
-    local y = mob:getYPos()
-    local z = mob:getZPos()
+entity.onMobSpawn = function(mob)
+    mob:setLocalVar('petCount', 0)
+    mob:setLocalVar('petTimer', 0)
+end
 
-    for i = mobId + 1, mobId + 3 do
-        local wynav = GetMobByID(i)
-        if wynav and not wynav:isSpawned() then
-            local repopWynavs = wynav:getLocalVar('repop') -- see Wynav script
-            if mob:getBattleTime() - repopWynavs > 10 then
-                wynav:setSpawn(x + math.random(1, 5), y, z + math.random(1, 5))
-                wynav:spawn()
-                wynav:updateEnmity(target)
-            end
-        end
+entity.onMobFight = function(mob, target)
+    local petCount = mob:getLocalVar('petCount')
+    local petTimer = mob:getLocalVar('petTimer')
+
+    if
+        petCount < 3 and
+        petTimer < os.time() + 30
+    then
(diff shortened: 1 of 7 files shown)


zach2good:
Skimmed for 5 seconds, saw a manually managed linked list. If we want to support multiple pets, look at how we manage trusts: a vector. We don't want to introduce more raw pointers to disappearing entities. That's how you get crashes.

0x05010705:
Noted - Will look into refactoring this using a vector.
Am I correct in assuming you are referencing the `std::vector<Trust_t*> g_PTrustList` ?

zach2good:
The global trust vector is information for spawning trusts. I'm talking
about the vector of trusts that exists on char entities for storing their
active trust data, where we check to see if it's empty a lot

0x05010705:
Just a follow up on this:  Swapped to a vector, but in doing so, ran across some other issues with pets now despawning as soon as they go inactive.  This did not happen while using linked lists, so I am guessing I need to play "chase the pointer" for a bit.  Working through this and figuring out how best to resolve the issue, but outside of that, everything else seems to be working.

PR #6498, opened 2024-12-08
Lua, SQL: Implement Zipacna pathing and interaction with Cermet Doors

hooksta4:
**_I affirm:_**

Currently, Zipacna only spawns in one place and paths through the doors regardless if they are open or closed. 
This PR addresses the following: (I made no modifications to Zipacna's abilities or stats)
- adds the 4 spawn points located as shown here. 
image
image
image
https://www.youtube.com/watch?v=nWjGMWHsg9k

The doors are labelled as such (from west side to east side)
Basement Loop - |1st Blue Door | Middle Blue Door | Middle Yellow Door | 1st Yellow Door| - Basement loop

- There are 4 doors controlled by a Yellow or Blue Monolith. When one is activated the other "side doors" are closed. 
** *Note - It is possible that Zipacna can get stuck between the middle yellow and first yellow door as well as the middle blue door and first blue door. If someone switches while he is pathing between**

Zipacna's behavior is to path to a door. If the door is closed, it will turn around and go to the next door in the sequence.
Example: Middle Yellow door was closed, Zipacana will turn around and go to the middle blue door. If the middle blue door is open, Zipacna will proceed to the 1st Blue door, make a loop in the basement and then return to the first blue door. If however, the yellow switch is then triggered, the middle blue door is now closed and once Zipacna arrives will again turn around and go to the middle yellow door. 

Due to there needing to be checks on both sides of the door, vars were used to track the last door Zip passed through. It is hard to follow but the video shows an hour and a half of Zipacna pathing with all the scenarios I could come up with that could happen. https://youtu.be/rWGuV6pqJww

TODO: I believe it is more of a navmesh thing but if he is engaged and the players wipe/zone Zipacna will likely path through walls back to its spawn point, at which point, will begin his pathing again. So suck less?

!spawnmob 17502570 (Will spawn in 1 of 4 locations.
!gotoid 17502570 - follow Zipacna and you will observe it path to each sides respective first doors. At that point, the pathing between doors takes over.

I understand that there is most likely not a way to clean up the path points, but if there is a way to clean up the door logic, I am open to suggestions.


scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,557 @@
...
+    if
+        pos.x >= 141 and
+        pos.y >= 16 and
+        pos.z >= 440
+    then
+        --print("Zip spawned at Yellow basement Spawn Point- going to First Yellow door", pos) useful for debugging if zip gets lost somehow.
+        mob:pathThrough(yellowBasementSpawnPath, xi.path.flag.COORDS)
+    end
+
+end
```
Xaver-DaRed, line 337: Shouldnt you be using `elseif` for this ones instead of `if`s?

Also, there are only 4 options where it can spawn, all of them with severely diferent X positions. Can you check if this keeps working if you simplify the checks to only check for the X position?

Could also edit the X check to be more generous in case it gets slightly off.

Also, you have to allign comments properly, and remove the prints.

Xaver-DaRed: End result would look like this
```lua
    -- Zip popped on blue short side (!pos -196, 0, 396)
    if pos.x <= -195 then
        mob:pathThrough(blueShortSideSpawnPath, xi.path.flag.COORDS)

    -- Zip popped at the blue basement spawn point (!pos -121, 16, 420)
    elseif pos.x <= -120 then
        mob:pathThrough(blueBasementSpawnPath, xi.path.flag.COORDS)

    -- Zip popped at the Yellow Short Side Spawn Point - going to first yellow door (!pos 191, 0, 399)
    elseif pos.x >= 190 then
        mob:pathThrough(yellowShortSideSpawnPath, xi.path.flag.COORDS)

    -- Zip popped in the yellow side basement - going to first yellow door
    elseif pos.x >= 140 then
        mob:pathThrough(yellowBasementSpawnPath, xi.path.flag.COORDS)
    end
```

hooksta4: Ill give it a go - I think we can get away without the Z check but the Y will be needed I would imagine.

hooksta4: Looks like I was wrong - seems to be just fine with the x -

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,523 @@
...
+    -- begin of door checks. System in place is to check distance betwen mob and the NPC door (mapped to IDs.lua in local firstBlueDoorCoord, etc)
+    if
+        mob:checkDistance(firstBlueDoorCoord.x, firstBlueDoorCoord.y, firstBlueDoorCoord.z) <= 5
+    then
+        if -- POS from path to first blue door if Zip is coming from the basement
+            pos.x <= -162.000 and
+            pos.y <= 1 and
+            pos.z >= 379 and
+            getLastDoor ~= 1
+        then
```
zach2good, line 362: I'm struggling to understand what you're checking here?

You check the distance to the door (you can be using `mob:checkDistance(firstBlueDoor) <= 5`), and then you're checking some other point? Can you extract that to something named? Is it a part of one of the paths?

Then you're choosing which path to do depending on whether or not the door is closed. Fine, but the "if path is closed" is one other other other paths in reverse, no? I asked you in ASB or HXI discord to look at the `xi.path.flag.REVERSE` flat that `Guivre` uses so you don't have to have so much path data.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,523 @@
...
+                firstBlueDoor and
+                firstBlueDoor:getAnimation() == xi.anim.OPEN_DOOR
+            then
+                mob:pathThrough(firstBlueToBasement, xi.path.flag.COORDS)
+            end
+        end
+    end
+
+    -- middle blue door
+    if
```
zach2good, line 396: You repeat this exact pattern 4 times using different points and paths, extract it out into a function or a table, etc.

sql/nm_spawn_points.sql
```diff
@@ -5660,6 +5660,10 @@ INSERT INTO `nm_spawn_points` VALUES (17498522,46,-150.407,48.460,-333.947);
 INSERT INTO `nm_spawn_points` VALUES (17498522,47,-138.264,48.466,-341.724);
 INSERT INTO `nm_spawn_points` VALUES (17498522,48,-130.664,48.315,-338.463);
 INSERT INTO `nm_spawn_points` VALUES (17498522,49,-164.707,48.330,-330.263);
+INSERT INTO `nm_spawn_points` VALUES (17502570,0,-196.187,0.51,396.813); -- Zipacna
+INSERT INTO `nm_spawn_points` VALUES (17502570,1,-121.073,16.575,420.915); -- Zipacna
+INSERT INTO `nm_spawn_points` VALUES (17502570,2,191.685, 0.289, 399.782); -- Zipacna
+INSERT INTO `nm_spawn_points` VALUES (17502570,3,141.353,16.575,440.782); -- Zipacna
```
zach2good, line 5666: I don't know how this system works, will it just pick randomly from the points it has listed?

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,523 @@
...
+    { x = -157, y = 0, z = 379 },-- Middle blue door
+
+}
 entity.onMobSpawn = function(mob)
+    local pos = mob:getPos()
     mob:setMod(xi.mod.REGAIN, 200)
-    mob:pathThrough(pathNodes, xi.path.flag.PATROL)
+
+    -- Zip popped on blue short side (!pos -196, 0, 396)
+    if pos.x <= -195 then
```
zach2good, line 322: It would be more obvious if you specify the spawn points as x/y/z tables, with a comment saying this is what's listed in the nm spawn points SQL, and then do `checkDistance()` on each of them

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,523 @@
...
+    { x = -60, y = 0, z = 433 },
+    { x = -60, y = 0, z = 450 },
+    { x = -62, y = 0, z = 455 },
+    { x = -67, y = 0, z = 458 },
+    { x = -75, y = 0, z = 460 },
+    { x = -76, y = 0, z = 460 },
+}
+
+    -- Path from the middle blue door to the first blue door
+local middleBlueToFirstBlue =
```
zach2good, line 299: The naming of these is inconistant, you're mixing up "position" and "colour" in the naming of these. This is also presumably just the reverse of `firstBluetoBlueMiddle`. See my comment somewhere below about using the reverse flag.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,523 @@
...
+    { x = 184, y = 0, z = 380 },
+    { x = 176, y = 0, z = 379 },
+    { x = 157, y = 0, z = 380 },
+}
+
+    -- From first blue door to the blue basement loop
+local firstBlueToBasement =
+{
+    { x = -162, y = 0, z = 380 }, --First Blue Door
+    { x = -171, y = 0, z = 379 },
```
zach2good, line 97: Comment formatting:
```lua
-- From the first blue door to the blue basement loop
local firstBlueToBasement =
{
    { x = -162, y = 0, z = 380 }, -- First Blue Door
    { x = -171, y = 0, z = 379 },
```

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,523 @@
 -----------------------------------
 ---@type TMobEntity
 local entity = {}
+local ID = zones[xi.zone.VELUGANNON_PALACE]
 
-local pathNodes =
+local firstBlueDoorCoord = { x = -160, y = -2.974, z = 380 }
+local middleBlueDoorCoord = { x = -80, y = -2.974, z = 460 }
+local middleYellowDoorCoord = { x = 80, y = -2.974, z = 460 }
+local firstYellowDoorCoord = { x = 160, y = -2.974, z = 380 }
```
zach2good, line 12: Do you need these, if you're looking up the door entities themselves? You can just use `getPos()` on them, and `checkDistance()` for comparisons

zach2good (changes requested):

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,523 @@
...
+                middleYellowDoor and
+                middleYellowDoor:getAnimation() == xi.anim.CLOSE_DOOR
+            then
+                mob:pathThrough(middleYellowToFirstYellow, xi.path.flag.COORDS)
+            elseif
+                middleYellowDoor and
+                middleYellowDoor:getAnimation() == xi.anim.OPEN_DOOR
+            then
+                mob:pathThrough(middleYellowToMiddleBlue, xi.path.flag.COORDS)
+                mob:setLocalVar('[Zipacna]lastDoor', 3)
```
zach2good, line 473: Instead of setting these, would it work to use `not entity:isFollowingPath()` to make sure you aren't constantly resetting the path with `pathThrough()`?

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
hooksta4: If I am understanding your question correctly, there needs to be a check on both sides of the door. The reverse paths can work for the path between middle doors to the first doors but will not for the basement routes. I did not include them as I thought it would be confusing. Regardless, I feel all the paths would still need to be there due to coming from one side or the other. 

Maybe my approach is not the best way for this - thinking on it, I might be able to first check the status of the door (open/closed) and then determine which side of the door Zipacna is on. And from there, path accordingly.

sql/nm_spawn_points.sql
hooksta4: That is my understanding, yes. Prior to having `INSERT INTO `nm_spawn_points` VALUES (17502570,0,-196.187,0.51,396.813); -- Zipacna` and
`UpdateNMSpawnPoint(mob:getID())` - it would spawn at the same point that was listed in mob_spawn_points.sql - now it randomly picks between the 4 locations.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
hooksta4: Indeed, the reverse will not work for the basement loops as Zip goes in the same direction regardless. The reverse flag would have it go counter to that. For the sake of uniformity, I opted to not include those for the middle doors as I thought it might be confusing.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,523 @@
...
+    -- middle yellow door
+    if
+    mob:checkDistance(middleYellowDoorCoord.x, middleYellowDoorCoord.y, middleYellowDoorCoord.z) <= 5 then
+        if
+            pos.x <= 78 and
+            pos.y <= 1 and
+            pos.z <= 461 and
+            getLastDoor <= 3
+        then
+            print ("12check distance to middle yellow door is a success! from blue middle door")
```
Frankie-hz, line 444: Remove prints

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
zach2good: Do you even need to know which side of the door you're on? You already know which path you're using so if you're within whatever range of the door and it's closed, you know to reverse the current path. The side of approach is irrelevant because you already know which path you're using. If you don't, you can put your paths in a table and keep track of the index as a local var

zach2good:
> TODO: I believe it is more of a navmesh thing but if he is engaged and the players wipe/zone Zipacna will likely path through walls back to its spawn point, at which point, will begin his pathing again. So suck less?

This isn't a navmesh issue. This is a pathing issue. This would be a very good opportunity for you to write a helper function that takes a current patrol path, finds the nearest point to a given entity on that path, and then uses that slice of the patrol path to call to `pathThrough`. Essentially a "resume this path from the closest point" - rather than returning to spawn or the very start of the path

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
Removed all of these `pos.x` checks in lieu of `checkDistance` calls using a table for the spawn points now listed at top of file.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
Extracted these repeated calls into a helper function that determines the path finding logic.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
Completely changed the door logic and used a local var (scoped to the lua) to determine the direction Zip is going.
Also updated to use `checkDistance` calls and `xi.path.flag.REVERSE` and shortened his pathing node list.
The pathing node list has also been converted to a table.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
This has been extracted into a table at the start of the file.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
With the removal of excess local vars and using an indexed table to determine pathing, this should resolve most of this I believe.  I am still not a large fan of the naming convention myself, but I kept things as close to his as I could.

If he needs to do extra modifications on names, I will leave that to him for now.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
These should all be fixed now I believe.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
Swapped these to `GetNPCByID` and then using `getPos()` to determine their locations instead of hard-coding them

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
Since these have been extracted to a helper function, and `pathThrough()` is only getting called when in the presence of a door now, I am not sure that we need to use `isFollowingPath()` as the paths now end at each door where the check is performed to find a new path.

Might need to check the usage in the updated ammended commit to see if there is still a need to swap in some `isFollowingPath()` logic.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Helping Hook out with this - So will comment on some of this on his behalf.
These have been removed.

0x05010705:
Well, went to amend his commit, but I don't have Push Access to his fork - So I will send the updated changes to @hooksta4 and have him amend this PR from his end.

0x05010705:
This PR can be closed - Was not able to get the changes amended to this fork/branch.
This was re-opened as a new PR with the requested changes that I can continue to make adjustments on if needed.
New PR: https://github.com/LandSandBoat/server/pull/6516

WinterSolstice8:
as directed, superceded by #6516

PR #6516, opened 2024-12-12
[Lua/SQL] Zipacna Fixes for Pathing and Door Interactions

0x05010705:
**_I affirm:_**

Currently, Zipacna only spawns in one place and paths through the doors regardless if they are open or closed.
This PR addresses the following: (I made no modifications to Zipacna's abilities or stats)
- adds the 4 spawn points located as shown here.
393557417-d9385703-103a-4d8a-95d6-3f5bc5cd3a24
393557420-833a10fd-b3bd-4b09-8b42-36c9a87db275
393557428-58f35a18-16f1-4345-9ae7-99b6331d1983

    https://www.youtube.com/watch?v=nWjGMWHsg9k

The doors are labelled as such (from west side to east side)
Basement Loop - |1st Blue Door | Middle Blue Door | Middle Yellow Door | 1st Yellow Door| - Basement loop

- There are 4 doors controlled by a Yellow or Blue Monolith. When one is activated the other "side doors" are closed.
** Note - It is possible that Zipacna can get stuck between the middle yellow and first yellow door as well as the middle blue door and first blue door. If someone switches while he is pathing between **

Zipacna's behavior is to path to a door. If the door is closed, it will turn around and go to the next door in the sequence.
Example: Middle Yellow door was closed, Zipacana will turn around and go to the middle blue door. If the middle blue door is open, Zipacna will proceed to the 1st Blue door, make a loop in the basement and then return to the first blue door. If however, the yellow switch is triggered and the middle blue door is now closed. once Zipacna arrives at the door again, he will turn around and go to the middle yellow door.

### **NOTE: This addresses all the issues that were brought up in the original PR opened by @hooksta4 (https://github.com/LandSandBoat/server/pull/6498)**
- Removed Hard-Coded Position Checks (x/y/z)
- Updated to use `GetNPCByID()` and `getPos()` to get positions on doors
- Updated to use `checkDistance()` when needed
- Updated to use `xi.path.flag.REVERSE` where applicable to remove some of the extra path nodes
- Converted that path nodes into a table instead of several local variables
- Created a file-based scope variable `currentDirection` to track the direction of Zipacna (vs using a Local Var on the mob)
- Extracted the 4x replicated door logic into a helper function `setZipPath()` to handle pathing logic
- Fixed all comment formatting errors
- Removed un-needed local vars on Zipacna
- Removed excess prints that were left in the lua file

1 - `!spawnmob 17502570`
2 - `!gotoid 17502570`
3 - Follow Zipacna through the zones and up to blue/yellow doors
Note that he now paths to the doors, and if closed, turns around and paths the opposite direction.
4 - Trigger the doors to open and close so that he can path to the different segments
Note that even if he is 'trapped' between two doors, he will just path back and forth between the two doors until he finds an open door to continue.


Xaver-DaRed (approved):
Just one thing:
```lua
local tableName =
{
    [index] =
    {
        { data }
    }
}
```
Thats the style we use for tables

0x05010705:
Noted.  Updated and Amended.

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,349 @@
...
-    { x = -202, z = 391 },
+    blueShort   = {x = -196, y = 0,  z = 389},
+    blueLong    = {x = -121, y = 16, z = 420},
+    yellowShort = {x =  191, y = 0,  z = 399},
+    yellowLong  = {x =  141, y = 16, z = 440},
 }
 
-entity.onMobSpawn = function(mob)
+-- Direction 0 = West -> East; 1 = East -> West
+local currentDirection = 0
```
zach2good: This can be named, like an enum;
```lua
local currentDirection =
{
    WEST_EAST = 0,
    EAST_WEST = 1, 
}
```

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,349 @@
...
+        { x = 164,  y = 0,   z = 380 },
+    },
+}
+
+local setZipPath = function(mob, door, label)
+    if door then
+        if door:getAnimation() == xi.anim.OPEN_DOOR then
+            if currentDirection == 0 then -- West to East
+                mob:pathThrough(paths[label + 1], xi.path.flag.COORDS)
+                if label + 1 == 5 then
```
zach2good: In this context, `label` is hard to reason about. This could also be made into an enum

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,349 @@
...
 entity.onMobDeath = function(mob, player, optParams)
 end
 
+entity.onMobRoam = function(mob)
+    local firstBlueDoor    = GetNPCByID(ID.npc.B_DOOR_OFFSET + 1) -- North-West Blue Door _4x9
+    local middleBlueDoor   = GetNPCByID(ID.npc.B_DOOR_OFFSET)     -- North-Central Blue Door _4x8
+    local firstYellowDoor  = GetNPCByID(ID.npc.Y_DOOR_OFFSET + 6) -- North-East Yellow Door _4x6
+    local middleYellowDoor = GetNPCByID(ID.npc.Y_DOOR_OFFSET + 7) -- North-Central Yellow Door _4x7
+
+    -- check that we got all door objects (and for CI sake)
```
zach2good, line 338: 👍

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,349 @@
...
+        { x = 191,  y = 0,   z = 405 },
+        { x = 192,  y = 0,   z = 395 },
+        { x = 193,  y = 0,   z = 389 },
+        { x = 187,  y = 0,   z = 382 },
+        { x = 184,  y = 0,   z = 380 },
+        { x = 164,  y = 0,   z = 380 },
+    },
+}
+
+local setZipPath = function(mob, door, label)
```
zach2good: This is exactly the sort of data/driver separation I was after, thanks 👍

zach2good (approved):
Very tidy 👍

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: If I change this to an enum, value assignment can longer be done on `currentDirection` directly without adding a new member.  Would something like this be okay to keep these values associated with each other?
```lua
local direction = 
{
    WEST_EAST = 0,
    EAST_WEST = 1,
    CURRENT   = 0,
}
```
Where `CURRENT` will track the numerical data, but `direction.WEST_EAST` and `direction.EAST_WEST` can be used for value assignment?

Xaver-DaRed: I would go with defining
```lua
local direction = 
{
    WEST_EAST = 0,
    EAST_WEST = 1, 
}
```
And then just using the `currentDirection` variable to store it.

Another option to store that info is using an entity local variable (getLocalVar, setLocalVar)

But in any case, for now, I would just define the local enum and replacing the 0s and 1s with `direction.WEST_EAST` and `direction.EAST_WEST` and thats it

Xaver-DaRed: Having said that, 0 and 1 are no better to read than EAST_WEST / WEST_EAST, IMO

Maybe `TO_WEST` and `TO_EAST` would be easier/better to read at a glance

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: I will change `label` to `currPath` to be more descriptive about what it is
I also added an enum `paths` that replaces the checks for 1/2/3/4/5 values
_Note: I changed the table of path nodes from `paths` to `pathNodes`_

0x05010705:
- Added `direction` to enum 0/1
- Added member `CURRENT` to `direction` enum/table for direct value modifications
- Updated `paths` to `pathNodes` to be more descriptive about its contents
- Added `paths` to enum 0/1/2/3/4/5 to be more descriptive about its contents
- Updated hard-coded values to use new enum values
- Updates `label` to be `currPath` to be less ambiguous

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: RIP - I did not refresh this page before amending again.
I ended up just adding `CURRENT` into the table/enum for now.
Being able to assign `direction.CURRENT = direction.EAST_WEST` and `direction.WEST_EAST` are definitely better than just 0/1.

I also opted to try and keep `CURRENT` in the `direction` table to keep that information centralized and not spread across multiple isolated variables.

0x05010705: I can definitely change the values to `TO_EAST` and `TO_WEST` real quick though.

0x05010705:
- Updated `WEST_TO_EAST` to `TO_EAST`
- Updated `EAST_TO_WEST` to `TO WEST`

0x05010705:
- Removed excess/outdated comments

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,355 @@
...
+    yellowShort = {x =  191, y = 0,  z = 399},
+    yellowLong  = {x =  141, y = 16, z = 440},
+}
+
+local direction =
+{
+    TO_EAST = 0,
+    TO_WEST = 1,
+    CURRENT = 0,
+}
```
zach2good: Your usage here is logically unsafe. You're using TO_EAST/TO_WEST as enums for comparison, but later on you're using and changing CURRENT. Even though Lua doesn't have a dedicated const/non-mutable enum type, that's what should be implied by this usage.

Consider this:
```lua
-- This shouldn't be changed at runtime, used like an enum
local pathingDirection =
{
    TO_EAST = 0,
    TO_WEST = 1,
}

-- Do whatever you like with this
local currentDirection = pathingDirection.TO_EAST
```

zach2good (changes requested):

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,355 @@
...
+
+local paths =
+{
+    BLUE_TO_BASEMENT   = 1,
+    BLUE_TO_BLUE       = 2,
+    BLUE_TO_YELLOW     = 3,
+    YELLOW_TO_YELLOW   = 4,
+    YELLOW_TO_BASEMENT = 5
```
zach2good: Needs trailing comma

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
```diff
@@ -4,122 +4,355 @@
...
+        { x = 198,  y = 0,   z = 409 },
+        { x = 191,  y = 0,   z = 405 },
+        { x = 192,  y = 0,   z = 395 },
+        { x = 193,  y = 0,   z = 389 },
+        { x = 187,  y = 0,   z = 382 },
+        { x = 184,  y = 0,   z = 380 },
+        { x = 176,  y = 0,   z = 379 },
+        { x = 157,  y = 0,   z = 380 },
+    },
+    [1] =
```
zach2good: Why are you mixing string indexes for paths and numeric indexes for paths? Am I understanding that now these should all be indexed and looked up with their enum? `paths.BLUE_TO_BASEMENT` etc?

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
zach2good: You could even enforce this (at the CI level) with LLS annotations: https://luals.github.io/wiki/annotations/#enum

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Fixed

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Fixed

scripts/zones/VeLugannon_Palace/mobs/Zipacna.lua
0x05010705: Let me see what I can come up with on this.

0x05010705: - Updated the `paths` enum to include the spawn point paths (thus allowing them to be referenced by their enum as well).
- Updated the `pathNodes` table to use the `paths` enums instead of hard-coded numbers.
- Adjusted the location of the spawn nodes in `pathNodes` to the end to match the order they appear in the `paths` enum.
- Re-Tested everything to ensure the next indices are working as expected.

zach2good (approved):

PR #6519, opened 2024-12-12
[core] Use prepared statements for blobs

WinterSolstice8:
**_I affirm:_**

Use prepared statements for blobs
Code is rough, meant for review & CI

!monstrosity with no errors
set blue magic spells, /logout, /login, see same blue magic spells

### DIFF
--- a/src/map/monstrosity.cpp
+++ b/src/map/monstrosity.cpp
@@ -225,41 +225,46 @@ void monstrosity::WriteMonstrosityData(CCharEntity* PChar)
     }
 
     const char* Query = "REPLACE INTO char_monstrosity SET "
-                        "charid = '%u', "
-                        "current_monstrosity_id = '%d', "
-                        "current_monstrosity_species = '%d', "
-                        "current_monstrosity_name_prefix_1 = '%d', "
-                        "current_monstrosity_name_prefix_2 = '%d', "
-                        "current_exp = '%d', "
-                        "equip = '%s', "
-                        "levels = '%s', "
-                        "instincts = '%s', "
-                        "variants = '%s', "
-                        "belligerency = '%d', "
-                        "entry_x = '%.3f', "
-                        "entry_y = '%.3f', "
-                        "entry_z = '%.3f', "
-                        "entry_rot = '%u', "
-                        "entry_zone_id = '%d', "
-                        "entry_mjob = '%d', "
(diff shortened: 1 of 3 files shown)


src/map/monstrosity.cpp
```diff
@@ -225,41 +225,46 @@ void monstrosity::WriteMonstrosityData(CCharEntity* PChar)
...
+
+    auto equip     = db::encodeToBlob(PChar->m_PMonstrosity->EquippedInstincts);
+    auto levels    = db::encodeToBlob(PChar->m_PMonstrosity->levels);
+    auto instincts = db::encodeToBlob(PChar->m_PMonstrosity->instincts);
+    auto variants  = db::encodeToBlob(PChar->m_PMonstrosity->variants);
+
+    std::basic_istream<char> equipStream(&equip);
+    std::basic_istream<char> levelsStream(&equip);
+    std::basic_istream<char> instinctsStream(&equip);
+    std::basic_istream<char> variantsStream(&equip);
```
zach2good, line 255: I'd like to be able to get rid of this intermediate step, so we can do this:

```cpp
    db::preparedStmt(Query,
              PChar->id,
              PChar->m_PMonstrosity->MonstrosityId,
              PChar->m_PMonstrosity->Species,
              PChar->m_PMonstrosity->NamePrefix1,
              PChar->m_PMonstrosity->NamePrefix2,
              PChar->m_PMonstrosity->CurrentExp,
              db::encodeToBlob(PChar->m_PMonstrosity->EquippedInstincts)
```

and then once we have specialisations for structs and std::arrays:

```cpp
    db::preparedStmt(Query,
              PChar->id,
              PChar->m_PMonstrosity->MonstrosityId,
              PChar->m_PMonstrosity->Species,
              PChar->m_PMonstrosity->NamePrefix1,
              PChar->m_PMonstrosity->NamePrefix2,
              PChar->m_PMonstrosity->CurrentExp,
              PChar->m_PMonstrosity->EquippedInstincts // <-- Just the array/struct
```

zach2good: If the `basic_istream` is the thing that's holding the data, that could be moved inside of the `db` function call. Which is better, because  then if we're later offloading queries to worker threads, all of this will be copied to the worker thread without concern - which might be a bit messier if it's an intermediate step that exists outside of `db` stuff

PR #6541, opened 2024-12-17
[core] Dirty & Undirty inventory when changing gear

WinterSolstice8:
**_I affirm:_**

Replaces & improves upon #4885

Retail sends dirty/undirty packets for _every effected container_ (found via capture) when changing gear (see https://github.com/atom0s/XiPackets/blob/main/world/server/0x001D/README.md for the dirty/undirty stuff)
This also fixes temp items disappearing on gear equip (Thanks for the report via old PR, mowford)

From #4885

Before this change, replicate the issue via:
- `!exec player:addTempItem(5393)`
- open quick item menu to see temp item available
- Then equip any pieced of gear and see that the temp item disappears from quick item menu

same process for unequipping an item.

After change see that the item stays visible after equipping and unequipping gear

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -969,6 +969,23 @@ void CCharEntity::PostTick()
         pushPacket(new CCharAppearancePacket(this));
     }
 
+    // notify client containers are dirty and then no longer dirty
+    if (!dirtyInventoryContainers.empty())
+    {
+
+        // Notify client containers were dirty
+        // Note: Retail sends this in the same chunk as the inventory equip packets, but it doesnt seem to matter as long as it arrives
+        for (const auto& [container, dirty]: dirtyInventoryContainers)
+        {
+            pushPacket(new CInventoryFinishPacket(container));
+        }
+
+        dirtyInventoryContainers.clear();
+
+        // Notify client containers are now ok
+        pushPacket(new CInventoryFinishPacket());
+    }
+
     if (ReloadParty())
     {
         charutils::ReloadParty(this);
--- a/src/map/entities/charentity.h
+++ b/src/map/entities/charentity.h
@@ -23,6 +23,7 @@
 #define _CHARENTITY_H
 
 #include "event_info.h"
+#include "item_container.h"
 #include "monstrosity.h"
 #include "packets/char.h"
 #include "packets/entity_update.h"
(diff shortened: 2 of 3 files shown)


Xaver-DaRed (approved):

src/map/entities/charentity.cpp
```diff
@@ -969,6 +969,23 @@ void CCharEntity::PostTick()
         pushPacket(new CCharAppearancePacket(this));
     }
 
+    // notify client containers are dirty and then no longer dirty
+    if (dirtyInventoryContainers.size() > 0)
```
zach2good: `!dirtyInventoryContainers.empty()`

src/map/entities/charentity.h
```diff
@@ -511,6 +511,9 @@ class CCharEntity : public CBattleEntity
     bool getBlockingAid() const;
     void setBlockingAid(bool isBlockingAid);
 
+    // Send updates about dirty containers in post tick
+    std::map<uint8, bool> dirtyInventoryContainers;
```
zach2good: You could replace the uint8 with CONTAINER_ID

src/map/entities/charentity.cpp
```diff
@@ -969,6 +969,23 @@ void CCharEntity::PostTick()
...
+    {
+
+        // Notify client containers were dirty
+        // Note: Retail sends this in the same chunk as the inventory equip packets, but it doesnt seem to matter as long as it arrives
+        for (auto container: dirtyInventoryContainers)
```
zach2good: `const auto [container, dirty]`

zach2good (approved):
Cool!

src/map/entities/charentity.cpp
WinterSolstice8: done (this is way better than using .first .second etc)

src/map/entities/charentity.h
WinterSolstice8: done

src/map/entities/charentity.cpp
WinterSolstice8: done

src/map/entities/charentity.cpp
WinterSolstice8: Hmm. Clang doesn't like this:
```
 /home/runner/work/server/server/src/map/entities/charentity.cpp:978:25: error: loop variable '[container, dirty]' creates a copy from type 'std::pair<const CONTAINER_ID, bool> const' [-Werror,-Wrange-loop-construct]
        for (const auto [container, dirty]: dirtyInventoryContainers)
                        ^
/home/runner/work/server/server/src/map/entities/charentity.cpp:978:14: note: use reference type 'std::pair<const CONTAINER_ID, bool> const &' to prevent copying
        for (const auto [container, dirty]: dirtyInventoryContainers)
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                        &
```
Looks like it wants a ref. will play around with this

WinterSolstice8: It wanted: `const auto& [container, dirty]`

zach2good (approved):

PR #6542, opened 2024-12-17
Mob run speed changes

cocosolos:
**_I affirm:_**

Follow-up to #6303 

There is indeed an issue with kiting currently, though it could have been reported in a way more conducive to finding a solution.

The step distance calculation has been changed, and now seems more in line with retail speeds in various conditions. I would appreciate more testing, but I'm pretty confident in the current results. I renamed the setting and mob mod to make it a little more clear what it's used for.

Mob controller controls if mob should run or not, so remove the distance checks from the speed calculations. Rework GetSpeed to UpdateSpeed. Mobs only need to update their speed during pathfinding, and players only for update packets, so this allows us to read speed without necessarily recalculating (fixes an issue in the entity update packet where the run speed wasn't changing like it should).

~~I also renamed speedsub to speedBase in a separate commit, since I think it's more descriptive. I can drop this if wanted.~~

## Demonstration

Current: https://www.youtube.com/watch?v=I_euKvJ2zFc

Improved: https://www.youtube.com/watch?v=CDAWdI0Vmh4

Retail: https://www.youtube.com/watch?v=-rFpXKWLUvY

This is the route from the Aydeewa survival guide to the TVR RoE NM. Note towards the end of the video where the funguars deaggro, which matches retail.

Kite things with different speeds, compare to retail.

Batallia Downs has quick access to Goblins (40 speed) and tigers (60 speed).

### DIFF
--- a/src/map/ai/controllers/mob_controller.cpp
+++ b/src/map/ai/controllers/mob_controller.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
 ===========================================================================
 
   Copyright (c) 2010-2015 Darkstar Dev Teams
@@ -633,6 +633,7 @@ void CMobController::FaceTarget(uint16 targid)
     {
         PMob->PAI->PathFind->LookAt(targ->loc.p);
     }
+    PMob->UpdateSpeed();
 }
 
 void CMobController::Move()
--- a/src/map/ai/helpers/pathfind.cpp
+++ b/src/map/ai/helpers/pathfind.cpp
@@ -361,17 +361,23 @@ void CPathFind::FollowPath(time_point tick)
 void CPathFind::StepTo(const position_t& pos, bool run)
 {
     TracyZoneScoped;
-    float speed = GetRealSpeed();
-
-    int8 mode = 2;
-
-    if (!run)
+    bool speedChange = false;
+    if (auto* PBattleEntity = dynamic_cast<CBattleEntity*>(m_POwner))
     {
-        mode = 1;
-        speed /= 2;
+        speedChange = PBattleEntity->speed != PBattleEntity->UpdateSpeed(run);
     }
 
-    float stepDistance = (speed / 10) / 2;
+    float speed = m_POwner->speed;
+
+    if (const auto* PMobEntity = dynamic_cast<CMobEntity*>(m_POwner))
+    {
+        if (PMobEntity->speed == 0 && (m_roamFlags & ROAMFLAG_WORM))
+        {
+            speed = 20;
+        }
+    }
+
+    float stepDistance = speed / (run ? 50 : 40);
     float distanceTo   = distance(m_POwner->loc.p, pos);
     float diff_y       = pos.y - m_POwner->loc.p.y;
 
@@ -434,12 +440,9 @@ void CPathFind::StepTo(const position_t& pos, bool run)
         }
     }
 
-    m_POwner->loc.p.moving += (uint16)((0x36 * ((float)m_POwner->speed / 0x28)) - (0x14 * (mode - 1)));
+    m_POwner->loc.p.moving += speedChange ? 0x28 : 0x35;
 
-    if (m_POwner->loc.p.moving > 0x2fff)
-    {
-        m_POwner->loc.p.moving = 0;
-    }
+    m_POwner->loc.p.moving %= 0x2000;
 
     m_POwner->updatemask |= UPDATE_POS;
 }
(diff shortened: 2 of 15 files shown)


sql/npc_list.sql
```diff
@@ -26,7 +26,7 @@ CREATE TABLE `npc_list` (
   `pos_z` float(7,3) NOT NULL DEFAULT 0.000,
   `flag` int(10) unsigned NOT NULL DEFAULT 0,
   `speed` tinyint(3) unsigned NOT NULL DEFAULT 40,
-  `speedsub` tinyint(3) unsigned NOT NULL DEFAULT 40,
+  `speedbase` tinyint(3) unsigned NOT NULL DEFAULT 40,
```
Xaver-DaRed: Capture tools use speedsub as the name. I would keep the original name unless the change is necesary.

cocosolos: Yeah not necessary, just better aligns with what it actually is and XIPackets repo documentation.

Something else I noticed somewhat related to this is that in incoming packet 0x00E at position 0x25 the server sends the entity model size, which I think is used for calculating melee range. Right now I don't think any of the capture tools parse this specifically and I think most model sizes we are using are mostly just made up? (not sure on that) So I would like to see that field added to whatever capture tool is capturing this stuff. I assume the same tool is what's using speedsub, so idk if this would be a good time to change it, or keep speedsub and add model size, or whatever you guys wanna do (not sure who maintains that stuff).

I do have a half finished packet parser that can go back over old caps and can pull that model info out as well, but would be nice to have it parsed out along with the other stuff. It's no issue removing that commit here though if that's what you wanna do. Let me know, I'll probably keep this in draft for a few days until I can get on retail anyway.

src/map/ai/helpers/pathfind.cpp
```diff
@@ -361,17 +361,18 @@ void CPathFind::FollowPath(time_point tick)
...
-        mode = 1;
-        speed /= 2;
+        if (speed == 0 && (m_roamFlags & ROAMFLAG_WORM))
+        {
+            speed = 20;
+        }
     }
 
-    float stepDistance = (speed / 10) / 2;
+    float stepDistance = speed / (run ? 50 : 40);
```
cocosolos: Just wanted to comment on this. I'm pretty sure everything to do with worm speed and movement is wrong, but this works for now I think. Will revisit in the future.

Interestingly, this new calculation for stepDistance produces the same values as before the run speed boost changes:

```
40 / 2 / 10 / 2 = 1 = 40 / 40
40 / 10 / 2 = 2 = 100 / 50
60 / 2 / 10 / 2 = 1.5 = 60 / 40
60 / 10 / 2 = 3 = 150 / 50
```

But doing it this way allows us to send the correct speed in update packets (not sure how gravity and stuff was handled before, but it looks correct now).

src/map/ai/helpers/pathfind.cpp
```diff
@@ -434,12 +435,9 @@ void CPathFind::StepTo(const position_t& pos, bool run)
         }
     }
 
-    m_POwner->loc.p.moving += (uint16)((0x36 * ((float)m_POwner->speed / 0x28)) - (0x14 * (mode - 1)));
+    m_POwner->loc.p.moving += speedChange ? 0x28 : 0x35;
```
cocosolos, line 443: I'm not sure what this was all about, but quickly looking at some caps I was seeing a pretty consistent 53 move increase, with occasional 40s. Rarely I would see something +/- 1 or 2 (a couple 39, 51, 52, but mostly 53 and 40). 40 seemed to coincide with melee attacks/speed changes (range related) so that's how I ended up with this. I noticed that if I didn't decrease this during speed changes, the mob would teleport forward occasionally when the distance got too large (for instance, when kiting a tiger up to about 30', it would teleport to 20').

src/map/ai/helpers/pathfind.cpp
```diff
@@ -434,12 +435,9 @@ void CPathFind::StepTo(const position_t& pos, bool run)
...
 
-    if (m_POwner->loc.p.moving > 0x2fff)
-    {
-        m_POwner->loc.p.moving = 0;
-    }
+    m_POwner->loc.p.moving %= 0x2001;
```
cocosolos: This is 13 bits, but I saw it going from 8179 to 39. This could be `(8179 + 53) % 8193` or `(8179 + 52) % 8192`. I went with the former to stick with the 53, but I guess the latter makes more sense if you consider the +/-1 variance. It seems to work fine though so I don't think it matters either way (this way just double counts 0).

sql/npc_list.sql
WinterSolstice8: entity size is strange, player size is also different between races, I had tested galka vs hume (different sizes in the packets), and they both have the exact same "strikes true" range when using /ra. Very weird.

image

src/map/ai/controllers/mob_controller.cpp
```diff
@@ -1,4 +1,4 @@
-﻿/*
```
WinterSolstice8, line 1: Looks like the files BOM (byte order mark) got changed

sql/npc_list.sql
cocosolos: What kind of weapon is that? If we disregard the player model, which I think is done in some or all cases, then 8.3 - 2.8 = 5.5, which looks like the sweet spot for guns. I'm curious what your max melee is too.

src/map/ai/controllers/mob_controller.cpp
cocosolos: I think it may have been wrong? Shows as LF for me. I can't even change it to CRLF, editorconfig changes it to LF on save.

sql/npc_list.sql
WinterSolstice8: Oh, that's an old pic when I was testing ranged distance. I think it was crossbow?

Didn't check melee distance at that time

cocosolos: What plugin is that and where can I get it? After some preliminary testing I'm seeing some very interesting results.

```
Model size: 0.3
Max Melee (Taru): 2.6
Max Melee (Hume): 2.7

Model size: 1.1
Max Melee (Taru): 3.4
Max Melee (Hume): 3.5
Max Melee (Galka): 3.9

Model size: 2.7
Max Melee (Taru): 5.0
Max Melee (Hume): 5.1
Max Ranged (Both): 25.0

Model size: 3.4
Max Melee (Galka): 6.2
```

This is on flat ground. Being on raised ground seems to increase this, so vertical distance doesn't seem to be a factor (probably some vertical limit instead of being factored into distance). From your screenshot it looks like Hume is 0.4, I'm guessing Taru is 0.3, which would make melee distance something like `enemy model size + player model size + 2.0`. Ranged seems to be point to point 25.0 with no model size adjustments. Model size also doesn't seem to be sent in every packet.

src/map/ai/controllers/mob_controller.cpp
Xaver-DaRed: This happens constantly. I'm sure if one were to check them all, they would be pretty inconsistent

sql/npc_list.sql
WinterSolstice8: I just hacked up ashita's distance

```
    local player = GetPlayerEntity()
    -- Ensure there is a target..
    local target = GetEntity(AshitaCore:GetMemoryManager():GetTarget():GetTargetIndex(0));
    if (target ~= nil) then
        distance.font.text = ('%.1f:%.1f; TargetSize: %.1f, YourSize: %.1f, TargetModelSize:%f'):fmt(math.sqrt(target.Distance), math.sqrt(target.Distance)-target.ModelHitboxSize-player.ModelHitboxSize, target.ModelHitboxSize, player.ModelHitboxSize, target.ModelSize);
    else
        distance.font.text = '';
    end
```
note that this is junk and shouldn't be used
`math.sqrt(target.Distance)-target.ModelHitboxSize-player.ModelHitboxSize`

I think windower has an equivalent for the model hitbox size but not 100% sure. Note that model size is independent of the hitbox size, theoretically SE could send any arbitrary number for either

cocosolos: I'm not sure if Windower tracks this, but it makes sense that Ashita does. [It's not in their fields file](https://github.com/Windower/Lua/blob/45db20d088daf72d97e358c3f4b79861d9e3d92c/addons/libs/packets/fields.lua#L1472), which is what packetviewer uses, but they do have some model stuff in their wiki. I'll have to look later. I was just using packetviewer to look at the update packet.

> Note that model size is independent of the hitbox size, theoretically SE could send any arbitrary number for either

I think this is the hitbox size. [This is what I'm looking at](https://github.com/atom0s/XiPackets/tree/main/world/server/0x000E#g), which would be ModelHitboxSize assuming Ashita would use the same wording. Considering the model sizes + 2 melee distance holds across multiple different mobs showing different model sizes I think I'm on the right track here. I know there's some other model size related things too, like S/M/L players and a bit that makes mobs bigger. I wonder if any of that is factored in as well. I'm honestly surprised player race makes a difference in melee range.

cocosolos: To the original topic, I reverted the speedsub -> speedbase change. This model size stuff will need additional captures. NPCLogger saves the raw packet so can probably reparse a lot of that stuff too.

src/map/entities/baseentity.cpp
```diff
@@ -231,3 +231,8 @@ bool CBaseEntity::IsDynamicEntity() const
 {
     return this->targid >= 0x700;
 }
+
+uint8 CBaseEntity::UpdateSpeed(bool run)
```
WinterSolstice8: `run` is unused now. Is this intentional?

WinterSolstice8: If it is, denote it's intentional with

```suggestion
uint8 CBaseEntity::UpdateSpeed(bool run)
{
    std::ignore = run;
```

cocosolos: This was only used for battle entities, so I just moved it up and added a type check in pathfind.

WinterSolstice8: That works

WinterSolstice8 (approved):

sql/npc_list.sql
WinterSolstice8: > I think this is the hitbox size. [This is what I'm looking at](https://github.com/atom0s/XiPackets/tree/main/world/server/0x000E#g), which would be ModelHitboxSize assuming Ashita would use the same wording.

Yup, that's the one.

PR #6554, opened 2024-12-18
Update Lord's Cuirass to be craftable

bliven-was-taken:
**_I affirm:_**

Uncomment out Lord's Cuirass. The recipe is correct as is, but it was commented out because of conflicting reports when the topaz database was pulled in 2020. 

Comment on website states recipe is wrong on ffxiah and should only be 2 gold ingots. Recipe as listed on ffxiah has 9 ingredients which I don't believe is possible. 
https://www.ffxiah.com/item/13757/lords-cuirass

This is the ffxiclopedia update which corrected the recipe from 3 gold ingot to 2 gold ingot. 
https://ffxiclopedia.fandom.com/wiki/Lord%27s_Cuirass?oldid=183741

Bg wiki states the recipe is as listed.
https://www.bg-wiki.com/ffxi/Lord%27s_Cuirass

2004 post that lists most of the correct items, it says gold board instead of sheet and dark instead of darksteel
https://ffxi.allakhazam.com/db/item.html?fitem=3356&post=1&replyto=107847415562020&p=1#m107847415562020


WinterSolstice8 (approved):

zach2good (approved):

Xaver-DaRed (approved):
Japanesse wiki seems to agree

Issue #6407, opened 2024-11-04
🔨 Lua: `math.random()` calls should always be bounded


zach2good:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature

Looking at this PR: https://github.com/LandSandBoat/server/pull/6406

I see we have 100 results for `math.random()` with no arguments, which returns a float between 0-1, which we typically are multiplying by 100 and then flooring.

We have 4 results for the regex: `random\(\d\)`.

I think we should be using `math.random(1, 100)` in pretty much every case, unless we're picking from a container, matching onto stats, etc.

PR #6558, opened 2024-12-19
[Lua] Made math.randoms more explicit

Flibe-XI:
**_I affirm:_**

Address #6407.  Attempt to standardize lua math.random calls to return ints from a bounded range rather than floats.  There are more instances of this, but not trying to overload this PR, as the specific instances vary a lot in context, and some require more testing than others.

Be summoner, use blood pact "Glittering ruby", get random stat buff.

### DIFF
--- a/scripts/actions/abilities/pets/glittering_ruby.lua
+++ b/scripts/actions/abilities/pets/glittering_ruby.lua
@@ -10,26 +10,23 @@ end
 
 abilityObject.onPetAbility = function(target, pet, petskill, summoner, action)
     --randomly give str/dex/vit/agi/int/mnd/chr (+12)
-    local effect = math.random()
-    local effectid = xi.effect.CHR_BOOST
+    local effects =
+    {
+        xi.effect.STR_BOOST,
+        xi.effect.DEX_BOOST,
+        xi.effect.VIT_BOOST,
+        xi.effect.AGI_BOOST,
+        xi.effect.INT_BOOST,
+        xi.effect.MND_BOOST,
+        xi.effect.CHR_BOOST,
+    }
+
+    effectId = utils.randomEntry(effects)
+    effectPower = math.random(12, 14)
 
     xi.job_utils.summoner.onUseBloodPact(target, petskill, summoner, action)
 
-    if effect <= 0.14 then --STR
-        effectid = xi.effect.STR_BOOST
-    elseif effect <= 0.28 then --DEX
-        effectid = xi.effect.DEX_BOOST
-    elseif effect <= 0.42 then --VIT
-        effectid = xi.effect.VIT_BOOST
-    elseif effect <= 0.56 then --AGI
-        effectid = xi.effect.AGI_BOOST
-    elseif effect <= 0.7 then --INT
-        effectid = xi.effect.INT_BOOST
-    elseif effect <= 0.84 then --MND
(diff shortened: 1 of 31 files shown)


Xaver-DaRed (approved):

scripts/actions/abilities/pets/glittering_ruby.lua
```diff
@@ -10,34 +10,30 @@ end
...
+        xi.effect.STR_BOOST,
+        xi.effect.DEX_BOOST,
+        xi.effect.VIT_BOOST,
+        xi.effect.AGI_BOOST,
+        xi.effect.INT_BOOST,
+        xi.effect.MND_BOOST,
+        xi.effect.CHR_BOOST,
+    }
+
+    effectId = math.random(effects[1], effects[#effects])
```
zach2good: You can use `utils.randomEntry` for this sort of thing

scripts/actions/abilities/pets/glittering_ruby.lua
```diff
@@ -10,34 +10,30 @@ end
...
-    elseif effect <= 0.56 then --AGI
-        effectid = xi.effect.AGI_BOOST
-    elseif effect <= 0.7 then --INT
-        effectid = xi.effect.INT_BOOST
-    elseif effect <= 0.84 then --MND
-        effectid = xi.effect.MND_BOOST
-    end
-
-    target:addStatusEffect(effectid, math.random(12, 14), 0, 90)
+    target:addStatusEffect(effectId, math.random(12, 14), 0, 90)
```
zach2good: I'd probably extract this out to be like:
```
local effectPower = math.random(12, 14)
... then use it ...
```

scripts/actions/abilities/pets/spring_water.lua
```diff
@@ -32,7 +32,7 @@ abilityObject.onPetAbility = function(target, pet, petskill, summoner, action)
     target:wakeUp()
     target:delStatusEffect(xi.effect.SILENCE)
 
-    if math.random() > 0.5 then
+    if math.random(1, 2) == 1 then
```
zach2good: I feel like there has to be a nicer way of representing "random 50% chance". I've had a little poke around and I can't see any helpers, but I vaguely remember there being one. 

Failing that, it might be better to do `math.random(1, 100) > 50` (if that evaluates to the same, I'm not paying huge attention to the bounds here)

zach2good (approved):
Looks good, just minor nits 👍

zach2good:
If you're going to take this on, here's the regex I just used to look up all empty/unbounded random calls in Lua: `math\.random\(\s*\d*\s*\)`

I use VSCode's search with regex mode on for 127 results, or so

scripts/actions/abilities/pets/spring_water.lua
Flibe-XI: I like `<= 50` as that would be clearer and also be consistent with the instances with `rand(1,4) == 1` that I would like to convert to `rand(1, 100) <= 25`.  This way everything is very obvious what the chance is.  Will also continue looking for that util.

zach2good:
Ah, we used to have this:
```lua

function utils.chance(likelihood)
    return math.random(100) <= likelihood
end
```

But we removed it to make sure that the limits of each call were more obvious at each callsite. So just go for 1,100, or whatever the range is that you're using

Issue #6518, opened 2024-12-12
Ashita keeps asking me for my username and password


AlexandreCA:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## OS / platform the server is running (if known)

**Ubuntu**

## Branch affected by issue 

`base`

## Steps to reproduce

Nothing special to do. Since 3 server updates. Ashita keeps asking me for my username and password. If I update the database by doing an express update. This fixes the problem temporarily.

![image](https://github.com/user-attachments/assets/14e16c95-1301-4e01-be50-834270e5f606)

WinterSolstice8:
You need to use xiloader from here. the builtin xiloader from Ashita doesn't support the TLS connection yet
https://github.com/LandSandBoat/xiloader/releases

Edit: I see you are actually using the right one, I made an assumption based on the title, sorry

WinterSolstice8:
I don't see why updating the DB could fix your problem, are there any errors in xi_connect?

AlexandreCA:
Exactly none. I came across this fix unexpectedly. I don't know why it does this. The guy who calls himself General on the Discord would have heard/seen this problem from someone else. Not knowing where it could come from or how to cause it, I unfortunately can't say more.

FFXILevelDown:
This is General, We ran into this issue yesterday where I had to repeatedly update someone's password in sql. Once I set the password in sql, he could log in, if we tried to reset the password thru Ashita, it would register the change but would state it is an invalid login / password and I would have to reset it in sql again. Work has had me tied up so I haven't looked into the issue yet past experiencing the issue. Since I haven't had time to look into it to provide my own data,  I suggested he open an issue with his findings.

WinterSolstice8:
I have a fix that prevents password changing not working. Thanks for the tip. I don't know if this solves OPs issue though.

AlexandreCA:
I am currently applying the update. I will follow up in a few days to say whether or not things have been fixed or not.

[...]

PR #6564, opened 2024-12-20
[login] Add Task Manager ticks

WinterSolstice8:
**_I affirm:_**

Adds task manager ticks to Login

Related: #6518

Wait a while after running login and hopefully see `(C++) Pinging database to keep connection alive` eventually

### DIFF
--- a/src/login/connect_server.cpp
+++ b/src/login/connect_server.cpp
@@ -81,8 +81,10 @@ ConnectServer::ConnectServer(int argc, char** argv)
         handler<view_session> view(io_context, settings::get<uint32>("network.LOGIN_VIEW_PORT"));
         handler<data_session> data(io_context, settings::get<uint32>("network.LOGIN_DATA_PORT"));
         asio::steady_timer    cleanup_callback(io_context, std::chrono::minutes(15));
+        asio::steady_timer    tick_callback(io_context, std::chrono::seconds(1)); // Start at one second, TaskMgr will report back with future delays
 
         cleanup_callback.async_wait(std::bind(&ConnectServer::periodicCleanup, this, std::placeholders::_1, &cleanup_callback));
+        tick_callback.async_wait(std::bind(&ConnectServer::tick_callback, this, std::placeholders::_1, &tick_callback));
 
         // NOTE: io_context.run() takes over and blocks this thread. Anything after this point will only fire
         // if io_context finishes!
@@ -157,6 +159,21 @@ void ConnectServer::periodicCleanup(const asio::error_code& error, asio::steady_
             // reset timer
             timer->expires_at(timer->expiry() + std::chrono::minutes(15));
             timer->async_wait(std::bind(&ConnectServer::periodicCleanup, this, std::placeholders::_1, timer));
+        }
+    }
+}
+
+void ConnectServer::tick_callback(const asio::error_code& error, asio::steady_timer* timer)
+{
+    if (!error)
+    {
+        if (Application::IsRunning())
+        {
+            duration next = CTaskMgr::getInstance()->DoTimer(server_clock::now());
+
+            // Reset timer
+            timer->expires_after(next);
+            timer->async_wait(std::bind(&ConnectServer::tick_callback, this, std::placeholders::_1, timer));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


zach2good (approved):

StabbyCutyou:
Hello, unwarranted question from someone just trying to learn the codebase:

if ping_connection calls getState() https://github.com/LandSandBoat/server/blob/724226264029861686b61108af450eb8e5f415c4/src/common/taskmgr.cpp#L41

And getState() adds a task which calls ping_connection with a type of CTaskMgr::TASK_INTERVAL
https://github.com/LandSandBoat/server/blob/724226264029861686b61108af450eb8e5f415c4/src/common/database.cpp#L136

And CTaskMgr::TASK_INTERVAL type Tasks get re-enqueued after a run of DoTimer()
https://github.com/LandSandBoat/server/blob/724226264029861686b61108af450eb8e5f415c4/src/common/taskmgr.cpp#L117-L120

Won't this slowly build up a never ending list of Tasks to ping the DB, as ping_connection keeps calling getState() which keeps adding a new task to call ping_connection, and that task is pushed back onto the TaskManager stack after each run?

Apologies if i'm just mis-reading the code and causing a distraction.

WinterSolstice8:
It returns early here https://github.com/LandSandBoat/server/blob/724226264029861686b61108af450eb8e5f415c4/src/common/database.cpp#L90

WinterSolstice8:
Just went to double check ,and after an hourish the debugger thinks there's only one task here: 
image

WinterSolstice8:
image
Finally triggered

StabbyCutyou:
Womp womp

Apologies! Appreciate you taking the time to explain it!

Xaver-DaRed (approved):

zach2good:
If this works great, but inside of getState is meant to be self-repairing, so then it doesn't work :(

StabbyCutyou:
I noticed that as well but I assumed lack of familiarity with the codebase meant I was missing something more obvious.

Is it possible that isvalid() and reconnect() are implemented in a way where they aren't aware of the server having closed the connection, and are returning true (and thus, the broken connection) even though they haven't actually validated / properly reconnected? Ie: they're designed to repair the client in the event of a state that the client is aware of / caused (which in this case, they wouldn't know that the server timed out the connection)?

I poked around the driver itself but I'm supremely out of my element with cpp, so figured I'd raise it as a possibility. For all I know, I'm not even looking at the right thing:

https://github.com/ericcurtin/mariadb-connector-cpp/blob/60a87352d2edfad375eb60eafd6d52ba84eb9404/driver/mysql_connection.cpp#L1550

And

https://github.com/ericcurtin/mariadb-connector-cpp/blob/60a87352d2edfad375eb60eafd6d52ba84eb9404/driver/mysql_connection.cpp#L1590

Substantially lowering the idle timeout on Maria's side might be one way to validate this more quickly. If I didn't have to work through the weekend, I'd be happy to donate time trying to help test a fix here.

WinterSolstice8:
Jury's still out on if it fixed anything

I never had the problem on this machine, this is simply adding task manager ticks that happens to use sql ping, but I do agree that the "self healing" bit does not seem to work for others

StabbyCutyou:
The original problem, as I experienced it, would only happen in specific circumstances - namely, the connection would need to be closed on the remote db server without the client being aware.

In my example, it happens consistently due to idleness (i set my server up, and have been playing around with it sporadically. I've been reading the code a lot more than I've actually used the emulator tbh, so it's easy to repro for me, and my db and app servers are not the same machines, so I'm not going over the local loopback).

If your server is highly trafficked, you might literally never experience it.

zach2good:
Since login, search, and world server are all refactors to use the new `db` namespace, I'd like to have this PR go out into the wild for a few weeks and see if it does the reconnection logic without periodic pings: https://github.com/LandSandBoat/server/pull/6568

If we get back reports that it doesn't work, then this PR can come in on top of it

StabbyCutyou:
I can do an update tonight and let the server run for a few days and then see if the error recurs.

Appreciate the attention on the issue!

zach2good:
This is another db improvement, which should solidify the self-healing of connections if they break: https://github.com/LandSandBoat/server/pull/6571

WinterSolstice8:
I don't think this is necessary anymore

PR #6566, opened 2024-12-20
Move Draw-In to Lua

cocosolos:
**_I affirm:_**

> // TODO: can this be moved to scripts entirely?

Other implementations of draw-in rely heavily on mob mods and hard coded values. With the variety of types of draw ins, it should just be pulled out into the scripts. This provides a framework for creating one or multiple conditions to trigger draw in, draw-in position that can be set on the fly and not necessarily at the mobs position, and options to place the player somewhere around a point (center of hitbox, in front, behind, some specific spot(s) on the map, etc).

Ports https://github.com/AirSkyBoat/AirSkyBoat/pull/2031 arenaDrawIn function and expands it to handle basically all types of draw in. Gets rid of the mob mods and moves most functionality out of core into Lua. Some of the conditions need to be verified but work well enough for now (for example, Cassie may have boundaries at each door/hallway instead of a boundary around her spawn point). A draw_in mixin was made to provide an easy way to add basic draw in that triggers when the target is >2x melee range and draws in to center of mob. A mimic family mixin provides similar utility, drawing in to the front of the mob. For things that don't fit the mixins, it's easy enough to rig the functionality you need. I converted all existing draw-in to this system so there's plenty of different kinds of examples. We still check for a valid floor position, so I'm not sure if this is a big loss.

[Here's a video showing some of the changes.](https://youtu.be/S60fghK2nOs)

16806227 Tiamat ID
17379783 Mimic

!gotoid
!mobhere

Spawn some stuff that draws in, try to escape.


Xaver-DaRed:
As a general comment, while I like what I see in general, it worries me all of it is done in `onMobFight`.

Unfortunately, we do not have something that ticks less times, like onMobAttackCheck, or something not prone to tick 3 times per second.

This has nothing to do with you or the PR. Just writing my thoughts.

cocosolos:
> As a general comment, while I like what I see in general, it worries me all of it is done in `onMobFight`.
> 
> Unfortunately, we do not have something that ticks less times, like onMobAttackCheck, or something not prone to tick 3 times per second.
> 
> This has nothing to do with you or the PR. Just writing my thoughts.

True, and I think there's some added variability because things can delay the combat tick, like mobskills (I may be wrong, there's a lot going on). That being said, it's currently being handled in the Move function which is also called on the combat tick. The position based conditions will be cheap though, and the distance based ones can probably used distance squared to lighten them up a bit.

TracentEden2:
Please check this PR https://github.com/LandSandBoat/server/pull/6007 and take that into account. The PR is still pending as I was waiting for a reply from Zach after I replied to a comment of his. Thanks!

cocosolos:
> Please check this PR #6007 and take that into account. The PR is still pending as I was waiting for a reply from Zach after I replied to a comment of his. Thanks!

Yeah I saw that and #6532 and they both go for relatively complicated approaches to modification that rely heavily on mob mods to do specific behavior and I think this is just a better approach. This provides a real framework that can be as extensible as needed without needing any additional mob mods or dealing with bitmasks. For example both of those implementations have a specific flag or mod just for drawing in to the front of the mob, when there's a huge variety or potential locations. Roc draws in to some area behind it, Naga Raja and probably the other DI wyrms draw in to 90 degrees at their side. Mobs like Tiamat draw in if you cross some boundary that's completely unrelated to the mobs position. That is all easy to deal with with this system and doesn't rely on any extra flags or whatever. There's no reason so much of the draw in code needs to be in the core.

WinterSolstice8:
Is this PR supposed to still be in draft?

cocosolos:
> Is this PR supposed to still be in draft?

I have some coordinates I need to add for Hydra and Khimaira and want to rename utils.sameSide to utils.sameSideOfLine but otherwise I think this is all set. Should have that done some time today.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

zach2good (approved):

PR #6568, opened 2024-12-20
db: Make db:State self-healing on invalid connection (timeouts)

zach2good:
**_I affirm:_**

`connection->reconnect()` seemingly doesn't work. But `->isValid()` wraps `mysql_ping`, so it's a good indicator of if the connection is going to work or not. We now rebuild the connection completely if `isValid()` fails.

### DIFF
--- a/src/common/database.cpp
+++ b/src/common/database.cpp
@@ -45,46 +45,19 @@ namespace
     };
 } // namespace
 
-int32 ping_connection(time_point tick, CTaskMgr::CTask* task)
-{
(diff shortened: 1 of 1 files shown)


zach2good:
Tested by stripping out the ping job that was in there so that it doesn't try to reconnect and pings every 5s. Then restarting my MariaDB service repeatedly and see the new logging for "rebuilding connection" instead of the pings failing because `->reconnect()` doesn't do what it promises

claywar (approved):

PR #6577, opened 2024-12-22
[lua] Properly apply SJ_RESTRICTION effect in Dynamis Dreamlands

sruon:
**_I affirm:_**
- [X] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.

`SJ_RESTRICTION ` effect applied upon entering CoP Dreamlands currently does not execute the underlying lua file, leading to subjob abilities remaining visible to the client and usable unless the client forced a recalculation of the abilities list through the use of an ability modifying the list of abilities (Call Wyvern, Light Arts...)

After troubleshooting, it appeared to me the [sj_restriction.lua](https://github.com/LandSandBoat/server/blob/71887f9d6865f3da82c2cacdb895e5b4d9582c0b/scripts/effects/sj_restriction.lua#L8) was not being executed to begin with (whereas doing `!addeffect SJ_RESTRICTION` would properly do)

This PR changes the way the `SJ_RESTRICTION `effect is applied upon entering Dynamis through the use of different parameters. I am not quite clear on the distinction between the various `addStatusEffect `calls, there might be another underlying issue there.

Steps to reproduce issue on base branch:
```
!gotoid 17261192
!changejob DNC 99
!changesjob WAR49
Enter by selecting "Ready! (Support jobs restricted)"

Result: All WAR abilities are available and working
```

```
!gotoid 17261192
!changejob DNC 99
!changesjob WAR49
Enter by selecting "Ready! (Support jobs restricted)"

Result: WAR abilities are not available
```

### DIFF
--- a/scripts/globals/dynamis.lua
+++ b/scripts/globals/dynamis.lua
@@ -466,7 +466,7 @@ xi.dynamis.zoneOnZoneIn = function(player, prevZone)
                 playerArg:messageBasic(xi.msg.basic.UNABLE_TO_ACCESS_SJ)
             end)
 
-            player:addStatusEffect(xi.effect.SJ_RESTRICTION, 0, 0, 0, 7200)
+            player:addStatusEffect(xi.effect.SJ_RESTRICTION, 0, 0, 0, 0, 0)
         end
 
         player:addStatusEffectEx(xi.effect.DYNAMIS, 0, 0, 3, 3600)


WinterSolstice8 (commented):
I think they meant to use addStatusEffectEx here, because that parameter index is expiry for "Ex". 7200 seconds = 2 hours, which is likely not a coincidence.

Does the effect disappear correctly on zone?

sruon:
> I think they meant to use addStatusEffectEx here, because that parameter index is expiry for "Ex". 7200 seconds = 2 hours, which is likely not a coincidence.
> 
> Does the effect disappear correctly on zone?

Had to go back and test. The status effect does get removed. I tested with HPing and also using Warp.

WinterSolstice8 (approved):

PR #6582, opened 2024-12-23
[lua,sql] Rolanberry Fields NM audit

CriticalXI:
**_I affirm:_**

- Adds missing TP move Flurry of Rage and sets Eldritch Edge to exclusively use this move
   - https://www.youtube.com/watch?v=HVZJDumonG8
   - Did a capture myself that I didn't record back in 2021 for this NM to obtain the animation ID
- Black Triple Stars now despawns at daytime
- Simurgh now uses Mighty Strikes, draws in, and has extra evasion.
   - https://www.youtube.com/watch?v=jeE6tU5OfW8
   - https://www.youtube.com/watch?v=AxhBLHCh5G0
- Ravenous Crawler now exclusively uses Sticky Thread.
  - https://www.youtube.com/watch?v=EfHCbzraVRY
- Adjusted drop lists of all NMs to match ffxidb/ffxiclopedia (whichever had more reported drops)
- Added several spawn points to all NMs to add variance to spawn locations

Use !spawnmob to test changes to each NM in game


CriticalXI:
Pending completion of #6566 for proper draw in behavior for Simurgh.

scripts/zones/Rolanberry_Fields/mobs/Simurgh.lua
```diff
@@ -2,11 +2,35 @@
...
+    require('scripts/mixins/rage'),
+    require("scripts/mixins/job_special")
+}
 -----------------------------------
 ---@type TMobEntity
 local entity = {}
 
+entity.onMobInitialize = function(mob)
+    mob:addMod(xi.mod.SLEEPRES, 100)
+    mob:setMod(xi.mod.EVA, 400)
```
WinterSolstice8, line 13: horizon recently had a PR that moved this evasion mod out to onMobSpawn because apparently setMod for evasion doesn't work in onMobInitialize. I guess stats are re-calculated after that is called? not sure. Maybe the next spawn has it wrong and the first one is fine

additionally, we use `mob:addImmunity(xi.immunity.SLEEP)` for immunities now iirc

I'm not a mob/combat guy typically so I will defer to others if i'm wrong

WinterSolstice8:
CI is complaining that `getMeleeRange` isn't defined. I assume that function needs to be ported from ASB too
edit: that might be in the drawin PR

CriticalXI:
That's in the draw in PR that I linked to in my first comment.

scripts/zones/Rolanberry_Fields/mobs/Simurgh.lua
Frankie-hz: Yeah when the mob spawns the stats are calculated so it uses the evasion in the mobinit as its base evasion.

You can use what winter said for immunities. Sleepres is only for traits at this moment if I’m not mistaken.

CriticalXI: Adjusting this immunity has shown me that the blanket sleep immunity has been removed. Also adjusting this for Eldritch Edge to put the immunities in lua instead of mob_pools.

WinterSolstice8: Oh yeah, it's light sleep or dark sleep (separated) now

scripts/zones/Rolanberry_Fields/mobs/Evil_Weapon.lua
```diff
@@ -19,7 +19,7 @@ entity.onMobDeath = function(mob, player, optParams)
 end
 
 entity.onMobDespawn = function(mob)
-    xi.mob.phOnDespawn(mob, eldritchPHTable, 20, math.random(5400, 7200)) -- 90 to 120 minutes
+    xi.mob.phOnDespawn(mob, eldritchPHTable, 20, 5400) -- 90 minute minimum
```
LacunaXI: 20% spawn rate sounds way too high based on Clopedia player testimonies.

scripts/zones/Rolanberry_Fields/mobs/Ravenous_Crawler.lua
```diff
@@ -9,6 +9,10 @@ entity.onMobInitialize = function(mob)
     mob:addMod(xi.mod.REGAIN, 100)
```
LacunaXI: These NMs added in the 2009 updates were a bit copy-paste. Aside from sweeping immunity to gravity bind and sleep they generally have a potent Store TP and not a regain. Players would sometimes write this as regain on the wikis but it is very easy to test by turning your back with Stoneskin at <20% HP.

CriticalXI: Have you gotten any capture data to suggest what value to put in there for that? That is the main reason I didn't adjust that.

LacunaXI: Not for this specific NM.

You can view the boost to Store TP being widespread on all these NMs added around 2009 from the many time spawns that sit idle. Skirling Liger, Chaneque, Hyakinthos, Lamina, Olgoi-Khorkhoi and so on. Ravenous Crawler would not have a large regain based on the consistency of design around the NMs from the update it was added in.

CriticalXI: Doesn't even seem like Store TP. In the Ravenous Crawler video, the crawler never really lands a hit but it's frequently using Sticky Thread. It's also obviously not regain because I don't see any of the nov 2009 NMs using a TP move on pull like a mob with regain would do. Doesn't really seem like an obvious answer here to me.

WinterSolstice8: Sticky thread might be on a timer?

LacunaXI: > Doesn't even seem like Store TP. In the Ravenous Crawler video, the crawler never really lands a hit but it's frequently using Sticky Thread. It's also obviously not regain because I don't see any of the nov 2009 NMs using a TP move on pull like a mob with regain would do. Doesn't really seem like an obvious answer here to me.

Store TP also increases the amount of TP gained by a player or monster when being hit. The behavior shown in the video is similar to that of other NMs mentioned before. It is just deceptive from the video without any controls to contrast with. Similar NMs mentioned reach 1,000+ TP without being touched in 4-5 attack rounds (depending on delay).

scripts/zones/Rolanberry_Fields/mobs/Evil_Weapon.lua
CriticalXI: Bumped it down to 10% after reading over the spawn time talk

scripts/zones/Rolanberry_Fields/mobs/Ravenous_Crawler.lua
CriticalXI: Adjusted to store tp on Ravenous Crawler and Eldritch Edge.

WinterSolstice8 (approved):

Xaver-DaRed:
Changes look good to me.

However, could you move the new `onMobDespawn` sections to be after `onMobDeath`? That's the usual order.

CriticalXI:
Got that adjusted for all the mobs.

Xaver-DaRed (approved):

PR #6588, opened 2024-12-25
Bind spells duration has a floor of 5 seconds if not fully resisted.

blueBounder:
**_I affirm:_**

This is a follow up on an earlier PR #6194 that implemented bind's random duration.
This earlier PR implemented bind's random duration with a floor of 1 second when not fully resisted, leading to instances of the spell's effect wearing off almost immediately.
[Further capture data](https://docs.google.com/spreadsheets/d/1MnIZySb1c03Ak9JHBuyaUAze76aFb0Pj6_Mvh6FbUYA/edit?pli=1&gid=0#gid=0) supports a floor for bind's duration when not fully resisted of 5 seconds.
Further discussion can be found in discord thread [here](https://discord.com/channels/933423693848260678/1180272981210046504).

1) Find a mob that is not bind immune
2) Cast bind and wait for the spell to wear off naturally
3) Verify time difference is between 5 and 60 seconds
4) Repeat many, many times.
5) Find a mob that is stun immune- verify change did not break this as well.

### DIFF
--- a/scripts/globals/spells/enfeebling_spell.lua
+++ b/scripts/globals/spells/enfeebling_spell.lua
@@ -449,8 +449,10 @@ xi.spells.enfeebling.useEnfeeblingSpell = function(caster, target, spell)
     -- STEP 5: Exceptions.
     ------------------------------
     -- Bind: Dependant on target speed.
+    -- Bind: Duration floor of 5 seconds.
     if spellEffect == xi.effect.BIND then
         potency = target:getSpeed()
+        duration = utils.clamp(duration, 5, 60)
 
     -- TODO: This is unnecesary, but, for now, we will comply with core.
     elseif spellEffect == xi.effect.SLEEP_I then


scripts/globals/spells/enfeebling_spell.lua
```diff
@@ -449,8 +449,10 @@ xi.spells.enfeebling.useEnfeeblingSpell = function(caster, target, spell)
     -- STEP 5: Exceptions.
     ------------------------------
     -- Bind: Dependant on target speed.
+    -- Bind: Duration floor of 5 seconds.
     if spellEffect == xi.effect.BIND then
         potency = target:getSpeed()
+        duration = utils.clamp(duration, 5, 60)
```
lapislosh, line 455: this probably makes more sense on line 259 which is already clamping from 1-60?

blueBounder: I went that way originally but later realized we would be skewing the duration slightly longer  by doing so. Since the result of line 259 is later multiplied by the resistance factor which will be between 0.5-1, to achieve an end result floor of 5, we'd have to clamp 259 between 10-60. That's my reasoning anyhow. Happy to clamp on 259 though if it wouldn't affect the distribution too much.

WinterSolstice8: Do we know if getting 1/2 resisted puts the floor below 5 seconds? If not, this is the appropriate place

blueBounder: In the observed data the minimum bind seen was no less than 5sec- so i believe a 1/2 resist would result in the minimum 5sec duration.

lapislosh: the linked capture data does show some binds at 3-4 seconds so I would assume the clamp would need to happen before the resistance check. the original PR data even shows a 1s bind which I presume is a 1/8th resist or something

blueBounder: The original PR #6194 was implemented based off of retail capture data and did not include any floor but did establish the mean and standard deviation. 
The purpose of this PR is to add the floor, as SE changed bind many times over the years- durations, floors, even the name of the spell at one point. The floor of 5s is based on the video capture data from toau era. In that set the shortest bind observed is 7s. It seemed more reasonable 5s would be the floor than 7.

All that said- the goal of this PR was to achieve a minimum duration for bind of 5s after resists given that no bind shorter than this was observed in toau era evidence.

WinterSolstice8: good enough for me

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #6589, opened 2024-12-26
[lua] Ensure white procs can't happen without SJ_RESTRICTION effect in Dynamis

sruon:
**_I affirm:_**

Currently the Dynamis proc system checks if the player is missing a subjob before rolling for a white proc at a 1% rate, which is the correct behavior for CoP Dynamis zones. However the logic is applied across the board, even for non-CoP zones (where white procs are not a thing on retail https://www.bg-wiki.com/ffxi/Category:Dynamis#Dynamis_Proc_System)

If the player never obtained a subjob or managed to remove their subjob through a change job glitch, they would be able to get white procs in non-CoP Dynamis zones.

The PR changes the subjob check to instead check if the player has the SJ_RESTRICTION effect applied, which in normal conditions is only possible when entering CoP Dynamis zones.

A zone check may also work instead of checking for SJ_RESTRICTION.

It's best to temporarily update the line to proc 100% of the time as the white proc rate is only 1% natively
```
-- Pre PR behavior
if player:getSubJob() == xi.job.NONE then

-- Post PR beheavior
if player:hasStatusEffect(xi.effect.SJ_RESTRICTION) then
```
```
!zone Dynamis-Bastok
# Add TE KIs to enable proc system
!addkeyitem CRIMSON_GRANULES_OF_TIME
!addkeyitem AZURE_GRANULES_OF_TIME
!addkeyitem AMBER_GRANULES_OF_TIME
!addkeyitem ALABASTER_GRANULES_OF_TIME
!addkeyitem OBSIDIAN_GRANULES_OF_TIME
# Force the SJ_RESTRICTION effect to remove your subjob
!addeffect SJ_RESTRICTION
```
Find a mob with the following job: RNG, NIN, BST, MNK, THF
Use job abilities to proc.

Result on base branch: White proc and a 100 will drop
Result with this PR applied: Red proc and no 100 will drop

### DIFF
--- a/scripts/globals/dynamis.lua
+++ b/scripts/globals/dynamis.lua
@@ -823,7 +823,7 @@ xi.dynamis.procMonster = function(mob, player)
 
         local extensions = getExtensions(player)
         if extensions > 2 then
-            if player:getSubJob() == xi.job.NONE and math.random(1, 100) == 1 then
+            if player:hasStatusEffect(xi.effect.SJ_RESTRICTION) and math.random(1, 100) == 1 then
                 mob:setLocalVar('dynamis_proc', 4)
                 mob:addStatusEffect(xi.effect.TERROR, 0, 0, 30)
                 mob:weaknessTrigger(3)


WinterSolstice8 (approved):

zach2good (approved):

sruon:
Wasn't aware of the 72 characters limit on commit messages, should be good this time.

PR #6602, opened 2024-12-28
Core: Add empty CFellowEntity class

zach2good:
**_I affirm:_**

Add empty CFellowEntity class


claywar (approved):

zach2good:
image
image

zach2good:
This logic isn't equivalent. Before these changes `isMob()` on a Trust would be false, but now it's true. Will need to audit whether or not this is good/safe.

zach2good:
Reworked this PR to add an empty CFellowEntity class type, added some const, and the one new relevant binding

PR #6609, opened 2024-12-30
[Lua, Fishing] Inside the belly convert to IF and refactor

Flibe-XI:
**_I affirm:_**

Convert quest "Inside the Belly" to IF.
- Refactored cutscene onTrigger for fish recommendations to be easier to read, eliminating long if/else chain
- Changed cutscene variations to be based on fishing level instead of player main job level
- Factored out onTrigger/onTrade into local functions
- Changed how roll and sum are calculated, such that instead of the roll being divided by 10 to match the range of the fish reward chance values, the fish reward values are multiplied by 10 to match the roll, in the range of (1, 1000).  There is no mathematical difference in the outcome odds.

!addquest 4 22 -- the real gift
!delquest 4 26 -- inside the belly
Set fishing skill to 29
!setskill 48 29
Talk to zaldon, doesn't start the quest
!additem 13808 -- fisherman's tunica
Equip tunica, (+1 to fishing skill, for a total of 30)
Talk to Zaldon, quest should start
Talk to Zaldon again, should recommend 4 fish to catch
Trade Zaldon various fish, and receive correct rewards, at correct rates


scripts/quests/otherAreas/Inside_the_Belly.lua
```diff
@@ -0,0 +1,724 @@
...
+    },
+}
+
+local function tradeFish(player, fishId)
+    quest:setLocalVar(player, 'fishId', fishId)
+    quest:setLocalVar(player, 'itemIdx', 0)
+
+    local rewards = fishRewards[fishId].items
+    local roll    = math.random(1, 1000)
+    local found   = false
```
Flibe-XI: `found` variable is unnecessary, can be deleted

zach2good (approved):
This retains the safety block of:
```lua
    -- NOTE: We confirm the trade now, and not at the end of the cutscene as normal
    --     : because the cutscene gives away whether or not the trade was successful
    --     : or not, and it's possible for players to cheese this trade by force-dc-ing.
    player:confirmTrade()
```
    
So probably fine 👍

Flibe-XI:
> This retains the safety block of:
> 
> ```lua
> 
>     -- NOTE: We confirm the trade now, and not at the end of the cutscene as normal
> 
>     --     : because the cutscene gives away whether or not the trade was successful
> 
>     --     : or not, and it's possible for players to cheese this trade by force-dc-ing.
> 
>     player:confirmTrade()
> 
> ```
> 
>     
> 
> So probably fine 👍 

I also changed the reward function to give Gil even in the event of a failed trade, because we keep the fish regardless, so it's as if the trade went through even though it fails (ie in the case of already having a rare reward, as you can't really have a full inventory), so might as well get Gil in return

claywar:
@Flibe-XI I understand that this has been merged, but can you confirm that this actually awards an item in cases where applicable?

Flibe-XI:
I tested the changes on a number of different fish.  Trading many of the same type to get a rough estimate to confirm specific success rates weren't completely bonkers.  It does still award items.

claywar:
Understood, and I could be missing something.  Could you show me where the item is given?

Flibe-XI:
> Understood, and I could be missing something. Could you show me where the item is given?

Terribly sorry about that, I must not have saved my changes on the last commit after doing some testing.  See #6670

PR #6611, opened 2024-12-30
[search] Rewrite search in ASIO

WinterSolstice8:
**_I affirm:_**

Rewrite search in ASIO

TODO: I did a lot but I didn't do everything.


src/search/search_session.cpp
```diff
@@ -0,0 +1,869 @@
...
+#include "packets/linkshell_list.h"
+#include "packets/party_list.h"
+#include "packets/search_comment.h"
+#include "packets/search_list.h"
+
+
+// A single IP should only have one request in flight at a time, so we are going to
+// be tracking the IP addresses of incoming requests and if we haven't cleared the
+// record for it - we block until its done
+extern shared_guarded<std::unordered_set<std::string>> gIPAddressesInUse;
```
zach2good: Instead of globals and externs, can we make gIPAddressesInUse a member of search_server and pass references to it around? Acess to this is shared_guarded so it's safe to access from multiple threads

src/search/search_session.cpp
```diff
@@ -0,0 +1,869 @@
...
+    socket_.lowest_layer().set_option(asio::socket_base::reuse_address(true));
+    ipAddress = socket_.lowest_layer().remote_endpoint(ec).address().to_string();
+
+    // If someone from the same IP is making a request, block for 10 ms in this thread repeatedly until the request is done.
+    // TODO: do we need a better timeout to drop the request if data never comes from a DIFFERENT socket?
+    // Someone doing this maliciously would just lock themselves out...
+    uint32_t numWaits = 0;
+    while (isIPAddressInUse(ipAddress))
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds (10));
```
zach2good: Rather than a full thread sleep, can we just use the asio countdown timer/countdown? Then you could keep the ip/numWaits as part of search server or something?

zach2good (approved):
Looking good 👍

src/search/handler.h
```diff
@@ -0,0 +1,67 @@
...
+private:
+    void do_accept()
+    {
+        // clang-format off
+        acceptor_.async_accept(
+        [this](std::error_code ec, asio::ip::tcp::socket socket)
+        {
+            if (!ec)
+            {
+                if constexpr (std::is_same_v<T, search_session>)
```
zach2good: Is there any way we can have a single, common handler across all processes, templated or something?

src/search/handler.h
```diff
@@ -0,0 +1,67 @@
...
+                {
+                    auto auth_handler = std::make_shared<T>(std::move(socket));
```
zach2good: Should this be search_handler? This is the thing that's doing shared_from_this() all the way down the chain right?

src/search/handler.h
WinterSolstice8: Without using extern globals for the IP address whitelist and such? Not entirely sure it would look clean, but possible

src/search/handler.h
WinterSolstice8: probably

src/search/search_session.cpp
WinterSolstice8: Is that functionally different? clueless.png

WinterSolstice8: Regardless, done. used an asio steady timer

src/search/handler.h
WinterSolstice8: done

WinterSolstice8:
Tested just about everything I could, it all works

PR #6613, opened 2024-12-30
[lua] Rename all incorrectly named Mamool Ja Scripts so they work

WinterSolstice8:
**_I affirm:_**

Rename all incorrectly named Mamool Ja Scripts so they work

See weapon break mixin be applied


claywar (approved):

WinterSolstice8:
Just a note here for future searches -- The renaming was only needed here because I run Linux and filenames are case sensitive. On windows this wouldn't have been noticeable

PR #6619, opened 2024-12-31
Search: Minor cleanup and refactoring of handler<T>

zach2good:
**_I affirm:_**

Just a little tidying

### DIFF
--- a/src/search/CMakeLists.txt
+++ b/src/search/CMakeLists.txt
@@ -6,11 +6,11 @@ set(SOURCES
     data_loader.cpp
     data_loader.h
     handler.h
-        search_handler.h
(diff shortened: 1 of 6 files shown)


zach2good:
@WinterSolstice8 If we construct `asio::io_context io_context;` with no arguments, is it just single-threaded on the calling thread that starts it? ie. Do we even need to bother guarding stuff?

zach2good:
Ah yes, it just blocks the calling thread, so it looks like we can do away with any of the locking unless we want to introduce more threads through asio::thread_pool or manually and post io_context.run() to them, etc.

WinterSolstice8 (approved):

Issue #6610, opened 2024-12-30
🐛 Player speed not reflecting Settings


FFXILevelDown:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base`

## Steps to reproduce
Updated to base and speed does not reflect what is in settings anymore. It defaults player speed to retail regardless of settings. Tried to alter speed in settings, reboot, rebuild etc and no change. 
## Expected behavior

Unless this was intentional, this should allow adjustment of speed thru SPEED_MOD in map.lua settings. The !setspeed command still works :).

Mortalelite:
I also tested this and was no change to speed_mod https://github.com/LandSandBoat/server/blob/base/settings/default/map.lua#L109 using Positive or Negative results. Still ran same speed unless you used the command.

GeneralLD:
Just pulled the update to test this out. The speed adjustment worked however it is not affecting the character animation. I move across the screen at the modded speed set in settings, but the character animation is slower than retail speed. Tested with three different servers up, one with the fix, one before speed adjustment was originally implemented in PR and one without fix but using the speed adjustment from PR and the animation is noticeable.

PR #6631, opened 2025-01-01
Speed settings overhaul

cocosolos:
**_I affirm:_**

Split the single movement speed setting into multiple settings to improve ease of use and customizability. Base speed, speed limit, and mount speed are all now directly set instead of an offset to retail values.

Animation speed setting added that controls player movement animation speed. This is a little hard to explain. Take your current speed and divide it by your base speed. Normally this would be 50/50=1.0, this is the rate at which your running animation plays, 1x. Flee brings you to 80/50=1.6x animation speed. If you increase base speed to 80, then you still have a 1x animation speed. This setting is applied as a divisor to the base speed, so if you want to increase the base movement speed and also increase animation speed, you would increase this value (e.g. to make it always appear as if the player has a flee effect you would set BASE_SPEED to 80 and ANIMATION_SPEED_DIVISOR to 1.6). Will fix an additional issue brought up in #6610.

Speedsub was changed to animationSpeed to reflect this, and baseSpeed was added to track the base speed of an entity. For most entities these will be the same, but allows for player animation adjustments. This was left as speedsub in the DB so as to not require any changes to capturing tools.

Let's see if I can break this down a little more. What LSB and capture tools call `speedsub`, would be what Windower and Ashita refer to as "animation speed" (or SpeedBase in XiPackets documentation.) However, this is not necessarily the base movement speed. Other than player customization mentioned above, I have noticed with trusts on retail that they have an animation speed of 50 and normally a speed of 155. However, if you do some maneuvering and kind of kite them then run back into them you can get them to stop "running" and see their speed change to 62. This fits with the recent mob run speed changes where `62 * 2.5 = 155`. So we can see that trusts have a base speed of 62, run speed of 155 (normal 2.5x modifier), and 50 animation speed ("speedsub"). Here's a short clip demonstrating this:

https://github.com/user-attachments/assets/d06fe49f-735b-4bda-aa88-735c51f949fa

Side note, if you look at some old videos ([example1](https://www.youtube.com/watch?v=DBKbGXc7ZDk) [example2](https://youtu.be/8f9wpRTzHow?t=129)) the players appear to be moving with the same animation speed as current retail. With these changes, you can set BASE_SPEED to 40 and leave ANIMATION_SPEED_DIVISOR at 1.0 to achieve these results, whereas before you would be animating at 40/50=0.8x animation speed, making it appear as if your movement is slowed down (which it is compared to current retail, but doesn't appear to have been the case when the movement speed actually was 40.)

Play around with changing these settings:

```lua
xi.settings.map =
{
    BASE_SPEED = 80,
    SPEED_LIMIT = 255,
    MOUNT_SPEED = 255,
    ANIMATION_SPEED_DIVISOR = 2.0,
}
```

### DIFF
--- a/settings/default/map.lua
+++ b/settings/default/map.lua
@@ -105,12 +105,18 @@ xi.settings.map =
     -- Enable/disable jobs other than BST and RNG having widescan
     ALL_JOBS_WIDESCAN = true,
 
-    -- Modifier to apply to player speed. 0 is the retail accurate default. Negative numbers will reduce it.
-    SPEED_MOD = 0,
+    -- Base player movement speed
+    BASE_SPEED = 50,
 
-    -- Modifier to apply to mount speed. 0 is the retail accurate default. Negative numbers will reduce it.
-    -- Note retail treats the mounted speed as double what it actually is.
-    MOUNT_SPEED_MOD = 0,
+    -- Player movement speed limit
+    SPEED_LIMIT = 80,
+
+    -- Mount speed, expressed as player speed. Can surpass speed limit.
+    MOUNT_SPEED = 80,
+
+    -- Player animation speed divisor
+    -- Raising this increases the players movement animation speed
+    ANIMATION_SPEED_DIVISOR = 1.0,
 
     -- Multiplier for speed of engaged mobs when their target is out of range.
     -- The default for almost all mobs on retail is 2.5x their normal speed.
--- a/src/map/entities/baseentity.cpp
+++ b/src/map/entities/baseentity.cpp
@@ -39,7 +39,7 @@ CBaseEntity::CBaseEntity()
 , m_TargID(0)
 , animation(0)
 , animationsub(0)
-, speedsub(50 + settings::get<int8>("map.SPEED_MOD"))
+, baseSpeed(settings::get<uint8>("map.BASE_SPEED"))
 , namevis(0)
 , allegiance(ALLEGIANCE_TYPE::MOB)
 , updatemask(0)
@@ -53,7 +53,8 @@ CBaseEntity::CBaseEntity()
 , m_nextUpdateTimer(std::chrono::steady_clock::now())
 {
     TracyZoneScoped;
-    speed = speedsub;
+    speed          = baseSpeed;
+    animationSpeed = static_cast<uint8>(std::clamp<float>((baseSpeed / settings::get<float>("map.ANIMATION_SPEED_DIVISOR")), std::numeric_limits<uint8>::min(), std::numeric_limits<uint8>::max()));
 }
 
 CBaseEntity::~CBaseEntity()
--- a/src/map/entities/baseentity.h
+++ b/src/map/entities/baseentity.h
@@ -303,20 +303,21 @@ public:
 
     bool IsDynamicEntity() const;
 
-    uint32          id;           // global identifier unique on the server
-    uint16          targid;       // local identifier unique to the zone
-    ENTITYTYPE      objtype;      // Type of entity
-    STATUS_TYPE     status;       // Entity status (different entities - different statuses)
-    uint16          m_TargID;     // the targid of the object the entity is looking at
-    std::string     name;         // Entity name
-    std::string     packetName;   // Used to override name when being sent to the client
-    look_t          look;         //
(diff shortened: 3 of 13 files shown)


WinterSolstice8:
There is a bastok mission CS (i'm not sure which one, an early one) in Metalworks where there's 2-3 NPCs walking down the stairs from the president's office. On Horizon it looked really jank due to speedsub and they were "stutter stepping" a bit.

Does this PR fix that? I'll see if I cant find the CS id for you.

edit: this was a side effect of them changing it back to era speed if that wasn't obvious

WinterSolstice8:
Here's an example on current `base` because I think Naji has the wrong base speed (40?) "wrong", maybe it's still 40 on a cap on retail. not sure.

in metalworks, cs ID 713; Naji looks jank walking up the stairs and is stutter stepping but YOU don't

cocosolos:
> There is a bastok mission CS (i'm not sure which one, an early one) in Metalworks where there's 2-3 NPCs walking down the stairs from the president's office. On Horizon it looked really jank due to speedsub and they were "stutter stepping" a bit.
> 
> Does this PR fix that? I'll see if I cant find the CS id for you.
> 
> edit: this was a side effect of them changing it back to era speed if that wasn't obvious

I'm pretty sure that issue always happens, even at 50 speed. None of these speed settings should have been affecting NPCs AFAIK since they're all set from the DB and all mostly have 40 speed anyway I think. I know exactly what you're talking about too because I've gone through Bastok missions and that part in particular where they're walking down the steps at the President's place was really jarring. I doubt this fixes that since this should only really have an affect on players, trusts, and dynamic entities that don't have their speed explicitly set.

WinterSolstice8:
Makes me wonder if it's bugged on retail... hmm...

cocosolos:
> Makes me wonder if it's bugged on retail... hmm...

I wanna say no, but I can't remember if/when I actually tested that. I'm sure there's a capture though. It kind of reminds me of issues I saw when the `moving` field in pathfinding was wrong, but I'm not sure if/how CS NPCs use that.

WinterSolstice8:
I took this just now. He moves normally.
[2025-01-01 11-39-49.webm](https://github.com/user-attachments/assets/645286d0-c752-473a-aa07-82879dc3afed)

WinterSolstice8 (approved):

Xaver-DaRed (approved):

cocosolos:
@WinterSolstice8 Okay this is kind of funny. Check out [this capture](https://drive.google.com/open?id=1NEl7FbisvotfKl8Ab6JqV3t3KfHuq6WG). In the NPCLogger Metalworks table:

```lua
    [17748013] = {['id']=17748013, ['name']="Naji", ['polutils_name']="Naji", ['npc_type']="Equipped NPC", ['index']=45, ['x']=66.865, ['y']=-13.999, ['z']=-4.562, ['r']=128, ['flag']=14, ['speed']=50, ['speedsub']=50, ['animation']=0, ['animationsub']=1, ['namevis']=0, ['status']=0, ['flags']=27, ['name_prefix']=32, ['look']="0100050100101B2000300C401850006000700000", ['raw_packet']="0E2609002DD00E012D005780E1BA8542E7FB5FC1E7FB91C00E00000032326400001B00000000002000000100000000000100050100101B2000300C4018500060007000000000000000000000"},
```

`['speed']=50, ['speedsub']=50`

LSB:

```sql
INSERT INTO `npc_list` VALUES (17748013,'Naji','Naji',128,66.865,-13.999,-4.562,14,40,40,0,1,0,0,27,0x0100050100101B2000300C401850006000700000,32,NULL,1);
```

`40,40`

What happens when you change them to 50?



https://github.com/user-attachments/assets/d1aa24b0-d85f-42d0-aca6-ed8747200aeb

WinterSolstice8:
Interesting. Maybe all NPCs were shifted to 50 when players were?

cocosolos:
Uh wow yeah it looks like most NPCs are all 50/50 now. Raminel is 40/50 and his movement is broken on retail but works fine on LSB since he's still 40/40 aka 1.0x animation speed lmao.

Xaver-DaRed:
#6632

PR #6639, opened 2025-01-02
[ci] Add math.random bounds checking

claywar:
**_I affirm:_**

* Adds CI rule for lua_stylecheck to support #6407
* Adds lower and/or upper bounds to remaining math.random() calls.

NOTE: Lua's math.random function is [0,1) when using no parameters.  In cases like this, any < check has been moved to <=

CI will fail until issues are resolved in modified Lua scripts

### DIFF
--- a/tools/ci/tests/stylecheck.lua
+++ b/tools/ci/tests/stylecheck.lua
@@ -240,3 +240,9 @@ x ==   y -- FAIL
 x  ~= y -- FAIL
 
 x = { x  + y } -- PASS
+
+-- math.random boundaries
+math.random(1, 100) -- PASS
+math.random(1) -- FAIL
+math.random() -- FAIL
+math.random(1, 2, 3) -- FAIL
--- a/scripts/zones/Arrapago_Remnants/npcs/Armoury_Crate.lua
+++ b/scripts/zones/Arrapago_Remnants/npcs/Armoury_Crate.lua
@@ -16,8 +16,8 @@ entity.onTrigger = function(player, npc)
     local second = { 5366, 5368, 5369, 5370, 5372, 5376, 5377, 5378, 5379, 5380, 5381, 5382 }
 
     player:addTreasure(5365) player:addTreasure(5365) player:addTreasure(5373) player:addTreasure(5375)
-    player:addTreasure(5374) player:addTreasure(first[math.random(#first)]) player:addTreasure(first[math.random(#first)])
-    player:addTreasure(second[math.random(#second)]) player:addTreasure(second[math.random(#second)])
+    player:addTreasure(5374) player:addTreasure(first[math.random(1, #first)]) player:addTreasure(first[math.random(1, #first)])
+    player:addTreasure(second[math.random(1, #second)]) player:addTreasure(second[math.random(1, #second)])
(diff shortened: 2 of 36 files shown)


zach2good (approved):

Xaver-DaRed (approved):
Like it

Flibe-XI:
What about occourances of `math.random() * math.pi`?

WinterSolstice8:
> What about occourances of `math.random() * math.pi`?

Those should fall under the floating point specialization of xirand -- https://github.com/LandSandBoat/server/blob/fd1983ec94e05315b5a776f3f27381d628d0e307/src/common/xirand.h#L88

It's not effected by the same bounds issues as integers

WinterSolstice8:
You can read more at the cpp ref:

https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution
vs
https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution

WinterSolstice8:
```        // Bind math.randon(...) globally
        // clang-format off
        lua["math"]["random"] =
            sol::overload([]() { return xirand::GetRandomNumber(1.0f); },
                          [](int n) { return xirand::GetRandomNumber<int>(1, n + 1); },
                          [](float n) { return xirand::GetRandomNumber<float>(0.0f, n); },
                          [](int n, int m) { return xirand::GetRandomNumber<int>(n, m + 1); },
                          [](float n, float m) { return xirand::GetRandomNumber<float>(n, m); });
```

`math.random()` is using the floating point specialization

tools/ci/tests/stylecheck.lua
```diff
@@ -240,3 +240,9 @@ x ==   y -- FAIL
 x  ~= y -- FAIL
 
 x = { x  + y } -- PASS
+
+-- math.random boundaries
+math.random(1, 100) -- PASS
+math.random(1) -- FAIL
+math.random() -- FAIL
```
Flibe-XI, line 247: Won't this catch the occurrences of `math.pi * math.random()`? @WinterSolstice8

WinterSolstice8: I see what you mean, yeah

WinterSolstice8: `math.random(0.0, 1.0) * math.pi` is equivalent

Flibe-XI: These seem like things that are intentionally floats
[sickle slash](https://github.com/LandSandBoat/server/blob/base/scripts/actions/mobskills/sickle_slash.lua#L27)
[EES](https://github.com/LandSandBoat/server/blob/base/scripts/actions/mobskills/eagle_eye_shot.lua#L14)
[taru warp](https://github.com/LandSandBoat/server/blob/base/scripts/actions/mobskills/tarutaru_warp_ii.lua#L14)

Should they be converted to `math.random(float_n, float_m)`?

edit: I was typing, missed your two comments above before posting

WinterSolstice8: Those are easy converts, but I feel like EES and sickle slash aren't correct, that's a different problem however

PR #6640, opened 2025-01-02
[search] Wait for replies from client before sending more data

WinterSolstice8:
**_I affirm:_**

Apparently we are supposed to expect a reply and to wait for it. Otherwise bad things happen. (see #6633)

search linkshell, party, /sea all, search comment, AH item list, AH sale list, AH detailed sale list on 2+ clients at a time with no errors on _windows_ and linux

### DIFF
--- a/src/search/search.h
+++ b/src/search/search.h
@@ -40,4 +40,38 @@ struct search_req
     uint8       commentType;
 };
 
+class searchPacket
+{
+public:
+    // max size of search packet is 1024 in packets
+    static constexpr uint16_t max_size = 1024;
+
+    searchPacket(uint8_t* buffer, uint16_t length)
+    {
+        if (length > max_size)
+        {
+            size = 0;
+            ShowError(fmt::format("Error: search packet with size above {} requested!", max_size));
+            return;
+        }
+
+        std::memcpy(buff_.data(), buffer, length);
+        size = length;
+    }
+
+    uint16_t getSize()
+    {
+        return size;
+    }
+
+    uint8_t* getData()
+    {
+        return buff_.data();
+    }
+private:
+
+    std::array<uint8_t, max_size> buff_;
+    uint16_t                      size;
+};
+
 #endif
--- a/src/search/search_handler.cpp
+++ b/src/search/search_handler.cpp
@@ -71,10 +71,10 @@ void search_handler::start()
 {
     if (socket_.lowest_layer().is_open())
     {
-        do_read();
-
-        deadline_.expires_after(std::chrono::milliseconds(2000));
+        deadline_.expires_after(std::chrono::milliseconds(10000)); // AH searches can take quite a while
         deadline_.async_wait(std::bind(&search_handler::checkDeadline, this));
+
+        do_read();
     }
 }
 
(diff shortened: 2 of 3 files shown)


claywar (approved):

src/search/search_handler.cpp
```diff
@@ -210,6 +225,13 @@ inline std::string searchTypeToString(uint8 type)
 
 void search_handler::read_func(uint16_t length)
 {
+    // if we already have a query in-flight...
+    if (!searchPackets.empty())
+    {
+        do_write();
+        return;
```
StabbyCutyou, line 232: Not a c guy, but shouldn't the method return a code to indicate it was unable to perform a read and instead needed to defer to write the awaiting packets to the socket (or otherwise could not perform what was asked of it)?

Would it be more ideal to make that the responsibility of the caller?

* Try to read
* Get told you can't due to x
* Resolve x however you need to
* Try to read again

zach2good (approved):

src/search/search_handler.cpp
WinterSolstice8: Error handling is done in the
`search_handler::do_write()` or `search_handler::do_write()` lambda. If it hits `search_handler::read_func()` that means there is NO error, which is either 

1) The clients initial request
2) The client's reply to our sending of search data

src/search/search_handler.cpp
```diff
@@ -90,26 +90,41 @@ void search_handler::do_read()
...
+void search_handler::do_write()
 {
+    auto packet = searchPackets.front();
+    auto length = packet.getSize();
+
+    std::memcpy(data_, packet.getData(), packet.getSize());
+
+    searchPackets.pop_front();
+
+    encrypt(length);
```
zach2good, line 113: Not relevant for this PR, but as future points: Should we:
- make `searchPacket` more abstract as `OwnedSpan` or similar
- make `searchPacket` more of it's own thing, where it can encrypt itself and we then just stuff it into the buffer, etc.
- Rename it to `SearchPacket`
- We now have `data_` and `searchPackets` - is this necessary?

src/search/search_handler.cpp
```diff
@@ -71,10 +71,10 @@ void search_handler::start()
 {
     if (socket_.lowest_layer().is_open())
     {
-        do_read();
-
-        deadline_.expires_after(std::chrono::milliseconds(2000));
+        deadline_.expires_after(std::chrono::milliseconds(10000)); // AH searches can take quite a while
```
zach2good, line 74: While I'm looking at this - what is this actually the deadline for? C2S packet, db access, S2C packet?

src/search/search.h
```diff
@@ -40,4 +40,38 @@ struct search_req
     uint8       commentType;
 };
 
+class searchPacket
```
zach2good, line 43: In the future, we might want to look at making this more abstract, or more specialised.

src/search/search_handler.h
```diff
@@ -110,4 +110,6 @@ class search_handler
     auto _HandleSearchRequest() -> search_req;
 
     blowfish_t blowfish;
+
+    std::deque<searchPacket> searchPackets;
```
zach2good, line 114: Under what circumstances will we end up with more than a couple of packets in here?

zach2good (approved):

src/search/search_handler.cpp
WinterSolstice8: ```We now have data_ and searchPackets - is this necessary?```
I suppose no, if we boost the internal buffer size of searchPackets to 4096 we could encrypt them ahead of time.

src/search/search_handler.cpp
WinterSolstice8: It was originally intended as a way to close a hung socket, but now it serves a few more purposes

If the query simply takes too long in general it will die, though that was not intended. There are several points where we could be waiting forever on a socket and we don't want to due to malicious actors.

I suppose we may want to repurpose it into the _total lifespan_ of a single request, with a shorter timeout for the original open.

Open -> few hundred MS waiting on arrival of first data packet -> close if it never comes
Open -> few hundred MS waiting on arrival of first data packet -> get request-> send reply -> we are now waiting _again_ for a reply that could never come. Repeat send reply -> get request a few times and this could be waiting a long time with someone with a malicious attack.

src/search/search.h
WinterSolstice8: agreed, it's probably useful elsewhere

src/search/search_handler.h
WinterSolstice8: AH list item packets are 20~ bytes and there are several hundred items in most equipment categories. So probably a good chunk of the time.

WinterSolstice8: ```    uint8 PacketsCount = (uint8)((ItemList.size() / 20) + (ItemList.size() % 20 != 0) + (ItemList.empty()));```

WinterSolstice8: Ah, maybe it is 20 _items_ per chunk. Still a good amount.

PR #6657, opened 2025-01-05
Fix TakeSwipeLungeDamage, CalculateBaseTP parameter types

claywar:
**_I affirm:_**

* TakeSwipeLungeDamage should be CBattleEntity type for Attacker, changed from CCharEntity.
* CalculateBaseTP had signature mismatch between defined/declared (Declared: int, Defined: int32) for delay.  Changed all to int32.
* Removes declaration for undefined function: CheckMobMultiHits

* No errors with mob type calling TakeSwipeLungeDamage should occur
* No other impact to gameplay

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -14770,8 +14770,8 @@ int32 CLuaBaseEntity::takeSwipeLungeDamage(CLuaBaseEntity* caster, int32 damage,
         return 0;
     }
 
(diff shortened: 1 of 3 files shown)


WinterSolstice8 (approved):

BoundingS:
Claywar, are you seeing the issue from low level mobs healing high level characters in discussion 6660 since the battleutils changes?

BoundingS:
You used a level 60 sand lizard - would you try with a low level mob please?  :)

PR #6671, opened 2025-01-07
[Combat] Expand status-effect table utility and apply usage to Weaponskills

Xaver-DaRed:
**_I affirm:_**

- Adds functionality to the statuseffect table.
- Applies that functionality to all effect-inducing weaponskill scripts which we have.
- Applies effect nullification from "incompatible effects" checks to spells.

This work is made towards the following objectives:
- Introducing "status effect resistance ranks" in small chunks while also not breaking our current functionality.
- Expanding our system to make mobskills and weaponskills check for immunities, resist triggers and effects that are nullified/nullify other effects, properly.

- Cast drown, shock and the other elemental DoT spells and have them be nullified as they should.
- Use effect-inducing weaponskills without them erroring.

### DIFF
--- a/scripts/globals/combat/status_effect_tables.lua
+++ b/scripts/globals/combat/status_effect_tables.lua
@@ -1,68 +1,149 @@
 -----------------------------------
 -- Table defining diferent status effect properties.
 -----------------------------------
+-- Info
+-- Resistance:  https://wiki-ffo-jp.translate.goog/html/1801.html?_x_tr_sl=ja&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=sc
+-- Effects:     https://wiki-ffo-jp.translate.goog/html/1720.html?_x_tr_sl=ja&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=sc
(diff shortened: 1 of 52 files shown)


scripts/globals/combat/status_effect_tables.lua
```diff
@@ -113,24 +189,33 @@ xi.combat.statusEffect.isTargetImmune = function(target, effectId, actionElement
...
-    if resistancePower == 0 then
+    local resistancePower = target:getMod(modifierId) + target:getMod(xi.mod.STATUSRES) + 5
+    if resistancePower <= 5 then
         return false
     end
 
-    local roll      = math.random(1, 100)
-    resistancePower = resistancePower + 5
-
     if actor:isNM() then
```
WinterSolstice8, line 204: We might want to add a TODO here, this isn't always true (resist terror from Sagasinger is not modified by NM)

WinterSolstice8 (approved):

scripts/globals/combat/status_effect_tables.lua
Xaver-DaRed: I wonder if that is something global for resistance mods from gear or a case by case thing.

WinterSolstice8 (approved):

PR #6674, opened 2025-01-08
[core] Misc fixes

WinterSolstice8:
**_I affirm:_**

Clamp lower end of auto attack damage to 0.
Only check roaming casts for light spirit,  this fixes non-light spirits just sitting there and not following you.

Summon non-light spirit, see it follow you
Get attacked by ultra low level mobs with >60~ VIT and never see a heal from an auto attack

### DIFF
--- a/src/map/ai/controllers/pet_controller.cpp
+++ b/src/map/ai/controllers/pet_controller.cpp
@@ -97,7 +97,7 @@ void CPetController::DoRoamTick(time_point tick)
         {
             return;
         }
-        else if (PetEntity->m_PetID <= PETID_DARKSPIRIT)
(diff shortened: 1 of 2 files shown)


zach2good (approved):

Xaver-DaRed (approved):

BoundingS:
The fix to return 0 damage if less than zero worked for the auto attack hits, but Qoofim is still bugging his enwater damage.  I noticed, it seems to be returning about 10% of Qoofim's damage as enwater damage if the melee hit damage was around 30 and up.  Around 20 damage, it started returning very large negatives like I posted in the bug, always around 131,000 HP recovered.  Is there a calculation for the enspell damage from mobs that also needs a 0 floor?

PR #6675, opened 2025-01-09
[feat] Adds A Chocobo Riding Game

CriticalXI:
**_I affirm:_**

- Adds [A Chocobo Riding Game](https://ffxiclopedia.fandom.com/wiki/A_Chocobo_Riding_Game)
- Adds functionality to three gate glyph teleport scrolls awarded from the race.

Captures are being tracked [here](https://docs.google.com/spreadsheets/d/1Kvb8IUd0ob9jqI3s2nuNEZwo_GtHE92bh91CySJviVI/edit?gid=0#gid=0)

1. Speak to any starter city or kazham chocobo NPC to begin a race.
2. Either manually navigate to the destination or !zone
3. Move to the gate for the destination zone and see a cutscene with a reward associated with the race you did

### DIFF
--- a/scripts/effects/mounted.lua
+++ b/scripts/effects/mounted.lua
@@ -23,6 +23,11 @@ end
 
 effectObject.onEffectLose = function(target, effect)
     target:setAnimation(xi.anim.NONE)
+
+    -- Remove CharVars from player participating in chocobo riding game
+    if target:isPC() then
+        xi.chocoboGame.dismountChoco(target)
+    end
 end
 
 return effectObject
--- a/scripts/globals/chocobo.lua
+++ b/scripts/globals/chocobo.lua
@@ -125,10 +125,11 @@ xi.chocobo.renterOnTrade = function(player, npc, trade, eventSucceed, eventFail)
     end
 end
 
-xi.chocobo.renterOnTrigger = function(player, eventSucceed, eventFail)
-    local mLvl   = player:getMainLvl()
-    local zoneId = player:getZoneID()
-    local info   = chocoboInfo[zoneId]
+xi.chocobo.renterOnTrigger = function(player, npc, eventSucceed, eventFail)
+    local mLvl      = player:getMainLvl()
+    local zoneId    = player:getZoneID()
+    local info      = chocoboInfo[zoneId]
+    local canRace   = xi.chocoboGame.raceCheck(player, npc)
 
     if info then
         if
@@ -136,19 +137,23 @@ xi.chocobo.renterOnTrigger = function(player, eventSucceed, eventFail)
             mLvl >= info.levelReq and
             (player:hasCompletedMission(xi.mission.log_id.WOTG, xi.mission.id.wotg.BACK_TO_THE_BEGINNING) or not info.past)
         then
-            local price = getPrice(zoneId, info)
-            player:setLocalVar('[CHOCOBO]price', price)
-
-            local currency = 0
-            if info.past then
-                currency = player:getCurrency('allied_notes')
+            if canRace then -- Check if NPC can start A Chocobo Riding Game
+                xi.chocoboGame.startRaceEvent(player, canRace, eventSucceed)
             else
-                currency = player:getGil()
+                local price = getPrice(zoneId, info)
+                player:setLocalVar('[CHOCOBO]price', price)
+
+                local currency = 0
+                if info.past then
+                    currency = player:getCurrency('allied_notes')
+                else
+                    currency = player:getGil()
+                end
+
+                local lowLevel = (mLvl < 20) and 1 or 0
+
+                player:startEvent(eventSucceed, price, currency, lowLevel)
             end
-
-            local lowLevel = (mLvl < 20) and 1 or 0
-
-            player:startEvent(eventSucceed, price, currency, lowLevel)
         else
             player:startEvent(eventFail)
         end
@@ -158,6 +163,8 @@ xi.chocobo.renterOnTrigger = function(player, eventSucceed, eventFail)
 end
 
 xi.chocobo.renterOnEventFinish = function(player, csid, option, eventSucceed)
+    local chocoGame = player:getLocalVar('tempDestCity')
+
     if csid == eventSucceed and option == 0 then
         local mLvl     = player:getMainLvl()
         local zoneId   = player:getZoneID()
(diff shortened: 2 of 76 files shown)


CriticalXI:
Sorry its a beefy PR 😓  Please let me know if I went about anything the wrong way, happy to tear the whole thing apart if I should have done this a different way.

I put this in a draft because I still need more captures to fully verify the race times. I also need to verify the windurst woods gate glyph teleport location.

Last thing todo still that I was hoping for some help with is to add a vanadiel day check to rotate between the different races in each starter city. I need to follow https://www.mithrapride.org/vana_time/chocoracetable.html and I wasn't quite sure how to accomplish this.

settings/default/main.lua
```diff
@@ -31,11 +31,11 @@ xi.settings.main =
...
-    ENABLE_ABYSSEA   = 1,
-    ENABLE_SOA       = 1,
-    ENABLE_ROV       = 1,
-    ENABLE_TVR       = 1,
-    ENABLE_VOIDWATCH = 1, -- Not an expansion, but has its own storyline. (Not Implemented)
+    ENABLE_ABYSSEA   = 0,
+    ENABLE_SOA       = 0,
+    ENABLE_ROV       = 0,
+    ENABLE_TVR       = 0,
+    ENABLE_VOIDWATCH = 0, -- Not an expansion, but has its own storyline. (Not Implemented)
```
Xaver-DaRed: Revert all changes to this file, please

scripts/globals/chocobo.lua
```diff
@@ -180,7 +187,13 @@ xi.chocobo.renterOnEventFinish = function(player, csid, option, eventSucceed)
             local price = player:getLocalVar('[CHOCOBO]price')
             player:setLocalVar('[CHOCOBO]price', 0)
 
-            if
+            if mLvl >= 20 then
+                duration = 1800 + (player:getMod(xi.mod.CHOCOBO_RIDING_TIME) * 60)
+            end
+
+            if chocoGame then -- Start A Chocobo Riding Game
```
claywar: Numeric values are not falsey on 0, so this will always default to true

scripts/globals/chocobo_riding_game.lua
```diff
@@ -0,0 +1,209 @@
...
+
+-- Rewards that all races give
+local defaultRewards =
+{
+    xi.item.CHOCOBO_TICKET,
+    xi.item.BUNCH_OF_GYSAHL_GREENS
+}
+
+-- Checks if the NPC can start a race and the player hasn't already participated this week
+xi.chocoboGame.raceCheck = function(player, npc, zone)
```
claywar: Given player and npc, there's probably not a good reason to need to pass zone into this function, as it can be derived

scripts/globals/chocobo_riding_game.lua
```diff
@@ -0,0 +1,209 @@
...
+        then
+            return destination
+        end
+    end
+
+    return false
+end
+
+-- Get the event param to tell the player which city race they are offering
+xi.chocoboGame.renterOnTrigger = function(player, destination, eventSucceed)
```
claywar: With this and onEventFinish, I'd almost prefer they're named something different if non-conforming to standard functions (onTrigger usually implies player, npc as args, whereas onEventFinish is player, csid, option, npc).

scripts/globals/chocobo_riding_game.lua
```diff
@@ -0,0 +1,209 @@
...
+        local recordTime = GetServerVariable('[ChocoGame][RecordTime]'..startingCity..'+'..destCity)
+
+        -- Parse record name
+        local recordName = ""
+        if recordId ~= 0 then
+            recordName = GetPlayerByID(recordId):getName()
+        end
+
+        -- Race complete event if a record for this path exists
+        if recordName then
```
claywar, line 151: As previously, might want to confirm that an empty string is interpreted as false

settings/default/main.lua
CriticalXI: doh, got that removed

scripts/globals/chocobo_riding_game.lua
CriticalXI: Renaming them to startRaceEvent and beginRace

scripts/globals/chocobo_riding_game.lua
CriticalXI: My testing showed this working correctly, the messaging from the NPC changed on my test depending on if a record existed or not

scripts/globals/chocobo_riding_game.lua
CriticalXI: Got that adjusted to get the zone from the player

scripts/globals/chocobo_riding_game.lua
```diff
@@ -0,0 +1,209 @@
...
+-- Effect is called when mounted effect is lost through scripts\effects\mounted.lua
+xi.chocoboGame.dismountChoco = function(player)
+    if player:getCharVar('[ChocoGame]StartTime') then
+        player:setCharVar('[ChocoGame]StartTime', 0)
+        player:setCharVar('[ChocoGame]StartingCity', 0)
+        player:setCharVar('[ChocoGame]DestCity', 0)
+    end
+end
+
+xi.chocoboGame.handleMessage = function(player, zoneIds)
```
Xaver-DaRed: Feeding player, you dont need to feed the zoneIds nor zone IDs.

```lua
xi.chocoboGame.handleMessage = function(player)
    local ID = zones[player:getZoneID()]
    ...
        player:messageSpecial(ID.text.TIME_ELAPSED, vanaHour, earthMin, earthSec)
    end
end
```

Xaver-DaRed:
I think I would separate the teleport items logic into a separate PR, considering they are "unrelated" enough and quite easy to immediately approve.
6 files are 6 files, after all.

CriticalXI:
Got that split out into #6676

CriticalXI:
I was able to figure out a solution to the logic for determining when which race is available. I think this PR should be good to go on the code itself from what I can see. I'll leave it up to you guys on how complete you want the captures to be, or if you want me to put an experimental toggle on it somehow?

scripts/effects/mounted.lua
```diff
@@ -23,6 +23,11 @@ end
 
 effectObject.onEffectLose = function(target, effect)
     target:setAnimation(xi.anim.NONE)
+
+    -- Remove CharVars from player participating in chocobo riding game
+    if target:isPC() then
```
zach2good, line 28: Could mounted ever apply to a non-PC?

zach2good (approved):

scripts/effects/mounted.lua
WinterSolstice8: 0x00E supports it, technically. There are mounted goblins in events, too. tldr maybe

CriticalXI: Happy to remove it if needed, just was playing it safe in case there was a situation I wasn't familiar with in retail.

PR #6678, opened 2025-01-09
[lua] Adjust mob add effect to account for edge cases

WinterSolstice8:
**_I affirm:_**

No more healing for 4 trillion hp, see comments in code

Get Qoofim to auto attack you until it damages you for < 50, not heals you for 4 trillion


WinterSolstice8:
Apparently this is in line with item add effects. Item add effects can only generate a bonus, and no penalty. So at least we're consistent after this PR

Xaver-DaRed (approved):

zach2good (approved):

PR #6679, opened 2025-01-09
Set the proper elements for Level 4 Light and Darkness skillchains

crallen:
**_I affirm:_**

Correctly sets the elements associated with Level 4 Light and Level 4 Darkness skillchains. This fixes an issue where magic bursts would not work correctly on these skillchains, and trusts (i.e. Shantotto II) would not attempt a magic burst on these skillchains.

1. Summon a trust that performs magic bursts, like Shantotto II.
2. Perform a skillchain that results in Level 4 Light or Darkness, Empyrean weapon skills can do this easily (Ukko's Fury -> Ukko's Fury, Tachi: Fudo -> Tachi: Fudo, etc)
3. The trust should execute a magic burst with a corresponding element.


Xaver-DaRed (approved):

WinterSolstice8 (approved):
nice find, thanks

PR #6681, opened 2025-01-09
[lua, sql] Yhoator Jungle NM Audit

CriticalXI:
**_I affirm:_**

- Adds a new lua binding setMaxMP to...set the max MP of a mob
- Edacious Opo-opo now has a 15 minute delay to respawn. Also adds latent effect to Nanban Kariginu dropped from this NM based off of my own capture below:
nanban_test
- Moves mob immunities from mob_pools to the mob luas
- NM droplists adjusted to the closest TH tier based off of ffxidb/ffxiclopedia reports

Woodland Sage Capture: https://www.youtube.com/watch?v=YJpvTYrfWDo

Getting captures for these changes is what led me down the chocobo riding game rabbit hole getting the prompt for the game getting a chocobo from Kazham.

Spawn each NM in game and confirm they reflect the proper changes listed in this PR


WinterSolstice8 (approved):

Xaver-DaRed (approved):

scripts/zones/Yhoator_Jungle/mobs/Acolnahuacatl.lua
```diff
@@ -9,6 +10,14 @@ entity.onMobInitialize = function(mob)
     mob:setMobMod(xi.mobMod.ADD_EFFECT, 1)
 end
 
+entity.onMobSpawn = function(mob)
+    mob:addImmunity(xi.immunity.LIGHT_SLEEP)
+    mob:addImmunity(xi.immunity.DARK_SLEEP)
+    mob:addImmunity(xi.immunity.GRAVITY)
+    mob:addImmunity(xi.immunity.BIND)
+    mob:setMod(xi.mod.STORETP, 25)
```
Xaver-DaRed: Have this store TP values been confirmed?

If not, could you add a `TODO` comment here and in the other "Store TP" mod down below stating that they need verification?

CriticalXI: Just went and got a capture and it pretty obviously has store tp to me. It uses TP moves pretty frequently while at high health. 

This is going back to a convo in my Rolanberry Fields NM audit where we discussed the store TP on all of these NMs that were added in the Nov 2009 patch. They all have matching bevhaior where they have the same immunties, only use 1 tp move, and have store tp.

https://youtu.be/VrP6haFN8-g

Xaver-DaRed: I think I didnt word myself properly. I have no doubt it has Store TP. What I am doing is questioning the value of the modifier.

"Store TP" can be, supposedly, calculated with relative ease by some1 that knows how TP works. Like WinterS.
So what I am asking here is if the value is 25 or if the value needs confirmation.

WinterSolstice8: Hard to prove exact numbers.  I don't think getting hit indicates the mobs store TP value to the player and NMs are uncharmable by default so you'd have to use something like Dark Seal absorb TP after one hit and then guess and check

CriticalXI: Thats a good point. I'll have to level up DRK to give that a try, I honestly had no clue how to verify the exact store TP amount.

LacunaXI: > Thats a good point. I'll have to level up DRK to give that a try, I honestly had no clue how to verify the exact store TP amount.

The Fatso Fargann jug pet is also an approach to get the exact TP.

Even when it comes to approximating these WotG NMs tend to get TP in ~5 hits unless they are a MNK and that is ~8 hits. Which can be contrast on your local build by checking the targets TP to see how many hits it will take at the assigned STP value.

WinterSolstice8: Oh, that jug is way more useful for that since you dont have to wait for dark seal. neat, thanks for that info

LacunaXI: You're welcome :)

CriticalXI: I've got a BST at 81 for that pet but can't find any of the item to summon. Will try and see if I can find someone on Asura with cooking today, otherwise I'll have to level up DRK for a bit.

CriticalXI: With Winter's help we were able to get a more accurate store tp value for Acolnahuacatl. I also went back and adjusted the store tp for a few NMs I recently adjusted as well with their more accurate store tp values.

LacunaXI: Awesome. Thank you for all the work.

PR #6693, opened 2025-01-12
add '.cache' folder to gitignore

Grahf0085:
**_I affirm:_**

On case sensitive systems the .cache folder isn't caught by the gitignore

Make a server on linux (and presumably macOS). The changes to the .cache folder will be tracked by git until you add "*.cache" to the gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,6 +109,7 @@ AppPackages/
 TestResults
 [Tt]est[Rr]esult*
 *.Cache
+*.cache
 ClientBin
 [Ss]tyle[Cc]op.*
 ~$*


Xaver-DaRed (approved):

WinterSolstice8 (approved):
hmm no cache folder here on ubuntu but seems reasonable

Grahf0085:
> hmm no cache folder here on ubuntu but seems reasonable

This is what I got after making the server earlier today: https://cloud.atavismxi.com/s/t7aHZwmXJznWWT9

PR #6699, opened 2025-01-12
[sql, lua] Moves silence and stun immunities to lua

Frankie-hz:
**_I affirm:_**

Moves silence and stun immunity column in sql to lua.

Should see no change.

See https://github.com/LandSandBoat/server/pull/6692 for more detailed information on how to test.


Xaver-DaRed:
Question:
I see onMobSpawn and onMobInitialize being used interchangebly

Shouldnt immunities be all in onMobInit?

zach2good:
Looking at the bindings, it doesn't seem to matter where they're put. Unless the immunities are changing at runtime, either is fine. If they're changing at runtime, probably need to fully reset them in onMobSpawn rather than anything on init

Frankie-hz:
> Question: I see onMobSpawn and onMobInitialize being used interchangebly
> 
> Shouldnt immunities be all in onMobInit?

Works on both, I can move them all to init if you want. If there was an init there I placed it there otherwise I made it onSpawn for new files. The idea I had behind that was if someone comes across adding stats to the mob (because most NMs have higher base stats/unique) they need to go in the spawn function otherwise it uses those as its base stats and it calculates wrong.

I would prefer them ALL go on the spawn function. I can move everything to the preferred one. I will fix the other PR as well once we decide

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #6706, opened 2025-01-13
[sql, cpp] Adds content tags to synth recipes

Frankie-hz:
**_I affirm:_**

Adds content tags to synth recipes.

THIS IS FOR TESTING PURPOSES ONLY - THIS CONTENT TAG IS NOT RIGHT
1) UPDATE synth_recipes SET content_tag = 'TOAU' WHERE ID = 20004;
2) !additem 4096
3) !additem 640 4
4) Craft using those items to make a copper ingot. It should not let you craft anything


src/map/utils/synthutils.cpp
```diff
@@ -87,6 +87,13 @@ namespace synthutils
 
         if (ret != SQL_ERROR && _sql->NumRows() != 0 && _sql->NextRow() == SQL_SUCCESS)
         {
+            // Check content tag first
+            char* contentTag = (char*)_sql->GetData(19);
```
zach2good, line 91: Normally I'd push for not using `char*` to represent strings. `std::string` is what we'd use here - but I'm literally rewriting this in the next couple of days so it doesn't matter 👍

zach2good (approved):

PR #6711, opened 2025-01-14
[core] Enable chat filter for others synth results (S2C 0x070)

WinterSolstice8:
**_I affirm:_**

Enable chat filter for others synth results (S2C 0x070). The server does NOT send packets when your synth filter is on via capture:

(capture is from an "other" char (casey) and the synther is robocop)
Filter off:
image
Filter on:
image

have 2 clients, synth on one, toggle filter on the other and see the chat filter work
image

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -406,6 +406,17 @@ void CCharEntity::clearPacketList()
     }
 }
 
+bool CCharEntity::isPacketFiltered(CBasicPacket* packet)
+{
+    // Filter others synthesis results
+    if (packet->getType() == 0x70 && playerConfig.MessageFilter.others_synthesis_and_fishing_results)
+    {
+        return true;
+    }
+
+    return false;
+}
+
 void CCharEntity::pushPacket(CBasicPacket* packet)
 {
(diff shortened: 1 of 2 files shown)


src/map/entities/charentity.cpp
```diff
@@ -414,6 +425,11 @@ void CCharEntity::pushPacket(CBasicPacket* packet)
 
     moduleutils::OnPushPacket(this, packet);
 
+    if (CCharEntity::isPacketFiltered(packet))
+    {
+        return;
```
zach2good, line 430: Reviewing from bed: if we return here doesn't this leak packet?

WinterSolstice8: Could be. that whole pushPacket is basically indecipherable and `packet` here is now unable to be dynamic casted for some unholy reason.

I'll run valgrind

WinterSolstice8: Looks like its probably not leaking but i'll remake this PR after everything is converted in #6713

zach2good:
I've taken (with --author) the logic from this PR into my incoming packet PR: https://github.com/LandSandBoat/server/pull/6713

Issue #5171, opened 2024-02-20
🔨 Death to raw operator `new`


zach2good:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature

I was looking at Teo's PR: https://github.com/LandSandBoat/server/pull/5163 and it struck me how we still have a lot of use of `new`. We have the capacity to use `unique_ptr`, but that's very verbose. I propose we template the `pushPacket` call and forward the args.

A la:
```cpp
PChar->pushPacket<CMessageBasicPacket>(PChar, PChar, 0, 0, MSGBASIC_CANNOT_USE_IN_AREA);
```

TeoTwawki:
>  propose we template the pushPacket call and forward the args.

oh yes please

PR #6713, opened 2025-01-14
Core: Remove CCharEntity::pushPacket(raw packet)

### LINKED ISSUE #6545: 🐛 Outgoing synth packet is always "your" synth result even though its from "others"
**_I affirm:_**

n/a

## Branch affected by issue 

`base`

## Steps to reproduce
enable filter of "other" synth results, see someone synth, get synth result anyway

Basically, we always send 0x06F when we should send 0x070 when it is received by "others". Our current implementation makes no distinction between "yours" and "others"

## Expected behavior

enable filter of "other" synth results, see someone synth, dont get their synth

zach2good:
**_I affirm:_**

Closes #6545

After looking here: I realised how easy it still is to leak packets: https://github.com/LandSandBoat/server/pull/6711

A big step towards: https://github.com/LandSandBoat/server/issues/5171

The scope of this blew up, as always. Now, `CBasicPacket` is ONLY a wrapper around `std::array<uint8, PACKET_SIZE> buffer_;`. It doesn't do non-owning, I've audited how they're constructed, etc. We also now NEVER `new` a packet - it's always constructed as a `unique_ptr`, and we are careful to pass around references or move them as needed. This also means that we have to be obvious when we wither `std::move()` or `->copy()` packets in places.

But now, there's no question about what happens where - we can see what our data is doing.

**Testing**

- [x] Log in
- [x] /wave
- [x] Zone into {Dynamis - Bastok}
- [x] Spells -> targeting failure
- [x] RA -> targeting failure
- [x] Action packet
- [x] Mounts
- [x] Chocobo raising menu
- [x] Train everything around
- [x] Crafting
- [x] Interacting with NPCs
- [x] Moghouse
- [x] Party
- [x] Trusts
- [x] Linkshell
- [x] IPC/ZMQ
- [x] Tell/Party/Shout/Yell
- [x] Events with updates (BCNM entry, Instance entry)
- [x] BCNMs
- [x] Instances
- [x] Trading
- [x] AH
- [x] Monstrosity

### DIFF
--- a/modules/custom/cpp/ah_announcement.cpp
+++ b/modules/custom/cpp/ah_announcement.cpp
@@ -179,7 +179,7 @@ class AHAnnouncementModule : public CPPModule
                                             name[0] = std::toupper(name[0]);
 
                                             // Send message to seller!
-                                            message::send(sellerId, new CChatMessagePacket(PChar, MESSAGE_SYSTEM_3,
+                                            message::send(sellerId, std::make_unique<CChatMessagePacket>(PChar, MESSAGE_SYSTEM_3,
                                                 fmt::format("Your '{}' has sold to {} for {} gil!", name, PChar->name, price).c_str(), ""));
                                         }
                                     }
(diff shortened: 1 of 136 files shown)


zach2good:
This is now very well tested and ready for review

Xaver-DaRed (approved):

PR #6728, opened 2025-01-16
[core] Add case for NPC entity on lua isSpawned()

WinterSolstice8:
**_I affirm:_**

Add case for NPC entity on lua isSpawned()

`!gotoname {name of npc} 1` of an NPC/qm that is up and see it tell you the QM/NPC is up rather than telling you it's not and going to its position anyway


Xaver-DaRed (approved):

claywar:
Slightly related, but I must have missed one in the search and replace just above.  We could eliminate the static cast there and clean up the code right above (else I can do that after this goes in)

claywar (approved):

zach2good (approved):

PR #6736, opened 2025-01-16
[core] Add CBattleEntity hasEnmity function

### LINKED ISSUE #6130: 🐛 Mount command does not respect gained enmity
**_I affirm:_**

Windows 11
## Branch affected by issue 

`base`

```Hax : !build
----== SystemMessage ==----
base-be17f7f3c3
Merge pull request #6121 from TiberonKalkaz/youre_a_wzard_galkas
Mon Aug 19 03:12:19 2024

client version
version 30240806_0
```    

## Steps to reproduce

- Gain agro / enmity from a monster
- Use mount

Video
https://cdn.discordapp.com/attachments/1095358627520458832/1275201865424048290/2024-08-19_15-52-37.mkv?ex=66c507c8&is=66c3b648&hm=20a9270f0702272bdb6c0d7f862c49256d9756ced07dc4505be6d8463dd96c59&
## Expected behavior
Players should not be able to use mounts with gained enmity.
> You are unable to mount while on a monster’s enmity list.
https://www.bg-wiki.com/ffxi/Category:Mounts

WinterSolstice8:
**_I affirm:_**

Fixes #6130

Aggro a mob, target yourself, see `!exec print(target:hasEnmity())` print true
Charm a mob, have your pet attack attack a mob and gain enmity, repeat !exec, see it return true
repeat for trusts

Normal pets, for some reason, don't get their notoriety container updated with things that don't like them. I don't know if that's intended or not, but it's out of scope. If a mob is directly attacking them the new function returns true. I think this is an acceptable edge case for now.

Aggro a mob, try to use /mount or summon a trust, see it fail

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -2691,3 +2691,32 @@ uint16 CBattleEntity::GetBattleTargetID() const
 {
     return m_battleTarget;
 }
+
+bool CBattleEntity::hasEnmityEXPENSIVE() const
+{
+    // TODO: This check seems to always fail for pets?
+    if (PNotorietyContainer->hasEnmity())
+    {
+        return true;
+    }
+
+    bool isTargeted = false;
+
+    // TODO: this is bad but because of how super tanking is implemented there's not much we can do without a larger refactor
+    if (loc.zone)
+    {
+        // clang-format off
+        loc.zone->ForEachMob([&](CMobEntity* PMob)
+        {
+            // Account for charmed mobs attacking normal mobs, etc
+            if (PMob->GetBattleTargetID() == targid && PMob->allegiance != allegiance)
+            {
+                isTargeted = true;
+                return;
+            }
+        });
+        // clang-format on
+    }
+
+    return isTargeted;
+}
--- a/src/map/entities/battleentity.h
+++ b/src/map/entities/battleentity.h
@@ -703,6 +703,8 @@ public:
     }
     CBattleEntity* GetBattleTarget();
 
+    bool hasEnmityEXPENSIVE() const; // Returns true if own notoriety container is not empty or mob in zone has entity listed as battle target
+
     /* State callbacks */
     /* Auto attack */
     virtual bool OnAttack(CAttackState&, action_t&);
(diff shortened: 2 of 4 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -2691,3 +2691,32 @@ uint16 CBattleEntity::GetBattleTargetID() const
 {
     return m_battleTarget;
 }
+
+bool CBattleEntity::hasEnmity() const
```
zach2good: I know this is going to be very ugly, but to sell how expensive this is (iterating possibly every mob in the zone if you aren't targetted) - can you call this `hasEnmityEXPENSIVE()` to make sure nobody uses it somewhere hot by accident?

src/map/entities/battleentity.cpp
```diff
@@ -2691,3 +2691,32 @@ uint16 CBattleEntity::GetBattleTargetID() const
...
+        return true;
+    }
+
+    bool isTargeted = false;
+
+    // TODO: this is bad but because of how super tanking is implemented there's not much we can do without a larger refactor
+    if (loc.zone)
+    {
+        // clang-format off
+        loc.zone->ForEachMob([&](CMobEntity* PMob)
```
zach2good, line 2709: When we eventually have spatial partitioning, we can do that sort of iteration here

WinterSolstice8: I don't think that will work here, because theoretically you could sleep a mob and Flee 100' away and it would STILL be mad at you

src/map/entities/battleentity.cpp
WinterSolstice8: sure, that works for me. I think the better longer-term fix here is to make a clone of the Notoriety Container but it only contains mobs that have aggroed you through PAI->Engage

zach2good: Yeah, thats the right answer ^

WinterSolstice8: The good news is this hasEnmity is sparsely used and only on-demand (trying to mount chocobo, call trust, use a chocobo whistle). All of these are rate limited by the client and/or other systems, so there's little risk of this being overused every tick

WinterSolstice8: Updated function name

zach2good (approved):

PR #6737, opened 2025-01-16
Lua: Update Orchestrion and Music IDs

zach2good:
**_I affirm:_**

I saw a new tunes keyitem and I went down the rabbithole a bit mapping these out

### DIFF
--- a/scripts/items/sheet_of_divine_tunes.lua
+++ b/scripts/items/sheet_of_divine_tunes.lua
@@ -1,0 +1,20 @@
+-----------------------------------
+-- ID: 6704
+-- Item: Divine Tunes
+-- Grants keyitem: SHEET_OF_DIVINE_TUNES
+-----------------------------------
+---@type TItem
+local itemObject = {}
+
+itemObject.onItemCheck = function(target, item, param, caster)
+    -- TODO: The error message returned by this is wrong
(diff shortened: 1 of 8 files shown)


scripts/items/sheet_of_divine_tunes.lua
```diff
@@ -0,0 +1,18 @@
...
+---@type TItem
+local itemObject = {}
+
+itemObject.onItemCheck = function(target, item, param, caster)
+    -- TODO: The error message returned by this is wrong
+    return target:hasKeyItem(xi.keyItem.SHEET_OF_DIVINE_TUNES) and 1 or 0
+end
+
+itemObject.onItemUse = function(target)
+    npcUtil.giveKeyItem(target, xi.keyItem.SHEET_OF_DIVINE_TUNES)
```
WinterSolstice8, line 17: P.S. this probably won't produce the right messaging of retail, so we need a TODO to cap it

See that right arrow all the way to the left?  the normal npcUtil.giveKeyItem doesn't generate this message
image

WinterSolstice8 (approved):

scripts/items/sheet_of_divine_tunes.lua
zach2good: Ohh quite right

claywar (approved):

PR #6751, opened 2025-01-18
Core: Refactor CZoneEntities::ZoneServer

zach2good:
**_I affirm:_**

This PR changes the pattern we have inside our "hot loop" (`ZoneServer`) of iterating with `const_iterator` and deleting things as we go, as they disappear, etc. This makes the logic and lifetimes of these entities very hard to reason about.

Instead, we will collect them up in `XToDelete` containers and clear them out at the end of the tick.

I've also written a macro (and thoroughly tested performance) that replaces these iterator loops and does the casting we want, while looking somewhat like a regular for loop. Performance is _exactly_ the same.

Moved a couple of lookups that were inside loops to be outside of loops - hopefully the compiler did this anyway.

General cleanup, refactoring, renaming, etc. Lifting intermediate containers out of the hot loop, pre-setting their capacity so they're less likely to expand at runtime, using the _ever so slightly faster_ distance calc helper in the hottest loop, tracy cleanup, TaskMgr cleanup, etc.

This is very much prep work for the spatial partitioning work I've been ramping up to do. Once this loop is clean and tidy, it'll be that much easier to change the logic.

To test:
- [x] Literally everything, jesus christ.
- [x] Instances
  - [x] !instance 7700 - Is aggro broken here? 
  - [x] Crash while !instance 7700 to exit?
  - [x] Amnaf not aggroing? -> doesn't aggro on `base` either
- [ ] Transport: `[01/23/25 09:52:48:438][map][error] CLuaBaseEntity::StartEventHelper: Could not start event, Character Entity already triggered. (CLuaBaseEntity::StartEventHelper:938)` is this my problem? It seems to work? Check on base

Done:
- [x] Where I've swapped loop styles, make sure I've not typo'd and iterated a different container (entity lists vs spawn lists)
- [x] Player zoning
- [x] DEs: Insertion, deletion, updating, etc.
- [x] Pets: Insertion, deletion, updating, etc.
- [x] Have I broken BCNM allies?
  - [x] The two Allies in Dawn seem to work fine 
- [x] BCNMs (Entered and completed Dawn - fine)
- [x] Have I fixed characters not being visible onZoneIn? -> Alas, seemingly not
- [x] Pets: PMaster problems? -> Nope
- [x] Trusts: All still sane?
- [x] NPCs & Shops: Still good?
- [x] Automatons?
- [x] Wyverns?
- [x] BST Jugs and Charm mobs
- [x] SMN Avatars
- [x] Mob vs Mob fighting
- [x] Aggro in general
- [x] Distance checking: Make sure I have converted all of the squared distances and true/false checks correctly. Each one. Individually.
- [x] Weather
- [x] Monstrosity + Feretory
- [x] Entering Mog Garden
- [x] Transport + Ferries

## Loop Proof

I was driving myself insane trying to get exactly the same performance results to show up in Tracy, so I decided it would be useful to compare the assembly output of the original loops with the new macro-based solution I've settled on:

<details>
  <summary>Code</summary>

```cpp
#include <unordered_map>
#include <string>
#include <iostream>

class CBaseEntity
{
public:
    virtual ~CBaseEntity() = default;

    void doSomethingBasic() const
    {
        std::cout << "Doing something basic" << std::endl;
    }
};

class CCharEntity : public CBaseEntity
{
public:
    void doSomethingCharacterful() const
    {
        std::cout << "Doing something characterful" << std::endl;
    }
};

using EntityList_t = std::unordered_map<int, CBaseEntity*>;

#define FOR_EACH_PAIR_CAST_SECOND(_collection, _type, _var)              \
    for (const auto& [_key, _value] : _collection)                       \
        if (auto _var = static_cast<_type>(_value); true)

#define FOR_EACH_PAIR_SECOND(_collection, _var)                          \
    for (const auto& [_key, _value] : _collection)                       \
        if (auto _var = _value; true)

void originalLoop(const EntityList_t& m_charList)
{
    for (EntityList_t::const_iterator it = m_charList.begin(); it != m_charList.end(); ++it)
    {
        CCharEntity* PCurrentChar = (CCharEntity*)it->second;
        PCurrentChar->doSomethingCharacterful();
    }
}

void macroLoopWithCast(const EntityList_t& m_charList)
{
    FOR_EACH_PAIR_CAST_SECOND(m_charList, CCharEntity*, PCurrentChar)
    {
        PCurrentChar->doSomethingCharacterful();
    }
}

void macroLoopNoCast(const EntityList_t& m_charList)
{
    FOR_EACH_PAIR_SECOND(m_charList, PCurrentEntity)
    {
        PCurrentEntity->doSomethingBasic();
    }
}

int main()
{
    EntityList_t m_charList;
    m_charList[1] = new CCharEntity();
    m_charList[2] = new CCharEntity();

    originalLoop(m_charList);
    macroLoopWithCast(m_charList);
    macroLoopNoCast(m_charList);

    for (auto& [key, value] : m_charList)
    {
        delete value;
    }

    return 0;
}
```

</details>

<details>
  <summary>Assembly</summary>

As you can see, all three styles of loop produce exactly the same 11 assembly instructions.

```asm
originalLoop(std::unordered_map<int, CBaseEntity*, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<int const, CBaseEntity*>>> const&):
        push    rbx
        mov     rbx, qword ptr [rdi + 16]
        test    rbx, rbx
        je      .LBB0_3
.LBB0_1:
        mov     rdi, qword ptr [rbx + 16]
        call    CCharEntity::doSomethingCharacterful() const
        mov     rbx, qword ptr [rbx]
        test    rbx, rbx
        jne     .LBB0_1
.LBB0_3:
        pop     rbx
        ret
```

```asm
macroLoopWithCast(std::unordered_map<int, CBaseEntity*, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<int const, CBaseEntity*>>> const&):
        push    rbx
        mov     rbx, qword ptr [rdi + 16]
        test    rbx, rbx
        je      .LBB2_3
.LBB2_1:
        mov     rdi, qword ptr [rbx + 16]
        call    CCharEntity::doSomethingCharacterful() const
        mov     rbx, qword ptr [rbx]
        test    rbx, rbx
        jne     .LBB2_1
.LBB2_3:
        pop     rbx
        ret
```

```asm
macroLoopNoCast(std::unordered_map<int, CBaseEntity*, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<int const, CBaseEntity*>>> const&):
        push    rbx
        mov     rbx, qword ptr [rdi + 16]
        test    rbx, rbx
        je      .LBB3_3
.LBB3_1:
        mov     rdi, qword ptr [rbx + 16]
        call    CBaseEntity::doSomethingBasic() const
        mov     rbx, qword ptr [rbx]
        test    rbx, rbx
        jne     .LBB3_1
.LBB3_3:
        pop     rbx
        ret
```

</details>

Godbolt: https://godbolt.org/z/GE4v4PE3x

### DIFF
--- a/src/map/ai/controllers/automaton_controller.cpp
+++ b/src/map/ai/controllers/automaton_controller.cpp
@@ -223,9 +223,8 @@ void CAutomatonController::DoCombatTick(time_point tick)
 
 void CAutomatonController::Move()
 {
-    float currentDistance = distanceSquared(PAutomaton->loc.p, PTarget->loc.p);
-
-    if ((shouldStandBack() && (currentDistance > 225)) || (PAutomaton->health.mp < 8 && PAutomaton->health.maxmp > 8))
+    if ((shouldStandBack() && !isWithinDistance(PAutomaton->loc.p, PTarget->loc.p, 15.0f)) ||
+        (PAutomaton->health.mp < 8 && PAutomaton->health.maxmp > 8))
     {
         PAutomaton->m_Behavior &= ~BEHAVIOR_STANDBACK;
     }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -391,7 +391,7 @@ jobs:
       - name: Build
         shell: cmd
         run: |
-          cmake --build build -j4
+          cmake --build build --config Release -j4
       - name: Archive Executables
         uses: actions/upload-artifact@v4
         with:
--- a/src/common/cbasetypes.h
+++ b/src/common/cbasetypes.h
@@ -9,29 +9,7 @@
 #include <utility>
 
 #include "logging.h"
-
-// The following definitions are set by CMake based on the architecture
-// #define ENV64BIT
-// #define ENV32BIT
-
-// Ensure one of the definitions is set
-#if !defined(ENV64BIT) && !defined(ENV32BIT)
-#error "Neither ENV64BIT nor ENV32BIT is defined"
-#endif
-
-// Debug mode
-#if defined(_DEBUG) && !defined(DEBUG)
-#define DEBUG
-#endif
-
-// Release mode
-#if !defined(_DEBUG) && !defined(RELEASE)
-#define RELEASE
-#endif
-
-// define a break macro for debugging
-#define XI_DEBUG_BREAK_IF(_CONDITION_) \
-    static_assert(false, "Use of XI_DEBUG_BREAK_IF is deprecated. Check your conditions and log appropriately instead.")
+#include "macros.h"
 
 // typedef/using
 using int8  = std::int8_t;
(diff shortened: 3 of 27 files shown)


claywar (approved):

src/map/ai/controllers/automaton_controller.cpp
```diff
@@ -223,9 +223,9 @@ void CAutomatonController::DoCombatTick(time_point tick)
 
 void CAutomatonController::Move()
 {
-    float currentDistance = distanceSquared(PAutomaton->loc.p, PTarget->loc.p);
+    const float currentDistance = distance(PAutomaton->loc.p, PTarget->loc.p);
 
-    if ((shouldStandBack() && (currentDistance > 225)) || (PAutomaton->health.mp < 8 && PAutomaton->health.maxmp > 8))
+    if ((shouldStandBack() && (currentDistance > 15)) || (PAutomaton->health.mp < 8 && PAutomaton->health.maxmp > 8))
```
cocosolos: Should this be checking `!isWithinDistance(PAutomaton->loc.p, PTarget->loc.p, 15.0f)` instead of using `distance`?

Xaver-DaRed (approved):

PR #6756, opened 2025-01-18
Chore: Audit mob droplist - Create 3302 to 3332

LacunaXI:
**_I affirm:_**

I have audited the entire mob_droplist.SQL and all dropids up until Seekers of Adoulin. This was a very time consuming task which involved utilizing FFXIDB, occasional retail data collection and watching countless videos for certain NMs such as Wyrms or Kings and beyond. Including Japanese videos with the Google image translation function on my phone.

The drop SQL was understandably a mess and this project should dramatically improve the accuracy of mob drops. When data was often missing for NMs I took the existing standard in the file and set the rate to Uncommon, 10%.

There is a lot of data pollution (AoE burning seems to throw things a bit?) or skewed data populating FFXIDB. The majority of the time this seems to be from TH0 being inflated by trust/pet TH and TH+ equipment. This makes it especially muddy where Common, 15% can appear as Very Common, 24%. This required paying greater attention to the other TH results on FFXIDB and establishing shared or similar patterns despite TH1/2 or general data often being limited. Over half a year of this pattern seeking has made discerning tiers from the data to become a newly acquired if not depressingly niche skill.

When a drop seemed between rates (particularly with Common and Very Common again) I would generally aim conservatively and downgrade a rank. There was also a loose theme in how retail was set with dungeons often having a rank increase. I figure this to be because mob respawn times originally being 2-3xs higher than foes outside.

There was often a good amount of shared consistency in drops and drop rates between families when dividing between fields and dungeons. Only beastmen in dungeons dropped player equipment. This vanilla equipment (Bronze Subligar) was always the same drop rate and so on. Retail testing showed these vanilla equipment drops are not pooled and multiple can drop but the drop rate falls dramatically the higher level the armor is. Starting at Very Rare, 1% for the lowest level equipment and quickly decreasing before eventually reaching Ultra Rare, 0.1%.

After auditing all 3,301 dropids I then went zone by zone looking at anything set to dropid 0. This lead to the creation of dropid 3302-3368. There will be more created as I go through the file still. As I made somewhere approaching 2,000 TODO notes to myself. I want to start at the end with these new dropids then return to the beginning of the file. I am breaking these PRs down into 30 dropids to make it manageable. This will still create over 110 PRs from start to finish. Please advise if it is more desirable for this to be larger or smaller in size per PR.

One final note is that Abyssea NMs are unique and will require their own project for anyone interested. Since the weakness triggering for Blue!! or Yellow!! dramatically influences drops. When it came to synth material drops I just removed these as that will need to be studied. These were already incorrectly and incompletely tossed in at random or often completely omitted.

There are some synth materials that can drop without Yellow!! and some which only drop with Yellow!! as well as Yellow!! opening additional slots. Any Blue!! influenced drops seem to generally be at a Rare, 5% base. While any equipment not influenced by weakness triggering is generally Common, 15% - Very Common, 24%. This is how I set these up in order to establish a cleaner basis with a TODO note on each. What the drop rate becomes for these Rare, 5% Blue!! equipment drops once triggered needs a little study as well.

- Audit every entry for retail accuracy.
- Adjust, reassign, or create dropids.
- Submit many PRs overtime and feel triumphant.

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -2500,7 +2500,6 @@ INSERT INTO `mob_droplist` VALUES (257,0,0,1000,816,140); -- Spool Of Silk Threa
 INSERT INTO `mob_droplist` VALUES (257,4,0,1000,4357,0);  -- Crawler Egg (Despoil)
 
 -- ZoneID: 137 - Berserker Demon
--- ZoneID: 137 - Inferno Demon
 INSERT INTO `mob_droplist` VALUES (258,0,0,1000,902,110);        -- Demon Horn (11.0%)
 INSERT INTO `mob_droplist` VALUES (258,0,0,1000,2825,@UNCOMMON); -- Square Of Cambric (Uncommon, 10%)
 INSERT INTO `mob_droplist` VALUES (258,0,0,1000,886,@RARE);      -- Demon Skull (Rare, 5%)
@@ -3846,7 +3845,6 @@ INSERT INTO `mob_droplist` VALUES (423,2,0,1000,816,0);       -- Spool Of Silk T
 
 -- ZoneID:  65 - Carriage Lizard
 -- ZoneID: 160 - Tormentor
--- ZoneID: 191 - Fume Lizard
 INSERT INTO `mob_droplist` VALUES (424,0,0,1000,926,180); -- Lizard Tail (18.0%)
 INSERT INTO `mob_droplist` VALUES (424,0,0,1000,852,30);  -- Lizard Skin (3.0%)
 INSERT INTO `mob_droplist` VALUES (424,0,0,1000,4362,30); -- Lizard Egg (3.0%)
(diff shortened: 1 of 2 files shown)


sql/mob_droplist.sql
```diff
@@ -26819,6 +26817,173 @@ INSERT INTO `mob_droplist` VALUES (3301,0,0,1000,936,@COMMON); -- Chunk Of Rock
...
+INSERT INTO `mob_droplist` VALUES (3304,0,0,1000,2522,@COMMON); -- Gnat Wing (Common, 15%)
+INSERT INTO `mob_droplist` VALUES (3304,0,0,1000,2490,@VRARE);  -- Forbidden Key (Very Rare, 1%)
+INSERT INTO `mob_droplist` VALUES (3304,4,0,1000,2522,0);       -- Gnat Wing (Despoil)
+
+-- ZoneID:  45 - Wiederganger
+INSERT INTO `mob_droplist` VALUES (3305,0,0,1000,1639,@UNCOMMON); -- Corse Robe (Uncommon, 10%)
+INSERT INTO `mob_droplist` VALUES (3305,0,0,1000,1614,@UNCOMMON); -- Corse Bracelet (Uncommon, 10%)
+INSERT INTO `mob_droplist` VALUES (3305,0,0,1000,2490,@VRARE);    -- Forbidden Key (Very Rare, 1%)
+
+-- ZoneID:  46 - Nothern Piranu
```
Frankie-hz, line 26843: Northern

Frankie-hz:
All the drop rates look reasonable based on the information we have and align with FFXIDB

WinterSolstice8 (approved):

UmeboshiXI:
Had this on my TODO list if you want to roll it into your changes. Wasn't sure if you got to this point yet.

https://github.com/LandSandBoat/server/issues/6762

LacunaXI:
> Had this on my TODO list if you want to roll it into your changes. Wasn't sure if you got to this point yet.
> 
> #6762

This was broken in #6382 and my branch for the initial work was older than this so I did not see it as I audited. I will correct it in my next PR. Good catch.

sql/mob_droplist.sql
```diff
@@ -26819,6 +26817,173 @@ INSERT INTO `mob_droplist` VALUES (3301,0,0,1000,936,@COMMON); -- Chunk Of Rock
...
+INSERT INTO `mob_droplist` VALUES (3325,0,0,1000,2755,@VCOMMON);   -- Ruszor Hide (Very Common, 24%)
+INSERT INTO `mob_droplist` VALUES (3325,0,0,1000,2824,@VCOMMON);   -- Square Of Shagreen (Very Common, 24%)
+INSERT INTO `mob_droplist` VALUES (3325,0,0,1000,2824,@VCOMMON);   -- Square Of Shagreen (Very Common, 24%)
+INSERT INTO `mob_droplist` VALUES (3325,0,0,1000,1311,@UNCOMMON);  -- Piece Of Oxblood (Uncommon, 10%)
+INSERT INTO `mob_droplist` VALUES (3325,0,0,1000,1312,@RARE);      -- Piece Of Angel Skin (Rare, 5%)
+INSERT INTO `mob_droplist` VALUES (3325,0,0,1000,19126,@UNCOMMON); -- Papilio Kirpan (Uncommon, 10%)
+
+-- ZoneID: 136 - Becut
+INSERT INTO `mob_droplist` VALUES (3326,0,0,1000,2815,@UNCOMMON);      -- Aptant Of Arkhe (Uncommon, 10%)
+INSERT INTO `mob_droplist` VALUES (3326,0,0,1000,2822,2811,@UNCOMMON); -- Aptant Of Primus (Uncommon, 10%)
```
mjsalerno, line 26945: This line causes an issue for me because it has too many arguments.
when updating the DB with dbtool.py i get
ERROR 1136 (21S01) at line 26945 in file: '/home/xi/server/sql/mob_droplist.sql': Column count doesn't match value count at row 1

WinterSolstice8:
```
--------------
INSERT INTO `mob_droplist` VALUES (3326,0,0,1000,2822,2811,@UNCOMMON)
--------------

ERROR 1136 (21S01) at line 26945 in file: Column count doesn't match value count at row 1
```
can confirm

WinterSolstice8: #6772

LacunaXI: Sorry for the mistake.

PR #6759, opened 2025-01-19
[sql, lua] Removal of mob_spawn_mods.sql

Frankie-hz:
**_I affirm:_**

Removes the mob_spawn_mods.sql

This SQL has been known to cause issue after issue with having hard-coded IDs. It is time is goes away once and for all and move everything into lua

You should observe no changes. All the mods should work when spawning said monster.


UmeboshiXI:
Something that I noticed that might need confirming:

On current retail, the Satellite weapons are "Normal" mobs now from what I understand. You just farm them to get an item to pop the corresponding animated weapon.

They likely may not need an Idle Despawn mod (May lead to weird behavior having them spawn in and out as normal mobs.)

scripts/enum/mob_mod.lua
```diff
@@ -34,7 +34,7 @@ xi.mobMod =
     IMMUNITY               = 23, -- immune to set status effects. This only works from the db, not scripts
```
Xaver-DaRed, line 34: We will have to handle this mobmod aswell

Xaver-DaRed (approved):

Frankie-hz:
> Something that I noticed that might need confirming:
> 
> On current retail, the Satellite weapons are "Normal" mobs now from what I understand. You just farm them to get an item to pop the corresponding animated weapon.
> 
> They likely may not need an Idle Despawn mod (May lead to weird behavior having them spawn in and out as normal mobs.)

A lot of this looks very suspicious to me but this file has been notorious for getting overlooked time and time again for ID shifts and mods in general. I am guessing a lot of this is inaccurate but this PR was simply to remove this sql from existence so all of this behavior is already happening, its just now finally being seen more clearly.

PR #6771, opened 2025-01-20
[lua] Resolve Automaton under/overflow with LEVEL_RESTRICTION

sruon:
**_I affirm:_**

When applying `LEVEL_RESTRICTION`, the automaton modifiers are reset by the `levelRestriction ` handler but the internal state of `automaton.lua` is not. 
When maneuvers wear after the `LEVEL_RESTRICTION` effect is applied, the `onManaveurLost` logic then tries to substract the modifiers it thinks are currently applied to the automaton. 

Depending on the buffs provided, this can lead to over or underflow of the modifier:
- Magniplug grants nearly capped weapon damage
- Mana Tank grants negative Refresh
- Auto-Repair kits gives a Poison-like effect
- Armor Plates makes the Automaton take more damage

Steps to reproduce ([video](https://youtu.be/GP4-TgT349Y)):
```
!changejob PUP 75
!addallattachments
Equip any frame/head
Equip "Magniplug" attachment (affects weapon/ranged damage)

Use "Activate" to summon automaton
(Use !getstats 3 on the automaton in between steps)
!addeffect FIRE_MANEUVER
!addeffect FIRE_MANEUVER
!addeffect FIRE_MANEUVER
!addeffect LEVEL_RESTRICTION 10
!deleffect FIRE_MANEUVER
!deleffect FIRE_MANEUVER
!deleffect FIRE_MANEUVER

Congratulations, you're the proud owner of an automaton with 65000+ weapon damage.
```

This PR does several things:
- Ensures _all_ modifiers are cleared and zero out related local variables when onAttachmentUnequip is called
- Adds a Pre/PostLevelRestriction to puppetutils, called by levelRestriction at different stages to ensure we can clean up existing state AND recalculate the attachments on the new automaton stats
- Adds a clearLocalVarsWithPrefix lua function, similar in spirit to clearVarsWithPrefix
- Adds setAttachment/setAutomatonHead/setAutomatonFrame lua functions, which I used for testing
- Enable Truesights attachment (unrelated to issue but it was a quick fix)
- Refactor Volt Gun attachment script to apply stats in a slightly different but still accurate way.

Several of these changes are not strictly related to the issue and could be removed from the PR, let me know.

Notes:
- Level restriction does not wipe maneuvers (or any other buff) today. On Retail level restriction wipes buffs in the majority of cases with only a handful of exceptions. This change should also work if this were to become a default behavior of levelRestriction and maneuvers were wiped.
- Initially tried to implement it using `onPetLevelRestriction` but it was too difficult without a major rewrite of `automaton.lua`

https://youtu.be/dBSDdiD2OT8

Collected data about every attachment  I could test (88/105)

https://docs.google.com/spreadsheets/d/1cEZ2CZkZEAgN-NbqpQzDVZT6BLkXy2SpQYkDX9v4Ykc/edit?usp=sharing

Maneuvers description:
- 0 maneuver, just onEquip bonus (without level restriction)
- 1 -> 3 maneuvers (without level restriction)
- 3 ->0 maneuvers (after level restriction)
- 1 -> 3 maneuvers (still under level restriction)

Test script [available here](https://gist.github.com/sruon/23f4842fa1001b3a550c13f1e910fdbc)
```
// Test without optical fiber, results in pup_attachments.csv next to xi_map.exe
!automatontest

// To test with optical fiber
!automatontest anythinghere
```

### DIFF
--- a/scripts/actions/abilities/pets/attachments/volt_gun.lua
+++ b/scripts/actions/abilities/pets/attachments/volt_gun.lua
@@ -6,7 +6,7 @@ local attachmentObject = {}
 
 local function calcEnspellDmg(pet, maneuvers)
     local skill = math.max(pet:getSkillLevel(xi.skill.AUTOMATON_MELEE), pet:getSkillLevel(xi.skill.AUTOMATON_RANGED), pet:getSkillLevel(xi.skill.AUTOMATON_MAGIC))
-    return (skill * 0.1) + ((skill * 0.05) * maneuvers)
+    return math.floor(skill / 10 + skill * maneuvers / 20)
 end
 
 attachmentObject.onEquip = function(pet, attachment)
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -720,16 +720,12 @@ void CLuaBaseEntity::setLocalVar(std::string const& var, uint32 val)
 
 void CLuaBaseEntity::clearLocalVarsWithPrefix(std::string const& prefix)
 {
-    auto& localVars = m_PBaseEntity->GetLocalVars();
-
-    auto iter = localVars.begin();
-    while (iter != localVars.end())
+    for (const auto& localVar : m_PBaseEntity->GetLocalVars())
     {
-        if (iter->first.rfind(prefix, 0) == 0)
+        if (starts_with(localVar.first, prefix))
         {
-            m_PBaseEntity->SetLocalVar(iter->first.c_str(), 0);
+            m_PBaseEntity->SetLocalVar(localVar.first, 0);
         }
-        ++iter;
     }
 
     return;
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -712,6 +712,30 @@ void CLuaBaseEntity::setLocalVar(std::string const& var, uint32 val)
 }
 
 /************************************************************************
+ *  Function: clearLocalVarsWithPrefix()
+ *  Purpose : Deletes all local variables with the given prefix.
+ *  Example : pet:clearLocalVarsWithPrefix("volt");
+ *  Notes   :
+ ************************************************************************/
+
+void CLuaBaseEntity::clearLocalVarsWithPrefix(std::string const& prefix)
+{
+    auto& localVars = m_PBaseEntity->GetLocalVars();
+
+    auto iter = localVars.begin();
+    while (iter != localVars.end())
+    {
+        if (iter->first.rfind(prefix, 0) == 0)
+        {
+            m_PBaseEntity->SetLocalVar(iter->first.c_str(), 0);
+        }
+        ++iter;
+    }
+
+    return;
+}
+
+/************************************************************************
  *  Function: resetLocalVars()
  *  Purpose : Reset local variables back to default (ex: on Mob disengage)
  *  Example : GetMobByID(Defender):resetLocalVars();
@@ -6615,6 +6639,11 @@ uint8 CLuaBaseEntity::levelRestriction(sol::object const& level)
 
         if (PChar->GetMLevel() != NewMLevel)
         {
+            if (PChar->PAutomaton)
+            {
+                // Call each attachment onUnequip handler and zero out localVars tracking applied buffs
+                puppetutils::PreLevelRestriction(PChar);
+            }
             charutils::RemoveAllEquipMods(PChar);
             PChar->SetMLevel(NewMLevel);
             PChar->SetSLevel(PChar->jobs.job[PChar->GetSJob()]);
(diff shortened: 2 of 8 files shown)


scripts/actions/abilities/pets/attachments/volt_gun.lua
```diff
@@ -4,37 +4,34 @@
...
+
+attachmentObject.onEquip = function(pet, attachment)
+    pet:setMod(xi.mod.ENSPELL_DMG, calcEnspellDmg(pet, 0))
+    xi.automaton.onAttachmentEquip(pet, attachment)
 end
 
 attachmentObject.onUnequip = function(pet, attachment)
-    pet:delMod(xi.mod.ENSPELL, xi.element.THUNDER)
     pet:delMod(xi.mod.ENSPELL_DMG, pet:getMod(xi.mod.ENSPELL_DMG))
-    pet:delMod(xi.mod.ENSPELL_CHANCE, 20)
```
Xaver-DaRed, line 17: Why are we removing the enspell element and chance modifiers?

scripts/actions/abilities/pets/attachments/volt_gun.lua
```diff
@@ -4,37 +4,34 @@
 ---@type TAttachment
 local attachmentObject = {}
 
-attachmentObject.onEquip = function(pet, attachment)
+local function calcEnspellDmg(pet, maneuvers)
     local skill = math.max(pet:getSkillLevel(xi.skill.AUTOMATON_MELEE), pet:getSkillLevel(xi.skill.AUTOMATON_RANGED), pet:getSkillLevel(xi.skill.AUTOMATON_MAGIC))
-    pet:addMod(xi.mod.ENSPELL, xi.element.THUNDER)
-    pet:addMod(xi.mod.ENSPELL_DMG, skill * 0.1)
-    pet:addMod(xi.mod.ENSPELL_CHANCE, 20)
+    return (skill * 0.1) + ((skill * 0.05) * maneuvers)
```
Xaver-DaRed: return math.floor(skill / 10 + skill * maneuvers / 20)

scripts/actions/abilities/pets/attachments/volt_gun.lua
WinterSolstice8: They were moved into scripts/globals/automaton.lua

src/map/utils/puppetutils.cpp
```diff
@@ -778,4 +778,50 @@ namespace puppetutils
...
+
+        if (PAutomaton)
+        {
+            for (int i = 0; i < 12; i++)
+            {
+                uint8 attachment = PAutomaton->getAttachment(i);
+
+                if (attachment != 0)
+                {
+                    CItemPuppet* PAttachment = (CItemPuppet*)itemutils::GetItemPointer(0x2100 + attachment);
```
WinterSolstice8:
```suggestion
                    CItemPuppet* PAttachment = dynamic_cast<CItemPuppet*>(itemutils::GetItemPointer(0x2100 + attachment));
                    if (PAttachment)
                    ...
```

WinterSolstice8: If it must be a hard cast, use `static_cast<CItemPuppet*>`, we prefer to use dynamic_casting and checking the result however

WinterSolstice8:
Dumb question:

Would this solve itself if maneuvers were removed on level sync?

sruon:
> Dumb question:
> 
> Would this solve itself if maneuvers were removed on level sync?

It would if the buffs were cleared before reducing the levels and clearing the modifiers. This is initially what I wanted to propose as a change after testing it on Retail.

However, there are a handful of cases where Level Restriction does not wipe buffs, which means this would still need to be handled.

> There are some very rare cases where the restriction removes no buffs at all, such as during [All by Myself](https://ffxiclopedia.fandom.com/wiki/All_by_Myself), [Eco-Warrior](https://ffxiclopedia.fandom.com/wiki/Eco-Warrior), and all "Trial-Size" avatar fights.

scripts/actions/abilities/pets/attachments/volt_gun.lua
sruon: Done and retested

WinterSolstice8 (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -711,6 +711,30 @@ void CLuaBaseEntity::setLocalVar(std::string const& var, uint32 val)
...
+
+    auto iter = localVars.begin();
+    while (iter != localVars.end())
+    {
+        if (iter->first.rfind(prefix, 0) == 0)
+        {
+            m_PBaseEntity->SetLocalVar(iter->first.c_str(), 0);
+        }
+        ++iter;
+    }
```
zach2good:
```cpp
for (const auto& localVar : m_PBaseEntity->GetLocalVars())
{
    if (starts_with(localVar, prefix))
    {
        m_PBaseEntity->SetLocalVar(localVar, 0);
    }
}
```

zach2good: Ah didnt see it was pairs, use: https://en.cppreference.com/w/cpp/language/structured_binding

zach2good:
```cpp
for (const auto& [localVar, _] : m_PBaseEntity->GetLocalVars())
{
    if (starts_with(localVar, prefix))
    {
        m_PBaseEntity->SetLocalVar(localVar, 0);
    }
}
```

PR #6777, opened 2025-01-21
Fix setSpeed Lua binding

cocosolos:
**_I affirm:_**

- Speed is dynamic and shouldn't be set directly. Make it protected and change speed by changing the base speed or apply an override modifier.
- Manually changing speed for run pathing is unnecessary (see Dark Ixion).
- Luopan and bind set base speed to 0, removing unnecessary "magic functionality" from MOVE_SPEED_OVERRIDE modifier.
- Remove references to old speed settings. Fix "speed" GM command.

Set job to GEO, cast a geo spell, see Luopan not follow you.
Use !speed GM command and see it work properly.
Bind a mob and see its speed go to 0 and not move. Watch as it moves again when the effect wears.
```
!exec player:addAssault(31)
!instance 5601
```
Watch Excaliace try to run and escape.

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -307,7 +307,7 @@ uint8 CBattleEntity::UpdateSpeed(bool run)
         outputSpeed = settings::get<uint8>("map.MOUNT_SPEED") / 2;
         outputSpeed *= (100 + getMod(Mod::MOUNT_MOVE)) / 100;
     }
-    else if (baseSpeed == 0 || getMod(Mod::MOVE_SPEED_OVERRIDE) > 255)
+    else if (baseSpeed == 0 || getMod(Mod::MOVE_SPEED_OVERRIDE) < 0)
     {
         outputSpeed = 0;
     }
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -306,96 +306,92 @@ uint8 CBattleEntity::UpdateSpeed(bool run)
     {
         outputSpeed = settings::get<uint8>("map.MOUNT_SPEED") / 2;
         outputSpeed *= (100 + getMod(Mod::MOUNT_MOVE)) / 100;
-        speed = std::clamp<uint8>(outputSpeed, std::numeric_limits<uint8>::min(), std::numeric_limits<uint8>::max());
-
-        return speed;
     }
-
-    // Gear penalties.
-    int8 additiveMods = static_cast<int8>(getMod(Mod::MOVE_SPEED_STACKABLE));
-
-    // Gravity and Curse. They seem additive to each other and the sum seems to be multiplicative.
-    float weightFactor = std::clamp<float>(1.0f - static_cast<float>(getMod(Mod::MOVE_SPEED_WEIGHT_PENALTY)) / 100.0f, 0.1f, 1.0f);
-
-    // Flee.
-    float fleeFactor = std::clamp<float>(1.0f + static_cast<float>(getMod(Mod::MOVE_SPEED_FLEE)) / 10000.0f, 1.0f, 2.0f);
-
-    // Cheer KI's
-    float cheerFactor = (99.0f + static_cast<float>(getMod(Mod::MOVE_SPEED_CHEER))) / 99.0f;
-
-    // Bolter's Roll. Additive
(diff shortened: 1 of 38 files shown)


WinterSolstice8 (approved):

src/map/entities/battleentity.cpp
```diff
@@ -388,12 +388,7 @@ uint8 CBattleEntity::UpdateSpeed(bool run)
     // TODO: Find exceptions. Add them here.
 
     // GM speed bypass.
-
-    if (getMod(Mod::MOVE_SPEED_OVERRIDE) > 255)
-    {
-        outputSpeed = 0;
-    }
-    else if (getMod(Mod::MOVE_SPEED_OVERRIDE) > 0)
```
Xaver-DaRed, line 396: I would really like to keep this functionality

WinterSolstice8: setBaseSpeed(0) will do this now if I understand correctly

cocosolos: > setBaseSpeed(0) will do this now if I understand correctly

This is the intent yes but there is one issue I just noticed. If base speed is set to 0, bolter's roll can still raise it. Instead, I can just return 0 if base speed is 0 (this wouldn't affect setting your speed to 0 with gear or other means). I would prefer something like that, where "base speed 0 means 0 full stop" as opposed to "big number means stop". IDK if "magic functionality" is a real term but that's what this feels like to me whereas setting base speed to 0 makes more logical sense.

cocosolos: Something like this:
```cpp
    if (getMod(Mod::MOVE_SPEED_OVERRIDE) > 0)
    {
        outputSpeed = getMod(Mod::MOVE_SPEED_OVERRIDE);
    }
    else if (baseSpeed == 0)
    {
        return baseSpeed;
    }
```

Xaver-DaRed: How about this:
```c++
    int16 speedOverride = getMod(Mod::MOVE_SPEED_OVERRIDE)

    if (speedOverride > 255)
    {
        return 0;
    }
    else if (speedOverride > 0)
    {
        return speedOverride;
    }
    else if (baseSpeed == 0)
    {
        return 0;
    }
```

Xaver-DaRed: Edited

cocosolos: What's the reason for keeping the mod like that though when setting base speed would do the same thing?

Xaver-DaRed: - It doesnt hurt, considering that if that modifier went over 255 it would overflow the speed output anyway
- It allows for GMs to set their speed to 0 using the !speed GM command instead of anything else
- Its a single line of code

cocosolos: I mean, it's clamped so it would never overflow. The GM command is a valid point I guess, I just don't like that pattern is all. Would it also be acceptable to have a negative modifier set it to 255? It's one of those things where if you know about it, you can use it instead of just setting the base speed, but if you don't know about it, you unexpectedly get bound if you go over.

cocosolos: Rearranged some stuff to reduce work being done, relevant part here:
```cpp
    else if (baseSpeed == 0 || getMod(Mod::MOVE_SPEED_OVERRIDE) > 255)
    {
        outputSpeed = 0;
    }
    else if (getMod(Mod::MOVE_SPEED_OVERRIDE) > 0)
    {
        // GM speed bypass.
        // Speed cap can be bypassed. Ex. Feast of swords. GM speed.
        // TODO: Find exceptions. Add them here.
        outputSpeed = getMod(Mod::MOVE_SPEED_OVERRIDE);
    }
    else
    {
```
everything else is just indentation.

cocosolos: @Xaver-DaRed how do you feel about this second commit? Using a negative mod to bind at 0 speed, otherwise clamping 0-255. Functionally it's the same, but is more intuitive imo, and fixes the remaining issues with the speed command.

WinterSolstice8: to me -1 does indicate something special,  you wouldn't normally set a speed of -1 after all

I have no objections

Xaver-DaRed (approved):

PR #6780, opened 2025-01-21
Core: Introduce luautils::callGlobal helper

zach2good:
**_I affirm:_**

Allow calling of functions in Lua from C++ without needing to manually prepare so many things for Lua, or to manually/dangerously access the lua state:
```cpp
        luautils::callGlobal<void>("xi.server.onServerStart");
```

etc.

### DIFF
--- a/src/common/filewatcher.cpp
+++ b/src/common/filewatcher.cpp
@@ -78,7 +78,7 @@ void Filewatcher::handleFileAction(efsw::WatchID watchid, std::string const& dir
     }
 }
 
-auto Filewatcher::getChangedLuaFiles() -> std::vector<std::pair<std::filesystem::path, Action>>
+auto Filewatcher::popChangedLuaFilesList() -> std::vector<std::pair<std::filesystem::path, Action>>
 {
     std::set<std::pair<std::filesystem::path, Action>> actions; // For de-duping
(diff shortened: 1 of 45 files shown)


zach2good:
Ready for review, works with hot-reloading. Will do some performance testing this week to see if we gain anything from the caching layer

zach2good:
Trying to pin down why this is crashing during CS event updates inside LuaJIT

zach2good:
This ended up shaking out a logical inconsistency that I can't understand why it wasn't crashing before.

When you extract something out of sol, it's lazy-evaluated. If you ask for `lua[1][2][3]` it won't evaluate until you _do_ something with it. If you do `lua[1][2][3]()` it evaluates right away. We were pulling out a `lua_nil` from something, holding onto it for a bit, and then `obj()[]`ing into it without any additional checks.

I've added those checks in now.

It was around event string updates, so I've run it through the chocobo racing intro, which is the premiere spot for big event updates with arguments and strings. 

I've logged in, I've done gobbie box, I've punched some lizards in the face, I've called out trusts and Sneak Attacked off of them, I've used items - everything seems stable now.

Ready to go 👍

claywar (approved):

PR #6782, opened 2025-01-21
[lua] Remove & Replace "xi.conq" alias with "xi.conquest"

WinterSolstice8:
**_I affirm:_**

Remove & Replace "xi.conq" alias with "xi.conquest"

xi.conq alias was manually removed, the rest was performed with this sed command:
`find -name "*.lua" -exec sed -i "s/xi\.conq\./xi\.conquest\./g" {} \;`

No error on load, `!updateconquest N` doesnt fail, etc


WinterSolstice8:
apparently xaver beat me, rip

PR #6785, opened 2025-01-21
[core] Use enum instead of defines for Item IDs

claywar:
**_I affirm:_**

Removes successive defines for Item IDs used in core, and replaces with enum.

All impacted items that are currently used revolve around Ninja tools:
1. Test casting spell with no item
2. Test casting spell with item
3. Use universal tool

### DIFF
--- a/src/map/items.h
+++ b/src/map/items.h
@@ -1,32 +1,34 @@
 ﻿#ifndef _ITEMS_H
 #define _ITEMS_H
 
-// Add items as needed, with 14K items in the database this
-// file would be huge if we put all of them in here.
+// Only add items as utilized in core.  There is no need to maintain a complete list
+// here.
 
-// Ninja tools
-#define ITEM_UCHITAKE        (1161)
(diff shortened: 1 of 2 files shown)


WinterSolstice8 (approved):

src/map/items.h
```diff
@@ -1,32 +1,34 @@
 ﻿#ifndef _ITEMS_H
 #define _ITEMS_H
 
-// Add items as needed, with 14K items in the database this
-// file would be huge if we put all of them in here.
+// Only add items as utilized in core.  There is no need to maintain a complete list
+// here.
```
Xaver-DaRed, line 5: I'm assuming this list supposed to be the c++ equivalent of enum/item.lua?

Xaver-DaRed (approved):

src/map/items.h
claywar: Yep, but that define style is definitely not the way to do this in the modern age.  Just updating to something a little more current

PR #6786, opened 2025-01-22
[core] [lua] Implement TH procs

WinterSolstice8:
**_I affirm:_**

Implements TH procs from here https://www.bg-wiki.com/ffxi/Treasure_Hunter
TODO includes SA/TA checking, but seems to be otherwise known complete
Cap TH on gear to 4 for non-THF main, 8 to THF main

Supercedes #3541

Use TH as described in https://www.bg-wiki.com/ffxi/Treasure_Hunter

See TH procs come before enspells and all other forms of add effects, only on first hit (retail source pic)
image

Breakpoint with various job point levels, feint effect levels, etc to see the numbers line up vaguely with retail
image

Check TH level caps as follows (setmod your TH to 10)
image

### DIFF
--- a/src/map/utils/battleutils.cpp
+++ b/src/map/utils/battleutils.cpp
@@ -1127,7 +1127,55 @@ namespace battleutils
 
         if (PAttacker->objtype == TYPE_PC)
         {
-            PChar = (CCharEntity*)PAttacker;
+            PChar = dynamic_cast<CCharEntity*>(PAttacker);
+
+            // TODO: cleanup lua further so we can handle all this add effect stuff somewhere else
+            // Treasure hunter takes priority over enspells
+            if (PChar &&
+                finaldamage > 0 &&
+                isFirstSwing &&
+                PDefender->objtype == TYPE_MOB &&
+                PChar->GetMJob() == JOB_THF &&
+                PChar->hasTrait(TRAITTYPE::TRAIT_TREASURE_HUNTER)) // TH trait as a requirement is assumed, but likely. Could this just be a level 15 check instead?
+            {
+                auto PMob = dynamic_cast<CMobEntity*>(PDefender);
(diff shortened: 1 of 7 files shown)


claywar (approved):

Xaver-DaRed (approved):
I forsee nightmares cleaning all the diferent en-like interactions

But thats a future-us problem

WinterSolstice8:
> I forsee nightmares cleaning all the diferent en-like interactions
> 
> But thats a future-us problem

yeah. It's on my list. we need a way to signify priority. I will probably end up rewriting the "handle enspell" to just call a lua function with a bunch of info fed into it

Xaver-DaRed:
Oh, btw
PR #3541 

May want to take a look at that if you havent already.

And in any case, that one can be closed once this one gets merged.

src/map/utils/battleutils.cpp
```diff
@@ -1127,7 +1127,55 @@ namespace battleutils
...
+            // Treasure hunter takes priority over enspells
+            if (PChar &&
+                finaldamage > 0 &&
+                isFirstSwing &&
+                PDefender->objtype == TYPE_MOB &&
+                PChar->GetMJob() == JOB_THF &&
+                PChar->hasTrait(TRAITTYPE::TRAIT_TREASURE_HUNTER)) // TH trait as a requirement is assumed, but likely. Could this just be a level 15 check instead?
+            {
+                auto PMob = dynamic_cast<CMobEntity*>(PDefender);
+                if (PMob && PMob->m_THLvl < (12 + PChar->getMod(Mod::TREASURE_HUNTER_CAP))) // TH proc cap is 12 + job gifts
```
zach2good, line 1142: Is there scope to add a setting around here for era-style TH application?

zach2good (approved):

zach2good:
(PS. Thanks for replacing my old PR, I was unlikely to finish that this year 😆 )

src/map/utils/battleutils.cpp
WinterSolstice8: The next PR, because this one was touching way too much stuff as it was

PR #6794, opened 2025-01-22
[core] Differentiate between player and mob for getWeaponDmg

CriticalXI:
**_I affirm:_**

- Adds a check to getWeaponDmg lua binding to differentiate between player and mob for the weapon damage calculations. 
   - Currently all mob skills are using the player damage formula for damage calculations when calling getWeaponDmg, not mobs. This should fix that for all skills, and also adds functionality to the WEAPON_BONUS mobmod being applied to various mobs.
- Minor change to move the attack value prints in getstats from defensive section to the offensive section.

- Use !getstats on a mob with WEAPON_BONUS mobmod applied and see that the weapon damage now reports the correct value


WinterSolstice8 (approved):

src/map/lua/lua_baseentity.cpp
```diff
@@ -14507,7 +14507,20 @@ uint16 CLuaBaseEntity::getWeaponDmg()
         return 0;
     }
 
-    return static_cast<CBattleEntity*>(m_PBaseEntity)->GetMainWeaponDmg();
+    uint16 weaponDamage = 0;
+
+    if (m_PBaseEntity->objtype == TYPE_MOB ||
+        m_PBaseEntity->objtype == TYPE_PET)
+    {
```
Xaver-DaRed, line 14515: What about trusts and such? Wouldnt it be easier to just check for PC?

CriticalXI: I was thinking that most things that aren't mobs would use the player damage formula, if I'm mistaken and its the other way around I can reverse the check to be only for PC. I was also curious about that for the player fellows as well.

WinterSolstice8: I feel like this is a problem for another day, I don't think anyone actually knows what the base damage of a trust is. there's no way its calculated based on level alone (like mobs/pets), it's probably completely arbitrary for each and every one

I would throw in a TODO and kick that can down the road

CriticalXI: Got a TODO for the trusts and fellows on there now

Xaver-DaRed (approved):

PR #6796, opened 2025-01-22
[lua, quest] Convert 'Community Service' quest to interaction framework

Grahf0085:
**_I affirm:_**

Converts the quest "Community Service" to the new framework and fixes some bugs like those listed here: https://github.com/AirSkyBoat/AirSkyBoat/issues/2697

I referenced this capture by Siknawz to code the quest: https://www.youtube.com/watch?v=dMCHSUnPDlg

The quest was coded so that even though you could accept the quest at 6pm you couldn't start until a few hours later. From my research I found that this was how the quest was in retail until mid 2005. Since then it was changed so that you could start the quest as soon as you accepted it. So CS 115 - to remind players to start later on after they accepted the quest - was removed

Cut scene 113 was added.  The NPC uses it after you light the lamps until the next morning when the lamps are unlit.

Some code related to the quest remains in Zone.lua:
1. All streetlamps are turned off at 7am
2. Everyone with a membership card is notified that Zauko is recruiting at 6pm
3. If no playable characters is doing the quest by 1am then a NPC named Vhana Ehgaklywha walks around the lights

The quest will only offer you he key item as a reward if you do not already have it. I believe it would be repeatedly offered the way it was coded.

You can now turn in the quest anytime - something the internet seems to agree with. Before you had to turn it in by 1am. You just have to light the lamps by 1am. Talking to the NPC to complete the quest can be done anytime.

Finally in most quests what changes the behavior of NPCs is whether the quest is available, accepted, or completed. That was not the case for this quest due to it being repeatable, only available during certain hours, and only one person on server can do it per night. I try to name variables and conditions appropriately for each check function.

Complete the quest "Community Service" in Lower Jeuno.


Grahf0085:
I know Lua Diagnostics is mad at me about the return but the quest wont work without the return :(

scripts/quests/jeuno/Community_Service.lua
```diff
@@ -0,0 +1,394 @@
...
+                [117] = function(player, csid, option, npc)
+                    local hasCompletedQuest = player:hasCompletedQuest(quest.areaId, quest.questId)
+
+                    if not hasCompletedQuest then -- first time doing quest - get 15 fame here and 15 from quest.rewards for 30 total - repetitions of quest is 15 fame
+                        player:addFame(xi.fameArea.JEUNO, 15)
+                    end
+
+                    if hasCompletedQuest and option == 1 then
+                        -- Lua Diagnostics complains about this return but key item wont be given without a return
+                        return npcUtil.giveKeyItem(player, xi.ki.LAMP_LIGHTERS_MEMBERSHIP_CARD) -- only get this key item for repeating the quest
```
claywar: This is an invalid return.  onEventFinish does _not_ process any return at all.  Just use npcUtil.giveKeyItem directly.

claywar: If you need to exit the function, just use return with no args.

Grahf0085: I spent like an hour trying to get this keyItem. I tried npcUtil.giveKeyItem in other places and it worked without the return. I tried player:addKeyItem here and it didn't work. I can remove the return but just for the record getting the keyitem isn't working for me without the return.

Grahf0085: > This is an invalid return. onEventFinish does _not_ process any return at all. Just use npcUtil.giveKeyItem directly.

And what's even weirder is that I would try getting different key items at this onEventFinish without the return and it worked fine.  Call me crazy.

Grahf0085: > If you need to exit the function, just use return with no args.

This is what I got in game when I returned with no args.
image
image

It says I got the key item but I didn't.

KnowOne134: do you have the update for a return of a keyItem?
``` lua
return quest:keyItem(key-item)
```

if not can use 
```lua
npcUtil.giveKeyItem(key-item)
return quest:noAction()
```

Grahf0085: > do you have the update for a return of a keyItem?
> 
> ```lua
> return quest:keyItem(key-item)
> ```
> 
> if not can use
> 
> ```lua
> npcUtil.giveKeyItem(key-item)
> return quest:noAction()
> ```

I'm not sure what you man by "do you have the update for a return of a keyitem".

But I tried your second block of code and it's not working:

image

image

Grahf0085: DIdnt have my key items open in the screenshot but it's not there.  No key item

KnowOne134: adds some prints in the logic to find where its failing, you sure it has an option of 1 to process? why no option check if you have not completed quest?
and i didnt realise this until now that this is onEventFinish code, there is no need to ever return in this block.
should just be:
```lua
if hasCompletedQuest and option == 1 then
    npcUtil.giveKeyItem(player, xi.ki.LAMP_LIGHTERS_MEMBERSHIP_CARD)
end
```
make sure you dont have the keyItem and everything spelled correctly

Grahf0085: I don't think the logic is failing. It gets to the right area and then prints that I got the key item. I would never get the text that I received the key item if the logic was failing right? 

There is an option check for not completing the quest - line 358 in the last screenshot.

I have tried dozens of times without  a return. It works fine without a return if I change it to other key items instead of the membership card. I can get the membership card if I just move the line to add the key item to an onTrigger. But it will not work without  a return on this onEventFinish. Ill make  a 1 min video this evening showing the quest and code side by side and I'll complete the quest without a return and with a  return.

It takes like 30 seconds to do the quest you just need to change server vars

Grahf0085: > adds some prints in the logic to find where its failing, you sure it has an option of 1 to process? why no option check if you have not completed quest? and i didnt realise this until now that this is onEventFinish code, there is no need to ever return in this block. should just be:
> 
> ```lua
> if hasCompletedQuest and option == 1 then
>     npcUtil.giveKeyItem(player, xi.ki.LAMP_LIGHTERS_MEMBERSHIP_CARD)
> end
> ```
> 
> make sure you dont have the keyItem and everything spelled correctly

https://cloud.atavismxi.com/s/MDf4LxX3ffS7yZy

It's like 40 seconds to complete the quest twice.  The first time I complete it it's the code from this PR and I don't get the key item. The only thing I changed was the times the quest is available. 

Then like half way through you'll see me add a return and then I get the key item.

PR #6800, opened 2025-01-23
[magic] Implement cardinal chant

Xaver-DaRed:
**_I affirm:_**

What title sais
https://www.bg-wiki.com/ffxi/Cardinal_Chant

About the direction enum: It's not made up. Taken from client
imagen

Be a GEO at different levels (for different cardinal chant tiers) and cast nukes, specially -ra spells


zach2good (approved):
Could have sworn we already had this 🤷 , nice work!

PR #6801, opened 2025-01-23
Core: Rollback pending packet changes to raw ptr forms

### LINKED ISSUE #6778: 🐛 Mobs/NPCs not clearing when out of range
**_I affirm:_**

Windows/Linux

## Branch affected by issue 

`base`

## Steps to reproduce

When a mob or npc moves out of your viewing range (50') the model doesn't despawn. You don't get any more updates, but the model stays where it was when it left your view until you get within range of the actual entity again. This only seems to apply when the mob/npc is the one to  move out of your range, as they disappear properly if you move out of their range.

Cactrot Rapido (17244539) and Raminel (17719326) make this easy to test. Just target them and watch as they move 50' away from you.

Maybe something to do with SpawnNPCList and SpawnMOBList. Unknown if pet, trust, and PC are affected as well.

## Expected behavior

Mobs/NPCs should disappear when they get out of your viewing range.

zach2good:
**_I affirm:_**

Might fix https://github.com/LandSandBoat/server/issues/6778

The lifetimes for these were tight before my changes, but they weren't well-defined. This sets them back to what they were, but I'm not happy about it ;_;

Not tested yet, I have a lot of work-work on before the weekend

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -422,11 +422,11 @@ void CCharEntity::pushPacket(std::unique_ptr<CBasicPacket>&& packet)
     {
         if (PendingPositionPacket)
         {
-            PendingPositionPacket = packet->copy();
+            PendingPositionPacket = packet.get();
         }
         else
         {
-            PendingPositionPacket = packet->copy();
+            PendingPositionPacket = packet.get();
             PacketList.emplace_back(std::move(packet));
         }
     }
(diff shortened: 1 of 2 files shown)


zach2good:
I suppose the way to do this with well-defined lifetimes and no raw ptrs would be to not insert new "pending" packets into `PacketList` at all, but to keep them updated and to the side. Then we'd lose their position in the queue of packets to be sent to the player. If this doesn't actually matter, we could unload "pending" packets first when `send_parse` runs, and then the rest of "PacketList".

cocosolos:
Tested this a little, looks like mobs and npcs are despawning properly.

WinterSolstice8 (approved):

Xaver-DaRed (approved):

PR #6818, opened 2025-01-24
[sql, lua] Setup system to move nm spawn points to lua

Frankie-hz:
**_I affirm:_**

**Goal:** Delete the nm_spawn_points.sql. This sql file has caused more numerous issues with having missed updated IDs just like the mob_spawn_mods.sql.

### Zone Initialization Updates
- When a zone starts up (e.g., server boot, maintenance, or after a crash), the `zoneObject.onInitialize` function is called.
- Inside this function, we:
  - `UpdateNMSpawnPoint`: Randomize or update the NMs potential spawn locations.
  - Set Respawn Time: Use `GetMobByID(...):setRespawnTime(...)` to schedule the next spawn after a certain interval.

### Mob Despawn Updates
- When the NM is defeated (`entity.onMobDespawn`), the server needs to do two things again:
   - Re-randomize/Update the NM spawn point to avoid the same spawn location twice.
   - Set a new respawn time.
   
### Drawbacks of the Current System
- Code Duplication/Multiple File References: Because both the zone initialization script and the mobs despawn script need nearly identical logic (updating spawn point and setting respawn time), there is repetition in two different files.

## New Proposed System (most of this is only for timed NMs)
In order to get the respawn time working for `onMobInitialize` we have to check a few things. In our current code we are defaulting the `m_AllowRespawn` to false if the mob is not  `SPAWNTYPE_NORMAL`. This means that anything you put inside the `onMobInitialize` was just being forced to false and was not spawning regardless of the respawn time set (unless scripted elsewhere). By adding a check to see if the current mob is `SPAWNTYPE_SCRIPTED` and checking if the mob as a respawn time forced to it (like onMobInitialize) this solved the problem.

During initial testing I realized that now some mobs were spawning immediately because for some reason they had respawnTimes set inside the sql but also had a spawntype of 128 (Scripted). If a mob is SCRIPTED then the respawn time needs to be 0 in the sql.

### Single Script for everything
- Instead of splitting logic between the zones onInitialize and the mob’s onMobDespawn, move all relevant code (spawn point updates, respawn timer settings) into the mob’s own Lua file. This consolidates the setup and ensures you dont have to maintain a zone-level script for the same NM.

### Random Spawn Function
- Uses a lua function `xi.mob.updateNMSpawnPoint` that draws from a table of possible coordinates or placeholders. This way, you can:
    - Update the NM’s spawn point.
    - Perform all of this without referencing SQL entries for coordinate data or repeating logic in multiple zone scripts.

### Benefits
- Less Duplication: All logic lives in one place -> the mob’s Lua script or the placeholders script.
- Simplified Maintenance: Only one file to edit when adjusting spawn logic or timers.
- You can now easily modify spawn locations, times, or extra conditions without tracking down multiple references in zone files and SQL.

### Lottery NMs
- The only change to lottery NMs will be an extra params you have to define in order to input the table into the current function. 
- This will allow the sql data to be moved to lua as it did with timed NMs.

This will be done over multiple PRs so a failsafe has been added for the still used `UpdateNMSpawnPoint` function.
I have given two examples in this PR, one for Tiamat (a timed NM) and one for Valkurm Emperor (a lottery NM).

The easiest way to test this is to set the `mob:setRespawnTime(math.random(86400, 259200))` in Tiamats lua to something like 120 so you only have to wait 2 minutes. During the server restart it should spawn at the specified tab in any one of the points inside the table.

To test the Lotto NM you will have kill its PH "x" amount of times until the NM spawns (or set its chance to 100) and check the position it pops with the table inside the placeholder lua.


Xaver-DaRed (approved):

scripts/globals/mobs.lua
```diff
@@ -31,13 +31,29 @@ local function lotteryPrimed(phList)
     return false
 end
 
+xi.mob.updateNMSpawnPoint = function(mob, spawnPoints)
+    -- This function is used to replace UpdateNMSpawnPoints() inside the Zone.lua files and the NM despawn scripts
+    -- Once UpdateNMSpawnPoints() is no longer used, this note can be removed
+    -- Spawnpoints is a table of {x = , y = , z = }
+    if spawnPoints ~= nil and #spawnPoints > 0 then
+        local chosenSpawn    = spawnPoints[math.random(#spawnPoints)]
```
zach2good: utils.randomEntry

PR #6822, opened 2025-01-24
[lua] Gambits support multiple predicates

sruon:
**_I affirm:_**

The existing `addSimpleGambit` lua interface only supports setting a single condition and a single action for each gambit.

This makes it impossible to define Gambits that need several conditions to match, or to perform several actions:
- Koru-Moru will spam Refresh on a player with Sublimation activated since the only condition is NOT_STATUS/Refresh.
- Valaineral uses Divine Emblem before Flash, if available.

This PR:
- Adds a new `addGambit` interface:
  - Accepts a table of arbitrary length for predicates, each entry describing the condition and the condition argument - also supports a table describing a single predicate
  - Accepts a table of arbitrary length for actions - also supports a table describing a single action.
  - Adds a deprecation warning to `addSimpleGambit`
  - Supports logic operators (`ai.l.OR` only at the moment) to indicate how to resolve a group of predicates
- Adds `removeGambit` and `removeAllGambits`
  - Adds a deprecation warning to `removeSimpleGambit` and `removeAllSimpleGambits`
- Modifies the `Gambit` class
  - Stores the targetType for the `Gambit` (used to be in predicates)
  - Stores a vector of `PredicateGroup` (instead of `Predicates`)
    - `PredicateGroup` stores a vector of `Predicate`, along with a logic operator indicating how the predicates should be resolved (AND/OR supported at the time)
- Modifies `CGambitsContainer::Tick`
  - Collects all potential targets before executing predicates
  - Will resolve all predicate groups before coming to a conclusion
  - Removed unnecessary duplicate code in the main loop
- Updates `Koru-Moru` to showcase usage
  - Will not refresh someone with Sublimation
  - Will not overwrite `Haste` on ranged characters

Examples:

```
-- Similar to addSimpleGambit, except arguments are grouped in tables
mob:addGambit(ai.t.SELF, { ai.c.MPP_LT, 5 }, { ai.r.JA, ai.s.SPECIFIC, xi.ja.CONVERT })

-- Several predicates, assumed to be AND'd if no logic operator is explicitely used
mob:addGambit(ai.t.CASTER, {
    { ai.c.NOT_STATUS, xi.effect.REFRESH },
    { ai.c.NOT_STATUS, xi.effect.SUBLIMATION_ACTIVATED },
    { ai.c.NOT_STATUS, xi.effect.SUBLIMATION_COMPLETE },
}, { ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.REFRESH })

-- Valaineral: Use Divine Emblem before Flash
-- Note: Not all succession of actions are supported at this time
mob:addGambit(ai.t.TARGET, { { ai.c.NOT_STATUS, xi.effect.FLASH } }, { 
  { ai.r.JA, ai.s.SPECIFIC, xi.ja.DIVINE_EMBLEM }, 
  { ai.r.MA, ai.s.SPECIFIC, xi.magic.spell.FLASH } 
})

-- Predicate group wrapped in a logic operator
-- AAEV: Uses Rampart when her target is under the effects of Chainspell, Manafont, or Astral Flow.
mob:addGambit(ai.t.TARGET, {
    ai.l.OR(
        { ai.c.STATUS, xi.effect.CHAINSPELL },
        { ai.c.STATUS, xi.effect.MANAFONT },
        { ai.c.STATUS, xi.effect.ASTRAL_FLOW }
    ),
}, { ai.r.JA, ai.s.SPECIFIC, xi.ja.RAMPART })
```

While this PR is not enough on its own to bring Trusts to their Retail parity, I believe it offers a framework that can be extended with ease to support the complex decision trees required for certain Trusts.

Enhancements I'd like to add in the near future:
- Deprecate the simple variants
- Update all trusts to use the new interface
- Support several actions seamlessly (i.e. cast 2x March)
- Support an arbitrary amount of predicates nesting
- Support an arbitrary amount of arguments for conditions (i.e. { HAS_EFFECT_COUNT, xi.effect.MARCH, 2 })

Ran it through Tracy and did not notice any meaningful deviation but I am not too familiar with the tool.

https://www.youtube.com/watch?v=FeNDQMbVgOo
https://www.youtube.com/watch?v=-AR1hJBqAms

Summon Koru-Moru as caster job with SCH subjob
Engage a mob
Ensure Koru-Moru casts Refresh when Sublimation is turned off
Ensure Koru-Moru does not cast Refresh when Sublimation is active
Ensure Koru-Moru uses his other gambits as expected

This (somewhat broken) [ArkEV script](https://gist.github.com/sruon/0c97e2de87b6b209e6298420b60f10d3) can be tested and she should use Rampart when a monster uses Manafont/Chainspell/Astral Flow.


scripts/actions/spells/trust/koru-moru.lua
```diff
@@ -24,7 +24,11 @@ spellObject.onMobSpawn = function(mob)
     mob:addSimpleGambit(ai.t.PARTY, ai.c.HPP_LT, 50, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.CURE)
 
     mob:addSimpleGambit(ai.t.MELEE, ai.c.NOT_STATUS, xi.effect.HASTE, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.HASTE)
-    mob:addSimpleGambit(ai.t.CASTER, ai.c.NOT_STATUS, xi.effect.REFRESH, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.REFRESH)
+    mob:addGambit(ai.t.CASTER, {
+        { cond = ai.c.NOT_STATUS, arg = xi.effect.REFRESH },
+        { cond = ai.c.NOT_STATUS, arg = xi.effect.SUBLIMATION_ACTIVATED },
+        { cond = ai.c.NOT_STATUS, arg = xi.effect.SUBLIMATION_COMPLETE },
+    }, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.REFRESH)
```
Xaver-DaRed: This looks good to me.

While you are in this file
imagen

Could you remove the duplicate dia entry?

scripts/actions/spells/trust/koru-moru.lua
```diff
@@ -24,7 +24,11 @@ spellObject.onMobSpawn = function(mob)
     mob:addSimpleGambit(ai.t.PARTY, ai.c.HPP_LT, 50, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.CURE)
 
     mob:addSimpleGambit(ai.t.MELEE, ai.c.NOT_STATUS, xi.effect.HASTE, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.HASTE)
-    mob:addSimpleGambit(ai.t.CASTER, ai.c.NOT_STATUS, xi.effect.REFRESH, ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.REFRESH)
+    mob:addGambit(ai.t.CASTER, {
+        { cond = ai.c.NOT_STATUS, arg = xi.effect.REFRESH },
```
zach2good: I think we don't need `cond` and `arg` here, the gambits are designed to somewhat be read aloud - I suppose the table of multiple conditions is just inferring conjunction and different conditions would be different gambits, inferring disjunction.

src/map/lua/lua_baseentity.cpp
```diff
@@ -15104,6 +15104,51 @@ void CLuaBaseEntity::trustPartyMessage(uint32 message_id)
...
+    {
+        ShowWarning("Invalid Entity calling function (%s).", m_PBaseEntity->getName());
+        return {};
+    }
+
+    using namespace gambits;
+    Gambit_t g;
+
+    auto target = static_cast<G_TARGET>(targ);
+    for (const auto& kvp : predicates)
```
zach2good: Instead of looking these up as key-value pairs, you can do another lookup inside here to look up child tables OR a single condition.

```
if (predicateArgument is a table) // should always be true
    for (object in predicateArgument)
        if (object is a table)
            extract kvp from object
        else
            is just a single condition, extract kvp from predicateArgument
```

zach2good: I think we can also get rid of the idea of "simple" gambits. A gambit is just a gambit at this stage 🤷

src/map/ai/helpers/gambits_container.cpp
```diff
@@ -277,33 +109,39 @@ namespace gambits
 
                 // TODO: This whole section is messy and bonkers
                 // Try and extract target out the first predicate
-                CBattleEntity* target = nullptr;
-                if (gambit.predicates[0].target == G_TARGET::SELF)
+                // All predicates in a gambit should be pointing to the same target
+                // TODO: What's the point of predicates holding the target?
```
zach2good: Good point, I don't remember - might be worth removing this for clarity?

zach2good (commented):
Thanks for picking this up, it's been a long time since I've looked at gambits but this looks like a step in the right direction 👍 

For some historical context: The reason we do this horrible "parse some bullshit syntax into conditions that C++ evaluates" instead of Lua logic is because of how often trust logic is called per-tick and how many trusts there could be in a zone: up to x5 the amount of players.

There was also the vague plan of turning this into a generic "AI" solution for all non-NM mobs, but that has fallen by the wayside.

Any improvements welcome as long as it remains performant 👍

src/map/lua/lua_baseentity.cpp
sruon: > I think we can also get rid of the idea of "simple" gambits. A gambit is just a gambit at this stage 🤷

I would probably make a separate PR updating each trust script to use the new function, once this one is merged.

What's the general policy on deprecating lua functions? I assume we don't want to break custom scripts relying on them.

zach2good: To be a bit ol' killjoy, this is what our license says:

```
This program is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  GNU General Public License for more details.
```

If it isn't in our repo we don't make any guarantees about it, the software is too far from being stable/done for us to guarantee anything.

So, break whatever you feel like!

sruon:
Round 2

- Made changes to `addGambit` 
  - Takes a table of actions, indicating multiple actions to be taken as reaction
    - This needs some more fixes in the `GambitContainer`, JA -> MA works for the most part but it doesn't ensure the MA is ready to use. MA -> MA doesn't work because of the cast time. Need some sort of queuing on the trust directly that could be checked before processing any other gambit.
  - Supports several syntax for the predicates and actions

```
{ ai.c.NOT_STATUS, xi.effect.REFRESH }
{ { ai.c.NOT_STATUS, xi.effect.REFRESH } }
{ { ai.c.NOT_STATUS, xi.effect.REFRESH }, { ai.c.NOT_STATUS, xi.effect.BALLAD } }
```

```
{ ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.REFRESH }
{ { ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.REFRESH } }
{ { ai.r.JA, ai.s.SPECIFIC, xi.ja.DIVINE_EMBLEM }, { ai.r.MA, ai.SPECIFIC, xi.spell.FLASH } }
```

- Introduces logic operators to group several predicates together
  - Only **OR** is supported at the moment
  - Ungrouped predicates are assumed to be **AND**

```
-- Uses Rampart when her target is under the effects of Chainspell, Manafont, or Astral Flow.
mob:addGambit(ai.t.TARGET, {
    ai.l.OR(
        { ai.c.STATUS, xi.effect.CHAINSPELL },
        { ai.c.STATUS, xi.effect.MANAFONT },
        { ai.c.STATUS, xi.effect.ASTRAL_FLOW }
    ),
}, { ai.r.JA, ai.s.SPECIFIC, xi.ja.RAMPART })
```

- Changes to `Gambit`
  - Stores a vector of `PredicateGroup`
    - `PredicateGroup `stores a vector of `Predicate` along with the logic operator
  - Moves target selector from the predicates into the main `Gambit` object

- Modified `GambitContainer::Tick()` and associated functions 
  - Will resolve all predicate groups and apply their logic operator
  - The design only supports a single predicate group at the moment but should be extendable to complex decision trees
  - Does not loop anymore for each action associated with a Gambit
    - If conditions resolved, it is assumed all actions must be performed

- Mades slight changes to `addSimpleGambit` to maintain compatibility for the time being

Been trying to implement several Trusts by following the wiki page outlining their behavior such as [ArkEV](https://gist.github.com/sruon/0c97e2de87b6b209e6298420b60f10d3) and overall quite happy with the [results](https://youtu.be/-AR1hJBqAms).

src/map/ai/helpers/gambits_container.cpp
```diff
@@ -673,177 +537,199 @@ namespace gambits
...
+                return std::ranges::all_of(predicateResults, [](const bool result)
+                                           { return result; });
             }
-            case G_CONDITION::HP_MISSING:
+            case G_LOGIC::OR:
             {
-                return (trigger_target->health.maxhp - trigger_target->health.hp) >= (int16)predicate.condition_arg;
-                break;
+                return std::ranges::any_of(predicateResults, [](const bool result)
+                                           { return result; });
```
zach2good, line 732: Fancy!

zach2good (commented):
Looking pretty cool 👍 

What I'd do if you're not comfortable being the one to remove `addSimpleGambit` is to just make `addSimpleGambit` forward all of its args to `addGambit` and print a message saying that it's deprecated and to use `addGambit` instead

sruon:
> Looking pretty cool 👍
> 
> What I'd do if you're not comfortable being the one to remove `addSimpleGambit` is to just make `addSimpleGambit` forward all of its args to `addGambit` and print a message saying that it's deprecated and to use `addGambit` instead

I absolutely want to remove `addSimpleGambit ` but that would make the PR a little too large. I will make it so that it forwards arguments to `addGambit` (+deprecation warning) and then in a separate PR update all the trust scripts and remove it.

zach2good (approved):
I think this multiple table entry handling will eventually lead to:
```
gambit:
    target
    { n conditions }
    { n reactions } -- where (target) and (n conditions) are checked for every reaction/step before execution
```

For things like Iroha's self-skillchains, etc.

scripts/actions/spells/trust/koru-moru.lua
```diff
@@ -19,26 +19,32 @@ spellObject.onMobSpawn = function(mob)
...
+    mob:addGambit(ai.t.CASTER, {
+        { ai.c.NOT_STATUS, xi.effect.REFRESH },
+        { ai.c.NOT_STATUS, xi.effect.SUBLIMATION_ACTIVATED },
+        { ai.c.NOT_STATUS, xi.effect.SUBLIMATION_COMPLETE },
+    }, { ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.REFRESH })
+    mob:addGambit(ai.t.TANK, { ai.c.NOT_STATUS, xi.effect.REFRESH }, { ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.REFRESH })
+    mob:addGambit(ai.t.RANGED, {
+            { ai.c.NOT_STATUS, xi.effect.FLURRY_II },
+            { ai.c.NOT_STATUS, xi.effect.HASTE }, -- No overwriting Haste
+        }, { ai.r.MA, ai.s.HIGHEST, xi.magic.spellFamily.FLURRY }) -- xi.effect.FLURRY_II is not a typo
```
WinterSolstice8: This comment probably needs to be moved up to the line where it actually has the flurry_II effect

Xaver-DaRed (approved):

PR #6825, opened 2025-01-24
[Lua] Halvung Bombs + Big Bomb NM

dallano:
**_I affirm:_**

- Big Bomb NM behaviour
- Shared common behaviour of Friar's Lantern mobs in Halvung Bomb ToAU skill list
- Heat Wave mobskill

## More Details:
- Heat Wave's power was determined by level / 2.05 as witnessed difference between Big Bomb + Friar's Lantern.  This value is witnessed to be 39 from big bomb, and 37 from Friar's Lanterns. (81 / 2.05 = 39, 77 / 2.05 = 37). This can be witnessed in linked personal captures
- Damage threshold of bomb size increase can be witnessed in all videos. I theorized critical might have a chance to trigger, but this was disproven in personal captures.
- HP value of big bomb can be witnessed in captures
- Created ToAU bomb mob skill list, as bombs in this era have unique skill list to earlier expansions.
- Attack and magic attack is a given with each increase in size. However this value is experimental and not a proven value. However in testing this value was comparable to the captures.
- Adjust Big Bomb's spell list to be Bomb Queen's as they are identical - thus saving the requirement of a new mob spell list

## Sources used:
Big bomb fight: https://www.youtube.com/watch?v=huhMJlo9EcY
Captures: https://youtu.be/X_1qLq81oWI   -   https://1drv.ms/u/c/87e665e10555c452/EVLEVQXhZeYggIfGAQAAAAAB-Y31wIIUy5BPJcWcoYOCVQ?e=i8jFUw 
Personal Captures: https://1drv.ms/f/s!AtUZibtOtozEhf1aynJwm5DhxJbMSw?e=d04BIJ

Fight the specific Friar's Lanterns in Halvung / Big Bomb to witness behaviour of mixin


scripts/zones/Halvung/mobs/Big_Bomb.lua
```diff
@@ -2,11 +2,17 @@
...
 local entity = {}
 
-entity.onMobInitialize = function(mob)
-    mob:setMobMod(xi.mobMod.IDLE_DESPAWN, 900)
+entity.onMobSpawn = function(mob)
+    mob:setMobMod(xi.mobMod.IDLE_DESPAWN, 180)
+    mob:addImmunity(xi.immunity.LIGHT_SLEEP)
+    mob:addImmunity(xi.immunity.DARK_SLEEP)
+    mob:addImmunity(xi.immunity.STUN)
+    mob:setMod(xi.mod.SILENCERES, 80) -- Approximation - requires further experimentation
```
Xaver-DaRed: This modifier is used for "Resist Silence" job trait, almost exclusively.
Did the message return `Resist! The big bomb resisted...` When casting Silence on the mob? (The `Resist!` at the beggining is whats important.)

After seing the videos I didnt see the trait being triggered. use SILENCE_MEVA modifier at a 100 or even 200 value as a placeholder until we implement status effect resistance ranks

Xaver-DaRed:
Cant access second link

sql/mob_skill_lists.sql
```diff
@@ -3950,7 +3950,13 @@ INSERT INTO `mob_skill_lists` VALUES ('Unfettered_Twitherym',2017,2952); -- blac
 
 INSERT INTO `mob_skill_lists` VALUES ('Unfettered_Twitherym_low_50',2018,2953); -- smouldering_swarm
 
--- Next available ID: 2019
+INSERT INTO `mob_skill_lists` VALUES ('Bomb_TOAU',2019,511); -- Berserk
```
Xaver-DaRed: From your own capture linked, the skill ID for berserk used by this mob is ID 593
`[Actor: 17031401 (Big Bomb)] Berserk > Cat: 11 ID: 593 Anim: 879 Msg: 194`

dallano:
Ah damn. Unfortunately it appears I lost the video I used during my captures. However the files themselves should have a majority of the content I witnessed.

scripts/zones/Halvung/mobs/Big_Bomb.lua
dallano: Upon re-reviewing the captures done by Siknawz, it appears Big Bomb is also immune to terror, and disease.

However, with silence it simply returned Resists... not Resist!

I will commit terror and disease immunities. Am I using the appropriate mod for silence? Or should I be using SILENCE_MEVA

sql/mob_skill_lists.sql
dallano: Interesting. I wasn't able to discern any new behaviors in berserk.

scripts/zones/Halvung/mobs/Big_Bomb.lua
Xaver-DaRed: I edited my original message before I saw your answer, but I'll repeat it.

Use SILENCE_MEVA at a 100-200 value until we get status effect resistance ranks implemented

sql/mob_skill_lists.sql
Xaver-DaRed: There are many copies of TP moves. Probably a mere matter of having different power than other copies

dallano: I can create a similar "dummy" file to the other bomb skills that were preexisting this PR. they were simply p1_hellstorm, p1_vulcanian_impact, etc.
That way the ID is reserved, and a separate file is made for further balancing if needed.

dallano: Similarily, heat wave from big bomb was ID 595

Xaver-DaRed: For now just correct the ID here in this skill list.

dallano: Before I commit changes, I'm going to get another capture of Friar's Lantern behaviors + berserk ID to confirm all is well.
Will link my findings!

dallano:
I found some fascinating behaviour with these bombs. (Curse them)

So the bombs that don't drop smoke filled glasses behave like any regular bomb (Self-destruct and Berserk only)
While the bombs that do drop the flask have access to the "ToAU" skills. HOWEVER. after getting them to weaponskill roughly 100 times, heat wave was ONLY witnessed when animation sub was greater than the default sub, NEVER saw a self-destruct at default sub, and only saw self-destruct when bomb is at its biggest size.

Will continue to do some testing and captures to confirm this behaviour before committing to the PR.

dallano:
Okay!! Some crazy discoveries were made by going in further depth:

These bombs use different abilities based on their animation sub. As an example they will only use berserk when default size, and heat wave isn't even usable until it has grown twice. I linked further captures in the link above and all evidence can be witnessed. Took a few hours of capture to prove this theory. That is why they use different IDs for their mobskills.

Commit with all changes regarding comments as well as new found behaviour incoming!

scripts/zones/Halvung/mobs/Big_Bomb.lua
dallano: All requested changes made <3

sruon:
What's the status on this PR? It would close #7433

dallano:
I'll rebase and resolve merge conflicts once I get wifi back here (Just moved)

dallano:
Going to reopen this PR

PR #6831, opened 2025-01-24
Use PacketList directly when sending

cocosolos:
**_I affirm:_**

Couple edge case fixes if a pending packet ends up being null. Mainly though this makes it so we use the PacketList directly when preparing packets to send instead of making a copy and using that. We currently make a copy of the packet list, process those copies, then delete the real packets. This apparently causes some sync issues (#6829) that should be fixed now.

I'm not sure why exactly this issue seems to be consistent, so this may require some further review, but this seems to fix it.

Log in with 2 characters, go to the same zone, both characters can see each other.

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -396,6 +396,11 @@ bool CCharEntity::isPacketListEmpty()
     return PacketList.empty();
 }
 
+auto CCharEntity::getPacketList() const -> const std::deque<std::unique_ptr<CBasicPacket>>&
+{
+    return PacketList;
+}
+
 void CCharEntity::clearPacketList()
 {
     while (!PacketList.empty())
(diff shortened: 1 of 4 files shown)


src/map/entities/charentity.cpp
```diff
@@ -439,40 +439,34 @@ void CCharEntity::pushPacket(std::unique_ptr<CBasicPacket>&& packet)
 void CCharEntity::updateCharPacket(CCharEntity* PChar, ENTITYUPDATE type, uint8 updatemask)
 {
     auto existing = PendingCharPackets.find(PChar->id);
-    if (existing == PendingCharPackets.end())
+    if (existing == PendingCharPackets.end() || !existing->second)
```
zach2good: I find this logical block quite hard to reason about, can you break these up even if it ends up with nested conditions?

I think it would also help to rename existing to itr, and break out conditions like:

`const bool hasPendingPacket = itr != PendingCharPackets.end();`

and to not use the itr pair directly, but to use:

`auto& pendingPacket = existing->second;`
`pendingPacket ->updateWith...`

src/map/entities/charentity.h
```diff
@@ -542,8 +541,7 @@ class CCharEntity : public CBattleEntity
 
     CItemEquipment* getEquip(SLOTTYPE slot);
 
-    // TODO: Don't use raw ptrs for this, but don't duplicate whole packets with unique_ptr either.
-    CBasicPacket* PendingPositionPacket = nullptr;
+    std::deque<std::unique_ptr<CBasicPacket>> PacketList; // The list of packets to be sent to the character during the next network cycle
```
zach2good: instead of making this public, can you keep it private and access it with: `auto getPacketList() -> std::deque<std::unique_ptr<CBasicPacket>>&`

src/map/map.cpp
```diff
@@ -1049,14 +1049,12 @@ int32 send_parse(int8* buff, size_t* buffsize, sockaddr_in* from, map_session_da
...
         do
         {
-            *buffsize       = FFXI_HEADER_SIZE;
-            auto packetList = PChar->getPacketListCopy();
-            packets         = 0;
+            *buffsize = FFXI_HEADER_SIZE;
+            packets   = 0;
 
-            while (!packetList.empty() && *buffsize + packetList.front()->getSize() < MAX_BUFFER_SIZE && static_cast<size_t>(packets) < PacketCount)
+            while (!PChar->PacketList.empty() && *buffsize + PChar->PacketList.front()->getSize() < MAX_BUFFER_SIZE && static_cast<size_t>(packets) < PacketCount)
```
zach2good: `isPacketListEmpty`

zach2good (approved):
Logically seems fine, I'd just like it to be made as clear and obvious about what's going on at every step as possible

src/map/entities/charentity.cpp
cocosolos: Updated the logic here, hopefully more clear.

src/map/entities/charentity.h
cocosolos: Good call, made it const too.

BoundingS:
This seems to cause a very game breaking bug if items get out of sync between mog house and player inventory, gil etc.  Would you take a look at the issue I posted and see what you think?

zach2good:
Please stop commenting on closed PRs, it's very hard for us to track. Open new issues and link prs you want to talk about

PR #6840, opened 2025-01-25
[sql] add content tags to about 200 COP era recipes

Grahf0085:
**_I affirm:_**

Adds about 200 content_tags to synthesis recipes from COP.  A little more than 200 to go.

EDIT: Also some of the sushi recipes that were added during TOAU were marked as TOAU. They are very similar to the COP era recipes for sushi but use fish from TOAU areas.

Run dbtool


sql/synth_recipes.sql
```diff
@@ -466,10 +466,10 @@ INSERT INTO `synth_recipes` VALUES (4015,0,0,83,0,0,0,0,0,0,0,4101,4243,695,721,
 INSERT INTO `synth_recipes` VALUES (4016,0,0,83,0,0,0,0,0,0,0,4099,4241,721,721,721,2533,2533,2533,2533,2535,358,358,358,358,1,1,1,1,'Credenza',NULL);
 INSERT INTO `synth_recipes` VALUES (4017,0,0,84,0,0,0,0,0,54,0,4099,4241,704,710,927,927,927,932,937,4109,18234,18234,18234,18234,66,99,99,99,'Mega Pump',NULL);
 INSERT INTO `synth_recipes` VALUES (4018,1,0,84,0,0,0,0,0,0,0,4100,4242,17056,0,0,0,0,0,0,0,720,720,844,844,1,1,1,1,'Mythic Wand (desynth)',NULL);
-INSERT INTO `synth_recipes` VALUES (4019,0,0,84,0,0,0,0,0,0,0,4099,4241,9486,0,0,0,0,0,0,0,12409,12409,12409,12409,1,1,1,1,'Numinous Shield',NULL);
+INSERT INTO `synth_recipes` VALUES (4019,0,0,84,0,0,0,0,0,0,0,4099,4241,9486,0,0,0,0,0,0,0,12409,12409,12409,12409,1,1,1,1,'Numinous Shield','COP');
```
Frankie-hz: This is not COP - this is the woodworking kit synth

Grahf0085: Thanks. It's so many numbers.  I ran this query and found one other recipe labeled as COP that uses a kit:
```
SELECT * FROM synth_recipes
WHERE content_tag = 'COP'
AND Ingredient1 BETWEEN 8805 AND 9523
```

Frankie-hz (approved):
Rest looks right

zach2good (approved):

PR #6795, opened 2025-01-22
[lua, sql] HNM Kings Rework

CriticalXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

- Reworks Adamantoise, Aspidochelone, Behemoth, King Behemoth, Fafnir, and Nidhogg to have correct stats and behavior

Stats and mob skill adjusts are based off of Jimmayus's monster stats capture sheet
Behe/KB: https://www.youtube.com/watch?v=uEayfQTiaac
Adamantoise/Aspis: https://www.youtube.com/watch?v=-5nZcruW-3w

## Steps to test these changes

Spawn each HNM and see that it reflects the changes listed

PR #6846, opened 2025-01-25
[SQL] Audit HNM King and Wyrm Drops

LacunaXI:
**_I affirm:_**

As discussed in #6795. This PR updates the drops for HNM land kings.


CriticalXI:
Somewhat related to this, but I will probably be getting to the rest of the HNMs in CoP or older zones over the next month. I would appreciate it if you could also submit these droplist reworks for things like Tiamat, Jormungand, etc if you have them.

LacunaXI:
> Somewhat related to this, but I will probably be getting to the rest of the HNMs in CoP or older zones over the next month. I would appreciate it if you could also submit these droplist reworks for things like Tiamat, Jormungand, etc if you have them.

I worked on all 3300+ drop tables. It is just a matter of submitting and merging.

I will add those Wyrms to this PR as well then. These were tricky but the shared aspects of certain rewards between each made it viable to patch every different wyrm video into a commonality. Searching for every video was the only option as FFXIDB has almost 0 data to work with.

If you ever hear a young person say they were born in the wrong time or era. Show them these old 240-360p HNM videos made with Movie Maker transitions. Where they overdubbed Lincoln Park and Evanescence or something silly like the Hampster Dance or Crazy Frog.

Xaver-DaRed (approved):

PR #6847, opened 2025-01-25
[sql] Adding Judges Lure, Judge Fly, Judge Minnow, and Goliath Worm to the …

StabbyCutyou:
…fishing_bait table

**_I affirm:_**

This PR enables the 3 Judge baits/lures alongside Goliath Worm to actually be usable when attempting to fish. 

Currently, [this check](https://github.com/LandSandBoat/server/blob/base/src/map/utils/fishingutils.cpp#L2030) will fail and prevent fishing altogether if you attempt to use these items.

For the 3 Judge items, there is not a practical way for me to verify any of their behaviors in retail, and so i've left their flags/configuration as vanilla as possible. I noticed that there are other baits/tackles/etc without entries in the affinities table, and I have confirmed that the Judge Lure will enable you to catch fish without any affinities, so I assume this should functionally be ok.

For the Goliath Worm, it seems to only be involved in a [WOTG quest ](https://ffxiclopedia.fandom.com/wiki/Songbirds_in_a_Snowstorm) and it is used to fish up key items. 

I could not find much else out about it, but a [Talk Page](https://ffxiclopedia.fandom.com/wiki/Talk:Goliath_Worm) entry suggests that it might not be able to fish anything up at all (or, that no one has mentioned it). Given that it seems to be used specifically for the quest only, it also seemed reasonable to set no affinities and keep the configuration otherwise vanilla. 

Grant yourself GM Status
Turn on GM Mode (optional)
Give yourself the Judge Rod, alongside any of the Judge baits/lures
Equip them
Be able to start the minigame
Give yourself a Goliath Worm
Equip it
Be able to start the minigame


claywar (approved):

StabbyCutyou:
I can fix the commit title, one moment.

Xaver-DaRed (approved):

Issue #6842, opened 2025-01-25
🐛 Something is VERY wrong with packets, GIL, items, Mog House, delivery?


BoundingS:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base` - current through now.

## Steps to reproduce

Any change in items between Mog House and character outside seems to throw everything out of sync.  I sent via delivery 3 items from one character to another.  I logged in on the 2nd character to retrieve these items.  Upon taking them in the Mog House deliver box, the 3rd item gave no retrieved message.  At that point, all 3 items disappeared from inventory.

I walked outside the Mog House.  Now 2 of the 3 items that gave the message appeared but not the 3rd.  However gil was now at 0.  I gave myself 1 gil, the gil came back using givegil.

I went back into a new zone, North Sandy, all items were back except the gil.  Back into moghouse, all items gone.  Something very crazy is going on.

Both characters affected.

Zoned to Ru Lude Garden to test Mog House there, all NPCs gone.  Mog House not functional.  !zone to Lower Jueno and walk back to Ru Lude, NPCs back, Mog House working.  All items still unsynced.  

Exited Mog House, all NPCs and save crystal gone.

Rolled server back to a few days ago, seems to clear up all issues.

## Expected behavior

All items and gil should sync always and NPCs shouldn't fail to load.

WinterSolstice8:
Sent items/gils to myself, logged back out and back in, got all the items and gil no problem. NPCs work fine, can't reproduce. The missing items until you zone is an ancient bug that nobody's been able to reliably reproduce, so it's not very fixable until we can figure that out (if you can, please let me know)

![Image](https://github.com/user-attachments/assets/e967ba4e-a21a-4919-adfe-305f9b945726)

![Image](https://github.com/user-attachments/assets/bba3525e-41e0-417b-a849-91c3db338ede)

Note: there's zero functional difference between sending to yourself and another char if you logout in between

That said --
[...]

PR #6848, opened 2025-01-26
Packet handling fixes

cocosolos:
**_I affirm:_**

- Use a copy of the PacketList in case there are any issues while sending.
- Pop packets from PacketList only after finished sending.
- Fix an issue with players using the wrong update packet.

Related issues: #6842, #6829, #6778, #6746

Zone in to Ru'Lude and see stuff. Zone in near another player and see eachother. Watch as things disappear when they get too far.

### DIFF
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -163,7 +163,7 @@ void CZoneEntities::TryAddToNearbySpawnLists(CBaseEntity* PEntity)
                 case TYPE_PC:
                 {
                     PCurrentChar->SpawnPCList[PEntity->id] = PEntity;
-                    PCurrentChar->updateEntityPacket(PEntity, ENTITY_SPAWN, UPDATE_ALL_CHAR);
+                    PCurrentChar->updateCharPacket(static_cast<CCharEntity*>(PEntity), ENTITY_SPAWN, UPDATE_ALL_CHAR);
                     break;
                 }
                 case TYPE_NPC:
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -401,6 +401,16 @@ auto CCharEntity::getPacketList() const -> const std::deque<std::unique_ptr<CBas
     return PacketList;
 }
 
+auto CCharEntity::getPacketListCopy() -> std::deque<std::unique_ptr<CBasicPacket>>
+{
+    std::deque<std::unique_ptr<CBasicPacket>> PacketListCopy;
+    for (const auto& packet : PacketList)
+    {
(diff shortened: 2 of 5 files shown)


BoundingS:
Thank you for looking at this!

cocosolos:
I think this fixes all the linked issues, though I'm too tired to thoroughly test right now. It looks like #6751 is the source of the player update issues. `CZoneEntities::TryAddToNearbySpawnLists` had PCs using `updateEntityPacket` instead of `updateCharPacket` (I was actually working on combining these 2 and doing some renaming when this issue came up, will have a followup maybe tomorrow for that + some other stuff).

NPCs load in Ru'Lude, players are visible when using !goto or zoning, NPCs disappear when they go out of range. IDK if I'm missing anything but seems alright.

src/map/zone_entities.cpp
```diff
@@ -163,7 +163,7 @@ void CZoneEntities::TryAddToNearbySpawnLists(CBaseEntity* PEntity)
                 case TYPE_PC:
                 {
                     PCurrentChar->SpawnPCList[PEntity->id] = PEntity;
-                    PCurrentChar->updateEntityPacket(PEntity, ENTITY_SPAWN, UPDATE_ALL_CHAR);
+                    PCurrentChar->updateCharPacket(static_cast<CCharEntity*>(PEntity), ENTITY_SPAWN, UPDATE_ALL_CHAR);
```
WinterSolstice8, line 166: This must have been a needle in a haystack. good job

WinterSolstice8 (approved):

src/map/zone_entities.cpp
cocosolos: TY, before this issue was reported I was actually working on combining `updateCharPacket` into `updateEntityPacket` so it was pretty fresh in my mind. The naming is pretty confusing so it'll be nice to just do it like it was here and have everything use `updateEntityPacket`. But the branch I was working on didn't have that merged so I was dealing with different layers of bugs interacting without realizing it for a while.

zach2good: PEntity and PCurrentChar are the same pointer here, static_cast is just type system shenanigans. This should just be PCurrentChar?

zach2good: Oh shit, I see it! YOU ABSOLUTE DETECTIVE!

zach2good (approved):

zach2good:
This also seems to now finish the job of my fix for "other players won't be visible if I zone next to them" due to the `TryInsertIntoSpawnList` call now doing what it's meant to

PR #6858, opened 2025-01-26
[lua] Shami/Shemo accept different seals type in single trade

sruon:
**_I affirm:_**

Make Shami and Shemo match their [retail behavior](https://www.youtube.com/watch?v=gv6_zycKxLA)

- Shami will accept several type of seals in a single trade (i.e. Beastmen's + Kindred)
- Enable trading to Shemo with similar logic
- Sets event parameter to 1 if you've used their service before, skipping their first dialog/menu

https://youtu.be/7ftGnHHseg0

```
!zone Port Jeuno
!gotoname Shami
!additem BEASTMENS_SEAL 99
!additem KINDREDS_SEAL 99
```

Trade 1 Beastmen's seal to Shami, ensure Shami accepts
Trade 1 Beastmen's seal + 1 Kindred's seal to Shami, ensure Shami accepts with a slightly different dialog
Trade garbage item along with seals, ensure the extra items are not lost
Purchase an orb from Shami and talk to him again, ensure the initial dialog is skipped

Trade 1 Beastmen's seal to Shemo, ensure Shemo accepts
Trade 1 Beastmen's seal + 1 Kindred's seal to Shemo, ensure Shemo accepts with a slightly different dialog
Trade garbage item along with seals, ensure the extra items are not lost
Trade seals with Shemo and talk to him again, ensure the initial menu is different

### DIFF
--- a/scripts/globals/seals.lua
+++ b/scripts/globals/seals.lua
@@ -1,0 +1,59 @@
+-----------------------------------
+-- Seals and Crests trade (Shami/Shemo)
+-----------------------------------
+xi = xi or {}
+xi.seals = xi.seals or {}
+
+---@class sealItems
+---@field [xi.item] { [integer]: integer, [integer]: integer }
+xi.seals.sealItems =
+{
+    -- Trade Item ID              Seal ID, Retrieve Option
+    [xi.item.BEASTMENS_SEAL       ] = { 0, 2 },
+    [xi.item.KINDREDS_SEAL        ] = { 1, 1 },
+    [xi.item.KINDREDS_CREST       ] = { 2, 3 },
+    [xi.item.HIGH_KINDREDS_CREST  ] = { 3, 4 },
+    [xi.item.SACRED_KINDREDS_CREST] = { 4, 5 },
+}
+
+---@nodiscard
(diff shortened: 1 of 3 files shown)


scripts/globals/seals.lua
```diff
@@ -0,0 +1,59 @@
...
+    [xi.item.BEASTMENS_SEAL       ] = { 0, 2 },
+    [xi.item.KINDREDS_SEAL        ] = { 1, 1 },
+    [xi.item.KINDREDS_CREST       ] = { 2, 3 },
+    [xi.item.HIGH_KINDREDS_CREST  ] = { 3, 4 },
+    [xi.item.SACRED_KINDREDS_CREST] = { 4, 5 },
+}
+
+---@nodiscard
+---@param trade CTradeContainer
+---@return table?
```
claywar: This shouldn't be nil-able, since it will always return a table (just perhaps an empty one)

sruon: Pushed an update for this but I noticed another logic issue. Will add a commit shortly.

WinterSolstice8 (approved):

PR #6859, opened 2025-01-27
Quest: Tigers Teeth to IF

KnowOne134:
**_I affirm:_**

moves quest to interaction framework.
Adds missing pre-quest default dialog.
Adds missing wrong trade CS.

!additem BLACK_TIGER_FANG
!pos -140 -6 -8 230
!delquest 0 23


scripts/quests/sandoria/Tigers_Teeth.lua
```diff
@@ -0,0 +1,99 @@
...
+                end,
+
+                onTrigger = quest:event(575):replaceDefault(),
+            },
+
+            onEventFinish =
+            {
+                [572] = function(player, csid, option, npc)
+                    quest:complete(player)
+                end,
```
Xaver-DaRed, line 65: Shouldnt this complete the trade aswell?

KnowOne134: explained up top it compltes here and still runs the event finish 572 in the quest_completed section. just a cool trickle down effect and why i have the fame for completes at 25 and you get the 5 fame in the completed section to addup to 30, the completed trade hits there as well

Xaver-DaRed: Could you add a comment in here stating that both onEventFinish, both in different sections, execute?

Its a bit counter-intuitive, tbh. Not saying its wrong, but it has certanly taken me by surprise

KnowOne134: i changed it to local var checks and not use the trickle down effect separating the sections

Xaver-DaRed (approved):

PR #6876, opened 2025-01-28
Chore: Tidy SQL usage in login, search, and world servers

zach2good:
**_I affirm:_**

Also tidies multiline strings

Testing:
- [x] Make a new character
- [x] Change password
- [x] Delete a character
- [x] Log in
- [x] /sea all
- [x] Send tell
- [x] Join party
- [x] Send party chat
- [x] Add someone to blacklist
- [x] Create and name a linkshell
- [x] Make a linkpearl and trade to someone else
- [x] Both equip linkpearls
- [x] Send linkshell chat - seen by all
- [x] List something on the AH
- [x] Buy something on the AH - get item
- [x] De-list something from the AH
- [x] AH: Price history shows
- [x] AH: Current listings show
- [x] AH: Listing per-item show
- [x] AH: Have someone buy something - gil shows up in dbox
- [x] Send something to another char in the dbox
- [x] Get something from another char in the dbox
- [x] Daily tally
- [x] Conquest
- [x] Cancel dbox sends

### DIFF
--- a/src/world/message_server.cpp
+++ b/src/world/message_server.cpp
@@ -113,7 +113,8 @@ std::string ipp_to_string(uint64 ipp)
     char target_address[INET_ADDRSTRLEN];
     inet_ntop(AF_INET, &target, target_address, INET_ADDRSTRLEN);
 
-    return fmt::format("{}:{}", str(target_address), port);
+    // This is internal, so we can trust it.
+    return fmt::format("{}:{}", asStringFromUntrustedSource(target_address), port);
 }
 
 void message_server_parse(MSGSERVTYPE type, zmq::message_t* extra, zmq::message_t* packet, zmq::message_t* from)
@@ -139,8 +140,9 @@ void message_server_parse(MSGSERVTYPE type, zmq::message_t* extra, zmq::message_
             return;
         }
 
+        // This is internal, so we can trust it.
         ShowDebug(fmt::format("Message: Received message {} ({}) from {}:{}",
-                              msgTypeToStr(type), static_cast<uint8>(type), str(from_address), from_port));
+                              msgTypeToStr(type), static_cast<uint8>(type), asStringFromUntrustedSource(from_address), from_port));
 
         while (rset->next())
         {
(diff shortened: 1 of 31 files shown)


src/world/message_server.cpp
```diff
@@ -351,15 +361,12 @@ void message_server_listen(bool const& requestExit)
...
-
-        // TODO: Is this still needed?
-        // sql_->TryPing();
     }
 }
 
 void cache_zone_settings()
 {
-    auto rset = db::query("SELECT zoneid, zoneip, zoneport, misc FROM zone_settings");
+    auto rset = db::preparedStmt("SELECT zoneid, zoneip, zoneport, misc FROM zone_settings");
```
StabbyCutyou, line 372: Since there are no params here, you should be fine to keep using query (unless the idea is to future proof against someone parameterizing this later)

WinterSolstice8: Prepared statements are faster because the database engine knows ahead of time about the query, so there is actually a beneficial change here

StabbyCutyou: Fair, however I thought the query cache usually provides that benefit (regardless of how the client invokes the call), but you'd be more familiar with mariadb and the c driver than I am, so I am happy to be told I am in the wrong here.

zach2good: For a query like this that has no params - sure, it doesn't really matter. As a general rule - the MariaDB C and C++ drivers aren't as battle-tested as we'd like, and a user could misconfigure their MariaDB service in any number of wacky ways, so I'd rather enforce our usage patterns through the code we control.

I'd like to eventually get to a point where there's only a single entry point to the db: `preparedStmt`. It allows us to do fancy things like make the call `constexpr` and register all the statements on startup etc.

Also performance and security, etc.

StabbyCutyou: That all makes sense. The main gain here being defense against injections for everything else, sticking to one entry point for all calls is a reasonable default. 

Will continue to lurk and stop interrupting y'all.

claywar (approved):

Xaver-DaRed (approved):

cocosolos:
Is that migration necessary? Triggers should be dropped when the table is dropped (it's not protected so it'll get dropped on update).

zach2good:
But when the trigger names change, the old one will be left behind forever

cocosolos:
I _think_ the triggers will be dropped when the associated table is dropped. So when `synth_recipes` is dropped, the associated trigger is also dropped without needing to manually drop it, then the new one gets imported with the new name.

https://dev.mysql.com/doc/refman/8.4/en/drop-table.html
> Dropping a table also drops any triggers for the table.

cocosolos:
Triggers on protected tables probably need a migration though, since those tables aren't dropped on update.

zach2good:
> > https://dev.mysql.com/doc/refman/8.4/en/drop-table.html
> 
> Dropping a table also drops any triggers for the table.

Ohh great, thanks @cocosolos

PR #6899, opened 2025-01-30
Lua: Lower Jeuno Gobbie box shouldn't show Tenshodo Coffer options

zach2good:
**_I affirm:_**

They're named the same, so this is a quick patch to stop the inert gobbie box coffer showing the hidden tenshodo box options

### DIFF
--- a/scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
+++ b/scripts/zones/Lower_Jeuno/npcs/Treasure_Coffer.lua
@@ -1871,7 +1871,16 @@ local argumentKeyItems =
     },
 }
 
+local isTenshodoCoffer = function(player)
+    local distance = player:checkDistance(41.169, 3.899, -51.005)
+    return distance < 6.0
+end
+
 entity.onTrigger = function(player, npc)
+    if not isTenshodoCoffer(player) then
+        return
+    end
+
     -- determines if server/player is eligible to receive a nexus cape
     local eligibleNexusCape = xi.settings.main.ENABLE_ACP == 1 and
                             xi.settings.main.ENABLE_AMK == 1 and


Xaver-DaRed:
I think a better solution would be to just rename the npc entries in lower jeuno to `Treasure_Coffer_Tenshodo` and `Treasure_Coffer_Gobbie` and then just rename the `Treasure_Coffer.lua` script in Lower Jeuno folder to `Treasure_Coffer_Tenshodo.lua`
imagen

claywar:
Agreed with this, and didn't realize you already had a PR up for it.  Separating the logic into two separate NPC scripts would be a better solution (and something I still want to do in other places as well for things like Phomiuna ladders, signposts)

cocosolos:
Are the gobbie coffers even supposed to be targetable?

Xaver-DaRed:
No they arent supposed to.
I took a cap of the NPC.
The gobbie coffer sql entry would have to be changed from
`INSERT INTO `npc_list` VALUES (17780999,'Treasure_Coffer','Treasure Coffer',85,13.699,0.000,9.310,1,50,50,0,4,36,0,3,0x0000C20300000000000000000000000000000000,0,'ABYSSEA',0);
`
to
`INSERT INTO `npc_list` VALUES (17780999,'Treasure_Coffer_Gobbie','Treasure Coffer',85,13.700,0.000,9.310,1,50,50,0,4,36,0,2051,0x0000C20300000000000000000000000000000000,0,'ABYSSEA',0);
`
So, in short:
- Change the pos, couse its minimally wrong
- Change the entity flags, which among other things, control if it can be targeted
- Change the npc name anyway out of principle. Makes it clearer what npc we are talking about

PR #6900, opened 2025-01-30
[lua] remove duplicate code for Eldieme Necropolis candles

Grahf0085:
**_I affirm:_**

Removes duplicate code for lighting the candles in Eldieme Necropolis.  Code is rafactored into one function that lives in globals.lua and called  by each candle NPC. The candles are sued to spawn seven different NMs. If this PR goes through there's a couple issues with the lamps I'd like to address next. Such as no fire animation when the lamp is lit, when checking a lit lamp it tells you it's unlit, etc

Trade a flintstone to the lamps. Should behave the same as it always has. Will tell you the lamp is lit and how many of the seven lamps are left to lite. After lighting all seven a NM will spawn at each lamp.

### DIFF
--- a/scripts/zones/The_Eldieme_Necropolis/npcs/_5fu.lua
+++ b/scripts/zones/The_Eldieme_Necropolis/npcs/_5fu.lua
@@ -1,38 +1,19 @@
 -----------------------------------
 -- Area: The Eldieme Necropolis
---  NPC: Tallow Candle
+-- NPC: Tallow Candle
 -- !pos 15.19 -18.30 339.80
 -----------------------------------
-local ID = zones[xi.zone.THE_ELDIEME_NECROPOLIS]
-local func = require('scripts/zones/The_Eldieme_Necropolis/globals')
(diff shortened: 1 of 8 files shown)


scripts/zones/The_Eldieme_Necropolis/npcs/_5fu.lua
```diff
@@ -3,36 +3,17 @@
 --  NPC: Tallow Candle
 -- !pos 15.19 -18.30 339.80
 -----------------------------------
-local ID = zones[xi.zone.THE_ELDIEME_NECROPOLIS]
 local func = require('scripts/zones/The_Eldieme_Necropolis/globals')
```
zach2good: I know this was already here, but while you're in here doing stuff can you change `func` to something more descriptive? In the Kuftal version of this we use `local kuftalGlobal = require('scripts/zones/Kuftal_Tunnel/globals')`, etc.

Grahf0085: Of course! How do we feel about this?

zach2good (approved):

PR #6905, opened 2025-01-31
[Lua] Bodyguard Follow Behavior Fix

dallano:
**_I affirm:_**

- This PR adds a new method of following a target as a train of mobs. The added function goes through the list all possible followers, adjacent to a mob's ID, and has each mob follow the next mob that is 1 ID closer to the leader - giving the behavior of a column following order. Should a mob in the middle of the train die, or not be spawned, it will follow the next mob in order.
- This behavior is used in a plethora of content and not only the ROTZ overworld NMs that spawn with x2 bodyguards. This PR intends to build the framework for future needed content.
- Side note: Previous behavior of the ROTZ overworld mixin caused the mobs to path inside one another

Links to captures with proof of follow behavior in ROTZ overworld NMs
https://www.youtube.com/watch?v=spfDvNUIUWM
https://www.youtube.com/watch?v=CV7WNt6vaWU
https://www.youtube.com/watch?v=wDdv3Ditz1U

!zone {Eastern Altepa Desert}
!spawnmob 17244372
!exec player:setPos(GetMobByID(17244372):getPos())
Observe behavior
Kill mob between NM and final mob and view adjustment behavior

### DIFF
--- a/scripts/globals/follow.lua
+++ b/scripts/globals/follow.lua
@@ -99,3 +99,34 @@ xi.follow.assignLeaderMod = function(mob, leaders, maxDistance)
         end
     end
 end
+
+-- This function causes the followers to organize in a column, and not
+-- stack upon one another. Must be called in leading mob's onMobRoam in
+-- order to maintain continuity of mob ordering in the line.
+--
+-- @param Original mob that leads the line of followers
+-- @param The amount of followers (assumed to be adjacent to origin mob's ID)
+-----------------------------------
+xi.follow.followLine = function(mob, numFollowers)
+    local mobID = mob:getID()
+
+    if
+        mob and
+        mob:isAlive()
+    then
+        for indexFollower = mobID + numFollowers, mobID + 1, -1 do
+            local follower = GetMobByID(indexFollower)
+
+            if follower and follower:isAlive() then
+                for indexLeader = indexFollower - 1, mobID, -1 do
+                    local newLeader = GetMobByID(indexLeader)
+
+                    if newLeader and newLeader:isAlive() then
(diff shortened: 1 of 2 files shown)


scripts/globals/follow.lua
```diff
@@ -99,3 +99,34 @@ xi.follow.assignLeaderMod = function(mob, leaders, maxDistance)
...
+-----------------------------------
+xi.follow.followLine = function(mob, numFollowers)
+    local mobID = mob:getID()
+
+    if
+        mob and
+        mob:isAlive()
+    then
+        for indexFollower = mobID + numFollowers, mobID + 1, -1 do
+            local follower = GetMobByID(indexFollower)
```
Frankie-hz, line 118: You are using mobID + 1 but does EVERY bodyguard follow this same behavior of follower = master + 1?

If this is used outside of just this instance does all mobs that follow their master follow this rule?

dallano: In all of my experience with following mobs, yes.
However, you're probably right in making it start based on a specific index rather than mobID + 1. That would allow for much more flexibility in strange edge cases

jmcmorris:
If I recall correctly this logic is attempted with Mother Globe and may be worth checking out. https://github.com/LandSandBoat/server/blob/base/scripts/zones/The_Shrine_of_RuAvitau/mobs/Mother_Globe.lua

scripts/globals/follow.lua
dallano: Upon review there are no non-custom mobs that would use a column follow behavior that aren't adjacent in mob ID.

PR #6919, opened 2025-02-01
Core: If inserted zone entity is a char check for moghouse status

zach2good:
**_I affirm:_**

Bug repro:

- Both be in MH in the same area
- One of you !goto "me"
- Other person will see the first person flash in for 1 400ms tick, and then disappear

Visible before, not visible after fix.

Also checked you can still see other players outside of MH.


zach2good:
Tested and ready to go

Xaver-DaRed (approved):

PR #6928, opened 2025-02-02
[core] Don't emit unnecessary party packets

sruon:
**_I affirm:_**

While investigating the party update packets storm (see #6914), I noticed several flaws in the logic:
- The ZoneServer forces everyone to update their party if they were within visible range of the now defunct Trust, even if they were not part of that Trust party. Possibly emitting hundreds of needless updates.
- charutils::ReloadParty loops over each party, and defers to CParty::ReloadParty which itself also loops over each party in the alliance, if the party is part of one.

Both combined, along with the issue outlined in #6914 caused an exponential amount of packets to be emitted if you were in an alliance and trusts had been spawned in your vicinity and not cleared from the ZoneServer.

This PR:
- Removes the request to ReloadParty from the ZoneServer clean up phase. As far as I can tell, it's not needed and the party correctly refreshes.
- Removes the loop in charutils::ReloadParty and simply defer to CParty::ReloadParty

This should theoretically help with the amount of packets sent in crowded zones.

I'll be testing some more shortly.

- Create a party with 1 char
- Create a party with another char
- One of the two summon and dismiss a Trust. Only the player that dismissed the trust should be receiving party packets (0xc8, 0xdd)

- Create an alliance
- Zone one of the character
- Verify everyone in each party is receiving party updates

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -6350,17 +6350,7 @@ namespace charutils
             }
 
             // once parties and alliances have been reassembled, reload the party/parties
-            if (PChar->PParty->m_PAlliance)
-            {
-                for (auto* party : PChar->PParty->m_PAlliance->partyList)
-                {
-                    party->ReloadParty();
-                }
-            }
-            else
-            {
-                PChar->PParty->ReloadParty();
-            }
+            PChar->PParty->ReloadParty();
         }
         else
         {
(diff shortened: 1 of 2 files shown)


WinterSolstice8 (approved):
LGTM as it's currently written

sruon:
Tested it now, and confirmed that RemoveTrusts is already requesting a party update so the call in ZoneServer was unnecessary.

zach2good (approved):

Issue #6897, opened 2025-01-30
Items traded in quests remain on hand.


eyes-and-brain:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

Items traded in quests remain in your possession
Correctly, the item should disappear after the trade.

Name of quest found
The Trader in the Forest
The Rescue

Flam9:
Looks like this can now be closed

PR #6931, opened 2025-02-02
[quest, lua] fix quest The Trader In The Forest to remove batagreens when traded

Grahf0085:
**_I affirm:_**

Fixes a part of #6897. When you trade batagreens to the NPC Abeaule they stay in your inventory. 

Complete the quest The Trade In The Forest


claywar (dismissed):
This needs to be in the 525 event finish since an item reward is given, and should only be confirmed if the complete is successful.

Xaver-DaRed (approved):

zach2good (approved):

PR #6932, opened 2025-02-03
[Quest][lua][sql] The Die is Cast

KnowOne134:
**_I affirm:_**

adds quest The Die is Cast
adds Random Ring

https://ffxiclopedia.fandom.com/wiki/The_Die_Is_Cast
 https://youtu.be/b7b5ABOi1n4

follow wiki steps
-- Ratihb: !pos 75.225 -6.000 -137.203 50
-- Ekhu Pesshyadha: !pos -13.043 0.999 103.423 50
-- Jijiroon: !pos 15.913 0.000 -32.676 53
-- qm9: !pos 311.088 -3.674 170.124 54


scripts/zones/Arrapago_Reef/mobs/Bukki.lua
```diff
@@ -2,11 +2,38 @@
...
+    mob:setMobMod(xi.mobMod.IDLE_DESPAWN, 300)
+    mob:setMobMod(xi.mobMod.GIL_MAX, -1)
+    mob:setLocalVar('tp_spam', math.random(5, 15))
+    mob:addListener('COMBAT_TICK', 'BUKKI_TICK', function(mob)
+        if mob:getHPP() <= mob:getLocalVar('tp_spam') and mob:getLocalVar('tp_spam') > 0 then
+            mob:setLocalVar('tp_spam', 0)
+            mob:setMagicCastingEnabled(false)
+            mob:setMobSkillAttack(788)
+        end
+    end)
```
Xaver-DaRed, line 37: To fix the CI:
```lua
mob:addListener('COMBAT_TICK', 'BUKKI_TICK', function(mobArg)
    if mobArg:getHPP() <= mobArg:getLocalVar('tp_spam') and mobArg:getLocalVar('tp_spam') > 0 then
        mobArg:setLocalVar('tp_spam', 0)
        mobArg:setMagicCastingEnabled(false)
        mobArg:setMobSkillAttack(788)
    end
end)
```

Xaver-DaRed: Athough, you could always just use the regular entity function for combat tick, i would think
```lua
    -- Logic
end
```

KnowOne134: ah thanks, couldnt figure out what it wanted, we dont have to use args

scripts/quests/ahtUrhgan/The_Die_is_Cast.lua
```diff
@@ -0,0 +1,159 @@
...
+                onTrigger = function(player, npc)
+                    if npcUtil.popFromQM(player, npc, ID.mob.BUKKI, { hide = 0 }) then
+                        return quest:messageSpecial(ID.text.SPINE_CHILL)
+                    end
+                end,
+            },
+
+            ['Bukki'] =
+            {
+                onMobDeath = function(mob, player, isKiller, noKiller)
```
Xaver-DaRed: `onMobDeath = function(mob, player, optParams)`

We now use optParams.isKiller and opt.Params.noKiller. So those 2 are part of the optParams tsable

claywar (approved):

PR #6933, opened 2025-02-03
[SQL] Audit mob droplist - 151 to 180

LacunaXI:
**_I affirm:_**

Continues #6756 

Fixing and validating all of the salvage drops is a project itself and outside of the scope of this audit. Salvage foes are being left with a TODO on them. The chariot boss at the end of the run is an exception as it was easy to verify the retail accuracy of these four.


Xaver-DaRed (approved):

KnowOne134:
im in the middle of salvage porting to LSB which is done on eden server, so I appreciate you skipping those tables

LacunaXI:
That is excellent, thank you very much.

PR #6938, opened 2025-02-04
[SQL] Audit mob droplist - 181 to 210

LacunaXI:
**_I affirm:_**

Continues the audit.


Xaver-DaRed (approved):

zach2good:
Doing good work, thanks!

LacunaXI:
> Doing good work, thanks!

Thank you and thank you for the quick merges. We are now 6% through all these PRs!

WinterSolstice8:
we're ok with doing more than 30 droplists at a time, especially if it's similar droplists or something. I know its a lot of PRs to get everything in here and I personally would like to help you be done with it sooner so that you dont spend a ton of time simply making PRs of already-done work

LacunaXI:
I chose 30 arbitrarily as it can be a lot to review at once. What would you think then? 50? More?

I still re-review close to every entry before I submit them anyway. Especially since there were over 1,500 TODOs along the way. So it takes a lot of time to submit despite being already-done from 1-3370.

WinterSolstice8:
I specifically asked about 50 earlier and that will be fine, but was noted that arbitrary numbers may not matter if groupings come up for a lot more

LacunaXI:
I am happy to do 50. Thank you.

The current PR #6952 is an example of groupings reducing the amount to review as more was removed than added. In that case 50 is still fine since is just less to review.

PR #6939, opened 2025-02-04
[quest, lua] fix quest The Rescue to remove Quadav Charm item when traded

Grahf0085:
**_I affirm:_**

Fixes the last issue with #6897

With these changes when I do the quest "The Rescue" the Quadav Charm item is removed from my inventory when I trade it

!gotoid17793026
Speak to Thunder Hawk to get quest
!additem 495 to get Quadav Charm
!gotoid 17379626 to get close to destination
Go to door at I8.
Trade Quadav Charm to door.
It should be removed from your inventory.

### DIFF
--- a/scripts/quests/otherAreas/The_Rescue.lua
+++ b/scripts/quests/otherAreas/The_Rescue.lua
@@ -93,6 +93,7 @@ quest.sections =
             onEventFinish =
             {
                 [1000] = function(player, csid, option, npc)
+                    player:confirmTrade()
                     npcUtil.giveKeyItem(player, xi.ki.TRADERS_SACK)
                 end,
             },


scripts/quests/otherAreas/The_Rescue.lua
```diff
@@ -93,7 +93,9 @@ quest.sections =
             onEventFinish =
             {
                 [1000] = function(player, csid, option, npc)
-                    npcUtil.giveKeyItem(player, xi.ki.TRADERS_SACK)
+                    if npcUtil.giveKeyItem(player, xi.ki.TRADERS_SACK) then
```
claywar: giveKeyItem always returns true, so I would suggest just confirm then give (that order, since the scream test will work if it fails).  No need for a condition here

Grahf0085: Not sure what scream test is but changes were made

Xaver-DaRed (approved):

PR #3978, opened 2023-06-08
[core] Refactor trigger areas and validate volumes on startup

zach2good:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

Will this make me finally refactor trigger areas to be typesafe? Maybe. Who knows.

PR #6940, opened 2025-02-04
Support overlapping TriggerAreas

### LINKED ISSUE #6894: 🐛 Overlapping TriggerAreas defeat debounce mechanism
**_I affirm:_**

Server: Ubuntu 24.04
Client: Windows 10

## Branch affected by issue 

`base`
Running on 9b0a788804e19b4795aaa4944ff6e655ba087e2f, but have reproduced it as further back as e6e34bbb37

## Steps to reproduce

Accept the quest "[Flyers For Regine](https://www.bg-wiki.com/ffxi/Flyers_for_Regine)" in Port Sandoria
Exit to Northern Sandoria
Approach Villion (V) from below (F3, up on a balcony)
Enter his TriggerArea
Move slightly, notice it only writes 1 log line and does not loop
Slowly move towards Guilberdrier (G) (F3, on the lower level)
Trigger Gs TriggerArea, while still being within Vs
Log is immediately spammed with Gs phrase once, then Vs trigger phrase on loop

The spam stops once you are out of range

## Expected behavior

You can be in multiple trigger areas at the same time with the debounce mechanism (not firing if you're already within the Trigger Area) working correctly for each.

## Notes

The design of this seems to be shared across Zone and Zone Instance fwiw. I notice that some of [this code ](https://github.com/LandSandBoat/server/commit/aa94b032e75f6f54382a9101096170d1fe7f111d)changed very recently, could be related, or could be a red herring given it was a large refactor and a lot of things got touched.

What looks like is happening is that when the [trigger area check](https://github.com/LandSandBoat/server/blob/base/src/map/zone.cpp#L1228) fires, and you are both:
- In an existing Trigger Area from the prior call
- Have just entered a new Trigger Area in this call

The second TriggerArea in the list (not necessarily the 2nd one you entered, afaik) is [stored on PChar](https://github.com/LandSandBoat/server/blob/base/src/map/zone.cpp#L1260) each loop while you're in both, resulting in the first one always firing the action for OnTriggerAreaEnter, because it will always [fail this check](https://github.com/LandSandBoat/server/blob/base/src/map/zone.cpp#L1245). Since m_InsideTriggerAreaID is only ever set outside the loop, it only ever stores the last one it encounters, meaning earlier triggerAreas in the list will always fire.

The [logs](https://gist.github.com/StabbyCutyou/bd0d8facbe532642060a2f3951957f3c) from the server reveal that the char suffers a massive packet backlog as well. Those should show a clean zone into Northern Sandy, approaching V, seeing the trigger, my putting something to into say to demarcate, then entering Gs trigger area and getting the flood.

The logic inside of that function is not what I would have expected, as I would have assumed you would want to record the zone as you enter it, and not as you finish the loop checking for all entries, among other things. That said, if the idea is that you can only be in one zone at a time, any choice is fine so long as it is consistent.

If this is the intended behavior, and it's just this quest that needs it's own internal debounce, that could be a quick solve that feels similar to the [Castle Z](https://github.com/dallano/server/blob/82650d4c5c366f113abce1831041ecac45e87628/scripts/zones/Castle_Zvahl_Keep/Zone.lua#L75) teleport fix that went in recently. But I wanted to raise it as a possible bug before trying to put a fix in for it, as it does seem like the [code that uses m_InsideTriggerAreaID](https://github.com/LandSandBoat/server/blob/9b0a788804e19b4795aaa4944ff6e655ba087e2f/src/map/lua/lua_baseentity.cpp#L2807) is designed to expect a single area only.

It could also be that what is actually needed are events for both entering an area, as well as existing inside of one on a tick, since both have valid applications.

StabbyCutyou:
**_I affirm:_**

Closes #6894

The current logic for handling being inside of a TriggerArea presumes you'll only ever be inside of a single one at a time, which causes issues for quests like Flyers for Regine. In that quest, multiple pairs of NPCs are close enough to have their trigger areas invoked at once, which causes the issue described in that bug (the last area in the list that you've entered will be recorded, but all prior trigger areas will fire on repeat until you leave them, since their debounce fails).

This change pivots m_InsideTriggerAreaID to be a bitmask instead of a single value, which allows us to shift 1 left by the numbers of bits represented by the value of the ID to act as unique bitmask values.

This also adds methods for interacting with m_InsideTriggerAreaID so that all the logic for manipulating areas is kept in one place, and I removed a lot of extraneous logic from zone/zone instance and pushed tracking the triggered area into the luatils method as that seemed to do all the real heavy lifting, and now when the method exits all the logic to invoke the trigger area (including recording it on the player) is in one place as well.

- Accept the quest "[Flyers For Regine](https://www.bg-wiki.com/ffxi/Flyers_for_Regine)" in Port Sandoria
- Exit to Northern Sandoria
- Approach Villion (V) from below (F3, up on a balcony)
- Enter his TriggerArea
- Move slightly, notice it only writes 1 log line for V, and does not loop
- Slowly move towards Guilberdrier (G) (F3, on the lower level)
- Trigger Gs TriggerArea, while still being within Vs
- Notice it only writes 1 log line for G and does not repeat Vs

Leaving and reentering the trigger areas cause them to fire again as normal.

I've also tested the Bastok Bridge area, as that one got a special call out in some older comments, and it also worked as expected.

## Notes

I expect I need to provide better evidence of it working than "trust me", so please let me know what would be sufficient. I have some videos of the quest and the bastok airship working, but just quick ones on my phone to show my partner. I can provide better proof however you'd like.

### DIFF
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -3098,6 +3098,26 @@ bool CCharEntity::OnAttackError(CAttackState& state)
     return false;
 }
 
+bool CCharEntity::isInTriggerArea(uint32 triggerAreaID)
+{
+    return m_InsideTriggerAreaID & (1 << triggerAreaID);
+}
+
+void CCharEntity::onTriggerAreaEnter(uint32 triggerAreaID)
+{
+    m_InsideTriggerAreaID |= (1 << triggerAreaID);
+}
+
+void CCharEntity::onTriggerAreaExit(uint32 triggerAreaID)
+{
+    m_InsideTriggerAreaID &= ~(1 << triggerAreaID);
+}
+
+void CCharEntity::clearTriggerAreas()
+{
+    m_InsideTriggerAreaID = 0;
+}
+
 bool CCharEntity::isInEvent()
 {
     return currentEvent->eventId != -1;
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -3100,22 +3100,25 @@ bool CCharEntity::OnAttackError(CAttackState& state)
 
 bool CCharEntity::isInTriggerArea(uint32 triggerAreaID)
 {
-    return m_InsideTriggerAreaID & (1 << triggerAreaID);
+    if (auto found = charTriggerAreaIDs.find(triggerAreaID); found != charTriggerAreaIDs.end()) {
+        return true
+    }
+    return false;
 }
 
 void CCharEntity::onTriggerAreaEnter(uint32 triggerAreaID)
 {
-    m_InsideTriggerAreaID |= (1 << triggerAreaID);
+    charTriggerAreaIDs.insert(triggerAreaID)
 }
 
 void CCharEntity::onTriggerAreaExit(uint32 triggerAreaID)
 {
-    m_InsideTriggerAreaID &= ~(1 << triggerAreaID);
+    charTriggerAreaIDs.erase(triggerAreaID)
 }
 
 void CCharEntity::clearTriggerAreas()
 {
-    m_InsideTriggerAreaID = 0;
+    charTriggerAreaIDs.clear()
 }
 
 bool CCharEntity::isInEvent()
--- a/src/map/zone_instance.cpp
+++ b/src/map/zone_instance.cpp
@@ -443,29 +443,21 @@ void CZoneInstance::CheckTriggerAreas()
             // TODO: When we start to use octrees or spatial hashing to split up zones,
             //     : use them here to make the search domain smaller.
 
-            uint32 triggerAreaID = 0;
             for (const auto& triggerArea : m_triggerAreaList)
             {
+                auto triggerAreaID = triggerArea->GetTriggerAreaID();
                 if (triggerArea->isPointInside(PChar->loc.p))
                 {
-                    triggerAreaID = triggerArea->GetTriggerAreaID();
-
-                    if (triggerArea->GetTriggerAreaID() != PChar->m_InsideTriggerAreaID)
+                    if (!PChar->isInTriggerArea(triggerAreaID))
                     {
                         luautils::OnTriggerAreaEnter(PChar, triggerArea);
                     }
-
-                    if (PChar->m_InsideTriggerAreaID == 0)
-                    {
-                        break;
-                    }
                 }
-                else if (triggerArea->GetTriggerAreaID() == PChar->m_InsideTriggerAreaID)
+                else if (PChar->isInTriggerArea(triggerAreaID))
                 {
                     luautils::OnTriggerAreaLeave(PChar, triggerArea);
                 }
             }
-            PChar->m_InsideTriggerAreaID = triggerAreaID;
         });
         // clang-format on
     }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -464,6 +464,15 @@
         "__split_buffer": "cpp",
         "__threading_support": "cpp",
         "__tree": "cpp",
-        "print": "cpp"
+        "print": "cpp",
+        "__bits": "cpp",
+        "__debug": "cpp",
+        "__errc": "cpp",
+        "__functional_base": "cpp",
+        "__mutex_base": "cpp",
+        "__nullptr": "cpp",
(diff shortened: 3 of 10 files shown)


src/map/entities/charentity.cpp
```diff
@@ -3098,6 +3098,26 @@ bool CCharEntity::OnAttackError(CAttackState& state)
     return false;
 }
 
+bool CCharEntity::isInTriggerArea(uint32 triggerAreaID)
+{
+    return m_InsideTriggerAreaID & (1 << triggerAreaID);
```
StabbyCutyou: I just realized that it might make sense to change the name of this, since it's no longer a single ID.

claywar:
Before reviewing as well, can you check the scenario where you zone in on top of a triggerArea? Might be an easy win

StabbyCutyou:
Yup let me give that a shot now

StabbyCutyou:
Confirmed, logging in and both Villion and Guilberdrier immediately give me the look.

src/map/entities/charentity.cpp
zach2good: From a quick glance, this isn't safe: 

```lua
    -- Yellow portals (From Sky Gods)
    zone:registerTriggerArea(33,  543, -73,  -19,  547, -69,  -15) -- Seiryu Island to SE Island
    zone:registerTriggerArea(34,  182, -73,  511,  186, -69,  515) -- Genbu Island to NE Island
    zone:registerTriggerArea(35, -432, -73,  332, -428, -69,  336) -- Byakko Island to NW Island
    zone:registerTriggerArea(36, -453, -73, -308, -449, -69, -304) -- Suzaku Island to SW Island

    -- Green portals
    zone:registerTriggerArea(37,  142, -41, -156,  145, -39, -153) -- Main Island (NE)
    zone:registerTriggerArea(38, -145, -41, -156, -142, -39, -153) -- Main Island (NW)
```

We have multiple zones with more than 32 registered trigger areas.

src/map/zone_instance.cpp
```diff
@@ -443,29 +443,21 @@ void CZoneInstance::CheckTriggerAreas()
             // TODO: When we start to use octrees or spatial hashing to split up zones,
             //     : use them here to make the search domain smaller.
 
-            uint32 triggerAreaID = 0;
             for (const auto& triggerArea : m_triggerAreaList)
             {
+                auto triggerAreaID = triggerArea->GetTriggerAreaID();
```
zach2good, line 448: Rather than using the mask I saw before, I don't think there's any harm (as long as it's properly maintained) for CCharEntity to have a `std::unordered_set` of `triggerAreaIDs` that they're currently inside - which we insert to when they enter, and remove from when they leave, etc. Since it's `unordered_set` insertion and removal are constant time (in the most common case).

That allows us to have an arbitrary number of trigger areas in a zone, with the possibility of players being in any number of them - and tracking entry/exit from them.

zach2good (commented):
In favour of this PR, it would make it possible for me to kill my old PR here: https://github.com/LandSandBoat/server/pull/3978

(or scale it back to a simple refactor without the validation)

src/map/zone_instance.cpp
StabbyCutyou: Originally my idea was to use a list, but I figured this would be a simpler refactor.

That said, I completely missed that there were cases of >32 triggers, and so my approach here ultimately wouldn't work.

I can start the changes to move to an unordered list tonight after I'm home from work.

StabbyCutyou:
Converted to draft while I'm thrashing on getting c++ syntax right.

StabbyCutyou:
Ah it still kicks off checks while in draft - I can close this and re-open once I have it working again, I don't wanna distract while i'm pushing builds to my server and testing.

zach2good:
Don't worry about the CI checks running, we bounce things of CI until they work _constantly_, just squash your commits when you're done and comment when you think you're ready for re-review (no need to tag anyone ,we see _everything)

zach2good:
Also, MSFT is paying for all the CI, use at will 💸

StabbyCutyou:
Hah, good to know. Sweet deal.

I didn't see this until I had already basically gotten the second pr ready (sick partner, WFH, time to code for fun in the morning) otherwise I would have respected your request here.

Thanks again!

PR #6942, opened 2025-02-04
[build] Load nested module CMakeLists.txt if present

sruon:
Allow modules to manipulate the build process for xi_map

**_I affirm:_**

Building a module that needs to link libraries and realized there was no easy way to change the build process without dirtying the main CMakeLists.

This makes the map process CMakeLists check if the module has a CMakeList.txt at the root and include it if so.

Not a CMake expert, open to other ideas.

Add a CMakeList.txt in a module, reconfigure cmake, see it get included.

### DIFF
--- a/src/map/CMakeLists.txt
+++ b/src/map/CMakeLists.txt
@@ -153,12 +153,18 @@ set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${CMAKE_SOURCE_D
 
 # Search for module directories
 set(module_include_dirs "")
+set(cmakelist_include_paths "")
 file(STRINGS ${CMAKE_SOURCE_DIR}/modules/init.txt INIT_FILE_ENTRIES REGEX "^[^#\n].*")
 foreach(entry ${INIT_FILE_ENTRIES})
     if (${entry} STREQUAL "")
         continue()
     endif()
     if (IS_DIRECTORY "${CMAKE_SOURCE_DIR}/modules/${entry}")
+        if (EXISTS "${CMAKE_SOURCE_DIR}/modules/${entry}/CMakeLists.txt")
+            list(APPEND cmakelist_include_paths "${CMAKE_SOURCE_DIR}/modules/${entry}/CMakeLists.txt")
+            message(STATUS "Adding module CMakeLists.txt to build: ${CMAKE_SOURCE_DIR}/modules/${entry}/CMakeLists.txt")
+            continue()
+        endif()
         file(GLOB_RECURSE module_files
             "${CMAKE_SOURCE_DIR}/modules/${entry}/*.cpp"
             "${CMAKE_SOURCE_DIR}/modules/${entry}/*.h")
@@ -174,6 +180,13 @@ endforeach()
 add_executable(xi_map
     ${SOURCES}
     ${resource})
+
+foreach(entry ${cmakelist_include_paths})
+    if (${entry} STREQUAL "")
+        continue()
+    endif()
+    include("${entry}")
+endforeach()
 
 if(WIN32)
     set_target_properties(xi_map PROPERTIES OUTPUT_NAME xi_map)


WinterSolstice8 (approved):
seems reasonable, but deferring to zach

zach2good:
Reply pending, please hold

zach2good:
**TLDR;**

This PR aside: what library are you trying to link and what are you trying to achieve? This opens up all sorts of nasty doors that I think fall outside of the scope of what our module system is trying to achieve:

https://github.com/LandSandBoat/server/wiki/Module-Guide

> Modules are designed first and foremost to ensure that server operators can safely keep their servers updated and that bug reports sent back upstream are valid and not the result of deep customisation.

^ Kind of implied by this: If you have used the base software with some reasonable modules and haven't done anything _too_ wacky, it's reasonable for us to try and support you - since it'll benefit the software as a whole.

**The Rest**

These changes would further reduce the already small pool of valid bug reports being sent to us. It also invalidates our "fact-finding" missions into server discords to find bug reports that users are unwilling to send in to us for whatever reasons. We've already seen out in the wild that operators are using modules to implement valid (often simple) bugfixes and missing retail content, with no view to ever send it back upstream for anyone else to benefit from. With these things in mind you can see why making C++ modules all-powerful wouldn't be in our best interests.

If operators/users want to achieve a level of customisation that requires them to modify the build process rather than just appending to it; linking new libraries, changing flags, swapping translation units, and whatnot - they've outgrown modules and are sufficiently advanced in need and proficiency that it's unreasonable for us to offer up any support.

I'm not trying to be a killjoy, we want people to use and enjoy the software we pour so many hours into upgrading and maintaining, but the mission statement is: `a server emulator for Final Fantasy XI`, and not `a customisation engine that just so happens to interface with the Final Fantasy XI client`.

zach2good:
We talked about this in DMs and I think I've come up with a design I'm happy with (I can implement it, or you can, whichever you're more comfortable with):

```
Lua modules are scraped by looking at init.txt at startup, and then scraping _everything_ per listing. (existing)
SQL modules are scraped by looking at init.txt with dbtool.py, and then scraping _everything_ per listing. (existing)

C++ modules will be scraped by looking at init.txt during CMake configuration time. If it reaches a directory with a `CMakeLists.txt` or `*.cmake` file in it, it will load that and stop walking that directory. Otherwise it will continue scraping for `*.cpp` files.
```

I think this will allow the "dream structure" of a complete module to look like:

```
<repo>
    modules/
        example_module/
            lua/
                <lua files loaded by init.txt & xi_map>
            src/
                <src files loaded by CMakeLists.txt>
            sql/
                <SQL files loaded by dbtool.py>
            CMakeLists.txt (listing src/, other libraries, etc.)
            README.md
        init.txt (listing "example_module/")
```

sruon:
> We talked about this in DMs and I think I've come up with a design I'm happy with (I can implement it, or you can, whichever you're more comfortable with):

I'll make the changes and submit another commit for review.

sruon:
Made the changes and updated the custom module to reflect usage.

> If it reaches a directory with a `CMakeLists.txt` or `*.cmake` file in it

I wasn't sure if you really meant *.cmake since you used CMakeLists.txt in the next example. 

I can add a glob for it but then I'm not sure this is not eventually going to cause issues with regards to ordering.

zach2good:
Lets just got with `CMakeLists.txt` for now, that's more standard. For the modules you've converted - I don't know if they should be treated as a single module. People treat them as one, but I wrote them to be optionally used separately, so maybe just leave them as they are and I'll make up an empty example module or something

zach2good:
Also, I think `GLOB` is never preferable, as it doesn't track changes and won't trigger a re-configure if the underlying files are added or removed unless you manually remove your cache, etc. (I think)

sruon:
Reverted changes to the custom module.

zach2good:
Looking good, hope this handles your needs 👍 

For passers-by to compare to my previous big statement: Turds are gonna turd. If people want to be bad actors, there's no way we can force them to not. It's also unreasonable for me to think modifying the build system will lead to _complex_ abuse like swapping translation units. Also, when people report bugs to us, one of our normal responses is: "have you been able to reproduce this on base LSB with no modules?" - which won't change with this PR, etc.

PR #6944, opened 2025-02-04
[lua, cpp] fixes mobskills base damage and ftp naming

Frankie-hz:
**_I affirm:_**

- Fixes the base damage used for mobs. For whatever reason mobs based damage was using `weaponDamage + fstr + dLvl? * ftp`
https://w.atwiki.jp/studiogobli/pages/13.html
https://www.bg-wiki.com/ffxi/Base_Damage
https://www.bg-wiki.com/ffxi/Physical_Damage

- Renames dmgmod to ftp. Based on retail research mob damage is extremely consistent and has been since the release of the game. Any mobskill thats damage cannot vary has the same ftp values at 3,000 TP vs 1,000 TP. After testing 10 different mobskills that have static damages with no damage varies - the damage is always the same from the TP values. This lines up with what you see in game all over the place. Here is just one example.

image

- Gets rid of the duplicate pdif calcs inside the mobskills.lua and uses the global function. There is an edge case that cannonball uses the DEF stat and not he ATT stat so I added a param to the `xi.combat.physical.calculateMeleePDIF`. It looks messy but I didn't know another way to do this. We will need this eventually for bluemage.lua when we convert that over to use the new global.

- Fixes an issue where Seismostomp was throwing errors https://github.com/LandSandBoat/server/issues/6953#issuecomment-2635554917

Go to any mob and !tp 3000 on it for it to use any physical mobskill. Since the base damage is now slightly lower you should see slightly lower damage than before based on if the mob is higher level than you or not.

### DIFF
--- a/scripts/actions/mobskills/abyssal_strike.lua
+++ b/scripts/actions/mobskills/abyssal_strike.lua
@@ -15,8 +15,8 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
     local accmod  = 2
-    local dmgmod  = 4.7
-    local info    = xi.mobskills.mobPhysicalMove(mob, target, skill, numhits, accmod, dmgmod, xi.mobskills.physicalTpBonus.DMG_VARIES, 1, 2, 3)
+    local ftp     = 4.7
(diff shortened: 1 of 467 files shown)


Xaver-DaRed (approved):
Everything looks good. Except for the cannpnball thing.

A work for another time, however, instead of feeding "is this one specific action" we could have fed "what do we use as attack" and improve both the functionality and flexibility.

But, thats a job for future-someone-else-lets-hope-not-me

WinterSolstice8 (approved):

PR #6946, opened 2025-02-04
Support overlapping TriggerAreas

### LINKED ISSUE #6894: 🐛 Overlapping TriggerAreas defeat debounce mechanism

StabbyCutyou:
**_I affirm:_**

Closes https://github.com/LandSandBoat/server/issues/6894
Replaces https://github.com/LandSandBoat/server/pull/6940
Potentially obviates some of the work in https://github.com/LandSandBoat/server/pull/3978

The current logic for handling being inside of a TriggerArea presumes you'll only ever be inside of a single one at a time, which causes issues for quests like Flyers for Regine. In that quest, multiple pairs of NPCs are close enough to have their trigger areas invoked at once, which causes the issue described in that bug (the last area in the list that you've entered will be recorded, but all prior trigger areas will fire on repeat until you leave them, since their debounce fails).

This change pivots m_InsideTriggerAreaID to be an unordered set instead of a single value, which allows us to store and arbitrary number of TriggerAreaIDs on a character. The datatype was recommended by one of the core maintainers here as having largely good performance characteristics.

I've also moved to a private section alongside some other unordered_set/maps, as the logic to handle the data should always be going through the helper methods and callers should not interact with the list itself. Additionally, i changed the name to `charTriggerAreaIDs` to be more in-keeping with its new neighbors.

This also adds methods for interacting with charTriggerAreaIDs so that all the logic for manipulating areas is kept in one place, and I removed a lot of extraneous logic from zone/zone instance and pushed tracking the triggered area into the luatils method as that seemed to do all the real heavy lifting, and now when the method exits all the logic to invoke the trigger area (including recording it on the player) is in one place as well.

- Accept the quest "[Flyers For Regine](https://www.bg-wiki.com/ffxi/Flyers_for_Regine)" in Port Sandoria
- Exit to Northern Sandoria
- Approach Villion (V) from below (F3, up on a balcony)
- Enter his TriggerArea
- Move slightly, notice it only writes 1 log line for V, and does not loop
- Slowly move towards Guilberdrier (G) (F3, on the lower level)
- Trigger Gs TriggerArea, while still being within Vs
- Notice it only writes 1 log line for G and does not repeat Vs

Leaving and reentering the trigger areas cause them to fire again as normal.

I've also tested the Bastok Bridge area, as that one got a special call out in some older comments, and it also worked as expected.

## Notes

I've verified that the test case from my original PR still works, but have not checked every triggerarea.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -464,6 +464,15 @@
         "__split_buffer": "cpp",
         "__threading_support": "cpp",
         "__tree": "cpp",
-        "print": "cpp"
+        "print": "cpp",
+        "__bits": "cpp",
+        "__debug": "cpp",
+        "__errc": "cpp",
+        "__functional_base": "cpp",
+        "__mutex_base": "cpp",
+        "__nullptr": "cpp",
+        "__std_stream": "cpp",
+        "__string": "cpp",
+        "__tuple": "cpp"
     },
 }
--- a/src/map/entities/charentity.cpp
+++ b/src/map/entities/charentity.cpp
@@ -181,17 +181,16 @@ CCharEntity::CCharEntity()
 
     m_PMonstrosity = nullptr;
 
-    m_Costume             = 0;
-    m_Costume2            = 0;
-    m_hasTractor          = 0;
-    m_hasRaise            = 0;
-    m_weaknessLvl         = 0;
-    m_hasArise            = false;
-    m_InsideTriggerAreaID = 0;
-    m_LevelRestriction    = 0;
-    m_lastBcnmTimePrompt  = 0;
-    m_AHHistoryTimestamp  = 0;
-    m_DeathTimestamp      = 0;
+    m_Costume            = 0;
+    m_Costume2           = 0;
+    m_hasTractor         = 0;
+    m_hasRaise           = 0;
+    m_weaknessLvl        = 0;
+    m_hasArise           = false;
+    m_LevelRestriction   = 0;
+    m_lastBcnmTimePrompt = 0;
+    m_AHHistoryTimestamp = 0;
+    m_DeathTimestamp     = 0;
 
     m_EquipFlag         = 0;
     m_EquipBlock        = 0;
@@ -3096,6 +3095,26 @@ bool CCharEntity::OnAttackError(CAttackState& state)
         return true;
     }
     return false;
+}
+
+bool CCharEntity::isInTriggerArea(uint32 triggerAreaID)
+{
+    return charTriggerAreaIDs.find(triggerAreaID) != charTriggerAreaIDs.end();
+}
+
+void CCharEntity::onTriggerAreaEnter(uint32 triggerAreaID)
+{
+    charTriggerAreaIDs.insert(triggerAreaID);
+}
+
+void CCharEntity::onTriggerAreaLeave(uint32 triggerAreaID)
+{
+    charTriggerAreaIDs.erase(triggerAreaID);
+}
+
+void CCharEntity::clearTriggerAreas()
+{
+    charTriggerAreaIDs.clear();
 }
 
 bool CCharEntity::isInEvent()
--- a/scripts/specs/core/CBaseEntity.lua
+++ b/scripts/specs/core/CBaseEntity.lua
@@ -725,9 +725,27 @@ end
 function CBaseEntity:isInMogHouse()
 end
 
----@nodiscard
----@return integer
-function CBaseEntity:getPlayerTriggerAreaInZone()
+---@param triggerAreaId integer
+---@return boolean
+function CBaseEntity:isPlayerInTriggerArea(triggerAreaId)
+end
+
+---@return nil
+function CBaseEntity:isPlayerInAnyTriggerArea()
+end
+
+---@param triggerAreaId integer
+---@return nil
+function CBaseEntity:onPlayerTriggerAreaEnter(triggerAreaId)
+end
+
+---@param triggerAreaId integer
+---@return nil
+function CBaseEntity:onPlayerTriggerAreaLeave(triggerAreaId)
+end
+
+---@return nil
+function CBaseEntity:clearPlayerTriggerAreas()
 end
 
 ---@param statusID integer
--- a/scripts/zones/Port_Bastok/npcs/_6kt.lua
+++ b/scripts/zones/Port_Bastok/npcs/_6kt.lua
@@ -35,7 +35,7 @@ entity.onTimeTrigger = function(npc, triggerID)
 
         --If a player is on the bridge, kick them off
         for _, player in pairs(players) do
-            if player:getPlayerTriggerAreaInZone() == 2 then
+            if player:isPlayerInTriggerArea(2) then
                 player:startEvent(70)
             end
         end
--- a/src/map/entities/charentity.h
+++ b/src/map/entities/charentity.h
@@ -479,8 +479,7 @@ public:
 
     std::unique_ptr<monstrosity::MonstrosityData_t> m_PMonstrosity;
 
-    uint32     m_InsideTriggerAreaID; // The ID of the trigger area the character is inside
-    uint8      m_LevelRestriction;    // Character level limit
+    uint8      m_LevelRestriction; // Character level limit
     uint16     m_Costume;
     uint16     m_Costume2;
     uint32     m_AHHistoryTimestamp;
@@ -586,6 +585,11 @@ public:
     int32 GetTimeCreated();
     uint8 getHighestJobLevel();
 
+    bool isInTriggerArea(uint32 triggerAreaId);
+    void onTriggerAreaEnter(uint32 tiggerAreaId);
+    void onTriggerAreaLeave(uint32 triggerAreaId);
+    void clearTriggerAreas();
+
     bool isInEvent();
     bool isNpcLocked();
     void queueEvent(EventInfo* eventToQueue);
@@ -659,6 +663,7 @@ private:
 
     std::unordered_map<std::string, std::pair<int32, uint32>> charVarCache;
     std::unordered_set<std::string>                           charVarChanges;
+    std::unordered_set<uint32>                                charTriggerAreaIDs; // Holds any TriggerArea IDs that the player is currently within the bounds of
 
     uint8      dataToPersist = 0;
     time_point nextDataPersistTime;
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -2799,13 +2799,12 @@ bool CLuaBaseEntity::isInMogHouse()
 }
 
 /************************************************************************
-*  Function: getPlayerTriggerAreaInZone
-*  Purpose : Returns the player's current trigger area inside the zone
-*  Example : local triggerAreaID = player:getPlayerTriggerAreaInZone()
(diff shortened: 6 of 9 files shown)


.vscode/settings.json
```diff
@@ -464,6 +464,15 @@
...
+        "print": "cpp",
+        "__bits": "cpp",
+        "__debug": "cpp",
+        "__errc": "cpp",
+        "__functional_base": "cpp",
+        "__mutex_base": "cpp",
+        "__nullptr": "cpp",
+        "__std_stream": "cpp",
+        "__string": "cpp",
+        "__tuple": "cpp"
```
StabbyCutyou, line 476: Let me know if I should revert this, I did not knowingly make any changes here and am unaware of what the impact of these might be.

src/map/entities/charentity.cpp
```diff
@@ -3098,6 +3097,30 @@ bool CCharEntity::OnAttackError(CAttackState& state)
     return false;
 }
 
+bool CCharEntity::isInTriggerArea(uint32 triggerAreaID)
+{
+    if (auto found = charTriggerAreaIDs.find(triggerAreaID); found != charTriggerAreaIDs.end())
+    {
+        return true;
+    }
+    return false;
```
zach2good: `return charTriggerAreaIDs.find(triggerAreaID) != charTriggerAreaIDs.end();`

.vscode/settings.json
zach2good: Don't worry about these, I dunno what the mechanic is that adds stuff to this, but it happens from time to time 🤷

zach2good (approved):
Looking good, just one small change 👍

src/map/entities/charentity.cpp
StabbyCutyou: oh wow that is painfully obvious, I'll patch that up later tonight.

src/map/lua/luautils.cpp
```diff
@@ -2050,6 +2050,8 @@ namespace luautils
             ShowError("luautils::onTriggerAreaEnter: %s", err.what());
             ReportErrorToPlayer(PChar, err.what());
         }
+        // Add the TriggerArea to the players cache of current TriggerAreas
+        PChar->onTriggerAreaEnter(PTriggerArea->GetTriggerAreaID());
```
zach2good: Are these necessary in the binding? The state of "am I in the trigger area or not" should be ZoneIn, ZoneOut, and CheckTriggerAreas - this should just be calling into Lua and not changing the underlying set?

The other areas should perhaps look like:
```
...
// <-- code to add to trigger area list
luautils::OnTriggerAreaEnter(PChar, triggerArea);
...
```

and the same for exit.

Ultimately it doesn't matter because it's appending to a set, so it's de-duplicated, but it's a bit out of place

StabbyCutyou: So, here was my thought process and you can feel to tell me that I'm wrong or you'd prefer see it differently (or that I've just misunderstood the question):

When I saw the original code, it (to me) looked like the logic for encapsulating everything you needed to do when someone entered a triggerarea was in 2 places:

1. The physical heavy lifting of actually performing all the logic necessary to invoke the result of stepping into (or stepping out of) a trigger area happened [here](https://github.com/LandSandBoat/server/blob/base/src/map/lua/luautils.cpp#L2001) (for entering, exit is just below). It is loading the lua function to call, it's checking things on PChar, it's handling if the lua invocation fails (and notably, does not return any result about if it did succeed or fail to the caller). This means we exit this function having potentially successfully invoked the trigger area, and potentially failed in some way. The caller won't know.

2. State management for PChar to record which trigger area was entered happened with the caller, notably [here](https://github.com/LandSandBoat/server/blob/base/src/map/zone.cpp#L1260) which would either set a TriggerAreaID or a 0, depending on what happened, but irrespective of the call to the actual code that was meant to apply the trigger area to the PChar. 

So my thought was, that luautils call should be handling everything, including state. Only it knows if the invocation to the trigger area succeeded or not, and so it should handle updating the PChar to reflect this. Because with this code change, the trigger area will not be invoked more than once if someone is already inside of it.

If there are other areas of the code that also might want to consider trigger areas, my assumption is that you'd want 1 place to handle all logic correctly, so that you don't need to remind people "Hey, also make sure you call X after you call this". 

**However**, actually making me write this all out, I realize my changes are still incomplete as I should have updated [this](https://github.com/LandSandBoat/server/blob/base/src/map/lua/luautils.cpp#L2095) to return early if it was not a success.

So I can either:

A - add an early return to actually correctly implement the change I intended to make
or
B - move this back into zone and zone instance

100% your call as to how you want to see the code maintained, happy to do either - that's just the justification for my current approach.

zach2good: So generally, there isn't any state-altering logic happening in any luautils or Lua bindings, it's all just preparing and looking things up ready to _yeet_ things into Lua, and then handle return cases or errors etc. I was mainly going after the idea that adding triggerAreaIDs shouldn't be in the bindings. The original logic had the state being changed inside `CheckTriggerAreas`, so that's probably where it should stay. We have a lot of code that follows the pattern of:

```
change some state
lua binding reporting that state has changed
```

zach2good: (if the lua function fails, that's Lua's fault. The underlying state of "This char is in this trigger area" and the associated list should still be correct)

StabbyCutyou: Gotcha, happy to unwind that change. I'll push some updates in a bit and re-test.

StabbyCutyou:
Going to re-release to my local server and ensure this still works

StabbyCutyou:
Well...

So it might actually make a difference, behaviorally. Clay was smart to suggest testing it while logging in.

With the state set in zone/zoneinstance, logging in does not result in the trigger area correctly pushing a message to my client about the NPCs looking at me.

With the state set in luautils, I get the message every time.

I want to take some time to actually understand what the issue is before trying to merge anything.

StabbyCutyou:
Aaaah - it's this:

`if (PChar->GetLocalVar("ZoningIn") > 0)`, the luautils call returns early if it detects loading.

So now the check runs during loading, I tick off that i'm inside, but the lua function aborts and I don't get my trigger effect.

With that in mind, if you still want the state in the zone/zoneinstance, I think luautils would have to return a boolean or an error code of some kind to let the caller know it can apply the state.

Let me know how you'd prefer to handle it.

StabbyCutyou:
Despite my many foolish compilation errors, the last issue with the build I assume is a flake

https://github.com/LandSandBoat/server/actions/runs/13147827228/job/36690124457?pr=6946#step:10:105

WinterSolstice8:
Yeah those socket binds are probably caused by CI runners not actually being isolated and something else is actually using that port

zach2good (approved):
Great, thanks for this 👍

StabbyCutyou:
Sorry for commenting on a closed PR, but this still has an outstanding issue:

https://github.com/LandSandBoat/server/pull/6946#issuecomment-2635291238

I can address it however you'd like, but the net effect ATM is logging in while in a trigger zone means the zone doesn't fire until you exit and leave again.

zach2good:
I suppose the logic should be:

```
trigger_area_logic_tick:
    if zoning_in:
        don't try and insert into the trigger area

    otherwise:
        do regular logic
```

`OnTriggerAreaEnter` shouldn't have early exit, it should just "do the thing", the early exit should be in CZone, or wherever that trigger logic lives - it's good to find and remove this stuff and make the lua bindings as stupid as possible

zach2good:
I'm about to land https://github.com/LandSandBoat/server/pull/6959 shortly, can you do your follow up afterwards?

StabbyCutyou:
Yea sure thing, and your suggested change sounds reasonable.

claywar:
One thing to also test here is if any change breaks Abyssea staging area logic.  Its essentially one giant trigger area, and can't remember how logout is handled, but could potentially cheese the timer logic.  I want to say the visitant effect is lost on zone (which logout should do as well), and this isn't something to worry about, but just tracking thoughts prior to sufficient caffeine.

StabbyCutyou:
I can try to test that out as well in the mean time, thanks for the heads up.

PR #6963, opened 2025-02-06
Relocate triggerarea zoning check

StabbyCutyou:
**_I affirm:_**

Follows up on [this](https://github.com/LandSandBoat/server/pull/6946#issuecomment-2636440177) outstanding issue from PR https://github.com/LandSandBoat/server/pull/6946

Login while having logged out in a trigger zone
Notice that the trigger zone takes effect once you have "zoned in"
Ex: NPCs for the [Flyers for Regine](https://www.bg-wiki.com/ffxi/Flyers_for_Regine) quest will look at you upon logging in

Prior to this, after the changes in the original PR, the lua would not have run but your character would still be marked as being inside the trigger area.

# Notes

Following up here: [here](https://github.com/LandSandBoat/server/pull/6946#issuecomment-2636450338)

Zoning into Abyssea - Konschtat has the visitant status message appear.
Logging out and back in also has the message appear.

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -2001,12 +2001,6 @@ namespace luautils
     {
         TracyZoneScoped;
 
-        // Do not enter trigger areas while loading in. Set in xi.player.onGameIn
-        if (PChar->GetLocalVar("ZoningIn") > 0)
-        {
-            return;
-        }
-
         std::string                 filename;
         std::optional<CLuaInstance> optInstance = std::nullopt;
         if (PChar->PInstance)
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -1224,6 +1224,12 @@ void CZone::CheckTriggerAreas()
         // TODO: When we start to use octrees or spatial hashing to split up zones,
         //     : use them here to make the search domain smaller.
 
+        // Do not enter trigger areas while loading in. Set in xi.player.onGameIn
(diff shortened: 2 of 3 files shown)


WinterSolstice8 (approved):
This is a pretty old bug. If you really want to make sure it nails a well known bug dead, try `!zone empyreal paradox` and see if it shows the CS to use the teleporter

StabbyCutyou:
Yea sure, let me give it a whirl

StabbyCutyou:
Sure enough, upon logging in, I get the event which activates the prompt to teleport. It was a little slow to happen, but it did happen.

https://imgur.com/a/VOwGLGV

Also, upon logging in again after logging out in Abyssea having not obtained visitant status, my timer was reset to 5 minutes. I think clay alluded to an issue with this, but I honestly do not remember how retail worked in that regard, nor am I sure if my change resulted in that behavior.

WinterSolstice8:
Excellent. As for the no-visitant status resetting to 5 minutes I think that is a minor bug. Not very exploitable compared to retail when you can just enter/exit repeatedly anyway. As long as you cant move past the barriers its not a big deal

StabbyCutyou:
When I ran past the barriers, I got a 10 second warning - [twice](https://imgur.com/a/zgQtEYc).

But the rest of the countdown happened normally, and I was warped back behind them a bit when it completed.

WinterSolstice8:
yeah thats a bit out of scope for this thankfully. long as it works

PR #6985, opened 2025-02-08
[core] Handle augments with multiple mods

sruon:
- Stores augment data loaded from DB into a vector for each augment ID.
- Iterate over vector when applying the mod

**_I affirm:_**

Augments are loaded from database at start but do not account for the possibility of an augment having several modifiers attached to it.

```
INSERT INTO `augments` VALUES (553,0,9,1,0,0); -- DEX+1 AGI+1
INSERT INTO `augments` VALUES (553,0,11,1,0,0); -- Cont.
```

The existing logic effectively overwrites the map entry with the latest row loaded from database, leading to augments only applying a portion of their effects.

This PR stores a vector for each augmentId in the augmentData map, and iterates over it when applying effects to items.

Bug introduced in #6445 

Add a rabbit charm with augment 553 (DEX+/AGI+)

!additem 13112 1 553 0

Before PR: Only AGI+1 is applied
With PR: DEX+1 and AGI+1 are applied

### DIFF
--- a/src/map/items/item_equipment.cpp
+++ b/src/map/items/item_equipment.cpp
@@ -46,7 +46,8 @@ namespace
         uint8  petType;
     };
 
-    std::unordered_map<uint16, AugmentDataRow> sAugmentData;
+    using AugmentDataRows = std::vector<AugmentDataRow>;
+    std::unordered_map<uint16, AugmentDataRows> sAugmentData;
 } // namespace
 
 void CItemEquipment::LoadAugmentData()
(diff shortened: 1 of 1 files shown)


src/map/items/item_equipment.cpp
```diff
@@ -426,28 +429,31 @@ void CItemEquipment::SetAugmentMod(uint16 type, uint8 value)
         ref<uint8>(m_extra, 0x01) |= 0x03;
     }
 
-    const auto& augmentData = sAugmentData[type];
+    const auto& augmentDataRows = sAugmentData[type];
```
WinterSolstice8: looks like this might be shadowing `using AugmentDataRows = std::vector<AugmentDataRow>;`, CI should probably trip over this

WinterSolstice8:
Looks like windows 64 bit CI is still tripping on itself, I'd ignore it if it keeps failing for cmake

src/map/items/item_equipment.cpp
sruon: changed the name to augmentDataModifiers

WinterSolstice8 (approved):

zach2good (approved):

PR #6986, opened 2025-02-08
Fix issue with GetBestAgainstTargetWeakness (#6974)

### LINKED ISSUE #6974: 🐛 Map crash after recent gambit changes.
**_I affirm:_**

Windows 10

`base`

## Steps to reproduce

Call a trust that uses best against target magic and as soon as they cast, map server crash.

https://github.com/LandSandBoat/server/pull/6967

## Expected behavior

Trust can cast best against target without a crash.

Eggowaffle1288:
**_I affirm:_**

map server crashing as per bug report 
Closes #6974
will now work if spell id is given or not

summon a trust with best against target gambit with no spell id, will cast best spell against target or best spell if cant find any weakness

### DIFF
--- a/src/map/mob_spell_container.cpp
+++ b/src/map/mob_spell_container.cpp
@@ -346,18 +346,61 @@ std::optional<SpellID> CMobSpellContainer::GetBestAgainstTargetWeakness(CBattleE
     // clang-format on
 
     std::size_t weakestIndex = std::distance(resistances.begin(), std::min_element(resistances.begin(), resistances.end()));
-
-    // TODO: Figure this out properly:
+    std::optional<SpellID> choice     = std::nullopt;
     auto Weakness_Element = weakestIndex + 1;
-    auto Spell_Element    = spell::GetSpell(spellId)->getElement();
-    if (Spell_Element == Weakness_Element)
+    if (spell::GetSpell(spellId) != 0)
     {
-        return spellId;
+    	auto Spell_Element    = spell::GetSpell(spellId)->getElement();
+    	if (Spell_Element == Weakness_Element)
+    	{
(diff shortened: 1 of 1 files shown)


Eggowaffle1288:
best of both worlds.
use specific spells or dont

zach2good (approved):
As a fix I think this is fine, thanks for coming back to patch it up 🙏 

In general, I think this function is a bit counter-intuitive:
```
std::optional<SpellID> CMobSpellContainer::GetBestAgainstTargetWeakness(CBattleEntity* PTarget, SpellID spellId)
```

You could pass in... Flare? and it would then pick something from { Fire I, Fire II, Fire III }, etc.

Since SpellFamily is groupings of spells, not overall spell element.

After this PR, I'd probably want this changed to:
- No additional argument
- Same logic for looking up the best weakness for the target
- Go through the spell list for all damaging spells that match the weakness element
- Pick the "best" of those

Eggowaffle1288:
I created the bug, I should fix it.

i think i understand what your saying.
i think what im trying to do should be pushed to another function then and this should be tweaked to do what you have suggested.
(i can work on it later tonight and get that done)

Yes if you pass flare it will check flare vs weakness and then cast it if weak to fire.

Just wanted to add a bit more functionality to the function for anyone who wants to be more specific with trusts like D. Shantotto or Kayeel-Payeel where they can cast ga spells or ancient spells.

I think I was reading the function name and interoperating it differently then what was intended.
most of the gambits i saw used a spell argument so i read that as

cast spell argument if its best against the targets weakness 
not 
cast your best spell against the target weakness.

just explaining my thought process incase someone comes in to find out why i did what i did.

PR #6988, opened 2025-02-08
Abyssea tether timer fix

StabbyCutyou:
**_I affirm:_**

This follows up on my [prior PR ](https://github.com/LandSandBoat/server/pull/6963) where I [noticed](https://github.com/LandSandBoat/server/pull/6963#issuecomment-2638567668) that the Abyssea tether timer was double-printing the timer warning for 10 seconds.

I made two changes here:

1. It was sending both `RETURNING_TO_SEARING_IN 7348` and `NO_VISITANT_WARD 7349` due to the way the logic was programmed for `onWardTriggerAreaLeave`, `tetherTimer == 11` and `tetherTimer <=6`. Since the messages are slightly different, I just removed the handling for `== 11`, as that was the cause of the double print in the `searingWardTimer` method.

2. I noticed the timer callback logic, and it seemed a bit curious to me. On a lark, I removed it entirely in favor of relying on `onEffectTick` to see how the timer status would behave, and... it was fine. The `onEffectTick` loop seems to invoke it at about the same cadence (I did not time it, just eyeballed it) and it looked to ultimately behave in the exact same manner. I figure if the outcome is the same, less code and less callbacks is easier to maintain, but i am happy to undo that part of the change if there is some other reason you all prefer that pattern.

Enter Abyssea Konschtat
Do not receive Vistant Status
Exit through the Searing ward
Watch the timer tick
Notice you only see the warning at 10 seconds, then 5,4,3,2,1
Get warped back inside the staging area

## Notes

While testing this, I noticed that the actual visitant timer which should be ejecting me out of Abyssea is not actually kicking me out. That was happening before I made any changes in this PR, and I can either try to debug and fix as part of this PR, or I could make it a different PR.

After testing this a bit, it seems like the issue is if you enter, do not gain actual visitant status (so, just the hidden temp status), logout, then log back in - the game hits an edge case and does not properly eject you. I'm making a separate branch to confirm the cause + try to put a fix in. I suspect something with `gameLogin` is not getting cleared properly, but I'll open that branch when I know for sure.

It seemingly works fine if you zone in clean, then wait the 5 minutes.

Edit - PR for the visitant status bug here https://github.com/LandSandBoat/server/pull/7000

### DIFF
--- a/scripts/globals/abyssea.lua
+++ b/scripts/globals/abyssea.lua
@@ -1206,11 +1206,11 @@ xi.abyssea.searingWardTimer = function(player)
     local tetherTimer = player:getLocalVar('tetherTimer')
 
     if tetherTimer > 1 then
-        if tetherTimer == 11 or tetherTimer <= 6 then
+        player:setLocalVar('tetherTimer', tetherTimer - 1)
+        if tetherTimer <= 6 then
             player:messageSpecial(ID.text.RETURNING_TO_SEARING_IN, tetherTimer - 1)
         end
 
-        player:setLocalVar('tetherTimer', tetherTimer - 1)
         player:timer(1500, function()
             xi.abyssea.searingWardTimer(player)
         end)


scripts/globals/abyssea.lua
```diff
@@ -1206,14 +1206,11 @@ xi.abyssea.searingWardTimer = function(player)
     local tetherTimer = player:getLocalVar('tetherTimer')
 
     if tetherTimer > 1 then
-        if tetherTimer == 11 or tetherTimer <= 6 then
+        player:setLocalVar('tetherTimer', tetherTimer - 1)
```
StabbyCutyou, line 1209: I moved this to be the first thing we did so it matched the way the logic in the `==1` case worked, which was to first and foremost update the tetherTimer value, which seemed most correct to me.

Happy to move it back if there is another reason for keeping it as it was.

WinterSolstice8 (approved):
Seems fine to me

StabbyCutyou:
It does seem like the change here would give folks more time - I think onEffectTick is ~3secs and the original approach with the callback was giving ~1.5secs between timer ticks. So my removal of the callback does have a drawback, although being slightly more generous with the eviction timer isn't the worst outcome - but that isn't really my call, so happy to change it back if need be.

scripts/globals/abyssea.lua
```diff
@@ -1206,14 +1206,11 @@ xi.abyssea.searingWardTimer = function(player)
...
+
+        if tetherTimer <= 6 then
             player:messageSpecial(ID.text.RETURNING_TO_SEARING_IN, tetherTimer - 1)
         end
-
-        player:setLocalVar('tetherTimer', tetherTimer - 1)
-        player:timer(1500, function()
```
claywar, line 1214: By removing this timer, you're relying on the effect tick of 3s as opposed to 1.5s.  I believe that makes this countdown much longer in that case (if I remember correctly, I went with 1.5s due to timing as well between client and server)

StabbyCutyou: Yea I realized that a bit late, originally I thought for some reason they had similar tick times and were redundant, but later realized in a comment my mistake.

I should be able to just drop that commit and then it'll just be the message fix.

StabbyCutyou:
Dropped the offending commit, should now be back to relying on the callback timer and just removes the clause that resulted in the double-printing. Just getting the formatting checks happy with me now.

I can give this another test later tonight just in case anything else subtle broke with the rebase.

StabbyCutyou:
Retested and still working as expected - zone in, leave the ward, given a single 10 second warning once i step outside, then the countdown warps me back.

Xaver-DaRed (approved):

PR #6989, opened 2025-02-08
Fix vertical rendering distance

### LINKED ISSUE #6969: 🐛 Several Pso'Xja elevators are broken
**_I affirm:_**

Win11

## Branch affected by issue 

`base`

## Steps to reproduce
- !zone Pso'Xja
- !gotoname @096 (@095 as well)
- Ride the elevator down

Character ends up stuck about 10ft up in the air and cannot exit the elevator.

## Expected behavior
Character can exit the elevator.

cocosolos:
**_I affirm:_**

Fixes #6969

Elevator going out of the 20' vertical range was causing issues. I set `ENTITY_VERTICAL_RENDER_DISTANCE` to 50 and it worked and at that point `isWithinDistance` can already handle the vertical check with `ENTITY_RENDER_DISTANCE` so I just got rid of the vertical stuff. Alternatively I can probably remove the vertical checks in just SpawnNPCs but that would also require reworking TryAddToNearbySpawnLists. I'm not sure how it works on retail or what the goals are here so let me know if a different approach is wanted.

See: #6969

### DIFF
--- a/src/map/zone_entities.cpp
+++ b/src/map/zone_entities.cpp
@@ -152,58 +152,60 @@ void CZoneEntities::TryAddToNearbySpawnLists(CBaseEntity* PEntity)
 
     FOR_EACH_PAIR_CAST_SECOND(CCharEntity*, PCurrentChar, m_charList)
     {
-        const auto isInHeightRange = isWithinVerticalDistance(PEntity, PCurrentChar);
-        const auto isInRange       = isWithinDistance(PEntity->loc.p, PCurrentChar->loc.p, ENTITY_RENDER_DISTANCE);
+        const auto isInRange = isWithinDistance(PEntity->loc.p, PCurrentChar->loc.p, ENTITY_RENDER_DISTANCE);
 
-        if (isInHeightRange && isInRange)
+        if (isInRange)
         {
-            switch (PEntity->objtype)
+            // Exclude NPCs from vertical rendering limits (elevators may go past the normal vertical range)
+            if (PEntity->objtype == TYPE_NPC)
             {
-                case TYPE_PC:
+                PCurrentChar->SpawnNPCList[PEntity->id] = PEntity;
+                PCurrentChar->updateEntityPacket(PEntity, ENTITY_SPAWN, UPDATE_ALL_MOB);
+            }
+            else if (isWithinVerticalDistance(PEntity, PCurrentChar))
+            {
+                switch (PEntity->objtype)
                 {
-                    auto* PChar = static_cast<CCharEntity*>(PEntity);
-                    if (PChar->m_moghouseID != PCurrentChar->m_moghouseID)
+                    case TYPE_PC:
                     {
-                        continue;
-                    }
+                        auto* PChar = static_cast<CCharEntity*>(PEntity);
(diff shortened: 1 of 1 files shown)


zach2good:
I took the concept of `isWithinVerticalDistance`

```cpp
bool isWithinVerticalDistance(CBaseEntity* source, CBaseEntity* target)
{
    return abs(target->loc.p.y - source->loc.p.y - 0.5) <= 20;
}
```

from Eden; where they sometimes also wrap it up in this:

```cpp
static bool shouldShowEntity(CBaseEntity* source, CBaseEntity* target)
{
    return target->status != Status::Disappear
           && source->InGridRange(target->gridloc)
           && isWithinVerticalDistance(source, target)
           && distance(source->loc.p, target->loc.p) < 50;
}
```

They use `isWithinVerticalDistance` in `SpawnPCs` and `SpawnPETs`, and `shouldShowEntity` only in `SpawnMOBs` - so I was over-zealous with those checks and they shouldn't be being applied to `SpawnNPCs`. 

Rather than ripping out the entire concept, I'd prefer if you just remove the vertical component from SpawnNPCs and adjust TryAddToSpawnList to something like:
```cpp
void CZoneEntities::TryAddToNearbySpawnLists(CBaseEntity* PEntity)
{
    TracyZoneScoped;

    FOR_EACH_PAIR_CAST_SECOND(CCharEntity*, PCurrentChar, m_charList)
    {
        const auto shouldAdd = [&]()
        {
            const auto isInHeightRange = isWithinVerticalDistance(PEntity, PCurrentChar);
            const auto isInRange       = isWithinDistance(PEntity->loc.p, PCurrentChar->loc.p, ENTITY_RENDER_DISTANCE);

            if (PEntity->objtype == TYPE_NPC)
            {
                return isInRange;
            }
            else
            {
                return isInHeightRange && isInRange;
            }
        }();
        if (shouldAdd)
        {
```

zach2good:
Maybe also a comment saying why they have some special logic

cocosolos:
Is there a reason that vertical distance is smaller than the regular rendering distance? `isWithinDistance` has a default `ignoreVertical = false` arg so vertical distance is already checked with that. It seems weird not to have the rendering field be a sphere.

zach2good:
Elevators aside, it doesn't seem particularly common to need to see other chars, pets, or mobs 50' above/below you - so a nice reduction in search space

cocosolos:
I changed it to just exclude NPCs from the additional vertical check, looks like the elevators work fine still. 20' does seem fine for a vertical range, it's just that particular elevator goes down like 3 floors and the NPC itself seems to be set in the middle about 25' away from the top and bottom platforms. Maybe elevators should be broadcasting their updates like the Altepa gate pilars do? Then you could cull other NPCs that aren't like environment objects. Think this is fine for now though.

WinterSolstice8:
I am pretty sure the Bastok drawbridge "assembling itself" & some airship jank (missing propellers?) is caused by the distance culling. Maybe even boats but SE took great care for boats to not be visible unless you're fairly close.

WinterSolstice8:
p.s. Windows CI has been losing its mind recently on cmake failures so don't worry about it

WinterSolstice8 (approved):

zach2good (approved):

PR #6990, opened 2025-02-08
Core: Convert transactions to use db::transaction

zach2good:
**_I affirm:_**

Lots of testing to do with this one. Quite frankly it's a miracle any of this was safe/correct to start with.

Also contains some misc cleanup and updates to helpers.


modules/custom/cpp/ah_announcement.cpp
```diff
@@ -183,17 +165,14 @@ class AHAnnouncementModule : public CPPModule
...
-
-                                        itemPurchasedSuccessfully = true;
                                     }
                                 }
-                            }); // TransactionWrapper
-                            // clang-format on
-
-                            if (itemPurchasedSuccessfully)
+                            });
+                            if (success)
```
WinterSolstice8, line 171: Is this if check even doing anything?

WinterSolstice8: looks like if success isnt true this happens:

`PacketParser[0x04E] = newHandler;`

Edit: thats inside the handler, github made it look like it was something else. Maybe it does do nothing?

src/map/utils/dboxutils.cpp
```diff
@@ -200,52 +200,48 @@ void dboxutils::SendOldItems(CCharEntity* PChar, uint8 action, uint8 boxtype)
...
+                if (PItem != nullptr)
                 {
-                    PItem->setSubID(_sql->GetIntData(1));
-                    PItem->setSlotID(_sql->GetIntData(2));
-                    PItem->setQuantity(_sql->GetUIntData(3));
+                    PItem->setSubID(rset->get<uint16>("itemsubid"));
+                    PItem->setSlotID(rset->get<uint8>("slot"));
+                    PItem->setQuantity(rset->get<uint32>("quantity"));
 
-                    if (_sql->GetUIntData(4) > 0)
```
WinterSolstice8, line 223: Glad this line is changed, no comment here wasn't helping it

WinterSolstice8 (approved):

Xaver-DaRed (approved):

modules/custom/cpp/ah_announcement.cpp
zach2good: The logic is:

If packet comes in, and is: `const auto action = data.ref<uint8>(0x04);`, do module logic. This very last sections returns out if everything was successful. If not it sends some packets saying things failed

zach2good:
Tested back and forth, up and down, inside out - everything is working exactly as it was before. 
- Sending to self
- Cancelling
- Receiving
- AH interaction
- Sending to other accounts
- Cancelling
- Trying to operate DBOX very quickly, etc.

PR #6993, opened 2025-02-08
[SQL] Adjust DRK trait "Resist Paralyze" acquisition level

Flam9:
**_I affirm:_**

The trait `Resist Paralyze` had the wrong content_tag set for Corsair/Dark Knight as well as the wrong level for tier 3 on Dark Knight.

Source: https://ffxiclopedia.fandom.com/wiki/Resist_Paralyze?diff=503918&oldid=499472

N/A

### DIFF
--- a/sql/traits.sql
+++ b/sql/traits.sql
@@ -335,7 +335,7 @@ INSERT INTO `traits` VALUES (49,'resist poison',11,60,3,241,20,NULL,0);
 INSERT INTO `traits` VALUES (49,'resist poison',11,81,4,241,25,NULL,0);
 INSERT INTO `traits` VALUES (50,'resist paralyze',8,20,1,242,10,NULL,0);
 INSERT INTO `traits` VALUES (50,'resist paralyze',8,40,2,242,15,NULL,0);
-INSERT INTO `traits` VALUES (50,'resist paralyze',8,50,3,242,20,NULL,0);
+INSERT INTO `traits` VALUES (50,'resist paralyze',8,60,3,242,20,NULL,0);
 INSERT INTO `traits` VALUES (50,'resist paralyze',8,75,4,242,25,NULL,0);
 INSERT INTO `traits` VALUES (50,'resist paralyze',8,81,5,242,30,'ABYSSEA',0);
 INSERT INTO `traits` VALUES (50,'resist paralyze',17,5,1,242,10,'TOAU',0);


LacunaXI:
I looked through the main update patch notes at http://www.playonline.com/pcd/verup/ff11us/update.html and did not see any mention of the Resist Paralyze trait being increased after the launch of Corsair. Patch notes did include mention of other trait changes like when an extra tier of Accuracy Bonus was added to Dragoon.

Checking the edit history of BG had these traits listed in 2007 where the FFXIClopedia link you shared was from 2008. https://www.bg-wiki.com/index.php?title=Resist_Paralyze&direction=next&oldid=22979

I would suspect that this trait was available before WOTG and was just not documented on the wiki.

Flam9:
Hi @LacunaXI,

Thanks for looking into this! I made changes based on your discoveries. Does this look fine?

LacunaXI:
You are welcome. Thank you for the changes.

I believe your change for tier 3 of the trait for DRK from 50 to 60 is correct. JP Wiki confirms it is level 60 and not 50. https://wiki.ffo.jp/html/3336.html Other traits follow this general 20-40-60 pattern too. BG should be corrected to reflect this.

As far as the content tag goes. Is there any evidence that DRK only obtained tier IV at 75 during TOAU and not earlier? Paladin gains Resist Sleep IV at 75 in the same level pattern as this for example.

I believe there may have been confusion on my prior remark about the trait not increasing since the launch of Corsair. I specifically was commenting to Corsair itself (http://www.playonline.com/pcd/update/ff11us/20060418lfQ7d3/detail.html) in the PR and not the entirety of the trait itself for DRK. Essentially I would expect that DRK always had access to tier IV as long as Paladin has by reaching level 75 and was just was not documented or known by players due to poor communication from Square.

Flam9:
You know, I had double checked the DRK change but I somehow tricked myself. You are absolutely right. Down to a tiny change. Thanks again!

Flam9:
CI Failed but this is a Squash and Merge situation anyways...

WinterSolstice8:
You can rebase and merge those two commits into one, it's just complaining that the commit itself has a short name

WinterSolstice8 (approved):

zach2good:
We can squash commits on the way in, that's fine, but the general idea is that commits are meant to be logical units of work that contributors can read and follow what's been going on. `True dat`, `lol CI is garbage`, and `yeet` aren't conveying anything to future readers.

zach2good:
^ The way that our CI is set up, we can't even get to the squash button until `Sanity_Checks` passes - so it looks like you've got some git learnin' to do

zach2good (approved):

Flam9:
np, fixed!

Xaver-DaRed (approved):

WinterSolstice8 (approved):

PR #6995, opened 2025-02-08
[Nyzul] Random layout generation implementation

Xaver-DaRed:
**_I affirm:_**

- Adds logic to generate a truly random layout for Nyzul Isle Investigation assault.
- Defines 1 of the 6 blocks of rooms, as an example.

Currently, while we randomize mobs and objectives, we do not truly randomize the layouts. This is because we do not handle door opening nor closing, for statrters. So this attempts to randomize in a logical way which doors open and close.

Process:
1. We select a block (Currently 1 only defined, but there are 6)
2. We select a room group at random to act as the start.
3. We loop using the previous group as a start and using the groups connected to it as the choices to continue. From there, we add the possible groups, while removing the ones we use AND we do not allow the ones we used to be re-added. All until we reach the requested amount of groups.
4. Once we have the room list, we populate several tables with the information we want, and we are done.

Experimental. Not implemented.

### DIFF
--- a/scripts/globals/nyzul/floor_layout.lua
+++ b/scripts/globals/nyzul/floor_layout.lua
@@ -1,0 +1,506 @@
+-----------------------------------
+-- Nyzul Isle: Floor generation methods and data.
+-----------------------------------
+local ID = zones[xi.zone.NYZUL_ISLE]
+-----------------------------------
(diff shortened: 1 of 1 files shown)


zach2good (approved):
Cool!

PR #6997, opened 2025-02-09
Mobskills: Remove usage of DMG_VARIES, implement ATK_VARIES

WinterSolstice8:
**_I affirm:_**

Nuke all DMG_VARIES with `sed -i -E "s/xi.mobskills.physicalTpBonus.DMG_VARIES, [+-]?([0-9]*[.])?[0-9]+, [+-]?([0-9]*[.])?[0-9]+, [+-]?([0-9]*[.])?[0-9]+/xi.mobskills.physicalTpBonus.NO_EFFECT, 0, 0, 0/g" *` in mobskills dir
Minor audit of FTPs that were effected with stuff in jimmy's sheet

Implement ATK_VARIES tp bonus

There are too many mobskills here to check, but given the nature of the edits it is likely nothing is wrong. I explicitly tested death scissors and a few others, but it's just not reasonable to test them all. I tried to make a smoke-test tool but it is not going to be a same-night thing.

### DIFF
--- a/scripts/actions/mobskills/abyssal_strike.lua
+++ b/scripts/actions/mobskills/abyssal_strike.lua
@@ -15,8 +15,8 @@ end
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
     local numhits = 1
     local accmod  = 2
-    local ftp     = 4.7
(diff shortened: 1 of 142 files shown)


Xaver-DaRed (approved):
Other than "could we stop using 1.0, 2.0, etc, because this isnt c++"  looks good to me

PR #7000, opened 2025-02-09
Move gameLogin 0 to end of onGameIn so it always fires

StabbyCutyou:
**_I affirm:_**

Moves the clearing of `gameLogin` from a conditional path which does not get invoked on game login, to a location where it will always fire.

Logout in Abyssea
Log back in
Wait out the timer
Notice you get the exit-event and a nice CS teleporting you to the zones exit point

## Without these changes

You can enter Abyssea, not get visitant status, logout, log back in, and allow the timer to expire.
You will not be removed from the zone due to the way the logic is currently handled (see below).

## Notes

This is what seems to be happening:

1. When we first login, [zoning](https://github.com/LandSandBoat/server/blob/base/src/map/utils/charutils.cpp#L717) is set to 2 - this invokes the call to set the `gameLogin` value initially
2. This means that when we now invoke `OnGameIn` [here](https://github.com/LandSandBoat/server/blob/base/src/map/utils/charutils.cpp#L864), `zoning == 1` will resolve to `false`
3. So when we ultimately invoke the lua method [here](https://github.com/LandSandBoat/server/blob/48ba9b1d551e41d15589b456c5f7ada89595c7d2/src/map/lua/luautils.cpp#L1873)...
4. We will not enter the conditional block that would clear gameLogin, because [zoning == false](https://github.com/LandSandBoat/server/blob/48ba9b1d551e41d15589b456c5f7ada89595c7d2/scripts/globals/player.lua#L152), skipping this [else](https://github.com/LandSandBoat/server/blob/935ec4a3c0098ff2648be28a456f2faf1abe6611/scripts/globals/player.lua#L156) block entirely, which means this [line](https://github.com/LandSandBoat/server/blob/48ba9b1d551e41d15589b456c5f7ada89595c7d2/scripts/globals/player.lua#L171) is never hit
5. Because of this, when the effect is lost, we fail [this](https://github.com/LandSandBoat/server/blob/48ba9b1d551e41d15589b456c5f7ada89595c7d2/scripts/effects/visitant.lua#L103) conditional check and never get [exited](https://github.com/LandSandBoat/server/blob/48ba9b1d551e41d15589b456c5f7ada89595c7d2/scripts/effects/visitant.lua#L108).

I'm uncertain if there is a better location for this, but just by the name of it, it would seem that it should get cleared out when the method exits. It only seems to be involved in Abyssea visitant status defense atm, so it doesn't seem like moving the call would have wide spread repercussions.

I've so far tested, post-fix:

1. Logging in with a character who lost temporary visitant from the bug, without eviction. They are given 30 seconds, then they are exited via the cutscene.
2. Logging in with a character who still has temporary visitant from a prior login, who has seen the 3 minute warning. They get the message about 5 minutes total, but the countdown timer resumes at the next warning interval - in this case, 2 minutes.
3. Zoning in fresh with a character who does not obtain proper visitant status. They are reported to receive 5 minutes, and at the end of the 5 minutes, are exited via the cutscene.

I have not yet tested the results with actual visitant status, although it seems like mostly the same effect, just with an icon present to indicate if it was real or temporary. But probably should still be tested.

I think the UX in point 2 is probably ok, but I can try to test retail to confirm - the 5 minute warning is generic to entering Abyssea without status, where as the timer intervals are specific to the status time remaining on your character.

### DIFF
--- a/scripts/globals/player.lua
+++ b/scripts/globals/player.lua
@@ -167,8 +167,6 @@ xi.player.onGameIn = function(player, firstLogin, zoning)
             player:messageSpecial(ID.text.ABYSSEA_TIME_OFFSET + 8)
             player:setPos(unpack(xi.abyssea.exitPositions[zoneID]))
         end
-
-        player:setLocalVar('gameLogin', 0)
     end
 
     -- Abyssea starting quest should be flagged when expansion is active
@@ -237,6 +235,10 @@ xi.player.onGameIn = function(player, firstLogin, zoning)
         -- Login Campaign rewards points once daily
         xi.events.loginCampaign.onGameIn(playerArg)
     end)
+
+    -- Enforce that gameLogin is always set to 0 once this method exits
+    -- This assists with ensuring Abyssea visitant status is handled properly on logins
+    player:setLocalVar('gameLogin', 0)
 end
 
 xi.player.onPlayerDeath = function(player)


claywar:
So for both of those statuses, they should continue to tick down regardless of online state.  Zoning into the same zone should have no impact on the timer or status in the area.  If you've got time on the clock, its time on the clock.  

EDIT: The only difference between the two is more of a state change to allow for other interactions, as opposed to just being time-based (invisible, vs visible status of the effect)

StabbyCutyou:
Yea that makes sense, and I think is consistent with what my testing showed as well.

Just wanted to try to be explicit about what I saw / what I had tested.

WinterSolstice8 (approved):

PR #7009, opened 2025-02-10
[Fix] Correct Aquarius Fransisca drop

Xaver-DaRed:
**_I affirm:_**

As title sais
https://www.ffxidb.com/zones/153/aquarius

As noted, the rates among TH tiers suggest that the drop is grouped, so TH doesnt apply, which means we can use the median to guess the rate at 90% 

If anything, it isnt guaranteed anyway.

None

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -1336,7 +1336,8 @@ INSERT INTO `mob_droplist` VALUES (147,4,0,1000,864,0); -- Handful Of Fish Scale
 INSERT INTO `mob_droplist` VALUES (148,0,0,1000,557,@RARE);   -- Ahriman Lens (Rare, 5%)
 
 -- ZoneID: 153 - Aquarius
-INSERT INTO `mob_droplist` VALUES (149,0,0,1000,17925,@ALWAYS); -- Fransisca (Always, 100%)
+INSERT INTO `mob_droplist` VALUES (149,1,1,@ALWAYS,17925,900); -- Fransisca (Group 1, Always, 100% - 90%)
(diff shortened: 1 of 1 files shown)


CriticalXI:
I don't believe that the drop is grouped. If an item is rare it won't drop whatsoever, so the reported drop rate from FFXIDB is lowered due to this. If anyone running the guildwork addon had francisca already in their inventory somewhere and they killed aquarius in a group where everyone had it already, then it would report a lower drop rate.

CriticalXI:
You can see this on Jelly ring, coffer keys, any item where the drop is rare the ffxidb rate is going to be lower than it actually is.

Xaver-DaRed:
This isnt a mere matter of data pollution due to how rare/ex items work.
We are also not speaking about a regular mob one can jusg kill hundreds of.

This is a NM on the route to Fafnir. One could argue it could have been killed by ppl on the way to it, but considering that in era you probably couldnt solo it, the fact its a lotery mob and the fact the only drop this mob has is the fransisca, I find it incredibly difficult to believe its a mere matter of data pollution. We are speaking about over 11.5% of the kills happened when everyone involved had the fransisca, a mob that had 0 reason to be killed for other than for it.

If we doubt such a significant error margin, at that point we would have to doubt every single drop rate we have, making the whole process pointless. This isnt a 1 or 5 %, this is 11.5%.

Since data pollution is a real thing, I opted to round the drop up instead of down. I will remind that whenever we are in doubt, the procedure is to always adjust in player's detriment, because that is less hurtfull to downstream servers than to adjust in players favor.

zach2good (approved):

PR #7010, opened 2025-02-10
[Chore] Rename magic accuracy modifiers and fix some modifier names in sql

Xaver-DaRed:
**_I affirm:_**

As title sais. Prep-work to untangle elemental affinities.

None. No real changes were made.
Review by commit, but its a matter of mass search/replace


src/map/modifier.h
```diff
@@ -86,14 +86,14 @@ enum class Mod
...
-    THUNDERATT = 36, // Thunder Damage
-    WATERATT   = 37, // Water Damage
-    LIGHTATT   = 38, // Light Damage
-    DARKATT    = 39, // Dark Damage
+    FIREATT      = 32, // Fire Damage
+    ICEATT       = 33, // Ice Damage
+    WINDATT      = 34, // Wind Damage
+    EARTHATT     = 35, // Earth Damage
+    THUNDERATT   = 36, // Thunder Damage
+    WATERATT     = 37, // Water Damage
```
Flam9, line 94: should these be `WATER_ATT` for consistency?

Xaver-DaRed: While I personally agree, I am doing this in small chuncks for a reason. So that is a change for another PR.

In an unrelated note, we dont even use those modifiers, reason why I didnt rename them: Left with that horrible name marks them as "I have an unresolved issue, please revisit"

WinterSolstice8 (approved):

PR #7011, opened 2025-02-10
[lua] rewrite logic to spawn skull NMs in Eldieme Necropolis

Grahf0085:
**_I affirm:_**

This PR makes the following changes when lighting the tallow candles in Eldieme Necropolis:

- flint stones are no longer consumed when trading them tot he tallow candle
- each candle stays lit for five minutes - in current code it was 40 seconds I think?
- added the "nothing happened" message special when trading a flint stone to the tallow candles when trading occurs outside time frame to trade to them
- an animation was added to the tallow candles when they are lit so they look like they're on fire when they're lit

This PR should fix the following issues that are present in the current code:

- checking a tallow candle that was already lit would tell you to trade a flint stone to it to light it - even though it was already lit
- trading multiple flint stones to the same candle would cause the count down to spawning the Skull NMs to progress - it should be only one trade per tallow candle. This behavior was sporadic - not sure what triggered it. I would just go to random candles and try trading multiple flint stones and sometimes that would cause this bug

I looked at this capture: https://www.youtube.com/watch?v=8bOepLkqArA

- Here it shows the flint stone isn't consumed when traded: https://youtu.be/8bOepLkqArA?feature=shared&t=207
- In the video the lamp it lit at 3:30 and goes out at 9:10. That means it stays lit for 5 minutes and 40 seconds. I'm wondering if my math is off or I'm missing something because 5 minutes and 40 seconds is a weird number. I coded it to stay lit for 5 minutes because that's what the ffxi encyclopedia says. Not sure how everyone feels about this.

1. !gotoid 17576335

- checking NPC should tell you that you can trade a flint stone to it
- trade a flint stone to NPC
- flint stone should stay in inventory
- lamp should start a fire
- should get dialogue saying the lamp was lit and how many remain to be lit
- checking lamp again should tell you it's lit

3. Repeat by adding one to id until you reach 17576342
4. All seven NMs should spawn
5. You should not be able to light the lamps again for an hour

### DIFF
--- a/scripts/zones/The_Eldieme_Necropolis/globals.lua
+++ b/scripts/zones/The_Eldieme_Necropolis/globals.lua
@@ -4,57 +4,61 @@
 -----------------------------------
 local ID = zones[xi.zone.THE_ELDIEME_NECROPOLIS]
 -----------------------------------
+local spawnSkulls = function()
+    -- Spawn all 7 Skulls
+    for skull = 1, 7 do
+        SpawnMob(ID.mob.LICH_C_MAGNUS + skull) -- IDs based off Lich C Magnus
+    end
 
--- Handle 7 Sins Skeleton NMs Spawns
-local skullTrade = function(player, npc)
-    local candleCount =
-    {
-        ID.text.SKULL_FIVE_REMAIN,
-        ID.text.SKULL_FOUR_REMAIN,
-        ID.text.SKULL_THREE_REMAIN,
-        ID.text.SKULL_TWO_REMAIN,
-        ID.text.SKULL_ONE_REMAIN,
-        ID.text.SKULL_SPAWN,
-    }
+    -- Used to stop Skulls from being spawned until an hour from being spawned
+    SetServerVariable('[ELDIEME]TimeToRespawnSkulls', os.time() + 3600)
+end
 
-    local tradeCount = GetNPCByID(ID.npc.CANDLE_OFFSET):getLocalVar('SkullTradeCount')  -- Track how many candles have been lit
-    local tradeWindow = GetNPCByID(ID.npc.CANDLE_OFFSET):getLocalVar('SkullTradeTimer') -- Track how much time before candles reset
-    local active = npc:getLocalVar('candleActive')                                      -- Track if current candle has already been lit
+local brazierMessages =
+{
+    ID.text.SKULL_SPAWN,
+    ID.text.SKULL_ONE_REMAIN,
+    ID.text.SKULL_TWO_REMAIN,
+    ID.text.SKULL_THREE_REMAIN,
+    ID.text.SKULL_FOUR_REMAIN,
+    ID.text.SKULL_FIVE_REMAIN,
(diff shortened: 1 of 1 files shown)


scripts/zones/The_Eldieme_Necropolis/globals.lua
```diff
@@ -4,57 +4,60 @@
...
+    local messageAfterLightingBrazier = brazierMessages[numberNotLit + 1]
+
+    player:messageSpecial(messageAfterLightingBrazier)
+
+    if numberNotLit == 0 then
+        spawnSkulls()
+    end
+end
+
+local eldiemeGlobal = {
```
Xaver-DaRed, line 61:
```lua
local nameOfTheTable =
{
```
This applies for all tables and such

scripts/zones/The_Eldieme_Necropolis/globals.lua
```diff
@@ -72,7 +75,7 @@ local eldiemeGlobal =
         -- gates are grouped in groups of five. even numbered groups share one animation, while odd numbered groups share the other.
 
         local animEven = (npc:getAnimation() == xi.anim.OPEN_DOOR) and xi.anim.CLOSE_DOOR or xi.anim.OPEN_DOOR
-        local animOdd  = (npc:getAnimation() == xi.anim.OPEN_DOOR) and xi.anim.OPEN_DOOR or xi.anim.CLOSE_DOOR
+        local animOdd = (npc:getAnimation() == xi.anim.OPEN_DOOR) and xi.anim.OPEN_DOOR or xi.anim.CLOSE_DOOR
```
Xaver-DaRed: This was properly aligned

Grahf0085: It's my auto formatter 
/cry

scripts/zones/The_Eldieme_Necropolis/globals.lua
Grahf0085: sorry about that

WinterSolstice8 (approved):

scripts/zones/The_Eldieme_Necropolis/globals.lua
```diff
@@ -4,57 +4,61 @@
...
 
-local eldiemeGlobal =
-{
+local lightBrazier = function(player, npc)
+    npc:setAnimation(xi.anim.OPEN_DOOR)
+
+    --unlight brazier after five minutes
+    npc:timer(300000, function()
+        npc:setAnimation(xi.anim.CLOSE_DOOR)
+    end)
```
TeoTwawki, line 47: Just so you are aware in the future, there is literally a function specifically to open a door, wait a specified number of seconds, and then close the same door, performing all 3 of these functions with one invocation without any seconds to milliseconds conversion or potential snags you can hit by invoking `timer()` directly. There is also a matching function to do a the reverse (closing a default open door, and then reopening it after a timeout).

`npc:openDoor(300)` would have done exactly what you did here in a single line.

PR #7017, opened 2025-02-11
[core] Add synth state

WinterSolstice8:
**_I affirm:_**

This fixes an issue with packet loss or d/cing causing materials to be lost

synth and stuff

### DIFF
--- a/src/map/ai/states/synth_state.cpp
+++ b/src/map/ai/states/synth_state.cpp
@@ -1,0 +1,100 @@
+/*
+===========================================================================
+
+  Copyright (c) 2025 LandSandBoat Dev Teams
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see http://www.gnu.org/licenses/
+
+===========================================================================
+*/
+
+#include "synth_state.h"
+
+#include "entities/battleentity.h"
+
+#include "ai/ai_container.h"
+#include "packets/action.h"
+#include "packets/lock_on.h"
+#include "utils/battleutils.h"
+#include "utils/synthutils.h"
+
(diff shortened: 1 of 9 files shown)


WinterSolstice8:
Drafted because I need to dial in the synth timer a bit better. It appears around 16 seconds but I havent done a ton of testing with packets

src/map/packet_system.cpp
```diff
@@ -2693,10 +2693,7 @@ void SmallPacket0x058(map_session_data_t* const PSession, CCharEntity* const PCh
 void SmallPacket0x059(map_session_data_t* const PSession, CCharEntity* const PChar, CBasicPacket& data)
 {
     TracyZoneScoped;
-    if (PChar->animation == ANIMATION_SYNTH)
-    {
-        synthutils::sendSynthDone(PChar);
-    }
+    // Do nothing. This is handled in synth state.
```
zach2good, line 2696: Could we compare the synth state start time if this packet comes in to catch turds?

src/map/ai/states/synth_state.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "packets/action.h"
+#include "packets/lock_on.h"
+#include "utils/battleutils.h"
+#include "utils/synthutils.h"
+
+CSynthState::CSynthState(CCharEntity* PChar, SKILLTYPE skill)
+: CState(PChar, PChar->targid)
+, m_PEntity(PChar)
+{
+    if (m_errorMsg)
```
zach2good: Is it possible for this state to be created with errorMsg populated?

zach2good (commented):
Good idea 👍

src/map/packet_system.cpp
WinterSolstice8: Timings haven't really been dialed in, and this entire state would mean premature synth results would be ignored to begin with. Probably a future problem if someone wants to tackle it

src/map/ai/states/synth_state.cpp
WinterSolstice8: Not sure, actually

WinterSolstice8:
Looks like with packet jitter its 15/16 seconds. I presume its supposed to be 16 seconds because it's a magic number and almost certainly not a coincidence.

image

src/map/ai/states/synth_state.cpp
WinterSolstice8: Removed the error handling, I don't think it's possible to hit anyway.

WinterSolstice8:
Good to go, force D/C after starting a synth completes now, unlike before.

zach2good (approved):

PR #7022, opened 2025-02-12
[core] Shared treasure pool, basic bindings for managing treasure pools

sruon:
**_I affirm:_**

Introduces a new type of treasure pool, tentatively called TREASUREPOOL_SHARED.

A shared treasure pool is intended to be used for content taking place in **non-instanced** areas, that requires unrelated characters (i.e. not in same PT/alliance) to be part of the same treasure pool while also being able to maintain several distinct treasure pools in the zone (unlike a Zone Pool). 
This type of pool requires direct management of its lifecycle and membership by the creator.

This type of treasure pool enables the "extended" treasure view.

To the best of my knowledge, only Einherjar possess this set of constraints but I could be wrong.

Also adds basic bindings to manage a treasure pool in lua directly:
- CLuaTreasurePool::addMember
- CLuaTreasurePool::delMember
- CLuaTreasurePool::addTreasure
- CLuaTreasurePool::flush
- CLuaZone::createSharedPool
- CLuaZone::deleteSharedPool

With two players, in the same zone and in the same party, execute this script

```
-- func: testpool
-- desc: 
---@type TCommand
local commandObj = {}

commandObj.cmdprops =
{
    permission = 1,
    parameters = ''
}

commandObj.onTrigger = function(player)
    local zone = player:getZone()
    local secondPlayer = GetPlayerByName("Celistine")

    local sharedPool = zone:createSharedPool()
    if not sharedPool then
        player:printToPlayer("Failed to create shared pool.")
        return
    end

    -- Add the current player to the shared pool
    sharedPool:addMember(player)

    -- Only the current player should see the Angelstone
    sharedPool:addTreasure(xi.item.ANGELSTONE)

    -- Only the 2nd player should see the Deathstone
    secondPlayer:addTreasure(xi.item.DEATHSTONE)

    -- Add a couple r/e item to pool to ensure they dont autodrop
    sharedPool:addTreasure(xi.item.HOFUD)
    sharedPool:addTreasure(xi.item.HOFUD)

    -- Flush the shared pool before deleting it
    sharedPool:flush()

    -- Remove the current player from the shared pool
    sharedPool:delMember(player)

    -- Delete the pool
    zone:deleteSharedPool(sharedPool)

    -- Should reattach to the other player party pool automatically!
    -- Both players should see this Larimar
    secondPlayer:addTreasure(xi.item.LARIMAR)
end

return commandObj
```


src/map/zone.cpp
```diff
@@ -837,6 +846,58 @@ CBaseEntity* CZone::GetEntity(uint16 targid, uint8 filter)
...
 
+/************************************************************************
+ *                                                                       *
+ *  Create shared treasure pools local to the zone.                      *
+ *                                                                       *
+ ************************************************************************/
+
+CTreasurePool* CZone::CreateSharedTreasurePool()
+{
+    CTreasurePool* sharedPool = new CTreasurePool(TREASUREPOOL_SHARED);
```
zach2good: Can you make this unique_ptr instead of raw `new` please? This will remove the need to the later `destroy`s

zach2good (commented):
I'd probably want you to include your `testpool` command in this PR too, we might look at setting it up in CI

src/map/lua/lua_treasure_pool.cpp
```diff
@@ -0,0 +1,183 @@
+﻿/*
+===========================================================================
+
+  Copyright (c) 2010-2015 Darkstar Dev Teams
```
zach2good: LandSandBoat

src/map/lua/lua_treasure_pool.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+#include "entities/charentity.h"
+#include "entities/npcentity.h"
+#include "lua_baseentity.h"
+#include "packets/char_status.h"
+#include "treasure_pool.h"
+#include "utils/mobutils.h"
+#include "zone_entities.h"
+
+CLuaTreasurePool::CLuaTreasurePool(CTreasurePool* PTreasurePool)
```
zach2good: You'll need to add a section for this in `sol_bindings` too, so you can pass around TreasurePool and not wrap things by hand

src/map/lua/lua_treasure_pool.cpp
```diff
@@ -0,0 +1,183 @@
...
+    {
+        ShowWarning("CTreasurePool::addTreasure() - Invalid itemID.");
+        return 0;
+    }
+
+    if (m_pLuaTreasurePool != nullptr)
+    {
+        if ((arg1 != sol::lua_nil) && arg1.is<CLuaBaseEntity*>())
+        {
+            // The specified PEntity can be a Mob or NPC
```
zach2good, line 143: I don't know treasure pools well, why is a mob valid here?

src/map/zone.cpp
zach2good: This will probably be a pain in the ass, let me know if you need some advice

src/map/zone.h
```diff
@@ -562,6 +562,11 @@ class CZone
     uint16 GetBackgroundMusicDay() const;
     uint16 GetBackgroundMusicNight() const;
 
+    CTreasurePool* CreateSharedTreasurePool();
```
zach2good: Use auto & trailing return type here so it's all aligned

src/map/zone.h
```diff
@@ -678,7 +683,8 @@ class CZone
     void LoadZoneLines();
     void LoadZoneWeather();
 
-    CTreasurePool* m_TreasurePool;
+    CTreasurePool*              m_TreasurePool;
+    std::vector<CTreasurePool*> m_SharedTreasurePools;
```
zach2good: No raw ptr members, please

src/map/lua/lua_zone.h
```diff
@@ -44,6 +46,9 @@ class CLuaZone
     void setLocalVar(const char* key, uint32 value);
     void resetLocalVars();
 
+    auto createSharedPool() -> std::optional<CLuaTreasurePool>;
```
zach2good, line 49: How will these be intended to be used? On zone setup? Like when a group enters dynamis, etc?

src/map/zone.cpp
sruon: I actually wanted to but figured consistency was important. I'll read up on it tomorrow and make the changes.

zach2good: No need to be consistent if the existing code is old, ugly, and fragile 😆

src/map/lua/lua_treasure_pool.cpp
sruon: Depending on if it's a NPC or a monster that drops the item, a flag is set in the packet. This tells the client to use a different format for the message.

```
You find a page from Balrahn's Reflections on the Margygr.

You find an adaman ingot in the Armoury Crate.
```

For what it's worth, this is nearly a 1:1 copy of CLuaBaseEntity:addTreasure

src/map/lua/lua_zone.h
sruon: It will make a lot of sense when I push my Einherjar PR. 

Basically Hazhalm is a non-instanced area with 10 chambers, each with their own treasure pool. Each chamber accepts up to 36 members, which may or may not be part of the same alliances/parties.

As far as I know, every other content use either a zone pool (old Dynamis, probably Dynamis-D), or takes place in an instance which sidesteps the issue entirely.

The flow is:
- When someone reserves a chamber, a new pool is created
- When members enter the chamber, they are added to the pool
- When members exit the chamber, they are removed from the pool
- When the chamber is cleaned up, the pool is released

sruon:
- Replaced every reference to raw CTreasurePool pointers with std::shared_ptr
- Had to add a self reference in `CTreasurePool::DelMember`, otherwise it blows up mid-air. Is that considered acceptable or should the `CCharEntity->PTreasurePool` only be nulled outside of `DelMember` wherever it's called?
- Removed explicit deletes, except when Zone is trying to reset to a fresh pool (replaced by .reset())
- Tested every single scenario I could think of with success
  - Solo -> Logout/Zone
  - Solo -> Party -> Back to solo
  - Solo -> Party -> Alliance -> Back to party -> Back to solo
  - Reran my script

Tried using unique_ptr at first but I'm unclear how that would work since several entities are holding references to the smart pointer.
- Global zone pool are created by CZone but are assigned to each PChar zoning in
- Party pools are created from the initial leader and assigned to each subsequent player joining

If there's a nuance I'm missing, please let me know!

Still need to address other comments raised.

src/map/lua/lua_treasure_pool.cpp
sruon: Updated

src/map/lua/lua_treasure_pool.cpp
sruon: Updated

src/map/zone.h
sruon: Updated

src/map/zone.h
sruon: Updated

src/map/lua/lua_treasure_pool.cpp
sruon: Actually looks like the SOL_BIND_DEF macro expects the Lua class to take a raw pointer. Not sure how to proceed!

zach2good: When you pass things held by a unique_ptr into Lua, you can just use `.get()` to get a raw pointer to the underlying object. Lua isn't going to delete or swap any objects that it gets

zach2good: I think we do this with `TriggerArea`s

zach2good:
Unfortunately, `std::shared_ptr` doesn't make the lifetimes of objects it holds explicit. It is generally considered bad practice to use unless you're being given an external resource and making `std::weak_ptr`s out of it for lookup/locking. The exception to this is `ASIO` code, which by convention uses `std::shared_ptr`.

A `CCharEntity` doesn't _really_ exist without a zone, it certainly doesn't tick without one. `CZone` should be holding a `std::unique_ptr` to a shared treasure pool (if there is one), and the char should be looking up to the zone it's in for access to that treasure pool. Aside from convenience, the char doesn't need the treasure pool as a member. If you _really_ want Char to hold the treasure pool as a member, you can use a ref to it.

I think the methods you've added to `zone.h` for managing the pool should be sufficient.

Back to `shared_ptr` as a design choice: If you have the method `DeleteSharedTreasurePool`, but it's lifetime is managed by `shared_ptr`, just because you've removed it from `m_SharedTreasurePools` doesn't guarantee that the treasure pool is actually gone. It's lifetime could be artificially extended in the `shared_from_this()` that exists somewhere, or the chars that have had it assigned to them.

zach2good:
If you're feeling _extra spicy_ and wanted to work on your C++, you could look at these sol binding macros:
```cpp
#define SOL_BIND_DEC(LuaType, CppType) \
    int sol_lua_push(sol::types<CppType*>, lua_State* L, CppType* obj);

#define SOL_BIND_DEC_SUB(LuaType, BaseCppType, CppType) \
    int sol_lua_push(sol::types<CppType*>, lua_State* L, CppType* obj);

#define SOL_BIND_DEF(LuaType, CppType)                                                      \
    int sol_lua_push(sol::types<CppType*>, lua_State* L, CppType* obj)                      \
    {                                                                                       \
        return obj ? sol::stack::push<LuaType>(L, obj) : sol::stack::push(L, sol::lua_nil); \
    }

#define SOL_BIND_DEF_SUB(LuaType, BaseCppType, CppType)                                                   \
    int sol_lua_push(sol::types<CppType*>, lua_State* L, CppType* obj)                                    \
    {                                                                                                     \
        return obj ? sol::stack::push<LuaType>(L, (BaseCppType*)obj) : sol::stack::push(L, sol::lua_nil); \
    }
```

and use `std::enable_if` to make them accept `std::unique_ptr` as well as `CppType*`

zach2good:
I think it's worth thinking about the different usage scenarios here (please correct me if I'm wrong, I don't know this feature well):
- A zone can have no shared pools, multiple shared pools, or a single big shared pool that everyone in the zone uses. Can multiple of these conditions be true at once? If not, we could be holding the pools in a `std::variant`
- If a pool has no members after its initial creation, it needs to be destructed. Objects shouldn't be managing their own lifetimes, so something external needs to manage it. It sounds like the zone itself should be handling this. You have `AddToZonePool`, so a removal equivalent would work - where it then nulls/destructs the treasure pool if it's empty after removing someone.

zach2good:
Another question: Can these types of treasure pools be used in zones with instances in them? Because then the treasure pools will need to live somewhere else.

A non instanced zone is: 1x Zone (CZone) -> 1x Collection of Entities (CZoneEntities)
An instanced zone is: 1x Zone (CZoneInstance) -> Multiple Collections of Entities (CInstance)

So if there's a zone-wide treasure pool, every instance could see/use it - even if they're different instances (Everbloom Hollow, etc.)

Xaver-DaRed:
> If a pool has no members after its initial creation, it needs to be destructed. Objects shouldn't be managing their own lifetimes, so something external needs to manage it. It sounds like the zone itself should be handling this. You have AddToZonePool, so a removal equivalent would work - where it then nulls/destructs the treasure pool if it's empty after removing someone.

Shouldnt the ~~instance~~ battlefield itself be the one in charge of creating/requesting an available pool on creation and then destroying/freing it on instance cleanup?

sruon:
Forgetting about this PR for one second, this is the flow today

- CZone with MISC_TREASURE flag set
  - Zone wide pool created at CZone instantiation
  - Every char that zones in get added to the zone wide pool (AddMember), then the pointer is assigned to PChar->PTreasurePool
  - Every char that zones out gets removed from the pool (DelMember), then PChar->PTreasurePool is nulled

https://github.com/LandSandBoat/server/blob/2fc9da9f72091502d5b782f5049b5eabcb9d4135/src/map/zone.cpp#L426-L429
https://github.com/LandSandBoat/server/blob/2fc9da9f72091502d5b782f5049b5eabcb9d4135/src/map/zone.cpp#L1054-L1059

- CZone without MISC_TREASURE flag set
  - Zone wide pool remains null at instantiation
  - Every char that zones in gets either a new CTreasurePool if solo, or get reattached to the first party member (present in same zone) treasure pool through AddMember (defaults to a new solo pool if no party member in the zone). This pool is assigned to CCharEntity->PTreasurePool
  - Every char that zones out gets removed from their pool (DelMember), then PChar->PTreasurePool is nulled

https://github.com/LandSandBoat/server/blob/2fc9da9f72091502d5b782f5049b5eabcb9d4135/src/map/zone.cpp#L1062-L1070
https://github.com/LandSandBoat/server/blob/2fc9da9f72091502d5b782f5049b5eabcb9d4135/src/map/zone.cpp#L1179-L1182

To summarize, a CCharEntity is always holding a CTreasurePool*, but a CZone can also eventually hold the pointer as well if it's a zone wide pool.

> Another question: Can these types of treasure pools be used in zones with instances in them? Because then the treasure pools will need to live somewhere else.

Have not seeing any indication of instanced zones having zone wide pools, but I also don't see what would stop this from happening presently. It should probably ignore the flag in that case and log the issue.

> A zone can have no shared pools, multiple shared pools, or a single big shared pool that everyone in the zone uses. Can multiple of these conditions be true at once? If not, we could be holding the pools in a std::variant

There's no zone or content in the game (that I know of), that would require having both a zone wide pool and a shared pool, but again there's nothing stopping that from happening presently.

Trying to summarize what's been said: We should be getting rid of CCharEntity->PTreasurePool and rely on asking the CZone for the player current treasure pool. Then the CZone is the only entity holding the pointers to all the treasure pools, regardless of their kind. Am I on the right track?

zach2good:
Thanks for the in-depth explanation! This sheds a lot of light on why TreasurePool is the way it is, and why it does `delete this;`.

I think the best way to handle this is to have a zone hold both a single shared PTreasurePool and then also an array of treasure pools. I think the best way to think about it is that solo player treasure pools are parties of one, then we only have 2 modes.

For now, checking against `m_ZoneTreasurePool` and branching depending on what we find is fine, but if you wanted to try and use `std::variant` and `std::visit`, that could be fun exercise.

Anywhere PChar is asking itself about a treasure pool, it should just be asking the zone it's in to do things to the relevant treasure pool. Same with CParty. Though, it's a lot more complicated inside CParty. 

Thinking about it, I'd probably pull all of this logic out into a `TreasurePoolContainer` or something, since `TreasurePool` already has a lot of its own logic.

sruon:
Working on a separate PR to overhaul Treasure Pool management, so I'll close this one for the time being.

PR #7029, opened 2025-02-13
[core] Always use rank C meva for mobs, rank G for trusts

WinterSolstice8:
**_I affirm:_**

Apparently, meva was based off the evasion rank of the mob, which was wrong. Now we force calculate rank C for all mobs meva

Set trusts to rank G meva (I dont have any proof for or against this FYI)

Fix a mod not working for pets

lvl 99 wyvern:
image
matches level 76 rank c
image
makes level 99 rank g
image

### DIFF
--- a/src/map/utils/mobutils.cpp
+++ b/src/map/utils/mobutils.cpp
@@ -106,8 +106,16 @@ namespace mobutils
 
     uint16 GetMagicEvasion(CMobEntity* PMob)
     {
-        uint8 mEvaRank = PMob->evaRank;
-        return GetBaseSkill(PMob, mEvaRank);
+        uint8 mlvl = std::min<uint8>(PMob->GetMLevel(), 99);
+
+        // Assume trusts have G rank meva like players
+        if (PMob->objtype == TYPE_TRUST)
+        {
+            return battleutils::GetMaxSkill(12, mlvl);
+        }
+
+        // Mobs have rank C magic evasion
+        return battleutils::GetMaxSkill(7, mlvl);
     }
 
     /************************************************************************
--- a/src/map/utils/petutils.cpp
+++ b/src/map/utils/petutils.cpp
@@ -1111,8 +1111,8 @@ namespace petutils
 
     void FinalizePetStatistics(CBattleEntity* PMaster, CPetEntity* PPet)
     {
-        // set C magic evasion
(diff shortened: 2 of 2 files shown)


LacunaXI:
Unfortunately Magic Evasion for trusts varies. While this could (hopefully) just be a base rate where certain trusts like Yoran-Oran (UC) or to a lesser extent Ygnas have increased values. There is not any source I know of to substantiate this further.

WinterSolstice8:
> Unfortunately Magic Evasion for trusts varies. While this could (hopefully) just be a base rate where certain trusts like Yoran-Oran (UC) or to a lesser extent Ygnas have increased values. There is not any source I know of to substantiate this further.

Yeah I was wondering about that, some trusts do feel like they resist more through unknown means, and then there's the campaigns they do where trusts get tons of meva and hp/mp. This is purely a "I am about 90% certain C rank all of the time is wrong" but I couldn't find ANY info on research

I'm not dead-set on changing it to rank g, I just am not sure the average trust should be rank c

Frankie-hz:
We had made it based off the EVA rank in case we found an edge case where a mobs meva was based off the rank. We have yet to find any evidence of that so looks good to me.

LacunaXI:
> I'm not dead-set on changing it to rank g, I just am not sure the average trust should be rank c

I couldn't say one way or another. However I imagine a good test for this if you are interested would be to compare your character being debuffed to Yoran-Oran (UC) and some other trusts. Yoran uniquely comes to mind for their reputation of frequently evading sleepga, silencega and so on while Apururu (UC) or other trusts very rarely did. Ygnas is listed this way as well on [JP wiki ](https://wiki.ffo.jp/html/27930.html) to compare with other trusts if you have them. A target like Asida in Escha Ruaun is probably good for this as it casts AoE enfeebles.

WinterSolstice8:
That would be a good test, comparing one trust to another can reveal how much more meva they have. iirc it's 1 meva = -0.5% hit rate when hit rate is >50%?

WinterSolstice8:
Now that I think about it maybe a colibri could work because we could reflect AOE spells

Xaver-DaRed (approved):

LacunaXI:
> That would be a good test, comparing one trust to another can reveal how much more meva they have. iirc it's 1 meva = -0.5% hit rate when hit rate is >50%?

This is what BG says but I have never done magic evasion testing. Most of our understanding on magic evasion likely comes from JP blogs like this one: https://sazitouhuu.blogspot.com/2019/11/21.html or testing from Lute if I went digging.

PR #7061, opened 2025-02-17
[Fix] Sky cermet doors opening on click when they shouldnt

Xaver-DaRed:
**_I affirm:_**

Doors that arent supposed to open on click need either:
- An empty script with an empty onTrigger function.
- An entry in default actions.
- An entry in an IF script.

Go to sky. It still sucks.


KnowOne134 (approved):
3rd option is to add to an if page also

Xaver-DaRed:
I toyed a little more with this and found a way not to create the scripts. Claywar's work is safe^^

claywar (approved):

PR #7065, opened 2025-02-17
[lua] add module for pre TOAU signet

Grahf0085:
**_I affirm:_**

This PR remove the def and evasion bonus signet had added to the player since 2007. Also, player will heal for less and lose TP when healing with signet. There is a link at top of each lua file.  If you open link and search for "signet" you will see the changes.

Destined to be wildly popular

Add the following to init.txt
```
 era/lua/cop_signet.lua
era/lua/cop_signet_healing.lua
```

Get signet from gate guard
Attack monster that's an even match or higher in region under your nations control. Your def shouldn't change.
!tp 2000
!hp 1
/heal
Should loose tp when healing.

### DIFF
--- a/modules/era/lua/cop_signet.lua
+++ b/modules/era/lua/cop_signet.lua
@@ -1,0 +1,47 @@
+-----------------------------------
+-- Pre TOAU Signet effect
+-- Pre 8 March 2007
+-- https://www.bg-wiki.com/ffxi/The_History_of_Final_Fantasy_XI/2007
+-----------------------------------
+require('modules/module_utils')
+-----------------------------------
+local m = Module:new('cop_signet')
+
+m:addOverride('xi.effects.signet.onEffectGain', function(target, effect)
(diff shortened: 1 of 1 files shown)


zach2good (approved):
Seems reasonable, you just need to shake it until you can get it to pass CI

Frankie-hz:
Since these were added in the same patch, is there a reason they can’t be in that same module file called cop signet? If we start breaking out every little difference in things into all separate modules it’s going to get extremely bloated

Grahf0085:
> Since these were added in the same patch, is there a reason they can’t be in that same module file called cop signet? If we start breaking out every little difference in things into all separate modules it’s going to get extremely bloated

Good call. I tried that and it seemed to work. Moved everything to one file.

Grahf0085:
> Seems reasonable, you just need to shake it until you can get it to pass CI

Thanks for supporting somewhat silly things like this.... the really old stuff

Xaver-DaRed (approved):

PR #7066, opened 2025-02-17
[Bug] Bastok 7 1 mission QM2 in Ro'Maeve not moving

Eggowaffle1288:
**_I affirm:_**

The QM that spawns Mokkurkalfi was not moving around the zone every 15mins like it should

would always default to table pos 1.
issue with npcUtil.pickNewPosition, passing true to the function seems to not allow it to move?  setting it to false generates a new position.  
Since this should move every 15 mins real time (1hour = 2.4mins 15mins ish = 6 hours), it doesnt make sense to move the QM every time the ontrigger happens?  so i moved the logic to the Ro'Maeve zone file.

use gm command !addmission 1 12
talk to Cid
go to Ro'Maeve, QM will now move from spot to spot like it should every 15 mins real time


Eggowaffle1288 (commented):
dont think the 100 is needed
removed it

Eggowaffle1288:
resubmitting it properly.
this formatting and i dunno what happened is getting outa hand

PR #7090, opened 2025-02-19
[Combat] Implement defaultIfNil utility function

Xaver-DaRed:
**_I affirm:_**

Implements an utility function which will default an unknown value to a known one, if the original value is nil, and will also report the defaulting.

None.

### DIFF
--- a/scripts/globals/combat/skill_ranks.lua
+++ b/scripts/globals/combat/skill_ranks.lua
@@ -1,7 +1,8 @@
 -----------------------------------
 -- Table defining the skill level max value based on actor level and the skill rank.
------------------------------------
 -- Skill caps per rank: https://wiki-ffo-jp.translate.goog/html/2570.html?_x_tr_sl=ja&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=sc
+-----------------------------------
+require('scripts/globals/utils')
 -----------------------------------
 xi = xi or {}
 xi.combat = xi.combat or {}
@@ -115,9 +116,9 @@ xi.combat.skillLevel.dataTable =
 
 xi.combat.skillLevel.getSkillCap = function(actorLevel, skillRank)
     -- Sanitize fed values
-    local levelToCheck = actorLevel or 1             -- Assume level 1
-    local rankToCheck  = skillRank or xi.skillRank.G -- Assume rank G
+    local levelToCheck = utils.defaultIfNil(actorLevel, 0)             -- Assume level 0
+    local rankToCheck  = utils.defaultIfNil(skillRank, xi.skillRank.G) -- Assume rank G
 
     -- Going to assume levels over 99 give 1 skill level, just like master levels.
-    return xi.combat.skillLevel.dataTable[utils.clamp(levelToCheck, 0, 99)][rankToCheck] + utils.clamp(levelToCheck - 99, 0, 156)
+    return xi.combat.skillLevel.dataTable[utils.clamp(levelToCheck, 0, 99)][utils.clamp(rankToCheck, 1, 12)] + utils.clamp(levelToCheck - 99, 0, 156)
 end
--- a/scripts/globals/utils.lua
+++ b/scripts/globals/utils.lua
@@ -1234,3 +1234,14 @@ function utils.drawIn(target, table)
     target:setLocalVar('[Draw-In]WaitTime', 0)
     return false
 end
+
+function utils.defaultIfNil(inputValue, defaultValue)
+    if inputValue == nil then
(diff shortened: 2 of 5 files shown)


scripts/globals/combat/skill_ranks.lua
```diff
@@ -115,9 +115,9 @@ xi.combat.skillLevel.dataTable =
 
 xi.combat.skillLevel.getSkillCap = function(actorLevel, skillRank)
     -- Sanitize fed values
-    local levelToCheck = actorLevel or 1             -- Assume level 1
-    local rankToCheck  = skillRank or xi.skillRank.G -- Assume rank G
+    local levelToCheck = actorLevel >= 0 and actorLevel or 0           -- Assume level 0
+    local rankToCheck  = skillRank > 0 and skillRank or xi.skillRank.G -- Assume rank G
```
zach2good: These will produce errors if actorLevel or skillRank are passed in as nil. 

Some test code:
```lua
local actorLevel = nil
local skillRank = nil

-- Sanitize fed values
local levelToCheck = actorLevel >= 0 and actorLevel or 0           -- Assume level 0
local rankToCheck  = skillRank > 0 and skillRank or 0 -- Assume rank G

print(levelToCheck, rankToCheck)
```

```
lua5.4: Main.lua:5: attempt to compare number with nil
stack traceback:
	Main.lua:5: in main chunk
	[C]: in ?
```

They need to be in this order to be safe:
```lua
-- Sanitize fed values
local levelToCheck = actorLevel and actorLevel >= 0 or 0 -- Assume level 0
local rankToCheck  = skillRank and skillRank > 0 or 0 -- Assume rank G
```

Xaver-DaRed: Pushed changes and tested properly
imagen

zach2good: These change's aren't what I meant:
```
    local levelToCheck = (actorLevel and actorLevel > 0) and actorLevel or 0           -- Assume level 0
    local rankToCheck  = (skillRank and skillRank > 0) and skillRank or xi.skillRank.G -- Assume rank G
```

You're checking the same condition twice.

It's better to take the example I gave:
```
-- Sanitize fed values
local levelToCheck = actorLevel and actorLevel >= 0 or 0 -- Assume level 0
local rankToCheck  = skillRank and skillRank > 0 or 0 -- Assume rank G
```

This pattern of nil checking with a default option is used in a bunch of places and it relies on short-circuit evaluation to work:

```
local thing = maybeNil and maybeNil > 0 or 0
```

If `maybeNil` is nil the rest of the `and` statement won't be evaluated and the other side of the `or` is taken - the same with `maybeNil > 0`

zach2good: I'd avoid adding any ()'s if it isn't needed

Xaver-DaRed: Using this code
```lua
-- Sanitize fed values
local levelToCheck = actorLevel and actorLevel >= 0 or 0 -- Assume level 0
local rankToCheck  = skillRank and skillRank > 0 or 0 -- Assume rank G
```
Would populate both variables (levelToCheck and rankToCheck) with a `true` value (or a 0 if nil)
In c++ it would be the same as doing this:
`levelToCheck = actorLevel ? actorLevel >= 0 : 0`
`rankToCheck  = skillRank ? skillRank >= 0 : 0`
imagen
imagen

zach2good: You are 100% correct and I'm talking out of my ass. 

A helper like this in utils would make this sort of thing a lot more readable:
```lua
function defaultIfNil(inputValue, defaultValue)
    if inputValue == nil then
        local info = debug.getinfo(2, 'Sl')
        print(string.format("nil value encounted at %s:%i, defaulting to %i", info.source, info.currentline, defaultValue))
        return defaultValue
    end
    return inputValue
end

local val = nil

local checkedVal = defaultIfNil(val, 0)
```

```
function(input)
    -- validation
    input = utils.defaultIfNil(input, 0)
    input = utils.clamp(input, min, max)

    -- now we can do maths

    -- clamp and floor on the way out, if needed
end
```

scripts/globals/utils.lua
```diff
@@ -1234,3 +1234,14 @@ function utils.drawIn(target, table)
     target:setLocalVar('[Draw-In]WaitTime', 0)
     return false
 end
+
+function utils.defaultValue(inputValue, defaultValue)
```
zach2good: `valueIfNil` or `defaultIfNil` is better now that I see it, sorry for the runaround

zach2good (approved):

Issue #911, opened 2021-11-02
Battlefield timers are not updated by time extentions


TeoTwawki:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened
- [x] I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)
- [x] I have specified what branch this happens on branch: base

**_Additional Information_** (Steps to reproduce/Expected behavior) **:** 
Battlefields that show the countdown clock in the top left corner do not get updated if the time is modified (time extensions) in any way. Chat log messages do function correctly with the updated time however, so this is a missing cosmetic feature and not a bugged time remaining. It will reach zero, stop, and then you will still have however many minutes/seconds remain before being kicked out.

zach2good:
Good first issue: look inside battlefield.cpp at entry when we send out the timer packet. We should also be sending that out again on time updates. Notice that we send an empty timer packet to remove the timer (I think)

almuth150:
It also sends the full time value when entering after the battlefield has been initiated. So players who enter after the first one have their timers off by a little bit. If you wanted (or the battlefield was started while someone was afk and they entered 3 minutes late 😔), their timer would be 3 minutes off. It should send the current remaining time to subsequent entrants.

claywar:
While thinking about this, can anyone confirm that the timer starts while an event is rolling for the first player to enter?  I would assume yes

PR #7096, opened 2025-02-19
Send new timer packet when battlefield time limit is adjusted

eqbot:
**_I affirm:_**

Resolves issue #911 

1. Enter a BCNM (I used Treasures and Tribulations)
2. Use setbattlefieldtime GM command to edit the time limit on the battlefield
3. Countdown clock now shifts with changes to the battlefield time limit

### DIFF
--- a/scripts/commands/setbattlefieldtime.lua
+++ b/scripts/commands/setbattlefieldtime.lua
@@ -1,0 +1,29 @@
+-----------------------------------
+-- func: setbattlefieldtime
+-- desc: Set time limit for an active battlefield
+-----------------------------------
+---@type TCommand
+local commandObj = {}
+
+commandObj.cmdprops =
+{
+    permission = 4,
+    parameters = 'i'
+}
+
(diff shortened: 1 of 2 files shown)


scripts/commands/setbattlefieldtime.lua
```diff
@@ -0,0 +1,29 @@
...
+    parameters = 'i'
+}
+
+local function error(player, msg)
+    player:printToPlayer(msg)
+    player:printToPlayer('!setbattlefieldtime <minutes>')
+end
+
+commandObj.onTrigger = function(player, minutes)
+    local battlefield     = player:getBattlefield()
```
Xaver-DaRed: Could you remove the whitespace padding here, please?

eqbot: Certainly. I accidentally let it ride along from a different file.

WinterSolstice8 (approved):

PR #7097, opened 2025-02-19
Fix daily tally limit type

### LINKED ISSUE #7092: daily_tally is not tabulating correctly
**_I affirm:_**

Windows10 

## Branch affected by issue 

`base`

## Steps to reproduce

Enable Daily Tally with default values. Wait for JST midnight. Value will change to -15536. Next JST it will jump to 50000 (or presumibly the max points).

## Expected behavior

Tally should only increment by the number of points designated in main.lua or be set to max points if the value exceeds the max.

cocosolos:
Need to fit 50,000 into a signed int to compare to -1.

**_I affirm:_**

Fixes #7092

Default `DAILY_TALLY_LIMIT` is 50k, -1 is the uninitialized state before you unlock it. When comparing to -1, 50k is treated as a negative signed integer so the update query was producing incorrect results.

If `char_points.daily_tally` is negative, talk to one of the gobs and it should set it to 50, then either wait till JP midnight or what I did was add `dailytally::UpdateDailyTallyPoints();` in the hourly update [here](https://github.com/LandSandBoat/server/blob/f1a5d793dc0b504a54e3a4eed03f5db6646ea72e/src/world/time_server.cpp#L64). See daily tally correctly increase.

### DIFF
--- a/src/world/daily_tally.cpp
+++ b/src/world/daily_tally.cpp
@@ -30,8 +30,8 @@ namespace dailytally
 {
     void UpdateDailyTallyPoints()
     {
-        uint16 dailyTallyLimit  = settings::get<uint16>("main.DAILY_TALLY_LIMIT");
-        uint16 dailyTallyAmount = settings::get<uint16>("main.DAILY_TALLY_AMOUNT");
+        int32 dailyTallyLimit  = std::clamp<int32>(settings::get<int32>("main.DAILY_TALLY_LIMIT"), std::numeric_limits<uint16>::min(), std::numeric_limits<uint16>::max());
+        int32 dailyTallyAmount = std::clamp<int32>(settings::get<int32>("main.DAILY_TALLY_AMOUNT"), std::numeric_limits<uint16>::min(), std::numeric_limits<uint16>::max());
 
         if (!db::preparedStmt("UPDATE char_points "
                               "SET char_points.daily_tally = LEAST(?, char_points.daily_tally + ?) "


src/world/daily_tally.cpp
```diff
@@ -30,7 +30,7 @@ namespace dailytally
 {
     void UpdateDailyTallyPoints()
     {
-        uint16 dailyTallyLimit  = settings::get<uint16>("main.DAILY_TALLY_LIMIT");
+        uint32 dailyTallyLimit  = settings::get<uint16>("main.DAILY_TALLY_LIMIT");
```
WinterSolstice8: Shouldn't this be int32?

cocosolos: The limit doesn't or shouldn't actually support negative numbers, it just needs to be big enough to hold a uint16 in a 32 bit space. Either should work in this case so whatever is preferred.

WinterSolstice8: -1 on any uint should be 2^bitsize-1 

It would have the same problem logically, i'm guessing the compiler (or the connector?) does some funky stuff with these prepared statements.

I would promote both`dailyTallyLimit` and `dailyTallyAmount` to int32, use `settings::get<int32>` but clamp to 0: `std::max<int32>(0, settings::get<int32>("main.DAILY_TALLY_AMOUNT")));`

cocosolos: We don't want to actually exceed 16 bits though, since that's what we send in the packet. I could put it in a int32 and clamp it to uint16 numeric limits, but that's basically already what it's doing (settings::get casts it to a uint16). The issue is that any sufficiently large uint16 is interpreted as a negative int16. If we use a 32 bit setting then we would have the same problem when the 32 sign bit starts getting flipped (though packet issues would arise before that).

WinterSolstice8:
```
        const auto dailyTally = rset->get<int32>("daily_tally");
        ref<uint16>(0x5E)     = dailyTally == -1 ? 0 : static_cast<uint16>(dailyTally);
```
Looks like it's doing what I would expect because we use -1 tally as a default
db is stored as signed, and type is used as a downcasted uint16

cocosolos: I hope this is okay. I clamped it to uint16 range since that's the valid packet range, but use an int32 to query the db so all types should match up.

WinterSolstice8: Ooh, I love that. That's super cool and very obvious on what the intent is

WinterSolstice8 (approved):

src/world/daily_tally.cpp
cocosolos: I feel like there was a reason I used -1 instead of null, but I can't remember now. Maybe could make this all uint16 and just use null instead of -1, will have to look into that eventually.

WinterSolstice8: That would be ideal, but we'd need a migration most likely. Maybe it's doable?

cocosolos: It looks like `CLuaBaseEntity::getCurrency` and in turn `charutils::GetPoints` are what's ultimately preventing null from being used here. Only other thing I can think of would be using something else to indicate whether they've done the into or not like a charvar but then you have a bunch of forever vars which seems worse than the current implementation. This works now anyway so just gonna leave as is.

zach2good (approved):

PR #7101, opened 2025-02-20
[core] BLU Triple Attack overwrites Double Attack

sruon:
**_I affirm:_**

Setting 4 BLU spells with `Double Attack` currently grants both `Double Attack` and `Triple Attack`.

Only `Triple Attack` is supposed to be set in this condition. Double Attack can still be obtained independently from a WAR subjob.

> Setting only one tier will result in the [Double Attack](https://www.bg-wiki.com/ffxi/Double_Attack) Job Trait instead.
> https://www.bg-wiki.com/ffxi/Triple_Attack

> Setting two tiers will result in the [Triple Attack](https://www.bg-wiki.com/ffxi/Triple_Attack) Job Trait instead.
> https://www.bg-wiki.com/ffxi/Double_Attack

>     Set 8 trait points for double attack and 16 for triple attack.
>     Setting triple attack removes the double attack trait, but it can be regained through subbing warrior which provides a higher double attack rate than the spell trait does.
> https://www.bg-wiki.com/ffxi/Blue_Mage_Job_Traits#Double_Attack_&_Triple_Attack

> Rank1 can be acquired when the attribute value is 8 or above . Furthermore, when the attribute value reaches 16 or above, the attribute changes to Triple Attack
> https://wiki.ffo.jp/html/1904.html

This PR adds a special condition when the check is performed. 

As far as I can tell this is the only case where a trait overwrite a different one, hence the direct change to `blueutils::CalculateTraits`

There might be something smarter to be done as this could break if the following query ever returned the traits in a different order.

https://github.com/LandSandBoat/server/blob/f1a5d793dc0b504a54e3a4eed03f5db6646ea72e/src/map/trait.cpp#L89-L92

- Change to BLU 99 with a non WAR subjob
- Equip the following spells
image
- Verify only Triple Attack is present

- Unequip one of the spells
- Verify only Double Attack is present

- Repeat first test with a /WAR subjob and verify both Triple Attack and Double Attack are present

### DIFF
--- a/src/map/utils/blueutils.cpp
+++ b/src/map/utils/blueutils.cpp
@@ -484,9 +484,11 @@ namespace blueutils
                                         break;
                                     }
                                 }
-                                else if (PTrait->getMod() == Mod::DOUBLE_ATTACK && iter->getMod() == Mod::TRIPLE_ATTACK)
+                                else if ((PTrait->getMod() == Mod::DOUBLE_ATTACK && iter->getMod() == Mod::TRIPLE_ATTACK) ||
+                                         (PTrait->getMod() == Mod::GILFINDER && iter->getMod() == Mod::TREASURE_HUNTER))
                                 {
                                     // Triple Attack (16 pts) overwrites Double Attack (8 pts)
+                                    // Treasure Hunter (18 pts) overwrites Gilfinder (12 pts)
                                     add = false;
                                     break;
                                 }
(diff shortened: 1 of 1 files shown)


WinterSolstice8:
Supposedly gilfinder upgrades to TH:
https://www.bg-wiki.com/ffxi/Gilfinder

WinterSolstice8 (approved):

sruon:
Pushed another commit to handle TH/Gilfinder as well.

WinterSolstice8 (approved):

PR #7105, opened 2025-02-20
Core: Remove unused anticheat helpers

zach2good:
**_I affirm:_**

I was just poking around and noticed that this was written 5 years ago and never taken into action. All the information about jail cells etc. is available in the jail GM command


claywar (approved):

Xaver-DaRed (approved):
Deleted by zach2good

PR #7110, opened 2025-02-21
Only sleep zones after mobs have pathed home

### LINKED ISSUE #5910: 🔨 Add binding to force zones to stay away for arbitrary amount of time
**_I affirm:_**

## Describe the feature

Downstream request.

`zone:setKeepalive(utils.minutes(15))` or similar. 

Currently we keep zones alive for 5 seconds or so after the last player exits before it goes to sleep. This binding would take whatever mechanism that is and set it to be an arbitrary timestamp that defaults to be `"time the last player left" + 5 seconds`, where the `5 seconds` can be overridden temporarily. Once the zone goes to sleep, that time will be reset to the default and will need to be set again if you want a higher timer.

This is thought of as **TEMPORARY!**

In an ideal world the keep-alive-ness of zones will be managed exclusively from the world server, where it'll send "wake up" and "go to sleep" messages, but our RPC systems in both C++ and Lua land are hard to use and fragile.

Once they're enhanced or replaced with gRPC or something gRPC-like, with a layer of Lua bindings on top, then we can do away with these arbitrary timers.

DrDoakHead:
**_I affirm:_**

This is an attempt to fix #5910.  All zones with mobs stay alive until the mob fully paths back and/or restores its own hp/mp, then it sleeps that zone.

I have run this on my own client/server setup and examined print statements when I exit the zone.  I was able to determine we needed to add the "isDead" call to weed out mobs that are fished up, only spawn at night, or have other uncertain pop conditions.

### DIFF
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -937,6 +937,14 @@ void CZone::ZoneServer(time_point tick)
     {
         m_BattlefieldHandler->HandleBattlefields(tick);
     }
+
+    while (!zoneTimersAsleepAndRunning.empty())
+    {
+        std::string zoneName = zoneTimersAsleepAndRunning.front();
+        zoneTimersAsleepAndRunning.pop();
+
+        CTaskMgr::getInstance()->RemoveTask(zoneName + "_empty_timer");
+    }
 }
 void CZone::SleepZone()
 {
@@ -947,6 +955,8 @@ void CZone::SleepZone()
 
         ZoneTimerTriggerAreas->m_type = CTaskMgr::TASK_REMOVE;
         ZoneTimerTriggerAreas         = nullptr;
+
+        zoneTimersAsleepAndRunning.push(m_zoneName);
     }
 }
 
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -695,21 +695,20 @@ void CZone::UpdateWeather()
 void CZone::CheckMobsPathedBack()
 {
     // clang-format off
-    CTaskMgr::getInstance()->AddTask("zone_empty_timer", server_clock::now(), this, CTaskMgr::TASK_INTERVAL, 5s,
+    CTaskMgr::getInstance()->AddTask(m_zoneName + "_empty_timer", server_clock::now(), this, CTaskMgr::TASK_INTERVAL, 5s,
     [](time_point tick, CTaskMgr::CTask* PTask)
     {
         bool allMobsHomeAndHealed = true;
-        // if the timer runs out, check if all the mobs have pathed home
         CZone* PZone = std::any_cast<CZone*>(PTask->m_data);
-        if (PZone && PZone->GetZoneEntities())
+        // if the timer runs out, check if all the mobs have pathed home
+        if (PZone && PZone->GetZoneEntities() && PZone->GetZoneEntities()->GetMobList().size() > 0)
         {
             EntityList_t mobListMap = PZone->GetZoneEntities()->GetMobList();
             for (const auto& pair : mobListMap)
             {
                 CMobEntity* mob = dynamic_cast<CMobEntity*>(pair.second);
-
-                // if the mob is not fully healed OR ((the mob can rome home AND is far from home) OR the mob can't rome home)
-                if (mob && !mob->isFullyHealed() || ((mob->CanRoamHome() && mob->IsFarFromHome()) || !mob->CanRoamHome()))
+                // if the mob is (not dead/despawned AND it is not fully healed) OR it is pathing home
+                if (mob && ((!mob->isDead() && !mob->isFullyHealed()) || mob->m_IsPathingHome))
                 {
                     // at least one mob is away from home or not fully healed
                     allMobsHomeAndHealed = false;
@@ -718,7 +717,7 @@ void CZone::CheckMobsPathedBack()
             }
         }
 
-        // if all mobs home and healed up, sleep the zone @todo and stop timer?
+        // if all mobs home and healed up, sleep the zone
         if (allMobsHomeAndHealed)
         {
             PZone->SleepZone();
@@ -784,9 +783,6 @@ void CZone::IncreaseZoneCounter(CCharEntity* PChar)
     if (!ZoneTimer && !m_zoneEntities->CharListEmpty())
     {
         createZoneTimers();
-
-        // the zone is active again, remove the timer, if its still going
-        CTaskMgr::getInstance()->RemoveTask("zone_empty_timer");
     }
 
     PChar->StatusEffectContainer->DelStatusEffectsByFlag(EFFECTFLAG_ON_ZONE_PATHOS, true);
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -692,20 +692,39 @@ void CZone::UpdateWeather()
     // clang-format on
 }
 
-void CZone::SetKeepAlive()
+void CZone::CheckMobsPathedBack()
 {
-    // create timer to keep zone alive for 15 minutes
-    CTaskMgr::getInstance()->AddTask("zone_empty_timer", server_clock::now(), this, CTaskMgr::TASK_ONCE, 15min,
+    // clang-format off
+    CTaskMgr::getInstance()->AddTask("zone_empty_timer", server_clock::now(), this, CTaskMgr::TASK_INTERVAL, 5s,
     [](time_point tick, CTaskMgr::CTask* PTask)
     {
-        // if the timer runs out, sleep the zone
+        bool allMobsHome = true;
+        // if the timer runs out, check if all the mobs have pathed home
         CZone* PZone = std::any_cast<CZone*>(PTask->m_data);
-        if (PZone)
+        if (PZone && PZone->GetZoneEntities())
+        {
+            EntityList_t mobListMap = PZone->GetZoneEntities()->GetMobList();
+            for (const auto& pair : mobListMap)
+            {
+                CMobEntity* mob = dynamic_cast<CMobEntity*>(pair.second);
+                if (mob && mob->IsFarFromHome())
+                {
+                    // at least one mob is away from home
+                    allMobsHome = false;
+                    break;
+                }
+            }
+        }
+
+        // if all mobs home, sleep the zone, @todo and stop timer?
+        if (allMobsHome)
         {
             PZone->SleepZone();
         }
+
         return 0;
     });
+    // clang-format on
 }
 
 /************************************************************************
@@ -722,7 +741,7 @@ void CZone::DecreaseZoneCounter(CCharEntity* PChar)
 
     if (m_zoneEntities->CharListEmpty())
     {
-        SetKeepAlive();
+        CheckMobsPathedBack();
     }
     else
     {
--- a/src/map/zone.cpp
+++ b/src/map/zone.cpp
@@ -692,40 +692,26 @@ void CZone::UpdateWeather()
     // clang-format on
 }
 
-void CZone::CheckMobsPathedBack()
+bool CZone::CheckMobsPathedBack()
 {
-    // clang-format off
-    CTaskMgr::getInstance()->AddTask(m_zoneName + "_empty_timer", server_clock::now(), this, CTaskMgr::TASK_INTERVAL, 5s,
-    [](time_point tick, CTaskMgr::CTask* PTask)
+    bool allMobsHomeAndHealed = true;
+    if (m_zoneEntities && m_zoneEntities->GetMobList().size() > 0)
     {
-        bool allMobsHomeAndHealed = true;
-        CZone* PZone = std::any_cast<CZone*>(PTask->m_data);
-        // if the timer runs out, check if all the mobs have pathed home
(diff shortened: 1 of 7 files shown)


zach2good:
Hello and welcome 👋 

I'm the original author for https://github.com/LandSandBoat/server/issues/5910 and as I was typing up a response saying I didn't remember anything about it - a bunch of things came back to me. So here's my mind dump:

- Number one priority/warning/issue: In terms of basic things the server can do, keeping a zone alive is the most expensive thing possible. All the other expensive things I normally complain about like Navmesh access, SQL queries, etc. all rely on a zone actually being awake. Forcing a zone awake is forcing every mob in that zone to tick, and then the server pays the cost. If a player isn't in the zone, then you're potentially paying this heavy cost for no reason.
- The **abuse vector** I want to avoid here is if you're a small-time operator running the server on your old laptop or a raspberry pi etc. and the keepalive timer is set to 15 mins, a bad actor can just go to as many zones as possible (or get their friends to help) and jack up your server workload. We're not at a point where a single process can handle 100+ active zones, so server performance would start to degrade.
- Years ago we settled on 5s as the time limit to allow up to 2 logical ticks to fire off before the zone goes back to sleep after the last player leaves. This was to help with my confrontation system, which needed that buffer time to clean itself up.
- There is an actual retail use case here though, so the whole idea doesn't need to be thrown out: I think it would be harder to "simulate" zones that have events that happen in them when players aren't there than it would be to just spin up the zone and have it do things. When mobs travel through zones before besieged, when mobs travel for campaign, etc. it would be easier to just make the zone wake up and do the logic as if the player was there. This lends itself more to having a global ask the zone to wake up, and then put it back to sleep when it's done though, rather than keeping a zone awake for 15 mins when the last player leaves.

As a general question though, aside from trying to close out that ticket, is there something you're trying to solve here?

WinterSolstice8:
> As a general question though, aside from trying to close out that ticket, is there something you're trying to solve here?

There is a legitimate problem here, if you train mobs (NMs) to the zoneline and leave, on retail they will path back as if they moved normally. Right now if you do that on LSB and you're the only player in the zone you left they won't path back at all, which is incorrect. I think this is a legitimate change. In the future we may want to tick zones that don't have players in them recently at _all_ (such as during campaign, which will be event driven to activate zones in that case)

WinterSolstice8:
Thinking further, maybe what we could do is keep the zones alive until all mobs have pathed back to "home" and finished despawning? Most mobs despawn on zone so most of the time zones will remain active for seconds. I can't think of other cases at this exact moment however

DrDoakHead:
@zach2good This is my first go at fixing something on LSB.  I basically just wanted to fix your ticket, which in there you suggested, "zone:setKeepalive(utils.minutes(15)) or similar"

It makes sense that keeping the zone alive with no one in it would drastically increase the server load.  I come from safety critical land where you need to account for WCET type things, as I think even in horizon, there is rarely a zone empty.  With a few exceptions.

Anyways, again, I was taking your ticket as gold, and trying to do what you asked.  I didn't put much thought into the "why".

DrDoakHead:
@WinterSolstice8 if that is what we would like to solve, there would probably be a better way than just waiting and assuming after x amount of time that they pathed back?  I'm not familiar enough with the code, but if the mob class had some sort of flag signifying it was back "home", we could check those flags for all mobs at some sort of periodic rate, then when true, sleep the zone?

cocosolos:
Something to consider too is mobs that despawn from various conditions while idle. Consider King Vinegarroon. If you zone in during his weather and he pops, then you leave the zone, he will stay there until someone enters the zone again and his roam tick will happen and he'll despawn, setting his ToD respawn time to that point instead of when his weather actually ended and he was supposed to despawn. Or if he spawns and you leave and come back a few hours later after he should have despawned and reset his timer, if it's earth weather again he'll still be up.

DrDoakHead:
@zach2good and @WinterSolstice8 does my code look correct?  Will it do the intended functionality?

- I was a little confused on the singleton timer class.  Do I need to pass it "this"?
- I noticed the removeTask will print a warning if the item is not in the queue.
- Are there any things that stick out to you as wrong or just against your general coding guidelines?

jmcmorris:
I happened to see this and wanted to share my thoughts on how this should likely be fixed ultimately but not in this PR. I imagine updating a zone is only expensive due to updating monsters. Instead of putting the entire zone to sleep we should have monsters sleep if they are within their home range and no players in the zone. This could easily be a check at the start of the zone update. With this logic it'd allow monsters to roam back to home from a zone line and also handle logic like KV despawning. Perhaps we'd need a "no sleep" flag for mobs like this.

DrDoakHead:
I'm not sure I'm following.  This is my first PR so I am very new :)
**Current** - Zone sleeps 5 seconds after last person leaves
**This PR** - Zone sleeps 15 minutes after last person leaves
**Purposed** - Zone sleeps after last person leaves and mobs have pathed back to "home"

If "current" works, how would "purposed" break things like KV?

cocosolos:
> If "current" works, how would "purposed" break things like KV?

Sorry I wasn't really clear, KV currently behaves like that so there's nothing to break. I was just commenting on what a future solution should consider.

zach2good (approved):
I like this approach, and I'm glad we had the things available for it: `GetMobList()`, `IsFarFromHome()`, and `SleepZone()` 👍 

Outside of the scope of this PR: I'd probably want to see a new zoneflag (ZONEMISC, maybe, if we have room?) that enables/disables this logic.

Also, we're passing around copies of these entity lists, someone else can fix that in another PR.

WinterSolstice8:
Maybe we should also add in a check for mobs being fully healed too?

When all mobs path path it will call this which will heal them instantly. Normally this would be impossible if a mob is afflicted by a DoT, so if we wait out being healed to max hp/mp that call to `m_zoneEntities->HealAllMobs();` isn't necessary. Buff timers will expire on their own naturally (they store timestamps iirc) so we don't need to wait for those. Just HP/MP
```
void CZone::SleepZone()
{
    if (ZoneTimer && m_zoneEntities->CharListEmpty())
    {
        ZoneTimer->m_type = CTaskMgr::TASK_REMOVE;
        ZoneTimer         = nullptr;
        ZoneTimerTriggerAreas->m_type = CTaskMgr::TASK_REMOVE;
        ZoneTimerTriggerAreas         = nullptr;
        m_zoneEntities->HealAllMobs();
    }
}
```

DrDoakHead:
I was able to setup my client/server and verify this.  Only to the point that i did damage to a mob, zoned, waited 10sec or so, and re-entered and the mob was pathing back.  I can't verify from a black box perspective that the zone was slept.

DrDoakHead:
do i need to make another PR with the little checkbox checked or is this ok?

WinterSolstice8:
You can edit the checkbox back in on the post, as for verifying if it actually slept, you can just add a quick debug print temporarily `ShowInfo("Zone is going to sleep.")`

DrDoakHead:
I am remembering times in garliage citadel and yuhtunga jungle where someone would train a bunch of mobs to the entrance and zone, then the mobs would path back instead of disappearing and wipe xp parties.

That behavior was induced not by a zone being empty though.  This won't induce that right?

WinterSolstice8:
Nope, only NMs will path back. Normal mobs will despawn. I guess it could be enabled to work again and this code would help it work properly, which is good

src/map/zone.cpp
```diff
@@ -692,6 +692,43 @@ void CZone::UpdateWeather()
...
+        CZone* PZone = std::any_cast<CZone*>(PTask->m_data);
+        if (PZone && PZone->GetZoneEntities())
+        {
+            EntityList_t mobListMap = PZone->GetZoneEntities()->GetMobList();
+            for (const auto& pair : mobListMap)
+            {
+                CMobEntity* mob = dynamic_cast<CMobEntity*>(pair.second);
+
+                // if the mob is not fully healed OR ((the mob can rome home AND is far from home) OR the mob can't rome home)
+                if (mob && !mob->isFullyHealed() || ((mob->CanRoamHome() && mob->IsFarFromHome()) || !mob->CanRoamHome()))
```
WinterSolstice8: CI is complaining about this:
```suggestion
                if (mob && (!mob->isFullyHealed() || ((mob->CanRoamHome() && mob->IsFarFromHome()) || !mob->CanRoamHome())))
```

DrDoakHead: i see that ty.  also with print statements im seeing other issues.  looking into them.

cocosolos: > Nope, only NMs will path back. Normal mobs will despawn. I guess it could be enabled to work again and this code would help it work properly, which is good

I think this is already supported with a global setting and some mob mods. This logic looks pretty similar to [this in mob_controller](https://github.com/LandSandBoat/server/blob/ce0aad91b7961b9b3d1561110f2a3401803d1c69/src/map/ai/controllers/mob_controller.cpp#L972-L997). Maybe we could add a member var to mob entity, something like `bool m_idle`, and set it in mob_controller on the roam tick, then just check that here.

DrDoakHead: we want to check if the hp and mp has been restored.

which I am running into issues...for example, mobs that are fished up don't report hp==maxhp.  And some random mobs in the zone aren't reporting being fully healed.  Still looking into those.

WinterSolstice8: My guess: mobs that aren't spawned are mucking up your check. I bet things would work better if you filtered out the mobs that are not spawned

DrDoakHead: wow yea, that was it.  thank you!

cocosolos: > we want to check if the hp and mp has been restored.

The HP check is fine, it's this part I'm looking at:

```cpp
((mob->CanRoamHome() && mob->IsFarFromHome()) || !mob->CanRoamHome())
```
This is basically a duplicate check of this in mob_controller (my comments added):
```cpp
            // PMob->m_idle = true;
            if (!PMob->getMobMod(MOBMOD_DONT_ROAM_HOME) && PMob->IsFarFromHome())
            {
                if (PMob->CanRoamHome())
                {
                    // ...
                    // PMob->m_idle = false;
                }
                else if (!(PMob->getMobMod(MOBMOD_NO_DESPAWN) != 0) && !settings::get<bool>("map.MOB_NO_DESPAWN"))
                {
                    PMob->PAI->Despawn();
                    return;
                }
            }
```
If one of the conditions we're looking for to keep the zone alive is that they're roaming back home and we're already checking for that, we should leverage that instead of duplicating the check. That way you can just check:
```cpp
if (mob && !mob->isFullyHealed() || !mob->m_idle)
```
I'm not sure the exact order this stuff is happening, but it also looks like if this roam check in zone fires before the one in mob_controller, the Despawn call will be delayed until the zone wakes back up.

DrDoakHead: agree.  no need to duplicate code.  I'll change things up and use that.  Thank you!

DrDoakHead:
this is not right, do not merge this.  i need to remove the timer once the zone sleeps.
this is resulting in multiple timers being created because theres no restriction on duplicate named timers
also, if try to remove it, and its not there, i get a WARN log message.  Theres no elegant way to iterate through it because its a queue

DrDoakHead:
this should be good now.  I don't like the idea of checking a queue every "tick" but the check is in constant time.  As is adding to the queue and deleting from it, and accessing the front.

zach2good:
I haven't been paying attention to the conversation here, but I just had a look over the code. I'm happy with the idea that we won't put a zone to sleep until all mobs have roamed home and healed back to full, but I think this `queue` is overcomplicating things.

In the original:
```
if (ZoneTimer && m_zoneEntities->CharListEmpty() && m_timeZoneEmpty + 5s < server_clock::now())
{
    // Sleep zone, remove timer, etc.
```

This uses https://en.wikipedia.org/wiki/Short-circuit_evaluation since all of these conditions are `&&`'d together : it'll check `ZoneTimer` and if it's false it'll bail out, etc.

So you can just add your new condition:
```
if (ZoneTimer && m_zoneEntities->CharListEmpty() && m_timeZoneEmpty + 5s < server_clock::now() && CheckMobsPathedBack())
{
    // Sleep zone, remove timer, etc.
```

zach2good:
And in this layout I suggested, `void CZone::CheckMobsPathedBack()` becomes a singular check like:
```cpp
bool CZone::CheckMobsPathedBack()
{
    for each mob in zone:
        check theyr'e pathed back and healed
```

DrDoakHead:
updated per suggestions

DrDoakHead:
also, is there a way i can do these CI checks locally?

claywar:
> also, is there a way i can do these CI checks locally?

All of the scripts are in tools/ci.  Just run them from the base directory of your repo (check build.yml for dependencies)

src/map/zone.h
```diff
@@ -583,6 +583,8 @@ class CZone
     bool CanUseMisc(uint16 misc) const;
     void SetWeather(WEATHER weatherCondition);
     void UpdateWeather();
+    bool CheckMobsPathedBack();
+    void SleepZone();
```
zach2good: This declaration for SleepZone is no longer needed

zach2good (approved):
Almost there 👍

src/map/zone.h
DrDoakHead: Great catch.  Updated

PR #7119, opened 2025-02-22
[lua][Quest] Her Majestys Garden to IF

KnowOne134:
**_I affirm:_**

moves quest to IF system
no logic changes were needed

quest can start and finish


KnowOne134:
noticed Medicine Wonman was already done so popped it in as converted as well

zach2good (approved):
Doing great work, thanks!

PR #7130, opened 2025-02-23
[Refactor] Standardize "damage taken" modifier usage

Xaver-DaRed:
**_I affirm:_**

This is current usage:
- Physical SDT modifiers: Defined as a float in database, fetched as an uint, stored in a base 1000 modifier where base 100% damage = value of 1000
- Elemental SDT modifiers: Defined as an int in database, fetched as an int, stored in a base 10000 modifier where base 100% damage = value of 0 (sane) BUT 50% less damage = 5000 and 50% more damage = -5000
- Other SDT modifiers: Defined as an int in database, fetched as an int, stored in a base 10000 modifier where base 100% damage = value of 0, 50% less damage = -5000 and 50% more damage = 5000

This is new usage:
- All SDT modifiers are defined as an int in database, fetched as an int, stored in a base 10000 modifier where base 100% damage = value of 0, 50% less damage = -5000 and 50% more damage = 5000

Fight mobs. Cast spells.


scripts/effects/tomahawk.lua
```diff
@@ -9,14 +9,13 @@ effectObject.onEffectGain = function(target, effect)
...
+        effect:addMod(physSDT[i], -reductionPhys)
     end
 
     for element = xi.element.FIRE, xi.element.DARK do
-        local elementSDTMod  = xi.combat.element.getElementalSDTModifier(element)
-        local sdtModMagic    = target:getMod(elementSDTMod)
+        local sdtModMagic    = target:getMod(xi.combat.element.getElementalSDTModifier(element))
         local reductionMagic = sdtModMagic * 0.25
 
         effect:addMod(elementSDTMod, -reductionMagic)
```
UmeboshiXI: Orphaned Variable.

UmeboshiXI:
Just thought of some not functional but for documentation. The mod enum has comments relating to the old physical SDTs not being the same base values as the others.

zach2good:
I don't know if this ever came up in the past, but should we rename these mods to imply that they're `1/10000` `1e4-based`, etc.

`SLASH_SDT` -> `SLASH_SDT_BASE_10K`? Just really hammer it home forever? I wouldn't be against us marking other things with types too: `..._FLOAT`, `_PERCENT`, `_something that indicates 0.0 to 1.0`, etc.

Xaver-DaRed:
I wouldnt be opposed to changing the modifier naming conventions. Or implementing a convention in the first place.

With this change, now all `_SDT` mods use 10k as base and all function the same (positive -> more damage, negative .> less damage)
This was more something that was bound to happen. We started this a year or 2 ago and we just didnt finish it.

zach2good:
I think because this has been an issue for so long, and it's just a "thing you have to know" once you're up in Lua - and nobody is going to go read the comments next to where these mods/enums are defined: we should add things like `_BASE_10K`, `_PERCENT`, etc. to things that are of an arbitrary scale. We _don't_ need to add this sort of annotation for things that line up in-game like DEF, etc.

zach2good (approved):

zach2good:
We talked about this offline - the naming convention changes can happen in a later PR 👍

PR #7132, opened 2025-02-24
[lua][Quest] Setting Sun to if

KnowOne134:
**_I affirm:_**

Moves the Quest: Setting Sun to IF system
no logic changes made.

quest starts and completes

!addquest 0 BLACKMAIL, !completequest 0 BLACKMAIL
!pos -137.070 10.999 161.855 231 and trigger npc to start quest outlined here:
https://ffxiclopedia.fandom.com/wiki/The_Setting_Sun


scripts/quests/sandoria/The_Setting_Sun.lua
```diff
@@ -0,0 +1,80 @@
...
+                        player:needToZone(true)
+                    end
+                end,
+            },
+        },
+    },
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_COMPLETED and
+                not player:needToZone()
```
Xaver-DaRed: This suggests that you must have zoned for that event to play.
However, this video shows otherwise: https://www.youtube.com/watch?v=ygF_INPGERQ

KnowOne134: your right i added the wrong direction for that cs, corrected

Xaver-DaRed (approved):

PR #7133, opened 2025-02-24
[lua][Quest] Starting a Flame to IF

KnowOne134:
**_I affirm:_**

moves quest to IF
no logic changes to quest

quest starts and completes still
30 fame default on 1st
5 fame after


scripts/quests/sandoria/Starting_a_Flame.lua
```diff
@@ -0,0 +1,73 @@
...
+                [36] = function(player, csid, option, npc)
+                    if player:getQuestStatus(xi.questLog.SANDORIA, xi.quest.id.sandoria.STARTING_A_FLAME) == xi.questStatus.QUEST_ACCEPTED then
+                        quest:complete(player)
+                    else
+                        player:addFame(xi.fameArea.SANDORIA, 5)
+                    end
+
+                    npcUtil.giveCurrency(player, 'gil', 100)
+                    player:confirmTrade()
+                end,
```
Xaver-DaRed, line 67: Why not complete the quest each time no matter status and have the rewards be in the reward section?
I dont think the 5 fame points are notizable to the point of us knowing if first completion grants them or not, no?

Xaver-DaRed: A simplification nitpick, pretty much

KnowOne134: the 1st time you complete the quest the npcUtil.completeQuest which the quest frame work triggers the logic through gives default 30 fame for area you are in if nil, so 1st complete is 30 fame, all others are 5

Xaver-DaRed (approved):

PR #7139, opened 2025-02-25
[lua, sql] Fix Imp horn breaking

Aeshur:
**_I affirm:_**

Changes Imp family mixin and their TP abilities to function off of the updated mob sub animation IDs. In a PR awhile back the majority of Imps were changed to use 4 as their default and thus caused these scripts to no longer function. Additionally I changed the remaining Imps that had it still set from 0 to 4.

`!gotoid 17100840`
`!addeffect mighty_strikes` (So you crit and break horn)
`!tp 3000` to the  target imp
See that imp's horn breaks and their TP ability usage changes as intended.


zach2good (approved):

Xaver-DaRed (approved):

jmcmorris:
Why did the imps default animsub get changed in the first place? That seems wrong and the real fix here would be to revert the default animsubs back to 0.

I just looked and am not seeing the mentioned imp animsub change. Regardless, I think that all imp animsub should default to 0. This should fix the issue and only require changes in mob_pools.sql rather than many files.

Aeshur:
> Why did the imps default animsub get changed in the first place? That seems wrong and the real fix here would be to revert the default animsubs back to 0.
> 
> I just looked and am not seeing the mentioned imp animsub change. Regardless, I think that all imp animsub should default to 0. This should fix the issue and only require changes in mob_pools.sql rather than many files.

https://github.com/LandSandBoat/server/discussions/7126

Personally I agree with you, but I just want these mobs to function correctly for my server and am planning on doing the other ones as well (Lamia, Qutrub, Marid, etc). They are also borked (e.g. Lamia defaulted to 12 rather than 0 now, so they dont use their weapon TP moves and their weapons dont break).

UmeboshiXI:
It seems like most of the normal non NM imps were using 4 as a default animationSub. That seems to have not been touched in years, same with Lamia, etc.

For Lamia and the others (Imp has its own mixin)
I think it was caused by this PR: https://github.com/LandSandBoat/server/pull/5992/

Before it just checked: if an animationSub was not '1'.

Now it is explicit which means all these cases where the mob's animation sub is not 0 by default don't function like they should.

image

Edit: Clarity

Xaver-DaRed:
This needs to be rebased so conflicts are resolved and is able to be merged.

Everything else is good.

PR #7144, opened 2025-02-27
[lua][command][quest] all in 1 quest command for IF

KnowOne134:
**_I affirm:_**

all in 1 command for quests. With menu!
for IF, reads all vars in the format 'Prog' 'Stage' 'Option' 'Wait' upon use for the quest.
can clears all the vars listed for quest
adds, deletes, completes quests with 1 button, no need to add then complete to go from accepted to complete anymore, same with a completed quest to avail.

image

image

image

none.

### DIFF
--- a/scripts/commands/quest.lua
+++ b/scripts/commands/quest.lua
@@ -1,0 +1,154 @@
+-----------------------------------
+-- func: quest <logid> <questid> optional <player-name>, logid and quest ids work by numbers or string.format
+-- desc: quest command built with menu control, can add/delete/complete quests, see status,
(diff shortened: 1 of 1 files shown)


scripts/commands/quest.lua
```diff
@@ -0,0 +1,152 @@
+-----------------------------------
+-- func: addallspells
```
CriticalXI: Looks like the info at the top of the file needs to be updated for this command

KnowOne134: ahh, true. my secret. i like to take a page edit what i need instead of create all from scratch

zach2good (approved):

PR #7148, opened 2025-02-27
Scaffolding for onTransportEvent to trigger NPCs and mobs.

clanofartisans:
**_I affirm:_**

Adds some scaffolding to allow `onTransportEvent` / `TransportDepart` to trigger NPCs and mobs, in addition to players. I'm hoping to use this to work with mobs on the ferry, manaclipper, and barge, as well as NPCs on the barge. Could potentially be used to bootstrap the Discerning Eye quests.

Board all the boats on all routes and make sure nothing blows up.

### DIFF
--- a/scripts/zones/Aht_Urhgan_Whitegate/Zone.lua
+++ b/scripts/zones/Aht_Urhgan_Whitegate/Zone.lua
@@ -82,10 +82,12 @@ zoneObject.onTriggerAreaLeave = function(player, triggerArea)
 end
 
 zoneObject.onTransportEvent = function(player, transport)
-    if transport == 46 or transport == 47 then
-        player:startEvent(200)
-    elseif transport == 58 or transport == 59 then
-        player:startEvent(203)
+    if player:getObjType() == xi.objType.PC then
+        if transport == 46 or transport == 47 then
+            player:startEvent(200)
+        elseif transport == 58 or transport == 59 then
+            player:startEvent(203)
+        end
     end
 end
 
--- a/scripts/zones/Bastok-Jeuno_Airship/Zone.lua
+++ b/scripts/zones/Bastok-Jeuno_Airship/Zone.lua
@@ -17,8 +17,11 @@ zoneObject.onZoneIn = function(player, prevZone)
     return cs
 end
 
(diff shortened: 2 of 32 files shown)


clanofartisans:
I haven't tested this thoroughly yet but wanted to get a vibe check from the LSB team to see if this is something that might be accepted, hence the Draft PR.

zach2good:
I can't sign off on adding plumbing without seeing how you intend to use it. My kneejerk reaction is to say no to repurposing this pathway that is pretty much exclusively for triggering player CS events into something that has other unrelated logic.

There might be other ways to achieve what you want without having to jam more things into this binding, can you explain what you're trying to do.

Looking at this quest:
https://ffxiclopedia.fandom.com/wiki/A_Discerning_Eye

Do we not have a general "the transport has started" single call binding, where you can then set up the NPCs?

clanofartisans:
> I can't sign off on adding plumbing without seeing how you intend to use it. My kneejerk reaction is to say no to repurposing this pathway that is pretty much exclusively for triggering player CS events into something that has other unrelated logic.
> 
> There might be other ways to achieve what you want without having to jam more things into this binding, can you explain what you're trying to do.
> 
> Looking at this quest: https://ffxiclopedia.fandom.com/wiki/A_Discerning_Eye
> 
> Do we not have a general "the transport has started" single call binding, where you can then set up the NPCs?

No problem! And no, I've looked high and low and this:
https://github.com/LandSandBoat/server/blob/ceaf9ab7235409e4a730bf1d825e46ec83c7dba5/src/map/zone_entities.cpp#L394 
Seems to be the only place transport events interact with Lua at the moment. I'm happy to rework it some other way if you think there's a better spot for it! :)

CriticalXI:
Thanks Hugin for starting this because I was also going to be sending something similar myself. I also needed a lua binding that triggers for transports that wasn't reliant on players using the transport, as currently a player is required to be zoning onto a transport to get anything triggering in lua at the moment. Needed for despawning a NM that spawns on boats, like Zoredonite.

clanofartisans:
I have a different idea on how to accomplish this better but I'll just open a new PR for it at this point. Thanks!

PR #7152, opened 2025-02-28
[core, lua] Treasure Pool rework

sruon:
**_I affirm:_**

- Attaches a CTreasurePoolContainer to CZone, whose job is to manage treasure pools lifecycle.
- Provides a CCharEntity::GetTreasurePool method to ask CZone for the current appropriate treasure pool
- Introduces a Management type for CTreasurePool, indicating if the treasure pool should be managed directly by player/party/alliance or not.
- Moves CTreasurePool ticks processing from zone_entities to the CTreasurePoolContainer
- Introduces a Shared TREASUREPOOLTYPE, set to 36 slots.
- Removes direct PTreasurePool pointer manipulations from CCharEntity/CParty/CAlliance
- Creates an (almost complete) set of Lua bindings to manage treasure pools
- Adds a !getpool debug command, listing target treasure pool members/loot (prints all pools in zone if no target)
- (**Behavior change**) Reworks how Seals are added to pool, to support party-scoped cooldown on seals. On retail, an alliance of 3 parties is 3 distinct sets of cooldown, not a single global one.

Still working through testing everything, especially alliance. Will provide full list of tests when able.

### DIFF
--- a/scripts/enum/treasure_pool.lua
+++ b/scripts/enum/treasure_pool.lua
@@ -7,7 +7,6 @@ xi = xi or {}
 xi.treasurePool =
 {
     -- Keep synced with TREASUREPOOLTYPE
-
     SOLO                     =   1,
     PARTY                    =   6,
     ALLIANCE                 =  18,
--- a/scripts/enum/treasure_pool.lua
+++ b/scripts/enum/treasure_pool.lua
@@ -1,0 +1,16 @@
+-----------------------------------
+-- Treasure Pool
+-----------------------------------
+xi = xi or {}
+
+---@enum xi.treasurePool
+xi.treasurePool =
+{
+    -- Keep synced with TREASUREPOOLTYPE
+
+    SOLO                     =   1,
+    PARTY                    =   6,
+    ALLIANCE                 =  18,
+    SHARED                   =  36,
+    ZONE                     =  128,
+}
--- a/src/map/alliance.cpp
+++ b/src/map/alliance.cpp
@@ -187,34 +187,20 @@ void CAlliance::delParty(CParty* party)
         return;
     }
 
-    bool releaseTreasurePool = false;
-    if (this->getMainParty() == party)
-    {
-        releaseTreasurePool = true;
-    }
-
     // Delete the party from the alliance list
     auto partyToDelete = std::find(party->m_PAlliance->partyList.begin(), party->m_PAlliance->partyList.end(), party);
 
     if (partyToDelete != party->m_PAlliance->partyList.end())
     {
-        std::vector<std::pair<CZone*, CTreasurePool&>> PTreasurePools;
-
         for (const auto& PMember : party->members)
         {
-            const auto PChar = static_cast<CCharEntity*>(PMember);
-            if (auto& PTreasurePool = PChar->GetTreasurePool(); PTreasurePool.IsManaged())
+            if (const auto PChar = dynamic_cast<CCharEntity*>(PMember))
             {
-                PTreasurePools.emplace_back(zoneutils::GetZone(PChar->getZone()), PTreasurePool);
-                PTreasurePool.DelMember(PChar);
+                if (auto& PTreasurePool = PChar->getTreasurePool(); PTreasurePool.isManaged())
+                {
+                    PTreasurePool.delMember(PChar);
+                }
             }
-        }
-
-        // Force release of treasure pools, otherwise the leader would get reattached to the same pool
-        for (const auto& poolToBeReleased : PTreasurePools)
-        {
-            auto [PZone, PTreasurePool] = poolToBeReleased;
-            PZone->ReleaseTreasurePool(PTreasurePool);
         }
 
         party->m_PAlliance->partyList.erase(partyToDelete);
@@ -225,15 +211,26 @@ void CAlliance::delParty(CParty* party)
         entry->ReloadParty();
     }
 
+    if (party == getMainParty())
+    {
+        auto* PLeaderEntity = getMainParty()->GetLeader();
+        if (auto* PLeader = dynamic_cast<CCharEntity*>(PLeaderEntity))
+        {
+            zoneutils::GetZone(PLeader->getZone())->releaseTreasurePool(PLeader->getTreasurePool());
+        }
+    }
+
     party->m_PAlliance = nullptr;
 
     // Reattach party members to a new pool
     for (const auto& PMember : party->members)
     {
-        const auto PChar = static_cast<CCharEntity*>(PMember);
-        if (auto& PTreasurePool = PChar->GetTreasurePool(); PTreasurePool.IsManaged())
+        if (const auto PChar = dynamic_cast<CCharEntity*>(PMember))
         {
-            PTreasurePool.AddMember(PChar);
+            if (auto& PTreasurePool = PChar->getTreasurePool(); PTreasurePool.isManaged())
+            {
+                PTreasurePool.addMember(PChar);
+            }
         }
     }
 
@@ -256,9 +253,9 @@ void CAlliance::addParty(CParty* party)
 
     for (auto* PChar : party->members)
     {
-        if (auto& PTreasurePool = static_cast<CCharEntity*>(PChar)->GetTreasurePool(); PTreasurePool.IsManaged())
+        if (auto& PTreasurePool = static_cast<CCharEntity*>(PChar)->getTreasurePool(); PTreasurePool.isManaged())
         {
-            PTreasurePool.DelMember(static_cast<CCharEntity*>(PChar));
+            PTreasurePool.delMember(static_cast<CCharEntity*>(PChar));
         }
     }
 
@@ -285,9 +282,9 @@ void CAlliance::addParty(CParty* party)
     for (std::size_t i = 0; i < party->members.size(); ++i)
     {
         auto* PChar = static_cast<CCharEntity*>(party->members.at(i));
-        if (auto& PTreasurePool = PChar->GetTreasurePool(); PTreasurePool.IsManaged())
+        if (auto& PTreasurePool = PChar->getTreasurePool(); PTreasurePool.isManaged())
         {
-            PTreasurePool.AddMember(PChar);
+            PTreasurePool.addMember(PChar);
         }
         charutils::SaveCharStats(PChar);
         PChar->m_charHistory.joinedAlliances++;
@@ -367,18 +364,53 @@ void CAlliance::assignAllianceLeader(const std::string& name)
 
         m_AllianceID = charid;
 
+        const auto PPrevLeaderEntity = aLeader->GetLeader();
+
         // in case leader's on another server
         this->aLeader = nullptr;
 
         for (auto* PParty : partyList)
         {
-            if (PParty->GetMemberByName(name))
+            if (auto PNewLeaderEntity = PParty->GetMemberByName(name))
             {
                 this->aLeader = PParty;
+                onLeaderChange(PPrevLeaderEntity, PNewLeaderEntity);
                 break;
             }
         }
 
         db::preparedStmt("UPDATE accounts_parties SET partyflag = partyflag | ? WHERE charid = ?", ALLIANCE_LEADER, charid);
+    }
+}
+
+// Notify zone of leader changes for treasure pool, for each zone where members are.
+void CAlliance::onLeaderChange(CBattleEntity* PPrevLeaderEntity, CBattleEntity* PNewLeaderEntity) const
+{
+    if (!PPrevLeaderEntity || !PNewLeaderEntity)
+    {
+        return;
+    }
+
+    std::vector<CBattleEntity*> poolMembers;
+    for (const auto& party : this->partyList)
+    {
+        for (auto& member : party->members)
+        {
+            poolMembers.push_back(member);
+        }
+    }
+
+    for (auto* PMemberEntity : poolMembers)
+    {
+        if (const auto* PMember = dynamic_cast<CCharEntity*>(PMemberEntity))
+        {
+            if (const auto PPrevLeader = dynamic_cast<CCharEntity*>(PPrevLeaderEntity))
+            {
+                if (const auto PNewLeader = dynamic_cast<CCharEntity*>(PNewLeaderEntity))
+                {
+                    zoneutils::GetZone(PMember->getZone())->reassignTreasurePool(PPrevLeader, PNewLeader);
+                }
+            }
+        }
     }
 }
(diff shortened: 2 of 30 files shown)


src/map/zone.h
```diff
@@ -678,7 +686,7 @@ class CZone
     void LoadZoneLines();
     void LoadZoneWeather();
 
-    CTreasurePool* m_TreasurePool;
+    std::unique_ptr<CTreasurePoolContainer> m_TreasurePoolContainer;
```
zach2good, line 689: If a treasure pool is just a wrapper around 2x unordered_maps and it is simple to initialize itself - there isn't a need for it to be a unique_ptr, it can just be a regular member of this class

src/map/zone.h
```diff
@@ -573,6 +575,12 @@ class CZone
     void   SetLocalVar(const char* var, uint32 val);
     void   ResetLocalVars();
 
+    auto CreateTreasurePool(uint32 poolOwnerId, TREASUREPOOLTYPE poolType) const -> CTreasurePool&;
+    auto GetTreasurePool(CCharEntity* PChar) const -> CTreasurePool&;
+    auto GetTreasurePools() const -> std::unordered_map<uint32, CTreasurePool>&;
+    void ReleaseTreasurePool(CTreasurePool& PTreasurePool) const;
+    auto ReassignTreasurePool(CCharEntity* PPrev, CCharEntity* PNew) -> bool;
+
```
zach2good: We don't have this written down anywhere, but we're trying to move towards the convention of member functions (even bound ones) being `lowerCamelCase` and free functions (including in namespaces, and bindings) can be `UpperCamelCase`

src/map/zone.cpp
```diff
@@ -101,6 +101,8 @@ CZone::CZone(ZONEID ZoneID, REGION_TYPE RegionID, CONTINENT_TYPE ContinentID, ui
     // settings should load first
     LoadZoneSettings();
 
+    m_TreasurePoolContainer = std::make_unique<CTreasurePoolContainer>(m_miscMask & MISC_TREASURE);
```
zach2good, line 104: When I recommended that you make this not a unique_ptr, I didn't see the reading of `m_miscMask` - and now that I'm looking at it: where/when is `m_miscMask` set? Presumably inside `LoadZoneSettings`?

So, no need to change from unique_ptr, this is an item of work for me at some point to pull out all the db information  _before_ we start constructing things

src/map/treasure_pool_container.h
```diff
@@ -0,0 +1,56 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see http://www.gnu.org/licenses/
+
+===========================================================================
+*/
+
+#ifndef _CTREASUREPOOLCONTAINER_H
```
zach2good: #pragma once - please

src/map/treasure_pool_container.cpp
```diff
@@ -0,0 +1,225 @@
...
+auto CTreasurePoolContainer::OnZoneTick(const time_point tick) -> void
+{
+    for (auto it = m_Pools.begin(); it != m_Pools.end();)
+    {
+        it->second.CheckItems(tick);
+
+        if (it->second.MemberCount() < 1)
+        {
+            ShowDebugFmt("Removing empty pool formerly owned by ID {}", it->first);
+            it = m_Pools.erase(it);
```
zach2good: I tend not to like this pattern of `I am going to iterate with raw iterators and remove things as I go`, can you change this to do similar to what we do in `ZoneEntities`: Where we collect the things we want to remove into a temporary container as we do the first pass, and then remove all the container things. In this case, then you'd flush afterwards.

src/map/treasure_pool_container.cpp
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    return false;
+}
+
+// Force releases a treasure pool from the container.
+// Used when dissolving a party or alliance.
+auto CTreasurePoolContainer::ReleaseTreasurePool(CTreasurePool& PTreasurePool) -> void
+{
+    for (auto it = m_Pools.begin(); it != m_Pools.end();)
```
zach2good: Same here - as a general rule this isn't the hot-path so we don't need to worry about the mega/micro performance benefits of this type of iteration

src/map/treasure_pool_container.cpp
```diff
@@ -0,0 +1,225 @@
...
+// Used when party or alliance leader changes.
+auto CTreasurePoolContainer::ReassignTreasurePool(const CCharEntity* PPrev, const CCharEntity* PNew) -> bool
+{
+    if (!PPrev || !PNew)
+    {
+        ShowError("CTreasurePoolContainer::ReassignTreasurePool() - PPrev or PNew was null.");
+        return false;
+    }
+
+    if (auto node = m_Pools.extract(PPrev->id); !node.empty())
```
zach2good, line 146: Can node be const?

src/map/treasure_pool_container.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+    if (withGlobalPool)
+    {
+        m_ZonePools.emplace(0, CTreasurePool(TREASUREPOOL_ZONE, TREASUREPOOL_UNMANAGED));
+    }
+}
+
+// Returns the appropriate treasure pool for PChar in current zone.
+// Priority: Zone wide -> Shared -> Alliance -> Party -> Solo
+auto CTreasurePoolContainer::GetTreasurePool(CCharEntity* PChar) -> CTreasurePool&
```
zach2good: I _love_ this function, it's such a good abstraction. Nice work

src/map/treasure_pool.h
```diff
@@ -26,18 +26,34 @@
...
+{
+    // Membership updated by party/alliance
+    TREASUREPOOL_MANAGED = 0,
+
+    // Membership updated by lua bindings or zone
+    TREASUREPOOL_UNMANAGED
+};
+
+// Update xi.treasurePool accordingly when making changes
 enum TREASUREPOOLTYPE
```
zach2good: While you're here, can you make this: `enum class TreasurePoolType : uint8`, and then the entries will be `Solo`, `Party`, etc.

src/map/treasure_pool.h
```diff
@@ -26,18 +26,34 @@
 
 #include <vector>
 
+enum TREASUREPOOLMANAGEMENT
```
zach2good: Same here, `enum class TreasurePoolManagement : uint8`, `Managed`, `Unmanaged`

src/map/party.cpp
```diff
@@ -1047,7 +1013,10 @@ void CParty::SetLeader(const std::string& MemberName)
         _sql->Query("UPDATE accounts_parties SET partyid = %u WHERE partyid = %u", newId, m_PartyID);
         _sql->Query("UPDATE accounts_parties SET allianceid = %u WHERE allianceid = %u", newId, m_PartyID);
 
-        m_PLeader = GetMemberByName(MemberName);
+        const auto PPrevLeader = dynamic_cast<CCharEntity*>(GetLeader());
+        m_PLeader              = GetMemberByName(MemberName);
+        const auto PNewLeader  = dynamic_cast<CCharEntity*>(m_PLeader);
+
```
zach2good, line 1017: As a pattern, if you're doing `dynamic_cast` the immediate next thing should be to check that it's valid. So you'd do this assignment to `m_PLeader` first, and then the two dynamic casts, and then check they're valid

src/map/party.cpp
```diff
@@ -116,6 +126,28 @@ void CParty::DisbandParty(bool playerInitiated)
...
+            if (auto* PChar = dynamic_cast<CCharEntity*>(member))
+            {
+                auto& PTreasurePool = PChar->GetTreasurePool();
+                PTreasurePools.emplace_back(zoneutils::GetZone(PChar->getZone()), PTreasurePool);
+                PTreasurePool.DelMember(PChar);
+            }
+        }
+
+        // Force release of treasure pools, otherwise the leader would get reattached to the same pool
+        for (const auto& poolToBeReleased : PTreasurePools)
```
zach2good: You can do this decomposition in the if block: `if (const auto& [PZone, PTreasurePool] : ...`

src/map/lua/lua_treasure_pool.h
```diff
@@ -0,0 +1,49 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see http://www.gnu.org/licenses/
+
+===========================================================================
+*/
+
+#ifndef _LUATREASUREPOOL_H
+#define _LUATREASUREPOOL_H
```
zach2good: #pragma once

src/map/lua/lua_treasure_pool.cpp
```diff
@@ -0,0 +1,190 @@
...
+{
+    sol::table result = lua.create_table();
+    int        index  = 1;
+
+    for (const auto& item : m_PLuaTreasurePool->GetItems())
+    {
+        sol::table itemRow   = lua.create_table();
+        itemRow["ID"]        = item.ID;
+        itemRow["SlotID"]    = item.SlotID;
+        itemRow["TimeStamp"] = item.TimeStamp.time_since_epoch().count();
```
zach2good: Off the top of my head, I don't think we often have `Capitalized` keys when we build Lua tables

src/map/lua/lua_treasure_pool.cpp
```diff
@@ -0,0 +1,190 @@
...
+ *  Notes   : ID of 0 means no item present in the given slot.
+ *  {
+ *    { ID = 1, SlotID = 0, TimeStamp = 0, Lotters = { { member = {...}, lot = 0 } } },
+ *    [...]
+ *  }
+ ************************************************************************/
+auto CLuaTreasurePool::getItems() const -> sol::table
+{
+    sol::table result = lua.create_table();
+    int        index  = 1;
```
zach2good: index should be an unsigned type like size_t, unless it can hold -1, etc.

src/map/alliance.cpp
```diff
@@ -237,6 +254,14 @@ void CAlliance::addParty(CParty* party)
         return;
     }
 
+    for (auto* PChar : party->members)
+    {
+        if (auto& PTreasurePool = static_cast<CCharEntity*>(PChar)->GetTreasurePool(); PTreasurePool.IsManaged())
```
zach2good: I know mobs can be in parties, but I don't know if they can have alliances - is it safe to assume entities in here are players and we can `static_cast` them? This isn't the hot path, so we could be checking their objType

scripts/enum/treasure_pool.lua
```diff
@@ -0,0 +1,16 @@
...
+-- Treasure Pool
+-----------------------------------
+xi = xi or {}
+
+---@enum xi.treasurePool
+xi.treasurePool =
+{
+    -- Keep synced with TREASUREPOOLTYPE
+
+    SOLO                     =   1,
```
zach2good, line 10: Spacing here is a bit extreme

zach2good (approved):
Excellent work - I couldn't find anything logically wrong, so I've been anal about naming conventions, arbitrary C++ things, and some "hot path" things. Do these at your leisure.

🎉

src/map/zone.cpp
sruon: Correct.

src/map/treasure_pool_container.cpp
sruon: I don't believe so but while looking at it I realized I don't fully grasp the intricacies of maps and std::move so I might rewrite it in a different way.

src/map/party.cpp
sruon: Changed the flow of that method slightly. PNewLeader will get a static_cast since it's expected in this block that the new leader is a CCharEntity but the PrevLeader may be nullptr (new party creation)

```
void CParty::SetLeader(const std::string& MemberName)
{
    if (m_PartyType == PARTY_PCS)
    {
        uint32 newId = 0;
        int    ret   = _sql->Query(
            "SELECT chars.charid from accounts_sessions JOIN chars ON chars.charid = accounts_sessions.charid WHERE charname = ('%s')", MemberName);

        if (ret != SQL_ERROR && _sql->NumRows() != 0 && _sql->NextRow() == SQL_SUCCESS)
        {
            newId = _sql->GetUIntData(0);
        }
        else
        {
            return;
        }

        _sql->Query("UPDATE accounts_parties SET partyflag = partyflag & ~%d WHERE partyid = %u AND partyflag & %d", ALLIANCE_LEADER | PARTY_LEADER, m_PartyID, PARTY_LEADER);
        _sql->Query("UPDATE accounts_parties SET partyid = %u WHERE partyid = %u", newId, m_PartyID);
        _sql->Query("UPDATE accounts_parties SET allianceid = %u WHERE allianceid = %u", newId, m_PartyID);

        const auto PPrevLeaderEntity = GetLeader();
        m_PLeader                    = GetMemberByName(MemberName);
        const auto PNewLeader        = static_cast<CCharEntity*>(m_PLeader);

        if (this->m_PAlliance && this->m_PAlliance->m_AllianceID == m_PartyID)
        {
            m_PAlliance->m_AllianceID = newId;
        }

        m_PartyID = newId;
        _sql->Query("UPDATE accounts_parties SET partyflag = partyflag | IF(allianceid = partyid, %d, %d) WHERE charid = %u", ALLIANCE_LEADER | PARTY_LEADER, PARTY_LEADER, newId);

        // Passing leader dismisses trusts
        for (auto* PMemberEntity : members)
        {
            if (auto* PMember = dynamic_cast<CCharEntity*>(PMemberEntity))
            {
                PMember->ClearTrusts();

                // In every zone where party members are, reassign the pool to the new leader
                if (PPrevLeaderEntity && PNewLeader)
                {
                    if (const auto PPrevLeader = dynamic_cast<CCharEntity*>(PPrevLeaderEntity))
                    {
                        zoneutils::GetZone(PMember->getZone())->reassignTreasurePool(PPrevLeader, PNewLeader);
                    }
                }
            }
        }
    }
    else
    {
        m_PLeader = members.at(0);
    }
}
```

sruon:
Pushed another commit that addresses review comments and contains bugfixes for alliances.

Most of the tests [are documented here](https://docs.google.com/document/d/1bdE3bKTZtseq3hZi3mLcOkfslbqR-kVVPoYmNqZKH_Q/edit?usp=sharing).

However, when testing across multiple map processes the logic falls apart since there's a strong expectation that party/alliance leaders are set and it's proving too janky to keep everything in sync when party leadership changes.

I'll close the PR and revisit when CParty and CAlliance are reworked.

PR #7158, opened 2025-02-28
[core][lua] Fix Chocobo Digging

UmeboshiXI:
**_I affirm:_**

Fix to allow players to chocobo dig again.

Adds missing `xi.player.onChocoboDig` function in player.lua

1. !additem BUNCH_OF_GYSAHL_GREENS 10
2. Do some digs. See that you can obtain items.

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -4980,7 +4980,7 @@ namespace luautils
     {
         TracyZoneScoped;
 
-        return callGlobal<bool>("xi.player.onChocoboDig", PChar);
+        return callGlobal<bool>("xi.chocoboDig.start", PChar);
     }
 
     // Loads a Lua function with a fallback hierarchy


src/map/lua/luautils.cpp
```diff
@@ -4980,7 +4980,34 @@ namespace luautils
     {
         TracyZoneScoped;
 
-        return callGlobal<bool>("xi.player.onChocoboDig", PChar);
```
zach2good, line 4983: The entry function is:
```lua
xi.chocoboDig.start = function(player)
```

So the only change that needs to happen here is:

```cpp
return callGlobal<bool>("xi.chocoboDig.start", PChar);
```

UmeboshiXI:
Added context:

This is the latest when digging was working: https://github.com/LandSandBoat/server/pull/6090

onChocoboDig was removed/adjusted at some point here: https://github.com/LandSandBoat/server/pull/6780

src/map/lua/luautils.cpp
UmeboshiXI: Adjusted/Pushed Change

zach2good (approved):

Xaver-DaRed (approved):

PR #7162, opened 2025-03-01
Core: Split instance main and entry zones

zach2good:
**_I affirm:_**

`Entry` and `Instance` zones are no longer tied together with Lua and other bits. They communicate with IPC messages. The loading of instances is now _partially_ shipped off to the Async worker thread, so it won't trigger the watchdog on weaker or congested systems.

- Single Process
  - !instance 7700
  - Wait to load in
  - !instance 7700 (to fail)
  - Be safely ejected to the entrance zone
  - !instance 7700 (to enter again)
  - Disconnect and shut everything down
  - Start everything back up and log in
  - Be safely ejected to the entrance zone 
- Multi-Process (with entry and instance zones on different processes)
  - !instance 7700
  - Wait to load in
  - !instance 7700 (to fail)
  - Be safely ejected to the entrance zone
  - !instance 7700 (to enter again)
  - Disconnect and shut everything down
  - Start everything back up and log in
  - Be safely ejected to the entrance zone 

Test with
- 7700
- 5500
- 9300
- 25900

From
- The correct entry zone (!instance)
- The correct entry zone (via NPC)
- A different zone from the entry zone

TODO:
- Check some assaults the whole way through (and the assault helper)
- Light in the Darkness
- Path of Darkness
- Behind the Sluices

### DIFF
--- a/scripts/commands/instance.lua
+++ b/scripts/commands/instance.lua
@@ -27,7 +27,7 @@ commandObj.onTrigger = function(player, instance_id)
         currentInstance:fail()
     else
         player:printToPlayer('Creating instance: ' .. instance_id)
-        player:createInstance(instance_id)
+        player:requestInstance(instance_id)
     end
 end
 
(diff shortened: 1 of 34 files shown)


zach2good:
I've moved too many things around and rebasing this would be harder than starting from scratch - will restart at some point

Issue #7047, opened 2025-02-15
🐛 Cannot send items with the "Alt" icon (aka ITEM_FLAG_MAIL2ACCOUNT) to same account character


almuth150:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Branch affected by issue 

`base`

## Steps to reproduce

Go to a delivery box and attempt to send an item with the Alt icon in the description to either yourself or another character on the same account. !additem 13566 if you don't have such an item.

## Expected behavior

After you enter the name of a valid character to be able to send such an item to, it should be placed into the delivery box to be able to be sent on confirmation. However, it fails to be put into the delivery box, giving the message about communications being congested.

DrDoakHead:
PChar->accid is coming up as 0, so I assume it is uninitialized.  The real answer here is to set the member variable accid on character creation I would think.  But I'm not sure where to look for that.

zach2good:
https://github.com/LandSandBoat/server/pull/7176

PR #7176, opened 2025-03-05
Items that are "EX" and "ALT" can now be sent to characters on the same account

DrDoakHead:
**_I affirm:_**

Closes issue #7047.

Do !additem 13566 to generate an item that can be sent between characters on the same account (or to yourself) that is also EX.  Verify the item can be sent and the "Communication seems congested" error does not appear on screen.

### DIFF
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -712,7 +712,7 @@ namespace charutils
             }
         }
         const auto [_, accid] = charutils::getCharIdAndAccountIdFromName(PChar->getName());
-        PChar->accid = accid;
+        PChar->accid          = accid;
 
         db::query(fmt::format("UPDATE char_stats SET zoning = 0 WHERE charid = {}", PChar->id));
 
--- a/src/map/utils/charutils.cpp
+++ b/src/map/utils/charutils.cpp
@@ -711,6 +711,8 @@ namespace charutils
                 PChar->clearCharVarsWithPrefix("jugpet-");
             }
         }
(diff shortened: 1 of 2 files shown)


DrDoakHead:
PChar->accid is coming up as 0, so I assume it was never set and initialized to 0. The real answer here is to set the member variable accid on character login I would think.  But I'm not sure where to look for that.

WinterSolstice8:
> The real answer here is to set the member variable accid on character login I would think. But I'm not sure where to look for that.

https://github.com/LandSandBoat/server/blob/3ae70de4ae2263a43b8b141d210697eeade1638a/src/map/utils/charutils.cpp#L344

DrDoakHead:
Thanks @WinterSolstice8!  Added it in there and verified it still works :)

src/map/utils/charutils.cpp
```diff
@@ -711,6 +711,8 @@ namespace charutils
                 PChar->clearCharVarsWithPrefix("jugpet-");
             }
         }
+        const auto [_, accid] = charutils::getCharIdAndAccountIdFromName(PChar->getName());
```
WinterSolstice8: this function (LoadChar) already queries from the chars table
https://github.com/LandSandBoat/server/blob/3ae70de4ae2263a43b8b141d210697eeade1638a/src/map/utils/charutils.cpp#L395

you can add `accid` to there and then set it like this

```
PChar->accid = rset->get<uint32>("accid");
```

zach2good (approved):

PR #7181, opened 2025-03-05
Core: Party system rewrite

zach2good:
**_I affirm:_**

My notes and thoughts are in `party_proxy.h`

The skinny is: The map process really don't need much to populate the packets that go to players. Only the content of `partyInfo_t` per member. If we were to serialize a whole alliance update/redefinition - it would only be 520 bytes, chump change for ZMQ and our serialization system.

So, each process that has a party on will have a `PartyProxy`, which will always be kept in line with what's on the world server through complete updates. By asking the party proxy "is this character in the same zone as me" we can swap out the proxy information for the actual entity information. 

Ideas:
- Mob parties and players parties don't need to be the same thing. `MobGroup` as an abstraction makes more sense than trying to jam them into `Party` and `PartyProxy` etc. `MobGroup` is overloaded in the db, so maybe just `MobParty` and `PlayerParty` are unrelated classes with similar names? I suppose the point is that `CBattleEntity*` knows about what "party" an entity is in. What does it need here?
- These flags are probably still useful, could we keep the exact same db layout?
```
enum PARTYFLAG : uint16
{
    PARTY_SECOND    = 0x0001,
    PARTY_THIRD     = 0x0002,
    PARTY_LEADER    = 0x0004,
    ALLIANCE_LEADER = 0x0008,
    PARTY_QM        = 0x0010,
    PARTY_SYNC      = 0x0100,
};
```
- Time member joined
- Leader, SyncTarget, Quartermaster
- PartyNumber, Alliance, PartyLeader, AllianceLeader
- Party EffectFlags packet (easy)
- Interaction with treasurepool

TODO: Better name for party_proxy.

### DIFF
--- a/src/map/packets/party_define.cpp
+++ b/src/map/packets/party_define.cpp
@@ -27,9 +27,9 @@
 #include "entities/trustentity.h"
 #include "utils/zoneutils.h"
 
-const char* partyQuery = "SELECT chars.charid, partyflag, pos_zone, pos_prevzone FROM accounts_parties \
-                                        LEFT JOIN chars ON accounts_parties.charid = chars.charid WHERE \
-                                        IF (allianceid <> 0, allianceid = %d, partyid = %d) ORDER BY partyflag & %u, timestamp";
+const char* partyQuery = "SELECT chars.charid, partyflag, pos_zone, pos_prevzone FROM accounts_parties "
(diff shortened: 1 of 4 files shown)


cocosolos:
Re: mob parties, I think this is mostly used for target finding (targeted/AoE buffs, abilities) and linking.

zach2good:
Others have eventual plans for this, so Im going to close it

PR #7184, opened 2025-03-06
[lua][command] !quest var target

KnowOne134:
**_I affirm:_**

fetch targets info instead of just player that ran command

!quest logid questid playerName
will fetch vars for the playerName target instead of yourself


zach2good (approved):

zach2good:
It seems like I didn't add `getCharVarsWithPrefix` to the CBaseEntity spec file `scripts/specs/core/CBaseEntity.lua`, can you add it while you're here please?

KnowOne134:
Can try. Not sure what I'm doing in there

zach2good:
```lua
---@param prefix string
---@return table
function CBaseEntity:getCharVarsWithPrefix(prefix)
end

```

KnowOne134:
Added

PR #7188, opened 2025-03-07
[core] Einherjar lamp Lua bindings

sruon:
**_I affirm:_**

<del>

Adds several Lua bindings to manage Einherjar lamps
- CLuaBaseEntity::createLamp -> creates a new Glowing Lamp with chamber data saved in exdata
  - Used for initial lamp and duplication of lamps
- CLuaBaseEntity::getLamps -> returns all Glowing Lamps in PC inventory
  - Needed to check for duplicates when PC drops a lamp. If duplicate then PC is not expelled, similar to how old Dynamis worked.
- CLuaBaseEntity::voidLamp -> nulls exdata of a Glowing Lamp and save to DB
  - Retail nulls out a specific lamp if you attempt to use it when reservation is expired
- CLuaBaseEntity::voidLamps -> nulls exdata of all Glowing Lamps and save to DB
  - Retail nulls out all lamps in inventory when expelling from chamber
</del>

- CLuaBaseEntity::findItems() -> Like findItem but return all matching items
  - Used to retrieve all Glowing Lamps from PC inventory
- CLuaBaseEntity::delItemAt() -> Like delItem but takes a specific container/slotId
  - Used to delete a specific Glowing Lamp in PC inventory, since delItem only matches first item it finds.

Can't really test without the rest of the code.

https://www.youtube.com/watch?v=cQpE1iCGGcc

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -4198,6 +4198,38 @@ bool CLuaBaseEntity::delItem(uint16 itemID, int32 quantity, sol::object const& c
 }
 
 /************************************************************************
+ *  Function: delItemAt()
+ *  Purpose : Deletes an item from a player's inventory at specified container/slot
+ *  Example : player:delItemAt(4102, 1, xi.inv.INVENTORY, 5)
+ *  Notes   :
+ ************************************************************************/
+bool CLuaBaseEntity::delItemAt(const uint16 itemID, const int32 quantity, uint8 containerId, const uint8 slotId)
+{
+    auto* PChar = dynamic_cast<CCharEntity*>(m_PBaseEntity);
+    if (!PChar)
+    {
+        ShowWarning("Invalid entity type calling function (%s).", m_PBaseEntity->getName());
+        return false;
+    }
+
+    if (containerId >= CONTAINER_ID::MAX_CONTAINER_ID)
+    {
+        ShowWarning("Lua::delItemAt: Attempting to delete an item from an invalid slot. Defaulting to main inventory.");
+        containerId = LOC_INVENTORY;
+    }
+
+    if (const auto* PItem = PChar->getStorage(containerId)->GetItem(slotId); PItem && PItem->getID() == itemID)
+    {
+        charutils::UpdateItem(PChar, containerId, slotId, -quantity);
+        PChar->pushPacket<CInventoryFinishPacket>();
+
+        return true;
+    }
+
+    return false;
+}
+
+/************************************************************************
  *  Function: delContainerItems()
  *  Purpose : Deletes all items from a specific player's container
  *  Example : player:delContainerItems(xi.inv.INVENTORY)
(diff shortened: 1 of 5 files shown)


src/map/lua/lua_baseentity.cpp
```diff
@@ -18861,6 +18861,150 @@ void CLuaBaseEntity::claimContestReward()
     }
 }
 
+/************************************************************************
+ *  Function: createLamp()
```
Xaver-DaRed: Can we have the name be verbose?
As in `createEinherjarLamp`

There are many lamps. We have nyzul using lamps, old dynamis used some similar thing... Just to make it immediately clear what this is used for

Xaver-DaRed: Speaking with Claywar, it seems this functions arent needed, since we have ways to set exdata in lua
`player:addItem({ id = itemID, exdata = { [10] = 10 } })`
add 1 item of itemID, with the exdata at index 10 (0-indexed!) set to 10

Similarly, with voidLamp, you can just delete the item and immediately add an empty one

sruon:
I had my code originally working with generic bindings doing the heavy lifting in Lua as opposed to the Einherjar specific bindings I originally proposed in this PR. 

Based on feedback, it appears this approach is preferred.

Therefore new proposal:
- findItems() -> Like findItem but return all matching items
- delItemAt() -> Like delItem but takes a specific container/slotId

Using these, combined with addItem allows me to perform all the required lamp interactions.

Xaver-DaRed (approved):

Xaver-DaRed:
If I may ask, what would the actual use of `delItemAt()` be?
I ask because players can and will move items among containers and can even use addons to automate it, leading to a potential abuse with such kind of interactions.

Just curiosity, tbh. I still like the additions.

sruon:
```
-- Chamber is expelling player, void ALL matching lamps
for _, item in ipairs(player:findItems(xi.item.GLOWING_LAMP)) do
    local lampData = xi.einherjar.decypherLamp(item)

    if lampData.chamber == self.id and lampData.enter == self.startTime then
        xi.einherjar.voidLamp(player, item)
    end
end

-- Void one lamp, this can be called in a loop by the chamber OR when PC uses an expired lamp
xi.einherjar.voidLamp = function(player, lampObj)
    player:delItemAt(xi.item.GLOWING_LAMP, 1, lampObj:getLocationID(), lampObj:getSlotID())
    player:addItem({
        id = xi.item.GLOWING_LAMP,
        exdata = {
            [0] = 0x0,
            [2] = 0x0,
            [8] = 0x0,
            [9] = 0x0,
            [10] = 0x0,
            [11] = 0x0,
            [12] = 0x0,
            [13] = 0x0,
            [14] = 0x0,
            [15] = 0x0,
        },
    })
end
```

I thought about the case you mentioned but this is all happening synchronously before any other player packet is processed.

Note: I still need to check if retail zero out lamps in other containers, if you put one in satchel for example.

zach2good (approved):

PR #7189, opened 2025-03-07
Fixed Ramponneau placeholder to match retail (16D).

kngrace:
**_I affirm:_**

Fixed Ramponneau placeholder to match retail (16D).

Utilized addons (renamer & hxui) to confirm which is the correct placeholder. The current/incorrect placeholder spawns at (F-10) in a pack of 10 toads instead of at (I-9)/(I-10) in a pack of 3 toads (one toad appears to be missing?). Once the current/incorrect placeholder is defeated at (F-10) the NM still spawns in its expected location of (I-9)/(I-10), making for a weird experience.

### DIFF
--- a/scripts/zones/West_Sarutabaruta_[S]/mobs/Toad.lua
+++ b/scripts/zones/West_Sarutabaruta_[S]/mobs/Toad.lua
@@ -10,7 +10,7 @@ local entity = {}
 
 local ramponneauPHTable =
 {
-    [ID.mob.RAMPONNEAU - 1] = ID.mob.RAMPONNEAU, -- 78.836 -0.109 -199.204
+    [ID.mob.RAMPONNEAU - 4] = ID.mob.RAMPONNEAU, -- 78.836 -0.109 -199.204
 }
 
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
If I may ask, where did you use those addons? Inretail?

From your description, it looks more a matter of the mob being wrongly placed, not the id being wrong

kngrace:
I used those addons on a private server running LSB code to determine which is the correct placeholder for LSB code. The incorrect mob is currently marked as a placeholder. The NM currently spawns at its expected location (I-9)/(I-10) and the corrected placeholder is also at (I-9)/(I-10). The incorrect placeholder (without my code change) is way over at (F-10).

Xaver-DaRed:
I want to be absolutely sure: Have you gotten the placeholder ID from retail (or from a souce based on retail) or are you basing this on the fact the placeholder should be positioned in certain place?

Xaver-DaRed (approved):

Xaver-DaRed:
Went to retail to confirm this myself, just in case. Checked out.

PR #7199, opened 2025-03-09
[BCNM] BCNM20 Charming Trio Updates

The-Aerec:
Update charming trio mobs with latest capture data

**_I affirm:_**

Implements changes to Charming Trio based on latest capture data from Siknoz:
https://docs.google.com/spreadsheets/d/1TnrBzUAQ0hyuFVIjf5OLviIfhGw4vxN1x_4zv9gG4N4/edit?usp=sharing

Updated changes to:
Spawn Position
Mob Levels
Mob Abilities + Ability lists
Mob Casting Cooldowns
Mob Resistances, flags, etc from Capture data. 

All the relevant supporting data is contained in the above Google Doc.

Each Leech in the BCNM only has a single skill they use, 2 of them have 'special' changes to these abilities:
Random Kiss - TP/MP/HP Drain Only 1 at a time
Absorbing Kiss - Drain all stats (Up to -8 per stat)
Deep Kiss - Drain ALL effects (except food/reraise)

Each Leechs level, HP, and other common mob data were updated based on capture.

Mob spell CD was observed to be ~45s after casting drain, and ~15s after casting aspir. These special cast cooldowns have been added to the new mob luas.

SSs from testing are linked at the bottom of the Google Doc.

Enter BCNM20 Charming Trio and fight away!


scripts/actions/mobskills/absorbing_kiss.lua
```diff
@@ -10,6 +10,15 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    if mob:getPool() == 3116 then -- BCNM20 Charming Trio: Pepper does all attribute drain
+        for i = 136, 142 do
+            xi.mobskills.mobDrainAttribute(mob, target, i, 8, 3, 120) -- Can drain up to -8 per stat
```
zach2good, line 19: Even with the comment this is very hard to parse. This is handled a little better in other places:
```lua
    -- str down - chr down
    local effectType = math.random(136, 142)

    skill:setMsg(xi.mobskills.mobDrainAttribute(mob, target, effectType, 10, 3, 120))
```

But it could be even easier:

```
enum/effects.lua:
    STR_DOWN                 = 136,
    DEX_DOWN                 = 137,
    VIT_DOWN                 = 138,
    AGI_DOWN                 = 139,
    INT_DOWN                 = 140,
    MND_DOWN                 = 141,
    CHR_DOWN                 = 142,

for effect = xi.effect.STR_DOWN, xi.effect.CHR_DOWN do
   ...
```

scripts/actions/mobskills/absorbing_kiss.lua
```diff
@@ -10,6 +10,15 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
...
+            skill:setMsg(xi.msg.basic.ATTR_DRAINED) -- No message
+        end
+
+        return 7
```
zach2good: What's 7 do? Do we have this mapped to something human readable?

scripts/actions/mobskills/deep_kiss.lua
```diff
@@ -10,6 +10,23 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
...
+        if count == 0 then
+            skill:setMsg(xi.msg.basic.SKILL_NO_EFFECT)
+            return count
+        end
+
+        for i = 1, count do
+            mob:stealStatusEffect(target, xi.effectFlag.DISPELABLE) -- Steal all status effects valid on the player
+        end
+
+        skill:setMsg(xi.msg.basic.EFFECT_DRAINED)
```
zach2good, line 25: I feel like a helper for this exists, but I don't know what it would be called

scripts/actions/mobskills/absorbing_kiss.lua
UmeboshiXI: I think this is a return for the purpose of the messaging. 

`ATTR_DRAINED = 369, -- <user> uses <ability>. <amount> of <target>'s attributes is drained.`

There are 7 effects so it would say target had 7 attributes drained.

scripts/actions/mobskills/deep_kiss.lua
UmeboshiXI:
```
xi.mobskills.mobDrainStatusEffectMove = function(mob, target)
    -- If target has Hysteria, no message skip rest
    if mob:hasStatusEffect(xi.effect.HYSTERIA) then
        return xi.msg.basic.NONE
    end

    -- try to drain buff
    local effect = mob:stealStatusEffect(target)

    if effect ~= 0 then
        return xi.msg.basic.EFFECT_DRAINED
    end

    return xi.msg.basic.SKILL_NO_EFFECT
end
```

Could probably incorporate it into this with an added optional parameter that dictates how many times `mob:stealStatusEffect(target)` is called.

The-Aerec: Unfortunately the above would have to be re-written quite a bit to deal with the added need for more than 1 effect. This can be done if that is the desire.

scripts/actions/mobskills/absorbing_kiss.lua
The-Aerec: Updated for clarity

scripts/actions/mobskills/absorbing_kiss.lua
The-Aerec: Updated for clarity, and Umeboshi is correct.

scripts/actions/mobskills/absorbing_kiss.lua
```diff
@@ -10,12 +10,25 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
+    local numAttriDrained = 1
+
+    if mob:getPool() == 3116 then -- BCNM20 Charming Trio: Pepper does all attribute drain
+        numAttriDrained = 7
+
+        for i = xi.effect.STR_DOWN, xi.effect.CHR_DOWN do
+            xi.mobskills.mobDrainAttribute(mob, target, i, 8, 3, 120) -- Can drain up to -8 per stat
```
WinterSolstice8, line 19: You need to check the return value of `mobDrainAttribute`

```
xi.mobskills.mobDrainAttribute = function(mob, target, typeEffect, power, tick, duration)
    if not drainEffectCorrelation[typeEffect] then
        return xi.msg.basic.SKILL_NO_EFFECT
    end

    local results = xi.mobskills.mobStatusEffectMove(mob, target, typeEffect, power, tick, duration)

    if results == xi.msg.basic.SKILL_ENFEEB_IS then
        mob:addStatusEffect(drainEffectCorrelation[typeEffect], power, tick, duration)

        return xi.msg.basic.ATTR_DRAINED
    end

    return xi.msg.basic.SKILL_MISS
end
```

If it returns `xi.msg.basic.ATTR_DRAINED` then increment `numAttriDrained`, and only set the skill message if  numAttriDrained > 0. Make sure to set `numAttriDrained` to 0 before the loop.

The original is probably slightly wrong, returning `1` likely has no obvious effect if the skill isn't set to `xi.msg.basic.ATTR_DRAINED`, though. So it may not need to change.

WinterSolstice8:
The current LLS error will be fixed with a rebase

zach2good (approved):

PR #7222, opened 2025-03-13
[core][lua] mobMagicalMove Refactor + Suicide Fix

UmeboshiXI:
**_I affirm:_**

### Mobskill changes to Core:

* Adds a way to track a mob's HP at the time a skill was started (Used for suicide skills that deal damage based on remaining HP).

### mobMagicalMove Refactor:

* Once/If the function is reviewed and "OK'd" after feedback, I'll proceed on adapting the other mob skills to use it.

* mobMagicalMove function now uses a param system similar to what player weaponskills use. 

* This allows us to set multiple TP effects if needed as well as more options to adjust mobskills (Such as if a skill has a flat accuracy bonus or similar that is not common to other skills). It also reduces clutter and makes it easier to read what a mob skill does.

* Handling of skill TP return has been moved to mobFinalAdjustments.

* Corrects bomb_toss and bomb_toss_suicide damage/ftp values. (Values pulled from Jimmayus Spreadsheet)

List of Params that can be used in a mobskill's lua file:

```
params.baseDamage   = mob:getWeaponDmg()
params.fTP          = 1
params.element      = xi.element.NONE
params.ignoreResist = true/false
params.damageVaries = { 1.00, 1.00, 1.00 }
```

* params.baseDamage - Allows you to change the base damage of a skill. Defaults to "getMainLvl() + 2" if nil.

* params.fTP - Acts as a linear multiplier to base damage. Acts like old dmgmod.

* params.element - Sets the element of the skill.

* params.ignoreResist - If true, skill will skip resist / DMGMAGIC steps. (See Nether Blast)

* params.damageVaries - Scales baseDamage of skill based on TP.

Included are 2 versions of bomb toss to use as tests. (Regular and Suicide versions)

Currently prints are enabled for debugging/testing purposes.

1. Find a goblin and spam !tp 3000 etc. Use !immortal on it to prevent it from dying.
1b. Use !exec target:useMobAbility(591 or 592, player) without aggroing it. You can set TP and then force its skill on you to test different TP ranges.

2. Play with other settings, see no errors (Hopefully).

### DIFF
--- a/scripts/actions/mobskills/bomb_toss.lua
+++ b/scripts/actions/mobskills/bomb_toss.lua
@@ -10,9 +10,15 @@ mobskillObject.onMobSkillCheck = function(target, mob, skill)
 end
 
 mobskillObject.onMobWeaponSkill = function(target, mob, skill)
-    local damage = mob:getWeaponDmg() * 3
+    local params = {}
 
-    damage = xi.mobskills.mobMagicalMove(mob, target, skill, damage, xi.element.FIRE, 1, xi.mobskills.magicalTpBonus.MAB_BONUS, 1)
+    params.baseDamage   = mob:getWeaponDmg()
+    params.fTP          = 3
+    params.element      = xi.element.FIRE
+    params.damageVaries = { 1.0, 1.0, 1.0 }
+    params.ignoreResist = false
+
+    local damage = xi.mobskills.mobMagicalMove(mob, target, skill, params)
     damage = xi.mobskills.mobFinalAdjustments(damage, mob, skill, target, xi.attackType.MAGICAL, xi.damageType.FIRE, xi.mobskills.shadowBehavior.IGNORE_SHADOWS)
 
     target:takeDamage(damage, mob, xi.attackType.MAGICAL, xi.damageType.FIRE)
--- a/scripts/actions/mobskills/bomb_toss_suicide.lua
+++ b/scripts/actions/mobskills/bomb_toss_suicide.lua
@@ -6,7 +6,8 @@
 local mobskillObject = {}
 
 mobskillObject.onMobSkillCheck = function(target, mob, skill)
-    -- notorious monsters shouldn't explode, nor dynamis
+    -- Notorious monsters and goblins in Dynamis do not explode.
+    -- TODO: Adjust weighting between regular Bomb Toss and Bomb Toss Suicide.
     if mob:isMobType(xi.mobType.NOTORIOUS) or mob:isInDynamis() then
         return 1
     end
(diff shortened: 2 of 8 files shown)


Frankie-hz:
I have been looking at this trying to think of a better way to do it without params but I don't know if there is a better way to do it. In terms of fitting everything we need, and having the flexibility to add new things it might be the best way to do it.

A general knowledge of how magical mobskills work and what we determined last year based off retail:
1) Most mobskills (close to probably 90%+) base damage is not based off of the mobs base damage but strictly use a `Level + 2`. My original idea was to have it set that as a base in the magical mob move but setting it like this is probably fine (see below on  just a few niche cases)
2) Some magical skills follow this formula -> `Level + 2 + dInt`, where `dInt` is `dInt * 1` so we would need a place to have that being an option.
3) Several skills, like manticores have a base damage of `Level * 4` or something like Behemoth where meteor damage is `14 + Level * 30` with non-elemental where this is the "total damage" not "base damage". In these cases we would probably just set FTP to 1 and use the damage formula in the baseDamage but I'll let others comment on that as well.
4) `damageVaries` and the scale at which it varies is still unknown (pending captures) - I would be ok coming up with a determination on a universal scalable formula for now until we figure out if its skill specific or it follows a certain scale (similar to how jugs have a universal damage scale based on TP for physical moves we may be able to come up with something for magical https://www.ffxiah.com/forum/topic/45830/killer-instinct-the-beastmaster-compendium/#magicalready ). 
5) `mAccVaries` based on percentage seems extremely unlikely - Do you have proof of this at all?
6) `fTP` :white_check_mark: 
7) `mAccBonus` -> not sure this actually exists or not because normally the mob itself just has an inherit macbonus addition not specific skills. Do you have examples of this happening?

UmeboshiXI:
Thanks for feedback. I was digging through ASB and reading JP documents and saw the dStat multiplier for avatars etc. I added that last night, just need to push an update.

> Several skills, like manticores have a base damage of Level * 4 or something like Behemoth where meteor damage is 14 + Level * 30 with non-elemental where this is the "total damage" not "base damage". In these cases we would probably just set FTP to 1 and use the damage formula in the baseDamage but I'll let others comment on that as well.

Yeah baseDamage defaults to mob damage if its not defined but you can plug in anything there like the behemoth equation and it will use that.

> mAccVaries based on percentage seems extremely unlikely - Do you have proof of this at all?

No hard proof atm.
I did see a couple macc varies on the spread sheet(Bomb toss being one of them).
Perhaps an option for a flat modifier based on TP scale would make more sense? There is not too much info on this atm tbh.
If MACC Varies(% or flat) is just a thing that doesn't exist I can remove it.

> mAccBonus -> not sure this actually exists or not because normally the mob itself just has an inherit macbonus addition not specific skills. Do you have examples of this happening?

Again no hard proof as I was working off of the comments and was adding functionality for things that may be needed based on the comments. Can cut it out if we don't need.

UmeboshiXI:
Pushed a change with your input. baseDamage is now defaults to getMainLvl() + 2. Removed MACC related things pending more information/captures. Added dStatMultiplier.

UmeboshiXI:
Saw that this was still open so closing now. Superseded by https://github.com/LandSandBoat/server/pull/8360

PR #7226, opened 2025-03-13
[sql] Add missing zonelines for Cirdas Caverns - Woh Gates

UmeboshiXI:
**_I affirm:_**

Adds missing zoneline entries for Cirdas Caverns > Woh Gates and Woh Gates > Cirdas Caverns.

Prior, attempting to zone would give an error:
```
SmallPacket0x5E: Zone line 812201850 not found (SmallPacket0x05E : 3027)
SmallPacket0x5E: Zone line 913861754 not found (SmallPacket0x05E : 3027)
```

POS after zoning is _**not**_  retail captured (Left a TODO) but will place the player next to the zone line with a small 10~ yalm buffer distance. Player is also pointed away from the zone line after they zone.

Zone between Cirdas and Woh Gates then back again. Get no errors, appear in the other zone as expected.


claywar:
What map pos are you looking at for this transition (from woh to cirdas)? Its an easy cap, so I can't see letting this in without good data.

TeoTwawki:
> POS after zoning is not retail captured (Left a TODO) but will place the player next to the zone line with a small 10~ yalm buffer distance. Player is also pointed away from the zone line after they zone.

Fun fact: retail doesn't use a single static position, and sort of randomly places you within a certain distance of it. So if a zoneline is _perfectly_ aligned with a cardinal direction all we need is the correct distance (and someday, which access to random the value of), but those angled ones will be tricky to handle the day we start doing what retail actually does instead of dropping people on a fixed point. 👨‍🏫 

Same behavior happens exiting the mog house. You -can- all land on top of each other and they do not explicitly check for it, so I guess SE lazy random'd that bit.

PR #7246, opened 2025-03-17
[core] MobMod non-exclusive Claim

sruon:
**_I affirm:_**
- [X] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [X] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.

Introduces a new Mob modifier allowing anyone to attack a given monster, regardless of who holds the claim currently.

Regular claim rules still apply:
- Mob will show red to the entity currently holding the "claim"
- Mob will show purple to everyone else
- Claim ownership will flap between the various entities, depending on who has the most enmity.
- Rewards will go to the last claiming entity (if any)

It's different from the current `setClaimable` feature in the following fashion:
- setClaimable(false) disables any reward processing (EXP/loot)
- setClaimable(false) forces the mob to show as white, which I don't think ever happens on Retail.

See retail examples:
- [Besieged](https://youtu.be/y0M0Jx1ppwU?t=338)
- [Campaign](https://www.youtube.com/watch?v=F75PvJkaEEk)
- [Wildskeeper Reive](https://www.youtube.com/watch?v=mEGPMXTW-tU)

This is a cosmetic change at best for most content, but Einherjar needs it to process database drop tables (EXP books from Regular + Odin)

Partially based on a custom patch that originated from Xaver.

With 2 chars:

P1: !fafnir
P1: Claim Fafnir

P2: Attempt to attack Fafnir. "Cannot attack. Your target is already claimed."

!setmobmod NON_EXCLUSIVE_CLAIM 1

P2: Attempt to attack Fafnir. Works.

Deal damage with both, see claim status flap and nameplate change from red to purple and vice versa.


zach2good:
I'm a little busy today so I can't look at this closely, but is this different to this? https://github.com/LandSandBoat/server/commit/a5b37e9bb6f95990dbdf248b7ba32a30198df7fa

zach2good:
Ah I see, very good. Would it be useful to combine these two systems? Either the binding or the mobmod with different values depending on if you want loot/exp

sruon:
> I'm a little busy today so I can't look at this closely, but is this different to this? [a5b37e9](https://github.com/LandSandBoat/server/commit/a5b37e9bb6f95990dbdf248b7ba32a30198df7fa)

Frankly I'm not sure what use case the current getClaimable/setClaimable implementation serve because I can't think of a scenario where a mob cannot be claimed on retail. 

**edit:** actually it might be used for claim shield in LSB

It's been used on certain servers to work around the issue of allowing multiple unrelated entities to attack a mob (custom content zone bosses for example) but it's not how retail works for this sort of content.

ASB may have something similar for old-style Dynamis, I didn't check.

sruon:
> Ah I see, very good. Would it be useful to combine these two systems? Either the binding or the mobmod with different values depending on if you want loot/exp

0 = unclaimable
1 = regular claim
2 = free for all
3 = free for all no rewards

Something like that? Then get rid of the m_IsClaimable and change the binding to apply the mobmod.

zach2good:
Sounds good to me! As long as those values get human-readable enums
somewhere (maybe below the main table in mobmods, etc.)

zach2good:
Also, 0 should be regular claim - the default

Xaver-DaRed (approved):

sruon:
- Removed m_IsClaimable from CMobEntity
- Changed the mobmod to CLAIM, setting the claim behavior
- Added the enums
  - Did not add the ones dealing with rewards/no rewards because MOBMOD_NO_DROPS and MOBMOD_EXP_BONUS can be used at the same time to get the right effect.
- Updated the Lua bindings to use the mobmod instead

Tested with the claim shield since that's the only part of the codebase using these.

There's another part of the patch that may need to be included, I am investigating so please do not merge yet.

scripts/enum/claim.lua
```diff
@@ -0,0 +1,12 @@
+-----------------------------------
+-- Claim type
+-----------------------------------
+xi = xi or {}
+
+---@enum xi.claimType
+xi.claimType =
+{
+    EXCLUSIVE     = 0, -- Regular exclusive claim behavior. Only one entity can attack.
```
zach2good: Only one entity can attack -> Only one entity and related group*, or something like that

scripts/enum/mob_mod.lua
```diff
@@ -91,4 +91,5 @@ xi.mobMod =
     SKIP_ALLEGIANCE_CHECK  = 80, -- Skip the allegiance check for valid target (allows for example a mob to cast a TARGET_ENEMY spell on itself)
     ABILITY_RESPONSE       = 81, -- Mob can respond to player ability use with onPlayerAbilityUse()
     RUN_SPEED_MULT         = 82, -- Multiplier for the speed of a mob while running (generally when the target is out of range) 100 = 1.00x
+    CLAIM                  = 83, -- Changes the claim behavior of the mob. See xi.claimType enum.
```
zach2good: CLAIM_TYPE

src/map/lua/lua_baseentity.cpp
```diff
@@ -13170,7 +13170,8 @@ void CLuaBaseEntity::setClaimable(bool claimable)
 {
     if (auto* PMob = dynamic_cast<CMobEntity*>(m_PBaseEntity))
     {
-        PMob->m_IsClaimable = claimable;
+        PMob->setMobMod(MOBMOD_CLAIM, claimable ? ClaimType::Exclusive : ClaimType::Unclaimable);
```
zach2good: I reckon we can just get rid of this binding and operate solely through updating the mobmod

src/map/entities/mobentity.h
```diff
@@ -102,6 +102,13 @@ enum BEHAVIOR : uint16
     BEHAVIOR_NO_TURN      = 0x400  // mob does not turn to face target
 };
 
+enum ClaimType : uint8
```
zach2good: Make this an enum class

src/map/lua/lua_baseentity.cpp
```diff
@@ -13187,7 +13188,7 @@ bool CLuaBaseEntity::getClaimable()
 {
     if (auto* PMob = dynamic_cast<CMobEntity*>(m_PBaseEntity))
     {
-        return PMob->m_IsClaimable;
+        return PMob->getMobMod(MOBMOD_CLAIM) < ClaimType::Unclaimable;
```
zach2good: Check for the two claimable types, rather than `< Unclaimable` (will work in line with enum class)

src/map/lua/lua_baseentity.cpp
sruon: I don't disagree but that would break the claim shield and I'm not sure how to fix that one just yet.

The sequence of event looks more or less like that:

* CMobEntity::Spawn -> 
  * CBaseEntity::Spawn -> 
    * SPAWN listener events (where claim shield makes changes) 
  * CMobEntity::CalculateStats
     * PMob->restoreMobModifiers() 
       * (this resets to m_mobModStatSave, which is I guess either DB values or anything stored after an explicit saveMobModifiers, like in this PR)

So the claim shield would need to execute after CMobEntity::Spawn has completed, or to call saveMobModifiers which is not exposed through Lua bindings today.

zach2good: claim shield (even though I wrote it) is a niceity for downstream, but absolutely isn't needed to retail parity. If it needs to break to enable actual retail things then so be it.

I'm surprised that changing a mobmod doesn't automatically call: saveMobModifiers, seems like a thing that should happen.

sruon: I'll try updating the claim shield, I think I can make it work with minimal changes.

sruon:
- Removed setClaimable/getClaimable bindings.
- Updated claim shield to use setMobMod
- Added other part of Xaver's patch in targetfind

Should be good this time around.

Xaver-DaRed:
Added `squash` tag so this gets squashed before merging, as we spoke about.

Looking good

zach2good (approved):

zach2good:
Just saw this in CI:
```
xi_map.exe: [03/21/25 10:30:01:954][map][info] Loading Mob scripts (zoneutils::LoadMOBList:648)
xi_map.exe: [03/21/25 10:30:04:006][map][lua] Applying Claimshield to  for [75](https://github.com/LandSandBoat/server/actions/runs/13989436143/job/39170668922?pr=7256#step:10:76)00ms (lua_print:203)
xi_map.exe: [03/21/25 10:30:04:270][map][lua] Applying Claimshield to  for 7500ms (lua_print:203)
xi_map.exe: [03/21/25 10:30:05:064][map][lua] Applying Claimshield to  for 7500ms (lua_print:203)
xi_map.exe: [03/21/25 10:30:05:300][map][lua] Applying Claimshield to  for 7500ms (lua_print:203)
xi_map.exe: [03/21/25 10:30:13:827][map][lua] Applying Claimshield to  for 7500ms (lua_print:203)
xi_map.exe: [03/21/25 10:30:13:827][map][lua] Applying Claimshield to  for 7500ms (lua_print:203)
xi_map.exe: [03/21/25 10:30:13:827][map][lua] Applying Claimshield to  for 7500ms (lua_print:203)
xi_map.exe: [03/21/25 10:30:13:827][map][lua] Applying Claimshield to  for 7500ms (lua_print:203)
xi_map.exe: [03/21/25 10:30:13:[82](https://github.com/LandSandBoat/server/actions/runs/13989436143/job/39170668922?pr=7256#step:10:83)7][map][lua] Applying Claimshield to  for 7500ms (lua_print:203)
```

I suspect it's just a problem with the logging, rather than just the claimshield

sruon:
The name was already missing when I was testing before making any change, did not investigate further.

PR #7253, opened 2025-03-18
Core: Replace socket.cpp with ASIO & Introduce MapSocket

zach2good:
**_I affirm:_**

I removed a lot of `socket.cpp` in https://github.com/LandSandBoat/server/pull/7238, but this fully removes the file.

We now have a simple wrapper around ASIO::io_context::run_for(200ms)` which acts exactly the same as our previous implementation, except cleared and more encapsulated. 

Also cleaned up a bunch of things, left a lot of comments, prepared for the upcoming death of `kernel.cpp`

## Testing

Logged in, ran around in Dynamis Bastok, zoned, etc.


zach2good:
Ready to go

claywar (approved):

PR #7258, opened 2025-03-19
use pre 2013 accuracy unless 2013 or later expansion is enabled

Grahf0085:
**_I affirm:_**

This page describes how in 2013 the amount DEX affected accuracy changed: https://www.bg-wiki.com/ffxi/The_History_of_Final_Fantasy_XI/2013 - (search for "Status modifiers for hand-to-hand, one-handed, and ranged weapons have been adjusted as given below."). The values LSB uses appears to be the post 2013 values (of course). I wanted to try to use the pre 2013 values when no expansion from 2013 or later is enabled.

I attempt to implement part of that change in this PR. It's not complete and the repeat code can be removed. But before I spend  a wittle time on it I wanted to see if this is something that would be entertained. 

TBD

### DIFF
--- a/src/map/entities/battleentity.cpp
+++ b/src/map/entities/battleentity.cpp
@@ -1003,16 +1003,17 @@ uint16 CBattleEntity::ACC(uint8 attackNumber, uint16 offsetAccuracy)
             }
             skill = SKILL_HAND_TO_HAND;
         }
-        int32 ACC = GetSkill(skill) + iLvlSkill;
-        ACC       = (ACC > 200 ? (int16)(((ACC - 200) * 0.9) + 200) : ACC);
+        int32 ACC           = GetSkill(skill) + iLvlSkill;
+        ACC                 = (ACC > 200 ? (int16)(((ACC - 200) * 0.9) + 200) : ACC);
+        float dexMultiplier = settings::get<bool>("main.USE_PRE_2013_DEX_MULTIPLIER") ? 0.50f : 0.75f;
         if (auto* weapon = dynamic_cast<CItemWeapon*>(m_Weapons[SLOT_MAIN]); weapon && weapon->isTwoHanded())
         {
-            ACC += (int16)(DEX() * 0.75);
+            ACC += (int16)(DEX() * dexMultiplier);
             ACC += m_modStat[Mod::TWOHAND_ACC];
         }
         else
         {
-            ACC += (int16)(DEX() * 0.75);
+            ACC += (int16)(DEX() * dexMultiplier);
         }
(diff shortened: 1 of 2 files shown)


src/map/entities/battleentity.cpp
```diff
@@ -1003,16 +1003,17 @@ uint16 CBattleEntity::ACC(uint8 attackNumber, uint16 offsetAccuracy)
             }
             skill = SKILL_HAND_TO_HAND;
         }
-        int32 ACC = GetSkill(skill) + iLvlSkill;
-        ACC       = (ACC > 200 ? (int16)(((ACC - 200) * 0.9) + 200) : ACC);
+        int32 ACC       = GetSkill(skill) + iLvlSkill;
+        ACC             = (ACC > 200 ? (int16)(((ACC - 200) * 0.9) + 200) : ACC);
+        bool use2013Dex = settings::get<bool>("main.ENABLE_SOA") || settings::get<bool>("main.ENABLE_ROV") || settings::get<bool>("main.ENABLE_TVR");
```
zach2good: To me, it would make more sense to break this out into its own setting, the same as we have for the Adoulin WS calculations.

So this would then look like:
```cpp
float dexMultiplier = settings::get<bool>("main.USE_2013_DEX_MULTIPLIER") ? 0.75f : 0.50f;

...

ACC += (int16)(DEX() * dexMultiplier);
```

Frankie-hz: I would also vote for a setting. If a setting is not what we want we could also move the calcs to lua and make a module.

Grahf0085: Will make sure it compiles tonight test tomorrow if that's ok

zach2good: It's important to remember that we are retail facing backwards, so the default for this setting should be whatever retail is, so the defaults should be either:

USE_PRE_2013_DEX_MULTIPLIER = false

or

USE_2013_DEX_MULTIPLIER = true

I'd probably vote for the "pre" version

Grahf0085: I hear you. I feel like this is what would best fit LSB. Default value is settings/main.lua keeps everything as it was. If you switch the value of the new setting to true it will use less DEX when calculating accuracy. 

I tested it with a one handed weapon. Used printf to check the values of the variables. Got a value of 302 for the ACC variable when the DEX multiplier was 0.75.

Then I changed settings/main.luas value for  USE_PRE_2013_DEX_MULTIPLIER  to true. DEX multiplier printed as 0.50 and ACC printed as 283.

zach2good (approved):

Xaver-DaRed (approved):

PR #7279, opened 2025-03-23
Core: Refactor main loop and add performance stats

zach2good:
**_I affirm:_**

- Makes the main loop much clearer about what it does, on what deadlines
- Adds optional debug performance logging
- Adds an artificial sleep Lua call and command to help with testing
- Renames many things to make it much clearer about what they're doing
- Adds rate-limited warnings for if performance is severely degraded
image


zach2good:
Ready to go

claywar (approved):

PR #7295, opened 2025-03-24
[sql] Add ITEM_FLAG_SCROLL to several spells

sruon:
**_I affirm:_**

Adds ITEM_FLAG_SCROLL to the following scrolls:
- Haste II
- Distract
- Distract II
- Frazzle
- Frazzle II
- Gekka Ichi
- Yain Ichi
- Flurry
- Flurry II

Attempting to learn these spells while already knowing them causes the client to hard crash because of a bogus ID in the packet

https://github.com/LandSandBoat/server/blob/b2f7f2a7feb2a2a950cee9bff23d92d7a78853b9/src/map/ai/states/item_state.cpp#L104-L108

Note that the flag is not set in the DATs for some reason

image

- !addallspells
- !additem scroll_of_flurry
- Use scroll
- Don't crash


Xaver-DaRed (approved):

zach2good (approved):

zach2good:
I'd probably leave comments next to the items where the dat's don't have the correct flags

PR #7302, opened 2025-03-25
[lua] Add mission acceptance text ID for Heaven's Tower

slashtangent:
**_I affirm:_**

Adds missing text ID for if the player accepts a mission in Heaven's Tower.
image

Used in all Windurst Missions:
``player:messageSpecial(zones[zoneId].text.YOU_ACCEPT_THE_MISSION)``

1. ``!zone <Heaven's Tower>``
2. ``!exec player:messageSpecial(zones[xi.zone.HEAVENS_TOWER].text.YOU_ACCEPT_THE_MISSION)``
3. Alternatively, start a Windurst mission in Heaven's Tower


zach2good (approved):

zach2good:
Thanks for the fix 👍 

Presumably, before this change there would be errors in the logs?

```cpp
void CLuaBaseEntity::messageSpecial(uint16 messageID, sol::variadic_args va)
```

If you tried to send a nil in here it should log, so people haven't been reporting issues 🤷

Xaver-DaRed (approved):

Issue #6717, opened 2025-01-14
🔨 CBasicPacket's children shouldn't have member vars


zach2good:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature

As just seen in https://github.com/LandSandBoat/server/pull/6713, packets should just be wrappers on top of `CBasicPacket` and only have constructors that write to the underlying buffer, and accessors that read that buffer in different ways.

zach2good:
Relevant commit: https://github.com/LandSandBoat/server/pull/6713/commits/29f953430ba2f0e688d8406ad9c44a3c36a08ac9

AJBats:
I'm looking for some good warmup issues to get into LSB contribution. I'm looking at this one and it seems like all of the remaining classes that extend CBasicPacket (searched for "public CBasicPacket") have no members in them. Perhaps this issue has already been resolved?

Let me know, I would love to help out!

zach2good:
Yeah, seems like this isn't needed anymore 👍

PR #7307, opened 2025-03-26
[core] Improve Error Condition in CPartyEffectsPacket

0x05010705:
Adds in additional checks into AddMemberEffects to account for negatives and values above 5.

**_I affirm:_**

Improves error catching logic in `CPartyEffectsPacket`
- Previously, it only accounted for if `members == 5`, but not if the count ever exceeded 5.
- The `members` private member is also a regular `int` type, so added a ` < 0 ` check as a safeguard to prevent it from ever handling a negative number.

NOTE: This would be extremely difficult to test in practice by ones self.
The goal here is to break the party interface and get an extra player into the party to exceed the 5 member limit specified in the code.  If you successfully exceed >5 members, the error check logic would fail because it was previously only checking for 5.

1 - Login with 8 characters
2 - Fill the party with 5 members
3 - Send out pending invites on players 6-8
4 - Attempt to join Players 6-8 at the same time

NOTE: Even with a bugged party size, the error checking still kicks in.

_Side Note: This was an issue on Horizon that was observed and fixed.  This seemed to be possibly related to slight delays in communication across clustered map server processes on different IPs and the communication between them._

### DIFF
--- a/src/map/packets/party_effects.cpp
+++ b/src/map/packets/party_effects.cpp
@@ -24,23 +24,31 @@
 #include "party.h"
 #include "status_effect_container.h"
 
-CPartyEffectsPacket::CPartyEffectsPacket()
+void CPartyEffectsPacket::AddMemberEffects(std::size_t partyIndex, CCharEntity* PMember)
 {
-    this->setType(0x76);
-    this->setSize(0xF4);
-}
-
-void CPartyEffectsPacket::AddMemberEffects(CBattleEntity* PMember)
-{
-    if (members == 5)
+    // Check for valid party size to prevent buffer being overrun (244 bytes).
+    // When using multiple map processess across different IPs, the latency
+    // in communication combined with players joining/leaving at the same time
+    // can cause the party size to be larger than the packet size.
+    if (partyIndex >= 5)
     {
         ShowWarning("Member count exceeds packet maximum.");
         return;
     }
 
-    ref<uint32>(members * 0x30 + 0x04) = PMember->id;
-    ref<uint16>(members * 0x30 + 0x08) = PMember->targid;
-    ref<uint64>(members * 0x30 + 0x0C) = PMember->StatusEffectContainer->m_Flags;
-    std::memcpy(buffer_.data() + (members * 0x30 + 0x14), PMember->StatusEffectContainer->m_StatusIcons, 32);
-    ++members;
+    ref<uint32>(partyIndex * 0x30 + 0x04) = PMember->id;
+    ref<uint16>(partyIndex * 0x30 + 0x08) = PMember->targid;
+    ref<uint64>(partyIndex * 0x30 + 0x0C) = PMember->StatusEffectContainer->m_Flags;
+    std::memcpy(buffer_.data() + (partyIndex * 0x30 + 0x14), PMember->StatusEffectContainer->m_StatusIcons, 32);
+}
+
+CPartyEffectsPacket::CPartyEffectsPacket(const std::vector<CCharEntity*>& membersList)
+{
+    this->setType(0x76);
+    this->setSize(0xF4);
+
+    for (std::size_t idx = 0; idx < membersList.size(); ++idx)
+    {
+        AddMemberEffects(idx, membersList[idx]);
+    }
 }
--- a/src/map/packets/party_effects.h
+++ b/src/map/packets/party_effects.h
@@ -23,6 +23,7 @@
 #define _CPARTYEFFECTSPACKET_H
 
 #include "common/cbasetypes.h"
+#include "entities/charentity.h"
 
 #include "basic.h"
 
@@ -31,12 +32,9 @@ class CBattleEntity;
 class CPartyEffectsPacket : public CBasicPacket
 {
 public:
-    CPartyEffectsPacket();
-    void AddMemberEffects(CBattleEntity* PMember);
+    explicit CPartyEffectsPacket(const std::vector<CCharEntity*>& membersList);
+    void AddMemberEffects(std::size_t partyIndex, CCharEntity* PMember);
     void AddMemberEffects(uint32 id);
-
-private:
-    int members{ 0 };
 };
 
 #endif
--- a/src/map/party.cpp
+++ b/src/map/party.cpp
@@ -1226,8 +1226,9 @@ void CParty::PushEffectsPacket()
 
         for (auto& PMember : members)
         {
-            auto* PMemberChar = static_cast<CCharEntity*>(PMember);
-            auto  effects     = std::make_unique<CPartyEffectsPacket>();
+            auto*                     PMemberChar = static_cast<CCharEntity*>(PMember);
+            std::vector<CCharEntity*> sameZoneMembers;
+
             for (auto& memberinfo : info)
             {
                 if (memberinfo.partyid == m_PartyID && memberinfo.id != PMemberChar->id)
@@ -1235,11 +1236,13 @@ void CParty::PushEffectsPacket()
                     auto* PPartyMember = zoneutils::GetChar(memberinfo.id);
                     if (PPartyMember && PPartyMember->getZone() == PMemberChar->getZone())
                     {
-                        effects->AddMemberEffects(PPartyMember);
+                        sameZoneMembers.push_back(PPartyMember);
                     }
                 }
             }
-            PMemberChar->pushPacket(std::move(effects));
+
+            // Make and send packet for PMemberChar
+            PMemberChar->pushPacket<CPartyEffectsPacket>(sameZoneMembers);
         }
         m_EffectsChanged = false;
     }


src/map/packets/party_effects.cpp
```diff
@@ -32,7 +32,7 @@ CPartyEffectsPacket::CPartyEffectsPacket()
 
 void CPartyEffectsPacket::AddMemberEffects(CBattleEntity* PMember)
 {
-    if (members == 5)
+    if ((members < 0) || (members >= 5))
```
zach2good: Looking at how this packet is called/built, the zero condition is never possible:
```cpp
void CParty::PushEffectsPacket()
{
    if (m_EffectsChanged)
    {
        auto info = GetPartyInfo();

        for (auto& PMember : members)
        {
            auto* PMemberChar = static_cast<CCharEntity*>(PMember);
            auto  effects     = std::make_unique<CPartyEffectsPacket>();
            for (auto& memberinfo : info)
            {
                if (memberinfo.partyid == m_PartyID && memberinfo.id != PMemberChar->id)
                {
                    auto* PPartyMember = zoneutils::GetChar(memberinfo.id);
                    if (PPartyMember && PPartyMember->getZone() == PMemberChar->getZone())
                    {
                        effects->AddMemberEffects(PPartyMember);
                    }
                }
            }
            PMemberChar->pushPacket(std::move(effects));
        }
        m_EffectsChanged = false;
    }
}
```

We're iterating and incrementing - it can't be less than zero:
```cpp
void CPartyEffectsPacket::AddMemberEffects(CBattleEntity* PMember)
{
    if (members == 5)
    {
        ShowWarning("Member count exceeds packet maximum.");
        return;
    }

    ref<uint32>(members * 0x30 + 0x04) = PMember->id;
    ref<uint16>(members * 0x30 + 0x08) = PMember->targid;
    ref<uint64>(members * 0x30 + 0x0C) = PMember->StatusEffectContainer->m_Flags;
    std::memcpy(buffer_.data() + (members * 0x30 + 0x14), PMember->StatusEffectContainer->m_StatusIcons, 32);
    ++members;
}
```

So, just the >= condition or whatever

zach2good (commented):
How often is this logging present on HXI?
```
ShowWarning("Member count exceeds packet maximum.");
```

zach2good:
What would be ideal here, would be if you did some work towards this: https://github.com/LandSandBoat/server/issues/6717

Where this loop:
```cpp
for (auto& PMember : members)
        {
            auto* PMemberChar = static_cast<CCharEntity*>(PMember);
            auto  effects     = std::make_unique<CPartyEffectsPacket>();
            for (auto& memberinfo : info)
            {
                if (memberinfo.partyid == m_PartyID && memberinfo.id != PMemberChar->id)
                {
                    auto* PPartyMember = zoneutils::GetChar(memberinfo.id);
                    if (PPartyMember && PPartyMember->getZone() == PMemberChar->getZone())
                    {
                        effects->AddMemberEffects(PPartyMember);
                    }
                }
            }
            PMemberChar->pushPacket(std::move(effects));
        }
```

Just builds a `std::vector` of `CCharEntity*` with the "same zone" condition.

Then change the constructor of `CPartyEffectsPacket` to only take a `const std::vector<CCharEntity*>&`, and then it builds it completely.

Then after it's built, you can loop `members` again to send copies of the packet to whoever wants them.

Using `if (auto* PChar = dynamic_cast<...>(...))` where necessary, etc.

0x05010705:
Hypothetical Implementation Here -- 

Modified Header:
```cpp
#include "entities/charentity.h"
class CCharEntity;
explicit CPartyEffectsPacket(const std::vector<CCharEntity*>& membersList);
```
Modified Constructor:
```cpp
CPartyEffectsPacket::CPartyEffectsPacket(const std::vector<CCharEntity*>& membersList)
    : CPartyEffectsPacket()
{
    for (auto* member : membersList)
    {
        AddMemberEffects(member);
    }
}
```
Modified Looping
```cpp
void CParty::PushEffectsPacket()
{
    if (m_EffectsChanged)
    {
        auto info = GetPartyInfo();

        std::vector<CCharEntity*> sameZoneMembers;
        for (auto& memberinfo : info)
        {
            if (memberinfo.partyid == m_PartyID)
            {
                if (auto* PPartyMember = dynamic_cast<CCharEntity*>(zoneutils::GetChar(memberinfo.id)))
                {
                    if (PPartyMember->getZone() == dynamic_cast<CCharEntity*>(m_PLeader)->getZone())
                    {
                        sameZoneMembers.push_back(PPartyMember);
                    }
                }
            }
        }
        auto effects = std::make_unique<CPartyEffectsPacket>(sameZoneMembers);
        for (auto& PMember : members)
        {
            if (auto* PChar = dynamic_cast<CCharEntity*>(PMember))
            {
                PChar->pushPacket(std::make_unique<CPartyEffectsPacket>(*effects));
            }
        }
    }
}
```
What I am not sure on here is how best to isolate who is in what zone per se.  Right now, I am checking against the party leader, but what if the party leader is in one zone and 3 other members are in another zone.  Those three member *should* get effect updates for the members that are next to them, correct?

zach2good:
Look at where this is being called from:

```cpp
void CCharEntity::PostTick()
...

if m_EffectsChanged ... (for gating, thank god!)

        if (PParty)
        {
            PParty->PushEffectsPacket();
        }
```

```cpp
        auto info = GetPartyInfo(); // get "rough" party information from the db

        for (auto& PMember : members) // For each member of this current party object
        {
            for (auto& memberinfo : info) // For each entry in the "rough" party information
            {
                if (memberinfo.partyid == m_PartyID && memberinfo.id != PMemberChar->id) 
                {
                    auto* PPartyMember = zoneutils::GetChar(memberinfo.id);
                    if (PPartyMember && PPartyMember->getZone() == PMemberChar->getZone()) // If they're in the same zone
                    {
...
                    }
                }
            }
            PMemberChar->pushPacket(std::move(effects)); // Send packet
        }
...
    }
```

You don't need to isolate anything, when you're building `sameZoneMembers` you're already checking for same-zone-ness, the original logic didn't have anything to do with PLeader?

zach2good:
I'd imagine something like this:
```cpp
    auto info = GetPartyInfo(); // get "rough" party information from the db

    for (auto& PMember : members) // For each member of this current party object
    {
        auto* PMemberChar = static_cast<CCharEntity*>(PMember); // Current iterated member

        std::vector<CCharEntity*> sameZoneMembers;
        for (auto& memberinfo : info) // For each entry in the "rough" party information
        {
            if (memberinfo.partyid == m_PartyID && memberinfo.id != PMemberChar->id) // not self
            {
                auto* PPartyMember = zoneutils::GetChar(memberinfo.id); // other member as a char
                if (PPartyMember && PPartyMember->getZone() == PMemberChar->getZone()) // If they're in the same zone
                {
                    sameZoneMembers.push_back(PPartyMember);
                }
            }
        }

        // Make and send packet for PMemberChar
        PChar->pushPacket<CPartyEffectsPacket>(sameZoneMembers);
    }
```

This constructor is _almost_ right:
```cpp
CPartyEffectsPacket::CPartyEffectsPacket(const std::vector<CCharEntity*>& membersList)
    : CPartyEffectsPacket()
{
    for (auto* member : membersList)
    {
        AddMemberEffects(member);
    }
}
```

Like I mentioned here: https://github.com/LandSandBoat/server/issues/6717

We're trying to get rid of "state" inside packets:

https://en.cppreference.com/w/cpp/ranges/enumerate_view

```cpp
CPartyEffectsPacket::CPartyEffectsPacket(const std::vector<CCharEntity*>& membersList)
    : CPartyEffectsPacket()
{
    for (std::size_t idx = 0; idx < membersList.size(); ++idx)
    {
        AddMemberEffects(idx, members[idx]);
    }
}
```

Then that function changes:
```cpp
void CPartyEffectsPacket::AddMemberEffects(std::size_t partyIndex, CCharEntity* PMember)
```

zach2good:
And then your new proposed change would be:
```cpp
    if (memberIndex >= 5)
    {
        ShowWarning("Member index exceeds packet maximum.");
        return;
    }
```

(you don't need to check for negative, since it's a size_t, and you're iterating forward-only

0x05010705:
That makes a hell of a lot more sense.  I wasn't thinking about the context of the call origination (and it was also very late when I was looking it over).  Very much appreciate the conversation on this and working on putting this together now.

0x05010705:
Tested the update with just two characters and things seem to be working.
Breakpoints on the new constructor seem to work just fine:
```
-- CharacterB used ability to add an effect (m_effectsChanges == true)
-- Update is sent to CharacterA in the same zone
[03/27/25 18:55:58:251][map][info] Creating party effects packet for 1 members. (CPartyEffectsPacket::CPartyEffectsPacket:51)
[03/27/25 18:55:58:251][map][info] Sending Packet to: Beasty (ID: 1) (CPartyEffectsPacket::CPartyEffectsPacket:54)

-- Tested again with reverse characters (CharacterA updated effect, CharacterB received update)
[03/27/25 18:55:58:251][map][info] Creating party effects packet for 1 members. (CPartyEffectsPacket::CPartyEffectsPacket:51)
[03/27/25 18:55:58:251][map][info] Sending Packet to: Beastwo (ID: 2) (CPartyEffectsPacket::CPartyEffectsPacket:54)

-- Tested again with CharacterA in another zone
-- The constructor is still hit
[03/27/25 19:00:04:615][map][info] Creating party effects packet for 0 members. (CPartyEffectsPacket::CPartyEffectsPacket:51)
```
```cpp
CPartyEffectsPacket::CPartyEffectsPacket(const std::vector<CCharEntity*>& membersList)
    : CPartyEffectsPacket()
{
    ShowInfo("Creating party effects packet for %u members.", membersList.size());
    for (std::size_t idx = 0; idx < membersList.size(); ++idx)
    {
        ShowInfo("Sending Packet to: %s (ID: %u)", membersList[idx]->name.c_str(), membersList[idx]->id);
        AddMemberEffects(idx, membersList[idx]);
    }
}
```
Unless I should alter `PushEffectsPacket` to not actually call `pushPacket()` without the `sameZoneMembers.size() > 0`
```cpp
void CParty::PushEffectsPacket()
{
    if (m_EffectsChanged)
    {
        auto info = GetPartyInfo();

        for (auto& PMember : members)
        {
            auto* PMemberChar = static_cast<CCharEntity*>(PMember);
            std::vector<CCharEntity*> sameZoneMembers;

            for (auto& memberinfo : info)
            {
                if (memberinfo.partyid == m_PartyID && memberinfo.id != PMemberChar->id)
                {
                    auto* PPartyMember = zoneutils::GetChar(memberinfo.id);
                    if (PPartyMember && PPartyMember->getZone() == PMemberChar->getZone())
                    {
                        sameZoneMembers.push_back(PPartyMember);
                    }
                }
            }

            // Make and send packet for PMemberChar
            PMemberChar->pushPacket<CPartyEffectsPacket>(sameZoneMembers);
        }
        m_EffectsChanged = false;
    }
}
```

zach2good:
I think it'll also be useful to leave behind a comment saying _why_ you're checking for an invalid party size (latency, etc.)

src/map/packets/party_effects.h
```diff
@@ -32,11 +33,9 @@ class CPartyEffectsPacket : public CBasicPacket
 {
 public:
     CPartyEffectsPacket();
```
zach2good: We can get rid of the default constructor now

zach2good (approved):

src/map/packets/party_effects.h
0x05010705: - Removed the old default constructor
- Moved `type` and `size` values into the new constructor

zach2good (approved):

PR #7319, opened 2025-03-28
login: Fix MariaDB PASSWORD()<>BCrypt conversion case

sruon:
**_I affirm:_**

Fix the prepared statement used to get PASSWORD() hash, in cases where we're looking to convert an existing MariaDB PASSWORD() password into a BCrypt hash.

The statement is converting to "SELECT 'PASSWORD(\'xyz\')'" currently, which causes the login to fail for accounts without a bcrypt hash.

There is another block a little further in the code that does the same thing, this PR aligns on the same code.

- Attempt to login with an account using a password created with PASSWOR()
- Login succeeds and the row is updated with a bcrypt hash

### DIFF
--- a/src/login/auth_session.cpp
+++ b/src/login/auth_session.cpp
@@ -179,26 +179,23 @@ void auth_session::read_func()
             {
                 // It's not a BCrypt hash, so we need to use Maria's PASSWORD() to check if the password is actually correct,
                 // and then update the password to a BCrypt hash.
-                const auto passColumn = fmt::sprintf("PASSWORD('%s')", password);
-                const auto rset       = db::preparedStmt("SELECT ?", passColumn);
+                const auto rset = db::preparedStmt("SELECT PASSWORD(?)", password);
                 if (rset && rset->rowsCount() != 0 && rset->next())
                 {
-                    if (rset->get<std::string>(passColumn) != passHash)
+                    if (rset->get<std::string>(0) != passHash)
                     {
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
shouldnt all accounts have a bcrypt hash in the first place?

zach2good (approved):
I hope there aren't people who haven't migrated their PWs to the bcrypt hashes still ;_;

EDIT: It's now been 13 months since those changes went live, god damn people need to update their stuff

sruon:
Not naming names but some are still creating with PASSWORD() from their website and then relying on the first login to convert. I would drop this feature entirely from the login process, but that's for another day.

zach2good:
> still creating with PASSWORD from their website and then relying on the first login to convert

image

I did have a time-gated bit of code in there that would stop compiling this year if the `PASSWORD` stuff wasn't removed, but it got taken out at some point. I really don't want to support this, like, at all.

EDIT: Here's that PR that removed the static_assert: https://github.com/LandSandBoat/server/pull/6625

zach2good:
It's so easy to just... use bcrypt:

https://pypi.org/project/bcrypt/
https://www.php.net/manual/en/function.password-hash.php
https://www.npmjs.com/package/bcrypt

I'm considering just ripping this out. We used to get shit for continuing to use `PASSWORD` because it wasn't best practice, and now people are falling back to using the old shit for _reasons_ 🤷

PR #7322, opened 2025-03-28
DB: Add rowsAffected() function and use in PacketSystem

zach2good:
**_I affirm:_**

- Make charutils entries for some things in SmallPacket0x00A
- Adopt db in a few more places
- Remove dead code
- Update SmallPacket0x01B with decomp structures
- Scope Creep: Added rowsAffected() to DB, used everywhere

## Testing

- Log in
- Zone
- Zone (multi process)
- Chat (multi process)
- Basic modules/override works
- Create/Name linkshell, equip and use
- Set Bazaar message, zone, check it
- Set Search message, zone, check it (/sea, etc.)
- Place furniture in Moghouse
- Equip a mannequin
- Zone back in
- Blacklist still works
- Delivery box still works
- AH still works
- Search server still works
- Session wiping (w/ KillSession, or whatever)


zach2good:
Tested and ready to go

Xaver-DaRed (approved):

PR #7327, opened 2025-03-30
Update CSPositionPacket

### LINKED ISSUE #7325: 🐛 BCNMs concurrency issue / black screen / multi-areas not working
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce
2 chars, not grouped.

Char A enters a BCNM (let's say "Up in Arms" in Waughroon Shrine, but as far as I can tell this affects _all_ BCNMs/ENMs). Char A receives the correct event update (0x05C) from the lua (_CUTSCENE return code), gets warped in the BCNM and can complete it with no issues.

Char B attempts to enter _any_ BCNM in the same zone (while A is still progressing their own). Char B client initiates the fade to black but never receives any follow-up 0x05C event update, leading to an infinite black screen. 

Looking at it with a debugger shows `registerBattlefield` correctly returning `INCREMENT_REQUEST`, but when it gets back to `battlefield.lua`, it just returns 0 and does not send any event update? 
Not sure how that's supposed to work.

https://github.com/LandSandBoat/server/blob/d2018a7caa078f3fe41780e415b8247197bbdf87/scripts/globals/battlefield.lua#L814-L831

!pos -342.69 104 -259 144
!additem moon_orb

## Expected behavior
Char B receives correct event update and gets warped in the next BCNM "area"

cocosolos:
CSPositionPacket is identical to PositionPacket.
Update and fix packet handling for incoming packet 0x05C (event update with position).

**_I affirm:_**

Follows up on #6902 by updating [the other position packet, 0x065](https://github.com/atom0s/XiPackets/blob/main/world/server/0x0065/README.md). Basically a copy/paste of CPositionPacket.

This capture was helpful in figuring out how 0x065 is used:
https://drive.google.com/file/d/12zdCSIEmHCw47s6G4DF_GsYdOSZkox6I/view

When selecting a battlefield, the client sends a [warp request (C2S 0x05C)](https://github.com/atom0s/XiPackets/blob/main/world/client/0x005C/README.md). If the server says the request is denied (like when a battlefield arena is occupied), it sends 0x065 with mode 2 to the client without changing the player position. The client then sends a warp request for the next arena and so on and if all arenas are full the cutscene ends without any actual position changes. If a request succeeds the server sends 0x065 with mode 1, and a matching 0x05B with mode 0 (a normal position update).

Fixes #7325

SmallPacket0x05E should behave as it did previously, with the exception that it now calls `PChar->setLocked(false)` as part of the mode 5 (reset).

```
!pos -342.69 104 -259 144
!additem moon_orb
```
Do this with 4 characters, all solo.
Have the first 2 characters enter any BCNM.
With the last 2 characters, trade the orb to the entrance so both characters have the BCNM selection menu open.
Attempt to enter a BCNM with both characters.

One of the last 2 characters should be denied entry, and the screen should fade back in. This character will still be in the lobby, while the other 3 should be in different arenas.

### DIFF
--- a/src/map/packet_system.cpp
+++ b/src/map/packet_system.cpp
@@ -2759,6 +2759,20 @@ void SmallPacket0x05A(MapSession* const PSession, CCharEntity* const PChar, CBas
  *                                                                       *
  ************************************************************************/
 
+// https://github.com/atom0s/XiPackets/blob/main/world/client/0x005B/README.md
+struct GP_CLI_EVENTEND
+{
+    uint16_t id : 9;
+    uint16_t size : 7;
+    uint16_t sync;
+    uint32_t UniqueNo;  // PS2: UniqueNo
+    uint32_t EndPara;   // PS2: EndPara
+    uint16_t ActIndex;  // PS2: ActIndex
+    uint16_t Mode;      // PS2: Mode
+    uint16_t EventNum;  // PS2: EventNum
+    uint16_t EventPara; // PS2: EventPara
+};
+
 void SmallPacket0x05B(MapSession* const PSession, CCharEntity* const PChar, CBasicPacket& data)
 {
     TracyZoneScoped;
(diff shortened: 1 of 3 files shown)


zach2good (approved):
Looks good 👍 

If you're using atom0s's decomp structures, can you add this to the bottom of your commit message?

```
Co-Authored-By: atom0s <atom0s@live.com>
```

PR #7329, opened 2025-03-30
[core] Refactor seals/geodes drops

sruon:
**_I affirm:_**

Refactor the way seals and geodes are added to the pool so it's slightly less of a mess.
- Extracted the logic to get list of eligible seals/geodes in their own methods
- Made the cooldown logic take an ID so we can apply the logic equally to seals and geodes
- Used the right enums for weathers, elements. 
  - Added some constexpr for the items themselves, I don't know if we ever intend to list all the items in an enum at some point.

Fixes:
- In my previous PR, I wasn't adding seals cooldown in all the relevant places, leading to low level mobs dropping seals without cooldown

Behavior changes:
- HKC will only drop from mobs above 80 (instead of 75). [ffo.jp](https://wiki-ffo-jp.translate.goog/html/22307.html?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp) and [bg-wiki](https://www.bg-wiki.com/ffxi/H._Kindred_Crest) both state 80, ffxiclopedia is the only source claiming 75.
- Only one random roll to decide if a geode/avatarite should be added, instead of 2. Similar to how seals work.
- Geodes/Avatarites will follow same rules as seals and be subject to a 5 minutes cooldown
  - Basing this off [Siknoz findings](https://discord.com/channels/392903136336936960/883227930828886060/1348760319874437208) and my own limited retail testing

Kill mobs of varying levels, see seals/geodes added to pool. Kill more during the 5 minutes window and see they don't get added.


src/map/entities/mobentity.cpp
```diff
@@ -683,30 +683,183 @@ void CMobEntity::DistributeRewards()
...
+// - Mob >= 50: Beastmen's Seal, Kindred's Seal
+// - Mob >= 70: Beastmen's Seal, Kindred's Seal, Kindred's Crest
+// - Mob >= 80: Beastmen's Seal, Kindred's Seal, Kindred's Crest, High Kindred's Crest
+// If Abyssea is not enabled, pool is limited to Beastmen's Seal and Kindred's Seal.
+auto CMobEntity::GetEligibleSeals() -> std::vector<uint16>
+{
+    constexpr uint16 BEASTMENS_SEAL      = 1126;
+    constexpr uint16 KINDREDS_SEAL       = 1127;
+    constexpr uint16 KINDREDS_CREST      = 2955;
+    constexpr uint16 HIGH_KINDREDS_CREST = 2956;
```
zach2good: You can extract these constants into an anon namespace at the top of the cpp file:
```cpp
namespace
{
    constexpr uint16 BEASTMENS_SEAL      = 1126;
    constexpr uint16 KINDREDS_SEAL       = 1127;
    constexpr uint16 KINDREDS_CREST      = 2955;
    constexpr uint16 HIGH_KINDREDS_CREST = 2956;
}
```

src/map/entities/mobentity.cpp
```diff
@@ -683,30 +683,183 @@ void CMobEntity::DistributeRewards()
...
+    return { BEASTMENS_SEAL };
+}
+
+// Return the list of Geode and Avatarites that can drop based on the mob's level.
+// Rules:
+// - Mob >= 50: Geodes of matching weather/day can drop. Weather takes priority.
+// - Mob >= 80: Avatarites of matching weather/day can also drop. Weather takes priority.
+auto CMobEntity::GetEligibleGeodes() -> std::vector<uint16>
+{
+    constexpr uint16 FLAME_GEODE   = 3297;
```
zach2good: Same with these

src/map/entities/mobentity.cpp
```diff
@@ -683,30 +683,183 @@ void CMobEntity::DistributeRewards()
...
+    }
+
+    return {};
+}
+
 void CMobEntity::DropItems(CCharEntity* PChar)
 {
     TracyZoneScoped;
+
+    constexpr int RECAST_SEAL           = 1;
```
zach2good: These too

src/map/entities/mobentity.cpp
```diff
@@ -683,30 +683,183 @@ void CMobEntity::DistributeRewards()
...
+    constexpr uint16 SHIVITE   = 3521;
+    constexpr uint16 GARUDITE  = 3522;
+    constexpr uint16 TITANITE  = 3523;
+    constexpr uint16 RAMUITE   = 3524;
+    constexpr uint16 LEVIATITE = 3525;
+    constexpr uint16 CARBITE   = 3526;
+    constexpr uint16 FENRITE   = 3527;
+
+    // clang-format off
+    std::map<uint8, uint16> geodeMap = {
```
zach2good: These maps are being rebuilt every time `GetEligibleGeodes` is called, so move them up into that other namespace

zach2good (approved):
Constants and lookups need moving out of the function calls, but otherwise looks good 👍

src/map/entities/mobentity.cpp
Xaver-DaRed: scr/map/items.h

I think we could add those items to the enum list and use that

src/map/entities/mobentity.cpp
```diff
@@ -683,30 +683,183 @@ void CMobEntity::DistributeRewards()
...
+    constexpr uint16 SHADOW_GEODE  = 3304;
+
+    constexpr uint16 IFRITITE  = 3520;
+    constexpr uint16 SHIVITE   = 3521;
+    constexpr uint16 GARUDITE  = 3522;
+    constexpr uint16 TITANITE  = 3523;
+    constexpr uint16 RAMUITE   = 3524;
+    constexpr uint16 LEVIATITE = 3525;
+    constexpr uint16 CARBITE   = 3526;
+    constexpr uint16 FENRITE   = 3527;
```
Xaver-DaRed: Same here as my previous comment

src/map/entities/mobentity.cpp
sruon: don't know how I missed it!

sruon:
Moved item IDs to the ITEMID enum (and reordered by ID while I was at it)
Moved other constants to an anonymous namespace

Xaver-DaRed (approved):

PR #6980, opened 2025-02-07
[Salvage][sql] Spawnpoint, Groups, Drops Zhaylom Remnants

KnowOne134:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my PR will be ignored.
- [x] I understand I should leave resolving conversations to the LandSandBoat team so that reviewers won't miss what was said.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have _**tested my code and the things my code has changed**_ since the last commit in the PR and will test after any later commits.

## What does this pull request do?

updates the database portion for zone Zhayolm Remnants
mob spawn points
mob groups
mob droplist

## Steps to test these changes

database builds

PR #7330, opened 2025-03-30
[SQL] Adjust Bhaflau Remnants Drops & Add Zhayolm Comments

LacunaXI:
**_I affirm:_**

Bhaflau Remnants:
- I noticed that group drop rates where inversed with individual item weights and created a PR to address this.
- @KnowOne134 please confirm if this is the case.
- Stuck a TODO on the Reactionary Rampart after checking all Bhaflau entries in the file. It looks as if it is a lone straggler.

Zhayolm Remnants:
- Added a TODO comment on the two drops that have listeners planned for them in the future according to #6980
- Adding comments is tedious work. I added these despite seeing they will be changing in the hopes that it slightly reduces future workload. It is easier to change comments or visually reference them while working than to search for and add them.

### DIFF
--- a/sql/mob_droplist.sql
+++ b/sql/mob_droplist.sql
@@ -1389,11 +1389,11 @@ INSERT INTO `mob_droplist` VALUES (151,0,0,1000,15718,@UNCOMMON); -- Freyas Lede
 -- ZoneID:  75 - Troll_Lapidarist
 -- ZoneID:  75 - Troll Ironworker
 -- ZoneID:  75 - Black Pudding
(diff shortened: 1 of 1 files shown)


Xaver-DaRed:
Putting this on hold until KO answers

KnowOne134 (approved):
looks good to me. Thanks for the assist on this, sql data entry is not something i take pleasure in doing.
looks like i didnt understand grouping layout, way different than eden codewise

Xaver-DaRed (approved):

LacunaXI:
> looks good to me. Thanks for the assist on this, sql data entry is not something i take pleasure in doing. looks like i didnt understand grouping layout, way different than eden codewise

You are welcome. I can not say it is something I take pleasure in either. Just so happen to be numb to it due to the daily SQL grind for so long.

Thank you for submitting this work.

PR #7333, opened 2025-03-30
Fix MDT-3 for Skormoth Mask

Harfainx:
Magic damage taken -3%. The value was "3". Corrected this to "-300", which is -3%. This is how all other MDT- gear is. It's currently giving MDT+0.03%.

**_I affirm:_**

Correct Skormoth Mask MDT to -3%. Currently gives +0.03%.

Check and see that now it's giving 0, and with this change it will give -3%. It's a single value SQL fix. No real steps needed to test.


sql/item_mods.sql
```diff
@@ -75051,7 +75051,7 @@ INSERT INTO `item_mods` VALUES (27783,25,26);   -- ACC: 26
 INSERT INTO `item_mods` VALUES (27783,29,4);    -- MDEF: 4
 INSERT INTO `item_mods` VALUES (27783,31,43);   -- MEVA: 43
 INSERT INTO `item_mods` VALUES (27783,68,38);   -- EVA: 38
-INSERT INTO `item_mods` VALUES (27783,163,3);   -- DMGMAGIC: 3
+INSERT INTO `item_mods` VALUES (27783,163,-300);   -- DMGMAGIC: 3
```
Xaver-DaRed, line 75054: Fix the comment aswell

Xaver-DaRed (approved):

Issue #3911, opened 2023-05-22
🔨 mobs can have multiple despoil items, with different rarities


TeoTwawki:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature
Behold Exhibit A, the Lagoon Sahagin.
Out of 50 despoil attempts I got: 
 - 9 Quus
 - 1 Coral Butterfly
 - A metric ton of TP over the course of my time spent doing this thanks to Job Points

Other notes:
 - 9 to 1 is a pretty significant difference
 - by 50 attempts, I mean 50 sahagins, I did try to despoil same sahagin more than once, but didn't count the same sahagin more than once
 - not a single time did I ever get both items from any of them. pretty sure you can not. I don't have a big enough sample to prove it though.
 - screenshots showing the both quus and the coral butterfly are available, but not in the same screen because of log spam
 - I clearly have to much time on my hands on Sundays
 - I need to start using +despoil gear (I had none on me for this trip)

And then my crab didn't evolve so I get to do this again ;.; (I have to feed it coral butterfly to evolve it)

So in closing, I now have fairly good albeit anecdotal evidence that its not an equal shot at these items, and rock solid proof that mobs can have multiple items in their pool. 

And I am very confident that steal works exactly the same way, and is likely not solely left up to a level check - level corrected sure, but I've noted very different success rates on different items in the same approximate level range. Anyone who's done plat beastcoins in movopolis should know what I am talking about here. At the least, we should be prepared for the scenario. Its pretty tough to definitively prove without massive samples.

PR #7339, opened 2025-03-31
[core] Randomize despoiled item

sruon:
**_I affirm:_**

Randomize between all possible despoilable items. Certain mobs have more than 1.

!zone Boyahda Tree
!gotoname Bark_Spider
Use "Despoil", obtain randomly "Spider Web" or "Rainbow Thread"

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -18283,34 +18283,37 @@ uint16 CLuaBaseEntity::getStealItem()
 
 /************************************************************************
  *  Function: getDespoilItem()
- *  Purpose : Used to return the Item ID of a mob's item which can be despoiled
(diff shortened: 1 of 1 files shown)


Xaver-DaRed (approved):

sruon:
Somewhat related to #3911 but we don't have item-specific despoil rates in the database at the moment.

LacunaXI:
FWIW I spent time waiting for the despoil timer for hours on different mobs. I think despoil items may have some difficulty or weight and are not random. This PR is still an improvement in the meantime.

It requires further study but more valuable items seem to be more difficult to pull and Despoil will fail repeatedly. If there are multiple options you seem to more often get a subjectively less valuable rewards despoiled instead. A Spider Web more often than a Rainbow Thread, etc.

PR #7354, opened 2025-04-01
[core] Fix crash on Fishing Bait Loss

0x05010705:
Fixes crash on hookedFish in BaitLoss where hookedFish is deref'd before checking for null

**_I affirm:_**

Fixes a crash in `BaitLoss()`
- `PChar->hookedFish` was not checked for null before being accessed in the `PChar->hookedFish->successType` comparison, leading to a crash.

1 - Go fishing with bait
2 - Try and trigger the following messages related to stamina checking:
```cpp
case FISHACTION_FINISH:
    ...
    else if (stamina <= 0x14)  // Lack of Skill!
                {
                    // you lost the catch due to lack of skill
                    // lose bait but keep lure
                    PChar->animation = ANIMATION_FISHING_LINE_BREAK;
                    BaitLoss(PChar, false, true);
                    PChar->pushPacket<CMessageTextPacket>(PChar, MessageOffset + FISHMESSAGEOFFSET_LOST_LOWSKILL);

                    if (PChar->hookedFish)
                    {
                        PChar->hookedFish->successtype = FISHINGSUCCESSTYPE_NONE;
                    }
                }
    else if (stamina <= 0x64)  // Line Breaks!
        ...
    else if (stamina <= 0x100) // You give up!
        ...
    else // implied (stamina > 0x100) // Lost catch!
        ...
```
In the above cases, `BaitLoss()` was called before the `if(PChar->hookedFish)` checks
3 - Repeat the steps above to check the remaining cases

NOTE: Under any of the above conditions, no crash is experienced when `BaitLoss()` is called.

### DIFF
--- a/src/map/utils/fishingutils.cpp
+++ b/src/map/utils/fishingutils.cpp
@@ -1361,17 +1361,24 @@ namespace fishingutils
                 return false;
             }
 
-            if (PChar->hookedFish->successtype != FISHINGSUCCESSTYPE_CATCHITEM)
+            if (PChar->hookedFish == nullptr)
             {
-                if (PBait->getQuantity() == 1)
+                ShowWarning("PChar->hookedFish was null.");
+            }
+            else
+            {
+                if (PChar->hookedFish->successtype != FISHINGSUCCESSTYPE_CATCHITEM)
                 {
-                    charutils::UnequipItem(PChar, SLOT_AMMO, false);
-                }
-                charutils::UpdateItem(PChar, PBait->getLocationID(), PBait->getSlotID(), -1);
+                    if (PBait->getQuantity() == 1)
+                    {
+                        charutils::UnequipItem(PChar, SLOT_AMMO, false);
+                    }
+                    charutils::UpdateItem(PChar, PBait->getLocationID(), PBait->getSlotID(), -1);
 
-                if (SendUpdate)
-                {
-                    PChar->pushPacket<CInventoryFinishPacket>();
+                    if (SendUpdate)
+                    {
+                        PChar->pushPacket<CInventoryFinishPacket>();
+                    }
                 }
             }
         }


zach2good (approved):

src/map/utils/fishingutils.cpp
```diff
@@ -1342,7 +1342,7 @@ namespace fishingutils
     {
         CItemWeapon* PBait = dynamic_cast<CItemWeapon*>(PChar->getEquip(SLOT_AMMO));
 
-        if (PBait != nullptr)
+        if (PBait != nullptr && PChar->hookedFish != nullptr)
```
Xaver-DaRed: Shouldnt this check go in line 1364 a little down bellow?

0x05010705: The implications of the change would be the same.  In both cases, where `PChar->hookedFish` is null, it would still return `true`.
I figured it made more sense to do both `nullptr` checks in the same place though.

Xaver-DaRed: Not really. There are checks down there with warnings returned. If hookedfish happened to be null, it would eat those warnings (without giving a warning either) if, for instance, the bait happened to also not be a weapon.

The reason that null check is there is because all the checks inside use `PBait`. So I believe the null check should be placed bellow and maybe even adding a warning aswell

Xaver-DaRed: So TLDR, while it really doesnt matter in the end, as you said, I'd like to at least have a warning returned for it at the very least

0x05010705: Yea, from the logging standpoint it definitely makes sense.  Updated.

zach2good:
Unrelated to this PR: CI is failing because of our CI runners upgrading and our recast dep being left behind, if you rebase after this PR merges the CI failures will go away: <https://github.com/LandSandBoat/server/pull/7359>

Xaver-DaRed (approved):

PR #7360, opened 2025-04-03
[core] Extend flexibility of Dynamic Entities

0x05010705:
Adds in optional table values to modify mob level, drops, skills, spells, respawn, and timed spawns.

**_I affirm:_**

Extends the flexibility of Dynamic Entities for `TYPE_MOB` to allow for modifying the following:
- Minimum Level
- Maximum Level
- Drop ID
- Skill List
- Spell List
- Respawn Times
- Spawn Type (Normal, Night, Lottery, Windowed, Scripted, etc..)

1 - Use the following script to create a command to spawn a DE Mob:
_NOTE: This script is crude and strictly for testing this PR_
```lua
---@type TCommand
local commandObj = {}

commandObj.cmdprops =
{
    permission = 1,
    parameters = ""
}

commandObj.onTrigger = function(player)
    local zone = player:getZone()

    if not zone then
        return
    end

    local mob = zone:insertDynamicEntity({
        objtype = xi.objType.MOB,
        name = "Test DE Mob",
        x = player:getXPos(),
        y = player:getYPos(),
        z = player:getZPos(),
        rotation = player:getRotPos(),
        groupId = 1,
        groupZoneId = zone:getID(),
        look = 357,
        minLevel = 10,  -- Set Min Level
        maxLevel = 20,  -- Set Max Level
        dropId = 805,   -- Modify to Fafnir Drop Table
        skillList = 48, -- Set custom skills (Bee)
        spellList = 79, -- Set custom Spell List (King Arthro)
        respawn = 60,   -- 60 Second Respawn
        spawnType = 0,  -- SPAWNTYPE_NORMAL (00:00 - 23:59)
    })

    if not mob then
        return
    end

    mob:setSpawn(player:getXPos(), player:getYPos(), player:getZPos(), player:getRotPos())
    mob:spawn()
end

return commandObj
```
2 - Use the new command to trigger a DE spawn with the added settings
3 - Check the Following Conditions:
- Mob Level (Variable between 10 to 20)
- Mob Drops on Death (Set to Fafnir Drops)
- Mob Spells (Should be Water Spells)
- Mob Skills (Should be Bee TP Moves)  _..and enjoy the distorted animations of a bee on a crab.._
- Check Respawn (Should be 1 Minute)

NOTE: You are now able to further modify Dynamic Entities from the Table Data when spawning a DE.

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -94,6 +94,7 @@
 #include "los/zone_los.h"
 #include "map_networking.h"
 #include "map_server.h"
+#include "mob_modifier.h"
 #include "mobskill.h"
 #include "monstrosity.h"
 #include "navmesh.h"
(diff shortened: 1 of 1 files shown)


src/map/lua/luautils.cpp
```diff
@@ -5496,6 +5497,56 @@ namespace luautils
                 }
             }
 
+            // Allow for overrides of defaults on TYPE_MOB
+            // If no value is specified, mob_groups.sql values are used
+            auto minLevel = table["minLevel"].get_or<uint8>(0);
```
zach2good: If these values aren't being changed after assignment, they should all be `const auto`

0x05010705: Updated the additions to `const auto`

zach2good (approved):

Xaver-DaRed (approved):

PR #7362, opened 2025-04-03
[core] Fix crash in Latent Effect Container

0x05010705:
Adds a null check on m_POwner and removes the implicit casts that was happening.

**_I affirm:_**

Prevents a crash in latent effect container that can happen if m_POwner becomes null (during zoning, logout, etc)
- Adds `nullptr` check before before triggering latents.
- Removes implicit casts on `m_POwner`

1 - Login with a two characters
2 - Equip one character with a weather-based latent effect
3 - Move both characters to an outside zone
4 - Start logging out with one character
5 - Right as the first character is about to logout, use `!setweather` to change the weather to whatever matches the latent
NOTE: The map process no longer crashes trying to call `ApplyLatentEffect()`

### DIFF
--- a/src/map/latent_effect_container.cpp
+++ b/src/map/latent_effect_container.cpp
@@ -659,15 +659,18 @@ void CLatentEffectContainer::CheckLatentsWeather()
 void CLatentEffectContainer::CheckLatentsWeather(uint16 weather)
 {
     ProcessLatentEffects([this, weather](CLatentEffect& latent) {
-        if (latent.GetConditionsID() == LATENT::WEATHER_ELEMENT)
+        if (auto owner = dynamic_cast<CBattleEntity*>(m_POwner))
         {
-            auto element = zoneutils::GetWeatherElement(battleutils::GetWeather((CBattleEntity*)m_POwner, false, weather));
-            return ApplyLatentEffect(latent, latent.GetConditionsValue() == element);
-        }
-        else if (latent.GetConditionsID() == LATENT::WEATHER_CONDITION)
-        {
-            auto element = battleutils::GetWeather((CBattleEntity*)m_POwner, false, weather);
-            return ApplyLatentEffect(latent, latent.GetConditionsValue() == element);
+            if (latent.GetConditionsID() == LATENT::WEATHER_ELEMENT)
+            {
+                const auto element = zoneutils::GetWeatherElement(battleutils::GetWeather(owner, false, weather));
+                return ApplyLatentEffect(latent, latent.GetConditionsValue() == element);
+            }
+            else if (latent.GetConditionsID() == LATENT::WEATHER_CONDITION)
+            {
+                const auto element = battleutils::GetWeather(owner, false, weather);
+                return ApplyLatentEffect(latent, latent.GetConditionsValue() == element);
+            }
         }
         return false;
     });


Xaver-DaRed (approved):
lgtm

src/map/latent_effect_container.cpp
```diff
@@ -659,15 +659,19 @@ void CLatentEffectContainer::CheckLatentsWeather()
 void CLatentEffectContainer::CheckLatentsWeather(uint16 weather)
 {
     ProcessLatentEffects([this, weather](CLatentEffect& latent) {
-        if (latent.GetConditionsID() == LATENT::WEATHER_ELEMENT)
+        auto owner = dynamic_cast<CBattleEntity*>(m_POwner);
+        if (owner != nullptr)
```
zach2good: if (auto POwner = dynamic_cast<CBattleEntity*>(m_POwner))

src/map/latent_effect_container.cpp
```diff
@@ -659,15 +659,19 @@ void CLatentEffectContainer::CheckLatentsWeather()
...
-            auto element = zoneutils::GetWeatherElement(battleutils::GetWeather((CBattleEntity*)m_POwner, false, weather));
-            return ApplyLatentEffect(latent, latent.GetConditionsValue() == element);
-        }
-        else if (latent.GetConditionsID() == LATENT::WEATHER_CONDITION)
-        {
-            auto element = battleutils::GetWeather((CBattleEntity*)m_POwner, false, weather);
-            return ApplyLatentEffect(latent, latent.GetConditionsValue() == element);
+            if (latent.GetConditionsID() == LATENT::WEATHER_ELEMENT)
+            {
+                auto element = zoneutils::GetWeatherElement(battleutils::GetWeather(owner, false, weather));
```
zach2good: const auto element

src/map/latent_effect_container.cpp
```diff
@@ -659,15 +659,19 @@ void CLatentEffectContainer::CheckLatentsWeather()
...
+                return ApplyLatentEffect(latent, latent.GetConditionsValue() == element);
+            }
+            else if (latent.GetConditionsID() == LATENT::WEATHER_CONDITION)
+            {
+                auto element = battleutils::GetWeather(owner, false, weather);
```
zach2good: const auto element

zach2good (changes requested):
More important than my styling suggestions - you have repro steps for this. This is a symptom of a character being null during iteration. _It absolutely should not be in the entity collections if it is null! 

```cpp
    FOR_EACH_PAIR_CAST_SECOND(CCharEntity*, PCurrentChar, m_charList)
    {
        PCurrentChar->PLatentEffectContainer->CheckLatentsWeather(weather);
        PCurrentChar->PAI->EventHandler.triggerListener("WEATHER_CHANGE", CLuaBaseEntity(PCurrentChar), static_cast<int>(weather), element);
    }
```

`PCurrentChar` should not be null here. This fix, while I'm sure it fixes this specific crash, is covering up a condition that is normally hard to replicate, so there's the opportunity here to hunt down and fix the fundamentally dangerous underlying issue - which this PR _doesnt_ do.

src/map/latent_effect_container.cpp
zach2good: As a thought process to try and take on board:

If there are two classes like `Entity` and `ContainerOwnedByEntity`, where `ContainerOwnedByEntity` is constructed with a reference or a pointer back to the entity that owns it (like `m_POwner`) - there is no scenario where that container should exist or have its functions called where `m_POwner` isn't valid. We have deep deep lifetime issues if this is the case, and I think it's _extremely unsafe_ to try and sweep it under the rug, even if it's a pain in the ass to try and track down the root cause.

0x05010705:
For now, I have added the styling suggestions you provided.

I am spending some time to try and actually catch this happening on my local but will take some time to actually capture the issue since it is definitely an edge case.  Currently trying to test a "race condition" scenario:

1 - Player is in a state of disconnecting (crash, logout, etc) --> Character is not de-allocated from the zone/map for ~30s
2 - As that disconnect is happening, the weather changes --> Causes the iterator to start up and process weather latents
3 - Player that is in state of disconnection is at the end of the iterator of a busy zone (the hard part to test)
4 - Player session is ended before the iterator has reached them and their Char object is now de-allocated (so now `m_CharList` potentially has a dangling pointer).
5 - Iterator reaches the de-allocated char and then tried to call `PChar->PLatentEffectContainer->CheckLatentsWeather(weather);` which is now triggering the crash.

0x05010705:
I stopped using a an Organization to maintain my fork (since I cant have two forks from the same parent repo).
Just realized that this was closed by proxy of that action since I re-created a fork on my profile.

I will re-submit this PR once I have additional information on this and reference this PR.
Just posting in here to say that I am not abandoning this fix/search for root cause. D:

PR #7376, opened 2025-04-07
Implement Todays Horoscope KSNM 30

RaraProjects:
**_I affirm:_**

Implements a largely fleshed out (but still experimental) version of the KSNM 30 Today's Horoscope.

Details and capture info can be found here:
https://docs.google.com/spreadsheets/d/1LziL_LFqF-04Xw6OHWh-cE9UnIcJ8R4KuUemWu_HqTE/edit?usp=sharing

High Level NM Notes:
1. Attacks with Hundred Fists speed.
2. Immune to many melee prevention statuses: paralyze, slow, etc.
3. Has a self-sleep mechanic after using Petribreath or being ranged for too long while kiting.
4. Regens with ~1% HP/tick while sleeping and will heal to full with self-sleep if not woken up.
5. Has a 'whoosh' animation when waking up from self-sleep.
6. Has a preference for using Petribreath.
7. Rumble and Ram Charge are used twice in succession.
8. Casts Paralyga and Stoneskin.

1. !additem 1178
9. Go to Horlais Peak
10. Enter Today's Horoscope KSNM 30


zach2good (approved):
I _LOVE_ the spreadsheet, thanks!

scripts/zones/Horlais_Peak/mobs/Aries.lua
```diff
@@ -0,0 +1,216 @@
...
+
+-----------------------------------
+-- Checks if the mob is asleep.
+-----------------------------------
+local isMobAsleep = function(mob)
+    return
+        mob:getStatusEffect(xi.effect.SLEEP_I) or
+        mob:getStatusEffect(xi.effect.SLEEP_II) or
+        mob:getStatusEffect(xi.effect.LULLABY)
+end
```
Xaver-DaRed: lullaby and sleep ii effects arent used, and im also pretty sure theres a function that does this already.

scripts/zones/Horlais_Peak/mobs/Aries.lua
```diff
@@ -0,0 +1,216 @@
...
+    local action = mob:getCurrentAction()
+
+    return
+        action == xi.act.MOBABILITY_START or
+        action == xi.act.MOBABILITY_USING or
+        action == xi.act.MOBABILITY_FINISH or
+        action == xi.act.MAGIC_START or
+        action == xi.act.MAGIC_CASTING or
+        action == xi.act.MAGIC_FINISH
+end
```
Xaver-DaRed, line 37: This function idk if we have already. But if we dont, I think its a prime candidate for an util global function

Xaver-DaRed:
Excellent work

scripts/zones/Horlais_Peak/mobs/Aries.lua
RaraProjects: It looks like there is an `isAsleep()` function for the `CBattleEntity` class, but I don't see anything that explicitly references sleep for the `CBaseEntity` class or in `CBaseEntity.lua`. I'll remove the `LULLABY` and `SLEEP_II` checks.

Xaver-DaRed: Well, aries is a battle entity. mob:isAsleep() should work

RaraProjects: I tried adding it to `CBaseEntity.lua`, but I'm still getting nil references when calling it. I think that I need to implement it in `lua_baseentity.cpp` and then do a `SOL_REGISTER` for it to successfully connect to `CBaseEntity.lua`, but I'm less confident with C++. I could also be completely wrong.

Xaver-DaRed: I ended up looking it up.

Use `mob:canUseAbilities()` in the lua script instead of the function you coded. It checks for those status effects among others

RaraProjects: I think I might be able to use `mob:canUseAbilities()` in place of `abilityBlocked()`, but I need to check specifically for sleep for the  `isAsleep` variable in `onMobFight()` because the mob gets a regen when it sleeps. If I use `mob:canUseAbilities()` for `isAsleep` then I'll end up adding regen every time the mob casts spells, uses an ability, has Amnesia, etc.

Xaver-DaRed (approved):

PR #7382, opened 2025-04-08
[lua] Converts Like Shining Subligar to IF

slashtangent:
**_I affirm:_**

Converts Like Shining Subligar to IF
Fixes an issue in which Heiji would keep track of how many you turned in, but would get amnesia when telling the player.

1. ``!zone Norg``
2. Talk to Heiji : ``!pos -1 -5 25 252``
3. Trade him Rusty Subligar until 10 or more are turned in.

## Captures
https://youtu.be/Eyy92WMd47o
https://1drv.ms/u/s!AlLEVQXhZeaHggljW0_MHUOSpPnu?e=vvnv9N


zach2good (approved):

scripts/globals/quests.lua
```diff
@@ -579,7 +579,7 @@ xi.quest.id =
         SECRET_OF_THE_DAMP_SCROLL       = 135, -- ± Converted
         THE_SAHAGINS_STASH              = 136, -- + Converted
         ITS_NOT_YOUR_VAULT              = 137, -- +
-        LIKE_A_SHINING_SUBLIGAR         = 138, -- +
+        LIKE_SHINING_SUBLIGAR           = 138, -- + Converted
```
claywar: Is there a reason you're changing the name? If it appears in the quest log as "Like a Shining Subligar" I'd prefer to keep it like that

scripts/quests/outlands/Like_Shining_Subligar.lua
```diff
@@ -0,0 +1,86 @@
...
+    title    = xi.title.LOOKS_SUBLIME_IN_A_SUBLIGAR,
+    item     = xi.item.SCROLL_OF_KURAYAMI_ICHI,
+}
+
+quest.sections =
+{
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_AVAILABLE and
+            player:getFameLevel(xi.fameArea.NORG) >= 3
```
claywar: Extra indent on multiline return

scripts/quests/outlands/Like_Shining_Subligar.lua
```diff
@@ -0,0 +1,86 @@
...
+        [xi.zone.NORG] =
+        {
+            ['Heiji'] =
+            {
+                onTrigger = function(player, npc)
+                    return quest:event(124, quest:getVar(player, 'Prog'))-- Update the player on the number of Rusty Subligars they have turned in
+                end,
+
+                onTrade = function(player, npc, trade)
+                    local subligar = trade:getItemQty(xi.item.RUSTY_SUBLIGAR)
```
claywar: Align assignments

claywar: Also, this subligar var is a bit ambiguous.  If its qty, try to explain that its quantity

scripts/quests/outlands/Like_Shining_Subligar.lua
```diff
@@ -0,0 +1,86 @@
...
+                    local turnedInVar = quest:getVar(player, 'Prog')
```
claywar: numTurnedIn or something like that? Adding Var to the end feels odd

scripts/globals/quests.lua
slashtangent: Oversight on my part. Like Shining Leggings was incorrectly named Like a Shining Leggings, and assumed this was the same. Just verified that this quest does include an "a". Reverting it back.

scripts/quests/outlands/Like_Shining_Subligar.lua
KnowOne134: well, it is progress into the quest, once hits the right number you finish quest

Xaver-DaRed (dismissed):

scripts/quests/outlands/Like_Shining_Subligar.lua
```diff
@@ -0,0 +1,86 @@
...
+                            return quest:progressEvent(125) -- complete quest
+                        elseif numTurnedIn <= 9 then -- turning in less than the amount needed to finish the quest
+                            player:tradeComplete()
+                            quest:setVar(player, 'Prog', totalSubligar)
+                            return quest:event(124, totalSubligar) -- Update player on number of leggings turned in
+                        end
+                    else
+                        return quest:event(124, numTurnedIn) -- Update player on number of leggings turned in, but doesn't accept anything other than leggings
+                    end
+                end,
```
Xaver-DaRed, line 73: On second review, the logic in the trade could use some improvements:
```lua
    local subligarsInTrade = trade:getItemQty(xi.item.RUSTY_SUBLIGAR)
    local subligarsStored  = quest:getVar(player, 'Prog')
    local subligarsTotal   = subligarsInTrade + subligarsTurnedIn

    -- No subligars in trade or items other than subligars.
    if
        subligarsInTrade == 0 or
        subligarsInTrade ~= trade:getItemCount()
    then
        return quest:event(124, subligarsStored)
    end

    -- Trade has only subligars. Check how many we have total.
    if subligarsTotal >= 10 then
        return quest:progressEvent(125) -- Complete quest
    else
        player:tradeComplete()
        quest:setVar(player, 'Prog', subligarsTotal)
        return quest:event(124, subligarsTotal)
    end
end,
```

Xaver-DaRed: This logic should be more clear, take care of what Claywar asked for and should also fix a weird check in there.

Xaver-DaRed (approved):

PR #7393, opened 2025-04-12
Fix several CoP 4-2 Ouryu mob issues

TracentEden2:
**_I affirm:_**

This PR fixes several issues with Ouryu based on captures. This is in preparation for adding Ouryu Cometh in a future PR (which will leverage some of the same code; also note some logic might eventually be put into a flying wyrm mixin).

The fixes include:

1. Add change of phase (flying <-> ground) after taking 1000 damage (based on capture)
2. Add mistmelt functionality (based on capture)
3. Add added effect enstone (based on capture and retail [videos](https://www.youtube.com/watch?v=NCzace9hbIc))
4. Add range and magic damage taken down mods (based on capture)
5. Add status immunities (based on capture)
6. Add fastcast (based on capture)
7. Correct the sight aggro range (based on capture)
8. Correct HP (based on capture)
9. Change MP to use default value instead of set value of 500 (as no evidence of 500 and ouryu uses more MP in captures)
10. Remove multiple use of invincible (as no evidence from captures)
11. Make the phase (flying <-> ground) change logic more robust and with clear functions
12. Move the 25% cmbDmgMult out of the mob pool SQL and to the mob script
13. Correct the cmbDelay in the mob pool SQL (based on capture)
14. Wakes up immediately if slept while in the air (implied by [JP wiki](https://wiki.ffo.jp/html/7555.html) and stated in retail [videos](https://www.youtube.com/watch?v=NCzace9hbIc))
15. Changed job to PLD/PLD (as per Jimmayus monster stat spreadsheet)

The captures used include those by Siknoz [here](https://youtu.be/ADCPIMlNa20) and [here](https://youtu.be/_fY6Cuv6MZE), Wiggo [here](https://youtu.be/wg9alh9kzUg), and Zach2Good [here](https://drive.google.com/open?id=1ljO1JDGyDB-Mx2jxun8vg-81Ph23LO9O). Thank you to all the people that capture!

Fight Ouryu


scripts/items/mistmelt.lua
```diff
@@ -0,0 +1,26 @@
...
+-- ID: 5265
+-- Item: Mistmelt
+-- Item Effect: Forces Ouryu to land
+-----------------------------------
+local itemObject = {}
+
+itemObject.onItemCheck = function(target, item, param, player)
+    local result = 0
+
+    if target:getName() ~= 'Ouryu' then
```
Xaver-DaRed: This is going to need a zone + mob ID reference check, and not a name check.

Why? because this item doesnt work on all Ouryus

TracentEden2: > This is going to need a zone + mob ID reference check, and not a name check.
> 
> Why? because this item doesnt work on all Ouryus

Thanks for the info, I have added a battlefield ID check (as I found this was the method in other cases like CCB pump).

scripts/zones/Monarch_Linn/mobs/Ouryu.lua
```diff
@@ -21,45 +101,67 @@ entity.onMobFight = function(mob, target)
...
+        mob:getAnimationSub() == allFlightPhaseAnimationSub and
+        (mob:hasStatusEffect(xi.effect.SLEEP_I) or
+        mob:hasStatusEffect(xi.effect.SLEEP_II) or
+        mob:hasStatusEffect(xi.effect.LULLABY))
+    then
+        mob:delStatusEffect(xi.effect.SLEEP_I)
+        mob:delStatusEffect(xi.effect.SLEEP_II)
+        mob:delStatusEffect(xi.effect.LULLABY)
+    end
+end
```
Xaver-DaRed, line 141:
```lua
-- Wake up from sleep immediately if flying.
if mob:getAnimationSub() == allFlightPhaseAnimationSub then
    mob:wakeUp()
end
```

TracentEden2: > ```lua
> -- Wake up from sleep immediately if flying.
> if mob:getAnimationSub() == allFlightPhaseAnimationSub then
>     mob:wakeUp()
> end
> ```

Thanks for the suggestion, I have updated the PR.

Xaver-DaRed (approved):

PR #7394, opened 2025-04-12
Adjust ipc stubs generation for build efficiency

cocosolos:
Writing every time causes unnecessary rebuilds.

**_I affirm:_**

Currently ipc_stubs.h is generated every time CMake is configured even if the contents don't change, which updates timestamps and triggers an unnecessary rebuild.

Looking at how CMake handles [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html) (used for generating version.cpp):
> The generated file is modified and its timestamp updated on subsequent cmake runs only if its content is changed.

This commit changes the script from writing directly to the file, to writing to a string stream then comparing to the current file if it exists, and only writing the file if the contents are different.

Not sure if there's a better way to do this in CMake or something, but I'm not sure the use or status of this process so went with a simple approach.

```sh
rm -rf build
time cmake -S . -B build
# before: 0m37.721s
# after:  0m36.670s
time cmake --build build -j$(nproc)
# before: 11m24.449s
# after:  11m34.762s
time cmake -S . -B build
# before: 0m5.576s
# after:  0m5.157s
time cmake --build build -j$(nproc)
# before: 4m13.273s
# after:  0m0.517s
```

I did test changing the output of generate_ipc_stubs.py and it does detect the changes and rewrites the file resulting in the (expected) ~4min rebuild.


zach2good (approved):
Cool, well spotted!

PR #7397, opened 2025-04-13
Knocking on Forbidden Doors mannequin fixes

overefined:
**_I affirm:_**

1. Mannequins obtained through trade or purchase from Fyi Chalmwoh are now the correct race and can be equipped with items
2. Fixes the issue where players could obtain a mannequin through trade without it being removed from their inventory
3. Purchasing a mannequin with a full inventory will no longer take the player's gil and not give the item
4. Fixes an SQL error when calling setMannequinPose

!completequest 4 29
!completequest 4 77
!completequest 4 78
!addkeyitem 713
!setgil 1000000
!pos -39.273 -16.000 70.126 249
!zone 238


zach2good (approved):
Well spotted, thanks!

PR #7399, opened 2025-04-13
Core: More SQL refactoring and tidying

zach2good:
**_I affirm:_**

- More prepared stmts
- More multiple strings and not backslash broken strings
- Clean up LoadChar a bit
- Remove raw new from guilds
- Adds `LIMIT 1` to a bunch of places

## TODO
- [x] Remove `_sql` legacy handle
- [x] Remove `sql.h/sql.cpp`
- [x] Remove all "naked" queries for prepared statements
- [ ] Address all leftover TODOs, SQL->GetData() is popular here
- [ ] Make sure we aren't reusing rsets (`const auto rset = ...` only!)
- [ ] Audit all `db::preparedStmt`s (530, or so) for %'s and proper use of rset
- [x] First-class support for enums in database, remove old support casting
- [x] Everything compiles again
- [ ] Log in, zone, form parties, send messages
- [ ] We're free of ALL Athena code!


zach2good:
Optional extra: constexpr checks on `preparedStmt` to ensure they aren't using {} or %u/%s/%i etc.

zach2good:
@sruon I think you've achieved everything I wanted to do in this PR? Did you also pick out the first-class enum support? If so this PR can be killed

sruon:
I'll pull out the enum class part this week and then close it

sruon:
Replaced by #8248

Issue #7405, opened 2025-04-14
🔨 Ameretat audit


sruon:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature

While testing Vampiric Root, I noticed the following mechanics are missing in LSB:
- [ ] Ameretats use Vampiric Root and no other TP move, if the target has any dispellable buff.
  - Note this does not extend to all NMs. Morbol Emperor in Einherjar will use other moves freely. 
- [ ] Ameretats have increased regen when in their spawn area 
  - Distance unknown - nearly 1% a tick?

Claims I've seen and I believe may be valid but have not tested:
- [ ] Ameretats have Endrain when outside of their spawn area
- [ ] Ameretats stop moving at night

Claims I've seen and do not believe are true:
- Vampiric Root absorbs food and reraises
  - [Tested and it isn't true](https://www.youtube.com/watch?v=BLtr-YrjMJE) but might be for some NMs?
  - [ffo.jp page](https://wiki-ffo-jp.translate.goog/html/24312.html?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp) 
- Vampiric Lash absorbs buffs, food and reraises in ToAU areas
  - Seems like the wiki may be mistaking it for Vampiric Root, either way Ameretat won't use Vampiric Lash if you have any absorbable buff...

PR #7406, opened 2025-04-14
[lua] Vampiric Root steals status effects

sruon:
**_I affirm:_**

Adds logic to steal all status effects from target upon use.

Note that wikis claim this is supposed to steal Reraise and Food, which I have proven to be false, _at least on base Ameretat mobs_

This move isn't in any skill list currently but I've tested it on my upcoming Einherjar Morbol boss.

Ameretats need a thorough audit (and possibly the whole Morbol family), I've found so many conflicting or downright wrong claims on the wikis that I can't trust any of the data there. 

Related to #7405

### DIFF
--- a/scripts/actions/mobskills/vampiric_lash.lua
+++ b/scripts/actions/mobskills/vampiric_lash.lua
@@ -4,7 +4,7 @@
 -- Type: Magical
 -- Utsusemi/Blink absorb: 1 shadow
 -- Range: Melee
--- Notes: In ToAU zones, this has an additional effect of absorbing all status effects, including food.
+-- Notes: (Unverified) In ToAU zones, this has an additional effect of absorbing all status effects, including food.
 -----------------------------------
 ---@type TMobSkill
 local mobskillObject = {}
--- a/scripts/actions/mobskills/vampiric_root.lua
+++ b/scripts/actions/mobskills/vampiric_root.lua
@@ -1,10 +1,10 @@
 -----------------------------------
 -- Vampiric Root
--- Deals dark damage to a single target. Additional effect: Drain
+-- Steals HP from a single target and absorbs positive status effects.
 -- Type: Magical
 -- Utsusemi/Blink absorb: 1 shadow
 -- Range: Melee
--- Notes: If used against undead, it will simply do damage and not drain HP.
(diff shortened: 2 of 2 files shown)


zach2good (approved):

scripts/actions/mobskills/vampiric_root.lua
```diff
@@ -1,10 +1,9 @@
 -----------------------------------
 -- Vampiric Root
--- Deals dark damage to a single target. Additional effect: Drain
+-- Steals HP from a single target and absorbs positive status effects.
 -- Type: Magical
 -- Utsusemi/Blink absorb: 1 shadow
 -- Range: Melee
--- Notes: If used against undead, it will simply do damage and not drain HP.
```
Xaver-DaRed, line 7: Why was this comment deleted?

scripts/actions/mobskills/vampiric_lash.lua
```diff
@@ -4,7 +4,6 @@
 -- Type: Magical
 -- Utsusemi/Blink absorb: 1 shadow
 -- Range: Melee
--- Notes: In ToAU zones, this has an additional effect of absorbing all status effects, including food.
```
Xaver-DaRed, line 7: Why was this comment deleted?

scripts/actions/mobskills/vampiric_root.lua
sruon: I'm not sure in what context this move would ever be used against an undead. 

But now that I'm looking at it, this may be related to Monstrosity? Happy to revert the change.

scripts/actions/mobskills/vampiric_lash.lua
sruon: I reverted the change and simply tagged it as Unverified for now. 

For the back story, Ameretats will never use Vampiric Lash if you have any dispellable buff. They will always use Vampiric Root.

I can't claim I've tested all ToAU Morbols yet, but the only 2 base mobs in the whole expansion (Ameretat and Greater Ameretat) definitely have that behavior and the JP wiki agrees.

I'll try and sort it out in #7405 at some point.

Xaver-DaRed (approved):

PR #7426, opened 2025-04-15
[core] Allow buffs to be stolen silently

sruon:
**_I affirm:_**

Adds a boolean to `CLuaBaseEntity::stealStatusEffect`, which when set will propagate all the way down to `RemoveStatusEffect` to disable the "effect wears off" message packet.

Djigga in Einherjar steal buffs on attacks but it doesn't show a log message for the AE itself or for the effect wearing off. See [retail video](https://youtu.be/J7hzFU7U7IM?t=322) for what it looks like, around 02:28:19 (client timestamp)

Also adds animation ID for this AE. It's not a very common one, I only know of another mob using it: Tunga in Abyssea-Grauberg

Screenshot 2025-04-15 014955

```
entity.onAdditionalEffect = function(mob, target, damage)
    if math.random(1, 100) <= 25 then
        local result = mob:stealStatusEffect(target, xi.effectFlag.DISPELABLE, true)
        if result == 0 then
            return 0, 0, 0
        end

        return xi.subEffect.STATUS_DRAIN, xi.msg.basic.NONE, 0
    end

    return 0, 0, 0
end
```

I tested with Absorb-Attri spell as well to make sure it still worked without it set.

### DIFF
--- a/src/map/status_effect_container.h
+++ b/src/map/status_effect_container.h
@@ -37,6 +37,12 @@
 
 class CBattleEntity;
 
+enum class EffectNotice : uint8
+{
+    ShowMessage = 0, // Display the "effect wears off" or "gains effect" message to the player
+    Silent      = 1, // Suppress the message entirely
+};
+
 class CStatusEffectContainer
 {
 public:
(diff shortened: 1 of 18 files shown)


src/map/status_effect_container.h
```diff
@@ -62,12 +62,12 @@ class CStatusEffectContainer
...
-    uint8          EraseAllStatusEffect();                 // erases all status effects
-    EFFECT         DispelStatusEffect(EFFECTFLAG flag);    // We delete the first positive effect
-    uint8          DispelAllStatusEffect(EFFECTFLAG flag); // dispels all status effects
-    CStatusEffect* StealStatusEffect(EFFECTFLAG flag);     // dispels one effect and returns it
+    EFFECT         EraseStatusEffect();                             // We delete the first negative effect
+    EFFECT         HealingWaltz();                                  // dancers healing waltz
+    uint8          EraseAllStatusEffect();                          // erases all status effects
+    EFFECT         DispelStatusEffect(EFFECTFLAG flag);             // We delete the first positive effect
+    uint8          DispelAllStatusEffect(EFFECTFLAG flag);          // dispels all status effects
+    CStatusEffect* StealStatusEffect(EFFECTFLAG flag, bool silent); // dispels one effect and returns it
```
zach2good: We should stop using bools for flags like this, in favour of:

```
enum class StealLogging
{
    Normal,
    Silent,
};
```

sruon:
```
enum class EffectRemovalNotice : uint8
{
    ShowMessage = 0, // Display the "effect wears off" message to the player or nearby players
    Silent      = 1, // Suppress the message entirely
};
```

Updated commit with this, let me know what you think

sruon: Actually I figured this could also be expanded to the AddEffect side

```
enum class EffectNotice : uint8
{
    ShowMessage = 0, // Display the "effect wears off" or "gains effect" message to the player
    Silent      = 1, // Suppress the message entirely
};
```

Then I realized `CStatusEffectContainer::AddStatusEffect` is not even honoring the boolean currently. The adventure continues.

zach2good (approved):

src/map/status_effect_container.h
sruon: Not looking to make that change as part of this PR but after looking at it more closely, I think `AddStatusEffect` should either:

- lose the silent/notice flag
- be brought in line with `RemoveStatusEffect` and emit the messages directly

Currently the "gains effects" messages are emitted directly from the spells, item or action that caused them.

PR #7441, opened 2025-04-17
Fix the discontinuity in craft skillup formula

wccbuck:
**_I affirm:_**

The formula LSB uses to calculate the chance a player will increase their crafting skill with a synth is piecewise when using the "Modern System." This means if the server variable CRAFT_MODERN_SYSTEM is set to true, a separate skillup formula is used for crafts _above_ a player's skill level versus crafts _below_ their skill level. The formula appears to be designed such that players should have a lower chance to skill up on crafts below their skill level, which makes sense.

_Edit: I realized after posting this that all of these percentages are assuming CRAFT_CHANCE_MULTIPLIER = 2. With default settings, you can cut all those numbers in half. My point still stands, though._

However, a problem arises when a player's skill level is fractional, such as 108.9. If such a player were attempting to craft a level 109 item, their chance to skill up would be only 2%. If instead they attempted a level 108 or level 110 synth, their chance to skill up would be much higher: 14.7% or 22.3% respectively. This discontinuity in the curve is counterintuitive.

My suggestion is to simply change the baseDiff check in doSynthSkillUp from `if (baseDiff > 0)` to `if (baseDiff > 1)`. This change would have no impact at all on the curve if a player's synth level was a round number. But in the above case, a player with 108.9 skill attempting a level 109 craft would have a 17.2% to skill up, a number that cleanly sits between 14.7% and 22.3%. 

The charts below show skill-up rates for the following player craft levels: 50.0, 50.5, 50.9, 108.0, 108.5, and 108.9. You can see the dip in skill-up rate occurs whenever the craft level is not a round number, but this suggested fix smooths the curve in those cases.

image

image

image

image

image

image

Thanks to @ariel-logos for bringing my attention to this issue.

Attempt a craft with craft level 0.1 higher than your skill level; repeat with 1 level below or 1 level above, and the difference is apparent.

### DIFF
--- a/src/map/utils/synthutils.cpp
+++ b/src/map/utils/synthutils.cpp
@@ -693,7 +693,7 @@ namespace synthutils
 
             if (settings::get<bool>("map.CRAFT_MODERN_SYSTEM"))
             {
-                if (baseDiff > 0)
+                if (baseDiff > 1)
                 {
                     skillUpChance = (double)baseDiff * craftChanceMultiplier * (3 - (log(1.2 + charSkill / 100))) / 5; // Original skill up equation with "x2 chance" applied.
                 }


wccbuck:
Important note for those not familiar with this part of the code: The formula I'm plotting in those curves is the base skill up rate, before considerations like breaks (which cut the skillup chance in half) or skill up amount.

Xaver-DaRed (approved):

PR #7445, opened 2025-04-18
Replace mob_skills.sql magic numbers

sruon:
**_I affirm:_**

Testing my hand at making these tables easier to maintain/edit.
Just wanted to try one column and then it spiraled out of control.

I can say it compiles and the table renders to the same values but I haven't tested more than that yet.

Is it even worth investing time or should these tables be moved directly to lua/c++?


Xaver-DaRed:
> Is it even worth investing time or should these tables be moved directly to lua/c++?

I don't know about this table concretely, however, many of the tables containing static data could use cleanup by moving the data out to either lua or c++, depending on what the info is and where the info should be handled in.

For instance, animation time like in this table is c++ land, but, say, idk, element in spell table, that can and probably should be defined in lua.

In short? Any effort to help maintain this tables so ppl dont have to work with a sea of numbers is well worth the effort, IMO.

sruon:
Closing this PR for the time being. I'm exploring ways to streamline the capture <> database process and that will help inform me on the best way to proceed.

I'll resubmit the enum changes separately.

PR #7453, opened 2025-04-19
Implement Seasons Greetings KSNM 30

RaraProjects:
**_I affirm:_**

Implements a largely fleshed out (but still experimental) version of the KSNM 30 Seasons Greetings.

Details and capture info can be found here:
https://docs.google.com/spreadsheets/d/1t0BYUAtGJ6mYiEIRNCpNfTz9hC2gJdTX42LNGdeNVM0/edit?usp=sharing

High Level Battle Notes:
1. There are four tree NMs (WHM, MNK, NIN, DRK).
2. Each tree represents a season.
3. The trees start out unaggroed and only aggro when engaged.
4. When one tree uses its 2HR the next tree will wake up.
5. The order is always Gilgagoge (Spring) > Goga (Summer) > Ulagohvsdi (Autumn) > Gola (Winter).
6. If you start with Gola (Winter) then it will wrap around to the Gilagoge (Spring).

| Feature        | Notes     |
|-----------------|-----------------|
| Skills    | Each tree has a set ability that it uses. It won't use other abilities. |
| Spells    | Spells seem to be pretty standard for ~75 NMs    |
| Additional Effects    | Each tree has a unique additional effect that it can apply on melee.    |
| Spell Resistances   | Immune to Break and the mages are immune to Silence    |
| EEM    | After getting a couple EEMs, it seems like each tree might have its own EEM. Needs further verification.    |
| MP    | I couldn't fully get a tree drained so I don't know how much MP they have. I just have placeholder MP in place now.    |

Additional Battle Notes:
1. There is an interesting scenario where the player's damage will suddenly jump by almost two orders of magnitude.
2. I thought it was originally related to 2HR usage but I saw cases where that wasn't the case.
3. I thought it could be related to crits, but I saw cases where that wasn't the case.
4. I wondered if there was some sort of rotation happen where it could be time based or rotating physical damage type.
5. Needs further testing.

Notable Code Changes:
1. Implemented en-dispel.
2. Created a core function (with the help of @0x05010705) that helps with interacting with other mobs in the battlefield that are not the acting entity or its target. This was needed to wake up the next tree in line when the engaged tree used its 2HR.

1. !additem 1175
4. Go to Balga's Dais
5. Enter Seasons Greetings KSNM 30


scripts/zones/Balgas_Dais/mobs/Gilagoge_Tlugvi.lua
```diff
@@ -0,0 +1,83 @@
...
+
+        if not battlefield then
+            return
+        end
+
+        local nextMobIDs =
+        {
+            balgasID.mob.GILAGOGE_TLUGVI + 1,
+            balgasID.mob.GILAGOGE_TLUGVI + 6,
+            balgasID.mob.GILAGOGE_TLUGVI + 11,
```
zach2good: If you already have the IDs you want, you can use `GetMobByID()` to look each one up. This just grabs the mob out of the entity array without any sort of iteration.

scripts/enum/mob_skills.lua
```diff
@@ -35,13 +35,22 @@ xi.mobSkill =
 
     SILENCE_GAS_1            =  314,
     DARK_SPORE_1             =  315,
+    DRILL_BRANCH             =  328,
```
Xaver-DaRed, line 39: Leave a blank line when between dark spore and drill branch pls

scripts/enum/mob_skills.lua
```diff
@@ -35,13 +35,22 @@ xi.mobSkill =
...
+    PINECONE_BOMB            =  329,
+    LEAFSTORM                =  331,
```
Xaver-DaRed, line 42: And between this 2 aswell. And down bellow theres a few other cases.

 If the enums dont follow a direct progression, separate them

RaraProjects:
Do you have any preferences for how I handle the cyclomatic complexity in `xi.mob.onAddEffect`? I was thinking a good starting point would be breaking the Status Effect, Dispel, and Immediate Effect scenarios into their own functions.

Update: I separated each scenario into their own function. They separate pretty cleanly.

scripts/zones/Balgas_Dais/mobs/Gilagoge_Tlugvi.lua
RaraProjects: The trick with this scenario is that although I know the IDs of the mob I'm looking for, I don't know which version of the mob I need at runtime because the mob lua is battlefield agnostic. For example, if three groups are fighting in Seasons Greetings at the same time, and the WHM (Gilagoge) from group two uses Benediction, I need to make sure that I'm only interacting with battlefield two's version of the MNK (Goga).

The looping ensures that I'm only ever interacting with the appropriate battlefield's version of the mob.

Possible Gogas:
1. 17375408 (Battlefield 1)
2. `17375413 (Battlefield 2)`
3. 17375418 (Battlefield 3)

Group Two Mobs: (only this set would be checked in the example above)
1. 17375412 (Gilagoge)
2. `17375413 (Goga)`
3. 17375414 (Ulagohvsdi)
4. 17375415 (Gola)

Let me know if I'm misunderstanding your comment!

RaraProjects: I think I understand what you're suggesting now. I think you're recommending that I grab the ID of the mob performing the 2HR and offset (+1) from that to get the next mob (with some special modulo handling to loop back around to the beginning if needed). I think that would work.

Is there value in keeping `getSpecificMob` for similar cases in other BCNMs like Requiem of Sin, etc.?

zach2good: If you wanted to work off of a specific base offset for a given battlefield, we have `battlefield:getArea()`, which give you (I think) 1-3, depending on which version of the battlefield you've been placed in. You can then use that to calculate which mob and offset you need.

There's some useful examples hanging around:
```lua
local baseID = ID.mob.ZEID_BCNM_OFFSET + (mob:getBattlefield():getArea() - 1) * 4

...

    local inst = mob:getBattlefield():getArea()
    local instOffset = ID.mob.WARLORD_ROJGNOJ + (14 * (inst - 1))

...


local promathia = ID.mob.PROMATHIA + (mob:getBattlefield():getArea() - 1) * 2
```

zach2good (approved):

PR #7467, opened 2025-04-22
Implement 'Enhance Burst Affinity', make BLU AF3 set bonus apply to magical spells

wccbuck:
**_I affirm:_**

This change implements the "Enhances Burst Affinity effect" modifier from Mavi/Hashishin Basmak and Assimilator's Shalwar (see [Burst Affinity](https://www.bg-wiki.com/ffxi/Burst_Affinity) for the full list of gear). The change also correctly applies the "Occ. augments blue magic spells" set bonus from BLU AF3 to magical spells instead of only physical. Finally, this change assigns an ID for the ENHANCES_CHAIN_AFFINITY mod and applies it to relevant gear, but does not implement the actual effect; that'll come at a later date.

Put a print in server/scripts/globals/blumagic.lua in useMagicalSpell that shows "bonusWSC" after adjustments.

!changejob BLU 99
!addspell 648 (regurgitation)
!additem 23256 (assimilator's shalwar +2)

Find an enemy, use burst affinity, use regurgitation. Without the pants, bonusWSC should be 1 (total WSC modifier should be x2).

Now put the pants on, !reset, use burst affinity, use regurgitation on a mob. Now bonusWSC should be 1.14 (total WSC modifier should be x2.14).


wccbuck:
Check failed due to commit message too long. Should I go back and amend the commit?

Mortalelite:
Commit message up to 72 characters so you would need to fix it

wccbuck:
It's been brought to my attention that `math.random(0, 99)` is not the proper way to handle random numbers here, but rather 1 to 100. Here I literally copied a line of code from further up in the file, under usePhysicalSpell, so I was not aware; and if it's better that this check be changed, we probably ought to update the physical spell one too.

Xaver-DaRed (approved):

PR #7477, opened 2025-04-24
Implement Circle ability dmg bonus

wccbuck:
**_I affirm:_**

Currently in LSB, all the "Circle" abilities (Holy/Arcane/Warding/Ancient Circle) simply add a certain amount of "Killer" property to every party member in range. (Killer Instinct isn't implemented yet.) However, [this SE release outlining the July 21, 2009 FFXI update](http://www.playonline.com/pcd/verup/ff11us/detail/4742/detail.html) describes a change to how "Circle" abilities work.

Here's the relevant tidbit:

"The following Circle-type abilities have had their effects adjusted and will now grant enhanced resistance, defense, and attack against the specified monster family to party members within their area of effect.
Holy Circle / Arcane Circle / Warding Circle / Ancient Circle / Killer Instinct"

The value of this attack/defense bonus has been determined by the community to be about equal to the power of the added "Killer" effect. Example testing:

* [Killer Effects + Circles, FFXIAH user Braden, 2016](https://www.ffxiah.com/forum/topic/48529/the-official-bgwiki-thread/6#3181764): user observes +15% damage dealt, -15% damage taken, and +15% killer effect when using circles against regular mobs, or 10% against NMs. This thread also supports the idea that the "flinch" chance is reduced by 1/3 when the enemy is an NM, which I have also included in this PR.
* [Holy Circle bg user Martel, 2011, part 1](https://www.bluegartr.com/threads/107650-Random-Question-Thread-XXIV-Occupy-the-RQT?p=4909394&viewfull=1#post4909394): user observes a 17% damage increase from Holy Circle, is confused. [Part 2](https://www.bluegartr.com/threads/107650-Random-Question-Thread-XXIV-Occupy-the-RQT?p=4909428&viewfull=1#post4909428): user realizes he had Gallant Leggings equipped, which added +2 to the base 15% potency of Holy Circle
* [Arcane Circle, bg user Byrthnoth, 2011](https://www.bluegartr.com/threads/107650-Random-Question-Thread-XXIV-Occupy-the-RQT?p=4909792#post4909792): user observes a 15% base damage increase when using Arcane Circle as main job DRK, and 5% increase when it is used as subjob.
* [Killer Instinct, bg user Byrthnoth, 2011](https://www.bluegartr.com/threads/120200-Killer-Trait-Testing?p=4643650&viewfull=1#post4643650): user tests the damage reduction component of Killer Instinct, and finds it is a separate multiplicative modifier on top of "damage taken"
* [Misc testing, bg user Byrthnoth, 2011](https://www.bluegartr.com/threads/120200-Killer-Trait-Testing?p=4643654&viewfull=1#post4643654): This user describes using a subjob DRK to use arcane circle and seeing a 5% damage increase, while using Killer Instinct as a main-job BST and seeing a 15% damage increase

Sidebar: At first glance, that last thread seemed to show some evidence that Killer Instinct gave bonus damage/dt based on *total killer effect*. But that was only because they were wearing a [Gausape](https://www.bg-wiki.com/ffxi/Ferine_Gausape_%2B1) which explicitly converts Killer effects to damage boost/damage taken, so we can consider that a completely unrelated detail.

In support of these findings, [the japanese FFXI wiki](https://wiki.ffo.jp/html/12868.html) describes Circle effects like so:

(Translated)
"Both damage up and damage reduction are effective against all types of damage , whether physical or magical . The effective value is the same as the race defense. The reduction when fighting NM is also the same. Damage reduction is multiplied separately from damage cut equipment , so for example, if you use a circle while your damage is -50% with cut equipment , the damage cut will be -50% x -15%, or -57.5%."

The [Holy Circle page](https://wiki.ffo.jp/html/2610.html) goes on to say that the holy circle damage bonus applies to all damage dealt, even damage from Atonement.

In summary, this change makes LSB's implementation of "Circle" abilities align with documented and tested retail behavior as it been since 2009. This can eventually be expanded to include all the different types of killer effects you can get from Killer Instinct, once that ability is implemented. 

NOTE: I know a lot of this stuff is being migrated out of c++ and into lua, but as of now, this seemed to be the most straightforward approach to implementing this change. I'm happy to try a different approach, like move circleDmgAdjust to globals.utils, or to the new "xi.combat.utilities" once that's a thing, and assist with the migration however I can.

Test all four circles (pld, sam, drg, and drk) to ensure they affect the potency of the following actions:
* Auto attacks (damage boost when attacking, damage reduced when defending)
* Weaponskills
* Damaging spells
* Drain/Drain II
* Spikes (I tested with Reprisal)
* Enspells (I tested with Enlight)
* Chi blast (it uses weird separate logic, test this on mnk/pld against a skeleton)

Repeat the main tests above (attacks, weaponskills, spells) with the appropriate AF piece that augments your circle, and see its potency increase by 2.

Repeat the main tests above (attacks, weaponskills, spells) against an NM instead of a standard mob, and see your potency decrease to 10 (down from 15). If you have your AF equipped, it should be 11 (down from 17). 

Repeat the main tests above (attacks, weaponskills, spells) with pld, sam, drg, or drk as your subjob, and confirm your circle ability potency drops to 5%.

### DIFF
--- a/scripts/globals/spells/damage_spell.lua
+++ b/scripts/globals/spells/damage_spell.lua
@@ -849,7 +849,6 @@ end
 xi.spells.damage.calculateNukeAbsorbOrNullify = function(target, spellElement)
     local nukeAbsorbOrNullify = 1
     local liementFactor       = target:checkLiementAbsorb(xi.damageType.ELEMENTAL + spellElement) -- Check for Liement.
-
     -- Absobtion by liement.
     if liementFactor < 0 then
         return liementFactor
@@ -1136,6 +1135,7 @@ xi.spells.damage.useDamageSpell = function(caster, target, spell)
     finalDamage = utils.clamp(finalDamage - target:getMod(xi.mod.PHALANX), 0, 99999)
     finalDamage = utils.clamp(utils.oneforall(target, finalDamage), 0, 99999)
     finalDamage = utils.clamp(utils.stoneskin(target, finalDamage), -99999, 99999)
+    finalDamage = caster:circleDmgAdjust(target, finalDamage)
 
     -- Handle final adjustments. Most are located in core. TODO: Decide if we want core handling this.
     -- Check if the mob has a damage cap
--- a/scripts/globals/spells/damage_spell.lua
+++ b/scripts/globals/spells/damage_spell.lua
@@ -886,6 +886,49 @@ xi.spells.damage.calculateNukeAbsorbOrNullify = function(target, spellElement)
     return nukeAbsorbOrNullify
 end
 
+xi.spells.damage.calculateCircleDmgMultiplier = function(caster, target)
+    local circleDmgMultiplier = 1
+    -- Modifies damage based on the attacker/defender ecosystem, if attacker and/or
+    -- defender has warding, ancient, arcane, or holy circle active
+    -- https://wiki.ffo.jp/html/12868.html
+
+    local ecoTable =
+    -- TODO: Expand this for Killer Instinct (BST merit ability)
+    {
+        [xi.eco.ARCANA] = xi.mod.ARCANE_CIRCLE_DMG_BONUS,
+        [xi.eco.DEMON ] = xi.mod.WARDING_CIRCLE_DMG_BONUS,
+        [xi.eco.DRAGON] = xi.mod.ANCIENT_CIRCLE_DMG_BONUS,
+        [xi.eco.UNDEAD] = xi.mod.HOLY_CIRCLE_DMG_BONUS,
+    }
+
+    -- Handle both cases separately. It's possible, however unlikely, that both target and
+    -- caster have a circle ability up which counters the other entity's ecosystem
+    local casterWeakCircle = ecoTable[caster:getEcosystem()]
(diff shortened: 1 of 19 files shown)


scripts/globals/spells/damage_spell.lua
```diff
@@ -1136,6 +1135,7 @@ xi.spells.damage.useDamageSpell = function(caster, target, spell)
     finalDamage = utils.clamp(finalDamage - target:getMod(xi.mod.PHALANX), 0, 99999)
     finalDamage = utils.clamp(utils.oneforall(target, finalDamage), 0, 99999)
     finalDamage = utils.clamp(utils.stoneskin(target, finalDamage), -99999, 99999)
+    finalDamage = caster:circleDmgAdjust(target, finalDamage)
```
Xaver-DaRed: No.

This is not the place to apply damage multipliers. Definately not after phalanx and stoneskin.
It also doesnt need a call to core to transform the damage. Thats the opposite direction we want to go

WinterSolstice8:
can you rebase? looks fine to me

Xaver-DaRed:
Superseded by PR #10488

PR #7495, opened 2025-04-26
[lua][fix] Fix 'augments SA'/'TA' math in weaponskill damage

wccbuck:
**_I affirm:_**

Vajra's "Enhances Sneak Attack/Trick Attack effect" applies a bonus percentage to sneak attack / trick attack damage. This is implemented correctly in attack.cpp, applying to the _entire_ damage instead of just the bonus from dex/agi respectively. However, in weaponskills.lua, in the SATA section for applying those bonuses to the first hit, AUGMENTS_SA / AUGMENTS_TA were only being applied to the bonus, not the full "finaldmg" of the first hit. This change just adds a pair of parentheses to correct this.

Equip a vajra, do a weaponskill with SA or TA

### DIFF
--- a/scripts/globals/weaponskills.lua
+++ b/scripts/globals/weaponskills.lua
@@ -538,11 +538,11 @@ xi.weaponskills.calculateRawWSDmg = function(attacker, target, wsID, tp, action,
     if attacker:getMainJob() == xi.job.THF then
         -- Add DEX/AGI bonus to first hit if THF main and valid Sneak/Trick Attack
         if calcParams.sneakApplicable then
-            finaldmg = finaldmg + calcParams.pdif * attacker:getStat(xi.mod.DEX) * (1 + attacker:getMod(xi.mod.SNEAK_ATK_DEX) / 100) * (1 + attacker:getMod(xi.mod.AUGMENTS_SA) / 100)
+            finaldmg = (finaldmg + calcParams.pdif * attacker:getStat(xi.mod.DEX) * (1 + attacker:getMod(xi.mod.SNEAK_ATK_DEX) / 100)) * (1 + attacker:getMod(xi.mod.AUGMENTS_SA) / 100)
         end
 
         if calcParams.trickApplicable then
-            finaldmg = finaldmg + calcParams.pdif * attacker:getStat(xi.mod.AGI) * (1 + attacker:getMod(xi.mod.TRICK_ATK_AGI) / 100) * (1 + attacker:getMod(xi.mod.AUGMENTS_TA) / 100)
+            finaldmg = (finaldmg + calcParams.pdif * attacker:getStat(xi.mod.AGI) * (1 + attacker:getMod(xi.mod.TRICK_ATK_AGI) / 100)) * (1 + attacker:getMod(xi.mod.AUGMENTS_TA) / 100)
         end
     end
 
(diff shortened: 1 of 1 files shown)


scripts/globals/weaponskills.lua
```diff
@@ -538,11 +538,11 @@ xi.weaponskills.calculateRawWSDmg = function(attacker, target, wsID, tp, action,
...
         if calcParams.sneakApplicable then
-            finaldmg = finaldmg + calcParams.pdif * attacker:getStat(xi.mod.DEX) * (1 + attacker:getMod(xi.mod.SNEAK_ATK_DEX) / 100) * (1 + attacker:getMod(xi.mod.AUGMENTS_SA) / 100)
+            finaldmg = (finaldmg + calcParams.pdif * attacker:getStat(xi.mod.DEX) * (1 + attacker:getMod(xi.mod.SNEAK_ATK_DEX) / 100)) * (1 + attacker:getMod(xi.mod.AUGMENTS_SA) / 100)
         end
 
         if calcParams.trickApplicable then
-            finaldmg = finaldmg + calcParams.pdif * attacker:getStat(xi.mod.AGI) * (1 + attacker:getMod(xi.mod.TRICK_ATK_AGI) / 100) * (1 + attacker:getMod(xi.mod.AUGMENTS_TA) / 100)
+            finaldmg = (finaldmg + calcParams.pdif * attacker:getStat(xi.mod.AGI) * (1 + attacker:getMod(xi.mod.TRICK_ATK_AGI) / 100)) * (1 + attacker:getMod(xi.mod.AUGMENTS_TA) / 100)
         end
     end
```
Xaver-DaRed, line 549:
```lua
if attacker:getMainJob() == xi.job.THF then
    -- Add DEX/AGI bonus to first hit if THF main and valid Sneak/Trick Attack
    if calcParams.sneakApplicable then
        local dexFactor       = math.floor(attacker:getStat(xi.mod.DEX) * (1 + attacker:getMod(xi.mod.SNEAK_ATK_DEX) / 100))
        local sneakAttAugment = 1 + attacker:getMod(xi.mod.AUGMENTS_SA) / 100
        finaldmg              = math.floor((finaldmg + dexFactor) * sneakAttAugment)
    end

    if calcParams.trickApplicable then
        local agiFactor       = math.floor(attacker:getStat(xi.mod.AGI) * (1 + attacker:getMod(xi.mod.TRICK_ATK_AGI) / 100))
        local trickAttAugment = 1 + attacker:getMod(xi.mod.AUGMENTS_TA) / 100
        finaldmg              = math.floor((finaldmg + agiFactor) * trickAttAugment)
    end
end
```

Xaver-DaRed: Floors and we have to remove those pdif multipliers aswell. They are already applied in line 327, after all, and they are not there in the core calculations you mentioned in the PR description to be correct.

wccbuck: The formula needs PDIF to also apply to the dex/agi mod that is added to sneak/trick attack, because those are added to base damage. But I do agree with simplifying. I'll make adjustments

wccbuck: I separated out dexFactor and agiFactor like you suggested to improve legibility. In addition, I moved the "AUGMENTS_SA/TA" code out of the "if main job is THF" code block, since, unlike "SNEAK_ATK_DEX" and "TRICK_ATK_AGI", there is nothing fundamental about AUGMENTS_SA/TA that would require THF main job.

Xaver-DaRed (approved):

PR #7496, opened 2025-04-26
Clean up attack class

overefined:
**_I affirm:_**

I originally had more substantial changes and meant to update https://github.com/LandSandBoat/server/pull/7494 instead of closing it.

### DIFF
--- a/src/map/attack.cpp
+++ b/src/map/attack.cpp
@@ -28,11 +28,6 @@
 #include "status_effect_container.h"
 #include "utils/puppetutils.h"
 
-/************************************************************************
- *                                                                      *
- *  Constructor.                                                            *
- *                                                                      *
- ************************************************************************/
 CAttack::CAttack(CBattleEntity* attacker, CBattleEntity* defender, PHYSICAL_ATTACK_TYPE type, PHYSICAL_ATTACK_DIRECTION direction, CAttackRound* attackRound)
 : m_attacker(attacker)
 , m_victim(defender)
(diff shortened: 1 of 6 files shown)


Xaver-DaRed (approved):

src/map/attack.cpp
```diff
@@ -30,7 +30,7 @@
 
 /************************************************************************
  *                                                                      *
- *  Constructor.                                                            *
+ *  Constructor.                                                        *
```
zach2good: If we're in here cleaning up, big block comments that don't provide any additional information (like this one) can go

src/map/attack.cpp
```diff
@@ -459,12 +459,12 @@ bool CAttack::CheckAnticipated()
     }
 }
 
-bool CAttack::CheckHadSneakAttack() const
+bool CAttack::IsSA() const
```
zach2good: Please use `isSneakAttack()` (and the TA equivalent), it parses easier

src/map/attack.h
```diff
@@ -70,38 +70,37 @@ class CAttack
...
-    float                     GetDamageRatio() const;               // Gets the damage ratio.
-    void                      SetGuarded(bool);                     // Sets the isGuarded flag.
-    bool                      IsGuarded();                          // Sets the isGuarded flag. Also alters the damage ratio accordingly.
-    bool                      IsEvaded() const;                     // Gets the evaded flag.
-    void                      SetEvaded(bool value);                // Sets the evaded flag.
-    bool                      IsBlocked() const;                    // Returns the blocked flag.
+    uint16                    GetAnimationID();                         // Returns the animation ID.
+    PHYSICAL_ATTACK_TYPE      GetAttackType() const;                    // Returns the attack type (Double, Triple, Zanshin ect).
+    void                      SetAttackType(PHYSICAL_ATTACK_TYPE type); // Sets the attack type.
+    PHYSICAL_ATTACK_DIRECTION GetAttackDirection() const;               // Returns the attack direction.
```
zach2good: If you use trailing return syntax here you can get rid of the wild spacing

zach2good (commented):
Seems fine, just needs some things renamed 👍

zach2good (approved):

zach2good:
Seems fine, trying to kick CI until it starts up again

zach2good:
You may have to force push to get CI to run again, doesn't look like I can force anything from my side

PR #7497, opened 2025-04-27
Move auto attack damage to lua

overefined:
**_I affirm:_**

Everything is conditioned on ENABLE_AUTO_ATTACK_LUA in map.lua
- Moves damage calculations from CAttack::ProcessDamage() to lua
- Adds lua versions of doConsumeMana and doSoulEater
- Adds lua binding for attackutils::CheckForDamageMultiplier

The first pull request working towards a version of https://github.com/LandSandBoat/server/pull/7461

Observe auto attack damage under the various affected conditions.

### DIFF
--- a/scripts/globals/combat/effect_utilities.lua
+++ b/scripts/globals/combat/effect_utilities.lua
@@ -1,0 +1,77 @@
+xi = xi or {}
+xi.combat = xi.combat or {}
+xi.combat.effect = xi.combat.effect or {}
+
+xi.combat.effect.handleConsumeMana = function(player)
+    local damage = 0
+
+    if player:hasStatusEffect(xi.effect.CONSUME_MANA) then
+        damage = damage + math.floor(player:getMP() / 10)
+        player:setMP(0)
+        player:delStatusEffect(xi.effect.CONSUME_MANA)
+    end
+
+    return damage
+end
+
+xi.combat.effect.handleSoulEater = function(player, damage)
+    if player:hasStatusEffect(xi.effect.SOULEATER) then
+        -- Souleater's HP consumed is 10% (base) + x% from gear (only highest) + x% from gear augments.
+        local souleaterBonus = player:getMaxGearMod(xi.mod.SOULEATER_EFFECT) * 0.01
+        local souleaterBonusII = player:getMod(xi.mod.SOULEATER_EFFECT_II) * 0.01
+        local stalwartSoulBonus = 1 - player:getMod(xi.mod.STALWART_SOUL) / 100
+        local bonusDamage = player:getHP() * (0.1 + souleaterBonus + souleaterBonusII)
+
+        if bonusDamage >= 1 then
+            player:addHP(-utils.stoneskin(player, math.floor(bonusDamage * stalwartSoulBonus)))
+
+            if player:getMainJob() == xi.job.DRK then
+                damage = math.floor(damage + bonusDamage)
+            else
+                damage = math.floor(damage + (bonusDamage / 2))
+            end
+        end
+    end
+
+    return damage
+end
+
+xi.combat.effect.handleRestraint = function(player)
+    if player:hasStatusEffect(xi.effect.RESTRAINT) then
+        local effect = player:getStatusEffect(xi.effect.RESTRAINT)
+        if effect == nil then
+            return
+        end
+
+        local power = effect:getPower()
+
+        if effect:getPower() < 30 then
+            local jpBonus = 0
+
+            if player:isPC() then
+                jpBonus = player:getJobPointLevel(xi.jp.RESTRAINT) * 2
+            end
+
+            -- Convert weapon delay and divide
+            -- Pull remainder of previous hit's value from Effect Sub Power
+            local boostPerRound = ((player:getBaseDelay() / 1000) * 60) / 385
+            local remainder = effect:getSubPower() / 100
+
+            -- Calculate bonuses from Enhances Restraint, Job Point upgrades, and remainder from previous hit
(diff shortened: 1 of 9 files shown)


scripts/globals/combat/effect_utilities.lua
```diff
@@ -0,0 +1,93 @@
+xi = xi or {}
+xi.combat = xi.combat or {}
+xi.combat.effect = xi.combat.effect or {}
+
+xi.combat.effect.handleStoneskin = function(defender, damage)
```
WinterSolstice8: Is this functionally different than [utils.stoneskin](https://github.com/LandSandBoat/server/blob/c0324c5693c35d850c8ca2c6db12b165885b35ae/scripts/globals/utils.lua#L473)? Seems to be essentially identical with some slightly different logic

WinterSolstice8:
Sorry for the super late review

scripts/globals/combat/effect_utilities.lua
overefined: I didn't even know utils.stoneskin existed. All I did was take the logic in the C++ HandleStoneskin and convert it to Lua. It seems like the only difference is utils.stoneskin includes

target:setMod(xi.mod.STONESKIN, 0)

while the C++ version does not

I feel like we should probably factor out some of those effect calculations in utils.lua to another file. Let me know what I should do for this PR though.

WinterSolstice8: Replace the call with utils.stoneskin just to be consistent with the rest of lua for now

I'll do a more in depth review in a few days when I'm off work, but at a glance it looks good to me. Disabled by default means i'm not too worried about it honestly

scripts/globals/combat/physical_utilities.lua
```diff
@@ -80,6 +80,102 @@ local shieldSizeToBlockRateTable =
...
+    if isH2H then
+        local naturalH2hDamage = math.floor(attacker:getSkillLevel(xi.skill.HAND_TO_HAND) * 0.11) + 3
+
+        if physicalAttackType == xi.physicalAttackType.KICK then
+            local kickDamage = naturalH2hDamage + attacker:getMod(xi.mod.KICK_DMG)
+            damage = math.floor(damageRatio * (kickDamage + bonusBasePhysicalDamage + xi.combat.physical.calculateMeleeStatFactor(attacker, victim)))
+        else
+            damage = math.floor(damageRatio * (baseDamage + naturalH2hDamage + bonusBasePhysicalDamage + xi.combat.physical.calculateMeleeStatFactor(attacker, victim)))
+        end
+    elseif slot == xi.slot.MAIN then
```
WinterSolstice8, line 122: it looks like xi.slot.SUB is missing from this else/if chain

src/map/attack.cpp
```diff
@@ -566,6 +566,67 @@ bool CAttack::CheckCover()
...
+        }
+
+        // Trick attack.
+        if (m_attacker->GetMJob() == JOB_THF && m_isFirstSwing && m_attackRound->GetTAEntity() != nullptr)
+        {
+            m_isTA = true;
+        }
+
+        auto calculateAttackDamage = lua["xi"]["combat"]["physical"]["calculateAttackDamage"];
+        auto result                = calculateAttackDamage(m_attacker, m_victim, GetWeaponSlot(), m_attackType, m_attackRound->IsH2H(), m_isFirstSwing, m_isSA, m_isTA, m_damageRatio, m_attacker->GetMainWeaponDmg());
```
WinterSolstice8: This is currently hardcoded to only use  m_attacker->GetMainWeaponDmg(), but this could be fixed here or in lua. you are passing in the slot so it may not be necessary to pass in weapon damage.

Need to have slot sub/ammo support but I think it should be passed into lua via the slot (and removed here)

scripts/globals/combat/physical_utilities.lua
```diff
@@ -80,6 +80,102 @@ local shieldSizeToBlockRateTable =
...
+    if damage < 0 then
+        damage = 0
+    end
+
+    -- Apply Restraint Weaponskill Damage
+    if isFirstSwing then
+        xi.combat.effect.handleRestraint(attacker)
+    end
+
+    return damage
```
WinterSolstice8, line 178: Can you add a TODO before the return

```suggestion
    -- TODO: add charutils::TrySkillUP call
    return damage
```

WinterSolstice8: Reread the PR, you have the skillup left over in C++, but i'd still eventually like it in lua (this is more of a reminder comment)

src/map/attack.cpp
```diff
@@ -566,6 +566,67 @@ bool CAttack::CheckCover()
...
+        if (result.valid())
+        {
+            m_damage = result.get<int32>(0);
+
+            // Set attack type to Samba if the attack type is normal.  Don't overwrite other types.  Used for Samba double damage.
+            if (m_attackType == PHYSICAL_ATTACK_TYPE::NORMAL && (m_attacker->StatusEffectContainer->HasStatusEffect(EFFECT_DRAIN_SAMBA) ||
```
WinterSolstice8: This block to set the attack type should be moved to after the calls above where SA/TA is set, otherwise it has no effect when `attackutils::CheckForDamageMultiplier`  is called within lua because it will not be a Samba attack (yet)

WinterSolstice8:
Overall, nice job. Just some minor changes and looks good to me.

Again, sorry for the late review.

WinterSolstice8 (approved):

WinterSolstice8:
Good work, thanks for this.

PR #7499, opened 2025-04-27
[core] Check for throwing weapons in baseentity getStat

wccbuck:
**_I affirm:_**

PR #7450 fixed the bug where ranged PDIF was based on marksmanship skill when using a thrown weapon, but the damage dealt was still based on marksmanship skill, not throwing skill. This PR fixes that issue.

On base branch, `!setskill MARKSMANSHIP 0`, then throw a shuriken at a mob. It will do very little damage. Then, `!setskill MARKSMANSHIP 300`, throw another shuriken, and you'll see a large increase in damage.

Do the same experiment with this PR and you'll see shuriken/boomerang/dart/pebble damage should depend only on throwing skill.

### DIFF
--- a/src/map/lua/lua_baseentity.cpp
+++ b/src/map/lua/lua_baseentity.cpp
@@ -14402,7 +14402,15 @@ uint16 CLuaBaseEntity::getStat(uint16 statId, sol::variadic_args va)
                 }
                 else
                 {
-                    value = PEntity->RATT(SKILL_MARKSMANSHIP); // TODO: does this edge case exist? will mobs or trusts hit this?
+                    PWeapon = dynamic_cast<CItemWeapon*>(PEntity->m_Weapons[SLOTTYPE::SLOT_AMMO]);
+                    if (PWeapon)
+                    {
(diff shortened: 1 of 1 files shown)


wccbuck:
I left in the default marksmanship for no ranged weapon and no ammo, because I don't know the answer to the question this comment poses: 
```
value = PEntity->RATT(SKILL_MARKSMANSHIP); // TODO: does this edge case exist? will mobs or trusts hit this?
```

I felt it was safer not to throw an error in the event I was unaware of a strange corner case.

sruon:
> I left in the default marksmanship for no ranged weapon and no ammo, because I don't know the answer to the question this comment poses: 
> 
> ```
> 
> value = PEntity->RATT(SKILL_MARKSMANSHIP); // TODO: does this edge case exist? will mobs or trusts hit this?
> 
> ```
> 
> 
> 
> I felt it was safer not to throw an error in the event I was unaware of a strange corner case.

I believe the preferred option for such cases is to at least log a message so it doesn't go unnoticed.

zach2good (approved):

zach2good:
Squashing

Issue #2502, opened 2022-08-11
🐛+🔨  Multiple misuses of xi.mod.DELAY and a general lack of how-to


TeoTwawki:
- [x] I have paid attention to this example and will edit again if need be to not break the formatting, or I will be ignored
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md)

## Explanation
 _Yeah, I am violating the template(s) a bit here._ :stuck_out_tongue:

### Short version:
`setDelay()` takes milliseconds as input instead of a weapon delay value. `xi.mod.DELAY` ~~is a percentage adjustment.~~ _(as of [#2526](https://github.com/LandSandBoat/server/pull/2526) DELAYP will be used for percent adjustments, while DELAY needs set up to use standard delay units)_ And the weapon delays in the database are not milliseconds. 240 is not 240ms nor 2400ms. Unification of everything using just standard delay values is desired. **You want 240 delay, you input 240. No conversions.**

tl;dr we need better functionality for the scripters, and some existing scripts need corrections.

### Long version:
Several PRs brought in some changes that were simply incorrect. People have difficulty gauging a mobs actual delay, and the way we hand delay changing in project is a chaotic mix of various kludges that leave the average user unable to determine what they should be doing. I already know, and even I have trouble following that code end to end. End users are usually watching some youtube video and trying to time things. This more often than not makes it even worse for a whole slew of reasons that would take this issue way off topic. Hopefully in the near future an addon I am fiddling with will make it easier for them.

In these wrong scripts, authors have assumed that setting the mod directly adds, subtracts, or overwrites the mobs delay. **None of that is true.** The modifier is in fact a percentage change. So when the ladybug mixin says -400, the ladybug is receiving a 400% attack speed **increase** that is clearly not retail.

- That mixin is not the only place to incorrectly use the modifier. A helper lua in sealions den sets it to zero. Goblin Wolfman uses the mod in the amount of 1500 which probably (needs checked) should have been a haste mod (NOT haste _effect_ mind you).
[...]

PR #7500, opened 2025-04-27
Update core clocks & time usage

### LINKED ISSUE #4517: 🐛 [Conquest] Tally results differ in chat & in Region map
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

After todays tallying I noticed a discrepancy between what the system chat said and what is on the Region map.
See how Sandy and Windy ranks are different.

Skerx_2023 09 17_171417
Skerx_2023 09 17_171430

## Expected behavior

They should be the same rankings.

cocosolos:
**_I affirm:_**

This switches the server clock from the system clock to a steady clock.

> Tasks: 0ms, Network: -2552ms, Total: -2551ms, Diff/Sleep: 2751ms

The system clock is susceptible to changes caused by [NTP](https://en.wikipedia.org/wiki/Network_Time_Protocol). This means that the clock can occasionally move backwards. We do not want that for measuring durations, so we use the steady clock. This clock only ever moves forward, but the actual value of the time point is meaningless and inconsistent between reboots. So we use the steady clock to measure durations while the process is running, and use the system clock to persist a point in time to the database (or send it to Lua). The system clock is also used when we need a specific real world time, such as the current UTC or JST time.

Two helper namespaces have been added. The `::timer` namespace is a wrapper around the steady clock and has utilities to convert the steady time to and from UTC time, and for counting the seconds or milliseconds in a given duration. The `::earth_time` namespace is a wrapper around the system clock and contains namespaces for UTC, JST, and local timezones which all have helper functions for getting time information (such as current hour, minute, second, day of the week, etc). earth_time also has helper functions to return a Unix timestamp, a Vana'diel timestamp in Earth seconds since the Vana'diel epoch, the current "game" weekday (Monday-Sunday), and a time point for the start of the next "game" week (Monday at midnight JST).

A Vana'diel clock class has been added to the `::xi` namespace. This allows us to use actual Vana'diel time durations directly, down to the Vana'diel millisecond. A `::vanadiel_time` namespace has been added to help working with this. It has helper functions to convert the Vana'diel time to/from Earth times/durations as well as the same time information helpers the earth_time namespaces have, as well as namespaces for moon and RSE information. Care must be taken when working with Vana'diel time and converting to Earth/timer time and durations (`std::chrono::seconds()` and `""s` works for earth_time and timer durations, vanadiel_time uses `xi::vanadiel_clock::seconds()`).

With all of this in place, I went through and replaced every instance I could find of an integer being used as a time value with the actual time types. This means it is no longer necessary to manually convert between e.g. milliseconds and seconds by multiplying and dividing. Internally everything is either a [time point](https://en.cppreference.com/w/cpp/chrono/time_point) or [duration](https://en.cppreference.com/w/cpp/chrono/duration) and can use the helper namespaces or the [std chrono library](https://en.cppreference.com/w/cpp/chrono). When an integer is needed such as when reading/writing to the database, Lua, or packets, only then should these time types be floored to the appropriate duration and converted to an integer representation.

There are 2 exceptions that were not converted. Gardening furnishing items read/write a Vana'diel timestamp from their exdata. This doesn't get saved into any kind of gardening class or struct so it's fine to just stay as an integer. The other exception is weapon/base delay. These are eventually converted into time durations, but they remain integers until after all the calculations are done. See #2502, it seems there's still some unknowns and what we have is best guess so I'm not touching that.

Other than just converting, I was able to clean up the logic in some places and made some changes where I thought could be improved. This includes rearranging how some of the hourly ticks happen (like RoE). I don't think there is any negative impact, but there is a lot here and it's always possible I missed something. I made very few changes to scripts, but did change a few things that were using the Vana'diel day (1-360) when the days since epoch is more appropriate. I also cleaned up a few quests/missions that were using extra char vars and complicated conversions to count days (`VanadielTime()` is basically the Vana'diel equivalent of `os.time()` and should be used with the vanaTime enums).

- (Temporarily) enable [performance logging](https://github.com/LandSandBoat/server/blob/f56d96f5ecf3f45c7cb871e522d37650e464c658/settings/default/logging.lua#L59) to confirm values make sense.
- Try the `!time` command in game.

image

- Confirm everything matches https://www.pyogenes.com/ffxi/timer/v2.html and retail

Lot's of stuff to test:
- [x] Vana'diel time Lua bindings (!time)
- [x] Status effects
- [x] Attack delays
- [x] Spells cast and recast, interrupts
- [x] Death homepoint time
- [x] Mob pathing and combat
- [x] Conquest (fixes #4517)
- [x] /playtime
- [x] Abilities recast
- [x] Charged abilities (sic, stratagems, quickdraw)
- [x] /heal ticks
- [x] Ranged attacks
- [x] EXP/CP chains
- [x] Watchdog
- [x] Petskills
- [x] BCNM
- [x] Guild pattern
- [x] Transport/elevators
- [x] Usable items
- [x] Weaponskills/skillchains
- [x] Drops and treasure pool
- [x] Rate limiting packets (/jump)
- [x] Raise/weakness
- [x] Synthesis
- [x] Fishing
- [x] Quests/missions
- [ ] Instances
- [ ] xi.mob.phOnDespawn (night/daytime only spawns)
- [ ] ???

There's probably more. It'll probably be obvious if something doesn't work. Still in the process of thoroughly testing but everything seems to be working right.


cocosolos:
Clang doesn't have time zone support until clang19 😭 (only up to 18.1.8 on the runner) 

Update: Linux builds fine on clang 18.1.3 so I changed the MacOS build to use clang 18.1.8 (`-DCMAKE_C_COMPILER=$(brew --prefix llvm@18)/bin/clang -DCMAKE_CXX_COMPILER=$(brew --prefix llvm@18)/bin/clang`) and it still fails.

zach2good:
> so I changed the MacOS build to use clang 18.1.8 

I know OSX is a pain in the ass and generally we run quite far behind for compilers and dependencies on all platforms, but I'd rather we stick to what is defaultly available from our CI images so we provide maximum ease of use and compatibility. Nobody wants proper C++20 and C++23 support more than I do ;_;

src/world/time_server.cpp
```diff
@@ -30,82 +30,105 @@
...
-    if (CVanaTime::getInstance()->getJstWeekDay() == 1 && CVanaTime::getInstance()->getJstHour() == 0 && CVanaTime::getInstance()->getJstMinute() == 0)
+    /**
+     * Earth-based ticks.
+     * Uses the JST equivalent of the current timer tick. (steady_clock -> system_clock)
+     */
+
+    // Earth time points
+    earth_time::time_point jstTime    = earth_time::time_point(timer::to_utc(tick));
+    auto                   jstHour    = earth_time::jst::get_hour(jstTime);
+    auto                   jstWeekday = earth_time::jst::get_weekday(jstTime);
```
zach2good: these can all be `const auto`

src/world/time_server.cpp
```diff
@@ -30,82 +30,105 @@
...
+
+    // Static variable for the next tick
+    static earth_time::time_point nextHourlyTick = std::chrono::ceil<std::chrono::hours>(jstTime);
```
zach2good: `static const auto`

src/world/time_server.cpp
```diff
@@ -30,82 +30,105 @@
...
-            lastTickedJstMidnight = tick;
         }
+
+        nextHourlyTick = std::chrono::ceil<std::chrono::hours>(jstTime);
     }
 
-    // 4-hour RoE Timed blocks
-    static time_point lastTickedRoeBlock = tick - 1h;
-    if (CVanaTime::getInstance()->getJstHour() % 4 == 0 && CVanaTime::getInstance()->getJstMinute() == 0)
+    /**
```
zach2good: Can we use blocks of // comments instead of these /**/ comments please?

src/world/time_server.cpp
```diff
@@ -30,82 +30,105 @@
...
+     * Vana'diel-based ticks.
+     * Uses the Vana'diel time equivalent of the current JST time. (steady_clock -> system_clock -> vanadiel_clock)
+     * Note: Vana'diel minute is equal to the tick interval (2400ms). It is possible to miss a minute if there is
+     *       variance in the tick time.
+     */
+
+    // Vana'diel time points
+    vanadiel_time::time_point vanaTime = vanadiel_time::from_earth_time(jstTime);
+    vanadiel_time::TOTD       vanaTotd = vanadiel_time::get_totd(vanaTime);
+    auto                      vanaHour = vanadiel_time::get_hour(vanaTime);
```
zach2good: `const auto`

src/world/world_server.h
```diff
@@ -27,6 +27,10 @@
 
 #include "http_server.h"
 
+static constexpr auto kTimeServerTickInterval = 2400ms;
```
zach2good: Do these need to be visible in the header? They were in an anon namespace in world_server.cpp because they were only needed there and wrapped to be inaccessible anywhere else

src/common/vana_time.h
```diff
@@ -2,6 +2,7 @@
 ===========================================================================
 
   Copyright (c) 2010-2015 Darkstar Dev Teams
+  Copyright (c) 2025 LandSandBoat Dev Teams
```
zach2good, line 4: You've essentially completely rewritten this, you don't need to keep around the DSP attribution unless you want to

zach2good (approved):
Great work, just some general style points about comments and `const auto`, but the content is all great 👍

src/world/world_server.h
cocosolos: I moved this so I could reference it in the time server but I ended up changing that and didn't need it so I'll move it back.

cocosolos:
Rebased to include the battleutils changes in #7506 (simple change from static_cast to chrono::floor) + requested changes. I'm gonna say this is ready enough and if anything breaks just ping me and I'll fix it (xi.mob.phOnDespawn is the only thing I'm not really sure on and that's only because I don't understand what was going on there before, but based on what I think it should be doing it should work right).

src/world/time_server.cpp
cocosolos: This is updated during ticks so went with `static auto`.

src/world/world_server.h
cocosolos: Moved this back but kept the name change so it matches the map constants.

zach2good:
Nice work!

Yeet!

PR #7501, opened 2025-04-27
[Chore] Weaponskill misc cleanup

Xaver-DaRed:
**_I affirm:_**

Clean logic. Use existing new functions instead of old ones. And general maintenance prep work for the long promised weaponskill rewrite.

Use Weaponskills.
Review by commit.

### DIFF
--- a/scripts/globals/combat/physical_utilities.lua
+++ b/scripts/globals/combat/physical_utilities.lua
@@ -25,24 +25,24 @@ xi.combat.physical = xi.combat.physical or {}
 -----------------------------------
 local wsElementalProperties =
 {
-    -- [Skillchain type             ] = { None, Fire, Ice, Wind, Earth, Thunder, Water, Light, Dark },
-    [xi.skillchainType.NONE         ] = {    0,    0,   0,    0,     0,       0,     0,     0,    0 }, -- Lv0 None
-    [xi.skillchainType.TRANSFIXION  ] = {    0,    0,   0,    0,     0,       0,     0,     1,    0 }, -- Lv1 Light
-    [xi.skillchainType.COMPRESSION  ] = {    0,    0,   0,    0,     0,       0,     0,     0,    1 }, -- Lv1 Dark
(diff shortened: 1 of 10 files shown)


zach2good (approved):

scripts/globals/combat/physical_utilities.lua
```diff
@@ -267,112 +252,41 @@ xi.combat.physical.calculateFTP = function(actor, tpFactor)
...
-            wsElementalProperties[scProp2][xi.element.WIND] == 1 or
-            wsElementalProperties[scProp3][xi.element.WIND] == 1)
-        then
-            weaponFtpBonus = 0.1
-        end
     end
 
-    -- Add all bonuses and return.
-    fTP = fTP + neckFtpBonus + waistFtpBonus + headFtpBonus + handsFtpBonus + weaponFtpBonus
+    fTP = fTP + actor:getMod(xi.mod.ANY_FTP_BONUS) / 256
```
almuth150, line 289: After looking up the Fotia Gorget on various wikis, I believe this bonus should apply if the WS has *a* skillchain attribute. ie. it shouldn't work for ones without one, such as Spirit Taker. BGwiki is the most unclear about it, saying it works with any attribute, while the fandom wiki lists various WSs that don't work with it - all of which do not have any SC attributes (oddly enough, it does not list Spirits Within, though if you go to the individual page for it, it does make note that it doesn't work with any WS gorgets). The JP wiki is harder for me to decipher due to machine translation, but it does appear to specify that it works with a WS with a SC attribute.

https://ffxiclopedia.fandom.com/wiki/Fotia_Gorget
https://www.bg-wiki.com/ffxi/Fotia_Gorget
https://wiki.ffo.jp/html/32703.html

Xaver-DaRed: Line 267 is an early return. After it, the ability is guaranteed to have an skillchain/elemental property.
All this kind of gear needs the WS to have some kind of skillchain/elemental property, in truth.

PR #7516, opened 2025-05-01
Add six new item mod IDs

wccbuck:
**_I affirm:_**

This PR adds, but **does not yet implement**, the following item IDs, and assigns them to appropriate gear:
* ENHANCES_FUTAE (e.g. Hattori Tekko)
* ENHANCES_ELEMENTAL_SEAL (e.g. Laevateinn)
* ELEMENTAL_DEBUFF_EFFECT (e.g. Agwu's Slops)
* ENF_MAG_DURATION (e.g. Regal Cuffs)
* REWARD_RECAST (e.g. Khimaira Bonnet)
* AUGMENTS_ABSORB_TP (e.g. Bale Gauntlets +1)

In addition, this PR fixes two bugs I noticed while implementing these changes. Namely:
* NIN_NUKE_BONUS_GEAR (Ninjutsu MAB) is now appropriately applied to ninja spells (it was using the wrong mod before)
* One of the versions of Laevateinn had "quick draw dmg" instead of buffing Vidohunir damage, not sure how that happened

The new item mods aren't hooked up yet so no real way to test those, but you can confirm that ninja MAB gear is now working by performing a ninja nuke with and without Koga Hatsuburi equipped.


src/map/modifier.h
```diff
@@ -1081,16 +1090,12 @@ enum class Mod
...
-    AUGMENTS_ABSORB         = 1337, // Non-Liberator Absorb Spell potency +x% (Augments "Absorb" effects)
-    ABSORB_EFFECT_DURATION  = 1138, // Absorb Spell duration +% ("Absorb" effect duration +x%)
-
     // IF YOU ADD ANY NEW MODIFIER HERE, ADD IT IN scripts/enum/mod.lua ASWELL!
 
     // The spares take care of finding the next ID to use so long as we don't forget to list IDs that have been freed up by refactoring.
     // 570 through 825 used by WS DMG mods these are not spares.
     //
-    // SPARE IDs: 1148 and onward, but skip 1337 (AUGMENTS_ABSORB)
+    // SPARE IDs: 1154 and onward, but skip 1337 (AUGMENTS_ABSORB)
```
zach2good: AUGMENTS_ABSORB exists now, is that part of this comment still relevant?

zach2good: Oh I see now, why is it there, and why is it so high?

wccbuck: I rearranged some of the dark knight related augments from the bottom of the list up to the other dark knight stuff. I don't know why AUGMENTS_ABSORB was 1337 instead of 1137, either it was a mistake or somebody just wanted their AUGMENTS_ABSORB mod id to be leet

zach2good: Since it's all arbitrary and internal, can you hook it up to be in line with the other numbers, then we can get rid of this exception

wccbuck: Sure thing

Xaver-DaRed:
We don't usually allow adding random modifier IDs without them being actually implemented.
The main reason why is because the chances of them being forgotten and then being implemented down the line resulting in a duplicate is quite high.

I would propose this new, non-implemented, modifiers to be all placed at the end of the list, under a comment that states `This modifier isnt implemented` so they are all together and so we can easily track them, in case some1 decides to implement them in the future.

wccbuck:
> We don't usually allow adding random modifier IDs without them being actually implemented. The main reason why is because the chances of them being forgotten and then being implemented down the line resulting in a duplicate is quite high.
> 
> I would propose this new, non-implemented, modifiers to be all placed at the end of the list, under a comment that states `This modifier isnt implemented` so they are all together and so we can easily track them, in case some1 decides to implement them in the future.

That's fair. To give context, I am intending on adding functionality to all of these item mods over the next week, and decided to make this PR first in order to prevent a mod ID conflict which has happened before. Would it be sufficient for me to replace my "// TODO" with "// NOT YET IMPLEMENTED" on each of these?

Xaver-DaRed:
```lua
-- TODO: this mods are not implemented.
MOD_NAME_1 = n1
MOD_NAME_2 = n2
...
```
Then once you (or whoever else) decide to implement any of them, they can be moved to a better place.

Just a matter of organizing stuff properly, to make it clear not for a select few, but for as many ppl as possible within reason

Xaver-DaRed (approved):

PR #7522, opened 2025-05-02
[sql] add content tags to 25 base game/ROZ synthesis recipes

Grahf0085:
**_I affirm:_**

This PR adds content tags to 25 items added to the game before COP. I wasn't 100% on what to call the content tag. I know this content tag, no matter what it's called, what affect anything in the game. But anyone could use the content tag to delete recipes they don't want, they could add a a content setting for base/roz to main.lua to control the recipes being added to the game, and it will narrow down what recipes were added during other expansions.

The full list of base/roz recipes I'm tracking is at: https://cloud.atavismxi.com/s/wYnABz7RH6sdz3K

I made this list many years ago...... can't remember how exactly I made it.

Run dbtool and check content tags of the items in the database.


LacunaXI:
Why set something to Base/ROZ? Wouldn't it be one or the other?

zach2good (changes requested):
image

You didn't test this. When recipes are loaded the content tag is read in as a string, and then when we look up to see if a set of ingredients and their associated recipe are valid we run through this routine:

```cpp
    bool IsContentEnabled(const std::string& contentTag)
    {
        TracyZoneScoped;

        if (contentTag.empty())
        {
            return true;
        }

        std::string contentVariable("ENABLE_");
        contentVariable.append(contentTag);

        bool contentEnabled            = settings::get<bool>(fmt::format("main.{}", contentVariable));
        bool contentRestrictionEnabled = settings::get<bool>("main.RESTRICT_CONTENT");
        if (!contentEnabled && contentRestrictionEnabled)
        {
            return false;
        }

        return true;
    }
```

This would end up trying to look up a setting in Lua named: `main.BASE/ROZ` which is invalid.

```lua

local main = {
  -- Fine
  TEST = true,
  
  -- Fine
  BASE_ROZ = true,
  
  -- Invalid key:
  -- lua5.4: Main.lua:9: '}' expected (to close '{' at line 2) near '='
  -- BASE/ROZ = true,
}
```

zach2good:
Given that these are our settings for enabling/disabling content:
```lua
    -- Enable Expansion (1 = Enabled, 0 = Disabled)
    ENABLE_COP       = 1,
    ENABLE_TOAU      = 1,
    ENABLE_WOTG      = 1,
    ENABLE_ACP       = 1,
    ENABLE_AMK       = 1,
    ENABLE_ASA       = 1,
    ENABLE_ABYSSEA   = 1,
    ENABLE_SOA       = 1,
    ENABLE_ROV       = 1,
    ENABLE_TVR       = 1,
    ENABLE_VOIDWATCH = 1, -- Not an expansion, but has its own storyline. (Not Implemented)
```

What was the point in tagging these without adding the new setting? (which should probably just be called ENABLE_ROTZ, since we target the english client and BASE & ROTZ shipped together)

Grahf0085:
> Why set something to Base/ROZ? Wouldn't it be one or the other?

Which one do you want?

Grahf0085:
> Why set something to Base/ROZ? Wouldn't it be one or the other?

Which one do you want?

> Given that these are our settings for enabling/disabling content:
> 
> ```lua
>     -- Enable Expansion (1 = Enabled, 0 = Disabled)
>     ENABLE_COP       = 1,
>     ENABLE_TOAU      = 1,
>     ENABLE_WOTG      = 1,
>     ENABLE_ACP       = 1,
>     ENABLE_AMK       = 1,
>     ENABLE_ASA       = 1,
>     ENABLE_ABYSSEA   = 1,
>     ENABLE_SOA       = 1,
>     ENABLE_ROV       = 1,
>     ENABLE_TVR       = 1,
>     ENABLE_VOIDWATCH = 1, -- Not an expansion, but has its own storyline. (Not Implemented)
> ```
> 
> What was the point in tagging these without adding the new setting? (which should probably just be called ENABLE_ROTZ, since we target the english client and BASE & ROTZ shipped together)

I think we're in a much better spot now. I just tried to make a workbench with `RESTRICT_CONTENT = 1` and `ENABLE_ROTZ = 0` and it didn't let me make a workbench. Then I set` ENABLE_ROTZ = 1` and it let me make a workbench.

PR #7526, opened 2025-05-04
[core] Implement ENHANCES_MONSTER_CORRELATION

wccbuck:
**_I affirm:_**

This PR adds and implements the mod "ENHANCES_MONSTER_CORRELATION", which, when applied to a pet, grants +X accuracy and +X% damage against enemies belonging to a weaker opposing ecosystem. This mod can be seen on equipment such as [Nukumi Cabasset](https://www.bg-wiki.com/ffxi/Nukumi_Cabasset).

The PR also does a little bit of cleanup in mod.lua and modifier.h, moving two recently implemented mod IDs out of the "unimplemented" section.

```
!changejob BST 99
!additem lucky_broth (Faithful Falcorr, bird pet)
!additem nukumi_cabasset (enhances monster correlation +22)
!zone [Den of Rancor] (you should end up by the cloister home point)
```

Use the jug pet and have it fight a pugil. Then, equip the Nukumi Cabasset. Your pet's accuracy should increase by 22, and its attack should increase by 22%. Perform the test again, this time on a non-aquan enemy, and you should see no change in your pet's acc and attack with or without the helmet. You can test this by adding some ShowDebug()s in ACC() and ATT() within battleentity.cpp.


src/map/entities/battleentity.cpp
```diff
@@ -1077,7 +1096,25 @@ uint16 CBattleEntity::ACC(uint8 attackNumber, uint16 offsetAccuracy)
...
 
+            if (getEcoStrBonusFunc.valid())
+            {
+                CBattleEntity* thisTarget = nullptr;
+                if (this->PAI->IsEngaged())
+                {
+                    thisTarget = this->GetBattleTarget();
+                }
+                if (
+                    thisTarget != nullptr &&
```
zach2good: While this is the formatting we use in Lua, we don't use this in C++-land, these conditions can go on single lines, and there can be a break in between the two if statements here

zach2good (approved):

wccbuck:
Well, I tried to squash through a merge with base and ended up with the biggest PR in existence. I was just trying to fix the impending ID conflict after Xaver's recent fix but here we go.

Will try to fix.

Xaver-DaRed (approved):

PR #7540, opened 2025-05-08
[fix] AH Announcements, AH timer issue

### LINKED ISSUE #7485: 🐛 AH announcements module not letting purchases proceed
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

- With AH announcements module enabled, attempt to buy an item off the AH (at any price, even well above stocked items price)
- Char receives "failed to buy" messages

CE reported this, I have not personally tried to reproduce it yet. 

There were some changes to the queries and the module was likely not updated. The common code needs to be refactored so it doesn't have to be maintained independently.

sruon:
**_I affirm:_**

2 for 1

- Reworks ah_announcement module to defer the bulk of the logic to auctionutils, so it stops breaking every other day. Closes #7485
  - auctionutils::PurchasingItems returns a boolean, indicating if a sale went through

- Fixes the check for user throttling when they check list of sales, as it is currently defaulting to the "Try again later" state
  - Since `m_AHHistoryTimestamp` is initialized to `timer::time_point::min()` (min value of a long long), the existing check no longer works.

Screenshot 2025-05-07 185929

Without modules enabled:
- Open Auction House
- Check Sales Status
- Ensure items for sale show up
- Try again during the 5s cooldown, ensure "Please try again" shows up
- Ensure items can be listed and purchased

With modules enabled:
- Ensure pagination on Sales Status works, along with the 1.5s cooldown
- Ensure a message is received when purchasing one of your own item

### DIFF
--- a/modules/custom/cpp/ah_announcement.cpp
+++ b/modules/custom/cpp/ah_announcement.cpp
@@ -23,6 +23,7 @@
 #include "map/item_container.h"
 #include "map/map_session.h"
 #include "map/zone.h"
+#include "utils/auctionutils.h"
 
 #include <functional>
 #include <numeric>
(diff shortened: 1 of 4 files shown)


cocosolos:
Oh good catch. Could also use `std::chrono::abs()` I think. Intuitively -5s duration is still a 5s duration but I guess you need to explicitly check that.

Edit: or `curTick > PChar->m_AHHistoryTimestamp + 5s` should work too. I looked over the rest of the uses of time_point::min() and it looks like everything else should be okay.

sruon:
Made the change based on your recommendation since that's easier to read and understand.

Xaver-DaRed (approved):

zach2good (approved):

PR #7541, opened 2025-05-08
[fix] Instance music overrides

### LINKED ISSUE #7527: 🐛 Instance music type mismatch
**_I affirm:_**

## Branch affected by issue 

`base`

## Steps to reproduce

Noticed map server throwing a lot of errors for instances with music overrides set to -1

```
  `music_day` smallint(3) NOT NULL DEFAULT '-1',
  `music_night` smallint(3) NOT NULL DEFAULT '-1',
  `battlesolo` smallint(3) NOT NULL DEFAULT '-1',
  `battlemulti` smallint(3) NOT NULL DEFAULT '-1',

INSERT INTO `instance_list` VALUES (5600,'shades_of_vengeance',56,79,30,127.000,-15.000,-303.000,0,-1,-1,-1,-1);
```

```
[05/03/25 21:38:15:075][map][error] Query: SELECT instance_name, time_limit, entrance_zone, start_x, start_y, start_z, start_rot, music_day, music_night, battlesolo, battlemulti FROM instance_list WHERE instanceid = ? LIMIT 1 (db::detail::ResultSetWrapper::get:399)
[05/03/25 21:38:15:075][map][error] ResultSetWrapper::get: SQL Error: Out of range value for column 'music_night' : value -1 is not in uint32_t range (db::detail::ResultSetWrapper::get:397)
[05/03/25 21:38:15:075][map][error] Key: music_night (db::detail::ResultSetWrapper::get:398)
```

I haven't looked more than that but I'm guessing this is setting all instances with such entries to the track 0 (aka ambient sounds)

https://github.com/LandSandBoat/server/blob/3ba3962aa757e61067e24c7581d4d871f1f4e1d2/src/map/instance.cpp#L102-L105

https://github.com/LandSandBoat/server/blob/base/src/map/instance.cpp#L320-L323

## Expected behavior

No override is set, defaults to zone music.

sruon:
**_I affirm:_**

Closes #7527

Get the music overrides as int16 and cast them to uint16 so the existing -1 values are correctly retrieved and understood to be equivalent to "no music override"

Pondered making deeper changes to the types but settled on bare minimum to restore previous functionality for now. 

!instance 7704
Hear Nyzul Isle music

### DIFF
--- a/sql/instance_list.sql
+++ b/sql/instance_list.sql
@@ -26,10 +26,10 @@ CREATE TABLE `instance_list` (
   `start_y` float(7,3) NOT NULL DEFAULT '0.000',
   `start_z` float(7,3) NOT NULL DEFAULT '0.000',
   `start_rot` tinyint(3) unsigned NOT NULL DEFAULT '0',
-  `music_day` smallint(3) NOT NULL DEFAULT '-1',
-  `music_night` smallint(3) NOT NULL DEFAULT '-1',
-  `battlesolo` smallint(3) NOT NULL DEFAULT '-1',
-  `battlemulti` smallint(3) NOT NULL DEFAULT '-1',
+  `music_day` smallint(3) unsigned,
+  `music_night` smallint(3) unsigned,
+  `battlesolo` smallint(3) unsigned,
+  `battlemulti` smallint(3) unsigned,
   PRIMARY KEY (`instanceid`)
 ) ENGINE=Aria TRANSACTIONAL=0 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
 /*!40101 SET character_set_client = @saved_cs_client */;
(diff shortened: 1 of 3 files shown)


Xaver-DaRed (approved):

zach2good:
Since this system isn't so complex, I might ask you to fix it: If we aren't using an override, using -1 as the signifier is pretty lame. Surely we should be using a nullable unsigned int, so if there isn't an override it's set to `NULL` in SQL, and then represented internally as `xi::optional<uint16>`.

This is all static data, so no need for a migration etc.

sruon:
Made the columns nullable and changed the type of the overrides for instances.

Tried to introduce first class support for xi::optional in the ResultSetWrapper to avoid the ternaries but couldn't get the template quite right.

Tests:
- !instance 7702 (has music overrides)
- !instance 6000 (does not)

zach2good (approved):

PR #7542, opened 2025-05-08
[lua] [quest] Convert quest Your Crystal Ball to IF

slashtangent:
**_I affirm:_**

Converts the quest "Your Crystal Ball" into If format.
Adds missing interactions.

1. ``!zone <Lower Jeuno>``
2. Talk to Kurou-Morou ``!pos -4 -6 -28 245``
3. ``!zone <Maze of Shakhrami>``
4. ``!additem AHRIMAN_LENS``
5. Trade Ahriman lens with Rockwell ``!pos -18 -13 181 198``
6. Wait 1 IRL minute
7. Check Rockwell to receive DIVINATION_SPHERE
8. ``!zone <Lower Jeuno>``
9. Trade Divination Sphere to Kurou-Morou ``!pos -4 -6 -28 245``
10. Quest Complete

## Captures
https://youtu.be/-31BWZxWKaA
https://drive.google.com/drive/folders/1s2hyC0vGnvYvlJhDkShhnGGWfm3y5EvI?usp=sharing


scripts/quests/jeuno/Your_Crystal_Ball.lua
```diff
@@ -0,0 +1,116 @@
...
+    {
+        check = function(player, status, vars)
+            return status == xi.questStatus.QUEST_ACCEPTED
+        end,
+
+        [xi.zone.LOWER_JEUNO] =
+        {
+            ['Kurou-Morou'] =
+            {
+                onTrigger = function(player, npc)
```
KnowOne134, line 51: can be simplified for next time
```lua
onTrigger = quest:event(191),
```

scripts/quests/jeuno/Your_Crystal_Ball.lua
```diff
@@ -0,0 +1,116 @@
...
+                        if os.time() >= quest:getVar(player, 'Wait') + 60 then -- 1 minute wait time
+                            return quest:progressEvent(52)
+                        else
+                            return quest:messageSpecial(mazeID.text.WAIT_A_BIT_LONGER, 0, xi.item.DIVINATION_SPHERE)
+                        end
+                    end
+                end,
+
+                onTrade = function(player, npc, trade)
+                    local progress = quest:getVar(player, 'Prog')
```
KnowOne134: move this local inside the tradeHas scope

scripts/quests/jeuno/Your_Crystal_Ball.lua
```diff
@@ -0,0 +1,116 @@
...
+                        end
+                    end
+                end,
+            },
+
+            onEventFinish =
+            {
+                [52] = function(player, csid, option, npc)
+                    quest:setVar(player, 'Prog', 2)
+                    player:addItem(xi.item.DIVINATION_SPHERE)
```
KnowOne134: addItem of npcUtil becuase no message?
is there a param silent to use?
if not we need to add a inv check to make sure fits and set var inside those conditions

scripts/quests/jeuno/Your_Crystal_Ball.lua
slashtangent: From our discussion and testing, shorthand like this isn't currently supported in IF.

KnowOne134: its either an eden only addon that was released or its currently broken. unsure if needs a error report as could just be something for our server

scripts/quests/jeuno/Your_Crystal_Ball.lua
slashtangent: Correct there is no message. From the capture:
image

KnowOne134 (approved):

scripts/quests/jeuno/Your_Crystal_Ball.lua
```diff
@@ -0,0 +1,119 @@
...
+            },
+
+            onEventFinish =
+            {
+                [52] = function(player, csid, option, npc)
+                    if player:getFreeSlotsCount() ~= 0 then
+                        quest:setVar(player, 'Prog', 2)
+                        player:addItem(xi.item.DIVINATION_SPHERE)
+                    else
+                        return quest:messageSpecial(mazeID.text.ITEM_CANNOT_BE_OBTAINED, xi.item.DIVINATION_SPHERE)
```
Xaver-DaRed: imagen
I think this means "remove the return"

slashtangent: The return is required here. Top is with return, bottom is with it removed.
image

claywar: onEventFinish is a void func.  There's no return.  If your logic needs to exit out early, then the return should be after the messageSpecial and no value provided.

slashtangent:
Updated to remove os.time()

Xaver-DaRed (approved):

PR #7563, opened 2025-05-13
[MMM] Moblin Maze Mongers minor work

Xaver-DaRed:
**_I affirm:_**

Minor work to implement maze mongers while I capture.
- Add item enums.
- Add initial cutscene structure.
- Add guesswork logic for prize shop.

imagen

None.


scripts/globals/moblin_maze_mongers/mmm_shop.lua
```diff
@@ -0,0 +1,78 @@
+-----------------------------------
+-- Moblin Maze Monglers: Shop
```
zach2good: Mongers vs Monglers in a few places in this PR

scripts/globals/moblin_maze_mongers/mmm_shop.lua
```diff
@@ -0,0 +1,78 @@
...
+-- "Take my gil" menu.      -->    7
+
+local optionTable =
+{
+    [35009] = { xi.item.MAZE_VOUCHER_02, 1, 100, 'moblin_marble' },
+    [ 2241] = { xi.item.MAZE_VOUCHER_09, 1, 100, 'moblin_marble' },
+}
+
+xi.mmm.shopOnTrigger = function(player)
+    if player:hasKeyItem(xi.ki.TATTERED_MAZE_MONGER_POUCH) then
```
zach2good, line 26: If you want to do this in chunks and merge into base, introduce a new setting and have it off by default - then you can drip in work as you want/need to, like I've done with Chocobo Raising

zach2good (approved):

scripts/globals/moblin_maze_mongers/mmm_shop.lua
Xaver-DaRed: I created a setting. All this content is gated behind a KI, so I limited the setting to affect getting it.

PR #7577, opened 2025-05-19
Adds several item enums.

clanofartisans:
**_I affirm:_**

Adds several items to the item.lua enums.

n/a


Xaver-DaRed:
https://github.com/LandSandBoat/server/pull/7570

Also, when it came to shops, the ones left without enums are left that way to signify "This needs to be confirmed in retail"

clanofartisans:
> #7570
> 
> Also, when it came to shops, the ones left without enums are left that way to signify "This needs to be confirmed in retail"

Ah, okay! Should the values be set to the "no fame" prices, which will be the highest value, then? It looks like the Patolle NPC I checked as an example has the minimum values (max fame) set in LSB code. Also, do these need full, formal captures as well? Or just checked?

image

Xaver-DaRed:
No fame isnt enough. no fame AND outsiders and/or the nation being on a lower rank raise the price even more

And a screenshot of retail is needed to verify price, items and item order

clanofartisans:
> No fame isnt enough. no fame AND outsiders and/or the nation being on a lower rank raise the price even more
> 
> And a screenshot of retail is needed to verify price, items and item order

Perfect, thank you!

Xaver-DaRed:
> > #7570
> > Also, when it came to shops, the ones left without enums are left that way to signify "This needs to be confirmed in retail"
> 
> Ah, okay! Should the values be set to the "no fame" prices, which will be the highest value, then? It looks like the Patolle NPC I checked as an example has the minimum values (max fame) set in LSB code. Also, do these need full, formal captures as well? Or just checked?
> 
> image

This is what I meant with how fameless isnt enough

imagen

clanofartisans:
> > > #7570
> > > Also, when it came to shops, the ones left without enums are left that way to signify "This needs to be confirmed in retail"
> > 
> > 
> > Ah, okay! Should the values be set to the "no fame" prices, which will be the highest value, then? It looks like the Patolle NPC I checked as an example has the minimum values (max fame) set in LSB code. Also, do these need full, formal captures as well? Or just checked?
> 
> This is what I meant with how fameless isnt enough
> 


Oh, nice, thanks for the example! My goal was just to add some things to the enum but figured I'd capture some vendors while I was at it, especially after your first comment. As always, it looks like this is going to end up a little more labor intensive than I'd planned lol

clanofartisans:
I haven't had a chance to do my own captures yet but it looks like https://github.com/LandSandBoat/server/pull/7581 removed the need for them in the files I had been working on anyway. I don't know if @Xaver-DaRed would be willing to take a fresh look just to merge some item enums?

clanofartisans:
It looks like https://github.com/LandSandBoat/server/pull/7635 fixed the Barvira ID so I'll rebase mine to avoid a merge conflict.

Xaver-DaRed (approved):
IDs are good. Rebase for the Barvira thing and it's gtg

PR #7589, opened 2025-05-25
[Mobs] Create flying wyrm mixin and clean flight logic

Xaver-DaRed:
**_I affirm:_**

Creates a mixin for flying wyrms flight logic, so we dont have the same logic repeated and coded diferently across the codebase

Fight flying lizards


TracentEden2:
Could this apply to Ouryu in Monarch Linn for The Savage as well?

Xaver-DaRed:
> Could this apply to Ouryu in Monarch Linn for The Savage as well?

It already does, I thought. The idea is for this to be applicable to all wyrms that fly. Which admitedly arent that many more than this ones.

imagen

This ouryu can be forced down with the mistmelt item, unlike the Ouryu in The Savage II

TracentEden2:
> > Could this apply to Ouryu in Monarch Linn for The Savage as well?
> 
> It already does, I thought. The idea is for this to be applicable to all wyrms that fly. Which admitedly arent that many more than this ones.
> 
> imagen
> 
> This ouryu can be forced down with the mistmelt item, unlike the Ouryu in The Savage II

Ah, maybe I am missing something but I only see changes in the PR to the Ouryu mob file for Ouryu Cometh and not to the Ouryu mob file for The Savage.

TracentEden2 (approved):
Looks good to me!

TracentEden2:
Actually, I do have two small comments. Would the logic changes mean that mobs like Tiamat could use Mighty Strikes while in the air? Also should we be calling `handlePhaseVariables` on roam ticks as well, since some damage can take place even before engage, like a first damage spell?

Xaver-DaRed:
I added a check just now to disable 2-houring while in mid-air.
2-hours already disabled changing phase after all
imagen

Xaver-DaRed:
I will get to this in the near future

Issue #7588, opened 2025-05-25
🔨 Mobskills: A binding to manipulate knockback values.


UmeboshiXI:
**_I affirm:_**
- [x] I understand that if I do not agree to the following points by completing the checkboxes my issue will be ignored.
- [x] I have read and understood the [Contributing Guide](https://github.com/LandSandBoat/server/blob/base/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/LandSandBoat/server/blob/base/CODE_OF_CONDUCT.md).
- [x] I have searched existing [issues](https://github.com/LandSandBoat/server/issues) to see if the issue has already been opened, and I have checked the commit log to see if the issue has been resolved since my server was last updated.

## Describe the feature

Will need more captures to 100% confirm but came across an article that states certain gear with movement penalties (Dusk gear for example) can stop the Amphiptere knockback aura from triggering on players. Will report back pending captures of this mechanic.

https://wiki.ffo.jp/html/22686.html

UmeboshiXI:
Can confirm. Quoting cocosolos on capture discord.

> -movement speed negates the knockback on Amphipteres. More than -10 is needed (3 or more pieces of dusk or any combination that brings you over -10 (for example, 2 dusk (-5 each) and 1 oily trousers (-4) works, or plumb boots (-20) alone). Will test on Lilith in a bit.


More info in thread:
https://discord.com/channels/443544205206355968/443893540922064896/1376405495845355661

sruon:
Now that we have knockback sorted in the action packet, we can take another look at this. 
Did you figure out the specifics of the mechanic wrt movement speed?

sruon:
I've been testing knockback moves for a few days and have come to the following conclusions:
- Knockback is a value 1-7 in the action packet (scale field)
  - This is already known and documented
  - Note that this _does not_ translate to yalms, read the documentation on the 0x0028 packet over at XiPackets
- Knockback reduction gear is a flat reduction. Wearing +2 reduction on a move with KB 5 leads to KB being set to 3.
- Having more reduction than the skill value effectively cancels knockback
- Evading the skill hit also evades the knockback (It's possible there are skills going against that rule but I haven't found any yet)
  - This was already confirmed by SE
- Being hit for 0 still processes the knockback
- Knockback reduction is processed individually, you can get hit with a knockback of level 3 while your mates are hit with 5
- It doesn't appear you can "resist" knockback
  - But I haven't tested enough to make that claim

[...]

PR #7590, opened 2025-05-26
[cpp] Lua binding for setKnockback

Frankie-hz:
**_I affirm:_**

Adds a lua binding for setKnockback for https://github.com/LandSandBoat/server/issues/7588

### DIFF
--- a/src/map/lua/lua_mobskill.cpp
+++ b/src/map/lua/lua_mobskill.cpp
@@ -124,6 +124,11 @@ uint8 CLuaMobSkill::getMobHPP()
     return m_PLuaMobSkill->getHPP();
 }
 
+void CLuaMobSkill::setKnockBack(uint8 knockback)
+{
+    m_PLuaMobSkill->setKnockback(knockback);
+}
+
 //======================================================//
 
 void CLuaMobSkill::Register()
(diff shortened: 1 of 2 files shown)


MowFord:
btw stumbled upon this when looking at the list of open PRs. You may want to check that this doesn't set the _global_ properties of the mobskill's knockback value. I ran into that with [this PR](https://github.com/LandSandBoat/server/pull/5292/files) and had to send a pointer to the copy of the mobskill (which doesn't happen until after `onMobskillCheck` is called and the targets are determined)

This may not be affected, but just thought i'd throw that out since it tripped me up for a bit.

Frankie-hz:
Thanks I'll take a look!

What seemed to be a simple implementation had turned out to be a mystery as to why this isn't working.

It's currently only working if the mobskills knockback in the sql is greater than 0. Using the binding on a skill that has a 0 in the sql doesn't seem to work and the m_Knockback is being set correctly.

What's odd is that this exact implementation is working for horizon and there are no code differences and even the cpp module horizon has for their rebase project works. Throwing me for a loop

MowFord:
> Thanks I'll take a look!
> 
> What seemed to be a simple implementation had turned out to be a mystery as to why this isn't working.
> 
> It's currently only working if the mobskills knockback in the sql is greater than 0. Using the binding on a skill that has a 0 in the sql doesn't seem to work and the m_Knockback is being set correctly.
> 
> What's odd is that this exact implementation is working for horizon and there are no code differences and even the cpp module horizon has for their rebase project works. Throwing me for a loop

Yea breakpoint when the knock back gets calculated would tell you if it's being set, but it sounds very similar to what I was experiencing trying to update skill chain properties on mob skills.

Edit: a breakpoint [here](https://github.com/LandSandBoat/server/blob/9e9f691cf11d626fd4daffd690895982ec4496fb/src/map/entities/battleentity.cpp#L2237). Thinking about it more it's likely the same issue i ran into in that other PR. You need to send the pointer of the skill to the lua contructor or when you call your `setKnockback` binding, you're just changing an ephemeral version of the PSkill from the mobskill state.

I could be completely wrong, as it's been quite a long time since i messed with that (also that mobskill skillchain properties PR never merged into lsb/base)

PR #7595, opened 2025-05-29
[Misc] Add mapping for survival guide mog tablet discount and fix Tenzen being unavailable during "Dawn"

Xaver-DaRed:
**_I affirm:_**

- Adds the logic to display the "free ride" text in survival guides. Doesnt actually implement mog tablets, obviously.
- The NPC fixes for Verena are confirmed in retail and proper, but the Tenzen fix will require an actual rewrite of the system to support mission stages checking, when more use-cases are found.

None.


zach2good (approved):

zach2good:
Relevant; if you wanted to pick up this design: https://github.com/LandSandBoat/server/pull/7365

PR #7605, opened 2025-06-01
update both stock tables for xi.shop.general rewrite

MowFord:
**_I affirm:_**

few oversights in the `local stock` table updates from https://github.com/LandSandBoat/server/pull/7553

visit the NPCs and see that they function properly before and after CoP progression thresholds

### DIFF
--- a/scripts/zones/Tavnazian_Safehold/npcs/Caiphimonride.lua
+++ b/scripts/zones/Tavnazian_Safehold/npcs/Caiphimonride.lua
@@ -10,9 +10,9 @@ local entity = {}
 entity.onTrigger = function(player, npc)
     local stock =
     {
-        16450, 2030, -- Dagger
-        16566, 9216, -- Longsword
-        17335,    4, -- Rusty Bolt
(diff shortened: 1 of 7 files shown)


Mortalelite:
I would say probably be better do like `xi.item.ITEMHERE` instead of been Item ID listed. Kinda like here https://github.com/LandSandBoat/server/pull/7581/commits/2277a111bd1b446fd07080596d2ce7137fa946ab Unless they want something different.

MowFord:
> I would say probably be better do like `xi.item.ITEMHERE` instead of been Item ID listed. Kinda like here [2277a11](https://github.com/LandSandBoat/server/commit/2277a111bd1b446fd07080596d2ce7137fa946ab) Unless they want something different.

Absolutely would be better, but I'm definitely not taking on that scope creep, just fixing an npc I broke

Xaver-DaRed (approved):

Xaver-DaRed:
Not using item enums on this is ok for the time being. It marks the shop as "Item order and prices still need to be audited"

PR #7607, opened 2025-06-02
Check-in Eden's `test/` folder

zach2good:
Bringing this in _completely unaltered_, so we can then operate on it freely afterwards.

Check-in Eden's `test/` folder

This is a non-working PR. This is mainly so we can get Eden's tests and have their testing API visible to us. There is still a lot of work needed in the core to be able to use this.

**_I affirm:_**

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,0 +1,39 @@
+file(GLOB_RECURSE test_sources CONFIGURE_DEPENDS *.cpp)
+
+add_executable(xi_test
+  ${test_sources}
+)
+
+
+find_package(Python3 COMPONENTS Interpreter)
+
+if (${Python3_FOUND})
+  # Generate Lua definition files from parsing bindings in C++ files
+  set(gen_lua_types_script ${CMAKE_SOURCE_DIR}/tools/generate_lua_types.py)
+  set(expected_output ${CMAKE_SOURCE_DIR}/test/scripts/definitions/generated/LuaSimulation.lua) # Just need to list at least one generated output
+  file(GLOB cpp_lua_bindings CONFIGURE_DEPENDS ${CMAKE_SOURCE_DIR}/test/lua_*.cpp)
+  file(GLOB h_lua_bindings CONFIGURE_DEPENDS ${CMAKE_SOURCE_DIR}/test/lua_*.h)
+
+  add_custom_command(
+    OUTPUT ${expected_output}
+    COMMAND ${Python3_EXECUTABLE} ${gen_lua_types_script}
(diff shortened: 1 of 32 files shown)


zach2good:
So, like I say: this is ready to merge as-is.

For anyone interested in picking this up or tinkering on this, here's a small list of the things that need to happen:
- Our CMake layout is somewhat different from Eden's, so all of that needs to change and we need to add this as a proper build target while linking the sources (or a lib, hint hint) from xi_map
- Eden auto-generates enums and we don't, I'd imagine we need to fill those in manually
- Eden has different styled/named bindings than us in some places, those need to be audited (but should be picked up by LLS)
- test/ needs to be hooked into LLS and other CI tooling
- Eden has renamed xi_map to xi_game, but I've never liked this choice. We need to name what's going on inside of xi_map (MapServer) to encapsulate all of the "game" logic. Eden has called this "engine", I think, but again, that doesn't feel right to me. Open to ideas. I started this sort of layout:
```cpp
    //
    // Accessors
    //

    auto networking() -> MapNetworking&;
    auto statistics() -> MapStatistics&;
    auto zones() -> std::map<uint16, CZone*>&; // g_PZoneList
    // gameState()
```

So, maybe `GameState` is the "world" or whatever? It's all very global at the moment, but that doesn't need to be fixed if we just link in all the cpp files where those globals are defined.
- Kind of implied as above, this is a VERY static usage of state, we're not going to run tests concurrently within a process, or even across multiple processes. Entirely single test at a time sequentially! (for now)
- The reporting and sinks are all good, they just need enough jiggling to work.
- The tests are good, would like to try and make our changes so they can remain untouched and we can share tests back and forth.
- `SimulationClient`, `SimulationWorld` (needs a new name, since xi_world "owns" world as a term), `TestChar`, `TestRunner`, `TestWrapper`, `PacketBuilder` are all sane ideas, would be good to keep those as-is.
- This API is great and very flexible for multiple players, etc.:
```
suite['Verify drop rates (Devil Manta)'] = function(world)
    local client, player = world:spawnPlayer({ zone = xi.zone.Cape_Terrigan })

...

    world:skipTime(20)
    world:tick()
```

zach2good:
Other points for the future:
- We want to inject/replace our random implementation with something pseudo-random during testing, so we can test maths/damage output/drop rates, etc.
- We need to be able to query zone alive-ness once a player is spawned into a zone (if that isn't already a test it should be one)
- In line with the other naming I've put into `MapServer`, maybe `MapState` is enough to encapsulate the game/world state and logic. This kind of lines up with my future plans of `MapExecutor` for splitting out workloads, etc.

Xaver-DaRed (approved):

PR #7611, opened 2025-06-03
Set up the Linkshell vendors as proper guild vendors, general cleanup

sruon:
**_I affirm:_**

- Sets up the linkshell vendors as proper guild vendors (open 12 to 20), mirroring their retail implementation. 
  - Actual stock is best guesses, she stopped selling to me after 25 new linkshell but _did not_ restock on the next day. Given the items limited usefulness as sellable items, I find it is probably acceptable for the time being.
- Updates Paunelie dialog to the correct one
- Replace linkshell item IDs with enums in several places
- Let linkpearls and pearlsacks hit the recycle bin. Only fully formed linkshells bypass it, after the client warns you twice through confirmation dialog with no extra flavor text.
- Load item flags when loading guild shop items
- Guild sell menu will no longer publish items with NOSALE flag. This is a very niche case but mirrors retail.
- Renamed a handful of IPC struct messages to better reflect intent

I'm aware this is introducing 2x C-style casts but I couldn't get dynamic_cast to work likely due to the convoluted lineage between the two types.

!pos -144.96 -3 3 230


src/map/utils/guildutils.cpp
```diff
@@ -82,9 +82,10 @@ namespace guildutils
         }
         for (auto* PGuildShop : g_PGuildShopList)
         {
-            fmtQuery = "SELECT itemid, min_price, max_price, max_quantity, daily_increase, initial_quantity \
-                    FROM guild_shops \
-                    WHERE guildid = %u \
+            fmtQuery = "SELECT gs.itemid, gs.min_price, gs.max_price, gs.max_quantity, gs.daily_increase, gs.initial_quantity, ib.flags \
+                    FROM guild_shops gs \
+                    JOIN item_basic ib ON gs.itemid = ib.itemid \
+                    WHERE gs.guildid = %u \
```
zach2good: While you're here, can you change this query to use `db::preparedStmt` and change the multi-line query with trailing backslashes to just be multiple string literals:
```
"line 1 " // note the space
"line 2 "
"final line" // note the no space, and no semicolon

File: src/map/ipc_client.cpp [hunk 2]
```diff
@@ -636,17 +636,17 @@ void IPCClient::handleMessage_LinkshellRemove(const IPP& ipp, const ipc::Linkshe
...
         {
-            PChar->PLinkshell1->RemoveMemberByName(message.victimName, (targetLS->GetLSType() == (uint8)LSTYPE_LINKSHELL ? (uint8)LSTYPE_PEARLSACK : message.linkshellType));
+            PChar->PLinkshell1->RemoveMemberByName(message.victimName, message.changerRank);
         }
     }
     else if (PChar && PChar->PLinkshell2 && PChar->PLinkshell2->getID() == message.linkshellId)
     {
         CItemLinkshell* targetLS = (CItemLinkshell*)PChar->getEquip(SLOT_LINK2);
-        if (targetLS && (message.linkshellType == LSTYPE_LINKSHELL || (message.linkshellType == LSTYPE_PEARLSACK && targetLS->GetLSType() == LSTYPE_LINKPEARL)))
+        if (targetLS && (message.changerRank == LSTYPE_LINKSHELL || (message.changerRank == LSTYPE_PEARLSACK && targetLS->GetLSType() == LSTYPE_LINKPEARL)))
```
zach2good: I don't know if it parses this way for you, but `changer` looks like a typo every time I see it. Is there a better word we could be using? "instigator", "requester", "executor"? Naming is hard

zach2good (approved):
Looking good, I didn't know this was all borked!

src/map/ipc_client.cpp
sruon: requester is probably what I would have used if there wasn't a "changerId" already. I'll make the changes.

sruon:
- Reviewed the other 2 NPCs for correct dialogs. Bastok one was already correct.
- Use prepared statements in guildutils. Retested a couple guild shops to ensure they loaded correctly.
- Aligned the IPC structs fields on "requester"

Should be good to go.

zach2good (approved):

PR #7613, opened 2025-06-03
Refresh potency affects battery charge

wccbuck:
**_I affirm:_**

When a character casts the blue mage spell Battery Charge, the caster's "ENHANCES_REFRESH" mod (increased by gear such as [Amalric Coif](https://www.bg-wiki.com/ffxi/Amalric_Coif)) is added to the spell's potency.

`!changejob BLU 99`

Cast the spell Battery Charge with Amalric Coif equipped (`!additem 25615`), and see that your amount of MP recovered per tic goes from +3 to +4.


zach2good (approved):
Seems reasonable 👍 

But I've never really used BLU, so would like someone else to weigh in

Xaver-DaRed (approved):

PR #7615, opened 2025-06-04
fix: Set DE look before inserting in zone

sruon:
**_I affirm:_**

Ensures DE model is set before inserting into the zone. DEs "mobs-but-actually-NPCs" with custom models were spawning invisible to clients in Garrison..

I need to take the time to do a write up, this was a swiss cheese of a clusterfuck to debug.

Given the conditions required and the limited subset of content using DEs, this likely only affects Garrison.

!garrison start

### DIFF
--- a/src/map/lua/luautils.cpp
+++ b/src/map/lua/luautils.cpp
@@ -5448,6 +5448,18 @@ namespace luautils
             }
         }
 
+        // Look must be set _before_ inserting the entity into the zone,
+        // else equipped NPCs/Mobs will be invisible.
+        if (table["look"].get_type() == sol::type::number)
+        {
+            PEntity->SetModelId(table.get<uint16>("look"));
+        }
+        else if (table["look"].get_type() == sol::type::string)
+        {
+            auto lookStr  = table.get<std::string>("look");
+            PEntity->look = stringToLook(lookStr);
+        }
+
         if (auto* PNpc = dynamic_cast<CNpcEntity*>(PEntity))
         {
             PNpc->namevis     = table.get_or<uint8>("namevis", 0);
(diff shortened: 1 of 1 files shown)


zach2good (approved):

zach2good:
> this likely only affects Garrison

Yeah, I'm struggling to think of other areas where DEs are used on retail. Quasilumins in Sea and Garrison NPCs. I'm sure we could scrape captures to find "unexpected" names in the dynamic ID space to single out where else they might exist on retail, but we don't really gain much by doing that. DEs are absolutely still client abuse in the way that big downstream projects are using them, and they are fulfilling people's use cases by complete accident.

sruon:
For records purposes, as it relates to the general "Entity Update" mess and may provide context in the future:

When starting a Garrison with 18 NPCs (forced arbitrarily for testing), the first 8 NPCs would spawn and act but not be visible to the client, and the other 10 NPCs would be visible to the client.

Looking at a capture revealed this pattern

image

An invisible NPC
image

A visible NPC
image

The crux of the issue is that DE NPCs were inserted into the zone in one synchronous cycle without their look being set, which led the initial queued packets to not include the model flag and informations. 

But as soon as the player emitted a player update packet, the SpawnMOB of the associated zone updated all pending NPC update packets with the (now set) look information, which led to the on/off behavior. 

Another issue is that once an entity is stored in the spawnlist of a PC, the server will not emit full updates until the PC forces it to be evicted, by for example, walking beyond the max rendering range.

This never occurred with other DE, such as !fafnir because they load proper look data from the database to begin with, when the entity is being instantiated several blocks above. The issue likely started around the time zone entities were reworked and calls were shuffled around.

The client actually emits a very specific packet when facing a bogus entity, which are logged as "Invalid NPC" but not acted on. This packet handler should respond with a full update.

